PR #25, opened 2021-04-08
Encode message when disconnecting user

NullSoldier:
The message wasn't encoded when disconnecting users when you are at your max connections. This causes a crash because you can't send an object over WebSocket, you need to encode it to a string, or a buffer.

We'll look at cleaning up the code in another PR so that our wrapper connection only ever gets emitted from WebSocketServer, so this type of issue cannot happen again in the future.

Fixes IRO-635


dguenther (approved):
thanks for the test 👍

PR #34, opened 2021-04-08
Sync from same peer backwards

leanthebean:
The current block syncing algorithm is as follows: 

1) startup -> determine your heaviest head (e.g. the head of the heaviest chain connected to genesis) 
2) ask the network for a block after your heaviest head, e.g. ('here's _my_ heaviest head, give me the block(s) that come next)
3) if a responding peer gets a request it can't fulfill (most likely because the syncing peer is on a fork that lost), it sends back _its_ heaviest head 
4) the requesting peer sees this heaviest head, and then attempts to sync backwards from the newly given heaviest head until it connects to genesis 

The issue we're seeing is that nodes seem to stop syncing. A node gets into a weird state of requesting a block in the backwards direction, not getting a response (or getting the same response that makes it go in the same cycle)
The issue (i think) currently is that in step 4) the request goes out to a random peer from your peer manager list. This means that a syncing node has a high probability of randomly choosing to ask a peer that is _also_ syncing (and therefore doesn't have that block).Even after several retries, the syncing peer gets stuck thinking there are no more blocks to sync because every peer its asking is coming back with nothing.


ironfish/src/network/messages.ts
```diff
@@ -35,7 +35,7 @@ export type LooseMessage = Message<MessageType, PayloadType>
 export type Message<
   T extends MessageType,
   P extends PayloadType = undefined
-> = P extends undefined ? { type: T } : { type: T; payload: P }
+> = P extends undefined ? { type: T } : { type: T; payload: P; peer?: Identity }
```
NullSoldier, line 38: ???

ironfish/src/networkBridge.ts
```diff
@@ -102,43 +103,46 @@ export class NetworkBridge {
...
+
+        const serializedHash = this.node.captain.chain.blockHashSerde.serialize(hash)
+
+        const request: BlockRequest = {
+          type: MessageType.Blocks,
+          payload: {
+            hash: serializedHash,
+            nextBlockDirection: nextBlockDirection,
+          },
+          peer: peer,
```
NullSoldier, line 121: if we pass this as an argument to request, we won't need to modify the core messagetype to contain this

dguenther: related, I think this will end up sending the peer identity along with the message in `request`

ironfish/src/network/peerNetwork.ts
```diff
@@ -348,8 +349,10 @@ export class PeerNetwork {
    */
   async request(
     message: Message<MessageType, Record<string, unknown>>,
+    peer?: Identity,
```
NullSoldier, line 352: think this needs to get passed down to the requesat function below right?

ironfish/src/captain/messages.ts
```diff
@@ -0,0 +1,190 @@
...
+import { SerializedBlock } from './anchorChain/blockchain/Block'
+
+/**
+ * The type of a Iron Fish message. This is an exhaustive list of
+ * the messages that are sent from Captain. Other messages may
+ * be sent by peerNetwork's internal mechanisms (for example, a peer list).
+ *
+ * Note: A Response to a Request must have the same MessageType
+ */
+export enum MessageType {
```
dguenther, line 16: I think @NullSoldier merged this into `network/messages.ts` in #35

NullSoldier (approved):

PR #36, opened 2021-04-09
Move verifier from captain to consensus

NullSoldier:
Part of the great repo cleanup, strategy will be here too eventually. They'll both be deleted to form new modules that make up the protocol / consensus.


dguenther (approved):
Looks good, some lint failures to fix first

PR #41, opened 2021-04-10
Disconnect WebRTC connection if send throws an error

### LINKED ISSUE #16: WebRTC crash: RTCDataChannel.readyState is not 'open'
I've seen this error come up in Discord several times:

```
(node:1) UnhandledPromiseRejectionWarning: InvalidStateError: RTCDataChannel.readyState is not 'open'
    at RTCDataChannel._send (<anonymous>)
    at RTCDataChannel.send (/usr/src/app/node_modules/wrtc/lib/index.js:49:8)
    at Peer.send (/usr/src/app/node_modules/simple-peer/index.js:239:19)
    at WebRtcConnection.send (/usr/src/app/node_modules/ironfish/build/src/network/peers/connections/webRtcConnection.js:37:23)
    at Peer.send (/usr/src/app/node_modules/ironfish/build/src/network/peers/peer.js:259:47)
    at PeerManager.sendTo (/usr/src/app/node_modules/ironfish/build/src/network/peers/peerManager.js:452:21)
    at PeerManager.handleSignalMessage (/usr/src/app/node_modules/ironfish/build/src/network/peers/peerManager.js:805:18)
    at PeerManager.handleMessage (/usr/src/app/node_modules/ironfish/build/src/network/peers/peerManager.js:554:18)
    at /usr/src/app/node_modules/ironfish/build/src/network/peers/peerManager.js:428:18
    at Event.emitAsync (/usr/src/app/node_modules/ironfish/build/src/event.js:61:39)
```

IRO-591

dguenther:
The webRtcConnection would occasionally be in state `CONNECTED`, but then throw an exception when sending a message. This doesn't seem like a logic error in our code, since the connection must receive an Identity message from the opposite party before setting the connection to 'CONNECTED'.

This PR wraps the `send` in a try/catch and sets the connection to `DISCONNECTED` if it throws an error. I added some tests for it, and it falls back to WebSockets if available.

One issue here is it doesn't propagate the failure to send if the latency simulator is enabled, but that's only used for latency-related testing, so I'm less concerned about it.

Fixes #16

### DIFF
--- a/ironfish/src/network/peers/connections/webRtcConnection.ts
+++ b/ironfish/src/network/peers/connections/webRtcConnection.ts
@@ -127,18 +127,29 @@ export class WebRtcConnection extends Connection {
   /**
    * Encode the message to json and send it to the peer
    */
-  send = (message: LooseMessage): void => {
+  send = (message: LooseMessage): boolean => {
     if (this.shouldLogMessageType(message.type)) {
       this.logger.debug(`${colors.yellow('SEND')} ${this.displayName}: ${message.type}`)
     }
 
     const data = JSON.stringify(message)
-    this.peer.send(data)
+    try {
+      this.peer.send(data)
+    } catch (e) {
+      this.logger.debug(
+        `Error occurred while sending ${message.type} message in state ${this.state.type}`,
+        e,
+      )
+      this.close(e)
+      return false
+    }
 
     // TODO: Switch network traffic to binary
     const byteCount = Buffer.from(data).byteLength
     this.metrics?.p2p_OutboundTraffic.add(byteCount)
     this.metrics?.p2p_OutboundTraffic_WebRTC.add(byteCount)
+
+    return true
   }
 
   /**
--- a/ironfish/src/network/peers/peer.ts
+++ b/ironfish/src/network/peers/peer.ts
@@ -361,26 +361,36 @@ export class Peer {
    * @param message The message to send.
    */
   send(message: LooseMessage): Connection | null {
-    if (this.state.type === 'CONNECTED') {
-      if (this.state.connections.webRtc?.state.type === 'CONNECTED') {
-        this.state.connections.webRtc.send(message)
-        return this.state.connections.webRtc
-      }
-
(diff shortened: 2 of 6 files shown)


ironfish/src/network/peers/connections/webRtcConnection.ts
```diff
@@ -127,18 +127,29 @@ export class WebRtcConnection extends Connection {
...
     if (this.shouldLogMessageType(message.type)) {
       this.logger.debug(`${colors.yellow('SEND')} ${this.displayName}: ${message.type}`)
     }
 
     const data = JSON.stringify(message)
-    this.peer.send(data)
+    try {
+      this.peer.send(data)
+    } catch (e) {
+      this.logger.debug(
```
NullSoldier, line 139: Since we're now ignoring all errors, is it possible to lead to bugs we don't know about now?

ironfish/src/network/peers/peer.ts
```diff
@@ -361,26 +361,36 @@ export class Peer {
...
-      this.logger.warn(
-        `${this.displayName} is in CONNECTED state but has no active connections, dropping message of type ${message.type}`,
-      )
-
-      return null
     }
 
     this.logger.debug(
-      `Tried to send a message ${message.type} to ${this.displayName} in state ${this.state.type}`,
+      `Sending ${message.type} message to ${this.displayName} failed on all connections, dropping message`,
```
NullSoldier, line 393: It prints that it failed on all connections, even if there are no connections right?

NullSoldier (approved):

ironfish/src/network/peers/peer.ts
dguenther: It should get caught by the initial `CONNECTED` check and return early if there are no connections. I'll just remove the `on all connections` part

ironfish/src/network/peers/connections/webRtcConnection.ts
dguenther: I'll follow up and make a more specific PR to try to catch this specific error

PR #44, opened 2021-04-11
Allow setConfig to parse string arrays

### LINKED ISSUE #40: config:set bootstrapNodes
"Could not convert % to array" error, regardless of how the command is formatted.
I also tried without quotation marks (not pictured below).
image

dguenther:
Allows for updating the bootstrapNodes config with config:set, like so:

```
ironfish config:set bootstrapNodes "test.bn1.ironfish.network,examplenode.com"
```

Added an example for it to `ironfish config:set --help`.

Fixes #40﻿

### DIFF
--- a/ironfish/src/rpc/routes/config/uploadConfig.ts
+++ b/ironfish/src/rpc/routes/config/uploadConfig.ts
@@ -73,10 +73,23 @@ export function setUnknownConfigValue(
   config.set(sourceKey, value as any)
 }
 
+// Expects string in CSV format with no brackets
+function stringToStringArray(value: string): string[] | null {
+  if (value === '') return []
+
+  // Strip the brackets and split on commas
+  const parsedValue = value.split(',')
+
+  // Trim whitespace, trim leading/trailing quotes if necessary
+  return parsedValue.map((v) => v.trim())
+}
+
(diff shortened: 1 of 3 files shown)


ironfish/src/rpc/routes/config/uploadConfig.ts
```diff
@@ -73,10 +73,36 @@ export function setUnknownConfigValue(
   config.set(sourceKey, value as any)
 }
 
+function stringToStringArray(value: string): string[] | null {
+  if (!value.startsWith('[') || !value.endsWith(']')) {
+    return null
+  }
+
+  // special-case empty array
+  if (value === '[]') return []
```
NullSoldier, line 82: If we cut out the bracket checking, and make this work purely off CSV (but don't support escaping for now), it can work as a normal set too.

```sh
ironfish config:set bootstrapNode test.bn1.ironfish.network
```

or

```sh
ironfish config:set bootstrapNode test.bn1.ironfish.network,test.bn2.ironfish.network
```

dguenther: Sounds good, I took out some of the special-casing and added an example to `ironfish config:set --help`

NullSoldier:
Works great now, thank you. This is a nice improvement.

NullSoldier (approved):

PR #53, opened 2021-04-14
Remove NetworkBridge

NullSoldier:
The idea behind the network bridge, is that we'd move the Node entirely
inside of a worker, and then serialize messages across from the network,
to the worker, using the NetworkBridge. We've changed strategies and
are moving toward a WorkerPool approach, where we spin up workers
based on the number of cores, and move tasks that will freeze up the main
thread onto those workers.


ironfish/src/sdk.ts
```diff
@@ -227,7 +238,7 @@ export class IronfishSdk {
 }
 
 /** Get the current platform or null if it cannot detect the platform */
```
dguenther, line 240: This comment is wrong now

ironfish/src/mining/director.ts
```diff
@@ -147,6 +150,10 @@ export class MiningDirector<
         this.logger.error(err)
       })
     })
+
+    if (options.graffiti) {
+      this.setBlockGraffiti(options.graffiti)
+    }
```
dguenther, line 156: should we do this with minerAccount as well?

dguenther:
This includes several unrelated refactors, it'd be nice in the future if we could split those out into separate PRs.

ironfish/src/sdk.ts
```diff
@@ -157,7 +162,11 @@ export class IronfishSdk {
   }
 
   async node({ databaseName }: { databaseName?: string } = {}): Promise<IronfishNode> {
+    const webSocket = (await require('ws')) as IsomorphicWebSocketConstructor
+    const webRTC = (await require('wrtc')) as IsomorphicWebRtc | undefined
```
dguenther, line 166: This variable casing doesn't match that of `WebRtcConnection` and `IsomorphicWebRtc`

dguenther (approved):

ironfish/src/mining/director.ts
NullSoldier: lets do it... I actually just did that in the PR that removes captain from the mining director

ironfish/src/sdk.ts
NullSoldier: fixed

ironfish/src/mining/director.ts
NullSoldier: done

ironfish/src/sdk.ts
NullSoldier: fixed

ironfish/src/mining/director.ts
NullSoldier: fixed

PR #54, opened 2021-04-14
Move blockSyncer to node

NullSoldier:
This is in preparation for deleting captain. It moves BlockSyncer from
captain.blockSyncer to node.syncer. It requires the network, chain, and
strategy.


ironfish/src/node.ts
```diff
@@ -282,7 +288,7 @@ export class IronfishNode {
   }
 
   onPeerNetworkReady(): void {
-    this.captain.onPeerNetworkReady()
+    void this.syncer.start()
 
     // this.captain.blockSyncer.treesSynced &&
```
dguenther, line 293: I think this could be deleted too if you want to delete the commented code

dguenther (approved):

ironfish/src/node.ts
NullSoldier: Fixed in another PR

PR #55, opened 2021-04-14
Remove captain from mining director

NullSoldier:
This removes captain usage from mining director, and moves the mining
event it had on it, into the director itself.

### DIFF
--- a/ironfish/src/mining/director.test.slow.ts
+++ b/ironfish/src/mining/director.test.slow.ts
@@ -28,6 +28,7 @@ import {
 } from '../captain/testUtilities'
 import { MemPool } from '../memPool'
 import { WorkerPool } from '../workerPool'
+import { Assert } from '../assert'
 
 // Number of notes and nullifiers on the initial chain created by makeCaptain
 const TEST_CHAIN_NUM_NOTES = 40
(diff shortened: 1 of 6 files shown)


ironfish/src/mining/director.test.slow.ts
```diff
@@ -51,7 +52,7 @@ describe('Mining director', () => {
   let targetSpy: jest.SpyInstance
   let targetMeetsSpy: jest.SpyInstance
   let isAddBlockValidSpy: jest.SpyInstance
-  let memPool: TestMemPool
+  let mempool: TestMemPool
```
dguenther, line 55: I think if we want to do casing of the `mempool` variable this way (either is fine by me), the class should be cased `TestMempool` and `Mempool`

ironfish/src/mining/director.test.slow.ts
```diff
@@ -325,17 +320,13 @@ describe('Recalculating target', () => {
     jest.setTimeout(15000000)
     captain = await makeCaptain(strategy)
     memPool = new MemPool(captain)
```
dguenther, line 322: here's another differing memPool casing

dguenther (approved):

ironfish/src/mining/director.test.slow.ts
NullSoldier: reverted

ironfish/src/mining/director.test.slow.ts
NullSoldier: reverted

PR #58, opened 2021-04-15
Verify transactions in a worker thread

dguenther:
Creates a WorkerPool class that starts several worker threads and distributes long-running jobs to them. This should unburden the main thread from block verification, at least, and we can move the other async work into the new WorkerPool as well.

Fixes IRO-649


NullSoldier (approved):

NullSoldier:
Were gonna look at not passing in the pool to the constructor in another PR.

PR #64, opened 2021-04-15
Reject incoming transactions with negative fees

dguenther:
Incoming transactions with negative fees indicate miner rewards. They'll still pass proof verification, but they should only ever appear on blocks by the miners who created them, not gossiped via the NewTransaction message.

Fixes IRO-650﻿

### DIFF
--- a/ironfish/src/consensus/verifier.ts
+++ b/ironfish/src/consensus/verifier.ts
@@ -183,6 +183,9 @@ export class Verifier<
     } catch {
       return Promise.reject('Could not deserialize transaction')
     }
+    if ((await transaction.transactionFee()) < 0) {
+      return Promise.reject('Transaction has negative fees')
+    }
     if (!(await transaction.verify()).valid) {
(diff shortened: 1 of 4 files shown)


ironfish/src/consensus/verifier.ts
```diff
@@ -183,6 +183,9 @@ export class Verifier<
     } catch {
       return Promise.reject('Could not deserialize transaction')
     }
+    if ((await transaction.transactionFee()) < 0) {
+      return Promise.reject('Transaction has negative fees')
+    }
```
NullSoldier, line 188: note: this needs to move to the mem accept() function later

NullSoldier (approved):

ironfish/src/consensus/verifier.ts
dguenther: I'd like to leave it here as well in the future, but yep, we still need to prevent readding the transaction to the mempool and broadcasting it

PR #68, opened 2021-04-15
Improve memory usage of block syncer

dguenther:
The main fix here is to cap the size of blocksForProcessing, since it can grow unbounded. Also made a change to avoid allocating a new Serde class when functions are called, most notably the transactionSerde and the Serdes in BlockHeader.

- Limit new Serde instance allocations
- Limit the number of entries in blocksForProcessing

### DIFF
--- a/ironfish/src/blockSyncer.ts
+++ b/ironfish/src/blockSyncer.ts
@@ -26,6 +26,7 @@ import { Strategy } from './strategy'
 
 export const MAX_MESSAGE_SIZE = 500000 // 0.5 MB
 export const MAX_BLOCKS_PER_MESSAGE = 1
+export const MAX_GOSSIPPED_BLOCKS = 200
 
 export const ALLOWED_TRANSITIONS_TO_FROM = {
   ['STARTING']: ['STOPPED'],
@@ -186,7 +187,8 @@ export class BlockSyncer<
     }
   >()
 
-  blocksForProcessing: BlockToProcess<E, H, T, SE, SH, ST>[]
+  gossippedBlocksForProcessing: BlockToProcess<E, H, T, SE, SH, ST>[]
+  requestedBlocksForProcessing: BlockToProcess<E, H, T, SE, SH, ST>[]
 
   logger: Logger
   peerNetwork: PeerNetwork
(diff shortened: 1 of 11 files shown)


ironfish/src/blockSyncer.ts
```diff
@@ -281,8 +283,14 @@ export class BlockSyncer<
       block.header.sequence <= this.blocksForProcessing[0].block.header.sequence
     ) {
       this.blocksForProcessing.unshift({ block, fromPeer, type })
+      if (this.blocksForProcessing.length > this.blocksForProcessingMax) {
```
NullSoldier, line 286: If we suddenly drop blocks here, what will cause these blocks to get re-requested? This seems like it's going to cause some issues syncing.

dguenther: Suggestions for improving this so the list of blocks to process doesn't grow unbounded? I think the answer to most of these is "Fix bugs in the syncer so it syncs"

leanthebean: hmm.. what if we additionally split out the queue to be for gossiped blocks (and we can do that with an lru that we have in other places of the codebase already) and for syncing blocks? if the memory is coming from gossiped blocks, we could put bounds around that specifically

ironfish/src/blockSyncer.ts
```diff
@@ -204,13 +205,14 @@ export class BlockSyncer<
...
 
-    this.hashSerde = new BufferSerde(32)
+    this.hashSerde = BlockHashSerdeInstance
     this.blockSerde = options.strategy.blockSerde
 
     this.blockSyncPromise = Promise.resolve()
     this.blockRequestPromise = Promise.resolve()
     this.recentBlocks = new LeastRecentlyUsed(500)
     this.blocksForProcessing = []
+    this.blocksForProcessingMax = 200
```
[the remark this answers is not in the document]
dguenther: Fixed this

ironfish/src/blockSyncer.ts
dguenther: Updated to add two queues

NullSoldier:
I did a bunch of testing on this PR, and the syncing. It looks like it stops the memory being unbounded, but gets stuck syncing in a predictable way. All you need to do to cause it to get stuck, is to set the max gossip pool to 1, to trigger it, then start syncing from another node with a miner enabled. If it gets done syncing, and there's anything in the gossip pool, not only will it not sync those blocks, but the original node will then keep trying to sync a block it already has from the new node in an infinite loop.

However, I'll still merge this. Our syncing is still very broken, and if people run it, at most they won't run out of memory when trying the node out.

NullSoldier (approved):

PR #74, opened 2021-04-17
Move AsyncTransaction into WorkerPool

dguenther:
Removes AsyncTransaction and associated files and adds `createTransaction` and `createMinersFee` to the workerPool.

I refactored createTransaction to pass in all the transaction-related state in one function call to avoid needing workers to hold state across multiple function calls like they did under AsyncTransaction.

Fixes IRO-651


NullSoldier (approved):
This looks like a great cleanup, thank you.

PR #86, opened 2021-04-21
Drop NewBlock gossip messages if connections have buffered messages

dguenther:
NewBlock gossip messages should be optional, though they may not be right now. If NewBlock messages are missed, our block syncer should be able to proactively detect when it's missing blocks and seek them out from peers.

Accomplishes two things:
* Decreasing memory consumption during high volume of NewBlock messages. The outgoing buffers for the WS and WebRtc connections were backing up with messages to forward.
* Improving network congestion -- this should indirectly prioritize sending other messages over NewBlock messages, which should at least prevent the case where a node is trying to respond to an RPC call, but the socket's buffer is backed up with NewBlock messages.

Fixes IRO-655


NullSoldier (approved):
Let's give it a shit

PR #92, opened 2021-04-22
Exchange chain info in networking handshake

NullSoldier:
This will now send new fields along in the networking handshake and will get stored on the peer on the other side.
 * Heaviest head hash
 * Heaviest head sequence
 * Heaviest head cumulative work

Version has also changed to just an integer, and agent is a new handshake message field that is now freeform entry, and not required to be in a specific format by their peers. Instead of forcing exact versioning, we now enforce that the peer version is at least `VERSION_PROTOCOL_MIN`

These 3 new fields are also now sent back in the getPeers() endpoint. They will be used during the new block syncer algorithm.


ironfish/src/network/peers/peerManager.ts
```diff
@@ -861,12 +862,10 @@ export class PeerManager {
       return
     }
 
-    if (!versionsAreCompatible(this.localPeer.version, version)) {
-      const error = `Peer version ${
-        message.payload.version
-      } is not compatible to ours: ${renderVersion(this.localPeer.version)}`
-
+    if (version < VERSION_PROTOCOL_MIN) {
+      const error = `Peer version ${message.payload.version} is not compatible to our minimum: ${VERSION_PROTOCOL_MIN}`
```
dguenther, line 866:
```suggestion
      const error = `Peer version ${message.payload.version} is not compatible with our minimum: ${VERSION_PROTOCOL_MIN}`
```

ironfish/src/network/messages.ts
```diff
@@ -86,20 +86,27 @@ export type Identify = Message<
...
   const payload = obj.payload as Identify['payload']
+
   return (
     obj.type === InternalMessageType.identity &&
     typeof payload === 'object' &&
     payload != null &&
     typeof payload.identity === 'string' &&
-    typeof payload.version === 'string'
+    typeof payload.agent === 'string' &&
+    typeof payload.version === 'number'
```
dguenther, line 109: I think this has a few more fields on it as well now

dguenther (approved):
Few minor change suggestions but looks good to me after those are fixed up

ironfish/src/network/messages.ts
NullSoldier: fixed

PR #93, opened 2021-04-23
Install wasm-pack from wasm-pack website

dguenther:
This speeds up builds significantly, since using `cargo install` compiles wasm-pack locally. It also might simplify the dependency installation process on Windows (not certain on that though).

### DIFF
--- a/ironfish-wasm/build.js
+++ b/ironfish-wasm/build.js
@@ -23,7 +23,7 @@ if(buildNode) {
     });
     if (result.error) {
         if (result.error.message.includes('ENOENT')) {
-            console.error('wasm-pack is not installed. See README.md for install instructions.')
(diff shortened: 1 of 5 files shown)


ironfish-wasm/build.js
```diff
@@ -23,7 +23,7 @@ if(buildNode) {
     });
     if (result.error) {
         if (result.error.message.includes('ENOENT')) {
-            console.error('wasm-pack is not installed. Run `cargo install wasm-pack`.')
+            console.error('wasm-pack is not installed. See README.md for install instructions.')
```
NullSoldier, line 26: WDYT about just telling them what to do here?

```suggestion
            console.error('wasm-pack is not installed. Install from https://rustwasm.github.io/wasm-pack')
```

NullSoldier (approved):

PR #94, opened 2021-04-23
Switch to larger bloom filter from evicting set

dguenther:
The EvictingSet's size of 1000 was too small to catch all duplicate gossip messages. I verified this by adding a second, larger evicting set and checking if they disagreed whether a nonce was in the set.

A bloom filter with an expected item size of 100000 and a false positive rate of 0.000001 should take up about 315KB (see this bloom filter calculator https://hur.st/bloomfilter/?n=100000&p=0.000001&m=&k=), and we're probably doubling that by using a rolling bloom filter.

I ran an implementation that used both the evicting set and the bloom filter for 10 minutes or so and didn't notice any situation where they disagreed on whether a nonce was in the set, which makes sense given a false positive rate of roughly 1 in a million.

Fixes IRO-659


NullSoldier (approved):
Tested locally, works great.

PR #103, opened 2021-04-27
Return Result from Wasm functions

dguenther:
Changes most Wasm functions to return `Result`s, which allows us to intercept Rust errors and return a JS error from Wasm. 

Also replaces console_error_panic_hook with a panic_hook that throws a more descriptive error directly in the case that something somewhere does `panic` or `unwrap` an Error. The previous error had the generic text `RuntimeError: unreachable executed`, but the message logged from console_error_panic_hook was more useful. In practice, this shouldn't be used often since it's common in Rust to return Result rather than panic or unwrap Results.

### DIFF
--- a/ironfish-rust/src/errors.rs
+++ b/ironfish-rust/src/errors.rs
@@ -23,7 +23,7 @@ pub enum SaplingKeyError {
 
 impl fmt::Display for SaplingKeyError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", self)
+        write!(f, "{:?}", self)
     }
 }
 
@@ -49,7 +49,7 @@ pub enum SaplingProofError {
 
 impl fmt::Display for SaplingProofError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", self)
+        write!(f, "{:?}", self)
     }
 }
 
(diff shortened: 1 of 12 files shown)


ironfish-rust/src/errors.rs
```diff
@@ -23,7 +23,7 @@ pub enum SaplingKeyError {
 
 impl fmt::Display for SaplingKeyError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", self)
+        write!(f, "{:?}", self)
```
NullSoldier, line 26: __Note for others__:
{} causes the `fmt::Display` trait to be used, {?} causes the `fmt::Debug` trait to be used for rendering `self`

https://doc.rust-lang.org/std/fmt/index.html#fmtdisplay-vs-fmtdebug

ironfish-wasm/src/lib.rs
```diff
@@ -59,17 +59,17 @@ pub fn create_key_to_js() -> Key {
 }
 
 #[wasm_bindgen(catch, js_name = "generateNewPublicAddress")]
-pub fn create_new_public_key_to_js(private_key: &str) -> Key {
-    console_error_panic_hook::set_once();
+pub fn create_new_public_key_to_js(private_key: &str) -> Result<Key, JsValue> {
```
NullSoldier, line 62: should JsValue be `WasmSaplingKeyError`?

NullSoldier: Looks like wasm-bindgen may not support this, even though WasmSaplingKeyError implements From<JsValue>

NullSoldier:
I tested this as manually as possible, just doing basic operations.

NullSoldier (approved):

PR #104, opened 2021-04-28
Sort the peers list by connection state by default

dguenther:
Adds a `sort` flag to the `peers:list` command that's set to sort the peers list by connection state by default. We'll probably want to do a more complex sort later -- I'm guessing this sorts lexicographically.

image

### DIFF
--- a/ironfish-cli/src/commands/peers/list.ts
+++ b/ironfish-cli/src/commands/peers/list.ts
@@ -10,6 +10,8 @@ import blessed from 'blessed'
 
 type GetPeerResponsePeer = GetPeersResponse['peers'][0]
 
+const STATE_COLUMN_HEADER = 'STATE'
+
 export class ListCommand extends IronfishCommand {
   static description = `List all connected peers`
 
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/peers/list.ts
```diff
@@ -10,6 +10,8 @@ import blessed from 'blessed'
 
 type GetPeerResponsePeer = GetPeersResponse['peers'][0]
 
+const StateColumnHeader = 'STATE'
```
NullSoldier, line 13: I think all the other files usually specify these as upper camel case with underscores, if you really want to pull this out.

```suggestion
const STATE_COLUMN_HEADER = 'STATE'
```

NullSoldier (approved):

NullSoldier:
Tested manually, LGTM

ironfish-cli/src/commands/peers/list.ts
dguenther: I'll change that, main reason I pulled it out was just to make sure if we rename it in one place it gets updated in the other

PR #105, opened 2021-04-28
Add enableSyncer config setting

dguenther:
Adds a config value to enable the block syncer, defaulted to true. When set to false, the node performs no action when receiving NewBlock and NewTransaction messages, and never sends block request messages to other nodes.

Further additions here would be to add a field to the IDENTIFY message to indicate whether the node syncs blocks, although with upcoming syncer changes, peers shouldn't be sending block requests to nodes without knowledge of which blocks they have.

This config value is only used for testing and may be removed in the future.

### DIFF
--- a/ironfish/src/fileStores/config.ts
+++ b/ironfish/src/fileStores/config.ts
@@ -23,6 +23,7 @@ export type ConfigOptions = {
   enableRpc: boolean
   enableRpcIpc: boolean
   enableRpcTcp: boolean
+  enableSyncing: boolean
   enableTelemetry: boolean
   enableMetrics: boolean
   getFundsApi: string
@@ -128,6 +129,7 @@ export class Config extends KeyStore<ConfigOptions> {
       enableRpc: true,
       enableRpcIpc: true,
       enableRpcTcp: false,
+      enableSyncing: true,
       enableTelemetry: false,
       enableMetrics: true,
       getFundsApi: DEFAULT_GET_FUNDS_API,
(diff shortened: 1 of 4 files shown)


ironfish/src/fileStores/config.ts
```diff
@@ -23,6 +23,7 @@ export type ConfigOptions = {
   enableRpc: boolean
   enableRpcIpc: boolean
   enableRpcTcp: boolean
+  enableSyncer: boolean
```
NullSoldier, line 26: this does more than enable or disable the syncer, but I guess it's okay for now to call it this. It just requires a caveat, that it really disables all syncing.

ironfish/src/network/peerNetwork.ts
```diff
@@ -190,23 +190,38 @@ export class PeerNetwork {
...
+        () => Promise.resolve(undefined),
+        () => Promise.resolve(undefined),
+      )
+      this.registerHandler(
+        NodeMessageType.NewTransaction,
+        RoutingStyle.gossip,
+        () => Promise.resolve(undefined),
+        () => Promise.resolve(undefined),
+      )
+    }
```
NullSoldier, line 224: Later, I assume we'll just undo this, remove the log when a message isn't handled, and handle the concept of "unknown" messages by increasing banscore inside of the peer network.

dguenther: Yep, I'm expecting we'll refactor all of this network code to be more specific to our protocol.

ironfish/src/fileStores/config.ts
dguenther: I'll rename this to `enableSyncing`

NullSoldier:
Tested this manually

* Started two nodes and node A `(enableSyncing=false)`, did not sync from nodeB
* Started two nodes with genesis block only, mined on nodeB, nodeA `(enableSyncing=false)` did not accept miner gossip messages

NullSoldier (approved):

PR #110, opened 2021-04-30
Creating ironfish-beta docker package

leanthebean:
Creating ironfish-beta docker package and github action script. This is so we can more easily test the staging branch.


NullSoldier (approved):

.github/workflows/deploy-node-github-beta.yml
```diff
@@ -0,0 +1,30 @@
+name: Deploy Node Image to Github
```
NullSoldier, line 1: Change this name

PR #118, opened 2021-05-05
Replace lru-cache with blru for custom Map support

NullSoldier:
Bcoin's LRU supports custom map type, and forking lru-cache and adding
that support doesn't seem worth it when we can just use bcoins blru.

https://github.com/bcoin-org/blru


dguenther (approved):
Looks like there's another use of this in `syncer` that'll be going away once we merge that

PR #119, opened 2021-05-05
Abort gossip by having handlers return boolean

NullSoldier:
Before there was no way to abort the gossip, now we can abort the gossip
by returning true or false from a gossip handler.

### DIFF
--- a/ironfish/src/network/peerNetwork.ts
+++ b/ironfish/src/network/peerNetwork.ts
@@ -40,6 +40,7 @@ import {
   isRpc,
   Rpc,
   CannotSatisfyRequestError,
+  IncomingGossipGeneric,
 } from './messageRouters'
 import { Peer } from './peers/peer'
 import { LocalPeer } from './peers/localPeer'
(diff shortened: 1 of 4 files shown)


ironfish/src/network/peerNetwork.ts
```diff
@@ -443,7 +445,10 @@ export class PeerNetwork {
 
     switch (style) {
       case RoutingStyle.gossip: {
-        this.gossipRouter.register(type, hdlr)
+        this.gossipRouter.register(
+          type,
+          hdlr as (message: IncomingGossipGeneric<T>) => Promise<boolean | void>,
```
dguenther, line 450: maybe `| boolean | void` on the end to match the other return types? or is that not needed now since onNewBlock is async

dguenther (approved):

ironfish/src/network/peerNetwork.ts
NullSoldier: Updated, added it so handlers in prod code or tests don't have to make it async if they don't need it to.

PR #123, opened 2021-05-07
Add --threads to miners:start command

dguenther:
Add a `--threads` option to the `miners:start` command to allow for mining on multiple cores without running separate instances of `miners:start`. The thread pool was already in place, just not configurable.

### DIFF
--- a/ironfish-cli/src/commands/miners/start.ts
+++ b/ironfish-cli/src/commands/miners/start.ts
@@ -1,7 +1,9 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { flags } from '@oclif/command'
 import cli from 'cli-ux'
+import os from 'os'
 import { miner, NewBlocksStreamResponse, PromiseUtils } from 'ironfish'
 import { IronfishCommand } from '../../command'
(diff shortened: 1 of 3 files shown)


ironfish-cli/src/commands/miners/start.ts
```diff
@@ -11,10 +12,23 @@ export class Miner extends IronfishCommand {
 
   static flags = {
     ...RemoteFlags,
+    threads: flags.integer({
+      char: 't',
+      default: 1,
```
NullSoldier, line 17: shouldn't this be -1 by default?

ironfish-cli/src/commands/miners/start.ts
```diff
@@ -11,10 +12,23 @@ export class Miner extends IronfishCommand {
...
 
   async start(): Promise<void> {
-    this.parse(Miner)
+    const { flags } = this.parse(Miner)
+
+    let threads = flags.threads
+    if (threads === 0 || threads < -1) {
+      throw new Error('--threads must be a positive integer or -1.')
+    } else if (threads === -1) {
+      threads = (await import('os')).cpus().length
```
NullSoldier, line 30: is there a reason to do a lazy import here, instead of at the top?

dguenther: Changed

dguenther:
Should I do `workers` for this too?

NullSoldier (approved):

PR #124, opened 2021-05-07
Add --workers and nodeWorkers config

dguenther:
Adds a --workers flag to the `start` command, as well as a `nodeWorkers` config value. This sets the number of workers in the worker pool to create -- since each worker requires a few hundred MB of memory, this can be useful for reducing memory use on systems with many CPU cores.

### DIFF
--- a/ironfish-cli/src/commands/start.ts
+++ b/ironfish-cli/src/commands/start.ts
@@ -49,6 +49,10 @@ export default class Start extends IronfishCommand {
       char: 'p',
       description: 'port to run the local ws server on',
     }),
+    workers: flags.integer({
+      description:
+        'number of CPU workers to use for long-running operations. 0 disables (likely to cause performance issues), -1 auto-detects based on CPU cores',
+    }),
     name: flags.string({
       char: 'n',
       description: 'name for the node',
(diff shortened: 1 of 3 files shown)


ironfish/src/fileStores/config.ts
```diff
@@ -74,6 +74,14 @@ export type ConfigOptions = {
...
   blockGraffiti: string
   nodeName: string
+  /**
+   * The number of CPU threads to us for long-running node operations, like creating
+   * transactions and verifying blocks. 0 disables threads (this is likely to cause
+   * performance issues), and -1 auto-detects based on the number of CPU cores.
+   * Each thread uses several hundred MB of memory, so try a lower value to reduce memory
+   * consumption.
+   */
+  nodeThreads: number
```
NullSoldier, line 84: kind of naming nit, but since this just changes how many workers we screate, can we just make this `workers`?

ironfish-cli/src/commands/start.ts
```diff
@@ -94,6 +99,9 @@ export default class Start extends IronfishCommand {
     if (flags.port != undefined && flags.port !== this.sdk.config.get('peerPort')) {
       this.sdk.config.setOverride('peerPort', flags.port)
     }
+    if (flags.threads != undefined && flags.port !== this.sdk.config.get('nodeThreads')) {
```
NullSoldier, line 102: copy paste error... flags.port !==

dguenther: Fixed this

ironfish/src/node.ts
```diff
@@ -238,7 +238,11 @@ export class IronfishNode {
 
     // Work in the worker pool happens concurrently,
     // so we should start it as soon as possible
-    this.workerPool.start(os.cpus().length)
+    let threads = this.config.get('nodeThreads')
+    if (threads === -1) {
+      threads = os.cpus().length
```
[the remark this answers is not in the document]
dguenther: Changed this to -1

PR #129, opened 2021-05-10
Reduce deserializations when adding transactions

dguenther:
Wraps the syncTransaction function in a withReference to avoid multiple deserializations, and combined the acceptTransaction and add functions to avoid having to construct the transaction hash twice.


NullSoldier:
Reviewed with https://github.com/iron-fish/ironfish/pull/129/files?diff=split&w=1 helped a lot

NullSoldier (approved):

PR #132, opened 2021-05-10
Wait for all promises to resolve or reject before shutting down

dguenther:
`Promise.all` will continue as soon as one of the promises rejects, so this switches to `Promise.allSettled` to wait for all promises to either resolve or reject.

I'm not sure that this was actually causing any issues, but it seemed like there could be cases where one of the shutdowns throws, then attempting to close the DBs could break one of the other shutdowns.

### DIFF
--- a/ironfish/src/node.ts
+++ b/ironfish/src/node.ts
@@ -266,7 +266,7 @@ export class IronfishNode {
   }
 
   async shutdown(): Promise<void> {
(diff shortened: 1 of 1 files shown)


NullSoldier:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

PR #139, opened 2021-05-11
Add new hash to next table

NullSoldier:
This table is used to store the heaviest head chain BlockHash -> BlockHash

### DIFF
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -26,6 +26,7 @@ import {
   HashToNextSchema,
   SequenceToHash2Schema,
   MetaSchema,
+  HashToNext2Schema,
 } from './schema'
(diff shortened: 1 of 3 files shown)


ironfish/src/blockchain/blockchain.ts
```diff
@@ -98,6 +99,8 @@ export class Blockchain<
   sequenceToHash: IDatabaseStore<SequenceToHashSchema>
   // Sequence -> BlockHash
   sequenceToHash2: IDatabaseStore<SequenceToHash2Schema>
+  // BlockHash -> BlockHash
+  hashToNext2: IDatabaseStore<HashToNext2Schema>
```
leanthebean, line 103: sounds like this is going to be refactored in future PRs

leanthebean (approved):

PR #143, opened 2021-05-13
Remove graph from Blockchain

NullSoldier:
Graphs were an idea that let us sync in any order, randomly. We're removing them to make things simpler and less error prone.

 - [x] Delete SequenceToHash (bigint -> BlockHash[])
 - [x] Add SequenceToHash (bigint -> BlockHash)
 - [x] Delete HashToNext (BlockHash -> BlockHashes[])
 - [x] Add HashToNext (BlockHash -> BlockHashes)
 - [x] Remove getHeaviestHead()
 - [x] Remove getLatestHeader()


dguenther (approved):
Noticed a couple potential typos, but looks good to me!

Issue #20, opened 2021-04-07
Import an Account


leanthebean:
<img width="469" alt="Screen Shot 2021-04-07 at 11 25 50 AM" src="https://user-images.githubusercontent.com/2285160/113915757-02a12000-9794-11eb-90fe-63f153f1d1dc.png">

It looks like there's either confusion around how to import an account without a file, or the command is not prompting the user to enter in account details.

PR #150, opened 2021-05-17
Add interactive account import to accounts:import

dev-null-noop:
Add interactive command line inputs when no importPath is specified on the command line. This is a potential fix for https://github.com/iron-fish/ironfish/issues/20


NullSoldier:
This needs to be rebased to get the fix here so actions can run, https://github.com/iron-fish/ironfish/pull/151

PR #154, opened 2021-05-17
interactive command line input

dev-null-noop:
provide interactive experience for accounts:import if no importPath variable is specified.


dev-null-noop:
provided as potential fix for https://github.com/iron-fish/ironfish/issues/20

PR #155, opened 2021-05-17
Support interactive import in accounts:import 

dev-null-noop:
Provide an interactive experience to the end user if no importPath is specified when account:import is called on the command line. 

Provides potential fix for https://github.com/iron-fish/ironfish/issues/20

### DIFF
--- a/ironfish-cli/src/commands/accounts/import.ts
+++ b/ironfish-cli/src/commands/accounts/import.ts
@@ -6,6 +6,7 @@ import { IronfishCommand } from '../../command'
 import { JSONUtils, PromiseUtils, Account } from 'ironfish'
 import fs from 'fs'
 import { flags } from '@oclif/command'
+import { cli } from 'cli-ux'
 
 export class ImportCommand extends IronfishCommand {
   static description = `Import an account`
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/accounts/import.ts
```diff
@@ -39,19 +40,25 @@ export class ImportCommand extends IronfishCommand {
...
+      })) as string
+      const importPublicAddress = (await cli.prompt('Enter the account public address', {
+        required: true,
+      })) as string
+      data = `{"name": "${importAccountName}", "spendingKey": "${importSpendingKey}",
+      "incomingViewKey": "${importIncomingViewKey}", "outgoingViewKey": "${importOutgoingViewKey}",
+      "publicAddress": "${importPublicAddress}"}`
     }
 
     if (data === null) {
```
NullSoldier, line 64: This looks like it will not allow using STDIN to import an account anymore. Instead of just deleting that functionality and replacing it, can you keep all 3 paths?

 *.Import by path
 * Import by STDIN
 * Import by interactive flow

NullSoldier: You could check if there's any data, and if there is any, then you can just import that versus perform interactive.

dev-null-noop: Updated the code to use process.stdin.isTTY to detect if stdin input is passed or not. This will now either accept a path, stdin input, or switch to interactive mode if neither of those inputs are present. Updated the import statements for fs and cli-ux. Tested locally and all methods checked out.

NullSoldier (approved):

ironfish-cli/src/commands/accounts/import.ts
NullSoldier: Looks great!

NullSoldier (approved):

PR #162, opened 2021-05-19
Update peer sequence in gossip / sync

NullSoldier:
This should update the sequence in 2 cases... 

1. If they gossip a block we update their known sequence if it's higher. We don't update their work/ head because we don't know that it's their head since they could have added it to a fork.
2. If we sync from them we know its their head, so we update their peer info to this block if its later than their existing announced one

### DIFF
--- a/ironfish/src/syncer.ts
+++ b/ironfish/src/syncer.ts
@@ -365,9 +365,11 @@ export class Syncer {
           break
         }
 
-        peer.sequence = block.header.sequence
-        peer.head = block.header.hash
-        peer.work = block.header.work
+        if (!peer.sequence || block.header.sequence > peer.sequence) {
+          peer.sequence = block.header.sequence
+          peer.head = block.header.hash
+          peer.work = block.header.work
+        }
 
         head = block.header.hash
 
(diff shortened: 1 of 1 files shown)


ironfish/src/syncer.ts
```diff
@@ -365,9 +365,11 @@ export class Syncer {
           break
         }
 
-        peer.sequence = block.header.sequence
-        peer.head = block.header.hash
-        peer.work = block.header.work
+        if (!peer.sequence || block.header.sequence > peer.sequence) {
```
leanthebean, line 368: should this be above the break?

ironfish/src/syncer.ts
```diff
@@ -438,7 +440,11 @@ export class Syncer {
       return
     }
 
-    await this.addBlock(peer, block)
+    const { block } = await this.addBlock(peer, newBlock)
+
+    if (!peer.sequence || block.header.sequence > peer.sequence) {
+      peer.sequence = block.header.sequence
```
leanthebean, line 446: should we also check if we need to update the peer's head and work?

ironfish/src/syncer.ts
NullSoldier: If we do that, we'll update their sequence to be equal to an invalid block, right?

ironfish/src/syncer.ts
NullSoldier: How would you know what a peers head and work are from a gossip message? They could have added it to a fork and then gossiped it.

ironfish/src/syncer.ts
leanthebean: hm okay. they did send it, but sure yeah i see your point

leanthebean (approved):

PR #164, opened 2021-05-19
Add peer banning system

NullSoldier:
This adds the peer banning system by identity. Right now the ban list is
not persisted, and no top of that if a peer restarts they get a new
identity.

Also all verification errors have instant ban score, I'll discriminate
on those later and make them worth different values.

### DIFF
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -55,6 +55,7 @@ import { createDB } from '../storage/utils'
 import LRU from 'blru'
 import { BufferMap } from 'buffer-map'
 import { BlockHeaderEncoding, TransactionArrayEncoding } from './encoding'
+import { BAN_SCORE } from '../network/peers/peer'
 
 export class Blockchain<
   E,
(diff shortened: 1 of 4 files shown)


ironfish/src/blockchain/blockchain.ts
```diff
@@ -268,6 +273,8 @@ export class Blockchain<
 
       return { isAdded: true, reason: null }
     })
+
+    return { ...result, score: result.reason ? BAN_SCORE.MAX : BAN_SCORE.NO }
```
NullSoldier, line 277: This won't be hard coded later, it'll be returned from all the validation functions.

ironfish/src/network/peers/peer.ts
```diff
@@ -578,5 +596,18 @@ export class Peer {
     this.onStateChanged.clear()
     this.onKnownPeersChanged.clear()
     this.onMessage.clear()
+    this.onBanned.clear()
+  }
+
+  punish(score: number, reason?: string): boolean {
```
dguenther, line 602: Should this take BAN_SCORE or is it more flexible?

ironfish/src/network/peers/peer.ts
```diff
@@ -176,6 +186,11 @@ export class Peer {
    */
   readonly onKnownPeersChanged: Event<[]> = new Event()
 
+  /**
+   * Fired when the peer should be banned
+   */
+  readonly onBanned: Event<[]> = new Event()
```
dguenther, line 192: It doesn't look like this is emitted-to anywhere, is it planned to be used later?

dguenther (approved):

ironfish/src/network/peers/peer.ts
NullSoldier: I don't think so. ban score is a number, and BAN_SCORE is really just suggested relative ban score values you can use across the code base. I'd like to keep this flexible. I'm not really sure if I want to keep BAN_SCORE or not.

ironfish/src/network/peers/peer.ts
NullSoldier: Fixed, that prevented the entire thing from working. I'll write a test here in a second.

PR #178, opened 2021-05-22
Move seed into blockchain and make head non-null

dguenther:
* Removes node.seed() and moves the contents into chain.open(). Added an `autoSeed` option to help the outdated jest tests pass.
* Changes chain.head, chain.latest, and chain.genesis into properties that throw if their backing field is null. These should typically only ever be null if the chain hasn't been initialized with the genesis block, then should be set after that.

### DIFF
--- a/ironfish/src/blockchain/blockchain.test.ts
+++ b/ironfish/src/blockchain/blockchain.test.ts
@@ -10,30 +10,15 @@ import { makeBlockAfter } from '../testUtilities/helpers/blockchain'
 describe('Blockchain', () => {
   const nodeTest = createNodeTest()
 
-  it('constructs an empty chain', async () => {
+  it('add genesis block', async () => {
     const { chain } = nodeTest
     await chain.open()
 
-    expect(await chain.notes.size()).toBe(0)
-    expect(await chain.nullifiers.size()).toBe(0)
-    expect(chain.isEmpty).toBe(true)
-    expect(chain.head).toBe(null)
-    expect(chain.latest).toBe(null)
-    expect(chain.synced).toBe(false)
-  })
+    const genesis = await chain.getBlock(chain.genesis)
+    Assert.isNotNull(genesis)
 
-  it('add genesis block', async () => {
-    const { node, chain } = nodeTest
-    await chain.open()
-
-    expect(chain.head).toBe(null)
-    expect(chain.hasGenesisBlock).toBe(false)
-    expect(chain.isEmpty).toBe(true)
-
-    const genesis = await node.seed()
-
-    expect(chain.head?.hash).toEqualHash(genesis.header.hash)
-    expect(chain.latest?.hash).toEqualHash(genesis.header.hash)
+    expect(chain.head.hash).toEqualHash(genesis.header.hash)
+    expect(chain.latest.hash).toEqualHash(genesis.header.hash)
     expect(chain.isEmpty).toBe(false)
     expect(chain.hasGenesisBlock).toBe(true)
     expect(await chain.notes.size()).toBeGreaterThan(0)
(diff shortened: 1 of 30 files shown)


ironfish/src/blockchain/blockchain.test.ts
```diff
@@ -10,30 +10,15 @@ import { makeBlockAfter } from '../testUtilities/helpers/blockchain'
...
-    const { node, chain } = nodeTest
+    const { chain } = nodeTest
     await chain.open()
 
-    expect(chain.head).toBe(null)
-    expect(chain.hasGenesisBlock).toBe(false)
-    expect(chain.isEmpty).toBe(true)
-
-    const genesis = await node.seed()
+    const genesis = await chain.getBlock(chain.genesis)
```
NullSoldier, line 17: ooh can we just delete this block fetch now?

ironfish/src/blockchain/blockchain.test.ts
```diff
@@ -46,12 +31,12 @@ describe('Blockchain', () => {
...
     strategy.disableMiningReward()
 
-    const genesis = await nodeTest.node.seed()
-    Assert.isNotNull(genesis)
-
     // G -> A1 -> A2
     //         -> B2 -> B3
 
+    const genesis = await chain.getBlock(chain.genesis)
+    Assert.isNotNull(genesis)
```
NullSoldier, line 38: can we delete this block fetch here too?

ironfish/src/blockchain/blockchain.ts
```diff
@@ -205,21 +244,58 @@ export class Blockchain<
...
+  }
+
   async open(): Promise<void> {
     if (this.opened) return
     this.opened = true
 
     await this.db.open()
 
-    this.genesis = await this.getHeaderAtSequence(GENESIS_BLOCK_SEQUENCE)
+    const genesis = await this.getHeaderAtSequence(GENESIS_BLOCK_SEQUENCE)
```
NullSoldier, line 257: Can we move this into a function on the class called seed?

ironfish/src/blockchain/blockchain.ts
```diff
@@ -1284,10 +1355,6 @@ export class Blockchain<
       return
     }
 
-    if (!this.head) {
```
NullSoldier, line 1287: Would you mind bringing this back so we don't have to know all the conditions to call this function? It also simplifies the call sites.

ironfish/src/blockchain/blockchain.ts
```diff
@@ -205,21 +244,58 @@ export class Blockchain<
...
+      this.genesis = genesisHeader
+      this.head = this.genesis
+      this.latest = this.genesis
+    }
 
     const headHash = await this.meta.get('head')
-    if (headHash) this.head = await this.getHeader(headHash)
+    if (headHash) {
+      const head = await this.getHeader(headHash)
+      if (head) {
```
NullSoldier, line 283: do we need to do this if check?

ironfish/src/blockchain/blockchain.ts
dguenther: This check won't ever be hit now that `this.head` throws if it's null. We don't have any checks around the callsite for this as far as I know

ironfish/src/blockchain/blockchain.ts
dguenther: What about an assert instead? If `head` is set in the meta table, but there's no corresponding header, we need to handle that case somehow so we don't set `this.head` to null.

ironfish/src/blockchain/blockchain.ts
dguenther: 👍 pulled this out

ironfish/src/blockchain/blockchain.ts
NullSoldier: ah an assert sounds good

ironfish/src/blockchain/blockchain.ts
NullSoldier: well I was thinking moving the check in open for _head but that's fine

NullSoldier (approved):

PR #180, opened 2021-05-26
Discard mined blocks if chain head changes or can't add block

dguenther:
Adds two checks, along with tests:
* If adding the block to the chain fails, does not send the block out to the network
* If a block is successfully mined, checks that the new block's previous hash is equal to the chain's heaviest head, else discards the block.

### DIFF
--- a/ironfish/src/mining/director.ts
+++ b/ironfish/src/mining/director.ts
@@ -425,6 +425,11 @@ export class MiningDirector<
       return
     }
 
+    if (!this.chain.head || !block.header.previousBlockHash.equals(this.chain.head.hash)) {
+      this.logger.debug('Discarding block that no longer attaches to heaviest head')
+      return
+    }
+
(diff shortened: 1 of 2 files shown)


leanthebean (approved):

ironfish/src/mining/director.ts
```diff
@@ -448,8 +453,12 @@ export class MiningDirector<
       ],
     })
 
-    await this.chain.addBlock(block)
-    this.onNewBlock.emit(block)
+    const { isAdded, reason } = await this.chain.addBlock(block)
+    if (isAdded) {
+      this.onNewBlock.emit(block)
+    } else {
+      this.logger.debug(`Failed to add mined block to chain with reason ${String(reason)}`)
```
NullSoldier, line 460: set to error?

NullSoldier (approved):

PR #190, opened 2021-06-02
Dont gossip blocks when we don't add the block

NullSoldier:
We should only gossip blocks when we added them, because if we gossip a
block, it signals to others that we also have that block on our chain.
It's useful because if someone gets our gossip block, and its an orphan,
they will want to download the chain from us to that oprhan. So if we
gossip without adding, then that won't work.

Many coins distinguish this as announce vs broadcast, where announce
tells people we have it, and broadcast gossips it without indicating we
have it or not. Announce sends entire hash, broadcast sends entire
block.

If you look at the code, there is one situation here where we have the
block but we don't gossip it, and that's when it's a duplicate and we
already have it. At that point we don't regossip it out. We can
determine if we should be gossiping it out later in this case.

### DIFF
--- a/ironfish/src/network/peerNetwork.ts
+++ b/ironfish/src/network/peerNetwork.ts
@@ -748,7 +748,7 @@ export class PeerNetwork {
     if (!peer) return false
 
     try {
-      await this.node.syncer.addNewBlock(peer, block)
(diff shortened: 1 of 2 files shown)


dguenther (approved):
Not totally sure if this will cause forking, but let's give it a shot since it should reduce network traffic

PR #192, opened 2021-06-02
Transfer ownership of orphan syncing to syncer

NullSoldier:
Right now, when were orphan syncing, and any error occurs, it'll hit the
error handler inside of PeerNetwork, which is used to determine if
gossiping should happen. Ideally, as soon as it determines its an orphan
we should no longer gossip.

To do this, instead of awaiting syncOrphan()
we just use startSync() directly. syncOrphan was used to guarantee that
the peer had the orphan block to begin with, but we don't need to do
that anymore because we only gossip blocks that we add.

Note that once startSync is called, it takes over responsibility for handling errors for things like peer disconnection errors, etc.


dguenther (approved):
This might need a rebase, i'm not 100% sure it'll merge in correctly as-is

NullSoldier:
Rebased

PR #194, opened 2021-06-02
Color blocks with no graffiti as gray

NullSoldier:
Blocks with no graffiti should be gray now


ironfish-graph-explorer/src/index.js
```diff
@@ -37,17 +37,23 @@ function getNodes() {
         const nodeOffsetX = (islandOffset * ISLAND_OFFSET) + (childOffset * CHILD_OFFSET)
         const nodeOffsetY = block.seq * LEVEL_OFFSET
 
+        const graffiti = block.graffiti.replace(/\0/g, '').trim()
```
NullSoldier, line 40: This replaces the null terminator with '', which these empty buffers are filled with

dguenther (approved):

PR #195, opened 2021-06-02
Provide more info about old clients connecting

NullSoldier:
- Use peer host as name if they are unidentified
 - Print out information about their version if they connect from an old
 version

### DIFF
--- a/ironfish/src/network/peers/peerManager.ts
+++ b/ironfish/src/network/peers/peerManager.ts
@@ -17,6 +17,7 @@ import {
 import {
   DisconnectingMessage,
   DisconnectingReason,
+  getIdentifyVersion,
   IncomingPeerMessage,
   InternalMessageType,
   isDisconnectingMessage,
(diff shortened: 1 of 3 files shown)


ironfish/src/network/peers/peerManager.ts
```diff
@@ -887,10 +888,16 @@ export class PeerManager {
...
-      this.logger.debug(
-        `Disconnecting from ${peer.displayName} - Sent unexpected message ${message.type} while waiting for identity`,
-      )
-      peer.close()
+      let error = `Disconnecting from ${peer.displayName} - Sent unexpected message ${message.type} while waiting for identity`
+
+      // If we can parse out their old version, lets display that instead
+      const oldVersion = getIdentifyVersion(message)
+      if (oldVersion) {
+        error = `Old peer attempted to connect to us from ${peer.displayName}: ${oldVersion}`
```
dguenther, line 896: Should we reorder the checks in `handleWaitingForIdentityMessage` to look at just the version field first? Maybe checking that the message matches the spec exactly is too aggressive, and we should check it field by field. Or we could log the entire message they send

NullSoldier:
I'm closing this because it was actually because of a bug where I deployed an old version to staging

PR #212, opened 2021-06-08
Move signal encryption/decryption into worker pool

dguenther:
This and WebRTC-related code are the two most common occurrences when profiling the main thread with enableSyncing set to false (which is what the bootstrap node on staging has set).


ironfish/src/workerPool/pool.ts
```diff
@@ -225,4 +228,48 @@ export class WorkerPool {
 
     return response.verified
   }
+
+  async boxMessage(
```
rohanjadvani, line 232: I'm not super familiar with the worker pool tests, but do you think we could add some test cases for these methods?

rohanjadvani (approved):

ironfish/src/network/peers/peerManager.ts
```diff
@@ -313,16 +313,31 @@ export class PeerManager {
...
       if (peer.state.identity === null) {
-        const message = 'Cannot establish a WebRTC connection without a peer identity'
-        this.logger.debug(message)
-        connection.close(new NetworkError(message))
+        errorMessage = 'Cannot establish a WebRTC connection without a peer identity'
+      }
+
+      // Ensure one or more brokering peers exists before encrypting the signaling message,
+      // but discard the brokering peer in case its state changes during encryption
+      if (this.getBrokeringPeer(peer) == null) {
```
NullSoldier, line 324: It's strange that this finds a brokering peer then throws it away, only to try again down the line. The general flow here is a bit confusing, I wish we could tighten this up but I'm okay if it merges.

ironfish/src/workerPool/pool.ts
NullSoldier: The methods are actually tested via the peer network which calls these functions already

NullSoldier (approved):

PR #213, opened 2021-06-09
refactor(ironfish): Remove usage of `looseNotes` and `looseNullifiers`

rohanjadvani:
### Description

* Update `makeBlockAfter` to check the `reason` instead of `valid` flag
* Remove `looseNotes` and `looseNullifiers` from the `Blockchain` class

### Tests

* Removed tests that were referencing `looseNotes` and `looseNullifiers`
* This is temporary, but skipping tests that use `hasValidSpends` in `blockchain.old.test.ts`. These tests will be moved over to the verifier and create block fixtures instead of fake blocks

### DIFF
--- a/ironfish/src/assert.ts
+++ b/ironfish/src/assert.ts
@@ -8,7 +8,8 @@ export class Assert {
   }
 
   static isUndefined<T>(x: undefined | T, message?: string): asserts x is undefined {
-    if (x !== undefined) throw new Error(message || `Expected ${String(x)} to be undefined`)
+    if (x !== undefined)
+      throw new Error(message || `Expected ${JSON.stringify(x)} to be undefined`)
   }
 
   static isNotUndefined<T>(x: undefined | T, message?: string): asserts x is T {
--- a/ironfish/src/assert.ts
+++ b/ironfish/src/assert.ts
@@ -7,6 +7,10 @@ export class Assert {
     throw new Error(x)
   }
 
+  static isUndefined<T>(x: undefined | T, message?: string): asserts x is undefined {
+    if (x !== undefined) throw new Error(message || `Expected ${String(x)} to be undefined`)
+  }
+
   static isNotUndefined<T>(x: undefined | T, message?: string): asserts x is T {
     if (x === undefined) throw new Error(message || `Expected value not to be undefined`)
   }
--- a/ironfish/src/blockchain/blockchain.old.test.ts
+++ b/ironfish/src/blockchain/blockchain.old.test.ts
@@ -35,36 +35,9 @@ describe('Note adding', () => {
   it('immediately adds in order notes to the tree', async () => {
     await blockchain.addNote(0, 'zero')
     await blockchain.addNote(1, 'one')
-    expect(blockchain.looseNotes[0]).toBe('zero')
-    expect(blockchain.looseNotes[1]).toBe('one')
     expect(await blockchain.notes.size()).toBe(2)
     expect(await blockchain.notes.get(0)).toBe('zero')
     expect(await blockchain.notes.get(1)).toBe('one')
-    expect(await blockchain.nullifiers.size()).toBe(0)
-    expect(listener).not.toBeCalled()
-  })
-  it('adds an out of order note only to the loose notes', async () => {
-    await blockchain.addNote(10, 'ten')
-    await blockchain.addNote(11, 'eleven')
-    await blockchain.addNote(12, 'twelve')
-    expect(blockchain.looseNotes[10]).toBe('ten')
-    expect(blockchain.looseNotes[11]).toBe('eleven')
-    expect(blockchain.looseNotes[12]).toBe('twelve')
-    expect(await blockchain.notes.size()).toBe(0)
-    expect(await blockchain.nullifiers.size()).toBe(0)
-    expect(listener).not.toBeCalled()
-  })
-  it('syncs loose notes to the tree when the gap fills in', async () => {
-    await blockchain.addNote(2, 'two')
-    await blockchain.addNote(1, 'one')
-    await blockchain.addNote(0, 'zero')
-    expect(blockchain.looseNotes[0]).toBe('zero')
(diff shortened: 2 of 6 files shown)


ironfish/src/blockchain/blockchain.old.test.ts
```diff
@@ -76,8 +49,7 @@ describe('Note adding', () => {
     await blockchain.addNote(0, 'zero')
     await blockchain.addNote(1, 'one')
     await blockchain.addNote(2, 'two')
-    await blockchain.addNote(2, 'not two')
-    expect(errorFn).toHaveBeenCalled()
+    await expect(blockchain.addNote(2, 'not two')).rejects.toMatchSnapshot()
```
NullSoldier, line 52: Maybe expect the specific error?

rohanjadvani: Sounds good - fixed

ironfish/src/blockchain/blockchain.old.test.ts
```diff
@@ -146,8 +85,7 @@ describe('Nullifier adding', () => {
     await blockchain.addNullifier(0, nullifier0)
     await blockchain.addNullifier(1, nullifier1)
     await blockchain.addNullifier(2, nullifier2)
-    await blockchain.addNullifier(2, nullifier0)
-    expect(warnFn).toHaveBeenCalled()
+    await expect(blockchain.addNullifier(2, nullifier0)).rejects.toMatchSnapshot()
```
NullSoldier: maybe expect the error here too?

rohanjadvani: Updated - removed snapshot here as well

ironfish/src/assert.ts
```diff
@@ -7,6 +7,10 @@ export class Assert {
     throw new Error(x)
   }
 
+  static isUndefined<T>(x: undefined | T, message?: string): asserts x is undefined {
+    if (x !== undefined) throw new Error(message || `Expected ${String(x)} to be undefined`)
```
dguenther, line 11: It would be nice to use `JSON.stringify` here so we get better logging for objects.

ironfish/src/blockchain/blockchain.old.test.ts
```diff
@@ -96,45 +70,12 @@ describe('Nullifier adding', () => {
...
-    expect(blockchain.looseNullifiers[1]).toEqualNullifier(nullifier1)
-    expect(blockchain.looseNullifiers[2]).toEqualNullifier(nullifier2)
-    expect(await blockchain.nullifiers.size()).toBe(3)
-    expect(await blockchain.nullifiers.get(0)).toEqualNullifier(nullifier0)
-    expect(await blockchain.nullifiers.get(1)).toEqualNullifier(nullifier1)
-    expect(await blockchain.nullifiers.get(2)).toEqualNullifier(nullifier2)
-    expect(await blockchain.notes.size()).toBe(0)
-    expect(listener).not.toBeCalled()
-  })
   it("warns if the note doesn't match the previously inserted note that position", async () => {
```
dguenther, line 79:
```suggestion
  it("throws if the nullifier doesn't match the previously inserted nullifier in that position", async () => {
```

ironfish/src/blockchain/blockchain.old.test.ts
```diff
@@ -35,39 +35,12 @@ describe('Note adding', () => {
...
-    expect(blockchain.looseNotes[1]).toBe('one')
-    expect(blockchain.looseNotes[2]).toBe('two')
-    expect(await blockchain.notes.size()).toBe(3)
-    expect(await blockchain.notes.get(0)).toBe('zero')
-    expect(await blockchain.notes.get(1)).toBe('one')
-    expect(await blockchain.notes.get(2)).toBe('two')
-    expect(await blockchain.nullifiers.size()).toBe(0)
-    expect(listener).not.toBeCalled()
-  })
   it("logs errors if the note doesn't match the previously inserted note that position", async () => {
```
dguenther, line 44:
```suggestion
  it("throws if the note doesn't match the previously inserted note that position", async () => {
```

ironfish/src/blockchain/blockchain.old.test.ts
```diff
@@ -35,39 +35,12 @@ describe('Note adding', () => {
...
     const errorFn = jest.fn()
     blockchain['logger'].mockTypes((type) => {
```
dguenther, line 46: This should be deletable now if we don't care to test whether the logger is called

ironfish/src/blockchain/blockchain.ts
```diff
@@ -959,43 +947,25 @@ export class Blockchain<
...
-            )
+            const message = `Tried to insert a note, but a different note already there for position ${index}`
+            this.logger.error(message)
+            throw new Error(message)
           }
           return
+        } else if (index > noteCount) {
+          const message = `Can't insert a note at index ${index}. Merkle tree has a count of ${noteCount}`
+          this.logger.error(message)
+          throw new Error(message)
```
dguenther, line 965: Could you add a test for this if there isn't one already?

ironfish/src/blockchain/blockchain.ts
```diff
@@ -959,43 +947,25 @@ export class Blockchain<
       [this.notes.counter, this.notes.leaves, this.notes.nodes],
```
dguenther, line 947: I think the comment for `addNote` should remove references to `looseNotes`. Also, notes can't come in any order now (unrelated to this PR).

ironfish/src/blockchain/blockchain.ts
```diff
@@ -1015,39 +985,23 @@ export class Blockchain<
       [this.nullifiers.counter, this.nullifiers.leaves, this.nullifiers.nodes],
```
dguenther, line 985: Same comment needs to be updated here (should have been `looseNullifiers` anyway 😅 )

dguenther (approved):

PR #218, opened 2021-06-11
refactor(ironfish): Update `sequence` type to `number`

rohanjadvani:


ironfish/src/account/accountsdb.ts
```diff
@@ -70,7 +70,7 @@ export class AccountsDB {
     value: {
       transaction: Buffer
       blockHash: string | null
-      submittedSequence: bigint | null
+      submittedSequence: number | null
```
dguenther, line 73: I think this might be a breaking change -- We serialize the value with `JsonEncoding`, which uses `IJSON` and appends an "n" to the end of bigints. You could probably test it by mining some blocks on the current version, checking your balance to make sure the wallet picked up the miners fee transactions, then switching to this branch and checking your balance.

I think it's okay to do a breaking change into staging since we already have breaking changes there anyway, we'll just want to check if that's the case, and tell people to reset their accounts database.

dguenther (commented):
Once you double-check whether this is a breaking change for the accounts DB this looks good to me!

ironfish/src/account/accountsdb.ts
dguenther: Oh, I gave bad repro steps for this, it's actually the pending transactions that have this field set to a number/bigint. These are better steps:
1. Start two clean nodes, mining on one of them
2. Stop mining, then use `accounts:pay` to send money from the mining node to the other node (so that the transaction doesn't end up on a block)
3. Stop both nodes, change to this branch, then start them again. Error should display:
```
(node:65298) UnhandledPromiseRejectionWarning: TypeError: Cannot mix BigInt and other types, use explicit conversions
    at Accounts.rebroadcastTransactions (/Users/derek/repos/ironfish/ironfish/build/src/account/accounts.js:516:30)
```

dguenther:
I posted incorrect repro steps originally, but I did manage to find an error related to the accounts db. Looks good!

dguenther (approved):

ironfish/src/primitives/blockheader.ts
```diff
@@ -330,7 +329,7 @@ export class BlockHeaderSerde<
     // as it can be from untrusted sources
     const header = new BlockHeader(
       this.strategy,
-      BigInt(data.sequence),
+      Number(data.sequence),
```
NullSoldier, line 332: we should convert SerializedBlockHeader.sequence to a number now too that it isnt a bigint

ironfish/src/storage/database/encoding.ts
```diff
@@ -25,9 +25,9 @@ export class StringEncoding<TValues extends string = string>
   }
 }
 
-export class BigIntEncoding implements IDatabaseEncoding<bigint> {
-  serialize = (value: bigint): Buffer => Buffer.from(value.toString(), 'utf8')
-  deserialize = (buffer: Buffer): bigint => BigInt(buffer.toString('utf8'))
+export class NumberEncoding implements IDatabaseEncoding<number> {
+  serialize = (value: number): Buffer => Buffer.from(value.toString(), 'utf8')
```
NullSoldier, line 29: You can also just use a faster write the number straight to the buffer now that its just a number

```js
const b = Buffer.alloc(64)
b.writeDoubleBE(value)
return b
```

rohanjadvani: Updated!

ironfish/src/storage/database/encoding.ts
```diff
@@ -26,8 +26,15 @@ export class StringEncoding<TValues extends string = string>
 }
 
 export class NumberEncoding implements IDatabaseEncoding<number> {
-  serialize = (value: number): Buffer => Buffer.from(value.toString(), 'utf8')
-  deserialize = (buffer: Buffer): number => Number(buffer.toString('utf8'))
+  serialize(value: number): Buffer {
+    const buffer = Buffer.alloc(64)
```
NullSoldier, line 30: oh shoot, this should be 8 as a double is 64bit which is 8 bytes

ironfish/src/storage/database/encoding.ts
```diff
@@ -26,8 +26,15 @@ export class StringEncoding<TValues extends string = string>
...
+    buffer.writeDoubleLE(value)
+    return buffer
+  }
+
+  deserialize(buffer: Buffer): number {
+    return buffer.readDoubleLE(0)
```
NullSoldier, line 36: the offset is optional, so we can exclude it because it defaults at 0.

ironfish/src/storage/database/encoding.ts
NullSoldier: I know I wrote that it should be 64 but that's not correct

NullSoldier (approved):

NullSoldier:
Looks great!

PR #219, opened 2021-06-14
Lock database before executing transaction handlers

dguenther:
The database code runs transaction handlers immediately and doesn't block them until they reach a database call, so logic in the transaction handlers that's dependent on external state could be incorrect. We were seeing issues that seemed to stem from this because we forgot to pass in the transactions in all cases, but it seems safer for now to lock around the entire transaction handler and reduce the amount of work being done inside of transaction handlers if possible.

### DIFF
--- a/ironfish/src/storage/database.test.ts
+++ b/ironfish/src/storage/database.test.ts
@@ -564,23 +564,23 @@ describe('Database', () => {
     it('should wait for a lock before executing the handler', async () => {
       await db.open()
 
-      let value = null
+      let value = ''
 
       const [waitingPromise, waitingResolve] = PromiseUtils.split<void>()
 
       // Queue up two transactions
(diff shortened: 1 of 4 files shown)


ironfish/src/storage/database.test.ts
```diff
@@ -559,6 +560,43 @@ describe('Database', () => {
...
+        value = 't2'
+        await waitingPromise
+      })
+
+      const t3 = db.transaction([db.metaStore], 'readwrite', async () => {
+        value = 't3'
+        await waitingPromise
+      })
+
+      await PromiseUtils.sleep(0)
```
NullSoldier, line 587: Wish we had a comment here, no one will be able to figure out why this is here because of how Lock works

NullSoldier (approved):

PR #222, opened 2021-06-16
feat(ironfish-cli): Support rescanning without an active node

rohanjadvani:
* Add an `offline` flag to the `accounts:rescan` command
* Export `ScanState` and add a `getScan()` method (for testing)


ironfish-cli/src/commands/accounts/rescan.ts
```diff
@@ -22,34 +22,56 @@ export class RescanCommand extends IronfishCommand {
       description:
         'clear the in-memory and disk caches before rescanning. note that this removes all pending transactions',
     }),
+    offline: flags.boolean({
```
dguenther, line 25: This should probably be named `local` to match up with the similar flag in the `config` commands (e.g. `config:get`)

ironfish-cli/src/commands/accounts/rescan.ts
```diff
@@ -22,34 +22,56 @@ export class RescanCommand extends IronfishCommand {
...
+    }: {
+      sequence: number
+    }): void => {
+      cli.action.status = `Scanning Block: ${sequence}, ${Math.floor(
         (Date.now() - startedAt) / 1000,
       )} seconds`
     }
+    const updateCliStatusWithStartTime = updateCliStatus(Date.now())
+
+    if (offline) {
```
NullSoldier, line 61: We can reduce a lot of this code to just use `getConnectedClient` from `config:show` which at some point needs to be expanded out.

ironfish/src/account/accounts.ts
```diff
@@ -184,6 +184,10 @@ export class Accounts {
     return false
   }
 
+  getScan(): ScanState | null {
+    return this.scan
```
NullSoldier, line 188: I don't think this should be necessary you can just use accounts.scan

ironfish-cli/src/commands/accounts/rescan.ts
dguenther: Ah, I hadn't seen that before, that's convenient.

dguenther (changes requested):
It would help to move `this.accounts.load()` out of `node.ts/openDB` and into `node.ts/start`. This would help avoid any crashes resulting from the Accounts database being corrupt, which is possible in the case where we're trying to rescan it.

These steps should reproduce a crash:

1. Clear all your databases
2. Replace the body of `accountsdb.ts/loadTransactionsIntoMap` to:
```ts
    await this.transactions.clear()
    await this.database.transaction([this.transactions], 'readwrite', async (tx) => {
      // @ts-expect-error error
      await this.transactions.add(Buffer.alloc(32), {
        asdf: 'qwer',
      })
    })
```
3. Run `ironfish start`, kill the node, then run `ironfish start` again. This error should show up:
```
TypeError: Cannot read property 'length' of undefined
    at passArray8ToWasm0 (/Users/derek/repos/ironfish/ironfish-wasm/nodejs/ironfish-wasm-nodejs.js:54:28)
    at Function.deserialize (/Users/derek/repos/ironfish/ironfish-wasm/nodejs/ironfish-wasm-nodejs.js:798:20)
    at IronfishTransaction.takeReference (/Users/derek/repos/ironfish/ironfish/build/src/primitives/transaction.js:23:87)
    at IronfishTransaction.withReference (/Users/derek/repos/ironfish/ironfish/build/src/primitives/transaction.js:37:34)
    at IronfishTransaction.transactionHash (/Users/derek/repos/ironfish/ironfish/build/src/primitives/transaction.js:93:21)
    at AccountsDB.loadTransactionsIntoMap (/Users/derek/repos/ironfish/ironfish/build/src/account/accountsdb.js:159:50)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async Accounts.loadTransactionsFromDb (/Users/derek/repos/ironfish/ironfish/build/src/account/accounts.js:199:9)
    at async Accounts.load (/Users/derek/repos/ironfish/ironfish/build/src/account/accounts.js:623:9)
    at async IronfishNode.openDB (/Users/derek/repos/ironfish/ironfish/build/src/node.js:143:9)
```

ironfish/src/rpc/routes/accounts/rescanAccount.ts
```diff
@@ -35,6 +35,7 @@ router.register<typeof RescanAccountRequestSchema, RescanAccountResponse>(
       if (request.data.reset) {
         await node.accounts.reset()
       }
+      await node.accounts.updateHead()
```
dguenther, line 38: Rather than calling `updateHead` here, I think we should update `scanTransactions` to set `this.headHash` to `chain.head` if `this.headHash` is null before the scan begins.

NullSoldier (approved):

dguenther (approved):

PR #231, opened 2021-06-18
Remove loadGenesisBlock

NullSoldier:
This was old and isnt being used anymore because all the old fake tests
that used it are gone.


dguenther (approved):

dguenther:
A few more things need to be removed but looks good to me

PR #233, opened 2021-06-18
Support export starting from end of chain

NullSoldier:
This allows you to pass negative numbers to export the last N blocks
from the end. If you pass -5 from start, on a chain with a head of
sequence 10, it will export blocks 6-10 (the last 5 blocks)

### DIFF
--- a/ironfish-cli/src/commands/chain/export.ts
+++ b/ironfish-cli/src/commands/chain/export.ts
@@ -67,12 +67,21 @@ export default class Export extends IronfishCommand {
     Assert.isNotNull(node.chain.head, 'head')
     Assert.isNotNull(node.chain.latest, 'latest')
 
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/chain/export.ts
```diff
@@ -82,7 +91,7 @@ export default class Export extends IronfishCommand {
       format: 'Exporting blocks: [{bar}] {value}/{total} {percentage}% | ETA: {eta}s',
     }) as ProgressBar
 
-    progress.start(stop - start, 0)
+    progress.start(stop - start + 1, 0)
```
NullSoldier, line 94: this was an off by 1 error, because this range is inclusive.

dguenther (approved):

PR #237, opened 2021-06-21
feat(ironfish-cli): Add a `local` flag to `accounts:export`

rohanjadvani:
Allow users to get account information without an online node.


dguenther (approved):
Looks good! This could use a rebase, but feel free to merge after that

PR #242, opened 2021-06-22
Fix ironfish reset by not checking upgrade

NullSoldier:
There are instances where we want to open the DB without running any
upgrades. This introduces a new upgrade argument to accounts and
blockchain open() methods. It only checks for mandatory upgrade when
true is passed.

Later we'll remove this in favor of a better system when we have time
and have added DB checks.

### DIFF
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -249,11 +249,13 @@ export class Blockchain<
     return genesisHeader
   }
 
-  async open(): Promise<void> {
+  async open(options?: { upgrade?: boolean }): Promise<void> {
     if (this.opened) return
     this.opened = true
 
-    await this.db.open({ upgrade: this.forceUpgrade })
+    const upgrade = options?.upgrade ?? true
+
+    await this.db.open({ upgrade: upgrade ? this.forceUpgrade : undefined })
 
     let genesisHeader = await this.getHeaderAtSequence(GENESIS_BLOCK_SEQUENCE)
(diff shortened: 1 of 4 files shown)


ironfish/src/blockchain/blockchain.ts
```diff
@@ -249,11 +249,11 @@ export class Blockchain<
     return genesisHeader
   }
 
-  async open(): Promise<void> {
+  async open(options?: { upgrade?: boolean }): Promise<void> {
     if (this.opened) return
     this.opened = true
 
-    await this.db.open({ upgrade: this.forceUpgrade })
+    await this.db.open({ upgrade: options ? this.forceUpgrade : undefined })
```
dguenther, line 256: Should this take the upgrade from `options`?

NullSoldier: opps, fixed.

ironfish/src/node.ts
```diff
@@ -215,12 +215,12 @@ export class IronfishNode {
     })
   }
 
-  async openDB(): Promise<void> {
+  async openDB(options?: { upgrade?: boolean }): Promise<void> {
```
dguenther, line 218: Should the `start` command be passing in `{ upgrade: true }`? Or maybe that should be the default?

NullSoldier: Yes, I made the default true now.

dguenther: Reset might need to pass in `upgrade: false` now

dguenther (approved):

PR #244, opened 2021-06-22
Use node-datachannel instead of simple-peer

dguenther:
We've seen some issues with node-webrtc including segfaults and difficulty building for less-common platforms like ARM Macs. This PR tests out node-datachannel as a replacement, since we really only need the datachannel part of WebRTC.

This will require some work in the future to support browsers -- at that point we should either add support for node-datachannel to simple-peer (unfortunately it won't be a drop-in replacement), or implement an equivalent to simple-peer in WebRTCConnection.

Fixes IRO-684

## TODO

* [x] Rebase on staging once staging is re-branched from master
* [x] Test building on Windows to see if README needs to be updated. On M1, it needs OpenSSL to be installed and an env var to be set pointing at the root directory. (e.g. `brew install openssl`, then `export OPENSSL_ROOT_DIR=/opt/homebrew/opt/openssl@1.1`)

### DIFF
--- a/ironfish/src/network/peers/connections/webRtcConnection.ts
+++ b/ironfish/src/network/peers/connections/webRtcConnection.ts
@@ -2,15 +2,15 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-import nodeDataChannel from 'node-datachannel'
 import type { Logger } from '../../../logger'
 import colors from 'colors/safe'
+import nodeDataChannel from 'node-datachannel'
+import { Assert } from '../../../assert'
 import { Event } from '../../../event'
 import { MetricsMonitor } from '../../../metrics'
 import { LooseMessage, NodeMessageType, parseMessage } from '../../messages'
 import { Connection, ConnectionDirection, ConnectionType } from './connection'
 import { NetworkError } from './errors'
-import { Assert } from '../../../assert'
 
 export type SignalData =
   | {
@@ -187,7 +187,9 @@ export class WebRtcConnection extends Connection {
    * Encode the message to json and send it to the peer
    */
   send = (message: LooseMessage): boolean => {
-    if (!this.datachannel) return false
+    if (!this.datachannel) {
+      return false
+    }
 
     if (message.type === NodeMessageType.NewBlock && this.datachannel.bufferedAmount() > 0) {
       return false
@@ -236,7 +238,9 @@ export class WebRtcConnection extends Connection {
     }
 
     this.setState({ type: 'DISCONNECTED' })
+
     this.datachannel?.close()
+
     try {
       this.peer.close()
     } catch (e) {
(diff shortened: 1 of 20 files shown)


dguenther:
I tested this on my Windows machine and it worked fine without having to perform any additional setup steps. I added a section to the README with the error I ran into when attempting to get it working on an M1 Mac.

This should be ready for review now.

ironfish/src/network/peers/connections/webRtcConnection.ts
```diff
@@ -44,43 +70,58 @@ export class WebRtcConnection extends Connection {
       this.addLatencyWrapper()
     }
 
-    // TODO: This is using google STUN internally, we need to
-    // make it use any of the websocket peers
-    this.peer = new SimplePeer({ initiator, wrtc })
+    // TODO: Use our own STUN servers
```
NullSoldier, line 73: As far as I can tell, google's stun server is okay for production use.

ironfish/src/network/peers/connections/webRtcConnection.ts
```diff
@@ -171,6 +238,13 @@ export class WebRtcConnection extends Connection {
     }
 
     this.setState({ type: 'DISCONNECTED' })
-    this.peer.destroy()
+    this.datachannel?.close()
+    try {
```
NullSoldier, line 242:
```suggestion

    try {
```

ironfish/src/network/peers/peer.test.ts
```diff
@@ -5,7 +5,24 @@
...
+        onError: () => {},
+        onClosed: () => {},
+        onMessage: () => {},
+        close: () => {},
+        isOpen: () => {},
+        sendMessage: () => {},
+      })
+    },
+  }
+})
```
NullSoldier, line 25: this has to be copied into every test now with web rtc?

ironfish/src/network/peers/peerManager.ts
```diff
@@ -745,6 +742,7 @@ export class PeerManager {
     for (const peer of this.peers) {
       this.disconnect(peer, DisconnectingReason.ShuttingDown, 0)
     }
+    nodeDataChannel.cleanup()
```
NullSoldier, line 745: should this be on app teardown intead of here? If it isn't going to be globally torn down, should it at least by in PeerNetwork?

NullSoldier:
I think if we run a few days test, and see that there's no adverse difference. I'm okay with this.

It's unfortunate that we need to bring this in to bring in the added complexity straight into our code. It definitely makes our code base harder to understand, and harder to work with. The abstraction that existed before made this code easier to sort through.

ironfish/src/network/peers/connections/webRtcConnection.ts
dguenther: I think the main potential issues were geo-based and reliance on third-party servers.

ironfish/src/network/peers/peerManager.ts
dguenther: I'll just delete this, I don't think we need it at this point

ironfish/src/network/peers/peer.test.ts
dguenther: I moved this into the setup file

NullSoldier (approved):

PR #249, opened 2021-06-23
Show error message and stack in logs

NullSoldier:


dguenther (approved):
Putting this in master since we're debugging the deployment right now.

PR #253, opened 2021-06-25
chore: Update TypeScript version to 4.3.4 and compiler to ES2020

rohanjadvani:


ironfish/src/account/accounts.ts
```diff
@@ -574,10 +574,10 @@ export class Accounts {
     let confirmedBalance = BigInt(0)
 
     for (const note of notes) {
-      const value = note.note.value()
-      unconfirmedBalance += BigInt(value)
+      const value = note.note.value().valueOf()
```
NullSoldier, line 577: I'm down in another PR, but shouldn't this method just return `bigint` to avoid this call?

ironfish/src/testUtilities/fake/strategy.ts
```diff
@@ -147,8 +147,16 @@ export class TestStrategy
...
+    if (
+      typeof headerSequence !== 'bigint' &&
+      typeof headerSequence !== 'string' &&
+      typeof headerSequence !== 'number'
+    ) {
+      throw new Error(`Invalid sequence type in header`)
+    }
 
-    const sequence = BigInt(header['sequence'])
+    const sequence = BigInt(headerSequence)
```
NullSoldier, line 159: didn't we convert sequences to numbers?

ironfish/src/network/peers/peer.ts
```diff
@@ -289,14 +289,25 @@ export class Peer {
...
+    } else if (wsConnection) {
+      return {
+        type: 'CONNECTING',
+        identity: this.state.identity,
+        connections: {
+          webSocket: wsConnection,
+        },
+      }
+    } else {
+      throw new Error('At least one of webRtcConnection or wsConnection must be defined')
```
dguenther, line 310: Going to look at this a bit and see if there's a better way we can type this

ironfish/src/testUtilities/fake/strategy.ts
rohanjadvani: Yeah, sequence should be a `number` now.

ironfish/src/network/peers/peer.ts
```diff
@@ -289,14 +289,25 @@ export class Peer {
...
+          webRtc: webRtcConnection,
+          webSocket: wsConnection,
+        },
+      }
+    } else if (wsConnection) {
+      return {
+        type: 'CONNECTING',
+        identity: this.state.identity,
+        connections: {
+          webSocket: wsConnection,
```
dguenther, line 306: This should carry forward the `webRtcConnection` as well -- once that's updated, this looks good.

I'm not sure if there was a regression in TS here, or if I'm missing something, or it's not capable of narrowing types in these cases. I'd expect the following two code blocks to be equivalent, but the one with the `OR` operator has a TS error:

1.
```js
if (webRtcConnection) {
    return {
        type: 'CONNECTING',
        identity: this.state.identity,
        connections: {
            webRtc: webRtcConnection,
            webSocket: wsConnection,
        },
    }
}

if (wsConnection) {
    return {
        type: 'CONNECTING',
        identity: this.state.identity,
        connections: {
            webRtc: webRtcConnection,
            webSocket: wsConnection,
        },
    }
}
```

2.
```js
if (webRtcConnection || wsConnection) {
    return {
        type: 'CONNECTING',
        identity: this.state.identity,
        connections: {
            webRtc: webRtcConnection,
            webSocket: wsConnection,
        },
    }
}
```

dguenther (approved):

PR #255, opened 2021-06-28
feat(ironfish-cli): Add `graffiti` flag to start command

rohanjadvani:


ironfish-cli/src/commands/start.ts
```diff
@@ -90,45 +95,46 @@ export default class Start extends IronfishCommand {
...
-      this.sdk.config.setOverride('isWorker', flags.worker)
+    if (forceMining !== undefined && forceMining !== this.sdk.config.get('miningForce')) {
+      this.sdk.config.setOverride('miningForce', forceMining)
     }
-    if (
-      flags.forceMining !== undefined &&
-      flags.forceMining !== this.sdk.config.get('miningForce')
-    ) {
-      this.sdk.config.setOverride('miningForce', flags.forceMining)
+    if (graffiti !== undefined && graffiti) {
```
NullSoldier, line 121: minor question, but now this is the only one that still sets override even if it's the same so the behaviour is slightly different.

rohanjadvani: Woops, totally forgot to finish the rest of the check here. Updated and added a test.

NullSoldier (approved):

NullSoldier (approved):

PR #256, opened 2021-06-28
Add database version and migration errors

NullSoldier:
https://linear.app/ironfish/issue/IRO-699/db-versions-and-migration-error

This adds a new mechanism so we can throw errors when database schemas
change. It will now enforced the DB version and a reset in a less hacky
way. A simple global database version is stored in the DB, and specified
to `db.upgrade(version)`.

If the version mismatches, then an error is thrown. If you are writing a
migration, you can avoid calling this method to run your migration logic
on old code.

### DIFF
--- a/ironfish/src/account/accountsdb.ts
+++ b/ironfish/src/account/accountsdb.ts
@@ -92,14 +92,12 @@ export class AccountsDB {
       key: keyof AccountsDBMeta
       value: AccountsDBMeta[keyof AccountsDBMeta]
     }>({
-      version: 1,
       name: 'meta',
       keyEncoding: new StringEncoding<keyof AccountsDBMeta>(),
       valueEncoding: new JsonEncoding(),
     })
 
     this.accounts = this.database.addStore<{ key: string; value: Account }>({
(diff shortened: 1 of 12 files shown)


ironfish/src/account/accountsdb.ts
```diff
@@ -130,16 +126,20 @@ export class AccountsDB {
...
   }
 
-  async open(_options?: { upgrade?: boolean }): Promise<void> {
+  async open(options: { upgrade?: boolean } = { upgrade: true }): Promise<void> {
     await this.files.mkdir(this.location, { recursive: true })
+
     await this.database.open()
+
+    if (options.upgrade) {
+      await this.database.upgrade(1)
```
dguenther, line 141: Maybe `const` this version somewhere?

dguenther (approved):

ironfish/src/account/accountsdb.ts
NullSoldier: done

rohanjadvani (approved):

PR #262, opened 2021-06-29
Run miner task on WorkerPool

dguenther:
Moves the mineHeader task into our WorkerPool to remove the need for a different WorkerPool dependency. I also restructured the miner into a class to make it easier to mock the WorkerPool in tests.

Fixes IRO-683


rohanjadvani (approved):

ironfish/src/workerPool/worker.ts
```diff
@@ -151,6 +155,38 @@ function handleUnboxMessage({
...
+    ])
+
+    const blockHash = hashBlockHeader(headerBytes)
+
+    if (Target.meets(new Target(blockHash).asBigInt(), target)) {
+      return { type: 'mineHeader', initialRandomness, miningRequestId, randomness }
+    }
+  }
+  return { type: 'mineHeader', initialRandomness }
+}
```
NullSoldier, line 188: Can we move this into the miner? And have the worker pool call it? I wanted to move the guts of these things into their respective modules but until we had something this chunky here it didn't seem too bad until now.

dguenther: 👍 I'll move this

NullSoldier (approved):

PR #263, opened 2021-06-29
chore: Update some old dependencies

rohanjadvani:
This code change updates some out-of-date dependencies.


ironfish/src/network/messageRouters/globalRpc.test.ts
```diff
@@ -162,7 +150,11 @@ describe('Global Rpc', () => {
     await expect(promise).toRejectErrorInstance(CannotSatisfyRequestError)
   })
 
-  it('throws when peers available but none respond', async () => {
+  // TODO: It looks like in Jest 27.0.3, there is a regression introduced when
+  // using `setTimeout` and `setImmediate`. We should re-enable this test once
+  // that regression has been addressed.
+  // eslint-disable-next-line jest/no-disabled-tests
```
dguenther, line 156: Is there a Jest issue for this? Only asking because it's possible we're doing something hacky here that they actually fixed. I remember having to mess with this quite a bit to get it to pass as expected

ironfish/src/mining/miner.ts
```diff
@@ -179,7 +179,7 @@ async function miner(
 
       tasks[randomness] = enqueue(
         piscina,
-        Buffer.from(blockToMineResult.value.bytes),
+        Buffer.from(blockToMineResult.value.bytes.data),
```
dguenther, line 182: This is fine, but I'm guessing this logged a type error somewhere? We should probably simplify `newBlockData` at some point

ironfish/src/network/messageRouters/globalRpc.test.ts
rohanjadvani: I think I'm running into a similar issue described [here](https://github.com/facebook/jest/issues/11607). I'm happy to re-factor this test offline if you think there's a better way we can handle this case.

ironfish/src/mining/miner.ts
rohanjadvani: Yeah, the build was failing as we upgraded some packages due to types

ironfish/src/network/messageRouters/globalRpc.test.ts
dguenther: Ah, nice find. If you want to file a separate issue to investigate refactoring this test, we can address that in a separate branch.

dguenther:
The slow tests look like they might be timing out on something here, and it might need a rebase as well, otherwise looks good to me!

ironfish/src/mining/miner.ts
NullSoldier: What type does this match? I don't see a Buffer.from constuctor that accepts this? Is this matching the interface for one of the array types? Like ArrayBuffer or UInt8Array?

NullSoldier: How was this ever working before? What constructor made this work before? Did bytes match the interface for one of the array types?

ironfish/src/network/messageRouters/globalRpc.test.ts
NullSoldier: Hey, I looked deeply at this problem and it seems like this version of jest is maybe buggy and unstable. There's also no link to the issue here. If setTimeout and setInterval are buggy and broken, it seems like maybe we shouldn't upgrade to them. If something as core as their microtask queue is broken, who knows what else is going wrong. Should we wait to update til this can get fixed, or another version comes out then we can try that?

NullSoldier: We reverted jest here as the resolution.

ironfish/src/mining/miner.ts
NullSoldier: It looks like the old typescript types were wrong, even though buffer does take this as an input.

NullSoldier (approved):

PR #265, opened 2021-06-29
Add verify command to HTTP api to verify build

NullSoldier:
The command just logs and creates a new server, which would have
caught the last error we had in the build.


dguenther (approved):

dguenther:
Some lint failures on this and then it should be good

rohanjadvani (approved):

PR #267, opened 2021-07-02
Rename sequence -> height

NullSoldier:
Most coins use the term "height" while we use "sequence". We're changing  that to be in line with most crypto.

We had to leave two places with sequence, because we discovered block hashes aren't calculated using values, but a JSON string-ified buffer of the header. This means that hashes change when the JSON format changes at all,  including names of fields. We WILL fix this in a future change, but it'll require restarting the testnet.

 * https://github.com/iron-fish/ironfish/blob/master/ironfish/src/primitives/blockheader.ts#L189
 * https://github.com/iron-fish/ironfish/blob/master/ironfish/src/testUtilities/fake/strategy.ts#L151


dguenther (approved):
Looks good once tests pass/rebase

PR #268, opened 2021-07-02
Delete unused fake test code

NullSoldier:


dguenther (approved):
Lint failure but looks good when they all pass!

PR #269, opened 2021-07-02
feat(ironfish-http-api): Create new package for v2 HTTP API 

rohanjadvani:
Bootstraps a NestJS application with Express to serve a health check for the new HTTP API


ironfish-http-api-v2/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "author": "Iron Fish <contact@ironfish.network> (https://ironfish.network)",
+  "repository": "iron-fish/ironfish",
+  "license": "MPL-2.0",
+  "homepage": "https://github.com/iron-fish/ironfish",
+  "keywords": [
+    "http",
+    "api"
+  ],
+  "scripts": {
+    "build": "tsc -b",
```
NullSoldier, line 15: what's the recommended way to dev? Make a change and re-run build and start?

NullSoldier (approved):

ironfish-http-api-v2/package.json
rohanjadvani: For now, yeah we need to re-build and start. I wanted to lay out the groundwork for the server in this PR, but will add auto + hot reloading in another diff.

ironfish-http-api-v2/src/main.ts
```diff
@@ -0,0 +1,29 @@
...
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { NestFactory } from '@nestjs/core'
+import { ExpressAdapter, NestExpressApplication } from '@nestjs/platform-express'
+import compression from 'compression'
+import express from 'express'
+import helmet from 'helmet'
+import http from 'http'
+import { AppModule } from './app.module'
+
+const PORT = process.env.PORT || 8003
```
NullSoldier, line 12: It would be nice if it printed the port out in the logs when it started listening, small detail but useful when there's many projects who use different ports.

NullSoldier (approved):

ironfish-http-api-v2/src/main.ts
rohanjadvani: Sounds good - I added a Linear ticket for this

PR #273, opened 2021-07-02
feat(ironfish-http-api): Add Procfile for Heroku deployments

rohanjadvani:


ironfish-http-api-v2/tsconfig.json
```diff
@@ -1,10 +1,21 @@
...
     "outDir": "build",
     "resolveJsonModule": true,
     "rootDir": "src",
+    "skipLibCheck": true,
+    "sourceMap": true,
+    "strict": true,
+    "target": "ES2020",
     "tsBuildInfoFile": "./build/tsconfig.tsbuildinfo"
   },
   "include": ["src"],
```
NullSoldier, line 21: Why is this one getting materialized and options duplicated? Does it have issues with extending configs?

rohanjadvani: I was thinking we should deploy only this package to Heroku since the entire repository has some dependencies that aren't necessary for the API (ex. wasm). It seems cleaner to have everything deployed directly from this folder (which is why I refactored `tsconfig.json`) rather than pushing a subtree. What do you think?

NullSoldier (approved):

dguenther (approved):
👍 maybe we can consider pulling this out into another repository at some point, since it's independent from the core ironfish project. Or moving website + block explorer into this monorepo

PR #275, opened 2021-07-06
Abstract our partial header serializer

NullSoldier:
This moves the partial header serializer out so it can be tested, it
also upgrades the test so that it doesn't use the fake strategy.


ironfish/src/primitives/blockheader.ts
```diff
@@ -185,30 +185,16 @@ export class BlockHeader<
    * This is used for calculating the hash in miners and for verifying it.[]
```
leanthebean, line 185: could you remove the [] in the comment? think that's a typo

dguenther (approved):
LGTM

PR #280, opened 2021-07-08
feat(http-api): Add deploy script for staging API

rohanjadvani:
Deploy to staging every time a commit has been pushed to `beta`

### DIFF
--- a/.github/workflows/deploy-api-v2.yml
+++ b/.github/workflows/deploy-api-v2.yml
@@ -1,0 +1,20 @@
+name: Deploy HTTP API
+on:
+  push:
+    branches:
+      - beta
+
+jobs:
+  deploy:
+    name: Staging API Deploy
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out Git repository
+        uses: actions/checkout@v2
+      - name: Add Heroku remote
+        run: heroku git:remote --app $HEROKU_APP_NAME
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-api-v2.yml
```diff
@@ -0,0 +1,20 @@
...
+
+    steps:
+      - name: Check out Git repository
+        uses: actions/checkout@v2
+      - name: Add Heroku remote
+        run: heroku git:remote --app $HEROKU_APP_NAME
+        env:
+          HEROKU_APP_NAME: ${{ secrets.HEROKU_APP_NAME }}
+      - name: Push to Heroku
+        run: git push heroku -f `git subtree split --prefix ironfish-http-api-v2 HEAD`:master
```
NullSoldier, line 20: How is this working? Doesn't this require that you build it and push up the built files? or does the node buildpack now automatically install and build?

rohanjadvani: This pushes the sub-folder for the new API, and the `heroku/nodejs` buildpack installs dependencies / runs the build

NullSoldier: nice, that's new since I last used it. We used to have to use multibuildpack with custom buildpack and do the build ourselves.

NullSoldier (approved):

PR #284, opened 2021-07-08
Upgrade block.test.ts with real tests

NullSoldier:
Removed all usage of TestStrategy in block tests


rohanjadvani (approved):
Looks like some minor lint warnings, but lgtm otherwise

PR #290, opened 2021-07-08
feat(http-api): Create migration for `accounts` table

rohanjadvani:


ironfish-http-api-v2/prisma/schema.prisma
```diff
@@ -1,8 +1,17 @@
...
+
 datasource db {
   provider = "postgresql"
   url      = env("DATABASE_URL")
 }
 
-generator client {
-  provider = "prisma-client-js"
+model accounts {
+  id             String   @id @default(cuid()) @db.VarChar
```
NullSoldier, line 11: Is there a reason this is a string and not just an auto incrementing ID? It's not really important I guess but it's a trend I don't get. People just started using strings for ID's rather than normal auto incrementing IDs and I'm not sure why. URLS get really ugly, and harder to remember and link around, and data is harder to debug.

I'm not advocating for change but numbers are just way simpler to me.

NullSoldier (approved):

ironfish-http-api-v2/prisma/schema.prisma
rohanjadvani: Yeah, that works for me. I've generally used ulid/uuids in the past, but I don't have a strong opinion either way.

PR #295, opened 2021-07-12
Add flag for logging per-peer network traffic

dguenther:
Adds a `--logPeerMessages` flag and a `peers:show` command to allow for viewing per-peer network messages. This should make it easier to debug differences in WebRTC signaling behavior.

Some special-casing is required to handle brokered messages and signal message encryption, which makes the logic more complex than it should be.


rohanjadvani (approved):

NullSoldier:
Tested manually, works well.

NullSoldier (approved):

PR #302, opened 2021-07-14
Remove test strategy from verifier.test

NullSoldier:
Also fixed an issue where verifier for transactions was broken. The
error around deserialization wasn't actually happening because
deserialization is lazy. I refactored the code to force deserialization
to occur so the proper error happens and is thrown at the expected
time. The tests were re-written to handle that.

I also moved some code out into fixtures for a common construction case.


dguenther (approved):
Fixture might need to be regenerated for this, but looks good

PR #304, opened 2021-07-16
Remove TestStrategy from blockchain.old.test

NullSoldier:
This removes the last of use of TestStrategy from the block chain, code
that does not test our actual code, but a fake separate strategy. By not
testing out actual code, were not giving ourselves actual security
through tests and missing bugs.

I also notices many tests inside of blockchain.old.test were verifier
tests, not blockchain test. This makes that better.


dguenther (approved):
This looks good to me, just needs a rebase I think

PR #310, opened 2021-07-20
Remove generics

NullSoldier:
https://linear.app/ironfish/issue/IRO-681/remove-e-h-t-se-sh-st-generics-element-hash-transaction

Generics were originally designed to allow our coin to be composable, but added a lot of extra complexity to the code base. We've been working for the past few months to remove some complexity, including materializing generics with the exception of non continer types like MerkleTree. As well as remove TestStrategy which used the generics. The ultimate goal is to make the code base easier to understand to developers and the open source community.


leanthebean (approved):
🚢 🚢 🚢 🚢 🚢 🚢 🚢

PR #314, opened 2021-07-22
Added validation on account import.

ei0haro:
To reduce the risk of importing an invalid account, there should be some basic validation that the keys are correct.


ei0haro:
Today there is no validation on account-import. That means that you can end up with an account with invalid keys.
This pull-request adds simple validation on account import. 

The change is tested manually.

PR #321, opened 2021-07-26
Make backup command not export accounts

NullSoldier:
Change export command to not backup accounts unless you explicitly pass
--accounts flag to it.

### DIFF
--- a/ironfish-cli/src/commands/backup.ts
+++ b/ironfish-cli/src/commands/backup.ts
@@ -24,6 +24,11 @@ export default class Backup extends IronfishCommand {
       allowNo: true,
       description: 'wait for the database to stop being used',
     }),
+    accounts: flags.boolean({
+      default: false,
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/backup.ts
```diff
@@ -24,6 +24,11 @@ export default class Backup extends IronfishCommand {
       allowNo: true,
       description: 'wait for the database to stop being used',
     }),
+    accounts: flags.boolean({
+      default: false,
+      allowNo: true,
+      description: 'wait for the database to stop being used',
```
rohanjadvani, line 30: Should we change this `description`?

NullSoldier: fixed

dguenther (approved):

rohanjadvani (approved):

PR #323, opened 2021-07-27
Adding basic validation on account import

ei0haro:
Adding library to validate Account keys.
Adding basic validation on account import, this to reduce the risk of adding an account with invalid keys.

### DIFF
--- a/ironfish/src/account/accounts.ts
+++ b/ironfish/src/account/accounts.ts
@@ -17,7 +17,7 @@ import { IDatabaseTransaction } from '../storage'
 import { PromiseResolve, PromiseUtils, SetTimeoutToken } from '../utils'
 import { WorkerPool } from '../workerPool'
 import { Account, AccountDefaults, AccountsDB } from './accountsdb'
-
+import { validateAccount } from './validator'
 const REBROADCAST_SEQUENCE_DELTA = 5
 
 type SyncTransactionParams =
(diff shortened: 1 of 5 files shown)


ironfish/src/account/index.ts
```diff
@@ -2,4 +2,5 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 export * from './accounts'
+export * from './account-validator'
```
NullSoldier, line 5: rename this file to validator

ironfish/src/account/accounts.ts
```diff
@@ -789,6 +794,20 @@ export class Accounts {
...
+      ...toImport,
+    }
+
+    this.accounts.set(account.name, account)
+    await this.db.setAccount(account)
+
+    return account
+  }
+
+  validateAccount(toImport: Partial<Account>): void {
```
NullSoldier, line 810: This method feels like it should also be in `accounts/validator.ts`

NullSoldier:
Two small comments. Overall, we did a call on Discord but the summary around this PR is that it gets us halfway there. The Rust code is the only one that knows what the accounts are and if they are valid, so ultimately the rust could should do the validation. This PR does add value though, so we're going to accept it then work on exposing validation in the rust library as a new method `validate_keys(...)`

NullSoldier:
We need to figure out how to get the pipelines running on third party contributor PRs

NullSoldier (approved):

PR #327, opened 2021-08-02
Add API block syncer

NullSoldier:
This adds a new RPC command that follows the chain starting at an
arbitrary head, if any. Once it connects, and blocks are added, the
follow RPC command will stream back the blocks and the block event based
on the chain operations that occur. It only streams back blocks on the
main chain. This will be changed later.

The new command now will accept and API endpoint and token for an Ironfish
API and POST those blocks to the endpoint. That API can be found here,
https://github.com/iron-fish/ironfish-api

It will also pull the API host and endpoint from these environment variables
 * IRONFISH_API_HOST
 * IRONFISH_API_TOKEN 

```bash
> ironfish chain:sync
Connecting to node and fetching head...
connected: 00000004b7b8ecc6e945096bc809dde30ac5996e4c09ee1eada83439e6200210 - 44966 - 31h 35m 31s
connected: 0000000e8e5d297073ffddaa4ad2305836a18c5f5e1b8b20f84a3834488d337a - 44967 - 31h 35m 30s
connected: 00000003ca24388c238e49dc2d3f85540807eb50040350d564d1a0262f6953f9 - 44968 - 31h 19m 9s
connected: 0000000e8f3d5d53b34616671df9f38fad46dfdd2e3bb01efc9537b2ea48713f - 44969 - 31h 19m 8s
connected: 00000003f6b49bf564e4ab5f5b09813d694c37189417fd0e6918e57f23eb75d4 - 44970 - 31h 4m 6s
```

### DIFF
--- a/ironfish-cli/src/api.ts
+++ b/ironfish-cli/src/api.ts
@@ -1,0 +1,64 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import axios, { AxiosRequestConfig } from 'axios'
+import { FollowChainStreamResponse } from 'ironfish'
+
+export class IronfishApi {
+  host: string
+  token: string
+
+  constructor(host: string, token = '') {
+    if (host.endsWith('/')) {
+      host = host.slice(0, -1)
+    }
+
+    this.host = host
+    this.token = token
+  }
+
+  async head(): Promise<string | null> {
+    const response = await axios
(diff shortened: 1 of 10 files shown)


ironfish-cli/src/api.ts
```diff
@@ -0,0 +1,64 @@
...
+      host = host.slice(0, -1)
+    }
+
+    this.host = host
+    this.token = token
+  }
+
+  async head(): Promise<string | null> {
+    const response = await axios
+      .get<{ hash: string }>('https://api-staging.ironfish.network/blocks/head')
```
rohanjadvani, line 23: Can we replace this with `this.host`?

NullSoldier: fixed

rohanjadvani (approved):

PR #335, opened 2021-08-17
Send transactions back in block-info command

NullSoldier:


rohanjadvani (approved):

ironfish/src/rpc/routes/chain/getBlockInfo.ts
```diff
@@ -42,19 +62,41 @@ router.register<typeof GetBlockInfoRequestSchema, GetBlockInfoResponse>(
...
     if (!header) {
       throw new ValidationError(`No block with hash ${request.data.hash}`)
     }
 
+    const block = await node.chain.getBlock(header)
+    if (!block) {
+      throw new ValidationError(`No block with hash ${request.data.hash}`)
+    }
+
+    const transactions: GetBlockInfoResponse['block']['transactions'] = []
```
brekk, line 76: For my own understanding, why is the type here in brackets? (I am assuming this is a TS thing I'm unfamiliar with.)

ironfish/src/rpc/routes/chain/getBlockInfo.ts
```diff
@@ -42,19 +62,41 @@ router.register<typeof GetBlockInfoRequestSchema, GetBlockInfoResponse>(
...
+    })
+
     request.status(200).end({
       block: {
         graffiti: header.graffiti.toString('hex'),
         hash: request.data.hash,
         previousBlockHash: header.previousBlockHash.toString('hex'),
         sequence: Number(header.sequence),
         timestamp: header.timestamp.valueOf(),
+        transactions: transactions,
```
brekk, line 99: Again, might pitch the object property value shorthand here as well.
```suggestion
        transactions,
```

ironfish/src/rpc/routes/chain/getBlockInfo.ts
NullSoldier: If you have a type, you can reference inner types using brackets in Typescript without redeclaring them.

```typescript
type Foo = {
    bar: {
        baz: number
    }
}

type Baz = Foo['bar']['baz']
// Type of baz is number
```

ironfish/src/rpc/routes/chain/getBlockInfo.ts
NullSoldier: In reference to the other conversation here, https://github.com/iron-fish/ironfish/pull/336#discussion_r691532460

PR #336, opened 2021-08-17
Improve mined command usability

NullSoldier:
The mined command was good, but very slow, this makes a few
improvements

 - Processes and scans 100 blocks at a time now
 - Show time estimate using time estimater
 - Make iterateTransaction method more robust
 - Add option to also scan and include forks

### DIFF
--- a/ironfish-cli/src/commands/miners/mined.ts
+++ b/ironfish-cli/src/commands/miners/mined.ts
@@ -2,7 +2,14 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import cli from 'cli-ux'
-import { AsyncUtils, GENESIS_BLOCK_SEQUENCE, MathUtils, oreToIron } from 'ironfish'
+import {
+  AsyncUtils,
+  GENESIS_BLOCK_SEQUENCE,
+  MathUtils,
+  Meter,
+  oreToIron,
+  TimeUtils,
+} from 'ironfish'
 import { parseNumber } from '../../args'
 import { IronfishCommand } from '../../command'
 import { RemoteFlags } from '../../flags'
@@ -46,10 +53,14 @@ export class MinedCommand extends IronfishCommand {
     const { start, stop } = await AsyncUtils.first(stream.contentStream())
     this.log(`Exporting mined block from ${start} -> ${stop}`)
 
+    const speed = new Meter()
+
     const progress = cli.progress({
-      format: 'Exporting blocks: [{bar}] {value}/{total} {percentage}% | ETA: {eta}s',
+      format:
+        'Exporting blocks: [{bar}] {value}/{total} {percentage}% | ETA: {estimate} | SEQ {sequence}',
     }) as ProgressBar
 
+    speed.start()
     progress.start(stop - start + 1, 0)
 
     for await (const { sequence, block } of stream.contentStream()) {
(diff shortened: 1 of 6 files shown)


ironfish-cli/src/commands/miners/mined.ts
```diff
@@ -71,7 +82,12 @@ export class MinedCommand extends IronfishCommand {
         )
       }
 
-      progress.update(sequence - start)
+      speed.add(1)
+
+      progress.update(sequence - start, {
+        estimate: TimeUtils.renderEstimate(sequence - start, stop - start, speed.rate5s),
+        sequence: sequence,
```
brekk, line 89: Trivial, but you could use object property value shorthand here and just do 
```suggestion
        sequence,
```

ironfish/src/rpc/routes/mining/exportMined.ts
```diff
@@ -57,44 +59,64 @@ router.register<typeof ExportMinedStreamRequestSchema, ExportMinedStreamResponse
...
 
+    let promise = Promise.resolve()
+    let waiting = 0
+
     for (let i = start; i <= stop; ++i) {
-      const headers = await node.chain.getHeadersAtSequence(i)
+      promise = promise.then(async () => {
+        const headers = scanForks
+          ? await node.chain.getHeadersAtSequence(i)
+          : [await node.chain.getHeaderAtSequence(i)]
```
brekk, line 78: Again, trivial, but since you're doing the same work either way, would it make sense to do:
```ts
const sequenceHeaders = await node.chain.getHeadersAtSequence(i)
const headers = scanForks ? sequenceHeaders : [sequenceHeaders]
```
?

rohanjadvani (approved):

brekk (approved):
With the caveat that I'm not well-versed in this codebase (and thus my opinion carries less weight), this LGTM.

ironfish-cli/src/commands/miners/mined.ts
NullSoldier: I'm willing to agree to some type of standard but here's my general programming style for when I use that or not...

 - Use shorthand if all properties are shorthand
 - If there are any non shorthand properties, I expand all of them (usually)

I'm gonna merge this but let's keep the conversation going. What is your preference for using short hand construction?

ironfish/src/rpc/routes/mining/exportMined.ts
NullSoldier: They are actually 2 separate functions, one is plural one is not. The plural one pulls all headers from sequence, the non plural one only pulls the one from the main chain. It's a bit sneaky, but take a look.

ironfish-cli/src/commands/miners/mined.ts
brekk: I think my preference is to spend less time redeclaring things when there's ambient variables you can soak up. I'm curious why you expand all of them if any are not-shorthand? For me each declaration is a cognitive step and I try to name things as I go so that I can ideally use shorthand everywhere, i.e.

```js
// rename from generic to specific in one
Object.fromEntries(blah).map(([contextSpecificName, value]) => ({contextSpecificName, value}))
// inline-rename
listOfObjects.map(({key: contextSpecificName, value}) => ({contextSpecificName, value}))
```

ironfish/src/rpc/routes/mining/exportMined.ts
brekk: Ah! I totally missed that `s` in there.

ironfish-cli/src/commands/miners/mined.ts
NullSoldier: Mainly the reason is traditional javascript reasons. Using tools to rename variables might inadvertently also change payloads without realizing it. It can still happen but it's less of a bug in typescript. Renaming foo below, would cause a bug if you also didn't realize this changes the API's payload returned.

```typescript
function request(res) {
    const foo = 2
    res.send({ foo })
}
```

In any case, I'm down to just prefer short hand versions. I'll update this PR and generally use that going forward if you have a preference.

NullSoldier: Updated

PR #337, opened 2021-08-18
Add test for transaction hashes

NullSoldier:
Add a test to make sure miners fees don't have the same transaction
hash.

### DIFF
--- a/ironfish/src/primitives/transaction.test.slow.ts
+++ b/ironfish/src/primitives/transaction.test.slow.ts
@@ -1,0 +1,29 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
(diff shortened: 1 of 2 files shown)


brekk (approved):
:shipit:

ironfish/src/primitives/transaction.test.slow.ts
```diff
@@ -0,0 +1,29 @@
...
+    const transactionB = await nodeTest.strategy.createMinersFee(
+      BigInt(0),
+      1,
+      account.spendingKey,
+    )
+
+    const hashA = transactionA.transactionHash()
+    const hashB = transactionB.transactionHash()
+
+    expect(hashA.equals(hashB)).toBe(false)
```
rohanjadvani, line 27: Could we do 
```ts
expect(hashA).not.toBe(hashB);
```

NullSoldier: toBe uses reference equalilty, so if you want to check that the buffers contents are actually equal or not we need to use Buffer.equals to check that the byte sets are the same or not.

rohanjadvani (approved):

PR #340, opened 2021-08-27
IRO-1020: Fix Account Export Flow

deekerno:
Changelog
---
- Account information is now exported to a default filename if a directory is given as the export path.

### DIFF
--- a/ironfish-cli/src/commands/accounts/export.ts
+++ b/ironfish-cli/src/commands/accounts/export.ts
@@ -3,6 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { flags } from '@oclif/command'
 import fs from 'fs'
+import path from 'path'
 import jsonColorizer from 'json-colorizer'
 import { IronfishCommand } from '../../command'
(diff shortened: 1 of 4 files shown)


ironfish-cli/src/commands/accounts/export.ts
```diff
@@ -47,8 +50,34 @@ export class ExportCommand extends IronfishCommand {
...
+            this.sdk.fileSystem.join(resolved, `ironfish-${account}.txt`),
+            output,
+          )
+        } else {
+          this.log(`There is already an account backup at ${exportPath}`)
+          const confirmed = await cli.confirm(
+            `\nOverwrite the account backup with new file?\nAre you sure? (Y)es / (N)o`,
+          )
+          if (!confirmed) {
+            this.exit(1)
```
NullSoldier, line 67: Just a side note, but you don't need to put the write file in an else statement because this.exit() returns never, because it calls process.exit()

deekerno: Oh yeah, good catch for sure.

NullSoldier (approved):

PR #343, opened 2021-09-13
IRO-1111: Add pool speed status to workers info command

yuryironfish:
Add pool speed status to worker get status command
image

### DIFF
--- a/ironfish/src/rpc/routes/workers/getStatus.ts
+++ b/ironfish/src/rpc/routes/workers/getStatus.ts
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import * as yup from 'yup'
+import { MathUtils } from '../../../utils'
 import { WorkerPool } from '../../../workerPool'
 import { ApiNamespace, router } from '../router'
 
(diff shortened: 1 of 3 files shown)


ironfish/src/rpc/routes/workers/getStatus.ts
```diff
@@ -55,15 +68,42 @@ router.register<typeof GetWorkersStatusRequestSchema, GetWorkersStatusResponse>(
...
+      request.stream({
+        started: node.workerPool.started,
+        workers: node.workerPool.workers.length,
+        executing: node.workerPool.executing,
+        queued: node.workerPool.queued,
+        capacity: node.workerPool.capacity,
+        change: MathUtils.round(node.workerPool.change?.rate5s ?? 0, 2),
+        speed: MathUtils.round(node.workerPool.speed?.rate5s ?? 0, 2),
+        jobs,
+      })
```
NullSoldier, line 106: Can we move this logic into getJobs() and rename it so it isn't duplicated in 3 spots? Kind of like how we do in getStatus() here? https://github.com/iron-fish/ironfish/blob/master/ironfish/src/rpc/routes/node/getStatus.ts#L150-L194

yuryironfish: Fixe3d.

ironfish/src/rpc/routes/workers/getStatus.ts
```diff
@@ -72,21 +85,30 @@ router.register<typeof GetWorkersStatusRequestSchema, GetWorkersStatusResponse>(
...
-    const job = pool.stats.get(name)
+    const job = node.workerPool.stats.get(name)
 
     if (job) {
       result.push({ name: name, ...job })
     }
   }
-
-  return result
+  const response: GetWorkersStatusResponse = {
```
NullSoldier, line 103: Lets just return this

NullSoldier (approved):

PR #344, opened 2021-09-14
IRO-433: Add support memo field into accounts pay command

yuryironfish:


ironfish-cli/src/commands/accounts/pay.ts
```diff
@@ -54,6 +58,7 @@ export class Pay extends IronfishCommand {
     let fee = flags.fee as unknown as number
     let to = flags.to
     let from = flags.account
+    const message = flags.memo as string
```
NullSoldier, line 61:
```suggestion
    const message = flags.memo
```

NullSoldier, line 61: this already comes in as undefined | string so you don't need to typecast it

yuryironfish: fixed.

NullSoldier (approved):

PR #346, opened 2021-09-16
IRO-523: Support consistent identities by persisting identity

yuryironfish:


ironfish/src/fileStores/internal.ts
```diff
@@ -6,10 +6,12 @@ import { KeyStore } from './keyStore'
 
 export type InternalOptions = {
   isFirstRun: boolean
+  secretKey: string
```
NullSoldier, line 9: Our users might think this is their wallet secret key. What do you think about something more specific like `networkIdentity`?

ironfish-cli/src/commands/start.ts
```diff
@@ -224,4 +243,20 @@ export default class Start extends IronfishCommand {
...
+    const networkIdentity = this.sdk.internal.get('networkIdentity')
+    if (
+      !this.sdk.config.get('generateNewIdentity') &&
+      networkIdentity !== undefined &&
+      networkIdentity.length > 31
+    ) {
+      const hex = Uint8Array.from(Buffer.from(networkIdentity, 'hex'))
+      privateIdentity = tweetnacl.box.keyPair.fromSecretKey(hex)
+    } else {
+      privateIdentity = tweetnacl.box.keyPair()
```
NullSoldier, line 258: I think you can delete this line of code and this all still works.

NullSoldier (approved):

PR #349, opened 2021-09-21
IRO-1123: Send Transactions as a Part of Block Sync Payload

deekerno:
Changelog
---
- Add transactions array to response object
- Tweak process for converting commitments and nullifiers to hex strings
- Add `main`, `transactions`, and `transactions_count` to block level sync payload
- Add `fee` to transaction level sync payload
- Small grammar correction

### DIFF
--- a/ironfish-cli/src/api.ts
+++ b/ironfish-cli/src/api.ts
@@ -37,6 +37,7 @@ export class IronfishApi {
       previous_block_hash: block.previous,
       difficulty: block.difficulty,
       graffiti: block.graffiti,
+      main: block.main,
       transactions: block.transactions,
       transactions_count: block.transactions.length,
     }))
--- a/ironfish-cli/src/api.ts
+++ b/ironfish-cli/src/api.ts
@@ -37,7 +37,8 @@ export class IronfishApi {
       previous_block_hash: block.previous,
       difficulty: block.difficulty,
       graffiti: block.graffiti,
-      transactions_count: 0,
+      transactions: block.transactions,
+      transactions_count: block.transactions.length,
     }))
 
     const options = this.options({ 'Content-Type': 'application/json' })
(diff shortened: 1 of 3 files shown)


yuryironfish (approved):
LGTM

ironfish-cli/src/api.ts
```diff
@@ -37,7 +37,9 @@ export class IronfishApi {
       previous_block_hash: block.previous,
       difficulty: block.difficulty,
       graffiti: block.graffiti,
-      transactions_count: 0,
+      main: block.main,
+      transactions: block.transactions,
+      transactions_count: block.transactions.length,
```
NullSoldier, line 42: Why are we sending this up? Isn't this just a derived value from transactions?

NullSoldier (approved):

ironfish/src/rpc/routes/chain/followChain.ts
```diff
@@ -74,35 +112,60 @@ router.register<typeof FollowChainStreamRequestSchema, FollowChainStreamResponse
...
-      send(header, 'disconnected')
+    const onRemove = async (header: BlockHeader) => {
+      const block = await node.chain.getBlock(header)
+      Assert.isNotNull(block)
+      void send(block, 'disconnected')
     }
 
     const onFork = (block: Block) => {
-      send(block.header, 'fork')
+      void send(block, 'fork')
```
NullSoldier, line 168: This could cause problems but we can just see what happens for now if we don't await this.

ironfish-cli/src/api.ts
deekerno: Good point; changed it here and implemented the requisite change in the API.

ironfish/src/rpc/routes/chain/followChain.ts
deekerno: Changed to `await` and everything continues to work as expected.

PR #350, opened 2021-09-23
WIP: Refactor mining

NullSoldier:
This is WIP and still under construction, but this is designed to solve 3 different problems

- The architecture of the mining director does not make it easier for people to report the current status of their miner, and therefore they have no idea if its mining or not. In fact, often when it says "Mining a block" it's actually not mining at all
- It's impossible for the director to know what its miners are doing and what their status / hash speed is because there is no concept of a miner, only block streams.
- Multiple threads all use the same search space because the mining director does not distribute search space at all.
- Having to "Prime" the miner is a waste of CPU and doesn't need to be done

It will do this by using a small stratum style protocol.
- Miners connect over RPC and get an ID and a token to authenticate requests with
- MinerJobs are distributed to miners
- If the miner doesn't have a job or needs more jobs because it finished it can prompt the director for a new job.
- Search space is distributed as a random and randomMax value which correlate to a number of bits of search space assigned to each miner
- Miners are trusted in this code in terms of assignments

**Tasks:**
- When miner requests assign new search space
- Add stats API for miner to submit stats like hash speed and determine miner job progress
- Director should auto send job when miner hash space is low
- Director should send new job with flush = true when target changes


Curay1954:
Sumcoin index, I did@everything you explained@in your message to try to get my coins back but didn’t work. Here is the problem my coins were moved to another slice wallet that I don’t know who is the owner and I don’t know who did the transaction on July 29
Since that date I haven’t got a response from anybody and I don’t know where is my money. All I know@is that they asked 45600 in fees to release my money to the@bank which I thought it was outrageous but the got the fees and still no money .
How can I get my coins back?’what do I need to do?
Raymond Bailly
Curay1954

Curay1954:
Sumcoin index, I did@everything you explained@in your message to try to get my coins back but didn’t work. Here is the problem my coins were moved to another slice wallet that I don’t know who is the owner and I don’t know who did the transaction on July 29
Since that date I haven’t got a response from anybody and I don’t know where is my money. All I know@is that they asked 45600 in fees to release my money to the@bank which I thought it was outrageous but the got the fees and still no money .
How can I get my coins back?’what do I need to do?

Raymond Bailly
Curay1954

PR #355, opened 2021-09-28
feat(ironfish): Move transaction verification to mem pool

rohanjadvani:


ironfish/src/consensus/verifier.ts
```diff
@@ -176,27 +168,12 @@ export class Verifier {
       throw new Error('Transaction cannot deserialize')
     }
```
NullSoldier, line 169: This is a memory leak now, if you look here were not using `returnReference()` if no exception is thrown. We could move this to a finally clause and remove it in the catch()

ironfish/src/memPool/memPool.ts
```diff
@@ -42,7 +42,12 @@ export class MemPool {
   /**
    * Accepts a transaction from the network
    */
-  acceptTransaction(transaction: Transaction): boolean {
+  async acceptTransaction(transaction: Transaction): Promise<boolean> {
+    const { valid } = await this.chain.verifier.verifyTransaction(transaction)
+    if (!valid) {
```
NullSoldier, line 47: We should add a log here about the invalid transaction and the reason

NullSoldier: We should add ban score for peers here later, I just haven't thought through all the details yet.

ironfish/src/consensus/verifier.ts
rohanjadvani: Fixed

ironfish/src/memPool/memPool.ts
```diff
@@ -42,15 +42,24 @@ export class MemPool {
...
     const hash = transaction.transactionHash()
+    const renderedHash = hash.toString('hex')
+    if (!valid) {
+      if (reason) {
+        this.logger.debug(`Invalid transaction '${renderedHash}': ${reason}`)
+      }
+      return false
+    }
+
     if (this.transactions.has(hash)) {
```
NullSoldier, line 56: We should re-order these checks

NullSoldier (approved):

PR #357, opened 2021-09-28
WIP: IRO-1130: Update max and target peers to 8

yuryironfish:
Note from @NullSoldier : This cannot merge yet because with connection instability it creates an environment that's hard to establish connections because connecting connections take up slots in the connection limit.


.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
+{
```
leanthebean, line 1: what is this? looks like visual studio code preferences?

dguenther:
It'll be a bit before we're ready for this -- we need to fix the connection attempts timing out first.

PR #376, opened 2021-10-05
WIP: IRO-1170: Add isMinersFee method to transaction class

yuryironfish:


ironfish/src/primitives/transaction.ts
```diff
@@ -92,6 +92,10 @@ export class Transaction {
     })
   }
 
+  isMinersFee(): boolean {
+    return this.spendsLength() === 0 && this.notesLength() === 1
```
NullSoldier, line 96: oh I think we can also check if the transactionFee() is negative too actually

yuryironfish: so the condition should be:

```
spendsLength == 0 AND
notesLength == 1 AND
transactionFee <= 0

```
right?

PR #393, opened 2021-10-18
IRO-1179 Fix invalid spend error when reorganizing fork

dguenther:
Fixes an error where spend commitments/rootHashes were being validated against the note tree while syncing a fork, but before the tree has been reorganized to that of the fork.

Will add a test for this to make sure it's fixed as expected.


deekerno:
🔥 🔥 🔥

PR #408, opened 2021-10-27
IRO-1193: Implement PeerAddress and AddressManager - Part 1

deekerno:
This first part of the PR is intended to provide the foundation for peer persistence while the following part will implement more complex logic.

Changelog
---
- Implemented PeerAddress type
- Implemented HostsStore to persist hosts to storage
- Implemented PeerAddressManager to hold possible addresses for use in PeerManager
- Adjusted PeerNetwork and PeerManager instantiations to include PeerAddressManager
- Implemented tests for PeerAddressManager


ironfish/src/network/peers/peerManager.ts
```diff
@@ -779,9 +788,11 @@ export class PeerManager {
     }
   }
 
-  start(): void {
-    this.distributePeerListHandle = setInterval(() => this.distributePeerList(), 5000)
-    this.disposePeersHandle = setInterval(() => this.disposePeers(), 2000)
+  async start(): Promise<void> {
+    await Promise.allSettled([
```
dguenther, line 792: I don't think this is necessary now

ironfish/src/network/peers/peerAddressManager.test.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import { mockHostsStore } from '../testUtilities'
+import { PeerAddressManager } from './peerAddressManager'
+
+jest.useFakeTimers()
+
+describe('PeerAddressManager', () => {
+  it('constructor load hosts from HostsStore', () => {
+    const peerAddressManager = new PeerAddressManager(mockHostsStore())
+    expect(peerAddressManager.priorConnectedPeerAddresses).toMatchObject([
```
dguenther, line 14: I suppose you could add a test for `possiblePeerAddresses` as well

ironfish/src/network/peers/peerManager.ts
deekerno: 👍🏽

ironfish/src/network/peers/peerAddressManager.test.ts
deekerno: 👍🏽

dguenther:
Okay cool, this looks like a good start! I'm going to hit approve on it, but let's get feedback from @NullSoldier before merging in case he has input on usage of KeyStore.

dguenther (approved):

ironfish/src/fileStores/hostsStore.ts
```diff
@@ -0,0 +1,22 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { FileSystem } from '../fileSystems'
+import { PeerAddress } from '../network/peers/peerAddress'
+import { KeyStore } from './keyStore'
+
+export type HostsOptions = {
+  priorConnectedPeers: PeerAddress[]
```
NullSoldier, line 9: NIT: I think the word `Connected` indicates `Prior` already. What do you think about simplifying this name?

ironfish/src/fileStores/index.ts
```diff
@@ -4,3 +4,4 @@
 export * from './config'
 export * from './fileStore'
 export * from './internal'
+export * from './hostsStore'
```
NullSoldier, line 7: the base storage is called FileStore because it's an abstract store, but none of the other stores like `config` or `internal` have this `Store` suffix. What do you think about making them consistent?

ironfish/src/network/peerNetwork.ts
```diff
@@ -396,12 +399,14 @@ export class PeerNetwork {
...
-    this.webSocketServer?.close()
-    this.updateIsReady()
+  async stop(): Promise<void> {
+    await Promise.allSettled([
+      (this.started = false),
+      this.peerConnectionManager.stop(),
+      this.peerManager.stop(),
+      this.webSocketServer?.close(),
+      this.updateIsReady(),
+    ])
```
NullSoldier, line 409: This is creating a bug now, if any of these fail to stop the stop() will resolve successfully.

Also why is the non async thing inside of this promise?

ironfish/src/fileStores/index.ts
deekerno: Sure, will do.

ironfish/src/fileStores/hostsStore.ts
deekerno: How about `priorPeers`?

ironfish/src/network/peerNetwork.ts
deekerno: That was probably a holdover from a prior implementation that I had attempted; I'll remove.

ironfish/src/network/peers/peerAddressManager.ts
```diff
@@ -0,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { HostsStore } from '../../fileStores'
+import { PeerAddress } from './peerAddress'
+
+export class PeerAddressManager {
+  hostsStore: HostsStore
```
NullSoldier, line 8: maybe just call this hosts to keep it simple?

deekerno: 👍🏽

ironfish/src/network/peers/peerManager.ts
```diff
@@ -83,6 +84,12 @@ export class PeerManager {
    */
   peers: Array<Peer> = []
 
+  /**
+   * PeerAddressManager allows the storage and persistence of addresses that can
+   * map to peers.
+   */
```
NullSoldier, line 90: This is comment is describing the type, not the instance here. I feel like this comment should be on the type itself instead of here.

ironfish/src/network/peers/peerAddressManager.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  constructor(hostsStore: HostsStore) {
```
NullSoldier, line 10: ok, if this thing isn't the authority on its persistence, who is? The SDK, the node?

ironfish/src/network/peers/peerManager.ts
```diff
@@ -144,13 +151,15 @@ export class PeerManager {
...
     localPeer: LocalPeer,
+    peerAddressManager: PeerAddressManager,
     logger: Logger = createRootLogger(),
     metrics?: MetricsMonitor,
     maxPeers = 10000,
     targetPeers = 50,
     logPeerMessages = false,
   ) {
     this.logger = logger.withTag('peermanager')
+    this.peerAddressManager = peerAddressManager
```
NullSoldier, line 162: nit: can we just call this addresses or something simple? or addressManager?

ironfish/src/network/peers/peerAddressManager.ts
deekerno: Yep, there are modifications to both the node and SDK.

ironfish/src/network/peers/peerManager.ts
deekerno: 👍🏽

ironfish/src/network/peers/peerManager.ts
deekerno: 👍🏽

ironfish/src/node.ts
```diff
@@ -102,6 +107,7 @@ export class IronfishNode {
       chain: chain,
       strategy: strategy,
       metrics: this.metrics,
+      peerAddressManager: new PeerAddressManager(this.hosts),
```
NullSoldier, line 110: Why does the node keep track of the hosts? What does it use it for? Is there other systems later besides networking that will use the hosts?

ironfish/src/node.ts
```diff
@@ -156,6 +164,11 @@ export class IronfishNode {
       await internal.load()
     }
 
+    if (!hosts) {
+      hosts = new HostsStore(files, dataDir)
+      await hosts.load()
+    }
+
```
NullSoldier, line 171: Again, it looks like this store is going to be node level instead of being networking related. What's the larger vision around this hosts list store? Will block chain use it? Will mem pool use it for something?

NullSoldier (approved):

PR #409, opened 2021-10-28
Don't rebroadcast invalid transactions

NullSoldier:
This prevents a very specific problem where we create a transaction on a
fork that is no longer valid on the main chain. The issue is that we'll
permanently rebroadcast this transaction but no one will think it's
valid. This causes the transaction DDOS mechanism that is taking down
the network. We don't delete these transactions because they may become
valid if we re-org back to our original chain.

This is NOT the permanent solution because it will check the
transaction forever, and never unmark the notes as spent that were spent
on this fork.

The correct solution is...

1. Transaction expiration to eventually remove this transaction
2. Properly unmark notes as spent by deleting the transaction that was
   created on a fork when the block reorg occurs.

These fixes are coming in later patches.

### DIFF
--- a/ironfish/src/account/accounts.test.ts
+++ b/ironfish/src/account/accounts.test.ts
@@ -1,7 +1,13 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { createNodeTest } from '../testUtilities'
+import { genesisBlockData, GENESIS_BLOCK_SEQUENCE, VerificationResultReason } from '..'
+import {
+  createNodeTest,
+  useAccountFixture,
+  useMinerBlockFixture,
+  useTxFixture,
+} from '../testUtilities'
 import { makeBlockAfter } from '../testUtilities/helpers/blockchain'
 
 describe('Accounts', () => {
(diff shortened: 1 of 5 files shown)


ironfish/src/account/accounts.test.ts
```diff
@@ -44,4 +50,78 @@ describe('Accounts', () => {
     expect(node.accounts['headHash']).toEqual(blockB3.header.hash.toString('hex'))
     expect(getTransactionsSpy).toBeCalledTimes(8)
   }, 8000)
+
+  it.only('should reproduce issue', async () => {
```
leanthebean, line 54: can we make this comment a bit more descriptive? e.g. what issue?

ironfish/src/account/accounts.ts
```diff
@@ -746,15 +753,30 @@ export class Accounts {
       // TODO: This algorithm suffers a deanonim attack where you can watch to see what transactions node continously
```
leanthebean, line 753: deanonim == deanonymization?

ironfish/src/memPool/memPool.ts
```diff
@@ -118,18 +118,19 @@ export class MemPool {
...
         return false
       }
     }
+
     const validity = await transaction.verify()
     if (!validity.valid) {
       return false
     }
 
+    // it's faster to check if spends have been seen or not, so do that first
```
leanthebean, line 133: looks like we're not doing that first anymore

NullSoldier: we never were, deleted comment.

ironfish/src/account/accounts.ts
NullSoldier: Typo is unrelated to this PR, but I'll just fix it here this time.

ironfish/src/account/accounts.test.ts
NullSoldier: changed the test name and removed the only

leanthebean (approved):
lfg

deekerno (approved):

PR #416, opened 2021-10-29
refactor(cli,ironfish,wasm): Remove redundant `transaction` prefix from `transactionHash`

rohanjadvani:


dguenther:
We might have to bump the version on that codecov action -- it hasn't run in a while 😅

dguenther (approved):

PR #417, opened 2021-10-29
feat(ironfish): Add priority queue for transaction fees in mempool

rohanjadvani:


ironfish/src/memPool/memPool.test.ts
```diff
@@ -50,18 +50,31 @@ describe('MemPool', () => {
...
-    it('returns transactions from the node mempool', async () => {
+    it('returns transactions from the node mempool sorted by fees', async () => {
       const { node } = nodeTest
       const { accounts, memPool } = node
       const accountA = await useAccountFixture(accounts, 'accountA')
       const accountB = await useAccountFixture(accounts, 'accountB')
-      const { transaction } = await useBlockWithTx(node, accountA, accountB)
-
-      await memPool.acceptTransaction(transaction)
+      const { transaction: firstTransaction } = await useBlockWithTx(node, accountA, accountB)
```
NullSoldier, line 58: this is such a tiny nit, but in other places we usually name this with a suffix like the accounts above so `transactionA` or `transaction1` and `transactionB` or `transaction2`

ironfish/src/memPool/memPool.ts
```diff
@@ -20,6 +27,11 @@ export class MemPool {
   constructor(options: { strategy: Strategy; chain: Blockchain; logger?: Logger }) {
     const logger = options.logger || createRootLogger()
 
+    this.queue = new FastPriorityQueue<TransactionFeeAndHash>(
```
NullSoldier, line 30:
```suggestion
    this.queue = new FastPriorityQueue<TransactionFeeAndHash>((a, b) => a.fee > b.fee)
```

ironfish/src/memPool/memPool.ts
```diff
@@ -75,7 +94,9 @@ export class MemPool {
     let deletedTransactions = 0
 
     for (const transaction of block.transactions) {
-      this.transactions.delete(transaction.transactionHash())
+      const hash = transaction.transactionHash()
+      this.transactions.delete(hash)
+      this.queue.removeOne((t) => t.hash.equals(hash))
```
NullSoldier, line 99: this could be done faster with `queue.remove(value)`, there would be a bit more work to do that but we can talk through that.

ironfish/src/memPool/memPool.ts
```diff
@@ -20,6 +27,11 @@ export class MemPool {
   constructor(options: { strategy: Strategy; chain: Blockchain; logger?: Logger }) {
     const logger = options.logger || createRootLogger()
 
+    this.queue = new FastPriorityQueue<TransactionFeeAndHash>(
+      (firstTransaction, secondTransaction) => {
+        return firstTransaction.fee > secondTransaction.fee
```
NullSoldier, line 32: For this comparator what happens if two fees are equal, is this stable sort? Should we have a second ordering by hash?

NullSoldier:
This looks pretty good, it makes me think we should perhaps make a new MempoolEntry type that replaces TransactionWithFeeAndHash

ironfish/src/memPool/memPool.ts
```diff
@@ -43,8 +55,14 @@ export class MemPool {
   }
 
   *get(): Generator<Transaction, void, unknown> {
-    for (const transaction of this.transactions.values()) {
-      yield transaction
+    while (!this.queue.isEmpty()) {
+      const feeAndHash = this.queue.poll()
+      if (feeAndHash) {
```
NullSoldier, line 60: This should be an assert I think, because it can't be falsy.

ironfish/src/memPool/memPool.ts
```diff
@@ -43,8 +55,14 @@ export class MemPool {
...
+        const transaction = this.transactions.get(feeAndHash.hash)
+        if (transaction) {
```
NullSoldier, line 62: I'd argue that this should be an assert to, because if it's falsy were ignoring a clear bug.

NullSoldier (approved):

PR #421, opened 2021-11-01
Add test for ChainProcessor

NullSoldier:


deekerno (approved):

ironfish/src/chainProcessor.test.ts
```diff
@@ -0,0 +1,67 @@
...
+import { makeBlockAfter } from './testUtilities/helpers/blockchain'
+
+describe('ChainProcessor', () => {
+  const nodeTest = createNodeTest()
+
+  it('processes chain', async () => {
+    const { strategy, chain } = nodeTest
+    strategy.disableMiningReward()
+
+    // G -> A1
```
yuryironfish: I think that should be:
// G -> A1 -> A2 -> A3
       -> B1  -> B2

yuryironfish (approved):
APPROVED

ironfish/src/chainProcessor.test.ts
NullSoldier: Yeah, updated the comment.

PR #424, opened 2021-11-04
feat(rust,wasm): Add `expiration_sequence` to Transactions

rohanjadvani:


dguenther (approved):
Looks good! This is probably going to require regenerating fixtures + the genesis block, which is a bit of a pain unfortunately

deekerno (approved):
Rust code looks good. We should probably look into removing those `.unwrap()`s at some point in the future.

PR #425, opened 2021-11-05
Parameterize block mining time and free time range

yuryironfish:
This pull request contains the following changes:
- parameterize average block mining time
- set average block mining time to 60 seconds
- parameterized freeze time range 
- set freeze time range to 10 seconds
- add / update the related tests

Now:
- If a new block is mined between 1 - 54 seconds -> then the difficulty \ target will be increased \ decreased
- If a new block is mined between 55 - 64 seconds -> then the difficulty and target will not be changed
- If a new block is mined 65+ seconds -> then the difficulty \ target will be decreased \ increased

Playground:
https://jsfiddle.net/2toqvnrd/3/

### DIFF
--- a/ironfish/src/consensus/consensus.ts
+++ b/ironfish/src/consensus/consensus.ts
@@ -52,4 +52,4 @@ export const TARGET_BLOCK_TIME_IN_SECONDS = 60
 /**
  * The time range when difficulty and target not change
  */
-export const FREEZE_TIME_IN_SECONDS = 10
+export const TARGET_BUCKET_TIME_IN_SECONDS = 10
--- a/ironfish/src/consensus/consensus.ts
+++ b/ironfish/src/consensus/consensus.ts
@@ -43,9 +43,11 @@ export const MAX_SYNCED_AGE_MS = 60 * 1000
 export const MAX_REQUESTED_BLOCKS = 50
 
 /**
- * The network has a target block time of 15 seconds
- *
- * NOTE: This is not used in target calculation, or IRON_FISH_YEAR_IN_BLOCKS.
- * Ideally, it should be in a future refactor.
+ * The network has a target block time of 60 seconds
  */
-export const TARGET_BLOCK_TIME_MS = 60 * 1000
+export const TARGET_BLOCK_TIME_IN_SECONDS = 60
+
+/**
+ * The time range when difficulty and target not change
+ */
+export const FREEZE_TIME_IN_SECONDS = 10
--- a/ironfish/src/consensus/consensus.ts
+++ b/ironfish/src/consensus/consensus.ts
@@ -43,7 +43,9 @@ export const MAX_SYNCED_AGE_MS = 60 * 1000
 export const MAX_REQUESTED_BLOCKS = 50
 
 /**
- * The network has a target block time of 60 seconds
+ * The average time that all blocks should be mined
+ *
+ * NOTE: This is not used in target calculation, or IRON_FISH_YEAR_IN_BLOCKS.
  */
 export const TARGET_BLOCK_TIME_IN_SECONDS = 60
 
--- a/ironfish/src/consensus/consensus.ts
+++ b/ironfish/src/consensus/consensus.ts
@@ -48,4 +48,4 @@ export const MAX_REQUESTED_BLOCKS = 50
  * NOTE: This is not used in target calculation, or IRON_FISH_YEAR_IN_BLOCKS.
  * Ideally, it should be in a future refactor.
  */
-export const TARGET_BLOCK_TIME_MS = 15 * 1000
+export const TARGET_BLOCK_TIME_MS = 60 * 1000
--- a/ironfish/src/primitives/__fixtures__/transaction.test.slow.ts.fixture
+++ b/ironfish/src/primitives/__fixtures__/transaction.test.slow.ts.fixture
@@ -2,69 +2,11 @@
   "Accounts produces unique transaction hashes": [
     {
       "name": "test",
-      "spendingKey": "4518015797a099049ac1b7320ada50680ac39d680ee7145177552fe20cb92896",
-      "incomingViewKey": "2bf2f75e34de3968588af71ff3451aac4c2bb4804d3fcb93bc5c95d1b5c85604",
-      "outgoingViewKey": "99366a8077f75387665c1cd16553d6496f32e99777ea809e9beb5f12f147f8a4",
-      "publicAddress": "955a20a24ee8bf115860e998f8a82ef24da9e66f04b244de53cdbe2bac386cab9a7cca38b70788364a200e",
+      "spendingKey": "b342b6f4a6e7e015cfebb0def4bf84170a137ab03a826a9a98029b971ea9dd77",
+      "incomingViewKey": "2862e0b4967ca0fbc374866f4c3770b434441a53a544392126cca56765bc7e06",
+      "outgoingViewKey": "0b0e4dabf8a443eee19e3723c347086c7a4cc742e8e56768385980d324929db3",
+      "publicAddress": "8020264e76d2d5bded8339fc5b0eb489d61549d394afe00744deab7e1aa84279198bef2880a8dfcde83e1e",
       "rescan": null
-    }
-  ],
-  "Accounts check if a transaction is a miners fee": [
-    {
-      "name": "test",
-      "spendingKey": "68e14e6a093c481d67f3621fccbf15765e37e030cfe52b339fd3d63a0150cd0f",
-      "incomingViewKey": "dc19cbf0313e1a7bd22451e121d18f7a4845bbfafae6f2e0bef32a7e3f5ba305",
-      "outgoingViewKey": "6c89892f8b76a334953fc4f4da3e071756a9c200aca8f3503087c75b14c01762",
-      "publicAddress": "4c62d2d37aa1458c81e5753366ec7bae6f1496a33938ba5f50a53760809397124545ba1a82898bbf6ef34a",
-      "rescan": null
-    }
-  ],
-  "Accounts check if a transaction is not a miners fee": [
-    {
-      "name": "testA",
-      "spendingKey": "bb70c25909c096be1e4ae37485e62233a4afa54ce15f2a611efe908e6b06d224",
-      "incomingViewKey": "4080a9c13d75893180032107ebf48e73527a50ad50206bb5f56431c40f59ec06",
-      "outgoingViewKey": "ac21daff4fcd5ead65cf92c6f201caec3889b01298803cc317be26235939bbc5",
-      "publicAddress": "bb1d044cb91e11a310342fa20f927ea998a618b348e191ef92462586537414e6fa4173fe316ec7b9ee3d97",
-      "rescan": null
-    },
-    {
-      "name": "testB",
-      "spendingKey": "9ed4828f428d800937fbcd8c19e0ebe6ba94611143d02b9ff8d069f40c8e59b7",
-      "incomingViewKey": "fb6219eced9fa68d47a590612fc4f6c4a97a1a805be3a5cfca564aeebea4c002",
-      "outgoingViewKey": "d1d7ef7cfd1ffc4140f023f8e6a5edaab20283213ea294b13e43fe83c74e7c2f",
-      "publicAddress": "769bf02082700c7e722d83af829ed44d852350220ddbdeb1d0c5361c5513c4f5d68140d1f39d84f74bdac1",
-      "rescan": null
-    },
-    {
-      "header": {
-        "sequence": 2,
-        "previousBlockHash": "A5656494D952D917A187A750B5C19C689C564E6F2CF0919FC5CD84B36FAF4D8E",
-        "noteCommitment": {
-          "commitment": {
-            "type": "Buffer",
-            "data": "base64://KQ5oEeJILYShvB4Z/FNuaEaKPEToqAF8UYL5VFJm0="
-          },
-          "size": 4
-        },
-        "nullifierCommitment": {
-          "commitment": "64DB0DD35E49EE735EECC79FDAE6764C42BFF27FEF35D73A6F404E666A1F13C6",
-          "size": 1
-        },
-        "target": "883423532389192164791648750371459257913741948437809479060803100646309888",
-        "randomness": 0,
-        "timestamp": 1635954493371,
-        "minersFee": "-500000000",
-        "work": "0",
-        "hash": "62609D481E6C926FA068065D4E2DABC5FDE58E6DAB29D757BC5C35AD6C3E2F18",
-        "graffiti": "0000000000000000000000000000000000000000000000000000000000000000"
-      },
-      "transactions": [
-        {
-          "type": "Buffer",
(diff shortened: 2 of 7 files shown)


NullSoldier:
This is a duplicate of this PR https://github.com/iron-fish/ironfish/pull/379 which also has existing feedback and context.

ironfish/src/primitives/target.test.ts
```diff
@@ -121,11 +121,29 @@ describe('Calculate target', () => {
...
+
+      expect(newDifficulty).toBeGreaterThan(difficulty)
+      expect(BigInt(difficulty) + diffInDifficulty).toEqual(newDifficulty)
+
+      expect(newTarget.targetValue).toBeLessThan(target.targetValue)
+    }
   })
 
-  it('keeps difficulty/target of parent block header if time differnece is between 10 and 20 seconds', () => {
+  it('keeps difficulty/target of parent block header if time differnece is between 55 and 64 seconds', () => {
```
leanthebean, line 144: you mean 65 seconds?

ironfish/src/primitives/target.test.ts
```diff
@@ -121,11 +121,29 @@ describe('Calculate target', () => {
 
       expect(newTarget.targetValue).toBeLessThan(target.targetValue)
     }
+
+    // for any time 35-45 seconds after the last block, difficulty should increase by previous block's difficulty / BigInt(2048) * 2
```
leanthebean, line 125: why is 35-45 seconds special?

ironfish/src/primitives/target.test.ts
yuryironfish: The seconds when the difficulty is not changing:
55 56 57 58 59 60 61 62 63 64 -> 10 numbers

ironfish/src/primitives/target.test.ts
yuryironfish: Previously when the target was 15 -> then the difficulty had increased from 1 to 10
but right now the difficulty will be increased by bucket length pattern
...
15 - 25 - * 2048 * 4
25 - 35 - * 2048 * 3
35 - 45 - * 2048 * 2
45 - 55 - * 2048
55 - 64 - nothing
...

ironfish/src/primitives/target.test.ts
leanthebean: sure but the for loop is exclusive. 
small nit, doesn't really matter, but the comment isn't really matching the code then

ironfish/src/primitives/target.test.ts
leanthebean: ok great, we should have tests to reflect that then.

ironfish/src/primitives/target.test.ts
```diff
@@ -64,81 +65,73 @@ describe('Target', () => {
 })
 
 describe('Calculate target', () => {
-  it('can increase target (which decreases difficulty) if its taking too long to mine a block (20+ seconds since last block)', () => {
+  it('increases difficulty if a new block is coming in before the target range time', () => {
     const now = new Date()
-    // for any time 20-29 seconds after the last block, difficulty should decrease by previous block's difficulty / BigInt(2048)
-    for (let i = 1; i < 10; i++) {
-      const time = new Date(now.getTime() + 20000 + i * 1000)
+    // for any time 0 - 55 seconds after the last block, difficulty should increase by previous block's difficulty / BigInt(2048) * m
```
leanthebean, line 70: what is `m` here?

ironfish/src/primitives/target.test.ts
```diff
@@ -64,81 +65,73 @@ describe('Target', () => {
...
-
-      const newDifficulty = Target.calculateDifficulty(time, now, difficulty)
-      const newTarget = Target.calculateTarget(time, now, target)
-
-      expect(newDifficulty).toBeLessThan(difficulty)
-      expect(BigInt(newDifficulty) + diffInDifficulty).toEqual(difficulty)
-
-      expect(newTarget.asBigInt()).toBeGreaterThan(target.asBigInt())
-    }
+      const bucketFromParent =
```
leanthebean, line 77: can we extract this to a method that all these tests can use? otherwise it's a lot of copy/paste code that's part of the test setup

ironfish/src/primitives/target.test.ts
```diff
@@ -64,81 +65,73 @@ describe('Target', () => {
...
       const newDifficulty = Target.calculateDifficulty(time, now, difficulty)
       const newTarget = Target.calculateTarget(time, now, target)
 
-      expect(newDifficulty).toBeGreaterThan(difficulty)
-      expect(BigInt(difficulty) + diffInDifficulty).toEqual(newDifficulty)
+      const diffInDifficulty = BigInt(newDifficulty) - difficulty
 
-      expect(newTarget.targetValue).toBeLessThan(target.targetValue)
+      expect(diffInDifficulty).toEqual(BigInt(0))
+      expect(newDifficulty).toEqual(difficulty)
```
leanthebean, line 107: `expect(diffInDifficulty).toEqual(BigInt(0))`
`expect(newDifficulty).toEqual(difficulty)`
are testing the same thing

ironfish/src/primitives/target.test.ts
```diff
@@ -64,81 +65,73 @@ describe('Target', () => {
...
+      expect(newTarget.targetValue).toEqual(target.targetValue)
     }
   })
 
-  it('keeps difficulty/target of parent block header if time differnece is between 10 and 20 seconds', () => {
+  it('dencreases difficulty if a new block is coming in after the target range time', () => {
```
leanthebean, line 112: typo: `dencreases` should be `decreases`

ironfish/src/primitives/target.test.ts
```diff
@@ -64,81 +65,73 @@ describe('Target', () => {
...
     const now = new Date()
-    for (let i = 10; i < 20; i++) {
+
+    // for any time more than 65 seconds after the last block, difficulty should decrease by previous block's difficulty / BigInt(2048) / n
```
leanthebean, line 115: this comment is wrong
for 65 - 74 seconds difficulty = parentDifficulty - parentDifficulty / 2048
for 75 - 85 seconds difficulty =  parentDifficulty - (parentDifficulty / 2048 * 2)
and so on 

could you make the comment correct and more clear?

ironfish/src/primitives/target.ts
```diff
@@ -140,14 +141,31 @@ export class Target {
...
-    )
+
+    const threshold = 1
+    const halfBucket = Math.floor(FREEZE_TIME_IN_SECONDS / 2)
+    const bucket =
+      Math.floor(
+        (diffInSeconds - TARGET_BLOCK_TIME_IN_SECONDS + halfBucket) / FREEZE_TIME_IN_SECONDS,
+      ) + 1
+
+    const max_change =
```
leanthebean, line 154: don't use snake_case here, use camelCase

ironfish/src/primitives/target.ts
```diff
@@ -140,14 +141,31 @@ export class Target {
...
 
     // diff = parent_diff + parent_diff / 2048 * max(1 - (current_block_timestamp - parent_timestamp) / 10, -99)
-    const difficulty = BigIntUtils.max(
-      BigInt(previousBlockDifficulty) + offset * sign,
-      Target.minDifficulty(),
-    )
+
+    const threshold = 1
+    const halfBucket = Math.floor(FREEZE_TIME_IN_SECONDS / 2)
+    const bucket =
```
leanthebean, line 149: can we extract this to a method and test that? it's really weird to have things like `halfBucket` here.

ironfish/src/primitives/target.ts
```diff
@@ -140,14 +141,31 @@ export class Target {
...
+    let targetDifficulty = null
+    if (bucket <= threshold) {
+      // Scale up
+      const multiplier = threshold - bucket
+      targetDifficulty = previousBlockDifficulty + BigInt(max_change) * BigInt(multiplier)
+    } else {
+      // Scale down
+      const multiplier = Math.min(bucket - threshold, 99)
+      targetDifficulty = previousBlockDifficulty - BigInt(max_change) * BigInt(multiplier)
+    }
```
leanthebean, line 166: what is `threshold` here? why is it 1? 

the basic idea here is if our target is 60s and our slippage is +/- 5 seconds, then first we try to figure out our bucket from target and adjust previous block's difficulty accordingly 

i think this code can be simplified quite a bit by something like: 

```
      const bucketFromTarget = findBucket(targetTime, slippage, blockTime)

      const diffInDifficulty = (parentDifficulty / BigInt(2048)) * BigInt(bucketFromTarget)

      const difficulty =
        blockTime < targetTime
          ? parentDifficulty + diffInDifficulty
          : parentDifficulty - diffInDifficulty
```

ironfish/src/primitives/target.ts
```diff
@@ -140,14 +141,31 @@ export class Target {
...
+
+    const difficulty = BigIntUtils.max(targetDifficulty, Target.minDifficulty())
```
leanthebean, line 168: it feels like we're clamping the adjustment in two places here: on line 164 `Math.min(bucket - threshold, 99)` and here as well which is confusing

ironfish/src/primitives/target.ts
```diff
@@ -140,14 +141,31 @@ export class Target {
...
+      BigInt(previousBlockDifficulty) / BigInt(DIFFICULTY_ADJUSTMENT_DENOMINATOR)
+
+    let targetDifficulty = null
+    if (bucket <= threshold) {
+      // Scale up
+      const multiplier = threshold - bucket
+      targetDifficulty = previousBlockDifficulty + BigInt(max_change) * BigInt(multiplier)
+    } else {
+      // Scale down
+      const multiplier = Math.min(bucket - threshold, 99)
```
leanthebean, line 164: When would this condition ever hit? it's basically saying choose the min of bucket from target, or 99. When would `bucket - threshold` ever be more than 99 for this condition to hit?  That would mean the block is trying to come in 1055 seconds from previous block, and we should set our difficulty to be the lowest difficulty possible which is basically checked on like 168 with `BigIntUtils.max(targetDifficulty, Target.minDifficulty())` ? 

I might be wrong, but I also don't see a test for this

leanthebean (changes requested):

ironfish/src/primitives/__fixtures__/transaction.test.slow.ts.fixture
```diff
@@ -2,38 +2,38 @@
   "Accounts produces unique transaction hashes": [
```
leanthebean, line 2: are you sure fixtures need to be updated for this change? we're just changing a static method on the Target class.

ironfish/src/primitives/target.ts
```diff
@@ -140,14 +141,31 @@ export class Target {
...
+    const halfBucket = Math.floor(FREEZE_TIME_IN_SECONDS / 2)
+    const bucket =
+      Math.floor(
+        (diffInSeconds - TARGET_BLOCK_TIME_IN_SECONDS + halfBucket) / FREEZE_TIME_IN_SECONDS,
+      ) + 1
+
+    const max_change =
+      BigInt(previousBlockDifficulty) / BigInt(DIFFICULTY_ADJUSTMENT_DENOMINATOR)
+
+    let targetDifficulty = null
```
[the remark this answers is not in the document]
leanthebean: or parentDifficulty or previousBlockDifficulty

ironfish/src/primitives/target.ts
```diff
@@ -140,16 +141,39 @@ export class Target {
...
     // diff = parent_diff + parent_diff / 2048 * max(1 - (current_block_timestamp - parent_timestamp) / 10, -99)
-    const difficulty = BigIntUtils.max(
-      BigInt(previousBlockDifficulty) + offset * sign,
-      Target.minDifficulty(),
-    )
 
-    return difficulty
+    const threshold = 1
+    const bucket = this.findBucket(diffInSeconds)
+    const maxChange = previousBlockDifficulty / BigInt(DIFFICULTY_ADJUSTMENT_DENOMINATOR)
```
leanthebean, line 149: why can't we just put 2048 here? is DIFFICULTY_ADJUSTMENT_DENOMINATOR ever used anywhere else?

leanthebean: also again, `maxChange` implies that we'll be changing the difficulty _maximum_ that number, which isn't true

leanthebean (changes requested):

ironfish/src/primitives/target.test.ts
```diff
@@ -112,7 +112,7 @@ describe('Calculate target', () => {
   it('dencreases difficulty if a new block is coming in after the target range time', () => {
     const now = new Date()
 
-    // for any time more than 65 seconds after the last block, difficulty should decrease by previous block's difficulty / BigInt(2048) / n
+    // for any time more than 65 seconds after the last block, difficulty should decrease by previous block's difficulty / BigInt(2048) / bucket
```
leanthebean, line 115: `previous block's difficulty / BigInt(2048) / bucket`

that's not the algorithm though.

ironfish/src/consensus/consensus.ts
```diff
@@ -43,9 +43,11 @@ export const MAX_SYNCED_AGE_MS = 60 * 1000
 export const MAX_REQUESTED_BLOCKS = 50
 
 /**
- * The network has a target block time of 15 seconds
- *
- * NOTE: This is not used in target calculation, or IRON_FISH_YEAR_IN_BLOCKS.
- * Ideally, it should be in a future refactor.
+ * The network has a target block time of 60 seconds
```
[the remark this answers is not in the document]
yuryironfish: It showed that IRON_FISH_YEAR_IN_BLOCKS was not used anywhere

ironfish/src/primitives/target.ts
```diff
@@ -165,9 +160,21 @@ export class Target {
...
+  /**
+   * Returns the bucket values based difference time between the current and previous block, target block mining time and free time range
+   * @param diffInSeconds different in seconds between blocks
+   * @returns the bucket number
+   */
+  static findBucket(diffInSeconds: number): number {
+    const halfBucket = Math.floor(FREEZE_TIME_IN_SECONDS / 2)
+    return (
+      Math.floor(
+        (diffInSeconds - TARGET_BLOCK_TIME_IN_SECONDS + halfBucket) / FREEZE_TIME_IN_SECONDS,
```
leanthebean, line 174: `FREEZE_TIME_IN_SECONDS` and `free time range` are confusing. we're not really freezing anything. maybe something like `ALLOWED_SLIPPAGE_FROM_TARGET_TIME` ?

ironfish/src/consensus/consensus.ts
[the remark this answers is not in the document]
leanthebean: oh. that was a good note actually. 

because we're changing block time, IRON_FISH_YEAR_IN_BLOCKS also has to change. 
if you look in consensus.ts: 

```
/*
 * A ratio of blocks per year that represents an approximation of how many blocks are considered a "year".
 * It's generally an approximation based on 15 second block times. It's used in calculating how much a miner
 * should get in rewards. Approximately (365 * 24  * 60 * 60) / TARGET_BLOCK_TIME_MS
 */
export const IRON_FISH_YEAR_IN_BLOCKS = 2100000
```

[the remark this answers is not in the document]
leanthebean: IRON_FISH_YEAR_IN_BLOCKS changes the miner's reward, so very important that this change also updates that. 

that was an excellent note that you deleted fwiw — i would have totally forgotten about that.

ironfish/src/primitives/target.test.ts
```diff
@@ -64,81 +65,86 @@ describe('Target', () => {
...
     }
   })
 
-  it('keeps difficulty/target of parent block header if time differnece is between 10 and 20 seconds', () => {
+  it('decreases difficulty if a new block is coming in after the target range time', () => {
     const now = new Date()
-    for (let i = 10; i < 20; i++) {
+
+    /**
+     * if new block comes after target block mining time + half bucket time, then difficulty is adjust as:
```
leanthebean, line 123: `half bucket time` ? 

I know what you mean, but our code is using static variables that don't mention bucket. we have something like `FREEZE_TIME_IN_SECONDS` which is very confusing. maybe we can rename `FREEZE_TIME_IN_SECONDS` to something less confusing and reference it here? or define bucket somewhere?

ironfish/src/consensus/consensus.ts
```diff
@@ -43,9 +43,13 @@ export const MAX_SYNCED_AGE_MS = 60 * 1000
...
  * NOTE: This is not used in target calculation, or IRON_FISH_YEAR_IN_BLOCKS.
- * Ideally, it should be in a future refactor.
  */
-export const TARGET_BLOCK_TIME_MS = 15 * 1000
+export const TARGET_BLOCK_TIME_IN_SECONDS = 60
+
+/**
+ * The time range when difficulty and target not change
+ */
+export const FREEZE_TIME_IN_SECONDS = 10
```
leanthebean, line 55: since we're using this to determine our bucket size, maybe something like BUCKET_THRESHOLD_IN_SECONDS? idk just trying to figure out how to make this clearer

ironfish/src/primitives/target.test.ts
leanthebean: we're never dividing by bucket

leanthebean (approved):

leanthebean (changes requested):
we have to understand why we need to update so many fixtures in order to get the tests to pass

leanthebean (approved):

NullSoldier (approved):

PR #426, opened 2021-11-05
feat(ironfish): Set expiration sequence for Transaction on creation

rohanjadvani:


ironfish/src/account/accounts.test.slow.ts
```diff
@@ -231,6 +233,11 @@ describe('Accounts', () => {
       generateKey().public_address,
     )
 
+    // Check the default expiration is DEFAULT_TRANSACTION_EXPIRATION_SEQUENCE_DELTA from head
```
NullSoldier, line 236: this comment is just a textual description of the code below

ironfish/src/account/accounts.test.slow.ts
```diff
@@ -249,6 +256,55 @@ describe('Accounts', () => {
     })
   }, 600000)
 
+  it('throws a ValidationError with an invalid expiration sequence', async () => {
+    // Initialize the database and chain
```
NullSoldier, line 260: I don't think this is initializing the database and chain.

NullSoldier, line 260: nit: Lets delete this comment.. I see the old code is wrong too I'll put up a PR to clean those up as well.

ironfish/src/consensus/consensus.ts
```diff
@@ -49,3 +49,9 @@ export const MAX_REQUESTED_BLOCKS = 50
  * Ideally, it should be in a future refactor.
  */
 export const TARGET_BLOCK_TIME_MS = 15 * 1000
+
+/**
+ * The default delta of block sequence for which to expire transactions from the
+ * mempool.
+ */
+export const DEFAULT_TRANSACTION_EXPIRATION_SEQUENCE_DELTA = 450
```
NullSoldier, line 57: I think this should go in the config, rather than the consensus layer.

ironfish/src/account/accounts.ts
```diff
@@ -588,18 +591,27 @@ export class Accounts {
...
   ): Promise<Transaction> {
     const heaviestHead = this.chain.head
     if (heaviestHead === null) {
       throw new ValidationError('You must have a genesis block to create a transaction')
     }
 
+    expirationSequence =
+      expirationSequence ??
+      heaviestHead.sequence + DEFAULT_TRANSACTION_EXPIRATION_SEQUENCE_DELTA
+    if (0 < expirationSequence && expirationSequence <= heaviestHead.sequence) {
```
NullSoldier, line 604: You can pass in a negative number for expiration here and it will continue. How about....?

```suggestion
    if (expirationSequence !== 0 && expirationSequence <= heaviestHead.sequence) {
```

NullSoldier:
Ok looks pretty good. Let's add it to the CLI in another PR too?

NullSoldier:
Looks like we also need to add this to the consensus layer too.

ironfish/src/account/accounts.test.slow.ts
```diff
@@ -249,6 +257,55 @@ describe('Accounts', () => {
...
+    const newBlock = await chain.newBlock([], minersfee)
+    const addResult = await chain.addBlock(newBlock)
+    expect(addResult.isAdded).toBeTruthy()
+
+    // Account should now have a balance of 500000000 after adding the miner's fee
+    await node.accounts.updateHead()
+    expect(node.accounts.getBalance(account)).toEqual({
+      confirmedBalance: BigInt(500000000),
+      unconfirmedBalance: BigInt(500000000),
+    })
```
NullSoldier, line 292: Do we need all this to test the input validation?

ironfish/src/rpc/routes/transactions/sendTransaction.ts
```diff
@@ -91,6 +93,8 @@ router.register<typeof SendTransactionRequestSchema, SendTransactionResponse>(
       BigInt(transaction.transactionFee),
       transaction.memo,
       transaction.toPublicKey,
+      node.config.get('defaultTransactionExpirationSequenceDelta'),
```
NullSoldier, line 96: This is the only oddity, but I'm going to approve it. I think taking both may make this API a little bit confusing to outsiders.

NullSoldier (approved):

PR #428, opened 2021-11-09
Build ironfish-rust as native NodeJS addon

dguenther:
Adds an ironfish-rust-nodejs package to the repo that uses Neon to build ironfish-rust as a native NodeJS addon instead of a WASM bundle. This should significantly improve performance on all machines.

Also renames ironfish-wasm to ironfish-rust-wasm, and cleans up a lot of code that was referring to wasm specifically in the ironfish package.

Fixes IRO-1199


dguenther:
This should be ready for testing. Ideally you should be able to run `yarn` in the root directory to rebuild, and everything should Just Work after that.

leanthebean:
I'm getting `sh: cargo-cp-artifact: command not found` when I try to do `yarn build` from the root directory,

ironfish/jest.setup.env.js
```diff
@@ -26,8 +26,8 @@ jest.mock('node-datachannel', () => {
 })
 
 beforeAll(() => {
-  // This causes the WASM to be initialized, which is 1 time 2 second cost for each test suite
-  if (process.env.TEST_INIT_WASM) {
+  // This causes Sapling to be initialized, which is 1 time 2 second cost for each test suite
```
deekerno, line 29: Nit: Is this supposed to be "**a** 1 time 2 second cost..." (alternatively, "a one-time two-second cost...")?

dguenther:
> I'm getting `sh: cargo-cp-artifact: command not found` when I try to do `yarn build` from the root directory,

If anyone else runs into this, you need to do `yarn` from the root directory to update your packages.

deekerno (approved):
Everything looks good to me, great job!

leanthebean:
Been running this for a few hours and the node fully synced with avg add block time of 15-25ms (down from 55-80ms previously on the same machine). 

Looks good to me! 🔥 🔥 🔥 🔥 🔥 

🚢

leanthebean (approved):

NullSoldier (approved):

PR #429, opened 2021-11-09
Send peer list to connected peers instead of identified peers

deekerno:
We were mistakenly sending out peer lists to disconnected and connecting peers as well as connected peers.


NullSoldier (approved):

NullSoldier:
Tested manually, this works.

PR #431, opened 2021-11-10
Updating IRON_FISH_YEAR_IN_BLOCKS due to block time change

leanthebean:
Please see this doc: 
https://docs.google.com/spreadsheets/d/1b2ZKY3vpEyk5GPB3TCjWFK4v_dfaokVsxydc8-NigAM/edit#gid=0

TL;DR 
We used to have 15 second block times and with our emissions curve (and a tiny bit of tweaking) the first year past launch had 5 coins per block as the block reward. 
We then changed our block time to 60 seconds and using the same equation our block rewards would be 19.9771 coins per block. 
Since we're using a decay function to lower inflation and decrease block reward per year, our equation makes it hard to have round numbers per block, especially if we'll be changing block time. 
In this PR the proposal is to **round** the block reward to the nearest 0.125 of a coin to make the block rewards easier to read and not be prone to rounding errors. 
Due to the decision to round the block reward to the nearest 0.125 of a coin our emissions curve is going to change. 
Before this PR, our emissions curve had an asymptotic cap (e.g. unlike Bitcoin we didn't have a hard cap, but block rewards gradually approached 0 after 80 or so years). 
In this PR, due to the rounding decision, our emissions would indeed reach 0 after a certain point, making Iron Fish be deflationary/have a fixed supply, which it arguably did before anyway. 

To see graphs and such of old emissions curve and the new one that resulted due to this rounding decision, please see: https://docs.google.com/spreadsheets/d/1b2ZKY3vpEyk5GPB3TCjWFK4v_dfaokVsxydc8-NigAM/edit#gid=0

### DIFF
--- a/ironfish/src/consensus/consensus.ts
+++ b/ironfish/src/consensus/consensus.ts
@@ -51,6 +51,5 @@ export const TARGET_BUCKET_TIME_IN_SECONDS = 10
  * A ratio of blocks per year that represents an approximation of how many blocks are considered a "year".
  * It's generally an approximation based on TARGET_BLOCK_TIME_IN_SECONDS second block times.
  * It's used in calculating how much a miner should get in rewards.
- * Approximately (365 * 24 * 60 * 60) / TARGET_BLOCK_TIME_IN_SECONDS
  */
 export const IRON_FISH_YEAR_IN_BLOCKS = (365 * 24 * 60 * 60) / TARGET_BLOCK_TIME_IN_SECONDS
--- a/ironfish/src/consensus/consensus.ts
+++ b/ironfish/src/consensus/consensus.ts
@@ -25,13 +25,6 @@ export const ALLOWED_BLOCK_FUTURE_SECONDS = 15
  */
 export const GENESIS_SUPPLY_IN_IRON = 42000000
 
-/*
- * A ratio of blocks per year that represents an approximation of how many blocks are considered a "year".
- * It's generally an approximation based on 15 second block times. It's used in calculating how much a miner
- * should get in rewards. Approximately (365 * 24  * 60 * 60) / TARGET_BLOCK_TIME_MS
- */
-export const IRON_FISH_YEAR_IN_BLOCKS = 2100000
-
 /**
  * The oldest the tip should be before we consider the chain synced
  */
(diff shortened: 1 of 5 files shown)


ironfish/src/consensus/consensus.ts
```diff
@@ -48,4 +41,17 @@ export const MAX_REQUESTED_BLOCKS = 50
...
+/**
+ * The time range when difficulty and target not change
+ */
+export const TARGET_BUCKET_TIME_IN_SECONDS = 10
+
+/*
+ * A ratio of blocks per year that represents an approximation of how many blocks are considered a "year".
+ * It's generally an approximation based on TARGET_BLOCK_TIME_IN_SECONDS second block times.
+ * It's used in calculating how much a miner should get in rewards.
+ * Approximately (365 * 24 * 60 * 60) / TARGET_BLOCK_TIME_IN_SECONDS
```
NullSoldier, line 55: This is exactly the code below so I don't think we need this written out twice.

```suggestion
```

ironfish/src/strategy.ts
```diff
@@ -76,12 +76,16 @@ export class Strategy {
...
     const annualReward = (GENESIS_SUPPLY_IN_IRON / 4) * Math.E ** (-0.05 * yearsAfterLaunch)
 
-    reward = this.convertIronToOre(annualReward / IRON_FISH_YEAR_IN_BLOCKS)
+    reward = this.convertIronToOre(this.mRound(annualReward / IRON_FISH_YEAR_IN_BLOCKS, 0.125))
     this.miningRewardCachedByYear.set(yearsAfterLaunch, reward)
 
     return reward
   }
 
+  mRound(num: number, threshold: number): number {
```
NullSoldier, line 85: What do you think about moving this here next to this, and giving it a better name? It seems useful. https://github.com/iron-fish/ironfish/blob/master/ironfish/src/utils/math.ts#L29-L32

NullSoldier: Fixed

ironfish/src/consensus/consensus.ts
NullSoldier: Fixed

ironfish/src/consensus/consensus.ts
```diff
@@ -48,4 +41,17 @@ export const MAX_REQUESTED_BLOCKS = 50
...
+ */
+export const IRON_FISH_YEAR_IN_BLOCKS = (365 * 24 * 60 * 60) / TARGET_BLOCK_TIME_IN_SECONDS
```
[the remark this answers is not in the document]
NullSoldier: I think the answer is that this curve should be frozen to 0 block activation height, even if we abstract out the consensus parameters later.

NullSoldier (approved):

yuryironfish (approved):

PR #432, opened 2021-11-10
feat(cli): Add expiration sequence flag to pay command

rohanjadvani:
This PR will eventually go into the [`mempool`](https://github.com/iron-fish/ironfish/tree/mempool) branch. We should merge #426 first.


ironfish-cli/src/commands/accounts/pay.ts
```diff
@@ -59,6 +64,7 @@ export class Pay extends IronfishCommand {
     let fee = flags.fee as unknown as number
     let to = flags.to
     let from = flags.account
+    const expirationSequence = flags.expirationSequence
```
NullSoldier, line 67: this is great, we just need to pass this into the API

NullSoldier (approved):

PR #436, opened 2021-11-16
Add BigInt helper to divide to a number

NullSoldier:
This is useful for getting ratios between two BigInt types


yuryironfish (approved):
APPROVED

PR #437, opened 2021-11-18
Refactor miner

NullSoldier:
Changed the miner so that it can show status of what it's doing. The
main changes are below.

 - Cleaned up Miner Code and tried to make it simpler
 - Miner now has a meter to measure hash rate
 - MiningDirector now sends the sequence through the MineRequest
 - Fixed newBlockStream not unsubscribing when request closes
 - Fixed bug with race condition when multiple block results come in at
   the same time
 - Only submit first result for each block instead of adding all forks.


ironfish/src/utils/async.ts
```diff
@@ -11,6 +11,15 @@ export class AsyncUtils {
     return results
   }
 
+  static async *transform<TIn, TOut>(
+    iter: AsyncGenerator<TIn, unknown, unknown>,
```
dguenther, line 15: Can this take `AsyncIterable` instead? I think it's more general

NullSoldier: Changed

dguenther (approved):

PR #452, opened 2021-11-23
Pin cargo-cp-artifact at 0.1.5

NullSoldier:
Don't like self upgrading libraries. You can get exploited that way.


yuryironfish (approved):
APPROVED

PR #470, opened 2021-11-24
WIP: IRO-1283: Buffer serialization of block

yuryironfish:


ironfish/src/primitives/blockheader.ts
```diff
@@ -237,62 +239,48 @@ export class BlockHeaderSerde implements Serde<BlockHeader, SerializedBlockHeade
...
-      Number(data.sequence),
-      Buffer.from(BlockHashSerdeInstance.deserialize(data.previousBlockHash)),
+      Number(br.readU64()),
+      br.readHash(),
       {
-        commitment: this.strategy.noteHasher
-          .hashSerde()
-          .deserialize(data.noteCommitment.commitment),
-        size: data.noteCommitment.size,
+        commitment: br.readHash(),
```
NullSoldier: Reading in an object literal like this can be dangerous because the keys could get read in the incorrect order. I think it's better to read all the data out then construct this object.

dguenther:
I'm going to close this PR since it's fairly old and we're not currently working on it. We can pick it back up when we decide to start working on it again.

PR #473, opened 2021-11-30
Remove TODO items from chainProcessor

NullSoldier:
After looking at the code, I think the proposed changes are the real fix
and not a temporary fix. We can't really lock the entire chain to update
because it would freeze syncing. Also accounts doesn't need to lock the
chain either. All that's required is that we freeze the head at the
beginning and always reference that as the target, similar to the
accounts.


leanthebean (approved):

yuryironfish (approved):
APPROVED

PR #474, opened 2021-11-30
Remove transactions from accounts when they expire

dguenther:
Once transactions have expired, we want to remove them from accounts so that the notes that were spent in the expired transaction become spendable again.


NullSoldier (approved):

NullSoldier:
Can you just regenerate the fixtures and then this is good to go?

PR #476, opened 2021-11-30
Fix broken test

NullSoldier:
This test needed one more block on the A chain because chain B actually
has 4 blocks because of useBlockWithTx actually generates 2 blocks not
1.


ironfish/src/blockchain/blockchain.test.ts
```diff
@@ -541,33 +541,38 @@ describe('Blockchain', () => {
...
     // If we add A1 to nodeB it should not re-org yet
     await expect(nodeB.chain).toAddBlock(blockA1)
-    expect(nodeB.chain.head.hash.equals(blockB3.header.hash)).toBe(true)
+    expect(nodeB.chain.head.hash.equals(blockB4.header.hash)).toBe(true)
 
     // This should succeed but before the fix it would fail
     const result = await nodeB.chain.addBlock(blockA2)
     expect(result).toMatchObject({ isAdded: true, reason: null })
 
     // Head should still be blockB3
```
dguenther, line 568:
```suggestion
    // Head should still be blockB4
```

ironfish/src/blockchain/blockchain.test.ts
```diff
@@ -541,33 +541,38 @@ describe('Blockchain', () => {
...
+
     const blockB1 = await useMinerBlockFixture(nodeB.chain)
     await expect(nodeB.chain).toAddBlock(blockB1)
 
     const blockB2 = await useMinerBlockFixture(nodeB.chain)
     await expect(nodeB.chain).toAddBlock(blockB2)
 
-    const { block: blockB3 } = await useBlockWithTx(nodeB)
-    await expect(nodeB.chain).toAddBlock(blockB3)
+    // blockB3 is created and added by the fixture automatically
```
dguenther, line 553:
```suggestion
    // blockB4 is created and added by the fixture automatically
```

NullSoldier: This suggestion makes the comment incorrect.

dguenther (approved):

PR #480, opened 2021-11-30
Only load data in open when upgrading the DB

NullSoldier:
When we are not upgrading the data version, it means we are doing a
migration which means data should not be loaded.


brekk (approved):
:shipit:

PR #486, opened 2021-11-30
Fix miner adding double spends

NullSoldier:
We don't filter out conflicting transactions when we get transactions from the mem pool. That needs to happen in the director. This adds a test + the fix.


ironfish/src/memPool/memPool.ts
```diff
@@ -56,8 +56,10 @@ export class MemPool {
   }
 
   *get(): Generator<Transaction, void, unknown> {
-    while (!this.queue.isEmpty()) {
-      const feeAndHash = this.queue.poll()
+    const clone = this.queue.clone()
```
dguenther, line 59: There might be potential for issues if the cloned queue desyncs from the main queue, e.g. if a block is connected and transactions get removed from the queue while iterating over the cloned one. Not familiar enough with how the constructed block gets verified to say for sure

ironfish/src/mining/director.ts
```diff
@@ -497,6 +498,34 @@ export class MiningDirector {
...
+          conflicted = true
+          break
+        }
+
+        nullifiers.add(spend.nullifier)
+
+        if (await this.chain.nullifiers.contains(spend.nullifier)) {
+          conflicted = true
+          break
+        }
```
dguenther, line 519: There may be an edge case here, but the behavior might be intentional, I think we discussed this a bit yesterday.

Say you have two transactions in the queue, each with two nullifiers:
```
T1: N1, N2
T2: N1, N3
```
Assume N2 is in the chain already, but N1 and N3 are not. When the loop processes T1, N1 will get added to `nullifiers`, N2 will cause `conflicted` to be true and break out of the loop, then when the loop processes T2, N1 will be found to already be in `nullifiers` and break out of the loop, though it could potentially be added to the block.

dguenther (approved):
I think having this is better than not. I left a couple comments on some potential edge cases in case you hadn't thought of them

ironfish/src/memPool/memPool.ts
NullSoldier: Yeah, what will happen is that the final mined block will be invalid. The only solution is that we would have to lock the blockchain database in the director when creating a block and getting transactions.

ironfish/src/mining/director.ts
NullSoldier: ah that's worth fixing, good find on my sloppy logic.

NullSoldier: Fixed

PR #489, opened 2021-12-01
Put miner's fee transaction first on blocks

dguenther:
We added some code a while ago to always put the miner's fee transaction first on a block, but didn't test it sufficiently, so actually it was ending up at the end (but block.minersFee was still returning the first transaction). This branch fixes the order so that the miner's fee transaction always comes first, and is checked by the verifier

This will require a network reset.


ironfish/src/mining/director.test.slow.ts
```diff
@@ -36,12 +36,13 @@ describe('Mining director', () => {
       const generateBlockSpy = jest
         .spyOn(miningDirector, 'generateBlockToMine')
         .mockReturnValue(Promise.resolve())
+      // onSynced can trigger while generating fixtures or if this test is run in isolation,
```
NullSoldier, line 39:
```suggestion

      // onSynced can trigger while generating fixtures or if this test is run in isolation,
```

ironfish/src/genesis/makeGenesisBlock.ts
```diff
@@ -76,19 +91,18 @@ export async function makeGenesisBlock(
   )
   transactionList.push(postedInitialTransaction)
 
-  // Temporarily add the note from the transaction to our merkle tree so we can construct
-  // a witness from it. It will be re-added later when the block is constructed.
+  // Temporarily add the miner's fee note and the from the transaction to our merkle tree
```
NullSoldier, line 94: `and the from the`?

NullSoldier (approved):

ironfish/src/genesis/makeGenesisBlock.ts
dguenther: Yep, fixed the typo

PR #494, opened 2021-12-01
Update docs with cmake information

k-xo:


dguenther (approved):
I'm going to merge this into master since it's a docs-only change.

PR #495, opened 2021-12-01
Fix accountsDB not closing properly

NullSoldier:
When accounts opens it does not list the DB as having been opened. So
when you close it, it doesn't close the DB connection.


yuryironfish (approved):
APPROVED

dguenther (approved):

PR #525, opened 2021-12-02
Various small improvements and code optimizations

bogdanteodoru:
- removed some methods that didn't appear to be used anywhere
- optimized the imports a bit
- minor typos fixed
- removed redundant code


ironfish/src/mining/director.ts
```diff
@@ -410,8 +410,7 @@ export class MiningDirector {
 
     this.recentBlocks.set(this.miningRequestId, newBlock)
 
-    const canRetry = target.asBigInt() < Target.maxTarget().asBigInt()
-    return canRetry
+    return target.asBigInt() < Target.maxTarget().asBigInt()
```
NullSoldier, line 413: This code is less clear now. It's unclear what is being returned from this method.

bogdanteodoru: Hey @NullSoldier,

From my knowledge doing that is not a best practice. I could on the other hand add some comments to explain the what the return is.

bogdanteodoru: Added comments for better readability

ironfish/src/assert.ts
```diff
@@ -42,13 +42,13 @@ export class Assert {
   }
 
   static isTrue(x: boolean, message?: string): asserts x is true {
-    if (x === false) {
+    if (!x) {
```
CumpsD, line 45: carefull with replacing `===`

```
let x = null

(!x)
true

(x === false)
false
```

bogdanteodoru: Yeah but based on the parameters which seem to be "boolean" that shoudn't happen. I see a strict "boolean" there, not a "boolean | null"

CumpsD: right, that's true, TS type checking can protect us, nvm :)

dguenther:
I'm closing this PR as we're not interested in reviewing code style changes at this time. If you'd like to open a new PR that includes just the typo fixes in comments, I'd be happy to take a look at that.

PR #543, opened 2021-12-03
Updated README to include cmake error.

phdjsep:
On MBP, encountered a `cmake` error on `yarn install` as well as the OpenSSL error, so including that note as well.


dguenther:
This was fixed in #494 which was submitted earlier, thanks for the PR though!

PR #575, opened 2021-12-05
Bugfix: ironfish-cli, unable to run "yarn clean" and "yarn prepack" under Windows

spale777:
Added OS agnostic way of removing build directory


dguenther:
Could you use the `rimraf` package to do this instead, so we can make the `postpack` script OS-agnostic as well? It should take adding `rimraf` as a devDependency here, then re-running `yarn install` in the root directory to update `yarn.lock`

spale777:
Sure thing, wanted to avoid adding any unnecessary deps but yeah no problem

spale777:
There you go all fixed up

ironfish-cli/package.json
```diff
@@ -19,7 +19,8 @@
     "oclif": "1.16.1",
     "tsc-watch": "4.2.9",
     "typescript": "4.3.4",
-    "yarn": "^1.22.10"
+    "yarn": "^1.22.10",
+    "rimraf": "^3.0.2"
```
dguenther, line 23: Last change, could you alphabetize this import?

dguenther (changes requested):
Nice, thanks for making that change! Just one more nitpick and then this'll be good to go.

ironfish-cli/package.json
spale777: Yup, done

dguenther (approved):
Thanks, looks good!

spale777:
@dguenther Why are tests failing for stuff that hasn't been touched ?

dguenther:
The tests pass, it just fails to upload code coverage due to a permissions issue.

dguenther (approved):

dguenther:
Attempting to fix the permissions issue in #736

PR #601, opened 2021-12-06
stop setting block hash to upper case in link

jakegrella:
hyperlink does not work when the hash is capitalized


NullSoldier:
Thanks, did you apply for test net credit?

jakegrella:
just did, yes

dguenther:
We fixed this on the API side to allow both uppercase and lowercase in the URLs, but I'm going to merge this since there's now no need for the node to change the case of the hash.

PR #609, opened 2021-12-06
change error message

mrrusik:
This is enhancement. 
Small modification in the error message.
Better delete `~/.ironfish/database` while you receive this error: `You are running a newer version of ironfish on an older database.`


ironfish/src/storage/levelup/database.ts
```diff
@@ -99,7 +99,7 @@ export class LevelupDatabase extends Database {
     if (current !== version) {
       throw new Error(
         `You are running a newer version of ironfish on an older database.\n` +
-          `Wipe your database using "ironfish reset" or delete your data directory at ~/.ironfish\n`,
+          `Wipe your database using "ironfish reset" or delete your data directory at ~/.ironfish/database\n`,
```
NullSoldier, line 102: This will not resolve the error, because you still need to delete ~/.ironfish/accounts too. So the previous message is better. You can test your change by temporarily bumping DATABASE_VERSION

mrrusik: So maybe better show error: `delete your data directory at ~/.ironfish/database and ~/.ironfish/accounts`
With that the file `config.json` remains and you do not need to do the configuration again

NullSoldier: This error is also pretty far down, so every time you add a database you need to update this spot, and people won't know to do that. Can you propose the change that's been spoken about here?

mrrusik: I change error message to _delete your data directory at ~/.ironfish/database and ~/.ironfish/accounts_
As I understand this command in reset.ts file 
 ```
   await Promise.all([
      fsAsync.rm(node.config.accountDatabasePath, { recursive: true }),
      fsAsync.rm(node.config.chainDatabasePath, { recursive: true }),
    ])
```
will delete ~/.ironfish/database and ~/.ironfish/accounts, when user exec ironfish reset

dguenther (approved):

PR #636, opened 2021-12-07
Grammar fixes

git-sgmoore:
Multiple minor grammar fixes

### DIFF
--- a/ironfish/README.md
+++ b/ironfish/README.md
@@ -7,13 +7,13 @@ Ironfish SDK wraps all of the generic components of [Captain](./src/captain/READ
 ## Components
 
 ### Strategy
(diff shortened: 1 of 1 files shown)


ironfish/README.md
```diff
@@ -26,10 +26,10 @@ This represents the IronfishConfig and all of it's options. It's a hierarchical
 ```
 
 ### FileSystem
-This is an abtraction on top of any file system related APIs like `path` and `fs` in node. It makes it so you can perform file and file system related methods in a way that works in the browser and node. `NodeFileSystem` is one implementation that works for node.
+This is an abtraction on top of any file system-related APIs like `path` and `fs` in node. It makes it so you can perform file and file system-related methods in a way that works in the browser and node. `NodeFileSystem` is one implementation that works for node.
```
NullSoldier, line 29: Isn't the word here "file system" not "system-related"? What's the explanation here?

PR #661, opened 2021-12-09
Redundant allocation when sending headerBytes to workers

kernel32uiuc:
According to [Buffer API](https://nodejs.org/api/buffer.html#static-method-bufferfromarray), Buffer.from() copies the passed buffer data onto a new Buffer instance if the argument is Uint8Array. When we send headerBytes to the worker, it is casted to Uint8Array and then we call Buffer.from(). It seems unnecessary and we can keep it in Buffer.

### DIFF
--- a/ironfish/src/workerPool/tasks/mineHeader.ts
+++ b/ironfish/src/workerPool/tasks/mineHeader.ts
@@ -8,7 +8,7 @@ import { mineHeader } from '../../mining/mineHeader'
 export type MineHeaderRequest = {
   type: 'mineHeader'
   batchSize: number
-  headerBytesWithoutRandomness: Uint8Array
+  headerBytesWithoutRandomness: Buffer
   initialRandomness: number
   miningRequestId: number
   targetValue: string
@@ -33,7 +33,7 @@ export function handleMineHeader(
 ): MineHeaderResponse {
   const result = mineHeader({
     batchSize,
-    headerBytesWithoutRandomness: Buffer.from(headerBytesWithoutRandomness),
+    headerBytesWithoutRandomness,
     initialRandomness,
     miningRequestId,
     targetValue,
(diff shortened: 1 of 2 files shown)


ironfish/src/workerPool/tasks/mineHeader.ts
```diff
@@ -8,7 +8,7 @@ import { mineHeader } from '../../mining/mineHeader'
 export type MineHeaderRequest = {
   type: 'mineHeader'
   batchSize: number
-  headerBytesWithoutRandomness: Uint8Array
+  headerBytesWithoutRandomness: Buffer
```
dguenther, line 11: NodeJS uses [object cloning](https://nodejs.org/api/worker_threads.html#considerations-when-cloning-objects-with-prototypes-classes-and-accessors) when transferring data to a worker thread, so `headerBytesWithoutRandomness` will end up as a `Uint8Array` when the MineHeaderRequest is serialized and sent to the worker.

So for this reason, I'd like to leave this type as `Uint8Array`. However, in `mining/mineHeader`, I think that type can be changed to `Uint8Array`, and then we can avoid the same allocation.

dguenther (changes requested):

ironfish/src/workerPool/tasks/mineHeader.ts
kernel32uiuc: Thanks for the explanation. Done with the changes.

dguenther (approved):
👍 Thanks for the change! Retargeting to staging and merging.

kernel32uiuc:
Is this PR also eligible to claim points for?

dguenther:
Yep! If you could submit it through the Claim Points form on our website, we'll get around to adding points for it in the near future.

Issue #507, opened 2021-12-01
Node crashes when running 'ironfish logs'


jackkates:
Node crashes with BigInt (maybe the difficulty?) serialization error when tailing the logs. Occurs in both verbose and normal mode, and happens especially if you are also running the miner. 

![image](https://user-images.githubusercontent.com/1587680/144322644-6c83c6d2-64f6-4eb3-96c7-3fe553a5476f.png)

```WebSocket server started at :::9033

Connected to the Iron Fish network

Starting sync from AxMVf78. work: +2670289461, ours: 4051, theirs: 4052

Finding ancestor using linear search on last 3 blocks starting at 00000...9dc00 (4051) from peer AxMVf78 at 4052

Found peer AxMVf78 ancestor 00000...9dc00, syncing from 4051 -> 4052 (1) after 1 requests

Requesting 20 blocks starting at 00000...9dc00 (4051) from AxMVf78

Finished syncing 1 blocks from AxMVf78

/usr/local/Cellar/ironfish/17/libexec/node_modules/js-message/Message.js:38

        return JSON.stringify(

                    ^

TypeError: Do not know how to serialize a BigInt

    at JSON.stringify (<anonymous>)

    at Message.getJSON [as JSON] (/usr/local/Cellar/ironfish/17/libexec/node_modules/js-message/Message.js:38:21)

    at Parser.format (/usr/local/Cellar/ironfish/17/libexec/node_modules/node-ipc/entities/EventParser.js:21:21)

    at Server.emit (/usr/local/Cellar/ironfish/17/libexec/node_modules/node-ipc/dao/socketServer.js:79:29)

    at IpcAdapter.emitStream (/usr/local/Cellar/ironfish/17/libexec/node_modules/ironfish/build/src/rpc/adapters/ipcAdapter.js:206:21)

    at Request.onStream (/usr/local/Cellar/ironfish/17/libexec/node_modules/ironfish/build/src/rpc/adapters/ipcAdapter.js:180:18)

    at Request.stream (/usr/local/Cellar/ironfish/17/libexec/node_modules/ironfish/build/src/rpc/request.js:41:14)

    at InterceptReporter.callback (/usr/local/Cellar/ironfish/17/libexec/node_modules/ironfish/build/src/rpc/routes/node/getLogStream.js:41:17)

    at InterceptReporter.log (/usr/local/Cellar/ironfish/17/libexec/node_modules/ironfish/build/src/logger/reporters/intercept.js:12:14)

[...]

PR #672, opened 2021-12-09
Prevent node crashing when logger fails to serialize BigInt by using string interpolation

Vovans:
Addressing https://github.com/iron-fish/ironfish/issues/507 
The problem was with a logger which unsuccessfully tried to serialize an object with a BigInt value (newBlock.header.target). This was the cause of a node crash. It would be possible to fix this by overriding the JSON.stringify, but it is deep down in js-message library, so an easier fix is just to interpolate the value to a string since it is only used for debugging purposes.

Maybe it is a good idea to wrap the InterceptReporter.log with a try/catch, since it is really not that good if logging crashes things.

### DIFF
--- a/ironfish-cli/src/commands/logs.ts
+++ b/ironfish-cli/src/commands/logs.ts
@@ -32,7 +32,6 @@ export default class LogsCommand extends IronfishCommand {
       })
     }
 
-
     this.exit(0)
   }
 }
--- a/ironfish-cli/src/commands/logs.ts
+++ b/ironfish-cli/src/commands/logs.ts
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { logType } from 'consola'
-import { ConsoleReporterInstance, IronfishNode } from 'ironfish'
+import { ConsoleReporterInstance, IJSON, IronfishNode } from 'ironfish'
 import { IronfishCommand } from '../command'
 import { RemoteFlags } from '../flags'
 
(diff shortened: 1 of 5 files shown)


dguenther (changes requested):
Thanks for looking at this, it'd be great to get this fixed! I think your comments are alluding to this, but it'd be nice to make this a more comprehensive fix in case we ever add more bigint logs in the future.

We have an extended JSON serializer in the codebase here to handle buffers and bigints: https://github.com/iron-fish/ironfish/blob/b2a53668e784ffeac1ea3fa9b7f6e5a0370064e8/ironfish/src/serde/iJson.ts#L11

I think if we call that on `args`, it would behave similarly to `console.log`, but avoid the error: https://github.com/iron-fish/ironfish/blob/b2a53668e784ffeac1ea3fa9b7f6e5a0370064e8/ironfish/src/rpc/routes/node/getLogStream.ts#L44

Then this test will need to be updated as well: https://github.com/iron-fish/ironfish/blob/b2a53668e784ffeac1ea3fa9b7f6e5a0370064e8/ironfish/src/rpc/routes/node/getLogStream.test.ts#L28

I'm open to considering other serialization strategies as well if they produce results similar to `console.log`. Thanks for looking at this!

Vovans:
Sorry, the last push was a mistake. I'll change it in a bit.

Vovans:
So the node is not getting crashed because of the problem in js-message library (it uses JSON.stringify, which can't handle bigints). Now we are serializing logObj.args before that. 

Additionally, we improve logs a bit. As was mentioned in this issue https://github.com/iron-fish/ironfish/issues/776, it was unexpected "\n"'s.

NullSoldier:
Can you rebase this PR and get the lint to pass before we review it? This fix is valuable and I don't want to steal it from you.

Vovans:
Rebased, and lint is passed

dguenther:
Sorry for the long response time on this over the holidays! I'm going to take over getting this merged today.

ironfish-cli/src/commands/logs.ts
```diff
@@ -23,11 +23,19 @@ export default class LogsCommand extends IronfishCommand {
     const response = this.sdk.client.getLogStream()
 
     for await (const value of response.contentStream()) {
+      let parsedArgs
+      try {
+        parsedArgs = IJSON.parse(value.args) as unknown[]
+      } catch {
+        this.logger.error(`Failed to deserialize args: ${value.args}`)
+        continue
+      }
```
NullSoldier, line 32: Should we let this crash, or render the error? So users can send it to us? These will be things we need to fix, right?

I'd like to either let it crash, or just use `ErrorUtils.renderError(error, true)`

dguenther: I'll re-throw the error here so it crashes

dguenther (approved):

PR #673, opened 2021-12-09
Use lower case hash in explorer links

alxs:
Lower case [works](https://explorer.ironfish.network/blocks/000000000072a6dee9200ab0fd60764a432b074cc594d84236172815bdb069dc) while upper case [doesn't](https://explorer.ironfish.network/blocks/000000000072A6DEE9200AB0FD60764A432B074CC594D84236172815BDB069DC).

### DIFF
--- a/ironfish-cli/src/commands/miners/mined.ts
+++ b/ironfish-cli/src/commands/miners/mined.ts
@@ -71,7 +71,7 @@ export class MinedCommand extends IronfishCommand {
         const amount = MathUtils.round(oreToIron(block.minersFee), 2)
 
         const link = linkText(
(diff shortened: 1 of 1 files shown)


dguenther:
Thanks for the PR! We ended up fixing this on the API side so that the API more generally handles uppercase/lowercase hashes, so I'm going to close this PR. But feel free to still claim points testnet points if you're interested -- we haven't open-sourced the testnet API yet or I would have asked if you'd like to make the change there instead.

alxs:
Works, thank you. Though if the API can handle both, the conversion is superfluous, so I would still merge this. But of course that's up to you.

PR #698, opened 2021-12-10
IRO-1379: Update blake3-wasm to blake3 (native)

yuryironfish:


leanthebean:
Does this increase hash rate? Looks like the most recent version of this library is 2.1.7, is there a reason we're using an older version?

NullSoldier:
I'm getting about 15% faster hash rate on this change.

NullSoldier (approved):

CumpsD:
Does this fix the OOM wasm error on startup for machines with a lot of cores/memory?

gjfrn:
As per https://github.com/iron-fish/ironfish/issues/732, no performance improvement was observed with zen2 architecture CPUs. This issue has now been resolved with PR #815 in 1.1.15.

Issue #702, opened 2021-12-11
Mined blocks added to forks are counted as 'mined' blocks. 


kernel32uiuc:
The blocks mined locally but added to forks should not be reported as mined blocks. status command will report those blocks as 'mined':

./ironfish status:

Version 0.1.9 @ src
Node STARTED
Mining STARTED - 1 miners, X mined
...

------------

Successfully mined block 00000000076ef9860eba6ba1009946f9d6f3573e3f5ad2d0fdeb06177a6ade01 (**15879**) has 2 transactions

Added block to fork seq: **15879**, head-seq: 15879, hash: 00000...ade01, head-hash: 00000...52ba7, work: 1221616441828749, head-work: 1221616441828749, work-diff: 0

deekerno:
Fixed as of #704.

PR #704, opened 2021-12-11
Fix mined stats and logs for forks

kernel32uiuc:
## Summary
There is too much confusion about whether we (as testnet users) have mined a block successfully or not. The reason is mostly about the logs. Ironfish logs the following message for the failed blocks as well. 
> _Successfully mined block 00000000076ef9860eba6ba1009946f9d6f3573e3f5ad2d0fdeb06177a6ade01 (15879) has 2 transactions_

miners:mined is also accurate, but it is very slow and occasionally hangs. That's why it is really important for users to easily understand whether they mine or not. 

This PR fixes this issue. When we log the 'Successfully mined block' message, it is now always accurate. 'status' command will also give the accurate stats for the current run.

Addressing #702.

## Breaking Change
```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -237,6 +237,7 @@ export class Blockchain {
 
     const result = await this.addBlock(genesis)
     Assert.isTrue(result.isAdded, `Could not seed genesis: ${result.reason || 'unknown'}`)
+    Assert.isTrue(result.isFork === false)
 
     const genesisHeader = await this.getHeaderAtSequence(GENESIS_BLOCK_SEQUENCE)
     Assert.isNotNull(
@@ -313,9 +314,11 @@ export class Blockchain {
 
   async addBlock(block: Block): Promise<{
     isAdded: boolean
+    isFork: boolean | null
     reason: VerificationResultReason | null
     score: number | null
   }> {
+    let isFork = null
     try {
       await this.db.transaction(async (tx) => {
         const hash = block.header.recomputeHash()
@@ -349,16 +352,18 @@ export class Blockchain {
 
         await this.connect(block, previous, tx)
 
+        isFork = !this.head.hash.equals(block.header.hash)
+
         await this.resolveOrphans(block)
       })
     } catch (e) {
       if (e instanceof VerifyError) {
-        return { isAdded: false, reason: e.reason, score: e.score }
+        return { isAdded: false, isFork: isFork, reason: e.reason, score: e.score }
       }
       throw e
     }
 
-    return { isAdded: true, reason: null, score: null }
+    return { isAdded: true, isFork: isFork, reason: null, score: null }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


ironfish/src/blockchain/blockchain.ts
```diff
@@ -237,6 +237,7 @@ export class Blockchain {
 
     const result = await this.addBlock(genesis)
     Assert.isTrue(result.isAdded, `Could not seed genesis: ${result.reason || 'unknown'}`)
+    Assert.isTrue(result.isFork === false)
```
NullSoldier, line 240:
```suggestion
    Assert.isFalse(result.isFork)
```

kernel32uiuc: isFork can be null, so explicit false check is needed.

sureshvenkata12:
@NullSoldier any issues with the PR?

kernel32uiuc:
@NullSoldier @dguenther any updates on the PR?

kernel32uiuc:
@NullSoldier I guess, this is not the direction you would like to proceed. Should I go ahead and abort this PR?

ironfish/src/mining/director.ts
```diff
@@ -463,20 +463,22 @@ export class MiningDirector {
...
 
-    this.blocksMined++
+    if (!isFork) {
+      this.logger.info(
+        `Successfully mined block ${block.header.hash.toString('hex')} (${
+          block.header.sequence
+        }) has ${block.transactions.length} transactions`,
+      )
+
+      this.blocksMined++
```
NullSoldier, line 480: Let's inverse this so it just returns and doesn't gossip a block on a fork.

ironfish/src/blockchain/blockchain.ts
```diff
@@ -347,18 +351,18 @@ export class Blockchain {
...
     } catch (e) {
       if (e instanceof VerifyError) {
-        return { isAdded: false, reason: e.reason, score: e.score }
+        return { isAdded: false, isFork: null, reason: e.reason, score: e.score }
       }
       throw e
     }
 
-    return { isAdded: true, reason: null, score: null }
+    return { isAdded: true, isFork: isFork, reason: null, score: null }
```
NullSoldier, line 365:
```suggestion
    const isFork = !this.head.hash.equals(block.header)
    return { isAdded: true, isFork: isFork, reason: null, score: null }
```

ironfish/src/mining/director.ts
cryptokur: Do you mean adding a new MINED_RESULT such as `MINED_RESULT.FORK` and return that in case of `isFork`

kernel32uiuc: something like that? 
`if (isFork) return MINED_RESULT.FORK`

ironfish/src/blockchain/blockchain.ts
kernel32uiuc: sounds good. Why do we need a mutex lock here? Or did you just make a typo?

kernel32uiuc:
@NullSoldier I addressed the feedback. Could you please review the changes?

dguenther:
Yep, I'll take a look at this and try to get it through!

ironfish/src/blockchain/blockchain.ts
NullSoldier: Two add blocks running at the same time can cause bugs.

NullSoldier (approved):
Directionally approved, derek is going to make small changes and to get this in.

kernel32uiuc:
> Directionally approved, derek is going to make small changes and to get this in.

glad that this got merged. thanks for the help

PR #736, opened 2021-12-13
Fix codecov token upload

dguenther:
## Summary

Attempt to fix codecov upload failing on PRs from forks.

The codecov action doesn't require a token for public repositories, so that's why I deleted the token from Rust CI. We should try out the github action for codecov upload for the JS CI when we have some time.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/.github/workflows/rust_ci.yml
+++ b/.github/workflows/rust_ci.yml
@@ -92,7 +92,6 @@ jobs:
       - name: Upload to codecov.io
         uses: codecov/codecov-action@v2.1.0
         with:
-          token: ${{secrets.CODECOV_TOKEN}}
           flags: ironfish-rust-nodejs
 
       # fmt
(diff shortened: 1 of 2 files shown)


brekk (approved):

.github/workflows/rust_ci.yml
```diff
@@ -137,7 +136,6 @@ jobs:
       - name: Upload to codecov.io
         uses: codecov/codecov-action@v2.1.0
         with:
-          token: ${{secrets.CODECOV_TOKEN}}
           flags: ironfish-rust-wasm
 
       # fmt
```
NullSoldier, line 141: Will other people be able to upload to our code cov?

dguenther: It's not anything we configure in our codecov settings, looks like they use the GitHub Actions environment variables to tag the build appropriately:

https://github.com/codecov/codecov-action/issues/29

https://github.com/codecov/codecov-action/releases/tag/v1.0.6

NullSoldier (approved):

PR #738, opened 2021-12-14
Set maximum node workers to 6 using new option "nodeWorkersMax"

NullSoldier:
## Summary
We're making this change because many users with very high core
machines are running out of memory because our node is spawning too many
workers. This is causing their machine to run out of memory more than
they expect.

Based on our optimizations I don't think anyone needs to use more than 6
workers. They are mostly idle all the time now. In the case that the
user needs more they can set "nodeWorkersMax" to -1 to disable the cap
completely.

This only is used when auto scaling is used, because we always respect the
end users desired worker configuration if they set it manually.

## Testing Plan

Ran the code, and checked how many workers were created. I also set it to -1 to ensure it still auto scaled up.

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/node.ts
+++ b/ironfish/src/node.ts
@@ -169,7 +169,7 @@ export class IronfishNode {
     if (workers === -1) {
       workers = os.cpus().length - 1
     }
-    const workerPool = new WorkerPool({ metrics, maxWorkers: workers })
+    const workerPool = new WorkerPool({ metrics, numWorkers: workers })
 
     strategyClass = strategyClass || Strategy
     const strategy = new strategyClass(workerPool)
--- a/ironfish/src/node.ts
+++ b/ironfish/src/node.ts
@@ -168,6 +168,11 @@ export class IronfishNode {
     let workers = config.get('nodeWorkers')
     if (workers === -1) {
       workers = os.cpus().length - 1
+
+      const maxWorkers = config.get('nodeWorkersMax')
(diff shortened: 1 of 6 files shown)


brekk (approved):

ironfish/src/node.ts
```diff
@@ -168,8 +168,13 @@ export class IronfishNode {
     let workers = config.get('nodeWorkers')
     if (workers === -1) {
       workers = os.cpus().length - 1
+
+      const maxWorkers = config.get('nodeWorkersMax')
+      if (maxWorkers !== -1) {
```
dguenther, line 173: This is a nitpick, but I wonder if we should at least log if `nodeWorkers` is explicitly set to a value that's greater than `nodeWorkersMax`. It might be confusing to pass e.g. `--workers 8` and have it appear that nothing happens.

dguenther (approved):

ironfish/src/node.ts
NullSoldier: By nothing happening, you mean that it doesn't cap it? Is the confusion that this only affects auto scaling? Like you would expect this to affect non auto scaling?

dguenther: Oh sorry, I misread it and thought it affected non-autoscaling. Looks good

dguenther:
Changing the base to `staging` and merging

PR #749, opened 2021-12-14
Update yarn.lock with blake3 change

dguenther:
## Summary

#698 replaced blake3-wasm with blake3, but we needed to update the `yarn.lock` with this change as well.

## Breaking Change

```
[ ] Yes
[x] No
```


yuryironfish (approved):
APPROVED

Issue #619, opened 2021-12-07
Failed to add mined block to chain with reason Invalid spend


gjfrn:
I have successfully mined and added 3 blocks so far, but my fourth one has failed due to 'Invalid spend':

Successfully mined block 00000000036e8cc05a70361127a9e52ab25708b6b1953550f1ec03a16b75f036 (12313) has 11 transactions
Failed to add mined block to chain with reason Invalid spend

gjfrn:
Appears to be a different issue than https://github.com/iron-fish/ironfish/pull/393.

gjfrn:
I have had this happen for the next two blocks that I mined:
Successfully mined block 0000000006103af6a808c892723c512d885ca85cf8789f76f74600e0c3353807 (12340) has 4 transactions
Failed to add mined block to chain with reason Invalid spend

Successfully mined block 000000000b46f32e950678b7c2948e8de66a5ade4ef2a0b62e3bfcedd7ba57e4 (12494) has 5 transactions
Failed to add mined block to chain with reason Invalid spend

gjfrn:
Two more examples today:
Successfully mined block 0000000006f8e7df22aa2d90f69a5b38474cbb34b55480a87602a8e6db4f700d (13550) has 3 transactions
Failed to add mined block to chain with reason Invalid spend
Successfully mined block 0000000004dbdff138f3ba6dff62b21eb690f3ec92c559b73af785d39172bed8 (13846) has 2 transactions
Failed to add mined block to chain with reason Invalid spend

gjfrn:
Seems there are a number of other users reporting this recently since my original report so definitely an ongoing issue.

NullSoldier:
Users are reporting a suspicious amount of invalid add blocks with invalid spends. See ther example below. In one case, a stuck block that took 4 hours to mine had this issue which was very frustrating because it took a long time to mine.


```
Successfully mined block 0000000005afd3ac33be1082740dd909d7261d5816c76438490befb43489fc33 (13561) has 2 transactions
Failed to add mined block to chain with reason Invalid spend
Successfully mined block 0000000008b696906d12909a6049e7779e544eff1ea5c56db03c30e33f1c9599 (13561) has 2 transactions
Failed to add mined block to chain with reason Invalid spend
```

1frkmtk1:
Same problem
The problem appeared after updating to version 0.1.9.
Graffiti: frkmtk

[...]

PR #750, opened 2021-12-14
Check transactions for invalid spends before adding to block

dguenther:
## Summary

Fixes an issue where mined blocks would get rejected when calling `addBlock`, with the reason given `Failed to add mined block to chain with reason Invalid spend`. See https://github.com/iron-fish/ironfish/issues/619 for an example. 

Factors out the spend verification from `verifyTransactionAdd` and calls it when we're fetching transactions out of the mempool. The mempool calls `transaction.verify` internally, but doesn't know about the chain, so that verification has to be done by the director now.

There may be more bugs causing this issue, but will need to mine for a bit to check it out, and this fixes at least the most obvious and testable one.

## Testing Plan

Added a unit test that fails on the current `staging`.

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/mining/director.ts
+++ b/ironfish/src/mining/director.ts
@@ -331,7 +331,7 @@ export class MiningDirector {
       throw Error('No miner account found to construct the transaction')
     }
 
-    const blockTransactions = []
+    const blockTransactions: Transaction[] = []
 
     // Fetch all transactions for the block
     for await (const transaction of this.getTransactions()) {
@@ -339,7 +339,6 @@ export class MiningDirector {
         break
       }
 
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
       blockTransactions.push(transaction)
     }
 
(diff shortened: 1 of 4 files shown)


ironfish/src/mining/director.ts
```diff
@@ -331,15 +331,14 @@ export class MiningDirector {
...
+    const blockTransactions: Transaction[] = []
 
     // Fetch all transactions for the block
     for await (const transaction of this.getTransactions()) {
       if (blockTransactions.length >= MAX_TRANSACTIONS_PER_BLOCK) {
         break
       }
 
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
       blockTransactions.push(transaction)
```
NullSoldier, line 342: Nice. I just saw this and fixed it in another PR. I can't believe this was using any here...

ironfish/src/mining/director.ts
```diff
@@ -503,22 +502,19 @@ export class MiningDirector {
...
-          return true
-        }
-
-        if (await this.chain.nullifiers.contains(spend.nullifier)) {
-          return true
-        }
-
-        return false
+      const conflicted = await AsyncUtils.find(transaction.spends(), (spend) => {
+        return Promise.resolve(nullifiers.has(spend.nullifier))
```
NullSoldier, line 506: I wonder if I can update the types here in another PR so that the async version can at least take a non async method.

NullSoldier:
This is a good fix, when I wrote the original code I should have just used the verifier myself instead of re-writing spend verification. This is a good lesson in not duplicating code everywhere.

NullSoldier:
> ```
> 
> ```

I'm confused, this PR is about not adding invalid spends to blocks. What does the worker pool and memory have to do with this PR??

NullSoldier (approved):

Issue #725, opened 2021-12-12
[Bug] Peers disconnect upon requesting 20 blocks


CumpsD:
# TL;DR

## Running from source (Linux)

```bash
sed -i 's%const REQUEST_BLOCKS_PER_MESSAGE.*%const REQUEST_BLOCKS_PER_MESSAGE = 2%' ironfish/src/syncer.ts
yarn start:once start
```

## Running from source (Windows)

* Open ironfish/src/syncer.ts in Notepad
* Edit `const REQUEST_BLOCKS_PER_MESSAGE = 20` to `const REQUEST_BLOCKS_PER_MESSAGE = 2` 
* Save the file
* `yarn start:once start`

## Running from Docker (Linux and Windows)

**❗This assumes you did NOT run your container with `--rm` else the changes are gone on restart!!❗**

I expect your container to be run with:
`docker run -d --name ironfish --network host ghcr.io/iron-fish/ironfish:latest ........`

Find NAME_OF_YOUR_DOCKER_CONTAINER with `docker ps -a`

![image](https://user-images.githubusercontent.com/1201644/146088820-6f730840-20ad-456a-83e2-d68c91e13b85.png)

```bash
docker exec -t NAME_OF_YOUR_DOCKER_CONTAINER sh -c "sed -i 's%REQUEST_BLOCKS_PER_MESSAGE.*%REQUEST_BLOCKS_PER_MESSAGE = 2%' /usr/src/app/node_modules/ironfish/build/src/syncer.js"
docker restart NAME_OF_YOUR_DOCKER_CONTAINER
docker logs -f NAME_OF_YOUR_DOCKER_CONTAINER
```

------

# Problem

A lot of issues on GH and Discord about peers disconnecting while syncing. I've done some debugging:

Reproduction of the bug:
![image](https://user-images.githubusercontent.com/1201644/145732921-c0c68338-2607-48c8-aefc-a439320e7440.png)

My initial start:
https://github.com/iron-fish/ironfish/blob/master/ironfish/src/syncer.ts#L358-L362
```
      this.logger.info(
        `Requesting ${this.blocksPerMessage} blocks starting at ${HashUtils.renderHash(
          head,
        )} (${sequence}) from ${peer.displayName}`,
      )
```

I noticed blocksPerMessage is `20`, being passed in the ctor: https://github.com/iron-fish/ironfish/blob/master/ironfish/src/syncer.ts#L65

[...]

PR #753, opened 2021-12-15
Changing max message size for both websocket and

leanthebean:
## Summary
This past weekend a lot of people reported not being able to sync (see https://github.com/iron-fish/ironfish/issues/725 for instance). The problem is that some blocks are starting to get beefy, and when you sync in batches of 20 blocks that message size exceeds the default webrtc max message size. 

This PR sets the max message size for both our webrtc and websocket connections to 256MB to alleviate the problem for now. The real fix is to make sure our batch size doesn't exceed a certain size, which will come later. 

## Testing Plan

Tested locally for both webrtc and websocket connections 

## Breaking Change

No

### DIFF
--- a/ironfish/src/consensus/consensus.ts
+++ b/ironfish/src/consensus/consensus.ts
@@ -39,7 +39,7 @@ export const MAX_REQUESTED_BLOCKS = 50
  * Max size for a message, for instance when requesting batches of blocks
  * TODO 256MB is way too big
  */
-export const MAX_MESSAGE_SIZE = 268435456 //256mb
+export const MAX_MESSAGE_SIZE = 256 * 1024 * 1024
 
 /**
  * The average time that all blocks should be mined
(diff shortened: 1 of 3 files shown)


ironfish/src/consensus/consensus.ts
```diff
@@ -35,6 +35,12 @@ export const MAX_SYNCED_AGE_MS = 12 * 60 * 60 * 1000
  */
 export const MAX_REQUESTED_BLOCKS = 50
 
+/**
+ * Max size for a message, for instance when requesting batches of blocks
+ * TODO 256MB is way too big
+ */
+export const MAX_MESSAGE_SIZE = 268435456 //256mb
```
NullSoldier, line 42:
```suggestion
export const MAX_MESSAGE_SIZE = 256 * 1024 * 1024
```

NullSoldier (approved):

PR #754, opened 2021-12-15
Allow syncer to sync with peers even when synced

NullSoldier:
## Summary
This issue was introduced in
https://github.com/iron-fish/ironfish/pull/647 which moved our sync
time definition back 10 hours. This solved another issue, but now when
people become synced they are suddenly stuck because the syncer won't
run anymore.

This is an experimental change, but it allows the syncer to run at any
time when someone has more work than us. Remember, that gossip also
still kicks off a sync as well.

## Testing Plan
Ran locally for awhile to test, but I am only syncing one block at a time because of the syncing bug outlined in the current issues. So I haven't been able to see what happens when you are fully synced.

## Breaking Change

[ ] Yes
[x] No


dguenther (approved):
Looks good, switching the target to `staging`

deekerno (approved):

PR #755, opened 2021-12-15
Don't require async predicate in AsyncUtils.find

NullSoldier:
## Summary
Just a simple fix that allows you to not half to return a promise in the
predicate.

## Testing Plan
Run lint

## Breaking Change

 * [ ] Yes
 * [x] No


dguenther (approved):
Looks good!

PR #758, opened 2021-12-15
Allow for Multiple Receipts in Transactions

deekerno:
## Summary
This PR continues from prior work in allowing for multiple outputs in transactions. It was not necessary to adjust transactions at the native code level and the worker pool already had multiple recipients. This is also the first step in allowing for bulk faucet transactions, which should speed up faucet processing considerably.

## Testing Plan
Tests have been adjusted to use the `receives` object (originally present in `createTransaction.ts` in the worker pool module) in the transaction request object; all tests pass locally.

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish-cli/src/commands/accounts/pay.ts
+++ b/ironfish-cli/src/commands/accounts/pay.ts
@@ -210,7 +210,9 @@ ${displayIronAmountWithCurrency(
       stopProgressBar()
 
       const transaction = result.content
-      const recipients = transaction.receives.map((receive) => receive.publicAddress).join(', ')
+      const recipients = transaction.receives
+        ?.map((receive) => receive.publicAddress)
+        .join(', ')
       this.log(`
 Sending ${displayIronAmountWithCurrency(amount, true)} to ${recipients} from ${
         transaction.fromAccountName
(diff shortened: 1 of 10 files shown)


ironfish-cli/src/commands/accounts/pay.ts
```diff
@@ -195,19 +195,26 @@ ${displayIronAmountWithCurrency(
...
+        ],
         fee: ironToOre(fee).toString(),
         expirationSequence,
       })
 
       stopProgressBar()
 
       const transaction = result.content
+      const recipients = transaction.receives
+        ?.map((receive) => receive.publicAddress)
```
dguenther, line 214: Can we delete the `?` here? I'm having trouble looking for the origin type to see if `transaction.receives` can ever be undefined/null

ironfish-cli/src/commands/accounts/pay.ts
```diff
@@ -195,19 +195,26 @@ ${displayIronAmountWithCurrency(
...
+        .join(', ')
       this.log(`
-Sending ${displayIronAmountWithCurrency(amount, true)} to ${transaction.toPublicKey} from ${
+Sending ${displayIronAmountWithCurrency(amount, true)} to ${recipients} from ${
```
dguenther, line 217: Just double-checking that this ends up being formatted okay

dguenther (approved):
This looks good, have you given it a test locally?

ironfish-cli/src/commands/accounts/pay.ts
deekerno: This does end up being formatted okay in my local testing as well as in isolation in a TypeScript REPL.

PR #769, opened 2021-12-16
Add timestamps in exportChain.ts

testingPurposes1:
## Summary
Some people would find it useful if exported chain will contain timestamps in it.
In this PR timestamp was added in the response schema, yup validation and populated to results.
One disadvantage of this PR is that chain:export will require offline node, because some variables and methods aren't accessible while node is running.
## Testing Plan
I ran chain:export and got this results

image

## Breaking Change

 [ ] Yes
 [x] No


NullSoldier:
This PR looks great. We can't merge community PR's yet until we figure out githubs and code covs build pipeline.

NullSoldier (approved):

shandongdacongge (approved):
it's ok

PR #773, opened 2021-12-17
Fix banning peers for expired transactions

NullSoldier:
## Summary

There is a bug in that we verify a transaction expiration based on the
current chain head, rather than on the block the transaction was added
to. This means if you sync an old fork from someone you are nearly
guaranteed to ban them.

Consider this chain
```
G -> A1 -> A2
  -> B1
```

If you have a transaction on B1 that expires at height 2 (expiration is
inclusive, meaning you have to include the transaction BEFORE that
expiration), then sync that chain from another node, you'll ban them.
From your perspective, B1 has an expired transaction because you
have block A2 as your head.

The fix is to validate transactions based on their block, and not your
current head.

## Testing Plan
Wrote a test, and ran my syncer for awhile and didn't see any banning.

## Breaking Change

 * [ ] Yes
 * [x] No

### DIFF
--- a/ironfish/src/account/accounts.ts
+++ b/ironfish/src/account/accounts.ts
@@ -701,7 +701,8 @@ export class Accounts {
 
     expirationSequence =
       expirationSequence ?? heaviestHead.sequence + defaultTransactionExpirationSequenceDelta
-    if (this.chain.verifier.isExpiredSequence(expirationSequence)) {
+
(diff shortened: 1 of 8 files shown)


deekerno (approved):
Looks good once the lint fixes are in.

deekerno (approved):

rohanjadvani (approved):

PR #775, opened 2021-12-17
Optimize MerkleTree.contains from O(n) to O(1)

leanthebean:
## Summary
In the merkle tree file, we have a `contained` method. Currently, there's a linear scan that goes through _**all**_ of the existing nullifiers when determining if a transaction is valid (e.g. not a double spend). As we get more transactions, this linear scan becomes more and more inefficient (e.g. beefy later blocks would take several seconds). 

This change adds another db store that lets you check in O(1) if an element exists in a merkle tree. 

## Testing Plan
Tested it locally and avg block time is about 20ms per transaction per block 

## Breaking Change

This requires a db upgrade. 

 * [x] Yes
 * [ ] No

### DIFF
--- a/ironfish/src/consensus/verifier.ts
+++ b/ironfish/src/consensus/verifier.ts
@@ -170,8 +170,10 @@ export class Verifier {
     tx?: IDatabaseTransaction,
   ): Promise<VerificationResult> {
     return this.chain.db.withTransaction(tx, async (tx) => {
+      const nullifierSize = await this.chain.nullifiers.size(tx)
+
       for (const spend of transaction.spends()) {
-        const reason = await this.verifySpend(spend, tx)
+        const reason = await this.verifySpend(spend, nullifierSize, tx)
 
         if (reason) {
           return { valid: false, reason }
@@ -303,12 +305,15 @@ export class Verifier {
       const spendsInThisBlock = Array.from(block.spends())
       const processedSpends = new BufferSet()
 
+      const previousNullifierSize =
+        block.header.nullifierCommitment.size - spendsInThisBlock.length
+
       for (const spend of spendsInThisBlock.values()) {
         if (processedSpends.has(spend.nullifier)) {
           return { valid: false, reason: VerificationResultReason.DOUBLE_SPEND }
         }
 
-        const verificationError = await this.verifySpend(spend, tx)
+        const verificationError = await this.verifySpend(spend, previousNullifierSize, tx)
         if (verificationError) {
           return { valid: false, reason: verificationError }
         }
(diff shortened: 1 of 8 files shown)


ironfish/src/consensus/verifier.ts
```diff
@@ -167,10 +167,8 @@ export class Verifier {
     tx?: IDatabaseTransaction,
   ): Promise<VerificationResult> {
     return this.chain.db.withTransaction(tx, async (tx) => {
-      const noteSize = await this.chain.notes.size(tx)
```
dguenther, line 170: Ah, looks like this might have been a bug, good spot! If possible it might be nice to pull this out into a separate PR so any issues with it aren't conflated with the LeavesIndex cache.

dguenther (approved):
From a code review standpoint, looks good -- I think @NullSoldier wants to add a perf test to this before it merges, and it'd be nice to run it on a laptop node for a bit to see if any issues happen to pop up.

dguenther:
It looks like there are some test failures in here as well

ironfish/src/consensus/verifier.ts
NullSoldier: Moved that fix here, https://github.com/iron-fish/ironfish/pull/798

NullSoldier:
Here is the perf test @dguenther https://github.com/iron-fish/ironfish/pull/799/files

ironfish/src/merkletree/merkletree.ts
```diff
@@ -339,6 +353,25 @@ export class MerkleTree<E, H, SE extends JsonSerializable, SH extends JsonSerial
...
+      {
+        index: index,
+        element: value.element,
+        merkleHash: value.merkleHash,
+        parentIndex: value.parentIndex,
+      },
+      tx,
+    )
+
+    await this.leavesIndex.put(this.hasher.merkleHash(value.element), index, tx)
```
NullSoldier, line 372:
```suggestion
    await this.leavesIndex.put(value.merkleHash, index, tx)
```

ironfish/src/consensus/verifier.ts
```diff
@@ -296,16 +294,15 @@ export class Verifier {
   async hasValidSpends(block: Block, tx?: IDatabaseTransaction): Promise<VerificationResult> {
```
NullSoldier, line 294: The errornously removed code was intending to check for  `the note being spent really existed in the tree at the time it was spent` but was not. Should this still check for that some how? The spend size is the size of the nullifier tree when the commitment was created, not the note tree right?

NullSoldier:
We still need to clean up the index in the `MerkleTree.truncate` function.

ironfish/src/consensus/verifier.ts
```diff
@@ -327,22 +331,21 @@ export class Verifier {
    * spend's spend root.
    *
    * @param spend the spend to be verified
-   * @param size the size of the nullifiers tree at which the spend must not exist
```
dguenther, line 330: This should probably be added back, or we should kill the whole comment so it's at least somewhat consistent

dguenther (approved):

ironfish/src/consensus/verifier.ts
NullSoldier: Fixed

NullSoldier (approved):

Issue #776, opened 2021-12-17
redundant line breaks in the log of invalid blocks


shandongdacongge:
There are redundant line breaks in the log of invalid blocks, affecting readability, as shown in the following figure

The source code is https://github.com/iron-fish/ironfish/blob/73a192a784e60bcaa7f0ff284cfa355e1572a081/ironfish/src/syncer.ts#L459

![1639721894(1)](https://user-images.githubusercontent.com/64308935/146498490-8f9392cd-36d8-45f2-a318-87445963a7f2.png)

deekerno:
Fixed as a result of #777 and went live as a part of #803.

PR #777, opened 2021-12-17
redundant line breaks in the log of invalid blocks

shandongdacongge:
redundant line breaks in the log of invalid blocks
This issue was introduced in #776 
There are redundant line breaks in the log of invalid blocks, affecting readability

This solved the issue.

## Breaking Change

 * [ ] Yes
 * [x] No


shandongdacongge:
remove the redundant line breaks in the log of invalid blocks

dguenther (approved):
Looks good, thanks!

shandongdacongge (commented):
OK

PR #784, opened 2021-12-17
Allow for Bulk Faucet Transactions

deekerno:
## Summary
These changes allow the faucet service command to get multiple requests from the API so that faucet transactions can be done in bulk.

## Testing Plan

1. Adjust your configuration with the following values:
-- `getFundsApi`: `'http://localhost:8003/faucet_transactions'`
-- `miningForce`:  true
2. Mock your own chain by doing the following:
-- Adjust `ironfish-cli/scripts/run-network-macos.sh` to include a miner, using `~/.ironfish1` as its data folder
-- Start one node using itself as a bootstrap node:
--- `yarn start:once start -b 127.0.0.1:9033 --datadir=~/batch_faucet_test`
-- Run `run-network-macos.sh` and allow for blocks to be mined on the chain.
3. Set the following environment variables:
-- `IRONFISH_API_HOST="http://localhost:8003"`
-- `IRONFISH_API_TOKEN="test"`
4. Start the Iron Fish API locally using the instructions in its README.
5. Open a terminal and check the balances of the accounts:
-- `yarn start:once accounts:balance --datadir=~/.ironfish2`
-- `yarn start:once accounts:balance --datadir=~/.ironfish3`
6. Send a request from each account to the faucet using one of the other created data folders:
--`yarn start:once faucet --datadir=~/.ironfish2`
--`yarn start:once faucet --datadir=~/.ironfish3`
7. At this time, you can check the `faucet_transactions` table in your local API database using your preferred tool.
-- You should see your pending requests in the table with their public keys corresponding to each data folder.
8. Open up another terminal and run `yarn start:once service:faucet --datadir=~/.ironfish1`. You should then the client connect to the node, fetch your requests from the local API, and batch them into one transaction. The client should then report that there are no faucet jobs as the amount of requests will be smaller than the max requests per transaction (this is set to 5 at this time).
9. Check the balances of the accounts once more; they should have increased by the intended faucet amount (5 ORE at this time).
10. Check the `faucet_transactions` table in your local API database for the two transactions; the requests should have a start date and a completion date, and they should also have the same transaction hash.

You can repeat steps 6-10 to generate more requests than are allowed to be included in a transaction; the client will continue to process them all until there are no more unfulfilled requests. Keep in mind that we enforce a limit on requests per public key or email; if you hit this limit, you can delete the second and third data folders and re-run the network mocking script to recreate a new account.

## Breaking Change

- [ ] Yes
- [x] No

The API currently supports behavior from both prior to this PR (only gets a single faucet request) and after this PR would be merged (gets multiple faucet requests). After this PR is merged, the API will be changed to only allow for multiple requests.

### DIFF
--- a/ironfish-cli/src/commands/service/faucet.ts
+++ b/ironfish-cli/src/commands/service/faucet.ts
@@ -186,9 +186,11 @@ export default class Faucet extends IronfishCommand {
     speed.add(1)
 
     this.log(
-      `COMPLETING: ${JSON.stringify(faucetTransactions, ['id', 'public_key'], '   ')} ${
-        tx.content.hash
-      } (5m avg ${speed.rate5m.toFixed(2)})`,
+      `COMPLETING: ${JSON.stringify(
+        faucetTransactions,
+        ['id', 'public_key', 'started_at'],
+        '   ',
+      )} ${tx.content.hash} (5m avg ${speed.rate5m.toFixed(2)})`,
     )
 
     for (const faucetTransaction of faucetTransactions) {
--- a/ironfish-cli/src/commands/service/faucet.ts
+++ b/ironfish-cli/src/commands/service/faucet.ts
@@ -8,6 +8,7 @@ import { RemoteFlags } from '../../flags'
 
 const FAUCET_AMOUNT = 5
 const FAUCET_FEE = 1
+const MAX_RECIPIENTS_PER_TRANSACTION = 5
 
 export default class Faucet extends IronfishCommand {
   static hidden = true
@@ -126,12 +127,15 @@ export default class Faucet extends IronfishCommand {
 
     const response = await client.getAccountBalance({ account })
 
-    if (BigInt(response.content.confirmed) < BigInt(FAUCET_AMOUNT + FAUCET_FEE)) {
+    if (
+      BigInt(response.content.confirmed) <
+      BigInt(FAUCET_AMOUNT * MAX_RECIPIENTS_PER_TRANSACTION + FAUCET_FEE)
+    ) {
       if (!this.warnedFund) {
         this.log(
-          `Faucet has insufficient funds. Needs ${FAUCET_AMOUNT + FAUCET_FEE} but has ${
-            response.content.confirmed
-          }. Waiting on more funds.`,
+          `Faucet has insufficient funds. Needs ${
+            FAUCET_AMOUNT * MAX_RECIPIENTS_PER_TRANSACTION + FAUCET_FEE
+          } but has ${response.content.confirmed}. Waiting on more funds.`,
         )
 
         this.warnedFund = true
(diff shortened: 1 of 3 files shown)


ironfish/src/webApi.ts
```diff
@@ -93,16 +93,16 @@ export class WebApi {
...
-        `${this.host}/faucet_transactions/next`,
+      const response = await axios.get<{ object: string; data: FaucetTransaction[] }>(
+        `${this.host}/faucet_transactions/next?count=${count}`,
         this.options(),
       )
 
-      return response.data
+      return response.data.data
     } catch (e) {
       if (IsAxiosError(e) && e.response?.status === 404) {
```
NullSoldier, line 107: I think the API here is a bit strange. It seems to return null because the old API used to return null when it got a 404, which was used by the API to indicate there were no results left to process. In an API that sends a list, it should just send an empty list when there are no items to process.

ironfish/src/webApi.ts
```diff
@@ -93,16 +93,16 @@ export class WebApi {
...
   }
 
-  async getNextFaucetTransaction(): Promise<FaucetTransaction | null> {
+  async getNextFaucetTransactions(count: number): Promise<FaucetTransaction[] | null> {
     this.requireToken()
 
     try {
-      const response = await axios.get<FaucetTransaction>(
-        `${this.host}/faucet_transactions/next`,
+      const response = await axios.get<{ object: string; data: FaucetTransaction[] }>(
```
NullSoldier, line 100: what is "object" here? This is a pretty weird response type from the API. What is it sending back?

ironfish-cli/src/commands/service/faucet.ts
```diff
@@ -143,38 +147,44 @@ export default class Faucet extends IronfishCommand {
...
 
-    this.log(
-      `COMPLETING: ${faucetTransaction.id} ${tx.content.hash} (5m avg ${speed.rate5m.toFixed(
-        2,
-      )})`,
-    )
+    const ids = faucetTransactions.map((ft) => ft.id).join(', ')
 
-    await api.completeFaucetTransaction(faucetTransaction.id, tx.content.hash)
+    this.log(`COMPLETING: ${ids} ${tx.content.hash} (5m avg ${speed.rate5m.toFixed(2)})`)
```
NullSoldier, line 184: Is this message easier to read this way? Do you prefer this over just doing what we did before?

deekerno: I don't prefer it actually; I intended on having a discussion on the topic, which is what we're in now! 👍🏽

ironfish-cli/src/commands/service/faucet.ts
```diff
@@ -126,12 +127,15 @@ export default class Faucet extends IronfishCommand {
...
+    ) {
       if (!this.warnedFund) {
         this.log(
-          `Faucet has insufficient funds. Needs ${FAUCET_AMOUNT + FAUCET_FEE} but has ${
-            response.content.confirmed
-          }. Waiting on more funds.`,
+          `Faucet has insufficient funds. Needs ${
+            FAUCET_AMOUNT * MAX_RECIPIENTS_PER_TRANSACTION + FAUCET_FEE
+          } but has ${response.content.confirmed}. Waiting on more funds.`,
         )
```
NullSoldier, line 139: It seems like we could just change this to do a best effort instead of failing if any of them cannot be sent. That way if we crank up the number, it'll still send some of them if it has partial money.

ironfish/src/webApi.ts
deekerno: The API is using the `List` type (which includes an object field [usually `list` in this case] and a data field) on this endpoint for consistency with other endpoints that return a list.

ironfish-cli/src/commands/service/faucet.ts
```diff
@@ -8,6 +8,7 @@ import { RemoteFlags } from '../../flags'
 
 const FAUCET_AMOUNT = 5
 const FAUCET_FEE = 1
+const MAX_RECIPIENTS_PER_TRANSACTION = 5
```
NullSoldier, line 11: If we have 5, it'll take 6 hours best case scenario, most likely double or triple this. If there any number want such a low number?

ironfish-cli/src/commands/service/faucet.ts
```diff
@@ -126,12 +127,15 @@ export default class Faucet extends IronfishCommand {
 
     const response = await client.getAccountBalance({ account })
 
-    if (BigInt(response.content.confirmed) < BigInt(FAUCET_AMOUNT + FAUCET_FEE)) {
+    if (
+      BigInt(response.content.confirmed) <
+      BigInt(FAUCET_AMOUNT * MAX_RECIPIENTS_PER_TRANSACTION + FAUCET_FEE)
+    ) {
```
NullSoldier, line 133: It seems like we could just change this to do a best effort instead of failing if any of them cannot be sent. That way if we crank up the number, it'll still send some of them if it has partial money.

ironfish-cli/src/commands/service/faucet.ts
deekerno: This was a test value intended for discussion.

ironfish/src/webApi.ts
deekerno: I've changed it to check for an empty list or a null value. This is to ensure that there is no downtime or errors when both the ability to return a single faucet request and the ability to return null are removed from the API. Once that happens, I can remove the null check and the empty list will be returned when there are no items to process.

ironfish/src/webApi.ts
NullSoldier: I mean that's fine, but I don't understand the point of the object field that's typed as a generic "string". What is the value? When can it not be list? Why is it returned to the front end in the first place? It's especially not even used here, so who would use it?

deekerno: Object field has been removed with no harm to functionality. 👍🏽

NullSoldier (approved):

PR #787, opened 2021-12-18
Abort the ongoing jobs from the previous request

kernel32uiuc:
## Summary
When a new mine request arrives, we should also discard the ongoing jobs on the workers from the previous request. When the batch size is large, those jobs might be long-running and harm the mining performance. 

## Testing Plan
Existing test suite.

## Breaking Change

```
[ ] Yes
[ X] No
```

### DIFF
--- a/ironfish/src/workerPool/pool.ts
+++ b/ironfish/src/workerPool/pool.ts
@@ -123,8 +123,7 @@ export class WorkerPool {
   }
 
   abortJobs(): void {
-    this.workers.forEach((worker) => 
-      worker.jobs.forEach((job) => job.abort()))
+    this.workers.forEach((worker) => worker.jobs.forEach((job) => job.abort()))
   }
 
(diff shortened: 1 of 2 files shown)


ironfish/src/workerPool/pool.ts
```diff
@@ -122,6 +122,10 @@ export class WorkerPool {
     await Promise.all(workers.map((w) => w.stop()))
   }
 
+  abortJobs(): void {
+    this.workers.forEach((worker) => worker.jobs.forEach((job) => job.abort()))
```
NullSoldier, line 126: It's possible this change doesn't do anything. Did you test that btw?

I saw some docs in node that say worker messages don't be processed if the event loop doesn't yield. We might need to introduce a yield in the miner event loop. I would be curious if the abort line in the actual mine job ever gets triggered.

kernel32uiuc: you are right. it does not do anything. I will abort the PR to look into the alternatives.

PR #798, opened 2021-12-21
Use nullifier tree size to verify spend

NullSoldier:
## Summary
We were using the notes tree size to verifier the spend, which is not
correct but shouldn't generally cause an issue because the note
tree is bigger than the nullifier tree.

## Testing Plan
Run node for awhile

## Breaking Change

```
[] Yes
[X] No
```

### DIFF
--- a/ironfish/src/consensus/verifier.ts
+++ b/ironfish/src/consensus/verifier.ts
@@ -170,10 +170,10 @@ export class Verifier {
     tx?: IDatabaseTransaction,
   ): Promise<VerificationResult> {
     return this.chain.db.withTransaction(tx, async (tx) => {
-      const noteSize = await this.chain.notes.size(tx)
+      const nullifierSize = await this.chain.nullifiers.size(tx)
 
       for (const spend of transaction.spends()) {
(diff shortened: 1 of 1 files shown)


deekerno (approved):

ironfish/src/consensus/verifier.ts
```diff
@@ -170,7 +170,7 @@ export class Verifier {
     tx?: IDatabaseTransaction,
   ): Promise<VerificationResult> {
     return this.chain.db.withTransaction(tx, async (tx) => {
-      const noteSize = await this.chain.notes.size(tx)
+      const noteSize = await this.chain.nullifiers.size(tx)
```
dguenther, line 173: This should be renamed to something like `nullifierSize` or `treeSize`. Otherwise, looks good

dguenther (changes requested):

ironfish/src/consensus/verifier.ts
NullSoldier: Fixed

dguenther (approved):

PR #808, opened 2021-12-21
Fix Faucet API Request

deekerno:
## Summary
The faucet was incorrectly configured to do the max amount of faucet requests possible in a transaction. Practically, we want to have an upper bound to this number. This PR also removes workarounds that were present for the API.

## Testing Plan
Manual testing through each case.

## Breaking Change

- [ ] Yes
- [x] No

### DIFF
--- a/ironfish/src/webApi.ts
+++ b/ironfish/src/webApi.ts
@@ -2,13 +2,9 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-import axios, { AxiosError, AxiosRequestConfig } from 'axios'
+import axios, { AxiosRequestConfig } from 'axios'
 import { FollowChainStreamResponse } from './rpc/routes/chain/followChain'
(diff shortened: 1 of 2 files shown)


ironfish/src/webApi.ts
```diff
@@ -93,23 +89,15 @@ export class WebApi {
...
-
-      return response.data.data
-    } catch (e) {
-      if (IsAxiosError(e) && e.response?.status === 404) {
-        return null
-      }
+    const response = await axios.get<{ data: FaucetTransaction[] }>(
+      `${this.host}/faucet_transactions/next?count=${count}`,
+      this.options(),
+    )
```
dguenther, line 98: Is removing this try/catch going to have ill effects if this `get` throws an Error? I suppose the Docker service should restart it even if it crashes. Just curious if there was a bug leading to this change

deekerno: The linter said it was an unnecessary try-catch wrapper after removing the check for a 404 status code. I'm assuming it doesn't, but I'll see if I can get a better answer.

dguenther (approved):

PR #810, opened 2021-12-21
Only verify spend root in hasValidSpends

NullSoldier:
This is causing add blocks to fail. It's verifying double spends inside of `hasValidSpends` which is called in `verifyConnectedBlock`. We should only verify the spend root after we connect the block, so it's split out here to do that.

The bug is detailed here:
1. Add block
2. Add nullifiers to chain from spends
3. Verify added block
4. Verify nullifiers don't exist in the chain

### DIFF
--- a/ironfish/src/consensus/verifier.ts
+++ b/ironfish/src/consensus/verifier.ts
@@ -170,8 +170,10 @@ export class Verifier {
     tx?: IDatabaseTransaction,
   ): Promise<VerificationResult> {
     return this.chain.db.withTransaction(tx, async (tx) => {
+      const nullifierSize = await this.chain.nullifiers.size(tx)
+
       for (const spend of transaction.spends()) {
-        const reason = await this.verifySpend(spend, tx)
+        const reason = await this.verifySpend(spend, nullifierSize, tx)
 
         if (reason) {
           return { valid: false, reason }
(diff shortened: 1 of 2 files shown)


deekerno (approved):

ironfish/src/consensus/verifier.ts
```diff
@@ -304,7 +307,7 @@ export class Verifier {
           return { valid: false, reason: VerificationResultReason.DOUBLE_SPEND }
         }
 
-        const verificationError = await this.verifySpend(spend, tx)
+        const verificationError = await this.verifySpendRoot(spend, tx)
```
dguenther, line 310: I might not have a handle on every place validation happens, but it looks like this change gets rid of the check for double spends across multiple blocks -- we should still be validating that somewhere.

dguenther (changes requested):

ironfish/src/consensus/verifier.ts
NullSoldier: I need you to elaborate. Here's the information I have. The stack where this line is only used in verifyConnectedBlock, which all these functions are only used in. So here is the call stack.

1. verifyConnectedBlock
2. hasValidSpends
3. verifySpendRoot

hasValidSpends used to call verifySpend which would also do a contains, not now it only does the second part of that function which is to verify the spend root.

dguenther: Yes, the `tree.contains` check is what I'm referring to, I think it's the only place we check if a nullifier already exists in the tree.

I said "across multiple blocks" because `hasValidSpends` has a BufferSet that ensures the same block can't have duplicate spends. However, the `tree.contains` check makes sure the spend hasn't been included in a prior block.

I'm not sure if, with this change, we're still checking if the nullifier already exists in the tree. Is there a test that tries adding a block with the same spend twice and checking that we get a double spend error?

dguenther (approved):

Issue #732, opened 2021-12-13
blake3 shows no hash rate improvement for Zen 2 CPUs


gjfrn:
Since upgrading to 0.1.10, I have not observed any significant hash rate improvement (certainly nothing like 15%) for the following AMD Zen2 CPUs:
Ryzen 5 3600
Ryzen 7 3700X
Ryzen 9 3900X

This is in reference to the following update: Use blake3 instead of blake3-wasm in our miner leading to a 15% increase in hash rate #698

shandongdacongge:
I think so .
the update has no improvement.

gjfrn:
Am pleased to report that this issue is resolved in 1.1.12 due to https://github.com/iron-fish/ironfish/pull/815

NullSoldier:
Ok, this looks like it has been resolved, yes?

gjfrn:
> Ok, this looks like it has been resolved, yes?

Correct. Closing this issue now.

PR #815, opened 2021-12-22
Switch blake3 to native addon

dguenther:
## Summary

The blake3 package doesn't seem to ship a native addon that works properly in node 16, so let's try blake-hash instead.

## Testing Plan

Manually tested by mining on a chain, and existing tests should pass.

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/package.json
+++ b/ironfish/package.json
@@ -7,9 +7,9 @@
   "main": "build/src",
   "license": "MPL-2.0",
   "dependencies": {
+    "@napi-rs/blake-hash": "1.3.1",
     "axios": "0.21.1",
     "bfilter": "1.0.5",
-    "blake3": "2.1.4",
     "blru": "0.1.6",
     "buffer": "6.0.3",
     "buffer-json": "2.0.0",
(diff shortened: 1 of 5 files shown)


sureshvenkata12:
did you notice any changes in the hashrate?

dguenther:
I tested on a mac laptop and it certainly seems quicker. Our hashrate display fluctuates all over the place, and I'm not 100% sure it's working as expected, but on average it did seem to stabilize at a higher number.

NullSoldier:
Tested manually

NullSoldier (approved):

gjfrn:
Am pleased to report that this resolves https://github.com/iron-fish/ironfish/issues/732 with hashrates now at an expected level of improvement, compared with the 0.1.10 change which did not result in any meaningful performance improvement.

sureshvenkata12:
@gjfrn how much improvement have you got?

gjfrn:
> @gjfrn how much improvement have you got?
As per #732, looks to be about 50%+ higher hashrate across 3900X, 3700X and 3600.

Kupuyc:
There are many complains about significant drop of hashrate (40-50%). For me it dropped from 5 mh/s to 3mh/s (lscpu shows AMD EPYC 7282).

PR #850, opened 2022-01-04
Fix facuet using wrong client

NullSoldier:
## Summary

If you try to use the faucet when the node is running, it will fail
silently but say it succeeded. This is because connect offline, then try
to connect using IPC and the error is not handled properly.

## Testing Plan
Request from faucet offline and online

## Breaking Change

```
[ ] Yes
[x] No
```


yuryironfish (approved):
APPROVED

PR #854, opened 2022-01-07
Small performance improvements to accounts store

dguenther:
## Summary

These are two separate changes that should produce very minor performance improvements to the accounts store. I discovered them while testing nulling out incoming block header hashes.

#### Avoid DB call when accounts head equals chain head

Right now, we call `chain.getHeader` with the accounts head hash, then check the result of that against the chain head hash to see if we need to scan new transactions. But since we already have the account hash, we can check it directly against the chain head hash and early-exit before calling `chain.getHeader`.

#### Cache headHash as a Buffer

We're storing headHash as a string in the accounts DB and caching it as a string on `this.headHash`, but on each event loop, we're converting it from a string to a Buffer in order to pass it to `blockchain.ts`. If we change the type of `this.headHash` to Buffer, we only need to convert it when serializing it to the database and when logging the hash.

## Testing Plan

Tested manually by adding an account with blocks on the chain and syncing to ensure the balance still updates.

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/account/accounts.ts
+++ b/ironfish/src/account/accounts.ts
@@ -59,7 +59,7 @@ export class Accounts {
 
   protected rebroadcastAfter: number
   protected defaultAccount: string | null = null
-  protected headHash: string | null = null
+  protected headHash: Buffer | null = null
   protected isStarted = false
   protected isOpen = false
   protected eventLoopTimeout: SetTimeoutToken | null = null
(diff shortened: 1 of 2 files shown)


ironfish/src/account/accounts.ts
```diff
@@ -350,9 +353,10 @@ export class Accounts {
     }
   }
 
-  async updateHeadHash(headHash: string | null): Promise<void> {
+  async updateHeadHash(headHash: Buffer | null): Promise<void> {
     this.headHash = headHash
-    await this.db.setHeadHash(headHash)
+    const hashString = headHash && headHash.toString('hex')
+    await this.db.setHeadHash(hashString)
```
NullSoldier, line 359: note: we should convert this to buffer later too

NullSoldier (approved):

NullSoldier:
Tested manually, and also did full rescan and that works too.

PR #855, opened 2022-01-07
Clear out header hash on incoming blocks

dguenther:
## Summary

Nulls out the hash field on headers of incoming blocks. There's no known bug as a result of this, but we should remove this field in the future because nodes should be calculating it themselves vs. assuming that other nodes have provided a correct hash.

## Testing Plan

Synced a chain from scratch on this PR with an account with transactions

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/network/peerNetwork.test.ts
+++ b/ironfish/src/network/peerNetwork.test.ts
@@ -200,11 +200,13 @@ describe('PeerNetwork', () => {
         message: {
           type: NodeMessageType.NewBlock,
           nonce: 'nonce',
-          payload: { block: Buffer.alloc(0) },
+          payload: { block: { header: { hash: '0' } } },
         },
       })
 
-      expect(peerNetwork['node']['syncer'].addNewBlock).toHaveBeenCalled()
+      expect(peerNetwork['node']['syncer'].addNewBlock).toHaveBeenCalledWith(peer, {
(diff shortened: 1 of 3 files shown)


NullSoldier:
I think this change makes the code more confusing, and doesn't have any actual affect on the code base. We recalculate the hash before verification. I'm willing to merge it in but it does make the code harder for other developers to work with.

dguenther:
The current implementation of blockHeader hashing is confusing and error-prone -- it's difficult to determine where it's necessary to call `recomputeHash` without looking through the whole code path for adding and verifying a block, so it's not easy to determine if this change has an effect on the codebase. For example, we call `recomputeHash` in `verifyTarget`, which gets called in `verifyBlockHeader`, which gets called in `addBlock` and `verifyBlock`. In `addBlock` we also call `recomputeHash`. It's tough to answer questions like whether one of those `recomputeHash` calls is redundant, or whether there any places where we use block header hash before calling `addBlock` or `verifyBlock`.

We don't have any in-depth code for verifying incoming network messages or stripping extra fields, like a `yup` schema. I could go down that route instead, but I do think it's a more substantial change that we'll expect to delete either way when we rework our network messages.

NullSoldier (approved):

NullSoldier (approved):

PR #858, opened 2022-01-08
Add displayName to Account and log when creating mining transaction

dguenther:
## Summary

- Create Account class and add displayName field

Adds a displayName field that includes a hash of the account's private+view keys along with the account name. The `displayName` field doesn't necessarily need to be serialized, but only needs to be generated once, so I created a class for Account that loads in a SerializedAccount object.

- Log account when generating mining reward transaction

Adds a log to the mining director when a mining reward transaction is generated, in order to make it clearer which account is generating the transaction.

## Testing Plan

Tested mining on a fresh chain manually.

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/mining/director.ts
+++ b/ironfish/src/mining/director.ts
@@ -371,7 +371,7 @@ export class MiningDirector {
       this._minerAccount.spendingKey,
     )
 
-    this.logger.info(
+    this.logger.debug(
       `Constructed miner's reward transaction for account ${this._minerAccount.displayName}, block sequence ${newSequence}`,
     )
 
--- a/ironfish/src/mining/director.ts
+++ b/ironfish/src/mining/director.ts
@@ -363,10 +363,16 @@ export class MiningDirector {
       totalTransactionFees += transactionFee
     }
 
+    const newSequence = blockHeader.sequence + 1
+
     const minersFee = await this.strategy.createMinersFee(
       totalTransactionFees,
-      blockHeader.sequence + 1,
+      newSequence,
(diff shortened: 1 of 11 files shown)


ironfish/src/mining/director.ts
```diff
@@ -363,12 +363,18 @@ export class MiningDirector {
...
+    const newSequence = blockHeader.sequence + 1
+
     const minersFee = await this.strategy.createMinersFee(
       totalTransactionFees,
-      blockHeader.sequence + 1,
+      newSequence,
       this._minerAccount.spendingKey,
     )
 
+    this.logger.info(
```
dguenther, line 374: Opinions on whether this log is too aggressive/frequent at `info` level? I think it'd be useful to see if there are any issues with generating a transaction for the wrong account right now, but we could bump it to `debug` later.

NullSoldier: I would advocate for changing this to debug sine the info is for debugging miner problems.

Right now we don't have any constant logs, but this will fill the terminal and make it hard to see what's happening in the node.

dguenther: Changed this -- I'm interested in brainstorming a way we can still save these logs without dumping them to the terminal. It's one of a category of logs where 99% of the time it's not useful, sometimes you retroactively would have wanted to enable it. For example, if someone mined a block on the main chain and is wondering why it's not showing up in their account.

NullSoldier (approved):

NullSoldier:
Tested manually including commands that changed

PR #881, opened 2022-01-12
Fix transaction withReference and use in followChain

dguenther:
## Summary

Added a `transaction.withReference` around the transaction use in `followChain` to avoid some duplicate deserialization calls.

I also noticed that `transaction.withReference` wasn't working correctly for async callbacks -- The callback would execute, then call the `finally` to drop the reference before the callback was finished. Fixed this without making `transaction.withReference` async by creating a separate `Promise.resolve()` to wrap the result and calling `finally` on that one.

## Testing Plan

Synced a fresh database locally and also ran a node that was already synced up. Ran followChain with these changes to make sure it still works

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/primitives/block.ts
+++ b/ironfish/src/primitives/block.ts
@@ -47,13 +47,15 @@ export class Block {
       t.takeReference()
     }
 
-    try {
-      return callback()
-    } finally {
+    const result = callback()
+
+    Promise.resolve(result).finally(() => {
       for (const t of this.transactions) {
         t.returnReference()
       }
-    }
+    })
+
+    return result
   }
 
   /**
--- a/ironfish/src/primitives/transaction.ts
+++ b/ironfish/src/primitives/transaction.ts
@@ -58,11 +58,14 @@ export class Transaction {
    */
   withReference<R>(callback: (transaction: TransactionPosted) => R): R {
     const transaction = this.takeReference()
-    try {
-      return callback(transaction)
-    } finally {
+
+    const result = callback(transaction)
+
+    Promise.resolve(result).finally(() => {
       this.returnReference()
-    }
+    })
+
+    return result
   }
 
   /**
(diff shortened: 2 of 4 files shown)


ironfish/src/primitives/block.ts
```diff
@@ -47,13 +47,15 @@ export class Block {
       t.takeReference()
     }
 
-    try {
-      return callback()
-    } finally {
+    const result = callback()
```
NullSoldier, line 50: Why not just await callback?

ironfish/src/primitives/transaction.ts
```diff
@@ -58,11 +58,14 @@ export class Transaction {
    */
   withReference<R>(callback: (transaction: TransactionPosted) => R): R {
     const transaction = this.takeReference()
-    try {
-      return callback(transaction)
-    } finally {
+
+    const result = callback(transaction)
```
NullSoldier, line 62: Why not just await transaction?

dguenther: > Fixed this without making transaction.withReference async by creating a separate Promise.resolve() to wrap the result and calling finally on that one.

It's functionally possible to use async/await here instead, but it'll be a larger code change because all of the getters on `transaction` use this function, and they'll all have to return e.g. `Promise<number>` instead of `number`, then that'll propagate out to everywhere else in the codebase. What do you think?

ironfish/src/primitives/block.ts
dguenther: This one is easier to change, but see other comment for reasoning

dguenther:
Investigating test failures on this -- it may end up changing the substance of the PR

dguenther:
All right, tests pass now.

ironfish/src/primitives/transaction.ts
NullSoldier: I can buy the reasoning.

NullSoldier (approved):

PR #888, opened 2022-01-12
feat(ironfish): Add heap and rss gauges to `metricsMonitor`

rohanjadvani:
## Summary

Add a `Gauge` metric, and collect RSS / Heap Used metrics every second in the metric monitor.

## Testing Plan

Manually tested.

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/metrics/gauge.ts
+++ b/ironfish/src/metrics/gauge.ts
@@ -2,17 +2,9 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 export class Gauge {
-  private _value: number
+  public value: number
 
   constructor() {
-    this._value = 0
-  }
-
-  get(): number {
-    return this._value
-  }
-
-  set(value: number): void {
-    this._value = value
+    this.value = 0
   }
 }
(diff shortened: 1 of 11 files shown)


ironfish/src/metrics/gauge.ts
```diff
@@ -0,0 +1,18 @@
...
+  constructor() {
+    this._value = 0
+  }
+
+  get(): number {
+    return this._value
+  }
+
+  set(value: number): void {
+    this._value = value
```
NullSoldier, line 16: Can we just get rid of these getters and setters and use public value? It's the same thing with a lot less code. If we need to add these later, it would be easy to do so.

ironfish/src/metrics/metricsMonitor.ts
```diff
@@ -23,6 +19,11 @@ export class MetricsMonitor {
   readonly p2p_OutboundTraffic_WS: Meter
   readonly p2p_OutboundTraffic_WebRTC: Meter
 
+  readonly heapUsed: Gauge
+  readonly rss: Gauge
+  private memoryInterval: ReturnType<typeof setInterval> | undefined
```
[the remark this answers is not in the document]
rohanjadvani: Yeah. It looks like we were using `ReturnType<typeof setInterval>` in a bunch of places, so I went off that. I updated those to use the shared type. Updated this type to use `null`, too.

deekerno (approved):

ironfish/src/metrics/meter.ts
```diff
@@ -22,7 +23,7 @@ export class Meter {
   private _rate5m: RollingAverage
   private _average: RollingAverage
   private _count = 0
-  private _interval: ReturnType<typeof setInterval> | null = null
+  private _interval: SetTimeoutToken | null = null
```
NullSoldier, line 26: Hey, is this compatible with intervals? Should we make a separate SetIntervalToken that is the return type of SetInterval?

NullSoldier (approved):

PR #890, opened 2022-01-13
feat(cli,ironfish): Add memory usage to status command

rohanjadvani:
## Summary

Log out node heap usage and rss with status command.

This PR is opened against #888 until that code change is merged.

## Testing Plan

Added unit test.

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish-cli/src/commands/status.ts
+++ b/ironfish-cli/src/commands/status.ts
@@ -8,8 +8,6 @@ import { Assert } from 'ironfish'
 import { IronfishCommand } from '../command'
 import { RemoteFlags } from '../flags'
 
-const BYTES_PER_MB = 1024 * 1024
-
 export default class Status extends IronfishCommand {
   static description = 'Show the status of the node'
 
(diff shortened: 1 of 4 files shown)


ironfish-cli/src/commands/status.ts
```diff
@@ -8,6 +8,8 @@ import { Assert } from 'ironfish'
 import { IronfishCommand } from '../command'
 import { RemoteFlags } from '../flags'
 
+const BYTES_PER_MB = 1024 * 1024
```
NullSoldier, line 11: Don't use this use https://github.com/iron-fish/ironfish/blob/master/ironfish/src/utils/file.ts#L27

ironfish/src/rpc/routes/node/getStatus.ts
```diff
@@ -18,6 +18,8 @@ export type GetStatusResponse = {
     status: 'started' | 'stopped' | 'error'
     version: string
     git: string
+    heapUsed: number
+    rss: number
   }
```
NullSoldier, line 23: I'm not 100% sure but I kind of want this to go into a memory section since this is for the process and not the node.

ironfish-cli/src/commands/status.ts
rohanjadvani: Ah didn't realize this was a util. Updated.

ironfish/src/rpc/routes/node/getStatus.ts
rohanjadvani: That makes sense. I moved the numbers.

NullSoldier (approved):

PR #894, opened 2022-01-14
Make batch size configurable for miner

kernel32uiuc:
## Summary
Playing with batchSize can impact the hash rate. A faster processor can benefit from a higher batchSize. Making it configurable can give more flexibility to miners. 

## Testing Plan
Manually tested via 
```
./ironfish config:set minerBatchSize <newBatchSize> 
./ironfish config:show
```
When I set a very low batch size, I noticed that the hash rate significantly slows down. 

## Breaking Change

```
[ ] Yes
[ X] No
```

### DIFF
--- a/ironfish-cli/src/commands/miners/start.ts
+++ b/ironfish-cli/src/commands/miners/start.ts
@@ -39,7 +39,8 @@ export class Miner extends IronfishCommand {
     }
 
     const client = this.sdk.client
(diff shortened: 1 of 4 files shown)


dguenther:
Nice addition! I'll do some testing on this then review/merge 🚢

NullSoldier (approved):

NullSoldier:
Tested locally, this works.

PR #908, opened 2022-01-18
Add TCP Secure flag to allow secure config over TCP

mat-if:
## Summary

Adds a `rpc.tcp.secure` flag for the node, which enables the ability to change sensitive config via tcp. Disabled by default.

## Breaking Change

```
[ ] Yes
[x] No
```


NullSoldier (approved):
This is a great PR, thank you.

PR #916, opened 2022-01-19
Move mining logic from NodeJS to Rust

mat-if:
## Summary

Moves the actual logic of hashing and comparing into Rust for some straight-forward performance improvement. Simple napkin math comparison of my computer shows a ~10-15x speedup.

## Testing Plan

Ran side-by-side with the original NodeJS implementation using 2 separate nodes

## Breaking Change

```
[ ] Yes
[x] No
```


ironfish-rust-nodejs/src/lib.rs
```diff
@@ -65,6 +84,31 @@ fn generate_new_public_address(mut cx: FunctionContext) -> JsResult<JsObject> {
...
+    });
+    // Argument 4
+    let batch_size = cx.argument::<JsNumber>(3)?.value(&mut cx) as i64;
+
+    // Execute batch mine operation
+    let mine_header_result = mining::mine_header_batch(
+        header_bytes, initial_randomness, target, batch_size
+    );
+
+    // Return result
```
mat-if, line 108: Added these because I found this code unnecessarily hard to parse at a glance for what it does, but I'm not going to fight for it

ironfish/src/mining/miner.test.ts
```diff
@@ -122,14 +122,14 @@ describe('Miner', () => {
 
 describe('mineHeader', () => {
   beforeEach(() => {
-    mocked(blockHeaderModule.hashBlockHeader).mockReset()
+    mocked(nativeModule.mineHeaderBatch).mockReset()
```
mat-if, line 125: Git makes this diff pretty gross, I just re-wrote the block entirely, so might be easier to review without a diff-view

ironfish-rust-nodejs/src/lib.rs
```diff
@@ -35,6 +40,20 @@ impl Key {
     }
 }
 
+impl ToObjectExt for mining::MineHeaderResult {
```
NullSoldier, line 43: Is there some way to say that Key implements this trait as well, since it matches?

mat-if: Good call, totally glossed over that. Added to Key

dguenther (approved):
Nice, looks good!

PR #917, opened 2022-01-19
Allow version to be driven by lib consumer

NullSoldier:
## Summary

This makes a change so that previously, the version sent out everywhere
and the agent string was driven by the library. This doesn't tell the
full story, because the real version is the person who is consuming the
library and potentially writing custom code on top of the
library.

This lets users who consume the library override the package and produce
their own agent and version library using the new Package type that you
can produce from a Package.json file and pass down into the node.

https://linear.app/ironfish/issue/IRO-1392/fix-node-versioning

```js
// old
Peer Agent    if/cli/36c71af
// new
Peer Agent    ironfish-cli/0.1.19/36c71af
```

## Testing Plan

Test using brew beta and locally

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish-cli/scripts/build.sh
+++ b/ironfish-cli/scripts/build.sh
@@ -18,8 +18,13 @@ if ! command -v git &> /dev/null; then
     exit 1
 fi
 
-echo "Inserting GIT hash into ironfish/package.json as gitHash"
+
 GIT_HASH=$(git rev-parse --short HEAD)
+
+echo "Inserting GIT hash into ironfish-cli/package.json as gitHash"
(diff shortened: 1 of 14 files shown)


ironfish-cli/scripts/build.sh
```diff
@@ -20,6 +20,7 @@ fi
 
 echo "Inserting GIT hash into ironfish/package.json as gitHash"
```
dguenther, line 21: Echo text should probably be updated here

dguenther (approved):

PR #920, opened 2022-01-20
Add debug command with accounts head info

mat-if:
## Summary

Create a `debug` command for the CLI which will house data that will be helpful in debugging issues. For now, just has accounts head information, and a few bits about the OS and tool versions.

Resolves IRO-1484

## Testing Plan

`yarn start:once debug`
`yarn start:once debug --datadir ~/.null`

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish-cli/src/commands/debug.ts
+++ b/ironfish-cli/src/commands/debug.ts
@@ -34,8 +34,8 @@ export default class Debug extends IronfishCommand {
     const telemetryEnabled = this.sdk.config.get('enableTelemetry').toString()
 
     this.log(`
-Ironfish version        ${Package.version} @ ${Package.git}
-Operating System        ${os.type()} ${process.arch}
+Iron Fish version       ${Package.version} @ ${Package.git}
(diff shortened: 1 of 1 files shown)


deekerno (commented):
Should the log output be more consistent with capitalization and branding? For example, "Ironfish version" should be "Iron Fish Version" if both words in "Operating System" are capitalized.

deekerno (approved):

NullSoldier:
Works for me

```
Iron Fish version       0.1.19 @ src
Iron Fish library       0.0.0 @ src
Operating system        Darwin x64
CPU model               Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
CPU threads             16
RAM total               64.00 GiB
Node version            v16.9.0
Telemetry enabled       true
Accounts head hash      00000000007c0153d832c4fe82bd7c794f284ecf10eb958e36594645b9aa6b67
Accounts head in chain  true
Accounts head sequence  76870
```

NullSoldier (approved):

PR #921, opened 2022-01-20
Format miner hash rate using metric system

NullSoldier:
## Summary

This PR wil now format the hash rate in metric increments similar to
file sizes while still printing the original value. It also adds tests
for the FileUtils, and fixes a few bugs I saw with rounding that
produced odd values.

```javascript
// old
Mining block 76858 on request 32... ⣷ 2276451 H/s
// new
Mining block 76858 on request 32... ⣷ 2.27 MH/s (2276451)
```

## Testing Plan

Added tests and tested miner locally

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/utils/file.ts
+++ b/ironfish/src/utils/file.ts
@@ -1,31 +1,55 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-type SizeSuffix = { B: string; KB: string; MB: string; GB: string }
+import { MathUtils } from './math'
 
-const fileSizeSuffix: SizeSuffix = { B: 'B', KB: 'KB', MB: 'MB', GB: 'GB' }
-const memorySizeSuffix: SizeSuffix = { B: 'B', KB: 'KiB', MB: 'MiB', GB: 'GiB' }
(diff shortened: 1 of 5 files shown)


ironfish/src/utils/file.ts
```diff
@@ -1,31 +1,55 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-type SizeSuffix = { B: string; KB: string; MB: string; GB: string }
+import { MathUtils } from './math'
 
-const fileSizeSuffix: SizeSuffix = { B: 'B', KB: 'KB', MB: 'MB', GB: 'GB' }
-const memorySizeSuffix: SizeSuffix = { B: 'B', KB: 'KiB', MB: 'MiB', GB: 'GiB' }
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
deekerno, line 6: This was probably unintentional.

deekerno (approved):
Looks good to me; left one comment.

ironfish/src/utils/file.ts
NullSoldier: Fixed

dguenther (approved):
Looks good! Seems like it doesn't touch anything except CLI display so should be safe

PR #927, opened 2022-01-23
Add GetUnspentNotes worker pool task

dguenther:
## Summary

Adds a worker pool task for getUnspentNotes, which is used when calculating an account's balance. Note that this does not make the balance calculation faster, just frees up the main thread when doing balance work.

I'm not sure if this is the right level of abstraction for worker pool tasks -- this'll probably have to be tweaked as more functionality in the accounts store is moved to the worker pool.

## Testing Plan

Balance calculation should work as before.

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/account/accounts.test.slow.ts
+++ b/ironfish/src/account/accounts.test.slow.ts
@@ -41,7 +41,7 @@ describe('Accounts', () => {
     await node.accounts.updateHead()
 
     // Initial balance should be 0
-    expect(node.accounts.getBalance(account)).toEqual({
+    await expect(node.accounts.getBalance(account)).resolves.toEqual({
       confirmed: BigInt(0),
       unconfirmed: BigInt(0),
     })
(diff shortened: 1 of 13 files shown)


mat-if (approved):
My first reaction was that  it might make more sense to have the loop over the returning `result.notes` in accounts.ts (https://github.com/iron-fish/ironfish/pull/927/files#diff-975265306b3138d4067a44c1f74a98ef80249217a05ac91dff96236b482c429bR608) in the workerPool's `getUnspentNotes` but that would require some fairly heavy refactoring to support the error cases which feels pretty far beyond the scope of the task, so I'm in favor of what you've done here

NullSoldier (approved):

NullSoldier:
Tested manually
 - Sent money between 2 accounts and checked balance on both
 - Checked balance on a large account, 2 smaller

PR #932, opened 2022-01-24
Avoid making BigUInts when comparing hashes

terite:
## Summary
When comparing hashes, comparing two `[u8; 32]` seems more efficient than
making a `BigUInt` to compare against target. On my laptop this improves
single-core hash rate from ~2.1 MH/s to ~4.1 MH/s.

This is my first time working in rust. I would greatly appreciate any critique, big or small, functional or aesthetic.

## Testing Plan
Added tests. Miner runs. Would like to confirm blocks are still found organically, but that's pretty time consuming.

## Breaking Change
No.

### DIFF
--- a/ironfish-rust/src/mining.rs
+++ b/ironfish-rust/src/mining.rs
@@ -42,12 +42,14 @@ pub fn mine_header_batch(
         found_match: false,
     };
 
+    let target_bytes = biguint_to_bytes(&target);
+
     for i in 0..batch_size {
         let randomness = randomize_header(initial_randomness, i, header_bytes);
         let hash = blake3::hash(&header_bytes);
-        let new_target = BigUint::from_bytes_be(hash.as_bytes());
+        let new_target_bytes = hash.as_bytes();
 
-        if new_target <= target {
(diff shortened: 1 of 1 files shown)


mat-if:
You can probably just remove the BigUInt stuff entirely, including the dependency altogether. It comes in as a slice from Neon here https://github.com/iron-fish/ironfish/blob/master/ironfish-rust-nodejs/src/lib.rs#L95 then converted to BigUInt then passed into this fn

dguenther:
I'm okay with merging this and doing the follow-up in a separate PR since it's not introducing any bugs, but it would be great to fix the CI warnings before merging, I think those are pretty reasonable to fix up.

terite:
@dguenther Lint changes fixed.

@mat-if I agree that seems like a good change, but doesn't seem likely to change performance and goes beyond the scope of this PR. Mind if I leave that for someone else?

mat-if:
Yeah, sorry, I should have been more clear, definitely fine leaving it for later

mat-if (approved):

PR #933, opened 2022-01-24
Add a github action to push version to API

mat-if:
## Summary

Push the CLI version to an API endpoint so we can surface it on the explorer, etc.

Keeping in draft until this is tested and moved into the deploy job, instead of being a separate job. Separate job is just for testing purposes. Also waiting on the API side PR to merge.

Closes IRO-1486

## Testing Plan

Run it manually, confirm in API: https://github.com/iron-fish/ironfish/runs/4940958759?check_suite_focus=true ✅ 

## Breaking Change

```
[ ] Yes
[x] No
```


dguenther:
Just offhand, this looks good to me if it runs/works! Feel free to ping me or someone for a follow-up review when you're ready

rohanjadvani (approved):

.github/workflows/deploy-node-aws.yml
```diff
@@ -30,3 +30,9 @@ jobs:
         env:
           REGISTRY_URL: ${{ secrets.AWS_NODE_REGISTRY_URL }}
           PACKAGE_NAME: ironfish
+
+      - name: Push version string to API
+        run: ./ironfish-cli/scripts/push-version.sh
+        env:
+          IRON_FISH_API_KEY: ${{ secrets.IRON_FISH_API_KEY }}
+          IRON_FISH_API_URL: ${{ secrets.IRON_FISH_API_URL }}
```
dguenther, line 38: I think we might not want it as part of this workflow, because sometimes we'll deploy test Docker images to AWS just for our API to use that we aren't ready to release publicly. The `deploy-node-github` one could be okay though -- the GitHub docker registry is only used by users as far as I know.

I don't mind having it as a separate workflow step to add to our deploy steps for now either. Another idea could be to have it run when we create a new release in GitHub -- not sure if that's a good idea or not

dguenther (changes requested):

dguenther (approved):

Issue #861, opened 2022-01-08
'yarn test' failing in windows command prompt


AmberKiso:
IronFish Version 0.1.15
Windows 10 Home

Copied from the discord chat to claim some points since I'm not going to be able compete in the mining arena on a consumer laptop. ;)
https://discord.com/channels/771503434028941353/828767731305676810/929192909469810688

> I'm getting the following error message when I've run 'yarn test' in a windows command prompt:
> FAIL ironfish src/workerPool/pool.test.ts (7.581 s)
>   ● Worker Pool › executes in worker
> 
>     Cannot find module 'C:\Users\mjolnir\Documents\GitHub\ironfish\ironfish\src\workerPool\worker.js'
> 
> Did I mis-install something? GitHub doesn't have that file in the workerPool directory...

Appears to be fixed with commit 859, but ironfish-cli tests then fail.

deekerno:
Thanks for letting us know! This issue is eligible for testnet credit, if you like!

PR #934, opened 2022-01-25
Add timeouts to select ironfish-cli tests

AmberKiso:
## Summary
After #861 was resolved by #859, the ironfish-cli tests would then fail. It seemed odd that something used consistently for all pre-commit validation would fail in Windows.

I took a guess that increasing the timeout would help. I couldn't find where jest.setTimeout.Error is set, so I added timeouts (another guess that these are the correct/"best" places) and things passed. I'm also guessing that the added setTimeout() calls have local scope and don't affect other tests.

I would appreciate some feedback on this fix. I **hate** fixing apparent timing issues by increasing values.

## Testing Plan

Ran `yarn test` a few times with changes. All passed.
Total execution time (From the `Done in Xs` completion message) was
62, 47, 51, 55, 50 seconds.
Surprising amount of variability with the results. (I'm running a i5-10210U @ 1.6Ghz.)

There is probably some risk that people running on slower processors or with simultaneous heavy workloads will exceed these timeouts.

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish-cli/src/commands/accounts/export.test.ts
+++ b/ironfish-cli/src/commands/accounts/export.test.ts
@@ -59,6 +59,7 @@ describe('accounts:export', () => {
   })
 
   describe('with no flags', () => {
+    jest.setTimeout(10000)
     test
       .stdout()
       .command(['accounts:export', 'default'])
--- a/ironfish-cli/src/commands/accounts/rescan.test.ts
+++ b/ironfish-cli/src/commands/accounts/rescan.test.ts
@@ -39,6 +39,7 @@ describe('accounts:rescan', () => {
   })
 
   describe('with no flags', () => {
+    jest.setTimeout(10000)
     test
       .stdout()
       .command(['accounts:rescan'])
--- a/ironfish-cli/src/commands/status.test.ts
+++ b/ironfish-cli/src/commands/status.test.ts
@@ -63,6 +63,7 @@ describe('status', () => {
   })
 
   describe('it logs out the status of the node', () => {
+    jest.setTimeout(10000)
     test
       .stdout()
       .command(['status'])


dguenther:
Thanks for putting this up in draft state for discussion! I'd rather figure out if there's an underlying issue with these tests on Windows vs adding timeouts to get them to pass. I haven't really worked with the CLI tests in depth -- they've been around for a while, and we've been maintaining them more than contributing to them.

I think we need to do some investigation around if there are specific issues with the library these tests are written in on Windows. The style of them is a little different than the Jest tests in the `ironfish` project.

AmberKiso:
Sure thing. Let me know how I can help.

dguenther:
Okay, here's what I'm suspecting is happening -- I don't think the CLI tests are using temporary directories for storage the way we do in the `ironfish` tests. So any tests that happen to involve writing to a database and don't have the storage layer mocked will operate out of the `~/.ironfish` folder by default, and if a lot of tests are running at the same time, they'll block each other waiting for the DB lock to be released.

(I actually noticed that the `start` CLI test doesn't have the network properly mocked, so it actually seems like it'll boot up a connection to the live testnet network if the test doesn't exit quickly...)

If you run `yarn test -w 2` in the `ironfish-cli` folder, for example, the timeouts aren't needed to make the tests pass (on my machine at least).

It would be nice to run the test suite on Windows as well as the default, Linux, so if you're interested in adding support for that to our GitHub Actions, feel free to take a try at it. I think I'd be okay with adding timeouts to the tests for now if we had a Windows GitHub Action set up, and the timeouts made the tests consistently pass on that. Then if you wanted to do more work to fix the issue with the database contention, you'd be more than welcome to!

dguenther:
This is probably going to be a fairly challenging task even for someone pretty familiar with the project, so no worries!

The `.github/workflows` folder has all of our GitHub actions in it, which show up in the `Actions` tab of this GitHub repository. It might take some work to figure out how to test/run them on a fork... I haven't done that enough to be sure how to do it right. 

Limiting the tests to just run one-at-a-time is an option, but I still don't like that the tests are actually booting up live network connections. And I agree that writing mocks for the storage layer and network layer is probably a little tough, we've already seen issues where if we forget to maintain them, the tests end up behaving in strange ways.

Along the lines of what the \ironfish project does, it mostly uses `os.tmpdir()` combined with a uuid as a data directory for the node. This comes from the createSetup function in this file: https://github.com/iron-fish/ironfish/blob/1f3dc311c6db990313f81ae7d8046d19e8b2ce93/ironfish/src/testUtilities/nodeTest.ts#L77

I think two of the best options would be either:
* Change every command in the CLI tests to include `--datadir <tmpdir>` (and also `--bootstrap ""` in the case of the start command). Not great that it requires you to remember to do that in every test when writing new tests.
* Replace or mock the `IronfishSdk.init` function to always pass `<tmpdir>` as a data directory (and `--bootstrap ""` for start commands). Avoids the need to override it in every function, but is probably slightly more challenging to implement right.

Also, I'd consider this a separate piece of work to running the Windows tests in GitHub Actions, now that we know that the issues with the tests on Windows aren't necessarily Windows-specific.

Feel free to close the PR if you're not interested in picking this up anytime soon -- just wanted to leave some info here if it'd be useful!

Issue #880, opened 2022-01-11
Node exit with error "Failed to iterate between blocks on diverging forks:"


vht210:
Update to 0.1.18
- Rung node for a while then it stopped with the error below (Reproduced with multiple  node/miners)

git/ironfish/ironfish/build/src/blockchain/blockchain.js:286
            throw new Error('Failed to iterate between blocks on diverging forks:' +
                  ^

Error: Failed to iterate between blocks on diverging forks: curr: 00000...24849, end: 00000...dd4cb, progress: 63094/63093
    at Blockchain.iterateToHashes (/home/tung/git/ironfish/ironfish/build/src/blockchain/blockchain.js:286:19)
    at iterateToHashes.next (<anonymous>)
    at Blockchain.iterateTo (/home/tung/git/ironfish/ironfish/build/src/blockchain/blockchain.js:265:26)
    at iterateTo.next (<anonymous>)
    at Blockchain.iterateAllTransactions (/home/tung/git/ironfish/ironfish/build/src/blockchain/blockchain.js:733:26)
    at async Accounts.scanTransactions (/home/tung/git/ironfish/ironfish/build/src/account/accounts.js:393:80)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

NullSoldier:
This issue is caused by two things
 * If you start a rescan using `ironfish accounts:rescan` when your wallet is on a fork
 * If you import an account using `ironfish accounts:import` when your wallet is on a fork

A temporary workaround is to clear your wallet stater using `ironfish accounts:rescan —reset`

NullSoldier:
This issue is happening because of how we handle rescans. 

1. When a rescan occurs, the wallet will never update itself until the rescan completes.
2. When the node starts and a rescan has started it will always try to rescan first

This means that because the rescan is broken, the wallet is stuck in a broken state permanently. Let's now dive into why the rescan is broken.


```
G -> A1 -> A2
        -> B2 -> B3

Accounts Head: A2
Chain Head:    B3
```

[...]

PR #937, opened 2022-01-26
Fix rescan when account head is on a fork

NullSoldier:
## Summary

This fixes the issue outlined in github issue [#880](https://github.com/iron-fish/ironfish/issues/880).
This is happening because of how we handle rescans.

1. When a rescan occurs, the wallet will never update itself until the rescan completes.
2. When the node starts and a rescan has started it will always try to rescan first

This means that because the rescan is broken, the wallet is stuck in a broken
state permanently. Let's now dive into why the rescan is broken.

```
G -> A1 -> A2
        -> B2 -> B3

Accounts Head: A2
Chain Head:    B3
```

When we do a rescan, we try to iterate from the genesis block (G) to the
accounts head (A2), but the block chain has re-orged to B3, the newer heaviest
chain. If the accounts head were to update, it would reorganize as well to B3.
Unfortunately, once a scan has started the accounts will wait for the rescan to
finish before updating the accounts head from A2 to B3.

Now the way we try to iterate from G -> A2 is that we use the heaviest chain
iteration. That means we're going to iterate in this order: `G -> A1 -> B2`
before realizing were not on target and fail with the error that's in the
ticket. It's now unrecoverable because there's no way out of this since accounts
is stuck at A2.

The fixes I chose were inspired by Bitcoin, which chose not to solve the problem
exactly. The worst case outcome are that if you are importing an account with a
transction on a fork, you'll potentially think that transaction is added to a
block on a fork, even if it never gets added to a block on the main chain.

The fixes are two fold:
1. Don't error when the account head is unreachable
2. Don't let iterateTo jump chains if we use iteratetoHashes() during a reorg

## Testing Plan
 - [x]  Write tests
 - [x]  Import accounts
 - [x]  Do rescans

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/blockchain/blockchain.test.ts
+++ b/ironfish/src/blockchain/blockchain.test.ts
@@ -164,7 +164,7 @@ describe('Blockchain', () => {
     expect(blocks[0].hash.equals(genesis.hash)).toBe(true)
   })
 
-  it('should not iterate and jump chains', async () => {
+  it('should not iterate and jump chains and throw error', async () => {
     const { strategy, chain } = nodeTest
     strategy.disableMiningReward()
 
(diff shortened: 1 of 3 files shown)


deekerno (approved):

ironfish/src/blockchain/blockchain.test.ts
```diff
@@ -164,6 +164,55 @@ describe('Blockchain', () => {
...
+    const blockB4 = await makeBlockAfter(chain, blockB3)
+
+    await expect(chain).toAddBlock(blockA1)
+    await expect(chain).toAddBlock(blockA2)
+    await expect(chain).toAddBlock(blockA3)
+
+    expect(chain.head.hash.equals(blockA3.header.hash)).toBe(true)
+    expect(chain.latest.hash.equals(blockA3.header.hash)).toBe(true)
+
+    const iter = chain.iterateTo(genesis, blockA2.header)
```
dguenther, line 195: I think this should iterate to A3, or else I'd expect the iterator will always be done when yielding A2 regardless of whether a fork was added.

dguenther (approved):
Looks good with the aforementioned change!

ironfish/src/blockchain/blockchain.test.ts
NullSoldier: Fixed by making tests more robust

PR #938, opened 2022-01-26
Peer Persistence

deekerno:
## Summary
This PR implements peer persistence by saving peer addresses to `hosts.json` on a defined interval and at shutdown; connected peers are saved to a `priorPeers` list and possible, un-connected peers are saved to a `possiblePeers` list. If there is persisted peer information at startup, the node will prioritize trying to connect to peers to which it was connected in the past before using the possible peers list.

This also decreases the amount of `peerListRequest` spam by changing the interval to 60 seconds and sending a request upon peer connection.

Note that there is room for optimization in the related code, e.g. capping the amount of `possiblePeers`.
## Testing Plan
Added tests. Ran the node and it successfully connected to the main network with no harmful effects. Starting the node with persisted host information led to a significant speedup in connecting to the network.

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/network/peerNetwork.ts
+++ b/ironfish/src/network/peerNetwork.ts
@@ -7,8 +7,8 @@ import { Assert } from '../assert'
 import { Blockchain } from '../blockchain'
 import { MAX_REQUESTED_BLOCKS } from '../consensus'
 import { Event } from '../event'
-import { FileSystem } from '../fileSystems'
 import { DEFAULT_WEBSOCKET_PORT } from '../fileStores/config'
+import { FileSystem } from '../fileSystems'
 import { createRootLogger, Logger } from '../logger'
 import { MetricsMonitor } from '../metrics'
 import { IronfishNode } from '../node'
@@ -292,7 +292,7 @@ export class PeerNetwork {
     })
   }
 
-  start(): void {
+  async start(): Promise<void> {
     if (this.started) {
       return
     }
@@ -369,7 +369,7 @@ export class PeerNetwork {
     }
 
     // Start up the PeerManager
-    this.peerManager.start()
+    await this.peerManager.start()
 
     // Start up the PeerConnectionManager
     this.peerConnectionManager.start()
@@ -397,12 +397,14 @@ export class PeerNetwork {
    * Call close when shutting down the PeerNetwork to clean up
    * outstanding connections.
    */
-  stop(): void {
-    this.started = false
-    this.peerConnectionManager.stop()
-    this.peerManager.stop()
-    this.webSocketServer?.close()
-    this.updateIsReady()
+  async stop(): Promise<void> {
+    await Promise.allSettled([
+      (this.started = false),
+      this.peerConnectionManager.stop(),
+      await this.peerManager.stop(),
+      this.webSocketServer?.close(),
+      this.updateIsReady(),
+    ])
   }
 
   /**
--- a/ironfish/src/network/peerNetwork.ts
+++ b/ironfish/src/network/peerNetwork.ts
@@ -7,6 +7,7 @@ import { Assert } from '../assert'
 import { Blockchain } from '../blockchain'
 import { MAX_REQUESTED_BLOCKS } from '../consensus'
 import { Event } from '../event'
+import { FileSystem } from '../fileSystems'
 import { DEFAULT_WEBSOCKET_PORT } from '../fileStores/config'
 import { createRootLogger, Logger } from '../logger'
 import { MetricsMonitor } from '../metrics'
@@ -58,7 +59,6 @@ import {
   NodeMessageType,
   PayloadType,
 } from './messages'
-import { AddressManager } from './peers/addressManager'
 import { LocalPeer } from './peers/localPeer'
 import { BAN_SCORE, Peer } from './peers/peer'
 import { PeerConnectionManager } from './peers/peerConnectionManager'
@@ -147,7 +147,8 @@ export class PeerNetwork {
     node: IronfishNode
     strategy: Strategy
     chain: Blockchain
-    addressManager: AddressManager
+    files: FileSystem
+    dataDir?: string
   }) {
     const identity = options.identity || tweetnacl.box.keyPair()
     const enableSyncing = options.enableSyncing ?? true
@@ -178,7 +179,7 @@ export class PeerNetwork {
 
     this.peerManager = new PeerManager(
       this.localPeer,
-      options.addressManager,
+      options.files,
       this.logger,
       this.metrics,
       maxPeers,
--- a/ironfish/src/network/peerNetwork.ts
+++ b/ironfish/src/network/peerNetwork.ts
@@ -3,12 +3,12 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 import tweetnacl from 'tweetnacl'
+import { HostsStore } from '..'
 import { Assert } from '../assert'
 import { Blockchain } from '../blockchain'
 import { MAX_REQUESTED_BLOCKS } from '../consensus'
 import { Event } from '../event'
 import { DEFAULT_WEBSOCKET_PORT } from '../fileStores/config'
-import { FileSystem } from '../fileSystems'
 import { createRootLogger, Logger } from '../logger'
 import { MetricsMonitor } from '../metrics'
 import { IronfishNode } from '../node'
@@ -147,8 +147,7 @@ export class PeerNetwork {
     node: IronfishNode
     strategy: Strategy
     chain: Blockchain
-    files: FileSystem
-    dataDir?: string
+    hostsStore: HostsStore
   }) {
     const identity = options.identity || tweetnacl.box.keyPair()
     const enableSyncing = options.enableSyncing ?? true
@@ -179,13 +178,12 @@ export class PeerNetwork {
 
     this.peerManager = new PeerManager(
       this.localPeer,
-      options.files,
+      options.hostsStore,
       this.logger,
       this.metrics,
       maxPeers,
       targetPeers,
       logPeerMessages,
-      options.dataDir,
     )
     this.peerManager.onMessage.on((peer, message) => this.handleMessage(peer, message))
     this.peerManager.onConnectedPeersChanged.on(() => this.updateIsReady())
--- a/ironfish/src/network/peerNetwork.ts
+++ b/ironfish/src/network/peerNetwork.ts
@@ -293,7 +293,7 @@ export class PeerNetwork {
     })
   }
 
-  async start(): Promise<void> {
+  start(): void {
     if (this.started) {
       return
     }
@@ -370,7 +370,7 @@ export class PeerNetwork {
     }
 
     // Start up the PeerManager
-    await this.peerManager.start()
+    this.peerManager.start()
 
     // Start up the PeerConnectionManager
     this.peerConnectionManager.start()
@@ -399,13 +399,11 @@ export class PeerNetwork {
    * outstanding connections.
    */
   async stop(): Promise<void> {
-    await Promise.allSettled([
-      (this.started = false),
-      this.peerConnectionManager.stop(),
-      await this.peerManager.stop(),
-      this.webSocketServer?.close(),
-      this.updateIsReady(),
-    ])
+    this.started = false
+    this.peerConnectionManager.stop()
+    await this.peerManager.stop()
+    this.webSocketServer?.close()
+    this.updateIsReady()
   }
 
   /**
--- a/ironfish/src/network/peerNetwork.ts
+++ b/ironfish/src/network/peerNetwork.ts
@@ -185,6 +185,7 @@ export class PeerNetwork {
       maxPeers,
       targetPeers,
       logPeerMessages,
+      options.dataDir,
     )
     this.peerManager.onMessage.on((peer, message) => this.handleMessage(peer, message))
     this.peerManager.onConnectedPeersChanged.on(() => this.updateIsReady())
--- a/ironfish/src/network/peers/addressManager.ts
+++ b/ironfish/src/network/peers/addressManager.ts
@@ -1,10 +1,10 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { Peer, PeerList } from '..'
 import { HostsStore } from '../../fileStores'
 import { FileSystem } from '../../fileSystems'
 import { ArrayUtils } from '../../utils'
+import { Peer, PeerList } from '..'
 import { ConnectionDirection, ConnectionType } from './connections'
 import { PeerAddress } from './peerAddress'
 
--- a/ironfish/src/network/peers/addressManager.ts
+++ b/ironfish/src/network/peers/addressManager.ts
@@ -1,8 +1,11 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Peer, PeerList } from '..'
 import { HostsStore } from '../../fileStores'
 import { FileSystem } from '../../fileSystems'
+import { ArrayUtils } from '../../utils'
+import { ConnectionDirection, ConnectionType } from './connections'
 import { PeerAddress } from './peerAddress'
 
 /**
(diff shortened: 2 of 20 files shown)


ironfish/src/node.ts
```diff
@@ -303,7 +304,7 @@ export class IronfishNode {
     await Promise.allSettled([
       this.accounts.stop(),
       this.syncer.stop(),
-      this.peerNetwork.stop(),
+      await this.peerNetwork.stop(),
```
dguenther, line 307: This doesn't need the await, it's handled by the wrapping `Promise.allSettled`

ironfish/src/sdk.ts
```diff
@@ -37,6 +37,7 @@ export class IronfishSdk {
   logger: Logger
   metrics: MetricsMonitor
   internal: InternalStore
+  hosts: HostsStore
```
dguenther, line 40: Do the node + sdk still need to construct the HostsStore? It looks like it's constructed at a lower level in the AddressManager now.

ironfish/src/network/peers/peerManager.ts
```diff
@@ -782,55 +793,54 @@ export class PeerManager {
     }
   }
 
-  start(): void {
-    this.distributePeerListHandle = setInterval(() => this.distributePeerList(), 5000)
-    this.disposePeersHandle = setInterval(() => this.disposePeers(), 2000)
+  async start(): Promise<void> {
+    await Promise.allSettled([
```
dguenther, line 797: This `Promise.allSettled` isn't needed, none of these functions involve promises.

ironfish/src/network/peers/peerManager.ts
```diff
@@ -782,55 +793,54 @@ export class PeerManager {
...
+      (this.requestPeerListHandle = setInterval(() => this.requestPeerList(), 60000)),
+      (this.disposePeersHandle = setInterval(() => this.disposePeers(), 2000)),
+      // eslint-disable-next-line @typescript-eslint/no-misused-promises
+      (this.savePeerAddressesHandle = setInterval(async () => {
+        await this.addressManager.save(this.peers)
```
dguenther, line 802: You can drop the async function and just do `void this.addressManager.save(this.peers)`. This will allow you to remove the disabled lint line as well.

ironfish/src/network/testUtilities/helpers.ts
```diff
@@ -94,6 +94,25 @@ export function getConnectedPeer(
...
+): { peer: Peer; connection: WebSocketConnection } {
+  const { peer, connection } = getConnectingPeer(pm)
+
+  if (!identity) {
+    identity = jest.requireActual<typeof import('uuid')>('uuid').v4()
+  }
+
+  if (!isIdentity(identity)) {
+    identity = mockIdentity(identity)
+  }
```
dguenther, line 109: The contents of this function can be replaced with `pm.getOrCreatePeer(null)`, since a new peer is disconnected by default. Likewise, it doesn't really make sense to return a connection from this function, because a Peer doesn't have any connections on itself when it's in the disconnected state.

ironfish/src/network/peers/peerManager.test.ts
```diff
@@ -558,13 +450,13 @@ describe('PeerManager', () => {
           sdpMid: '0',
         },
       })
-      expect(sendSpy).toBeCalledTimes(1)
+      expect(sendSpy).toBeCalledTimes(2)
```
dguenther, line 453: Marking this to check for later -- The peer was already in the connected state here, so what's triggering more than one send? Regardless, the test should be cleaned up a bit so it's just ensuring that the webRTC signaling data gets sent out, otherwise it'll be confusing why send is getting called twice.

ironfish/src/network/peers/addressManager.ts
```diff
@@ -2,24 +2,149 @@
...
+      identity: peer.identity ?? null,
+      name: peer.name ?? null,
+    }))
+
+    const possiblePeerStrings = this.possiblePeerAddresses
+      .filter((peerAddress) => !(peerAddress.address == null) && !(peerAddress.port == null))
+      .map(
+        (filteredPeerAddress) =>
+          //eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          `${filteredPeerAddress.address!}:${filteredPeerAddress.port!}`,
```
dguenther, line 47: You can use a flatMap to combine the filter and map calls without a need for a non-null assertion.

ironfish/src/network/peers/peerManager.ts
```diff
@@ -717,6 +727,7 @@ export class PeerManager {
         if (peer.state.type === 'CONNECTED') {
           this.updateIdentifiedPeerMap(peer)
           peer.onStateChanged.off(handler)
+          peer.send({ type: InternalMessageType.peerListRequest })
```
dguenther, line 730: This should probably be in its own handler -- This will only send the request once when the peer is initially connected/identified, and not at all if the peer is already identified. I think if we disconnect and reconnect at a later time, we'd still want to request peers right away.

dguenther:
Still looking through, but here are some initial comments.

ironfish/src/node.ts
NullSoldier, line 307: there shouldn't be await here right?

ironfish/src/network/peers/peerConnectionManager.ts
```diff
@@ -86,6 +86,13 @@ export class PeerConnectionManager {
       }
     }
 
+    if (this.peerManager.canCreateNewConnections()) {
+      const peer = this.peerManager.getRandomDisconnectedPeer()
+      if (peer) {
+        this.connectToEligiblePeers(peer)
```
dguenther, line 92: The `connectAttempts` variable was added above to avoid spending too much time establishing connections to new peers in one event loop. This new connection mechanism doesn't use it, but I think it probably should.

ironfish/src/node.ts
```diff
@@ -108,7 +108,8 @@ export class IronfishNode {
       chain: chain,
       strategy: strategy,
       metrics: this.metrics,
-      addressManager: new AddressManager(hosts),
+      files: this.files,
+      dataDir: dataDir,
```
NullSoldier, line 112: Can you look into this, something doesn't seem right here. It doesn't seem like it's ever being passed into the constructor.

ironfish/src/network/peers/addressManager.ts
```diff
@@ -2,24 +2,149 @@
...
 export class AddressManager {
-  hosts: HostsStore
+  hostsStore: HostsStore
 
-  constructor(hostsStore: HostsStore) {
-    this.hosts = hostsStore
+  constructor(files: FileSystem, dataDir?: string) {
+    this.hostsStore = new HostsStore(files, dataDir)
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    this.hostsStore.load()
```
dguenther, line 21: I think this lint error should be addressed rather than disabled, since there could be bugs related to other code executing before the loading has finished.

ironfish/src/network/peers/addressManager.ts
```diff
@@ -2,24 +2,149 @@
...
+    return this.hostsStore.getArray('possiblePeers')
+  }
+  /**
+   * Adds addresses associated to peers received from peer list
+   */
+  addAddressesFromPeerList(peerList: PeerList): void {
+    const newAddresses: PeerAddress[] = peerList.payload.connectedPeers.map((peer) => ({
+      address: peer.address,
+      port: peer.port,
+      identity: peer.identity ?? null,
```
dguenther, line 38: I don't think identity can be nullable in the peer list -- we should update the types if it is

ironfish/src/network/peers/peerManager.ts
```diff
@@ -1478,6 +1490,8 @@ export class PeerManager {
 
     let changed = false
 
+    this.addressManager.addAddressesFromPeerList(peerList)
```
dguenther, line 1493: We have a filter a couple lines down to make sure we filter ourselves out of the new peer set, and it'd probably be worth filtering ourselves out of the address manager as well. Would it make sense to name this something more generic, like `addPossiblePeerAddresses()`, and have the function take basically whatever the type of `newPeerSet.values()` is? Then this could be moved down below the `newPeerSet.delete` call.

ironfish/src/network/peers/peerManager.ts
```diff
@@ -866,6 +876,8 @@ export class PeerManager {
       }
       this.peers = this.peers.filter((p) => p !== peer)
 
+      this.addressManager.removePeerAddress(peer)
```
dguenther, line 879: This is a slightly tighter bound than we need to remove a peer from the peer addresses -- really we only care if the connection retry is in `willNeverRetryConnecting`, regardless of whether the peer has a connected peer.

Thinking about this also made me again consider whether `possiblePeers` is useful to keep around. I think if a peer address gets removed, it'll get added back into the possiblePeers next time a peerlist message is received. In other words, I think right now, on average `possiblePeers` will contain every peer with an address/port in `this.peers`.

ironfish/src/network/peers/peerManager.ts
```diff
@@ -782,55 +793,54 @@ export class PeerManager {
...
+    }
+  }
 
-      peer.send(peerList)
+  /**
+   * Gets a random disconnected peer address and returns a peer created from
+   * said address
+   */
+  getRandomDisconnectedPeer(): Peer | null {
+    const peerAddress = this.addressManager.getRandomDisconnectedPeerAddress(this.peers)
```
dguenther, line 836: This ends up filtering down to all known peer identities -- I think we could use `this.identifiedPeers.keys()` to get that same set.

ironfish/src/network/peers/peerManager.ts
```diff
@@ -782,55 +793,54 @@ export class PeerManager {
...
+      peer.send(peerListRequest)
+    }
+  }
 
-      peer.send(peerList)
+  /**
+   * Gets a random disconnected peer address and returns a peer created from
+   * said address
+   */
+  getRandomDisconnectedPeer(): Peer | null {
```
dguenther, line 835: Bit of a nitpick, but I think this should be named `create` rather than `get` since it can have the effect of adding a new peer to the peermanager.

ironfish/src/network/peerNetwork.ts
```diff
@@ -396,12 +398,14 @@ export class PeerNetwork {
...
-    this.started = false
-    this.peerConnectionManager.stop()
-    this.peerManager.stop()
-    this.webSocketServer?.close()
-    this.updateIsReady()
+  async stop(): Promise<void> {
+    await Promise.allSettled([
+      (this.started = false),
+      this.peerConnectionManager.stop(),
+      await this.peerManager.stop(),
```
dguenther, line 405: Similar comment here where the `await` inside of a Promise.allSettled isn't needed. I think you could drop the `Promise.allSettled` for now, since as far as I know there's only one async function here right now.

ironfish/src/network/peers/addressManager.ts
```diff
@@ -2,24 +2,149 @@
...
+
+    const priorConnectedAddressSet = new Set([...this.priorConnectedPeerAddresses])
+
+    const disconnectedPriorAddresses = this.filterConnectedAddresses(
+      priorConnectedAddressSet,
+      currentPeerAddresses,
+    )
+    if (disconnectedPriorAddresses.length) {
+      return ArrayUtils.sampleOrThrow(disconnectedPriorAddresses)
+    } else {
```
dguenther, line 90: This `else` could be flattened out to save a bit of nesting

ironfish/src/network/peers/peerManager.ts
NullSoldier: Why is the peer manager telling the AddressManager when to save? We're interesting the first top level system which doesn't control it's own event loop. So were diverging from other examples in the code base now right?

ironfish/src/network/peers/peerManager.ts
NullSoldier: It looks like this was copied from the other parts of our code base without reading what it does, because what promises is it waiting to settle on? If you aren't sure what a piece of our code is doing, you can always check the docs and play around with it locally before copying it. Check out the docs on `Promise.allSettled` here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

ironfish/src/network/peers/addressManager.ts
```diff
@@ -2,24 +2,149 @@
...
+    if (
+      this.possiblePeerAddresses.length === 0 &&
+      this.priorConnectedPeerAddresses.length === 0
+    ) {
+      return null
+    }
+
+    const currentPeerAddresses = new Set(
+      //eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      peers.filter((peer) => peer.state.identity !== null).map((peer) => peer.state.identity!),
```
dguenther, line 79: These are identities rather than addresses, so the name is slightly off here. `flatMap` could be used to avoid the eslint disabling here as well, otherwise I think you could have it take a list of identities rather than peers to avoid having to do the filter here

ironfish/src/network/peers/addressManager.ts
```diff
@@ -2,24 +2,149 @@
...
+        return ArrayUtils.sampleOrThrow(disconnectedPossibleAddresses)
+      } else {
+        return null
+      }
+    }
+  }
+
+  private filterConnectedAddresses(
+    addressSet: Set<Readonly<PeerAddress>>,
+    connectedPeerAdresses: Set<string>,
```
dguenther, line 106: These are identities, also a slight typo with `Addresses`

ironfish/src/network/peers/addressManager.ts
```diff
@@ -2,24 +2,149 @@
...
+   * Persist all currently connected peers and unused peer addresses to disk
+   */
+  async save(peers: Peer[]): Promise<void> {
+    const inUsePeerAddresses = peers
+      .filter(
+        (peer) =>
+          peer.state.type === 'CONNECTED' &&
+          !peer.getConnectionRetry(ConnectionType.WebSocket, ConnectionDirection.Outbound)
+            ?.willNeverRetryConnecting,
+      )
```
dguenther, line 140: This is a TS-enforced assertion, but I think the `?.` operator isn't necessary here if the ConnectionDirection is Outbound.

ironfish/src/network/peers/addressManager.ts
```diff
@@ -2,24 +2,149 @@
...
+      .map((peer) => ({
+        address: peer.address,
+        port: peer.port,
+        identity: peer.state.identity ?? null,
+        name: peer.name ?? null,
+      }))
+    this.hostsStore.set('priorPeers', [...inUsePeerAddresses])
```
dguenther, line 147: The value of `inUsePeerAddresses` is not exactly what we want for `priorPeers` -- I think we'd want all peers whom we've successfully established an outbound WebSocket connection to at least once in the past, but may or may not be currently connected to.

This is probably a non-trivial change, so now that I think about it, I'm not sure it's worth doing in this PR, but maybe a comment mentioning that would be nice.

ironfish/src/network/peers/peerManager.ts
```diff
@@ -87,17 +82,25 @@ export class PeerManager {
...
    * setInterval handle for peer disposal, which removes peers from the list that we
    * no longer care about
    */
   private disposePeersHandle: SetIntervalToken | undefined
 
+  /**
+   * setInterval handle for peer address persistence, which saves connected
+   * peers to disk
+   */
+  private savePeerAddressesHandle: ReturnType<typeof setInterval> | undefined
```
NullSoldier, line 102: It looks like above we use `SetIntervalToken` for this

ironfish/src/network/peers/addressManager.ts
```diff
@@ -2,24 +2,149 @@
...
+  addAddressesFromPeerList(peerList: PeerList): void {
+    const newAddresses: PeerAddress[] = peerList.payload.connectedPeers.map((peer) => ({
+      address: peer.address,
+      port: peer.port,
+      identity: peer.identity ?? null,
+      name: peer.name ?? null,
+    }))
+
+    const possiblePeerStrings = this.possiblePeerAddresses
+      .filter((peerAddress) => !(peerAddress.address == null) && !(peerAddress.port == null))
```
NullSoldier, line 43: Why not just write != null? This is a pretty complicated way to invert logic in a boolean expression.

ironfish/src/network/peers/addressManager.ts
NullSoldier: yeah in general, no null assertion is not good because it disables the type system, which is very good to us.

ironfish/src/network/peers/addressManager.ts
```diff
@@ -2,24 +2,149 @@
...
+          //eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          `${filteredPeerAddress.address!}:${filteredPeerAddress.port!}`,
+      )
+    const possiblePeerSet = new Set(possiblePeerStrings)
+
+    const dedupedAddresses = newAddresses.filter(
+      (newAddress) =>
+        !(newAddress.address == null) &&
+        !(newAddress.port == null) &&
+        !possiblePeerSet.has(`${newAddress.address}:${newAddress.port}`),
```
NullSoldier, line 55: You can invert a boolean expression in JS using `!=`

ironfish/src/network/peers/addressManager.ts
```diff
@@ -2,24 +2,149 @@
...
+        return null
+      }
+    }
+  }
+
+  private filterConnectedAddresses(
+    addressSet: Set<Readonly<PeerAddress>>,
+    connectedPeerAdresses: Set<string>,
+  ): PeerAddress[] {
+    const disconnectedAddresses = [...addressSet].filter(
```
NullSoldier, line 108: Why is this being copied, when filter will already copy it?

ironfish/src/network/peers/peerManager.ts
NullSoldier: It feels like ultimately we should remove known peer identities that aren't connected from the PeerManager.

ironfish/src/network/peers/peerManager.ts
```diff
@@ -782,55 +793,54 @@ export class PeerManager {
...
+   */
+  getRandomDisconnectedPeer(): Peer | null {
+    const peerAddress = this.addressManager.getRandomDisconnectedPeerAddress(this.peers)
+    if (peerAddress) {
+      const peer = this.getOrCreatePeer(peerAddress.identity)
+      peer.setWebSocketAddress(peerAddress.address, peerAddress.port)
+      peer.name = peerAddress.name || null
+      return peer
+    } else {
+      return null
```
NullSoldier, line 843: This is purely my opinion but I see this done a lot, and I'm kind of convinced looking at other code bases that it's simpler to have less nesting and have the main use case at the end.

```suggestion
    if (!peerAddress) {
          return null
    }
    
    const peer = this.getOrCreatePeer(peerAddress.identity)
    peer.setWebSocketAddress(peerAddress.address, peerAddress.port)
    peer.name = peerAddress.name || null
    return peer
```

ironfish/src/network/peers/peerManager.ts
NullSoldier: yeah we could do more to also just use the filtering that happens below but I don't think we need to do that.

ironfish/src/network/peers/peerManager.ts
deekerno: I'll adjust the bound. As for `possiblePeers`, I'm inclined to agree with you; I think it requires a fair bit of work to ensure that it functions as intended.

ironfish/src/network/peers/peerManager.ts
deekerno: That's fair; I'll find a way to persist currently connected peers without having to directly involve the PeerManager.

ironfish/src/node.ts
deekerno: You're correct, it was not; fixed.

ironfish/src/network/peers/addressManager.ts
deekerno: As we're using identities as the sole filter here, I went with the latter option. Thanks!

ironfish/src/network/peers/peerManager.ts
deekerno: I'm removing the `possiblePeers` store as I'm not so sure of its utility anymore; `addAddressesFromPeerList()` was only used to add to the possible peer store, so I'll remove this.

ironfish/src/network/peers/peerManager.ts
deekerno: Yeah, that makes sense. I'll make the change.

ironfish/src/network/peers/peerManager.ts
deekerno: After looking at this over the weekend, I don't think that there's a clean way to do this other than what is already being done. AddressManager needs to get the peer information somehow to be able to save it to disk and all of that information lives in PeerManager. We can either do it this way or have AddressManager contain a reference to PeerManager somehow, but I'm not sure that makes the code any easier to follow.

ironfish/src/network/peers/peerManager.ts
```diff
@@ -783,55 +796,54 @@ export class PeerManager {
...
+  }
 
-      peer.send(peerList)
+  /**
+   * Gets a random disconnected peer address and returns a peer created from
+   * said address
+   */
+  createRandomDisconnectedPeer(): Peer | null {
+    const peerAddress = this.addressManager.getRandomDisconnectedPeerAddress(
+      Array.from(this.identifiedPeers.keys()),
```
dguenther, line 837: Is this value supposed to be only connected/connecting peers? I'm pretty sure identifiedPeers can include disconnected peers as well, so this should be filtered by the peer state.

deekerno: It is, so I'll do a filter. Thanks!

dguenther (approved):

PR #944, opened 2022-01-27
feat(ironfish): Submit heap values on interval to telemetry

rohanjadvani:
## Summary

Adds an interval to submit memory from the metrics monitor to telemetry every 15 seconds.

## Testing Plan

Manually tested.

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/metrics/metricsMonitor.ts
+++ b/ironfish/src/metrics/metricsMonitor.ts
@@ -3,6 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 import { createRootLogger, Logger } from '../logger'
+import { submitMetric } from '../telemetry'
 import { SetIntervalToken } from '../utils'
 import { Gauge } from './gauge'
 import { Meter } from './meter'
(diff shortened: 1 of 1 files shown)


mat-if (approved):

ironfish/src/metrics/metricsMonitor.ts
```diff
@@ -78,4 +90,22 @@ export class MetricsMonitor {
...
+          value: this.heapUsed.value,
+        },
+        {
+          name: 'heap_total',
+          type: 'integer',
+          value: this.heapTotal.value,
+        },
+      ],
+    })
+  }
```
NullSoldier, line 110: If I was trying to find the place where telemetry is sent to the web, I would look in the telemetry system. How do you feel about moving this after your refactor of the telemetry system?

rohanjadvani: Yeah, that's the plan. This was mostly to keep the diff focused on uploading heap stats.

NullSoldier (approved):

PR #945, opened 2022-01-27
Add new testnet command to auto set up your node

NullSoldier:
## Summary
this new command will fetch a users test net user and auto configure
their node with their testnet user with their graffiti and node name.

## Testing Plan
Manually tested

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish-cli/src/commands/testnet.ts
+++ b/ironfish-cli/src/commands/testnet.ts
@@ -53,14 +53,12 @@ export default class Testnet extends IronfishCommand {
 
     let userId: number | null = null
 
-    const index = userArg.indexOf('/users/')
+    const index = userArg.indexOf('users/')
     if (index !== -1) {
-      const id = Number(userArg.slice(index + '/users/'.length))
+      userArg = userArg.slice(index + 'users/'.length)
+    }
 
-      if (!isNaN(id)) {
-        userId = id
-      }
-    } else if (!isNaN(Number(userArg))) {
+    if (!isNaN(Number(userArg))) {
       userId = Number(userArg)
     }
 
--- a/ironfish-cli/src/commands/testnet.ts
+++ b/ironfish-cli/src/commands/testnet.ts
@@ -53,8 +53,8 @@ export default class Testnet extends IronfishCommand {
 
     let userId: number | null = null
 
-    if (userArg.indexOf('/users/') !== -1) {
-      const index = userArg.indexOf('/users/')
+    const index = userArg.indexOf('/users/')
+    if (index !== -1) {
       const id = Number(userArg.slice(index + '/users/'.length))
 
       if (!isNaN(id)) {
--- a/ironfish-cli/src/commands/testnet.ts
+++ b/ironfish-cli/src/commands/testnet.ts
@@ -70,7 +70,7 @@ export default class Testnet extends IronfishCommand {
     }
 
     // request user from API
-    this.log(`Asking Ironfish who user ${userId} is...`)
+    this.log(`Asking Iron Fish who user ${userId} is...`)
 
     const api = new WebApi()
     const user = await api.getUser(userId)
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/commands/testnet.ts
```diff
@@ -0,0 +1,143 @@
...
+      userId = Number(userArg)
+    }
+
+    if (userId === null) {
+      this.log(`Could not figure out testnet user id from ${userArg}`)
+      this.exit(1)
+    }
+
+    // request user from API
+    this.log(`Asking Ironfish who user ${userId} is...`)
```
deekerno, line 73: Minor branding nit: Should be "Iron Fish".

deekerno (approved):
Manually tested, works on my end. Left one minor nit.

ironfish/src/webApi.ts
```diff
@@ -100,6 +108,13 @@ export class WebApi {
     return response.data.data
   }
 
+  async getUser(id: number): Promise<ApiUser | null> {
+    return await axios
+      .get<ApiUser>(`${this.host}/users/${id}`, this.options())
+      .then((r) => r.data)
+      .catch(() => null)
```
brekk, line 115: Are there error cases which necessitate this line?

ironfish-cli/src/commands/testnet.ts
```diff
@@ -0,0 +1,143 @@
...
+export default class Testnet extends IronfishCommand {
+  static hidden = false
+  static description = 'Set up your node to mine for the testnet'
+
+  static flags = {
+    [VerboseFlagKey]: VerboseFlag,
+    [DataDirFlagKey]: DataDirFlag,
+    confirm: flags.boolean({
+      default: false,
+      description: 'confirm without asking',
```
brekk, line 19: Capitalization required?

ironfish-cli/src/commands/testnet.ts
```diff
@@ -0,0 +1,143 @@
...
+      description: "Don't update your graffiti",
+    }),
+  }
+
+  static args = [
+    {
+      name: 'user',
+      required: false,
+      description:
+        'the user id or url to a testnet user like https://testnet.ironfish.network/users/1080',
```
brekk, line 36: Same

ironfish-cli/src/commands/testnet.ts
```diff
@@ -0,0 +1,143 @@
...
+    let userArg = ((args.user as string | undefined) || '').trim()
+
+    if (!userArg) {
+      userArg = (await cli.prompt(
+        'Enter the user id or url to a testnet user like https://testnet.ironfish.network/users/1080\nUser ID or URL',
+        {
+          required: true,
+        },
+      )) as string
+      this.log('')
```
brekk, line 51: This is to print a newline?

ironfish-cli/src/commands/testnet.ts
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    let userId: number | null = null
```
brekk, line 54: Does TS require this construct?

ironfish-cli/src/commands/testnet.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    if (userArg.indexOf('/users/') !== -1) {
+      const index = userArg.indexOf('/users/')
```
brekk, line 57:
```suggestion
    const index = userArg.indexOf('/users/')
    if (index !== -1) {
```

ironfish/src/webApi.ts
NullSoldier: If the user is not found you get a 404 which produces an exception

ironfish-cli/src/commands/testnet.ts
NullSoldier: Some times, but not in this case. If ommitted the type is "materialized" at the first line that requires it be a certain type.

ironfish-cli/src/commands/testnet.ts
NullSoldier: yes

ironfish-cli/src/commands/testnet.ts
NullSoldier: which part? I don't think so.

ironfish-cli/src/commands/testnet.ts
NullSoldier: fixed

PR #947, opened 2022-01-28
Convert native node code from neon to napi-rs

dguenther:
## Summary

napi-rs has a pretty convenient built-in system for generating prebuilt npm packages for Rust-based modules, plus has other features that would help us out (BigInt support, Promise support, automatically generating TS definitions and JS bindings). Since I'm working on generating prebuilt packages, I wanted to make this change first.

This is pretty likely to fail CI somewhere, so I'm putting this in draft until CI passes.

## Testing Plan

I tried generating a transaction, mining, and syncing blocks from another node.

## Breaking Change

This shouldn't break compatibility with other versions of the node, but it will require an upgrade of your Rust compiler if building from source.

```
[ ] Yes
[x] No
```


dguenther:
Okay, I think this one is ready for review now. I'm going to skim it myself too to make sure nothing unexpected slipped in.

ironfish-rust-nodejs/src/lib.rs
```diff
@@ -2,225 +2,81 @@
...
+    initial_randomness: i64,
+    target_buffer: Buffer,
+    batch_size: i64,
+) -> MineHeaderNapiResult {
     let mut target_array = [0u8; 32];
-    target_array.copy_from_slice(&target_slice[..32]);
-    // Argument 4
-    let batch_size = cx.argument::<JsNumber>(3)?.value(&mut cx) as i64;
+    target_array.copy_from_slice(&target_buffer[..32]);
+    print!("");
```
mat-if, line 68: is this still needed? guessing just missed it

mat-if (approved):
much nicer looking

PR #949, opened 2022-01-28
feat(ironfish): Generate telemetry node id in internal store

rohanjadvani:
## Summary

Generates a unique node id for telemetry in the internal store. This value is reset on every first run or set if it does not already exist on start.

## Testing Plan

Manually tested and updated unit test.

Screen Shot 2022-01-31 at 3 42 01 PM

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish-cli/src/commands/start.ts
+++ b/ironfish-cli/src/commands/start.ts
@@ -4,6 +4,7 @@
 import { flags } from '@oclif/command'
 import { Assert, IronfishNode, NodeUtils, PrivateIdentity, PromiseUtils } from 'ironfish'
 import tweetnacl from 'tweetnacl'
+import { v4 as uuid } from 'uuid'
 import { IronfishCommand, SIGNALS } from '../command'
 import {
   ConfigFlag,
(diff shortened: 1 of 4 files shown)


NullSoldier:
I'm not sure that we should do this, I would need to be convinced. Why do this instead of generate a new node secret we stored in internals. This also doesn't make sense because each node has many many public keys, one for each account. So which account are you using? That can also change over time.

rohanjadvani:
@NullSoldier makes sense, moved to the internal store.

ironfish-cli/src/commands/start.ts
```diff
@@ -204,6 +205,11 @@ export default class Start extends IronfishCommand {
       await this.firstRun(node)
     }
 
+    if (!node.internal.get('telemetryNodeId')) {
+      node.internal.set('telemetryNodeId', uuid())
+      await node.internal.save()
+    }
+
```
NullSoldier, line 212: We may have to move this to to the node later in case someone starts a node via the SDK.

NullSoldier (approved):

NullSoldier:
Tested manually, LGTM

PR #966, opened 2022-02-08
refactor(ironfish): Clean up existing Telemetry into single service

rohanjadvani:
## Summary

The existing telemetry implementation wasn't designed very intuitively. There was an existing `Telemetry` interface which was implemented by `NodeTelemetry` and `DisabledTelemetry`, each of which had a lot of side effects and returned new instances.

This code change:
* Refactors telemetry into a single service
* Adds a worker job for submitting telemetry to the API
* Convert `Metric` and `Field` from types to interfaces (inspired by TS [docs](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces))

## Testing Plan

Added unit tests and verified manually in InfluxDB for my `node_id`.

## Breaking Change

```
[ ] Yes
[x] No
```


ironfish/src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    try {
+      await this.pool.submitTelemetry(points)
+      this.logger.debug(`Submitted ${points.length} telemetry points`)
+    } catch (error: unknown) {
+      this.logger.error(`Error submitting telemetry to API: ${renderError(error)}`)
+
+      if (points.length < this.MAX_QUEUE_SIZE) {
+        this.logger.debug('Retrying telemetry submission')
+        this.points = points
```
NullSoldier, line 82: This needs to be concatted or else you'll wipe out queued points while the request is pending

NullSoldier: Later, we should dump points if we fail after a certain amount of times.

ironfish/src/workerPool/tasks/submitTelemetry.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+export type SubmitTelemetryResponse = {
+  type: 'submitTelemetry'
+}
+
+export async function submitTelemetry({
+  points,
+}: SubmitTelemetryRequest): Promise<SubmitTelemetryResponse> {
+  const api = new WebApi()
```
NullSoldier, line 19: I just noticed we don't consistency configure this from somewhere to let us test against staging, or let other users point at their own endpoint.

NullSoldier (approved):
This is a good PR, we can continue working on it after.

Issue #875, opened 2022-01-11
ironfish-master in windows still fail 1 test


QustomQure:
Summary of all failing tests
FAIL src/utils/currency.test.ts (5.482 s)
  ● Currency utils › displayIronAmountWithCurrency returns the right string

    expect(received).toEqual(expected) // deep equality

    Expected: "$IRON 0.00000002 ($ORE 2)"
    Received: "$IRON 0,00000002 ($ORE 2)"

       6 | describe('Currency utils', () => {
       7 |   test('displayIronAmountWithCurrency returns the right string', () => {
    >  8 |     expect(displayIronAmountWithCurrency(0.00000002, true)).toEqual('$IRON 0.00000002 ($ORE 2)')
         |                                                             ^
       9 |     expect(displayIronAmountWithCurrency(0.0000001, true)).toEqual('$IRON 0.00000010 ($ORE 10)')
      10 |     expect(displayIronAmountWithCurrency(0, true)).toEqual('$IRON 0.00000000 ($ORE 0)')
      11 |     expect(displayIronAmountWithCurrency(1, true)).toEqual(

      at Object.<anonymous> (src/utils/currency.test.ts:8:61)

QustomQure:
Graffiti Qualia

NullSoldier:
This is an interesting bug because it's related to localization and not windows. You can reproduce this on any platform by changing your local to any language that uses commas for number delimters. We should fix this, but is also an easy bug for someone to get PR credit for.

wd021:
just submitted PR for this - https://github.com/iron-fish/ironfish/pull/974

PR #974, opened 2022-02-09
Fix Issue #875

wd021:
Localization error is causing languages that use `,` instead of `.` to break (ie de_DE). Applying "en-US" in toLocaleString() to prevent this error.

### DIFF
--- a/ironfish/src/utils/currency.test.ts
+++ b/ironfish/src/utils/currency.test.ts
@@ -5,16 +5,20 @@ import { displayIronAmountWithCurrency, ironToOre, isValidAmount, oreToIron } fr
 
 describe('Currency utils', () => {
   test('displayIronAmountWithCurrency returns the right string', () => {
-    expect(displayIronAmountWithCurrency(0.00000002, true)).toEqual('$IRON 0.00000002 ($ORE 2)')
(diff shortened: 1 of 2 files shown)


Kupuyc:
Is it good approach - to hardcode locale? 🤔

Probably, it's better to use dotenv and set LC_ALL to "en_EN.UTF8", "de_DE.UTF8", etc. Doing so we can test localized messages by equivalence classes.

wd021:
Hardcoding because we're using en-US.UTF8 formatting for numerical displays.

Not sure what would be the need for setting de_DE.UTF8 or any other local format, it'd just fail the test. Unless we're localizing the whole app to account for different formats 🤷

NullSoldier:
I think we should make the test less restrictive, because I think this is going to make people in non en-us countries very unhappy.

wd021:
good point!

updated - tests now use `toLocaleString()`.

NullSoldier (approved):

NullSoldier:
Looks good, will merge if tests pass.

wd021:
@NullSoldier updated fix for lint issues.

wd021:
yes sir

dguenther (approved):
Looks good, thank you!

PR #977, opened 2022-02-10
Remove napi factory

dguenther:
## Summary

napi-rs has a concurrency bug when constructing objects using factory functions in worker threads. This change removes the use of factory functions in favor of constructors.

## Testing Plan

Tested that syncing a chain from a clean database doesn't cause incorrect block verification errors. Other than that, the strategy tests should pass.

## Breaking Change

```
[ ] Yes
[x] No
```


mat-if (approved):
was able to sync up with no issues 👍

PR #989, opened 2022-02-11
Increase verifier test coverage

AmberKiso:
## Summary
Improve verifier.ts test coverage.

## Expanded

- Mostly adding specific testcases to hit uncovered branches.
- Some changes to VerificationResultReason to expand error types and make errors clearer.
- Separated combined conditional at verifier.ts:69 to better cover both cases. It looks like the verifier.test.ts:88 covered a different case, but both are covered now.

One branch is not covered: verifier.ts:167 and I'm not sure how to hit it.
In order to get verifyTransaction to execute the catch{} block, miningPool.verify has to throw an error (or that is the first function that could). I haven't found any Jest magic to throw an error that gets passed into the function. Mocking verify to throw errors causes Jest to report the error. It's never getting processed by the code under test.

I'll put up another PR to break this out. I'm not sure if it's my lack of understanding of how throw-catch interact with Jest or an actual bug.

## Testing Plan

Ran `yarn test verifier` and `yarn test blockchain`. Both pass.
Also ran `yarn test:coverage:html` (Noticed failures in director.test.slow.ts - this also appears on the 0.1.21 base and I will file a bug shortly.)
`yarn lint` done. `yarn test` passes

New coverage metrics are:

> Statements 99.54% 441/443  
> Branches 99.05% 105/106 	
> Functions 100% 14/14

## Breaking Change

```
[ ] Yes
[X] No
```


AmberKiso:
I noticed the failures. Should I include the .fixture files to the commit?

dguenther:
Oh yep, we commit fixtures if your change causes them to need to be created/regenerated!

ironfish/src/consensus/verifier.ts
```diff
@@ -409,16 +413,19 @@ export enum VerificationResultReason {
...
+  INVALID_TRANSACTION_FEE = 'Transaction fee is incorrect',
   INVALID_TRANSACTION_PROOF = 'invalid transaction proof',
   INVALID_PARENT = 'invalid_parent',
   NOTE_COMMITMENT = 'note_commitment',
   NOTE_COMMITMENT_SIZE = 'Note commitment sizes do not match',
   NULLIFIER_COMMITMENT = 'nullifier_commitment',
   NULLIFIER_COMMITMENT_SIZE = 'Nullifier commitment sizes do not match',
   ORPHAN = 'Block is an orphan',
+  PREV_HASH_NULL = 'Previous hash is null',
+  PREV_HASH_UNEQUAL = 'previous hash not equal',
```
dguenther, line 428: Maybe this could be `PREV_HASH_INVALID`, and the string could be something like "previousBlockHash does not match expected hash"? Just brainstorming a bit -- I think "not equal" might briefly make readers wonder "not equal to what?"

ironfish/src/consensus/verifier.ts
```diff
@@ -86,7 +90,7 @@ export class Verifier {
     // minersFee should be (negative) miningReward + totalTransactionFees
     const miningReward = block.header.strategy.miningReward(block.header.sequence)
     if (minersFee !== BigInt(-1) * (BigInt(miningReward) + totalTransactionFees)) {
-      return { valid: false, reason: VerificationResultReason.INVALID_MINERS_FEE }
+      return { valid: false, reason: VerificationResultReason.INVALID_FEE_SUM }
```
dguenther, line 93: I think this could stay as INVALID_MINERS_FEE, since it's making sure that the minersFee derived field on the block header is actually correct.

ironfish/src/consensus/verifier.ts
```diff
@@ -66,10 +66,14 @@ export class Verifier {
...
 
       // Miner's fee should be only the first transaction
-      if ((i === 0 && fee > 0) || (i !== 0 && fee < 0)) {
+      if (i === 0 && fee > 0) {
         return { valid: false, reason: VerificationResultReason.INVALID_MINERS_FEE }
       }
 
+      if (i !== 0 && fee < 0) {
+        return { valid: false, reason: VerificationResultReason.INVALID_TRANSACTION_FEE }
+      }
```
dguenther, line 75: Nice change here, this is a good improvement 👍

ironfish/src/consensus/verifier.ts
```diff
@@ -66,10 +66,14 @@ export class Verifier {
       const fee = transactionFees[i]
 
       // Miner's fee should be only the first transaction
-      if ((i === 0 && fee > 0) || (i !== 0 && fee < 0)) {
+      if (i === 0 && fee > 0) {
         return { valid: false, reason: VerificationResultReason.INVALID_MINERS_FEE }
```
dguenther, line 70: This could potentially be something like MINERS_FEE_EXPECTED if changing the verification reason makes it easier to test. I don't really have a strong opinion here

ironfish/src/consensus/verifier.test.ts
```diff
@@ -85,7 +85,8 @@ describe('Verifier', () => {
       })
     })
 
-    it('rejects a block with miners fee as second transaction', async () => {
+    //this actually fails for transaction fee in the miners slot
```
dguenther, line 88: I think we can drop the comment here

ironfish/src/consensus/verifier.test.ts
```diff
@@ -85,7 +85,8 @@ describe('Verifier', () => {
...
+    it('rejects a block with transaction fee as first transaction', async () => {
```
dguenther, line 89: maybe "non-miner's transaction fee"? "transaction fee" can be a bit ambiguous between miner's fees and standard transaction fees.

dguenther (changes requested):
Mostly just some naming-related change requests, then if you can add in the fixtures as well, this'll be great, thanks!

dguenther:
Also, for the `catch{}` block test, maybe try `mockRejectedValue`? I'm not too sure offhand how to test that -- I would have expected that throwing an error from `mockImplementation` would have worked, so it might take a bit of research.

ironfish/src/consensus/verifier.ts
AmberKiso: Lets call it PREV_HASH_MISMATCH.

ironfish/src/consensus/verifier.ts
AmberKiso: Agree. The other 2 INVALID_MINERS_FEE I will change so we don't have different failure modes colliding with the same error code and to ensure our tests actually generate the failures we're expecting.

ironfish/src/consensus/verifier.ts
AmberKiso: Done

ironfish/src/consensus/verifier.test.ts
AmberKiso: Done

AmberKiso:
I'll put up my attempt for catch{} either tomorrow or Tuesday.

ironfish/src/consensus/verifier.ts
```diff
@@ -159,7 +163,7 @@ export class Verifier {
     }
 
     try {
-      return transaction.verify(options)
+      return await transaction.verify(options)
```
AmberKiso, line 166: Adding `await` fixed the issue I encountered with `throw()` not being processed properly.
It seemed (with my limited knowledge of ts) like the correct change to do since the call tree of `verifyTransaction` -> `transaction.verify` -> `workerPool.verify` are all `async` functions and just that `verifyTransaction` lacked `await` when calling one level deeper.

ironfish/src/consensus/verifier.ts
AmberKiso: Thank you!

AmberKiso:
Fixed up review comments.
Made strings in VerificationResultReason start with capitalized letters.
Added the last testcase and a code change to get to 100%.

dguenther:
Great, thanks for addressing those comments and sorry for the slow response -- I'm going to play around with these tests a bit and might push a commit or two onto this branch, then I'll merge away!

ironfish/src/consensus/verifier.ts
dguenther: Yep, this is a good catch -- Just to reiterate what you were saying, the issue is that without awaiting the promise, the `catch` here won't be hit if the promise happens to throw an error/reject.

dguenther (approved):

AmberKiso:
You're welcome!

PR #1000, opened 2022-02-13
add payment min fee notice

wd021:
When making a payment, I was only aware of the min payment fee after already submitting all the details.

Screenshot 2022-02-13 at 14 16 25

I propose adding the min fee message along with the `Enter the fee amount...` line.

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish-cli/src/commands/accounts/pay.ts
+++ b/ironfish-cli/src/commands/accounts/pay.ts
@@ -98,7 +98,7 @@ export class Pay extends IronfishCommand {
     if (!fee || Number.isNaN(Number(fee))) {
       fee = (await CliUx.ux.prompt('Enter the fee amount in $IRON', {
         required: true,
-        default: '0.00000001',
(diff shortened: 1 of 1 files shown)


mat-if (changes requested):
Is the issue that you were trying to enter a fee of 0? As far as I can tell, that's the only way to get to a fee validation error (besides not having enough $IRON etc).

Currently, if you use the pay command, and _don't_ enter a fee, it will use the default value, which is currently `0.00000001`. 

What you're changing here is the default value to a value that will not work. This sets the fee to `minimum $IRON 0.00000001`, which is not a valid amount because it is a bunch of text.

Instead, you probably want to add this string to the string that's being passed as the first value to `prompt`: `'Enter the fee amount in $IRON'`

wd021:
Ah, yea I just realized that.

It wasn't clear to me initially in the CLI prompt that `0.00000001` was the minimum until after I had entered with 1 too many zeros. I was intending to move up that information so the user knows beforehand.

As is the case, I'll close this PR.

PR #1005, opened 2022-02-14
Add new telemetry for block propagation

NullSoldier:
## Summary

This submits a new telemetry if the block has been successfully added.
I decided to only submit this telemtry so if invalid blocks are sent,
they aren't added to our data points. We also add data points for forks
and not forks, so that we can measure properly.

## Testing Plan
I will submit to local API with local influx attached.

## Breaking Change

```
[x] Yes
[ ] No
```

### DIFF
--- a/ironfish/src/telemetry/telemetry.ts
+++ b/ironfish/src/telemetry/telemetry.ts
@@ -166,4 +166,30 @@ export class Telemetry {
       ],
     })
   }
+
+  submitNewBlockSeen(block: Block, seenAt: Date): void {
+    this.submit({
+      measurement: 'propagation',
+      name: 'propagation',
+      timestamp: seenAt,
+      tags: [
+        {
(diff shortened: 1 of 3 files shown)


ironfish/src/telemetry/telemetry.ts
```diff
@@ -166,4 +166,34 @@ export class Telemetry {
...
+        },
+      ],
+      fields: [
+        {
+          name: 'block_timestamp',
+          type: 'integer',
+          value: block.header.timestamp.valueOf(),
+        },
+        {
+          name: 'block_sequence_field',
```
rohanjadvani, line 192: Do we need the sequence as both a tag and field?

NullSoldier: It was used, but no longer. I removed it from the tags.

rohanjadvani (approved):

PR #1006, opened 2022-02-14
Convert telemetry types to strict union types

NullSoldier:
## Summary
The interfaces before did not ensure that if you put "string" that you
then passed a string in the value. These union types now ensure that the
value matches the type that is specified.

## Testing Plan
Run tests, no testing needed.

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/telemetry/interfaces/tag.ts
+++ b/ironfish/src/telemetry/interfaces/tag.ts
@@ -1,7 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-export type Tag = {
+export interface Tag {
   name: string
   value: string
 }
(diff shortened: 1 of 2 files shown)


ironfish/src/telemetry/interfaces/tag.ts
```diff
@@ -1,7 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-export interface Tag {
+export type Tag = {
```
rohanjadvani, line 4: nit: As discussed offline, what do you think about keeping this as an [interface](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces)?

rohanjadvani (approved):

ironfish/src/telemetry/interfaces/tag.ts
NullSoldier: Reverted

PR #1007, opened 2022-02-14
Enhanced security/responsible disclosure policy

blrhc:
Improved the security policy based on widely accepted principles for responsible disclosure, as the previous version seemed a bit bare bones and vague.

```
[ ] Yes
[X] No
```

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,4 +2,19 @@
 
 ## Reporting a Vulnerability
 
-Please do not file a public issue on Github mentioning the vulnerability. To find out how to disclose a vulnerability, please email us at contact@ironfish.network.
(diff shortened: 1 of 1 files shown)


dguenther:
This is definitely something we want to solidify before mainnet launch, and we have it on our roadmap so we don't forget. But I think the change to the security policy needs to come as a result of the team setting up internal processes to handle responsible disclosures, so that disclosures get highlighted to the entire team as quickly as possible.

We're not ready to make those changes right now, so I'm going to close this PR, but I want to reiterate that we want to have a robust reporting policy and reward responsible disclosure, and improving this policy is something our team will do before mainnet launch.

Issue #941, opened 2022-01-27
Mining continues after node stopped 


AmberKiso:
### What happened?

3 windows command windows:
Window 1> `yarn start start` - normal startup and sync
2> `yarn start status` - returns

```
Version              0.1.20 @ src
Node                 STARTED
Heap Used            24.55 MiB / 26.39 MiB
Memory               240.88 MiB
P2P Network          CONNECTED - In: 0 B/s, Out: 0 B/s, peers 1
Mining               STOPPED - 0 miners, 0 mined
Mem Pool             0 tx
Syncer               IDLE @ 0 blocks per seconds
Blockchain           SYNCED @ HEAD 00000000000ffd6e92cf64aba975637b5dce7155e70947303831dbae709f72d0 (86686)
Workers              STARTED - 0 -> 0 / 6 - 0 jobs Δ, 0 jobs/s
```

3> `yarn start miners:start` - mining begins

2> ^C to get prompt, then `yarn start status` - returns

```
Version              0.1.20 @ src
Node                 STARTED
Heap Used            29.74 MiB / 33.64 MiB
Memory               401.46 MiB
P2P Network          CONNECTED - In: 34.59 KB/s, Out: 8.35 KB/s, peers 19
Mining               STARTED - 1 miners, 0 mined
Mem Pool             0 tx
Syncer               IDLE @ 0 blocks per seconds | avg time to add block 29.06 ms
Blockchain           SYNCED @ HEAD 0000000000036fb569b1b33f740090a47c969f9d3646eabef04dc6754e563bcf (86704)
Workers              STARTED - 0 -> 0 / 6 - 0 jobs Δ, 5.98 jobs/s
```
2> ^C to get prompt, then `yarn start stop` 
Window 1 reports `Not connected to the Iron Fish network` but mining continues in window 3.

2> ^C to get prompt, then `yarn start status`
Eventually I get an error message:

> $ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node bin/run status
> Error: connect ETIMEDOUT \\.\pipe\C:\Users\mjolnir\.ironfish\ironfish.ipc
>     at PipeConnectWrap.afterConnect [as oncomplete] (node:net:1161:16)
> error Command failed with exit code 1.

Mining **still** running.

3> ^C to kill mining and get prompt.
2> ^C to get prompt, then `yarn start status` Succeeds and shows correct state

```
Version              0.1.20 @ src
Node                 STOPPED
[...]

PR #1011, opened 2022-02-15
fix: `ironfish stop` is broken - issue #941

wd021:
## Summary

tldr: calling `yarn start stop` [doesn’t work properly.](https://github.com/iron-fish/ironfish/issues/941) Node shutdown function in `src/node.ts` never finishes because it hangs on `this.rpc.stop()`.

There are 2 ways to stop the node:

1. SIGINT / SIGTERM / SIGUS2 which initiates a `gracefulShutdown()` in `ironfish-cli/src/commands/`. This successfully processes `closeFromSignal()` in `./start.ts`, which handles `this.node?.shutdown()`.

1. `yarn start stop`. This initiates `start()` in `ironfish-cli/src/commands/stop`, which calls the sdkClient `stopNode()` function. This never completes as the shutdown() function in `src/node.ts` hangs on `this.rpc.stop()`

## Solution
Add an argument for the shutdown() function, ignoreRpc = false. when calling stopNode() from the sdkClient, we can pass a false value so it ignores the rpc.stop() call and continues with the shutdown() gracefully. 

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/node.ts
+++ b/ironfish/src/node.ts
@@ -307,12 +307,12 @@ export class IronfishNode {
     await this.shutdownPromise
   }
 
-  async shutdown(): Promise<void> {
+  async shutdown(ignoreRpc?: boolean): Promise<void> {
     await Promise.allSettled([
       this.accounts.stop(),
       this.syncer.stop(),
       this.peerNetwork.stop(),
-      this.rpc.stop(),
+      ...(ignoreRpc === true ? [] : [this.rpc.stop()]),
       this.telemetry.stop(),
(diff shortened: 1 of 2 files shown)


NullSoldier:
I think instead of doing this, stopping the RPC should just stop all adapters, which in turn should just end all in flight requests.

wd021:
gotcha, makes sense 👍

NullSoldier:
I just put up a real fix here, but we'll give you some testnet credit for the investigation and motivation to fix the bug.

PR #1013, opened 2022-02-15
Update RpcClient routes to use ApiNamespace enum

mat-if:
## Summary

Minor change to RpcClient to utilize the ApiNamespace enum for route definition. Since we have a non-stringly-typed enum already, using it here just minimizes the risk of a typo that breaks a route.

Also reorganized a couple routes to group routes by their namespace a little bit better.

## Breaking Change

```
[ ] Yes
[ ] No
```


NullSoldier (approved):
Looks great, thanks for making this change.

PR #1024, opened 2022-02-16
Mining Refactor

mat-if:
## Summary

Re-write mining code to simplify and add extra functionality

Main points:
- Replace existing mining endpoints with ones with more flexibility to better enable mining pool support, also allows room for future improvements as well
- Simplify code in MiningDirector and move code to new more sensible spots (MiningManager, Pool, etc) (For reference, MiningManager is the same general idea as the old MiningDirector)
- Add a mining pool
  - This includes a stratum-like protocol for communicating between pool and miner
- Refactor miner to utilize a pure rust thread pool instead of using the worker pool from the node for better CPU utilization and add support for the new endpoints and pool functionality

## Breaking Change

```
[ ] Yes
[ ] No
```


NullSoldier (approved):

deekerno (approved):
Absolute unit of a PR.

PR #1033, opened 2022-02-17
fix(ironfish): Pass in metrics and remove peer network from telemetry

rohanjadvani:
## Summary

The peer network and metrics from the node are [not passed](https://github.com/iron-fish/ironfish/blob/master/ironfish/src/node.ts#L191-L199) into the telemetry service, so traffic, peer counts, and heap usage are not collected. This code change fixes the dependencies (and refactors how this data is passed through to the API).

Changes:

1. Fix [duplicate](https://github.com/iron-fish/ironfish/blob/master/ironfish/src/telemetry/telemetry.ts#L100-L109) heap usage and send [heap total](https://github.com/iron-fish/ironfish/blob/81bb9395723f9293d8a665c7f776e186b28919b0/ironfish/src/telemetry/telemetry.ts#L98-L107)
1. Change types in traffic fields to [float](https://github.com/iron-fish/ironfish/blob/81bb9395723f9293d8a665c7f776e186b28919b0/ironfish/src/telemetry/telemetry.ts#L108-L117)
1. Fetch peer counts from a [gauge](https://github.com/iron-fish/ironfish/blob/81bb9395723f9293d8a665c7f776e186b28919b0/ironfish/src/telemetry/telemetry.ts#L121) in the metrics monitor

There were a couple considerations in removing the peer network as an argument in the telemetry constructor:
* There is [coupling](https://github.com/iron-fish/ironfish/blob/master/ironfish/src/node.ts#L110) between the peer network and node. That can be refactored in the future, but this means the peer network cannot be created in the [`init`](https://github.com/iron-fish/ironfish/blob/master/ironfish/src/node.ts#L131) method.
* Moving the telemetry service to the node's constructor makes dependencies difficult. The mining director [requires telemetry](https://github.com/iron-fish/ironfish/blob/master/ironfish/src/node.ts#L224) on creation. If the mining director is moved to the node's constructor, there are [circular dependencies](https://github.com/iron-fish/ironfish/blob/master/ironfish/src/node.ts#L96-L125) between the syncer, peer network, telemetry, and mining director. 
* The mining director has to be created before the peer network to avoid an undefined dereference on [`onNewBlock`](https://github.com/iron-fish/ironfish/blob/master/ironfish/src/network/peerNetwork.ts#L269-L271).

The considerations to use a gauge in the metric monitor:
* This [call](https://github.com/iron-fish/ironfish/blob/81bb9395723f9293d8a665c7f776e186b28919b0/ironfish/src/network/peerNetwork.ts#L190) could have been a method the telemetry service exposes. However, this would cause points to be submitted on every connect / disconnect state transition (which seems more granular / frequent than what we want).
* The metric monitor already is present in the peer network, so adding a gauge that is updated and processed as the telemetry service decides seemed like the cleanest option.

## Testing Plan

Ran node and verified points were present in Influx.

Screen Shot 2022-02-17 at 12 53 42 AM

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/metrics/metricsMonitor.ts
+++ b/ironfish/src/metrics/metricsMonitor.ts
@@ -12,10 +12,10 @@ export class MetricsMonitor {
   private _meters: Meter[] = []
   private readonly logger: Logger
 
+  readonly peersCount: Gauge
   readonly p2p_InboundTraffic: Meter
   readonly p2p_InboundTraffic_WS: Meter
   readonly p2p_InboundTraffic_WebRTC: Meter
-
   readonly p2p_OutboundTraffic: Meter
   readonly p2p_OutboundTraffic_WS: Meter
   readonly p2p_OutboundTraffic_WebRTC: Meter
@@ -29,10 +29,10 @@ export class MetricsMonitor {
   constructor({ logger }: { logger?: Logger }) {
     this.logger = logger ?? createRootLogger()
 
+    this.peersCount = new Gauge()
     this.p2p_InboundTraffic = this.addMeter()
     this.p2p_InboundTraffic_WS = this.addMeter()
     this.p2p_InboundTraffic_WebRTC = this.addMeter()
-
     this.p2p_OutboundTraffic = this.addMeter()
     this.p2p_OutboundTraffic_WS = this.addMeter()
     this.p2p_OutboundTraffic_WebRTC = this.addMeter()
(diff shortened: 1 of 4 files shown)


ironfish/src/metrics/metricsMonitor.ts
```diff
@@ -12,10 +12,10 @@ export class MetricsMonitor {
   private _meters: Meter[] = []
   private readonly logger: Logger
 
+  readonly peersCount: Gauge
```
NullSoldier, line 15: can we prefix this with p2p like the other ones?

NullSoldier (approved):

PR #1035, opened 2022-02-17
Fix default path of chain:export command

### LINKED ISSUE #997: Error in chain export
### What happened?

I ran
```
ironfish chain:export
```
It showed me the output
```
Exporting chain from 1 -> 40987 to /usr/src/ironfish-graph-explorer/src/data.json
Exporting blocks: [========================================] 40987/40987 100% | ETA: 0s
    Error: ENOENT: no such file or directory, open '/usr/src/ironfish-graph-explorer/src/data.json'
    Code: ENOENT
```

### Version

ironfish-cli/0.1.21 linux-x64 node-v16.13.0

### Relevant log output

_No response_

Kupuyc:
## Summary
Fixes #997

Default path of `ironfish chain:export` command is constant and relative - `../ironfish-graph-explorer/src/data.json`. It means command will always fail except when called from `~/ironfish/*` subfolder (in situation when `-p` flag is not set). It possible to fix such behavior using `process.env.PWD`.

## Testing Plan
There is no test for this command - it should be tested manually. Run `ironfish chain:export` w/o `-p` flag from different places.

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish-cli/src/commands/chain/export.ts
+++ b/ironfish-cli/src/commands/chain/export.ts
@@ -40,17 +40,12 @@ export default class Export extends IronfishCommand {
 
   async start(): Promise<void> {
     const { flags, args } = await this.parse(Export)
-    const path = this.sdk.fileSystem.resolve(
-      this.sdk.fileSystem.join(flags.path || this.sdk.config.dataDir, 'data.json'),
-    )
-    const pathAccessible = await this.sdk.fileSystem
-      .access(path)
-      .then(() => true)
-      .catch(() => false)
 
-    if (!pathAccessible) {
-      this.error(`Path "${path}" is not accessible. Check -p argument and try again.`)
-    }
+    const exportDir = flags.path
+      ? this.sdk.fileSystem.resolve(flags.path)
+      : this.sdk.config.dataDir
+
+    const exportPath = this.sdk.fileSystem.join(exportDir, 'data.json')
 
     const client = await this.sdk.connectRpc()
 
@@ -60,7 +55,7 @@ export default class Export extends IronfishCommand {
     })
 
     const { start, stop } = await AsyncUtils.first(stream.contentStream())
-    this.log(`Exporting chain from ${start} -> ${stop} to ${path}`)
+    this.log(`Exporting chain from ${start} -> ${stop} to ${exportPath}`)
 
     const progress = CliUx.ux.progress({
       format: 'Exporting blocks: [{bar}] {value}/{total} {percentage}% | ETA: {eta}s',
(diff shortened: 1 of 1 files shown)


mat-if:
Checked this out a bit, I'm still seeing the issue. Testing on a mac, `PWD` seems to just return the directory from which the command was executed, not sure if this is different on other operating systems. Compared with `__dirname` just to test:

image

`__dirname` docs for reference https://nodejs.org/docs/latest/api/modules.html#__dirname

Kupuyc:
Yeah... probably it is the best way. Let's ask team members in discord next monday.

ironfish-cli/src/commands/chain/export.ts
```diff
@@ -18,7 +18,11 @@ export default class Export extends IronfishCommand {
       char: 'p',
       parse: (input: string): Promise<string> => Promise.resolve(input.trim()),
       required: false,
-      default: '../ironfish-graph-explorer/src/data.json',
+      default: (): Promise<string> => {
+        Assert.isNotUndefined(process.env.PWD)
+
+        return Promise.resolve(`${process.env.PWD}/../ironfish-graph-explorer/src/data.json`)
```
NullSoldier, line 24: This PR still uses incorrect relative paths. How about we just set the default path to the data directory in a folder called exports? I would recommend maybe just removing this default entirely, then below use the data dir from `this.sdk.config.dataDir` where the data dir is stored if no path is specified.

Kupuyc: It's done by second way: default value of flag set to empty string and we will resolve path using `this.sdk.config.dataDir` in that case.

ironfish-cli/src/commands/chain/export.ts
```diff
@@ -18,7 +18,7 @@ export default class Export extends IronfishCommand {
       char: 'p',
       parse: (input: string): Promise<string> => Promise.resolve(input.trim()),
       required: false,
-      default: '../ironfish-graph-explorer/src/data.json',
+      default: '',
```
NullSoldier, line 21:
```suggestion
```

NullSoldier (approved):

ironfish-cli/src/commands/chain/export.ts
Kupuyc: Done.
Also, I added path accessibility verification.

NullSoldier (approved):

NullSoldier (approved):

NullSoldier:
I just removed the check access, and just had it recursively create the folder for better usability. It's now good to go.

Issue #961, opened 2022-02-03
Bug


b0b1kk:
### What happened?

A bug happened!
You can find error 1 below:

root@ironfish:~# ironfish faucet
yarn run v1.22.17
$ yarn build && yarn start:js faucet
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node bin/run faucet

::::::::::          :::::::::::::::::
::::::::::::       :::::::::::::::::::
:::::::::::::     :::::::::::::::::::::
::::::::::::::   ::::::::::::::::::::::::
 ::::::::::::: ::::::::::        :::::::::
   :::::::::::::::::::::          ::::::::::
   :::::::::::::::::::::          :::::::::
 ::::::::::::: ::::::::::        :::::::::
::::::::::::::   ::::::::::::::::::::::::
:::::::::::::     :::::::::::::::::::::
::::::::::::       :::::::::::::::::::
::::::::::           ::::::::::::::::

Enter your email to stay updated with Iron Fish: *******@gmail.com
Collecting your funds... Too many faucet requests
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

### Version

0.1.21 @ src    

### Relevant log output

```shell
root@ironfish:~# ironfish faucet
yarn run v1.22.17
$ yarn build && yarn start:js faucet
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node bin/run faucet

::::::::::          :::::::::::::::::
::::::::::::       :::::::::::::::::::
:::::::::::::     :::::::::::::::::::::
::::::::::::::   ::::::::::::::::::::::::
 ::::::::::::: ::::::::::        :::::::::
   :::::::::::::::::::::          ::::::::::
   :::::::::::::::::::::          :::::::::
 ::::::::::::: ::::::::::        :::::::::
::::::::::::::   ::::::::::::::::::::::::
:::::::::::::     :::::::::::::::::::::
::::::::::::       :::::::::::::::::::
::::::::::           ::::::::::::::::

Enter your email to stay updated with Iron Fish: *******@gmail.com
Collecting your funds... Too many faucet requests
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
[...]

PR #1048, opened 2022-02-19
Faucet improvements (fix bug #961)

wd021:
Improve Faucet
- email should be a required flag in the CLI prompt (because it is required by the client)
- `Too many faucet requests` and `You entered an invalid email.` errors result in `error Command failed with exit code 1.`, as seen in https://github.com/iron-fish/ironfish/issues/961. Improve error handling to exit gracefully here.

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish-cli/.yarnrc
+++ b/ironfish-cli/.yarnrc
@@ -1,0 +1,1 @@
+--silent true
--- a/ironfish-cli/.yarnrc
+++ b/ironfish-cli/.yarnrc
@@ -1,1 +1,0 @@
---silent true
--- a/ironfish-cli/src/commands/faucet.ts
+++ b/ironfish-cli/src/commands/faucet.ts
@@ -41,7 +41,7 @@ export class FaucetCommand extends IronfishCommand {
 
     if (!email) {
       email = (await CliUx.ux.prompt('Enter your email to stay updated with Iron Fish', {
-        required: false,
+        required: true,
       })) as string
     }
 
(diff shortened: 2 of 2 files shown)


ironfish-cli/src/commands/faucet.ts
```diff
@@ -41,7 +41,7 @@ export class FaucetCommand extends IronfishCommand {
 
     if (!email) {
       email = (await CliUx.ux.prompt('Enter your email to stay updated with Iron Fish', {
-        required: false,
+        required: true,
```
NullSoldier, line 44: Email isn't required, as it uses your public key if you don't specify an email.

NullSoldier:
This PR creates a bug in that an errored command now returns an exit code 0. I do agree the error that yarn outputs is really annoying. There is a recommended fix for this, and it's to use --silent in yarn. How about instead of this we just include this in all of our run commands?

ironfish-cli/src/commands/faucet.ts
wd021: Oh oops, you're right. I thought it was required cause leaving email blank results in `Collecting your funds... You entered an invalid email.`. Turns out it submits an empty string to the API, which always fails. So we need to pass `undefined` if a user skips email.

wd021:
Yea, that's a bit annoying. I think adding --silent in `.yarnrc` in the `ironfish-cli` folder is the best resolution. If necessary, perhaps add a line in the documentation for this.

ironfish-cli/.yarnrc
```diff
@@ -0,0 +1 @@
+--silent true
```
NullSoldier, line 1: This is the tricky thing I was trying to  point out. If you do this, it breaks all the logs for developers using yarn commands for install, build, etc because nothing is printed. Honestly, I think if you're using from source you have to be prepared to use engineering tooling and know how it works and be okay with these errors.

We're going to release NPM builds soon so 90% of users won't be running from source anymore and therefore should not be running yarn.

NullSoldier (approved):

NullSoldier:
This looks good, merging!

Issue #830, opened 2021-12-25
miners:mined miner [sic] formatting issues


AmberKiso:
IronFish Version       0.1.15
Running Docker 4.3.1, using their CLI
Windows 10 Home

Noticed a few minor formatting issues with the ironfish miners:mined command:
(Head node is roughly at 37838.)

> `# ironfish miners:mined -1000 1000`
Scanning for mined blocks...
Scanning for mined blocks from 36840 -> 36840
Scanning blocks: [----------------------------------------] 0/1 0% | ETA: {estimate} | SEQ {sequence}#

I'd expect the minimum start node to be at 0 (or 1 on how things are indexed), not negative.
The variables are not valid for estimate nor sequence.
This command also doesn't send a newline, so the # prompt is lost with the command output

> `# ironfish miners:mined 0 0`
Scanning for mined blocks...
Scanning for mined blocks from 1 -> 37885
Scanning blocks: [----------------------------------------] 61/37885 0% | ETA: N/A | SEQ 62

Double zero parameters = whole chain. Probably not what was intended.

> `# ironfish miners:mined -1000 -10000`
Scanning for mined blocks...
Scanning for mined blocks from 36875 -> 36875
Scanning blocks: [----------------------------------------] 0/1 0% | ETA: {estimate} | SEQ {sequence}#

> `# ironfish miners:mined -10000 -1000`
Scanning for mined blocks...
Scanning for mined blocks from 27876 -> 36876
Scanning blocks: [========================================] 8997/9001 99% | ETA: soon | SEQ 36873#

Testing negative numbers. Definitely an unexpected block range with the second example. Better to just refuse negative numbers.

> `# ironfish miners:mined 37000 99999999999`
Scanning for mined blocks...
Scanning for mined blocks from 37000 -> 37840
Scanning blocks: [========================================] 832/841 98% | ETA: soon | SEQ 37832#

>`# ironfish miners:mined 37840 37840`
Scanning for mined blocks...
Scanning for mined blocks from 37840 -> 37840
Scanning blocks: [----------------------------------------] 0/1 0% | ETA: {estimate} | SEQ {sequence}#

> `# ironfish miners:mined 37839 37840`
Scanning for mined blocks...
Scanning for mined blocks from 37839 -> 37840
Scanning blocks: [----------------------------------------] 0/2 0% | ETA: {estimate} | SEQ {sequence}#

> `# ironfish miners:mined 37800 37840`
Scanning for mined blocks...
[...]

PR #1051, opened 2022-02-20
Sanitize input parameters

### LINKED ISSUE #1034: miners:mined fractional inputs causes blocks to be missed
### What happened?

Running a bunch of miner:mined commands around my block. Found one with fractions that will cause it to be missed.
(Block at 75613)
miners:mined 75610 75615 -> HIT
miners:mined 75610.1 75700 -> MISS (unexpected)
miners:mined 75500.1 76000 -> MISS (unexpected)
miners:mined 75500 76000.1 -> HIT

I tried rounding down the parameters in my `truncate-mined-parameters` branch to fix #830, but this issue popped up.
It exists on 0.1.23.

### Version

0.1.23

### Relevant log output

```shell
C:\Users\mjolnir\Documents\GitHub\ironfish\ironfish-cli>yarn start:once miners:mined 75610 75615
yarn run v1.22.17
$ yarn build && yarn start:js miners:mined 75610 75615
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node bin/run miners:mined 75610 75615
Scanning for mined blocks...
Scanning for mined blocks from 75610 -> 75615
00000000014be60565690c0aa19d9cede762fc7adfb637ddd1f4cfe7b4a48eb3 RedFish 20 MAIN 75613: view in web
Done in 2.50s.

C:\Users\mjolnir\Documents\GitHub\ironfish\ironfish-cli>yarn start:once miners:mined 75610.1 75700
yarn run v1.22.17
$ yarn build && yarn start:js miners:mined 75610.1 75700
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node bin/run miners:mined 75610.1 75700
Scanning for mined blocks...
Scanning for mined blocks from 75610.1 -> 75700
Scanning blocks: [========================================] 75700/90.89999999999418 100% | ETA: 0 | SEQ 75700
Done in 2.29s.

C:\Users\mjolnir\Documents\GitHub\ironfish\ironfish-cli>yarn start:once miners:mined 75500.1 76000
yarn run v1.22.17
$ yarn build && yarn start:js miners:mined 75500.1 76000
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node bin/run miners:mined 75500.1 76000
Scanning for mined blocks...
Scanning for mined blocks from 75500.1 -> 76000
Scanning blocks: [========================================] 76000/500.8999999999942 100% | ETA: 0 | SEQ 76000
Done in 2.25s.

C:\Users\mjolnir\Documents\GitHub\ironfish\ironfish-cli>yarn start:once miners:mined 75500 76000.1
yarn run v1.22.17
$ yarn build && yarn start:js miners:mined 75500 76000.1
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node bin/run miners:mined 75500 76000.1
Scanning for mined blocks...
Scanning for mined blocks from 75500 -> 76000.1
00000000014be60565690c0aa19d9cede762fc7adfb637ddd1f4cfe7b4a48eb3 RedFish 20 MAIN 75613: view in web
Scanning blocks: [========================================] 76000.1/501.1000000000058 100% | ETA: 0 | SEQ 76000.1
Done in 4.32s.
```

AmberKiso:
## Summary
Sanitize miners:mined input parameters

##Strategy
Fix #1034 and the remaining issues with negative numbers, double zero parameters, and non natural numbers as inputs in #830.
Change first truncates START and STOP to whole numbers, then ensures 1 <= START <= STOP
Note that if STOP < START on the command line, you will scan one block, START.

## Testing Plan
Ran cases in #830 and the outputs were in line with new constraints on parameters.
Getting a bunch of the async timeout errors in `yarn test`

I am curious whether I should use MathUtils or Math for the floor function. MathUtils.max causes lint errors when assigned to args.max

## Breaking Change

```
[ ] Yes
[X] No
```


ironfish-cli/src/commands/miners/mined.ts
```diff
@@ -46,6 +46,14 @@ export class MinedCommand extends IronfishCommand {
 
     this.log('Scanning for mined blocks...')
 
+    //Truncate fractions from parameters
+    args.stop = MathUtils.floor(args.stop, 0)
```
NullSoldier, line 50: This is a bit of a typescript thing, but look below at lines 58. It's typing these differently as `number | null` which means the underlying value is nullable because above the required is false. If they were typed properly from our CLI library then it would cause an error here.

If you look at MathUtils.floor it doesn't take `number | null` only `number` so were actually hiding a bug here.  When typescript libraries can't type things properly, they'll use any which means the type system is disabled. Consider this code, I'd say it has unexpected output.

```typescript
function addOne(value: number): number {
    return value + 1
}

const a: any = 1
const b: any = null

console.log(addOne(a))
console.log(addOne(b))
```

ironfish-cli/src/commands/miners/mined.ts
```diff
@@ -46,6 +46,14 @@ export class MinedCommand extends IronfishCommand {
...
+    args.start = MathUtils.floor(args.start, 0)
+
+    //Fix parameters so 1 <= start <= stop
+    args.start = Math.max(args.start, 1)
+    args.stop = Math.max(args.stop, args.start)
```
NullSoldier, line 55: I think these 2 lines should be deleted, as they break logic we explicitly support. I read your ticket and I think there is a misunderstanding in how we treat negative numbers in all of our commands. Can you read the comment I wrote on your ticket, https://github.com/iron-fish/ironfish/issues/830#issuecomment-1048250105?

ironfish-cli/src/commands/miners/mined.ts
AmberKiso: Sure

ironfish-cli/src/commands/miners/mined.ts
AmberKiso: Agreed with the reasoning from the discord conversation (negative number use case). Will remove these lines.

AmberKiso:
Jason requested to move the fix to the RPC layer. Investigating the code, there's 3 RPC schema and associated CLI commands that take input start and stop parameters:
ShowChainRequestSchema - chain:show
ExportChainStreamRequestSchema - chain:export
ExportMinedStreamRequestSchema - miners:mined

Because each schema calls getBlockRange, I propose the fix goes in there. It's already doing a bunch of the min-max stuff there.

AmberKiso:
Tested miners:mined with cases from #1034, passes.
Still getting the async timeout errors in `yarn test`.

NullSoldier (approved):

NullSoldier:
This needs to be rebased, I can do it.

PR #1056, opened 2022-02-22
Fix parse error for hosts.json

deekerno:
## Summary
There's been some reports of a JSON parse error when reading `hosts.json`; it looks like a null Unicode marker or empty string is being written. While we investigate why that may be happening, it may be prudent to allow for the node to continue to start if there's an error. Thus, for the `hostsStore` class, we try to load `hosts.json` and if there's a `ParseJsonError`, then we get rid of the corrupted content by overwriting with a valid JSON structure (this file is meant to be ephemeral anyways).

## Testing Plan
All tests pass. Additionally, I tested the parsing with a correctly-formed file, an incorrectly-formed file, and a nonexistent file.

## Breaking Change

```
[ ] Yes
[x] No
```


brekk (approved):
Makes sense to me

PR #1065, opened 2022-02-23
Check if `ironfish` is present in PATH in `debug` command

mat-if:
## Summary

This will remove 1 more question/variable when trying to debug issues, which hopefully will speed up the turn around time on resolving issues for users.

image

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish-cli/src/commands/debug.ts
+++ b/ironfish-cli/src/commands/debug.ts
@@ -1,6 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { execSync } from 'child_process'
 import { FileUtils, IronfishPKG, NodeUtils } from 'ironfish'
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/debug.ts
```diff
@@ -33,6 +34,14 @@ export default class Debug extends IronfishCommand {
 
     const telemetryEnabled = this.sdk.config.get('enableTelemetry').toString()
 
+    let cmdInPath: boolean
+    try {
+      execSync('ironfish --help &>/dev/null')
```
dguenther, line 39: Is there a way we could get this to work on Windows as well? Or at least have a carve-out for Windows that just says "UNKNOWN". Maybe there's a way to run the command and check the error code, I'm not totally certain

mat-if: 🤦 I changed to this implementation from my original one because it wasn't Windows compatible, then I went and added `/dev/null` anyway. Changed it to just have node ignore stdio for the child process which seems to work just the same

NullSoldier (approved):

PR #1069, opened 2022-02-23
Improvements for getting started on MacOS using Homebrew to manage de…

lancearlaus:
## Summary

* Change .gitattributes CRLF handling to auto to avoid having a modified working copy upon repo clone on MacOS
* Add a Brewfile for use with `brew bundle` command to easily install needed dependencies
* Add direnv .envrc file to automatically manage needed environment variables
  * Note the use of `layout node` in the .envrc file which adds the project's node_modules/.bin directory to the PATH. This improves repeatability by using project-specific dependencies (removing dangling references to those installed in the environment). It would be nice to use this to remove the dependency on global installation of Node v16, but unfortunately yarn doesn't create the necessary symlink in .bin (npm does). 

## Testing Plan

* Clone repo to Mac without Homebrew or bundle modules installed
* Install Homebrew
* Run `brew bundle` from top level project directory to install build dependencies
* Install direnv into preferred shell
* Run `direnv allow .` in the top level project directory to enable automatic environment management
* Run `yarn install` per README instructions, which should not require setting any additional environment variables (e.g. OPENSSL_ROOT_DIR)
* Test CRLF handling on repo clone to Windows environment (which I don't have handy to test)

## Breaking Change

```
[ ] Yes
[x] No
```


mat-if:
Interesting! I haven't dug into the details yet, but was the change to yarn.lock intentional? Looks like npm overwrote all the domains for some reason

lancearlaus:
> Interesting! I haven't dug into the details yet, but was the change to yarn.lock intentional? Looks like npm overwrote all the domains for some reason

No, that change was not intentional and should be omitted from the PR (which is now the case)

PR #1077, opened 2022-02-25
Fix block syncer status

Kupuyc:
## Summary

Tech PR to fix wrong usage of type system: 
1. GetStatusResponseSchema::blockSyncer.status has wrong list of possible values
1. GetStatusResponse::blockSyncer.status was `string` instead of union of literals. Because of which condition [`content.blockSyncer.status === 'IDLE'`](https://github.com/iron-fish/ironfish/blob/1f3dc311c6db990313f81ae7d8046d19e8b2ce93/ironfish-cli/src/commands/status.ts#L75) was always false - content.blockSyncer.status could be `idle` but not `IDLE`.

## Testing Plan
Run `yearn test`. 

## Breaking Change

```
[ ] Yes
[x] No
```


dguenther (approved):
This is great, thanks!

PR #1088, opened 2022-03-01
added basic cli endpoint for expected fee

nmfaulkner:
## Breaking Change

```
[ ] Yes
[x] No
```


ironfish-cli/src/commands/accounts/pay.ts
```diff
@@ -3,6 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 import { CliUx, Flags } from '@oclif/core'
+import { Console } from 'console'
```
NullSoldier, line 6: This is unused

ironfish-cli/src/commands/accounts/pay.ts
```diff
@@ -70,16 +71,20 @@ export class Pay extends IronfishCommand {
...
+    //   )
+    //   this.exit(1)
+    // }
 
     if (!amount || Number.isNaN(amount)) {
       const response = await client.getAccountBalance({ account: from })
 
+      const estimatedFee = await client.getEstimatedFee()
+
+      console.log("Estimated fee is: " + estimatedFee.content.fee);
```
NullSoldier, line 86: This is printed before asking the amount, not the fee. If you pass the `--amount` flag then this won't be printed.

ironfish/src/memPool/memPool.ts
```diff
@@ -110,6 +110,16 @@ export class MemPool {
...
   }
 
+  async getEstimatedFee(): Promise<{ fee: bigint;}> {
+    let memPoolEntry = this.queue.peek();
+    if (memPoolEntry == undefined){
+      return {fee: BigInt(0)}
+    }
+    else{
+      return { fee: memPoolEntry.fee }
+    }
```
NullSoldier, line 120: This is more of a preference, but in these cases we usually don't use an else, just have the main use case as the last statement.

ironfish/src/memPool/memPool.ts
```diff
@@ -110,6 +110,16 @@ export class MemPool {
     return true
   }
 
+  async getEstimatedFee(): Promise<{ fee: bigint;}> {
```
NullSoldier, line 113: This function doesn't do anything asynchronous so I think we can drop the async and promise.

NullSoldier:
Do you want me to take this over? I can take it over.

NullSoldier:
I'm going to take this over to finish

Issue #699, opened 2021-12-10
accounts:pay allows for very long, very granular (small decimal), and leading zeroes. Outputs back to console.


zkstorm:
**Issue**

I attempted to enter 3 kinds of values when using accounts:pay:

1. Very large and absurd values (100+ 9s for example)
2. Large number of leading zeroes
3. Small granular amounts that shouldn't be allowed (0.0000000100000001 IRON for example).

This never led to an invalid transaction on the blockchain, but in all cases, it let me input these up to the point it created the transaction (2 and 3 were valid and went through, however, as expected). For each one, the command line echoed back the value that I input, which may be misleading. Some additional input checking and formatting would be helpful to confirm what is actually happening. 

Screenshots demonstrate these in sufficient detail and are easily understood. In one case, the amount of ORE I'm transacting with is marked as "Infinity".

**Expectation**

Invalid values should be either truncated (e.g. 0.00000001001 -> 0.00000001) or marked as invalid. Leading 0s should be formatted off of the echo-backs. 

**Supporting Resources**

<img width="965" alt="Screen Shot 2021-12-10 at 3 45 10 PM" src="https://user-images.githubusercontent.com/84751530/145645965-0b538a5d-82f9-46f2-ab32-118916654323.png">
<img width="965" alt="Screen Shot 2021-12-10 at 3 49 23 PM" src="https://user-images.githubusercontent.com/84751530/145645993-710947ff-2805-4d1a-902b-135828315977.png">
<img width="965" alt="Screen Shot 2021-12-10 at 3 50 44 PM" src="https://user-images.githubusercontent.com/84751530/145645998-3fbf4cdd-a887-4a4d-a618-61cf0e563ca6.png">

wd021:
working on improving this

wd021:
should be fixed with https://github.com/iron-fish/ironfish/pull/1089

lwisne:
We believe this is now closed. Let us know if you don't agree!

PR #1086, opened 2022-02-28
feat: new cli commands `accounts:notes` and `accounts:transactions`

wd021:
Proposing to add an `accounts:transactions` command to display transaction notes for an account. I've been debugging some account balance issues lately and found this to be quite helpful. Here's what it looks like

Screen Shot 2022-02-28 at 2 04 16 PM

If we add this, I'll open a PR to update docs on the website repo.

Is this a breaking change? If yes, add notes below on why this is breaking and
what additional work is required, if any.

```
[ ] Yes
[x] No
```

PR #1089, opened 2022-03-01
improving `accounts` command

wd021:
## Summary
Some improvements and fixes for the `accounts` command. 

1 - enhancing validation checks for `accounts:pay`
2 - fix long number error. amount and fee values need to be wrapped in `Number()` for `displayIronAmountWithCurrency()`, which fixes issue identified in https://github.com/iron-fish/ironfish/issues/699
3 - add max amount check for `isValidAmount()` to prevent absurdly large values, to fix https://github.com/iron-fish/ironfish/issues/699
4 - add `accountName` for `accounts:balance`.

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish-cli/src/commands/accounts/pay.ts
+++ b/ironfish-cli/src/commands/accounts/pay.ts
@@ -92,7 +92,7 @@ export class Pay extends IronfishCommand {
         ),
       )
 
-      if (Number.isNaN(amount)) {
+      if (Number.isNaN(input)) {
         this.error(`A valid amount is required`)
       }
 
@@ -160,7 +160,7 @@ export class Pay extends IronfishCommand {
 
     if (!isValidPublicAddress(to)) {
       this.log(`A valid public address is required`)
-      this.exit(0)
+      this.exit(1)
     }
 
     if (expirationSequence !== undefined && expirationSequence < 0) {
--- a/ironfish-cli/src/commands/accounts/pay.ts
+++ b/ironfish-cli/src/commands/accounts/pay.ts
@@ -32,7 +32,7 @@ export class Pay extends IronfishCommand {
     }),
     amount: Flags.string({
       char: 'a',
-      description: 'amount of coins to send',
+      description: 'amount of coins to send in IRON',
     }),
     to: Flags.string({
       char: 't',
@@ -63,8 +63,8 @@ export class Pay extends IronfishCommand {
     let fee = flags.fee as unknown as number
     let to = flags.to
     let from = flags.account
+    let memo = flags.memo
     const expirationSequence = flags.expirationSequence
-    const memo = flags.memo || ''
 
     const client = await this.sdk.connectRpc()
 
@@ -90,18 +90,18 @@ export class Pay extends IronfishCommand {
         },
       )) as number
 
-      if (Number.isNaN(amount)) {
+      if (Number.isNaN(amount) || !isValidAmount(amount)) {
         this.error(`A valid amount is required`)
       }
     }
 
-    if (!fee || Number.isNaN(Number(fee))) {
+    if (!fee || Number.isNaN(fee)) {
       fee = (await CliUx.ux.prompt('Enter the fee amount in $IRON', {
         required: true,
         default: '0.00000001',
       })) as number
 
-      if (Number.isNaN(fee)) {
+      if (Number.isNaN(fee) || !isValidAmount(fee)) {
         this.error(`A valid fee amount is required`)
       }
     }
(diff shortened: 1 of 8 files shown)


wd021:
will look into failed tests.

wd021:
fixed failed tests. also cleaned up some spacing/wording issues with the test script.

ironfish-cli/src/commands/accounts/pay.ts
```diff
@@ -95,7 +95,7 @@ export class Pay extends IronfishCommand {
       }
     }
 
-    if (!fee || Number.isNaN(Number(fee))) {
+    if (!fee || Number.isNaN(fee)) {
```
NullSoldier, line 98: `Number.isNaN` just checks the type of the input. In this case it will always return false because fee is type string.

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN

ironfish-cli/src/commands/accounts/pay.ts
```diff
@@ -150,6 +150,11 @@ export class Pay extends IronfishCommand {
       this.exit(0)
     }
 
+    if (!isValidPublicAddress(to)) {
+      this.log(`A valid public address is required`)
+      this.exit(0)
```
NullSoldier, line 155: This is an error so it should exit 1

```suggestion
      this.exit(1)
```

ironfish-cli/src/commands/accounts/pay.ts
```diff
@@ -159,10 +164,10 @@ export class Pay extends IronfishCommand {
       this.log(`
 You are about to send:
 ${displayIronAmountWithCurrency(
-  amount,
+  Number(amount),
   true,
 )} plus a transaction fee of ${displayIronAmountWithCurrency(
-        fee,
+        Number(fee),
```
NullSoldier, line 170: This is a good fix, but it actually reveals that the types for these 2 in this whole file are broken. I can fix them in another PR, or do you want to fix them in this PR? For one, they are strings that are just type-casted (not value casted) to numbers when they are actually strings the whole time above.

ironfish/src/rpc/routes/accounts/getBalance.ts
```diff
@@ -6,7 +6,7 @@ import { ApiNamespace, router } from '../router'
 import { getAccount } from './utils'
 
 export type GetBalanceRequest = { account?: string }
-export type GetBalanceResponse = { confirmed: string; unconfirmed: string }
+export type GetBalanceResponse = { accountName: string; confirmed: string; unconfirmed: string }
```
NullSoldier, line 9: maybe this is bad on our part, but when we accept an account name we use the variable `account` like above on line 8. Can you make this consistent for now so at least they don't diverge until we decide to make a change here?

ironfish/src/utils/currency.ts
```diff
@@ -6,10 +6,11 @@ const ORE_TICKER = '$ORE'
 const IRON_TICKER = '$IRON'
 const ORE_TO_IRON = 100000000
 export const MINIMUM_IRON_AMOUNT = 1 / ORE_TO_IRON
+export const MAXIMUM_IRON_AMOUNT = 1e12
```
NullSoldier, line 9: Do we need a maximum? We store most amounts in 64 bit ints in the rust, so technically the max is `1.8446744e+19`, I don't see a reason we need to enforce a max anywhere. I don't even see a reason we need to enforce a minimum either, but I've been meaning to remove the minimum too.

NullSoldier:
I think overall this PR does a few useful things, so I'd love to get it in. I wrote some comments on a few things.

ironfish-cli/src/commands/accounts/pay.ts
```diff
@@ -77,33 +77,41 @@ export class Pay extends IronfishCommand {
...
+            oreToIron(Number(response.content.confirmed)),
+            false,
+          )})`,
+          {
+            required: true,
+          },
+        ),
+      )
 
       if (Number.isNaN(amount)) {
```
wd021: updated to `Number.isNaN(input)`

ironfish/src/rpc/routes/accounts/getBalance.ts
wd021: makes sense. updated.

ironfish/src/utils/currency.ts
wd021: i updated the max value. as to our conversation on discord, I don't know if you still want to enforce this logic to resolve https://github.com/iron-fish/ironfish/issues/699 or remove this altogether.

ironfish/src/rpc/routes/accounts/getBalance.ts
wd021: this change applies to PR #1086 too. will update accordingly.

NullSoldier (approved):

NullSoldier:
Thanks, looks good now!

PR #1093, opened 2022-03-04
Add serializable interface and base serializable worker message class

deekerno:
_Please note that this PR is being opened against the `serialize-worker-pool` branch._

## Summary
Added a serializable interface; it's placed in a top-level common folder in order to aid future serialization efforts. Also, added a base serializable worker message class; each task request/response type will be adjusted to use this new class.

## Testing Plan
All tests pass.

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/common/serializable.ts
+++ b/ironfish/src/common/serializable.ts
@@ -1,0 +1,11 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import bufio from 'bufio'
+
+export interface Serializable {
+  serialize(bw: bufio.BufferWriter): Buffer
+  deserialize(buffer: Buffer): Serializable
+  getSize(): number
+}
--- a/ironfish/src/common/serializable.ts
+++ b/ironfish/src/common/serializable.ts
@@ -5,7 +5,7 @@
 import bufio from 'bufio'
 
 export interface Serializable {
-  serialize(bw: bufio.BufferWriter): Buffer
+  serialize(): Buffer
   deserialize(buffer: Buffer): Serializable
   getSize(): number
 }
--- a/ironfish/src/common/serializable.ts
+++ b/ironfish/src/common/serializable.ts
@@ -2,8 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-import bufio from 'bufio'
-
 export interface Serializable {
   serialize(): Buffer
   deserialize(buffer: Buffer): Serializable
--- a/ironfish/src/workerPool/serializableMessage.ts
+++ b/ironfish/src/workerPool/serializableMessage.ts
@@ -1,0 +1,44 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import bufio from 'bufio'
+import { Serializable } from '../common/serializable'
+
+export enum WorkerMessageType {
+  BoxMessage = 'boxMessage',
+  CreateMinersFee = 'createMinersFee',
+  CreateTransaction = 'createTransaction',
+  GetUnspentNotes = 'getUnspentNotes',
+  MineHeader = 'mineHeader',
+  Sleep = 'sleep',
+  SubmitTelemetry = 'submitTelemetry',
+  TransactionFee = 'transactionFee',
+  UnboxMessage = 'unboxMessage',
+  VerifyTransaction = 'verifyTransaction',
+}
+
+export abstract class SerializableWorkerMessage implements Serializable {
+  constructor(id: number, type: WorkerMessageType) {
+    this.id = id
+    this.type = type
+  }
+
+  id: number
+
+  type: WorkerMessageType
+
+  abstract serialize(bw: bufio.BufferWriter): Buffer
+
+  abstract deserialize(buffer: Buffer): Serializable
+
+  abstract getSize(): number
+
+  serializeWithMetadata(): Buffer {
+    const bw = bufio.write()
+    bw.writeU64(this.id)
+    bw.writeVarString(this.type)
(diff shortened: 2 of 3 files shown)


rohanjadvani:
Can you rebase `staging` into the destination branch?

ironfish/src/workerPool/serializableMessage.ts
```diff
@@ -0,0 +1,44 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import bufio from 'bufio'
+import { Serializable } from '../common/serializable'
+
+export enum WorkerMessageType {
```
rohanjadvani, line 8: We can leave this empty for now and add values as each job is refactored over.

ironfish/src/workerPool/serializableMessage.ts
```diff
@@ -0,0 +1,44 @@
...
+  VerifyTransaction = 'verifyTransaction',
+}
+
+export abstract class SerializableWorkerMessage implements Serializable {
+  constructor(id: number, type: WorkerMessageType) {
+    this.id = id
+    this.type = type
+  }
+
+  id: number
```
rohanjadvani, line 27: nit: Can you move these fields above the constructor for consistency with the rest of the code base?

```ts
export abstract class SerializableWorkerMessage implements Serializable {
  id: number
  type: WorkerMessageType

  constructor(id: number, type: WorkerMessageType) {
    this.id = id
    this.type = type
  }

  ...
```

ironfish/src/workerPool/serializableMessage.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  type: WorkerMessageType
+
+  abstract serialize(bw: bufio.BufferWriter): Buffer
+
+  abstract deserialize(buffer: Buffer): Serializable
+
+  abstract getSize(): number
+
+  serializeWithMetadata(): Buffer {
```
rohanjadvani, line 37: We may need to add more stuff here but I think this is fine for now.

ironfish/src/workerPool/serializableMessage.ts
```diff
@@ -0,0 +1,44 @@
...
+  GetUnspentNotes = 'getUnspentNotes',
+  MineHeader = 'mineHeader',
+  Sleep = 'sleep',
+  SubmitTelemetry = 'submitTelemetry',
+  TransactionFee = 'transactionFee',
+  UnboxMessage = 'unboxMessage',
+  VerifyTransaction = 'verifyTransaction',
+}
+
+export abstract class SerializableWorkerMessage implements Serializable {
```
rohanjadvani, line 21: What do you think about `WorkerMessage`?

```suggestion
export abstract class WorkerMessage implements Serializable {
```

ironfish/src/workerPool/serializableMessage.ts
```diff
@@ -0,0 +1,44 @@
...
+  constructor(id: number, type: WorkerMessageType) {
+    this.id = id
+    this.type = type
+  }
+
+  id: number
+
+  type: WorkerMessageType
+
+  abstract serialize(bw: bufio.BufferWriter): Buffer
```
rohanjadvani, line 31: I wonder if we should just return `BufferWriter` here instead. It feels a bit odd to take a writer and then render it per serialization. It might be cleaner to either (1) return the updated writer instance (`serialize: (BufferWriter) -> BufferWriter`) or (2) take no writer and just return a buffer which the caller can use (`serialize: () -> Buffer`). 

If doing the first option, we will need to update [this line](https://github.com/iron-fish/ironfish/pull/1093/files#diff-4c76277f743b576804dcad6904503ff146f239d76860b1b6d6c1afab89fa6729R41) to `this.serialize(bw)`.

deekerno: I think that the second option makes more sense as I would expect `serialize()` to return a Buffer.

ironfish/src/workerPool/serializableMessage.ts
deekerno: That sounds good; changed it.

ironfish/src/workerPool/serializableMessage.ts
deekerno: Changed. 👍🏽

ironfish/src/workerPool/serializableMessage.ts
rohanjadvani: Cool that sounds good 👍

ironfish/src/workerPool/serializableMessage.ts
deekerno: `serializeWithMetadata()` depends on the string values for enum members in order to write a VarString to a Buffer. If we make this empty for now, perhaps we should remove the `bw.writeVarString(this.type)` call for now, and then add it back once we start implementing each task.

rohanjadvani: Yeah let's omit it for now so it'll be more explicit for subsequent PRs.

ironfish/src/common/serializable.ts
```diff
@@ -0,0 +1,11 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import bufio from 'bufio'
+
+export interface Serializable {
+  serialize(bw: bufio.BufferWriter): Buffer
```
rohanjadvani, line 8: We should update this too

ironfish/src/workerPool/serializableMessage.ts
```diff
@@ -0,0 +1,43 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
rohanjadvani, line 1: Can we rename this file to `workerMessage` now the class is renamed?

deekerno: Done. 💯

ironfish/src/common/serializable.ts
deekerno: Got it.

ironfish/src/workerPool/serializableMessage.ts
deekerno: Changed.

rohanjadvani (approved):

PR #1101, opened 2022-03-09
Add custom eslint rule for imports ending in '.'

mat-if:
## Summary

image

Few more things I need to do:
- [ ] Add unit tests
- [ ] Verify if we absolutely have to create a separate project or if it can live in `eslint-config-ironfish`
- [ ] Settle on better naming/messaging

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/config/eslint-config-ironfish/index.js
+++ b/config/eslint-config-ironfish/index.js
@@ -17,6 +17,7 @@ module.exports = {
 
   plugins: [
     'header',
+    'ironfish',
     'jest', 
     'prettier', 
     'simple-import-sort',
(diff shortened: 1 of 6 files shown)


ironfish/src/serde/Serde.ts
```diff
@@ -0,0 +1,45 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
mat-if, line 1: Moved this from the serde/index.ts because VSCode won't pick up a better import suggestion otherwise, plus it just makes sense given the issue we're trying to solve

config/eslint-plugin-ironfish/index.test.js
```diff
@@ -0,0 +1,28 @@
...
+    {
+      code: "import foo from '..'",
+      errors: [{ message: /specific/i }],
+    },
+    {
+      code: "import foo from '../..'",
+      errors: [{ message: /specific/i }],
+    },
+  ],
+});
```
brekk, line 28: 🙌🏼

brekk (approved):
👍🏼

dguenther (approved):

PR #1126, opened 2022-03-15
Convert Sleep task to serialized form

dguenther:
## Summary

Converts the Sleep task to the new serialization format.

## Testing Plan

Added unit tests for the Sleep task.

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/workerPool/tasks/workerMessage.ts
+++ b/ironfish/src/workerPool/tasks/workerMessage.ts
@@ -8,8 +8,9 @@ import { Serializable } from '../../common/serializable'
 export enum WorkerMessageType {
   CreateMinersFee = 0,
   JobError = 1,
-  SubmitTelemetry = 2,
-  VerifyTransaction = 3,
+  Sleep = 2,
+  SubmitTelemetry = 3,
+  VerifyTransaction = 4,
 }
 
 export abstract class WorkerMessage implements Serializable {
--- a/ironfish/src/workerPool/tasks/workerTask.ts
+++ b/ironfish/src/workerPool/tasks/workerTask.ts
@@ -1,8 +1,9 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { WorkerMessage } from './workerMessage'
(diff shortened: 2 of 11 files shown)


ironfish/src/workerPool/tasks/workerTask.ts
```diff
@@ -1,8 +1,9 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { WorkerMessage } from './workerMessage'
+import type { Job } from '../job'
+import type { WorkerMessage } from './workerMessage'
 
 export abstract class WorkerTask {
-  abstract execute(request: WorkerMessage): WorkerMessage | Promise<WorkerMessage>
+  abstract execute(request: WorkerMessage, job: Job): WorkerMessage | Promise<WorkerMessage>
```
rohanjadvani, line 8: I think `job` might need to be optional

ironfish/src/workerPool/tasks/workerMessage.ts
```diff
@@ -9,6 +9,7 @@ export enum WorkerMessageType {
   JobError = 0,
   SubmitTelemetry = 1,
   VerifyTransaction = 2,
+  Sleep = 3,
```
rohanjadvani, line 12: nit: Can this be alphabetized?

ironfish/src/workerPool/tasks/workerTask.ts
dguenther: I think we want it to be required because the caller will always pass in a `Job` when calling `execute`. However, classes that extend `WorkerTask` don't necessarily have to add `job: Job` to the parameters of execute if it's unused because of this: https://github.com/Microsoft/TypeScript/wiki/FAQ#why-are-functions-with-fewer-parameters-assignable-to-functions-that-take-more-parameters

ironfish/src/workerPool/tasks/workerMessage.ts
dguenther: Yep, will do

dguenther: Oh, this is actually done now.

rohanjadvani (approved):

PR #1139, opened 2022-03-16
Move CreateTransaction to serializable worker message/task

deekerno:
## Summary
This PR moves the CreateTransaction task into the new serialization format.

## Testing Plan
Added test, and ran the node to ensure proper function.

## Breaking Change

```
[ ] Yes
[x] No
```


ironfish/src/workerPool/tasks/createTransaction.test.slow.ts
```diff
@@ -0,0 +1,147 @@
...
+  })
+  const nodeTest = createNodeTest()
+
+  describe('execute', () => {
+    it('creates the transaction', async () => {
+      const account = await useAccountFixture(nodeTest.accounts)
+      const minerTransaction = await useMinersTxFixture(nodeTest.accounts, account)
+
+      const spendNote = minerTransaction.getNote(0).decryptNoteForOwner(account.incomingViewKey)
+      if (spendNote === undefined) {
```
rohanjadvani, line 109:
```suggestion
Assert.isNotUndefined(spendNote)
```

ironfish/src/workerPool/tasks/createTransaction.test.slow.ts
```diff
@@ -0,0 +1,147 @@
...
+        throw new Error('spendNote should not be undefined')
+      }
+      for (let i = 0; i < minerTransaction.notesLength(); i++) {
+        const note = minerTransaction.getNote(i)
+        await tree.add(note)
+      }
+
+      const authPath = (await tree.witness(0))?.authenticationPath
+      if (authPath === undefined) {
```
rohanjadvani, line 118:
```suggestion
      Assert.isNotUndefined(authPath)
```

ironfish/src/workerPool/tasks/createTransaction.test.slow.ts
```diff
@@ -0,0 +1,147 @@
...
+            authPath,
+          },
+        ],
+        [{ publicAddress: account.publicAddress, amount: BigInt(1), memo: '' }],
+      )
+
+      const response = task.execute(request)
+      expect(response).toEqual(
+        new CreateTransactionResponse(
+          Buffer.from(response.serializedTransactionPosted),
```
rohanjadvani, line 141: We should make the assertion more strict. Right now the test is reconstructing a response object from a field the response already has, so the test is covering the object itself versus the actual execution logic. 

To have a more explicit assertion, we can (1) mock what `transaction.post` returns and assert that is in the response or (2) reconstruct the serialized transaction and verify that's what is returned. (1) is probably easier, and we can also verify `spend` and `receive` were called. 

There is an example for `createMinersFee` [here](https://github.com/iron-fish/ironfish/blob/serialize-worker-pool/ironfish/src/workerPool/tasks/createMinersFee.test.ts#L57-L60).

ironfish/src/workerPool/tasks/createTransaction.ts
```diff
@@ -24,39 +28,210 @@ export type CreateTransactionRequest = {
...
+
+  serialize(): Buffer {
+    const bw = bufio.write(this.getSize())
+    bw.writeVarString(this.spendKey)
+    bw.writeVarBytes(BigIntUtils.toBytesBE(this.transactionFee))
+    bw.writeU64(this.expirationSequence)
+    bw.writeU64(this.spends.length)
+
+    for (const spend of this.spends) {
+      bw.writeVarBytes(Buffer.from(spend.note))
```
rohanjadvani, line 62: Does this work?

```suggestion
      bw.writeVarBytes(spend.note)
```

ironfish/src/workerPool/tasks/createTransaction.ts
```diff
@@ -24,39 +28,210 @@ export type CreateTransactionRequest = {
...
+
+      for (const step of spend.authPath) {
+        switch (step.side) {
+          case Side.Left:
+            bw.writeU8(0)
+            break
+          case Side.Right:
+            bw.writeU8(1)
+            break
+          default:
```
rohanjadvani, line 75: Do we need this `default` branch?

rohanjadvani (changes requested):

ironfish/src/workerPool/tasks/createTransaction.ts
deekerno: It seems that the tests in `accounts.test.slow.ts` fail if `spend.note` is not explicitly a `Buffer`.

ironfish/src/workerPool/tasks/createTransaction.ts
deekerno: I placed it there to make sure to mirror some of the other switch-case uses in the code, but removing it has no ill effects.

rohanjadvani (approved):

PR #1146, opened 2022-03-17
Temp patch for node-ipc malware (CVE-2022-23812)

fkjadoon94:
## Summary

node-ipc 9.2.2 added the peacenotwar module, which performs an unauthorized file write on users' filesystem. Overriding temporarily for hotfix. 

## Testing Plan

TBD

## Breaking Change

```
[X ] Yes
[ ] No
```
Dep update could cause breaking changes, needs testing.

### DIFF
--- a/ironfish/package.json
+++ b/ironfish/package.json
@@ -87,6 +87,10 @@
   "resolutions": {
     "highlight.js": "10.4.1"
   },
+  "overrides": {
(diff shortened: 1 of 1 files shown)


dguenther:
Just to double-check, I believe this doesn't affect the `ironfish` package since the version is fixed at `9.1.3`?

PR #1163, opened 2022-03-18
Remove unused Note and Nullifier messages

dguenther:
## Summary

The note and nullifier messages aren't used anywhere, so we can remove them.

Fixes IRO-1688
Fixes IRO-1691

## Testing Plan

Tests should pass.

## Breaking Change

```
[ ] Yes
[X] No
```


deekerno (approved):
Looks good to me; there's a small lint error preventing the suite from passing.

PR #1168, opened 2022-03-23
feat(ironfish): Expand `parseMessage` to take `string | Buffer`

rohanjadvani:
## Summary

Expand the type of `parseMessage` to accept a `Buffer`. A lot of utility and helper methods throw an Error if buffers are passed. This is going to an intermediate branch and is unused, so it will be a temporary type guard for now.

## Testing Plan

N/A, will be tested as messages are converted over

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/rpc/routes/peers/getPeerMessages.ts
+++ b/ironfish/src/rpc/routes/peers/getPeerMessages.ts
@@ -16,6 +16,7 @@ type PeerMessage = {
         type: string
         payload: Record<string, unknown>
       }
+    | Buffer
   timestamp: number
   type: Connection['type']
 }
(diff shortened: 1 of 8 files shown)


ironfish/src/rpc/routes/peers/getPeerMessages.ts
```diff
@@ -16,6 +16,7 @@ type PeerMessage = {
         type: string
         payload: Record<string, unknown>
       }
+    | Buffer
```
dguenther, line 19: The `yup` validator in this file might have to be updated to accept some form of `Buffer` -- I'm a little surprised TS didn't complain about needing to update the `yup` validation, interesting

dguenther (approved):
I think this looks good to get it up and running, and we can revisit the specifics of implementing some of these once we get the first Buffer-based message working. (e.g. removing the `Error`s, and checking the message type when pushing to `loggedMessages`)

ironfish/src/rpc/routes/peers/getPeerMessages.ts
rohanjadvani: It seems like the schema [isn't used](https://github.com/iron-fish/ironfish/blob/master/ironfish/src/rpc/routes/peers/getPeerMessages.ts#L62). Will take a look in a subsequent PR.

PR #1169, opened 2022-03-23
feat(ironfish): Create abstract class for serializable network messages

rohanjadvani:
## Summary

Create new abstract class for network message conversion.

## Testing Plan

N/A

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/network/messages/networkMessage.ts
+++ b/ironfish/src/network/messages/networkMessage.ts
@@ -1,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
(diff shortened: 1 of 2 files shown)


deekerno (approved):

ironfish/src/network/messages/networkMessage.ts
```diff
@@ -0,0 +1,21 @@
...
+export enum NetworkMessageType {}
+
+export abstract class NetworkMessage implements Serializable {
+  private static id = 0
+
+  readonly networkId: number
+  readonly type: NetworkMessageType
+
+  constructor(type: NetworkMessageType, networkId?: number) {
+    this.networkId = networkId ?? NetworkMessage.id++
```
dguenther, line 15: Oh yeah, I hadn't thought of this before, but we probably won't end up needing message IDs on every message, just the RPCs. We can fix this at any time though, it's just getting the structure in place

dguenther (approved):

PR #1178, opened 2022-03-26
Properly dispose of WebRTC connections

mat-if:
## Summary

Noticed that my node was not properly disposing of some peers. This is the 'quick-fix', putting up here so I don't forget about it later, but the actual solution probably needs a little more thought.

The reproduction steps:
1. Connect via Websocket
2. Send Identity message to complete the initial Websocket "handshake"
3. Disconnect without ever upograding to WebRTC connection.

The WebRTC connection will retry until Infinity, but because tryDisposePeer only checks Websocket, it will stay forever. This is also a potential memory leak.

## Breaking Change

```
[ ] Yes
[ ] No
```


mat-if:
Documented in a larger doc around peer networking, going to close this since it has served it's purpose

PR #1181, opened 2022-03-28
RPC/CLI to watch transactions with a view key

mat-if:
## Summary

image

## Breaking Change

```
[ ] Yes
[ ] No
```


mat-if (commented):
lgtm

NullSoldier (approved):

PR #1189, opened 2022-03-31
Add index for mined blocks

deekerno:
## Summary
This adds an indexer for mined blocks. Adjustments to the `miners:mined` CLI command are coming in a separate PR.

## Testing Plan
Added tests and ensured that node runs as expected.

## Breaking Change

```
[ ] Yes
[x] No
```


ironfish/src/node.ts
```diff
@@ -284,6 +304,7 @@ export class IronfishNode {
     }
 
     this.telemetry.submitNodeStarted()
+    this.minedBlocksIndexer.start()
```
NullSoldier, line 307: The telemetry is wrong now based on this ordering right?

ironfish/src/indexer/indexer.ts
```diff
@@ -0,0 +1,32 @@
...
+  constructor({ files, location }: { files: FileSystem; location: string }) {
+    this.files = files
+    this.location = location
+    this.database = createDB({ location })
+  }
+
+  async openDB(options: { upgrade?: boolean } = { upgrade: true }): Promise<void> {
+    await this.files.mkdir(this.location, { recursive: true })
+    await this.database.open()
+    if (options.upgrade) {
```
NullSoldier, line 24: nit: extra space here? :)
```suggestion

    if (options.upgrade) {
```

ironfish/src/indexer/minedBlocksIndexer.ts
```diff
@@ -0,0 +1,245 @@
...
+      valueEncoding: new JsonEncoding(),
+    })
+
+    this.chainProcessor.onAdd.on(async (header) => {
+      const block = await chain.getBlock(header)
+      Assert.isNotNull(block)
+
+      const account = this.accounts.listAccounts().find((a) => isBlockMine(block, a))
+      if (account) {
+        const main = await this.chain.isHeadChain(header)
```
NullSoldier, line 88: this should always be true here

ironfish/src/indexer/minedBlocksIndexer.ts
```diff
@@ -0,0 +1,245 @@
...
+        const minersFee = this.chain.strategy.miningReward(header.sequence)
```
NullSoldier, line 89: You don't need to recalculate this, `header.minersFee`

ironfish/src/indexer/minedBlocksIndexer.ts
```diff
@@ -0,0 +1,245 @@
...
+    })
+
+    this.chainProcessor.onRemove.on(async (header) => {
+      if (await this.minedBlocks.has(header.hash.toString('hex'))) {
+        const block = await chain.getBlock(header)
+        Assert.isNotNull(block)
+
+        const account = this.accounts.listAccounts().find((a) => isBlockMine(block, a))
+        if (account) {
+          const main = await this.chain.isHeadChain(header)
```
NullSoldier, line 104: this should always be false

ironfish/src/indexer/minedBlocksIndexer.ts
```diff
@@ -0,0 +1,245 @@
...
+          const minersFee = this.chain.strategy.miningReward(header.sequence)
+
+          await this.put(header, account.name, main, minersFee)
```
NullSoldier, line 107: this is already stored, you just need to edit the existing one right?

ironfish/src/indexer/minedBlocksIndexer.ts
```diff
@@ -0,0 +1,245 @@
...
+        }
+      }
+      await this.updateHeadHash(header.previousBlockHash)
```
NullSoldier, line 110: what's the implication of not doing all this inside of a transaction?

ironfish/src/indexer/indexer.ts
```diff
@@ -0,0 +1,32 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { FileSystem } from '../fileSystems'
+import { IDatabase } from '../storage/database'
+import { createDB } from '../storage/utils'
+
+const DATABASE_VERSION = 1
+
+export abstract class Indexer {
```
NullSoldier, line 10: This is called Indexer, but what logic here is related to indexing? Isn't that all in MinedBlocksIndexer? Most of the abstraction isn't in the indexer.

ironfish/src/indexer/minedBlocksIndexer.ts
deekerno: Good catch, thanks!

ironfish/src/indexer/minedBlocksIndexer.ts
deekerno: Good catch, thanks!

ironfish/src/node.ts
deekerno: Good catch, thanks!

ironfish/src/indexer/minedBlocksIndexer.ts
deekerno: Changes would not be atomic, which could lead to non-linearizability. Adding the transaction.

ironfish/src/account/accounts.ts
```diff
@@ -32,6 +32,8 @@ export class Accounts {
     [account: Account | null, oldAccount: Account | null]
   >()
 
+  readonly onAccountCreated = new Event<[account: Account, imported: boolean]>()
```
NullSoldier, line 35: if something is imported, we're not creating an account, we're importing it. I think the bool parameter and the name of this are mutually exclusive.

ironfish/src/account/accounts.ts
```diff
@@ -957,6 +963,8 @@ export class Accounts {
 
     this.accounts.delete(name)
     await this.db.removeAccount(name)
+
+    await this.onAccountRemoved.emitAsync(name)
```
NullSoldier, line 967: Do we need async properties for these events? Should the accounts need to wait for the receiver to finish the event callback?

ironfish/src/indexer/minedBlocksIndexer.ts
```diff
@@ -0,0 +1,268 @@
...
+    })
+
+    this.chainProcessor = new ChainProcessor({ logger, chain, head: null })
+
+    this.chainProcessor.onAdd.on(async (header) => {
+      const block = await this.chain.getBlock(header)
+      Assert.isNotNull(block)
+
+      const account = this.accounts.listAccounts().find((a) => isBlockMine(block, a))
+      await this.database.transaction(async (tx) => {
```
NullSoldier, line 98: I think there's a few things here worth thinking through

1. It's possible to start a transaction here that does nothing in the current code
2. There's a data consistency issue with this transaction.

ironfish/src/indexer/minedBlocksIndexer.ts
```diff
@@ -0,0 +1,268 @@
...
+              hash: header.hash.toString('hex'),
+              sequence: header.sequence,
+              account: account.name,
+              minersFee: Number(header.minersFee),
+            },
+            tx,
+          )
+          const hashes: string[] = (await this.accountToBlockHashes.get(account.name, tx)) || []
+          hashes.concat(block.header.hash.toString('hex'))
+          await this.accountToBlockHashes.put(account.name, hashes, tx)
```
NullSoldier, line 113: this will store the entire first table into a value of row in the second table.

ironfish/src/indexer/minedBlocksIndexer.ts
```diff
@@ -0,0 +1,268 @@
...
+      Assert.isNotNull(block)
+
+      const account = this.accounts.listAccounts().find((a) => isBlockMine(block, a))
+      await this.database.transaction(async (tx) => {
+        if (account) {
+          await this.minedBlocks.put(
+            header.hash.toString('hex'),
+            {
+              main: true,
+              hash: header.hash.toString('hex'),
```
NullSoldier, line 104: We're double storing the hash here.

ironfish/src/indexer/minedBlocksIndexer.ts
NullSoldier: I'm unresolving, because a transaction wasn't added.

ironfish/src/indexer/minedBlocksIndexer.ts
```diff
@@ -0,0 +1,268 @@
...
+    })
+
+    this.accounts.onAccountCreated.on((_account, imported) => {
+      if (imported) {
+        this.chainProcessor.hash = null
+      }
+    })
+
+    this.accounts.onAccountRemoved.on(async (accountName) => {
+      await this.database.transaction(async (tx) => {
```
NullSoldier, line 145: Is this just used as a lock to avoid doing this while the chain processor is running?

ironfish/src/indexer/minedBlocksIndexer.ts
```diff
@@ -0,0 +1,268 @@
...
+    const minedBlocks = (await this.minedBlocks.getAllValues()).sort(
+      (a, b) => a.sequence - b.sequence,
+    )
+
+    if (includeForks) {
+      return minedBlocks
+    }
+
+    return minedBlocks.filter((block) => block.main === true)
+  }
```
NullSoldier, line 267: How do you filter on the sequence?

Also, is it possible for us to get away with just yielding blocks instead of accumulating and sorting all the blocks?

ironfish/src/account/accounts.ts
deekerno: Sure, I can replace this with `onAccountImported` since we don't really care about account creation in this indexer.

ironfish/src/indexer/minedBlocksIndexer.ts
deekerno: My intention was to filter in the RPC route and have this strictly return the desired blocks, but there's really no reason that it couldn't be done here by passing in options instead. Regarding your second question, since we're using hashes as the key for the store, `.getAllValues()` ends up being lexicographically sorted by hash. The current output of `miners:mined` is sorted by sequence.

ironfish/src/indexer/minedBlocksIndexer.ts
deekerno: Essentially, yes. I'm ensuring that it's an atomic change.

ironfish/src/indexers/minedBlocksIndexer.ts
```diff
@@ -0,0 +1,293 @@
...
+      .sort((a, b) => a.sequence - b.sequence)
+
+    if (!scanForks) {
+      minedBlocks = minedBlocks.filter((block) => block.main === true)
+    }
+
+    for (const block of minedBlocks) {
+      yield block
+    }
+  }
```
NullSoldier, line 292: We should make this actually able to iterate the results lazily, but to do that we need to avoid accumulating all the results. We can do that with an index by sequence.

ironfish/src/indexers/minedBlocksIndexer.ts
```diff
@@ -215,16 +199,36 @@ export class MinedBlocksIndexer {
...
   }
 
-  start(): void {
+  async start(): Promise<void> {
     if (this.isStarted) {
       return
     }
     this.isStarted = true
 
+    if (this.removalInProgress) {
```
NullSoldier, line 225: If you just remove this, I think it'll work the same. It's just an extra variable that's derived from `accountToRemove` anyway.

ironfish/src/indexers/minedBlocksIndexer.ts
```diff
@@ -215,16 +199,36 @@ export class MinedBlocksIndexer {
...
+      const accountName = await this.meta.get('accountToRemove')
+      if (accountName) {
+        await this.removeMinedBlocks(accountName)
+      }
+    }
```
NullSoldier, line 230: Code architecture wise, all this should be inside the event loop to avoid updating the head, and handling an account removal.

ironfish/src/indexers/minedBlocksIndexer.ts
```diff
@@ -166,14 +157,7 @@ export class MinedBlocksIndexer {
...
     this.accounts.onAccountRemoved.on(async (account) => {
-      await this.database.transaction(async (tx) => {
-        const hashes = await this.accountToBlockHashes.get(account.name, tx)
-        if (hashes) {
-          for (const hash of hashes) {
-            await this.minedBlocks.del(hash, tx)
-          }
-        }
-      })
+      await this.removeMinedBlocks(account.name)
```
NullSoldier, line 160: I think all this should be doing is queueing the "job" for removal, and letting the "scheduler" pick up that job. It ensures that no two things are running that could be conflicting. Like a prune job and updating the index.

ironfish/src/indexers/minedBlocksIndexer.ts
```diff
@@ -270,6 +274,37 @@ export class MinedBlocksIndexer {
     await this.meta.put('headHash', hashString, tx)
   }
 
+  async removeMinedBlocks(accountName: string): Promise<void> {
+    await this.meta.put('accountToRemove', accountName)
+
+    const hashes: Buffer[] = []
+    const sequences: number[] = []
+
+    await this.database.transaction(async (tx) => {
```
NullSoldier, line 283: We probably only need a transaction around the operations and not loading the entire database for pruning. Transactions cache all read state, so this will keep everything loaded in memory because of how this works. It also holds a really long lock.

ironfish/src/indexers/minedBlocksIndexer.ts
```diff
@@ -0,0 +1,344 @@
...
+    }
+  }
+
+  async updateHeadHash(headHash: Buffer | null, tx?: IDatabaseTransaction): Promise<void> {
+    const hashString = headHash && headHash.toString('hex')
+    await this.meta.put('headHash', hashString, tx)
+  }
+
+  async removeMinedBlocks(accountName: string): Promise<void> {
+    await this.meta.put('accountToRemove', accountName)
```
NullSoldier, line 271: I think this is left over

ironfish/src/indexers/minedBlocksIndexer.ts
```diff
@@ -272,7 +272,9 @@ export class MinedBlocksIndexer {
     for await (const [hash, block] of iterator) {
       if (block.account === accountName) {
         await this.database.transaction(async (tx) => {
-          await this.sequenceToHashes.del(block.sequence, tx)
+          let hashes = (await this.sequenceToHashes.get(block.sequence)) ?? []
```
NullSoldier, line 275: It looks like you're using null coalescence to default this to [] because it can be undefined, but how can it be undefined?

deekerno: That's a fair point. After looking through the code paths, it shouldn't ever be undefined if a block was added correctly. I think doing `Assert.isNotUndefined()` would be a good replacement.

ironfish/src/indexers/minedBlocksIndexer.ts
```diff
@@ -0,0 +1,369 @@
...
+    this.isStarted = false
+    this.rescan = false
+
+    this.meta = this.database.addStore<{
+      key: keyof MinedBlocksDBMeta
+      value: MinedBlocksDBMeta[keyof MinedBlocksDBMeta]
+    }>({
+      name: 'meta',
+      keyEncoding: new StringEncoding<keyof MinedBlocksDBMeta>(),
+      valueEncoding: new JsonEncoding(),
```
rohanjadvani, line 92: We now have binary encodings for all our stores in the node. Can you add encoders for each of these stores? You can look at the [PR](https://github.com/iron-fish/ironfish/pull/1256) here for examples. [Here](https://github.com/iron-fish/ironfish/blob/master/ironfish/src/account/database/accounts.ts) is how accounts are stored.

It might be useful to do these as separate PRs into this branch.

ironfish/src/indexers/database/accountsToRemove.ts
```diff
@@ -0,0 +1,44 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
rohanjadvani, line 1: Can you add unit tests for each of these encoders?

ironfish/src/indexers/database/meta.ts
```diff
@@ -0,0 +1,36 @@
...
+  headHash: string | null
+}
+
+export type MetaValue = MinedBlocksDBMeta[keyof MinedBlocksDBMeta]
+
+export class MetaValueEncoding implements IDatabaseEncoding<MetaValue> {
+  serialize(value: MetaValue): Buffer {
+    const bw = bufio.write(this.getSize(value))
+    if (value) {
+      bw.writeVarString(value, 'utf8')
```
rohanjadvani, line 17: Should this use `writeHash`?

ironfish/src/indexers/database/sequenceToHashes.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    while (reader.left()) {
+      hashes.push(reader.readHash())
+    }
+
+    return { hashes }
+  }
+
+  getSize(value: SequenceToHashesValue): number {
+    let size = 0
```
rohanjadvani, line 36:
```suggestion
    return 32 * value.hashes.length
```

ironfish/src/indexers/minedBlocksIndexer.ts
```diff
@@ -0,0 +1,395 @@
...
+} from './database/sequenceToHashes'
+
+const DATABASE_VERSION = 1
+const REMOVAL_KEY = 'accountsToRemove'
+
+const getMinedBlocksDBMetaDefaults = (): MinedBlocksDBMeta => ({
+  headHash: null,
+})
+
+type MinedBlocksDBMeta = {
```
rohanjadvani, line 39: This seems like a duplicate type now

ironfish/src/indexers/minedBlocksIndexer.ts
```diff
@@ -0,0 +1,395 @@
...
+      await this.accountsToRemove.put(REMOVAL_KEY, { accounts: accountsToRemove })
+    }
+
+    this.logger.debug(`Finished removing mined blocks for account ${accountName}`)
+  }
+
+  async getMinedBlock(
+    blockHash: Buffer,
+  ): Promise<
+    | { main: boolean; sequence: number; account: string; minersFee: number; hash: string }
```
rohanjadvani, line 342: Does this work?

```suggestion
    MinedBlockValue | undefined
```

ironfish/src/indexers/minedBlocksIndexer.ts
```diff
@@ -0,0 +1,395 @@
...
+  async *getMinedBlocks({
+    scanForks,
+    start,
+    stop,
+  }: {
+    scanForks?: boolean
+    start?: number
+    stop?: number
+  }): AsyncGenerator<
+    { main: boolean; sequence: number; account: string; minersFee: number; hash: string },
```
rohanjadvani, line 359:
```suggestion
    MinedBlockValue
```

ironfish/src/indexers/minedBlocksIndexer.ts
```diff
@@ -0,0 +1,395 @@
...
+      name: 'blocks',
+      keyEncoding: new BufferEncoding(),
+      valueEncoding: new MinedBlockValueEncoding(),
+    })
+
+    this.sequenceToHashes = this.database.addStore<{
+      key: number
+      value: SequenceToHashesValue
+    }>({
+      name: 'seqToHash',
```
rohanjadvani, line 106:
```suggestion
      name: 'sequenceToHashes',
```

ironfish/src/indexers/minedBlocksIndexer.ts
```diff
@@ -0,0 +1,395 @@
...
+      key: keyof MinedBlocksDBMeta
+      value: MinedBlocksDBMeta[keyof MinedBlocksDBMeta]
+    }>({
+      name: 'meta',
+      keyEncoding: new StringEncoding<keyof MinedBlocksDBMeta>(),
+      valueEncoding: new MetaValueEncoding(),
+    })
+
+    this.minedBlocks = this.database.addStore<{ key: Buffer; value: MinedBlockValue }>({
+      name: 'blocks',
```
rohanjadvani, line 97:
```suggestion
      name: 'minedBlocks',
```

ironfish/src/indexers/minedBlocksIndexer.ts
```diff
@@ -0,0 +1,395 @@
...
+      name: 'seqToHash',
+      keyEncoding: new U32Encoding(),
+      valueEncoding: new SequenceToHashesValueEncoding(),
+    })
+
+    this.accountsToRemove = this.database.addStore<{
+      key: string
+      value: AccountsToRemoveValue
+    }>({
+      name: 'accsToRemove',
```
rohanjadvani, line 115:
```suggestion
      name: 'accountsToRemove',
```

rohanjadvani (commented):
Let's be explicit with the names. We don't need to shorten these.

deekerno:
> Let's be explicit with the names. We don't need to shorten these.

@NullSoldier recommended keeping the names short due to how LevelDB represents rows under the hood. Happy to change it either way.

ironfish/src/indexers/database/meta.ts
deekerno: It should!

ironfish/src/indexers/database/accountsToRemove.test.ts
```diff
@@ -0,0 +1,34 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { AccountsToRemoveValue, AccountsToRemoveValueEncoding } from './accountsToRemove'
+
+describe('AccountsToRemoveValueEncoding', () => {
+  describe('with an empty list', () => {
```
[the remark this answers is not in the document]
deekerno: 👍🏽

ironfish/src/indexers/minedBlocksIndexer.ts
```diff
@@ -0,0 +1,391 @@
...
+    const { hashChanged } = await this.chainProcessor.update()
+
+    if (hashChanged) {
+      this.logger.debug(
+        `Updated MinedBlocksIndexer Head: ${String(this.chainProcessor.hash?.toString('hex'))}`,
+      )
+    }
+  }
+
+  async getHashesAtSequence(sequence: number, tx?: IDatabaseTransaction): Promise<Buffer[]> {
```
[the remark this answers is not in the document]
deekerno: Oh yes, it should, good catch.

ironfish/src/indexers/database/sequenceToHashes.test.ts
```diff
@@ -0,0 +1,23 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import {
+  SequenceToHashesValue,
+  SequenceToHashesValueEncoding,
+} from '../../blockchain/database/sequenceToHashes'
+
+describe('SequenceToHashesValueEncoding', () => {
+  describe('with a list of hashes', () => {
```
rohanjadvani, line 10: Same comment here.

ironfish/src/indexers/minedBlocksIndexer.ts
```diff
@@ -0,0 +1,383 @@
...
+
+          await this.updateHeadHash(header.hash, tx)
+        })
+      }
+    })
+
+    this.chainProcessor.onRemove.on(async (header) => {
+      await this.database.transaction(async (tx) => {
+        if (await this.minedBlocks.has(header.hash, tx)) {
+          const block = await this.chain.getBlock(header)
```
rohanjadvani, line 148:
```suggestion
          const block = await this.chain.getBlock(header, tx)
```

rohanjadvani (approved):

PR #1190, opened 2022-03-31
Improve json parse error checking

mat-if:
## Summary

- Update parse-json to 5.2.0 which exports JSONError, allowing us to use `instanceof` to type check it
- Add some basic unit test coverage

## Testing Plan

- Unit tests
- Manually tested: `rm hosts.json`, `touch hosts.json`, also tried `echo asdf > hosts.json`

## Breaking Change

```
[ ] Yes
[x] No
```


deekerno (approved):
Nice find!

PR #1196, opened 2022-04-05
STAGING -> MASTER

mat-if:
## Summary

Should be able to just slide through without needing any version changes since it's just a build change and the npm builds were never released for last release

## Breaking Change

```
[ ] Yes
[ ] No
```


brekk (approved):
:shipit:

PR #1199, opened 2022-04-05
Refactor default dataDir in node and sdk

danield9tqh:
## Summary
Doing some refactoring and moving the default dataDir up to the level of the node and sdk. We don't want to set defaults deep in the code

## Testing Plan
Running current test suite

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/network/testUtilities/mockHostsStore.ts
+++ b/ironfish/src/network/testUtilities/mockHostsStore.ts
@@ -51,7 +51,7 @@ class MockFileSystem extends FileSystem {
 
 class MockHostsStore extends HostsStore {
   constructor() {
-    super(new MockFileSystem())
+    super(new MockFileSystem(), "~/.ironfish")
     super.set('priorPeers', [
       {
         address: '127.0.0.1',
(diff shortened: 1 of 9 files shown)


ironfish/src/network/testUtilities/mockHostsStore.ts
```diff
@@ -51,7 +51,7 @@ class MockFileSystem extends FileSystem {
 
 class MockHostsStore extends HostsStore {
   constructor() {
-    super(new MockFileSystem())
+    super(new MockFileSystem(), "~/.ironfish")
```
NullSoldier, line 54: Small nit, not mandatory but we could just pass DEFAULT_DATA_DIR here from config.ts

ironfish/src/node.ts
```diff
@@ -154,17 +154,18 @@ export class IronfishNode {
   }): Promise<IronfishNode> {
     logger = logger.withTag('ironfishnode')
 
+    const dataDirOrDefault = dataDir || DEFAULT_DATA_DIR
```
NullSoldier, line 157: I think this is my preference but I think this is simpler to just default the variable to its default value in the case that you want to set a default. We actually do use that pattern too in a lot of places, instead of creating a new variable. It also makes the diff a little simpler. What do you think?

NullSoldier (approved):

NullSoldier:
Looks great! LGTM.

PR #1202, opened 2022-04-06
Add graffiti to telemetry API payload

mat-if:
## Summary

Note that this isn't adding graffiti to the telemetry payload itself, just to the API request. Will need accompanying API PR

## Breaking Change

```
[ ] Yes
[ ] No
```

### DIFF
--- a/ironfish/src/node.ts
+++ b/ironfish/src/node.ts
@@ -205,7 +205,7 @@ export class IronfishNode {
     const telemetry = new Telemetry({
       chain,
       logger,
-      graffiti: config.get('blockGraffiti'),
+      config,
       metrics,
       workerPool,
       defaultTags: [{ name: 'version', value: pkg.version }],
--- a/ironfish/src/node.ts
+++ b/ironfish/src/node.ts
@@ -205,6 +205,7 @@ export class IronfishNode {
     const telemetry = new Telemetry({
       chain,
       logger,
+      graffiti: config.get('blockGraffiti'),
       metrics,
       workerPool,
       defaultTags: [{ name: 'version', value: pkg.version }],
--- a/ironfish/src/telemetry/telemetry.ts
+++ b/ironfish/src/telemetry/telemetry.ts
@@ -7,7 +7,7 @@ import { Config } from '../fileStores/config'
 import { createRootLogger, Logger } from '../logger'
 import { MetricsMonitor } from '../metrics'
 import { Block } from '../primitives/block'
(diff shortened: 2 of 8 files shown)


ironfish/src/node.ts
```diff
@@ -205,6 +205,7 @@ export class IronfishNode {
     const telemetry = new Telemetry({
       chain,
       logger,
+      graffiti: config.get('blockGraffiti'),
```
NullSoldier, line 208: Should we pass in graffiti and get this adhoc incase they change their graffiti while the node is running? Otherwise they'll have to restart the node.

rohanjadvani: Agreed with @NullSoldier

ironfish/src/telemetry/telemetry.ts
```diff
@@ -35,13 +36,15 @@ export class Telemetry {
...
     workerPool: WorkerPool
+    graffiti?: string
     logger?: Logger
     metrics?: MetricsMonitor
     defaultFields?: Field[]
     defaultTags?: Tag[]
   }) {
     this.chain = options.chain
     this.workerPool = options.workerPool
+    this.graffiti = options.graffiti ?? ''
```
rohanjadvani, line 47: I think most people will have this enabled, but we could leave this as optional and conditionally check for assignment in the API.

mat-if: I thought about that, but I didn't see any examples of optional params in the worker pool message serializers so wasn't sure if that was going to be a lot of extra work for temporary code

rohanjadvani: Makes sense. [This](https://github.com/iron-fish/ironfish/blob/91fe59d1bf1981ed224ef5ea4465dd35ee386ba8/ironfish/src/network/messages/disconnecting.ts#L40-L66) is how I am doing it for network serialization if you'd want to do that here - either way.

mat-if: Ah perfect, thanks, I'll follow that model

mat-if: This is _kind_ of moot now that we pass in the config here instead of graffiti, but just to make sure: I had this technically kind of wrong, since `config.get('blockGraffiti')` will always return a string, `''` is the default value

NullSoldier (approved):

rohanjadvani (approved):

PR #1207, opened 2022-04-07
remove node from request object

danield9tqh:
## Summary
Node is not being used on the request object so simplifying this code. It seems like the request being sent to a node should not already know about the node it's being sent to

## Breaking Change

```
[ ] Yes
[X] No
```


mat-if (approved):

dguenther (approved):
looks good!

PR #1208, opened 2022-04-09
Allow for RPC response on different connection

dguenther:
## Summary

Switches the RPC router to store peers on requests rather than connections, allowing responses to be sent over a different connection (for example, when switching from a WebSocket to WebRTC connection).

Also fixes an issue with RPC IDs incrementing in the RPC router.

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/network/messageRouters/rpc.ts
+++ b/ironfish/src/network/messageRouters/rpc.ts
@@ -3,15 +3,13 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 import { createRootLogger, Logger } from '../../logger'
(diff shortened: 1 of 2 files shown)


ironfish/src/network/messageRouters/rpc.ts
```diff
@@ -142,12 +134,13 @@ export class RpcRouter {
           clearTimeout(timeout)
           reject(reason)
         },
+        peer: peer,
```
rohanjadvani, line 137:
```suggestion
        peer,
```

rohanjadvani (approved):

PR #1210, opened 2022-04-11
feat(ironfish): Yeet `getSize` from serializable interface

rohanjadvani:
## Summary

Remove the size from the network message header

## Testing Plan

Existing unit tests and ran node

## Breaking Change

```
[ ] Yes
[x] No
```


dguenther (changes requested):
We can probably drop it from being included on the message buffers, but still keep it on the message class so we can preallocate buffers.

dguenther (approved):

PR #1211, opened 2022-04-11
Remove node from ipc adapter

danield9tqh:
## Summary
Code cleanup, removing unused `node` field in the `IpcAdapter`. `IpcAdapter` accesses the node through the `Server` and should not need to know about the node directly. It is also not being used anywhere currently. Also adding stricter type validation to the `renderError` function since it is always called with an `Error` object

## Testing Plan
Current testing suite

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/rpc/adapters/ipcAdapter.ts
+++ b/ironfish/src/rpc/adapters/ipcAdapter.ts
@@ -7,7 +7,6 @@ import { v4 as uuid } from 'uuid'
 import * as yup from 'yup'
 import { Assert } from '../../assert'
 import { createRootLogger, Logger } from '../../logger'
-import { IronfishNode } from '../../node'
 import { YupUtils } from '../../utils/yup'
 import { Request } from '../request'
 import { ApiNamespace, Router } from '../routes'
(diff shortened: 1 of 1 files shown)


ironfish/src/rpc/adapters/ipcAdapter.ts
```diff
@@ -280,7 +274,7 @@ export class IpcAdapter implements IAdapter {
     this.server.emit(socket, 'stream', { id: messageId, data: data })
   }
 
-  renderError(error: unknown): IpcError {
+  renderError(error: Error): IpcError {
     let message = 'An error has occured'
     let stack = undefined
     let code: string = ERROR_CODES.ERROR
```
dguenther, line 280: I think the `error instanceof Error` check below can be removed, since now the types should guarantee an error is coming in. Other than that, looks good!

dguenther (approved):

ironfish/src/rpc/adapters/ipcAdapter.ts
danield9tqh: 👍 thanks, updated

PR #1214, opened 2022-04-12
Replace node-ipc server native node TCP functionality

danield9tqh:
## Summary
The node module `node-ipc` is a simple wrapper around native node `net` code. It's possible to replace this library with native TCP sockets from the `net`  lib. This PR is to replace the `node-ipc` server functionality but still expects requests to be formatted as if they are from `node-ipc` clients. Later we will replace the `node-ipc` client as well

## Testing Plan
Existing testing suite

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/rpc/adapters/tcpAdapter.ts
+++ b/ironfish/src/rpc/adapters/tcpAdapter.ts
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import net from 'net'
+import { v4 as uuid } from 'uuid'
 import * as yup from 'yup'
 import { createRootLogger, Logger } from '../../logger'
 import { YupUtils } from '../../utils/yup'
@@ -44,6 +45,7 @@ export class TcpAdapter implements IAdapter {
   server: net.Server | null = null
   router: Router | null = null
   namespaces: ApiNamespace[]
+  pending = new Map<string, { sock: net.Socket; reqs: Map<string, Request> }>()
 
   constructor(
     host: string,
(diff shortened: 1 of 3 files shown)


ironfish/src/rpc/adapters/tcpAdapter.ts
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  async onClientData(socket: net.Socket, data: Buffer): Promise<void> {
+    const dataString = data.toString('utf8').trim()
+    const result = await YupUtils.tryValidate(IncomingNodeIpcSchema, dataString)
+
+    if (result.error) {
+      this.logger.log(
+        `${this.host}:${this.port} Error parsing message : ${result.error.message} raw: ${dataString} `,
+      )
```
NullSoldier, line 115: I know were getting heavily into preference, but if there's one style thing I could advocate for our entire code base that I've adopted from bcoin that I think is very good is not nesting things, and bailing on error cases when there is no more execution flow. So return in this if statement, and unindent the rest of the code.

ironfish/src/rpc/adapters/tcpAdapter.ts
```diff
@@ -0,0 +1,207 @@
...
+          socket.write(res)
+        },
+        (data: unknown) => {
+          const res = this.encodeNodeIpc(this.constructStream(message.mid, data))
+          socket.write(res)
+        },
+      )
+
+      if (this.router == null) {
+        this.logger.log('handling incoming connection on unmounted adapter')
```
NullSoldier, line 132: same feedback here, because for every error case we have to add an indent, and also double check there's no more code after the else that will run here.

danield9tqh: Yea that makes sense. Just in case any code gets added on accidentally after 👍 . Fixed

NullSoldier (approved):

PR #1215, opened 2022-04-12
feat(ironfish): Add binary serialization to the networking layer

rohanjadvani:
## Summary

Migrate messages to use binary serialization when communicating with nodes / peers.

Aside from refactoring out the message routers, we specifically tried to avoid too much refactoring in this PR -- since it touches a lot of networking code, there are a lot of opportunity for changes, but the primary goal is to convert the message serialization to binary.

### Notable changes

* `disconnecting` message times are now ceiling-d to the next second
* Removed the router-based networking code (including `registerHandler`) and moved the still-used code into `peerNetwork.ts`
* `connection.ts` handles parsing the message type (and gossip nonce, RPC ID if applicable), then based on that type, instantiates one of the message classes in `ironfish/src/network/messages`.
* Added more bufio types
* Added a little-endian bigint conversion utility
* Commented on future work around partial header serialization

## Testing Plan

Added unit tests and manually ran node.

## Breaking Change

**_This branch requires a networking version bump, which will require all nodes to upgrade to the new version._**

```
[x] Yes
[] No
```


ironfish/src/network/messages/identify.ts
```diff
@@ -0,0 +1,98 @@
...
+  serialize(): Buffer {
+    const bw = bufio.write(this.getSize())
+    bw.writeBytes(Buffer.from(this.identity, 'base64'))
+    bw.writeVarString(this.name)
+    bw.writeU16(this.port)
+    bw.writeU16(this.version)
+    bw.writeVarString(this.agent)
+    bw.writeU32(this.sequence)
+    bw.writeHash(this.head)
+    bw.writeVarBytes(BigIntUtils.toBytesLE(this.work))
```
dguenther, line 60: This could probably be constrained -- it's a function of the target, I believe, but it'll likely be smaller than the full 256 bits of the target, so I think leaving it as a variable-length field isn't a huge deal.

dguenther (commented):
This should be ready for review now

mat-if (approved):

ironfish/src/network/peers/connections/connection.ts
```diff
@@ -189,11 +207,99 @@ export abstract class Connection {
...
+
+  private isGossipNetworkMessageType(type: NetworkMessageType): boolean {
+    return [NetworkMessageType.NewBlock, NetworkMessageType.NewTransaction].includes(type)
+  }
+
+  private parseBody({ rpcId, nonce, type, body }: NetworkMessageHeader): NetworkMessage {
+    switch (type) {
+      case NetworkMessageType.CannotSatisfyRequest:
+        Assert.isNotUndefined(rpcId)
+        return CannotSatisfyRequest.deserialize(rpcId)
```
danield9tqh, line 272: Not a blocker but something to think about. Should we match up the wrapping of serialize and deserialize functions a little better? Currently it looks like for serialize on messages that extend from multiple classes we are wrapping the serialize functionality

`NetworkMessage.serialize(RpcNetworkMessage.serialize(CannotSatisfyRequest.serialize()))` 

But for deserialize we just have a flat structure

`CannotSatisfyRequest.deserialize(rpcId)`

`rpcId` really would make more sense in the deserialize function of the `RpcNetworkMessage` and then we wouldn't have to do these Assert statements here as well

danield9tqh: Similar for `nonce` and `GossipNetworkMessage`

ironfish/src/network/messages/disconnecting.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+    return new DisconnectingMessage({
+      destinationIdentity,
+      disconnectUntil,
+      reason,
+      sourceIdentity,
+    })
+  }
+
+  getSize(): number {
```
danield9tqh, line 70: Is `size` the first thing serialized on every Serializable object? It does seems a little error prone to calculate the size of the object separately from the serialization logic. If they get out of sync (like we add a field to `serialize` but forget to add it to `getSize`). Is there a way we combine this logic; maybe get the size of the buffer in `NetworkMessage` and use that as the size?

rohanjadvani: We also use the size when allocating the buffer [here](https://github.com/iron-fish/ironfish/pull/1215/files#diff-933783bbd91ba8f941400e2ce015b750ea4359a65af898ab012e535110e4567cR43)

ironfish/src/network/peers/connections/connection.ts
dguenther: The wrapping described isn't quite correct -- `RpcNetworkMessage.serializeWithMetadata` overrides the `NetworkMessage.serializeWithMetadata` function, so it's more like

`RpcNetworkMessage.serializeWithMetadata(CannotSatisfyRequest.serialize())`

I think there might be some opportunity for improvement in the naming maybe? Like maybe renaming `serialize` to `serializeBody` and `serializeWIthMetadata` to `serialize`. I'm open to ideas around improving the structure of that, but don't have any immediate ideas.

The problem around deserialization is that TS doesn't allow for any sort of `static abstract` functions, so it's hard to express the pattern the same way. You'll notice the serialization interface doesn't actually mention any deserialization functions.

I think one improvement we could make is moving the deserialization logic into `RpcNetworkMessage` and `GossipNetworkMessage`, I'll work on that in this PR.

ironfish/src/network/messages/disconnecting.ts
dguenther: The size from `getSize` is just used to allocate the Buffer when serializing the message data. Tests will catch errors if the buffer is too small, but I agree it's a source of potential issues to keep the numbers in sync.

We could potentially do some abstraction around message field specification, where you'd define a message and its field types, and that would match up to the proper reader/writer and size behind the scenes, but that's probably outside the scope of this PR.

dguenther:
We need to address backward-compatibility with existing nodes on this branch as well (specifically, the new version shouldn't crash when old versions try to connect to it, and ideally it would ban the old versions)

ironfish/src/network/peers/connections/connection.ts
dguenther: I added a PR here that attempts a refactor: https://github.com/iron-fish/ironfish/pull/1258 We could potentially just merge them separately too, since this wouldn't be a breaking change.

danield9tqh: Yea I definitely like that change a lot better without the Asserts.

PR #1232, opened 2022-04-19
Add deposit CLI command

mat-if:
## Summary
image

## Breaking Change

```
[ ] Yes
[ ] No
```


ironfish-cli/src/commands/deposit.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  client: IronfishRpcClient | null = null
+  api: WebApi | null = new WebApi()
+
+  static flags = {
+    ...RemoteFlags,
+    fee: Flags.integer({
+      char: 'f',
+      default: 1,
+      description: `the fee amount in ORE, minimum of 1. 1 ORE is equal to ${MINIMUM_IRON_AMOUNT} IRON`,
```
leanthebean, line 34: We changed this to minimum of .1

NullSoldier: You are talking about the minimum transaction amount, this is the transaction fee.

ironfish-cli/src/commands/deposit.ts
```diff
@@ -0,0 +1,213 @@
...
+    const { canSend, errorReason } = await this.verifyCanSend(flags)
+    if (!canSend) {
+      Assert.isNotNull(errorReason)
+      this.log(errorReason)
+      this.exit(1)
+    }
+
+    const graffiti = this.sdk.config.get('blockGraffiti')
+
+    const balanceResp = await this.client.getAccountBalance()
```
NullSoldier, line 77: this needs to take the account name

ironfish-cli/src/commands/deposit.ts
```diff
@@ -0,0 +1,213 @@
...
+      default: 1,
+      description: `the fee amount in ORE, minimum of 1. 1 ORE is equal to ${MINIMUM_IRON_AMOUNT} IRON`,
+    }),
+    expirationSequence: Flags.integer({
+      char: 'e',
+      description: 'max number of blocks for the transaction to wait before expiring',
+    }),
+    account: Flags.string({
+      char: 'a',
+      default: 'default',
```
NullSoldier, line 42: this needs to not have a default for most commands to pick up the default account.

ironfish-cli/src/commands/deposit.ts
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    const { canSend, errorReason } = await this.verifyCanSend(flags)
+    if (!canSend) {
+      Assert.isNotNull(errorReason)
+      this.log(errorReason)
+      this.exit(1)
+    }
+
+    const graffiti = this.sdk.config.get('blockGraffiti')
```
NullSoldier, line 75: Should this come from the RPC layer so it'll match the node that sends the transaction, or no?

ironfish-cli/src/commands/deposit.ts
mat-if: Opted for loading the default account name from the RPC and passing it where needed to keep the validation logic simple. Explicit over implicit and all that.

NullSoldier: that sounds good to me!

NullSoldier (approved):
We can make further changes after this merges in.

PR #1242, opened 2022-04-27
refactor(ironfish): Remove workerPool from Transaction primitive

rohanjadvani:
## Summary

This code change reduces unnecessary coupling between a primitive and the worker pool. 

## Testing Plan

Run the node and existing unit tests.

## Breaking Change

```
[ ] Yes
[x] No
```


NullSoldier (approved):
Thank you so much for doing this PR. This is a piece of tech debt that I was very unhappy with. I'm so glad to see that you went out of your way to fix it. This type of PR exudes Iron Fish engineering values.

PR #1244, opened 2022-04-27
Update to supported ChaCha20Poly1305 library

dguenther:
## Summary

The rust-crypto library hasn't been updated in several years, and is reported as deprecated in the Rust security advisory DB: https://github.com/rustsec/advisory-db/blob/main/crates/rust-crypto/RUSTSEC-2016-0005.md

The ChaCha20Poly1305 crate has been kept up to date, and also has the benefit of an audit: https://crates.io/crates/chacha20poly1305

## Testing Plan

The hashing isn't supposed to change, so check to make sure that block generation and verification is compatible with the existing chain.

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish-rust-nodejs/Cargo.lock
+++ b/ironfish-rust-nodejs/Cargo.lock
@@ -3,6 +3,15 @@
 version = 3
 
 [[package]]
+name = "aead"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


dguenther:
I'll work on this more to see if I can make it backward-compatible.

dguenther:
Closing this for now -- I believe the algorithm we use is a draft version of chacha20 with a 64 bit nonce, but the commonly used version has a 96 bit nonce, so they're not directly compatible. This will need to be delayed till the next chain reset.

PR #1245, opened 2022-04-29
Update gitattributes to ignore certain file extensions when looking at line-endings

mat-if:
## Summary

Been running into this a lot when dealing with params files. Also have seen it pop up for the png file in the graph explorer module. According to the git docs, this should stop that from happening: https://git-scm.com/docs/gitattributes

## Breaking Change

```
[ ] Yes
[ ] No
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,3 +5,5 @@
 *.ts text
 *.js text
 *.json text
+*.png -text
+*.params -text


deekerno (approved):

dguenther (approved):
Nice, this is worth a try

PR #1251, opened 2022-05-02
feat(ironfish): Refactor serialization for accountsdb accounts store

rohanjadvani:
## Summary

Migrate `accounts` store in Accounts DB to binary serialization

## Testing Plan

Unit tests

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/account/database/accounts.ts
+++ b/ironfish/src/account/database/accounts.ts
@@ -16,9 +16,6 @@ export interface AccountsValue {
   rescan: number | null
 }
 
-// TODO: Remove after SDK release
-export type SerializedAccount = AccountsValue
-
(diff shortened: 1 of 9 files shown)


ironfish/src/account/database/accounts.ts
```diff
@@ -0,0 +1,74 @@
...
+export interface AccountsValue {
+  name: string
+  spendingKey: string
+  incomingViewKey: string
+  outgoingViewKey: string
+  publicAddress: string
+  rescan: number | null
+}
+
+// TODO: Remove after SDK release
```
dguenther, line 19: What's the dependency on the SDK release here?

dguenther (approved):

PR #1256, opened 2022-05-03
feat(ironfish): Add binary serialization to the database layer

rohanjadvani:
## Summary

Migrate encoding of keys and values in the database layer to use binary serialization.

## Testing Plan

Added unit tests and manually ran node.

## Breaking Change

```
[ ] Yes
[x] No
```


dguenther:
We need to address backward-compatibility with existing nodes on this branch as well (specifically, the new version shouldn't crash when old versions try to connect to it, and ideally it would ban the old versions)

ironfish/src/blockchain/database/transactions.ts
```diff
@@ -0,0 +1,46 @@
...
+import bufio from 'bufio'
+import { Transaction } from '../../primitives/transaction'
+import { WorkerPool } from '../../workerPool'
+
+export type TransactionsValue = {
+  transactions: Transaction[]
+}
+
+export class TransactionsValueEncoding implements IDatabaseEncoding<TransactionsValue> {
+  constructor(readonly workerPool: WorkerPool) {}
```
dguenther, line 15: This doesn't need the worker pool now

rohanjadvani: https://github.com/iron-fish/ironfish/pull/1265

PR #1259, opened 2022-05-04
feat(ironfish): Don't parse messages if websocket connection receives non-buffer

rohanjadvani:
## Summary

See title

## Testing Plan

Run node

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/network/peers/connections/webSocketConnection.ts
+++ b/ironfish/src/network/peers/connections/webSocketConnection.ts
@@ -78,11 +78,14 @@ export class WebSocketConnection extends Connection {
     }
 
     this.socket.onmessage = (event: MessageEvent) => {
+      if (!Buffer.isBuffer(event.data)) {
+        return
+      }
+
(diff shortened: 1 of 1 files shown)


deekerno (approved):

ironfish/src/network/peers/connections/webSocketConnection.ts
```diff
@@ -78,11 +78,14 @@ export class WebSocketConnection extends Connection {
     }
 
     this.socket.onmessage = (event: MessageEvent) => {
+      if (!Buffer.isBuffer(event.data)) {
+        return
```
dguenther, line 82: I think in this case we should close the connection immediately, like we do below if the peer sends an invalid message. That should free up connections for valid peers, as opposed to waiting for the connection to time out.

dguenther (changes requested):

dguenther (approved):

PR #1260, opened 2022-05-04
feat(ironfish): Bump `MAX_TRANSACTIONS_PER_BLOCK` to 300

rohanjadvani:
## Summary

See title

## Testing Plan

Run the node

## Breaking Change

```
[ ] Yes
[x] No
```


deekerno (approved):

dguenther:
This should actually be a breaking change, since it's a change to consensus requirements. I think we should target it into `chain-reset` just to keep staging backward-compatible for any phase 2 testing that's needed.

dguenther (approved):

PR #1264, opened 2022-05-04
Update reset command to work with database changes

dguenther:
## Summary

When starting a node after the database version has been increased, a warning pops up indicating the user should run `ironfish reset`.

However, the reset command also loads some database data in order to attempt to back up the accounts. This will fail, since the database structure has changed. Regardless, the command would never actually back up the accounts, since it doesn't try to load accounts from the database before writing out the backup file.

I think the main use case of the reset command should be to set the databases back to their original state, and upgrading the database should be covered by the database migration project, when that's completed. To this effect, I added an additional warning that running this command will delete your accounts.

### Example CLI output

```
/!\ WARNING: This will permanently delete your accounts. You can back them up by loading the previous version of ironfish and running ironfish export. /!\

Have you read the warning? (Y)es / (N)o: Y

You are about to destroy your node databases. The following directories will be deleted:

Accounts: /Users/derek/.ironfish-1/accounts/default
Blockchain: /Users/derek/.ironfish-1/databases/default

Are you sure? (Y)es / (N)o: N
```

## Testing Plan

Try running reset locally and ensure it deletes everything.

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish-cli/src/commands/reset.ts
+++ b/ironfish-cli/src/commands/reset.ts
@@ -44,7 +44,8 @@ export default class Reset extends IronfishCommand {
     confirmed = flags.confirm || (await CliUx.ux.confirm(warningMessage))
 
     if (!confirmed) {
-      this.exit(1)
+      this.log('Reset aborted.')
+      this.exit(0)
     }
 
     const accountDatabasePath = this.sdk.config.accountDatabasePath
@@ -59,7 +60,8 @@ export default class Reset extends IronfishCommand {
     confirmed = flags.confirm || (await CliUx.ux.confirm(message))
 
     if (!confirmed) {
-      this.exit(1)
+      this.log('Reset aborted.')
+      this.exit(0)
     }
 
     CliUx.ux.action.start('Deleting databases...')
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/reset.ts
```diff
@@ -34,66 +32,49 @@ export default class Reset extends IronfishCommand {
...
 
-    if (fs.existsSync(backupPath)) {
-      this.log(`There is already an account backup at ${backupPath}`)
+    confirmed = flags.confirm || (await CliUx.ux.confirm(warningMessage))
 
-      const confirmed = await CliUx.ux.confirm(
-        `\nThis means this failed to run. Delete the accounts backup?\nAre you sure? (Y)es / (N)o`,
-      )
+    if (!confirmed) {
+      this.exit(1)
```
rohanjadvani, line 47: If the user doesn't confirm a reset, I don't think the command necessarily failed. What do you think about logging some message and returning a 0?

dguenther: 👍 Updated this

rohanjadvani (approved):

PR #1268, opened 2022-05-05
Serialize randomness as a u64/BigInt to allow 2^64 search space

mat-if:
## Summary

This allows miners to use a bigger search space, which will allow more powerful mining rigs without them running out of search space and spending time idling. The other added benefit of this change is that it is more logical and inline with what one would expect. The previous limit of 2^53 was unexpected and could cause bugs with third party software down the road.

- Internal type is a BigInt (typescript side)
- Internal type is a u64 (rust side)
- Serialized as a big-endian hex string over network
- Pool keeps it as a hex string since it doesn't really need to do much with it
- 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER for further information on why this was capped at 2^53-1 prior to this PR

Fixes IRO-1854

## Testing Plan
Ran 2 nodes, ran pool, ran pool and solo miners

## Breaking Change

```
[ ] Yes
[ ] No
```


ironfish-rust/src/mining/mine.rs
```diff
@@ -64,11 +64,11 @@ mod test {
         let start = 42;
         let step_size = 1;
 
-        // Hardcoded target value derived from a randomness of 1, which is lower than 42
+        // Hardcoded target value derived from a randomness of 43, which is lower than 42
```
dguenther, line 67: Does this comment need updating to say 43 is higher than 42? I'm not 100% sure what's going on here

mat-if: Yeah.. technically, 43 is correct and 1 was incorrect. It is awkwardly worded, for sure. tldr: I run the test with a target of 0 and just print the target for randomness 43, then update it here. Honestly don't love the test either but wanted to test the batching

ironfish-rust/src/mining/thread.rs
```diff
@@ -96,10 +92,11 @@ fn process_commands(
             Command::NewWork(mut header_bytes, target, mining_request_id) => {
                 let mut batch_start = start;
                 loop {
-                    let batch_size = if batch_start + default_batch_size > MAX_SAFE_INTEGER {
-                        MAX_SAFE_INTEGER - batch_start
-                    } else {
+                    let remaining_search_space = u64::MAX - batch_start as u64;
```
dguenther, line 95: Can this `as u64` be removed? It seems like they both should be u64s now

ironfish-rust/src/mining/thread.rs
```diff
@@ -130,12 +127,12 @@ fn process_commands(
                         }
                     }
 
-                    batch_start += default_batch_size;
-                    if batch_start >= MAX_SAFE_INTEGER {
+                    if remaining_search_space < batch_size as u64 {
```
dguenther, line 130: Probably will work to remove the cast here too

dguenther, line 130: Also, should this be `==` rather than `<`? I saw up above that if `remaining_search_space <= default_batch_size`,`batch_size` gets set to `remaining_search_space`, so I'm not sure if there's a case where it'll be less.

Might be misreading, but just wanted to double-check the logic here

ironfish-rust-nodejs/src/lib.rs
```diff
@@ -98,7 +98,7 @@ impl ThreadPoolHandler {
     pub fn get_found_block(&self) -> Option<FoundBlockResult> {
         if let Some(result) = self.threadpool.get_found_block() {
             return Some(FoundBlockResult {
-                randomness: result.0 as f64,
+                randomness: format!("{:016x}", result.0 as u64),
```
dguenther, line 101: I think this cast can be removed

dguenther (commented):
I think that's all I have in terms of review, once comments are addressed this should be good to go! -- It might be neat to have a rust test for that wraparound case, but that might be too much work for this PR.

ironfish-rust/src/mining/mine.rs
dguenther: Ah gotcha, I understand it now, thanks!

ironfish-rust/src/mining/thread.rs
mat-if: Ah, you're absolutely correct. I meant for this to be `default_batch_size` instead of `batch_size`

dguenther (approved):
Looks good, thanks for those changes!

PR #1271, opened 2022-05-06
Deposit CLI command will exit early if account does not have required balance

mat-if:
## Breaking Change

```
[ ] Yes
[ ] No
```

### DIFF
--- a/ironfish-cli/src/commands/deposit.ts
+++ b/ironfish-cli/src/commands/deposit.ts
@@ -88,9 +88,15 @@ export default class Bank extends IronfishCommand {
     Assert.isNotUndefined(graffiti)
 
     const balanceResp = await this.client.getAccountBalance({ account: accountName })
-    const newBalance = oreToIron(
-      Number(balanceResp.content.confirmed) - ironToOre(IRON_TO_SEND) - fee,
-    )
+    const confirmedBalance = Number(balanceResp.content.confirmed)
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/deposit.ts
```diff
@@ -88,9 +88,15 @@ export default class Bank extends IronfishCommand {
     Assert.isNotUndefined(graffiti)
 
     const balanceResp = await this.client.getAccountBalance({ account: accountName })
-    const newBalance = oreToIron(
-      Number(balanceResp.content.confirmed) - ironToOre(IRON_TO_SEND) - fee,
-    )
+    const confirmedBalance = Number(balanceResp.content.confirmed)
+    if (confirmedBalance < ironToOre(IRON_TO_SEND) + fee) {
```
brekk, line 92: double checking, could this also be `<=`?

brekk (approved):

ironfish-cli/src/commands/deposit.ts
mat-if: should be okay. if you need 0.11 and you have 0.11, you can send, youll just end up with a balance of 0

lwisne (approved):

ironfish-cli/src/commands/deposit.ts
brekk: :shipit:

NullSoldier (approved):

PR #1275, opened 2022-05-07
Make default start start:once

NullSoldier:
## Summary

The default most people want is start:once while people who want to
develop and are experienced want the hot reloading setup, which should
be opt in to avoid confusion.

## Testing Plan

Run commands

## Breaking Change

`start:once` won't work, and `start` won't auto refresh. I'll put up a PR with any changes to the website too.

```
[x] Yes
[ ] No
```


brekk (approved):
Makes sense to me

mat-if (approved):

dguenther (approved):

PR #1276, opened 2022-05-07
Add ability for mining pool to ignore bad miners

mat-if:
## Summary

- Stratum server now has a set of client IDs that are known bad miners. When a client is added to this set, the server sends them a wait_for_work message.
- When the stratum server broadcasts messages, it will now skip broadcasting messages to bad clients
- When receiving work from miners, the pool will now clone the original block template instead of mutating it in place, which would cause issues later as it would try to re-broadcast for difficulty adjustment, etc

## Testing Plan

Tested locally with both a good and bad client simultaneously

## Breaking Change

```
[ ] Yes
[ ] No
```

### DIFF
--- a/ironfish/src/mining/pool.ts
+++ b/ironfish/src/mining/pool.ts
@@ -183,14 +183,17 @@ export class MiningPool {
       return
     }
 
-    const blockTemplate = this.miningRequestBlocks.get(miningRequestId)
+    const originalBlockTemplate = this.miningRequestBlocks.get(miningRequestId)
 
-    if (!blockTemplate) {
+    if (!originalBlockTemplate) {
       this.logger.warn(
         `Client ${client.id} work for invalid mining request: ${miningRequestId}`,
       )
       return
     }
+
+    const blockTemplate = Object.assign({}, originalBlockTemplate)
+    blockTemplate.header = Object.assign({}, originalBlockTemplate.header)
 
     const isDuplicate = this.isDuplicateSubmission(client.id, randomness)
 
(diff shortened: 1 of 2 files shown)


lwisne:
does the client id reset if the user updates to the new (correct) version and tries to mine again?

mat-if:
> does the client id reset if the user updates to the new (correct) version and tries to mine again?

Yep! It's an ephemeral ID assigned completely arbitrarily when a client connects

ironfish/src/mining/pool.ts
```diff
@@ -183,15 +183,18 @@ export class MiningPool {
...
 
-    if (!blockTemplate) {
+    if (!originalBlockTemplate) {
       this.logger.warn(
         `Client ${client.id} work for invalid mining request: ${miningRequestId}`,
       )
       return
     }
 
+    const blockTemplate = Object.assign({}, originalBlockTemplate)
```
lwisne, line 195: can't follow why you're doing this vs just using the original

NullSoldier (approved):

ironfish/src/mining/pool.ts
mat-if: from the pr description: "When receiving work from miners, the pool will now clone the original block template instead of mutating it in place, which would cause issues later as it would try to re-broadcast for difficulty adjustment, etc"

tldr: is that we have to change the randomness on the block template before we try to serialize it and fail (which is the primary change in this PR: the try/catch). by changing it to a bad value, it's mutating the original block template which then later blows up when it tries to check for other things (ie, other users submitting work or block being re-broadcast with updated difficulty)

lwisne: OK - this is pretty subtle, ideally we would have a test that illustrates the issue. and breaks if we do changes the block in place.

I would be fine in this case with a ticket to come back for it, given there doesn't appear to be an obvious easy place to put such a test ATM.

lwisne (approved):

PR #1277, opened 2022-05-07
Serialize memo as buffer in worker pool

mat-if:
## Summary

Very wip hacked together just to verify that this is fixing the issue
image

## Breaking Change

```
[ ] Yes
[ ] No
```


lwisne:
Is there a good place to create a test for this?

PR #1282, opened 2022-05-09
Worker pool createTransaction serializes memo as utf8

mat-if:
## Summary

This was getting clobbered as ascii which would've caused issues with syncing deposits

## Breaking Change

```
[ ] Yes
[ ] No
```

### DIFF
--- a/ironfish/src/workerPool/tasks/createTransaction.test.slow.ts
+++ b/ironfish/src/workerPool/tasks/createTransaction.test.slow.ts
@@ -97,7 +97,7 @@ describe('CreateTransactionRequest', () => {
         {
           publicAddress: '',
           amount: BigInt(5),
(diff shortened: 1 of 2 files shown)


ironfish/src/workerPool/tasks/createTransaction.test.slow.ts
```diff
@@ -97,7 +97,7 @@ describe('CreateTransactionRequest', () => {
         {
           publicAddress: '',
           amount: BigInt(5),
-          memo: '',
+          memo: '👁️🏃🐟',
```
brekk, line 100: :shipit:

brekk (approved):

deekerno (approved):

rohanjadvani (approved):

PR #1283, opened 2022-05-09
Any user-inputtable strings being serialized use utf8 encoding

mat-if:
## Summary

More serialization encoding changes for any user-inputtable strings. Continuation of #1282 

## Breaking Change

```
[ ] Yes
[ ] No
```


brekk (approved):
Soon I can use zero-width strings _everywhere_.

```
```

rohanjadvani (approved):

PR #1284, opened 2022-05-09
Miner cmd: add hashrate as rich cli ux action

mat-if:
## Summary

- Can be disabled with `--no-richOutput`

image

## Breaking Change

```
[ ] Yes
[ ] No
```

### DIFF
--- a/ironfish-cli/src/commands/miners/start.ts
+++ b/ironfish-cli/src/commands/miners/start.ts
@@ -139,6 +139,6 @@ export class Miner extends IronfishCommand {
       CliUx.ux.action.status = formatted
     }
 
-    this.updateInterval = setInterval(() => updateHashPower(), 1000)
+    this.updateInterval = setInterval(updateHashPower, 1000)
   }
 }
--- a/ironfish-cli/src/commands/miners/start.ts
+++ b/ironfish-cli/src/commands/miners/start.ts
@@ -2,13 +2,15 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import {
+  FileUtils,
   GraffitiUtils,
   isValidPublicAddress,
   MiningPoolMiner,
   MiningSoloMiner,
   parseUrl,
+  SetIntervalToken,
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/miners/start.ts
```diff
@@ -33,6 +37,11 @@ export class Miner extends IronfishCommand {
       char: 'a',
       description: 'the public address to receive pool payouts',
     }),
+    richOutput: Flags.boolean({
```
brekk, line 40: Minor but could the flags be like `--verbose` vs. `--quiet`? `--no-richOutput` seems very unintuitive (to me).

brekk (approved):
Minor suggestions, probably no blockers

ironfish-cli/src/commands/miners/start.ts
mat-if: `--verbose` exists and is used for, well, more verbose output. This is a fancy thing that hangs at the bottom of the terminal, which does need to able to be turned off because it doesn't play nice with certain TTY emulators, etc. I also don't love the name, so open to suggestions

ironfish-cli/src/commands/miners/start.ts
```diff
@@ -103,7 +119,26 @@ export class Miner extends IronfishCommand {
...
 
       miner.start()
+      if (flags.richOutput) {
+        this.displayHashrate(miner)
+      }
+
       await miner.waitForStop()
+      if (this.updateInterval) {
+        clearInterval(this.updateInterval)
+      }
```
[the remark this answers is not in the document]
mat-if: Oh, it's a typescript thing, not a clearInterval thing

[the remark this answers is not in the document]
brekk: ☝🏼 I count that as a failure of TS ^^

ironfish-cli/src/commands/miners/start.ts
brekk: Yeah, not sure then. `richOutput` makes me think of Ron Burgundy and "Corinthian leather" ^^

dguenther (approved):

PR #1300, opened 2022-05-10
If there is no default account on startup, create one

danield9tqh:
## Summary
After running ironfish reset, the node used to create an account for the user if one didn't exist on next startup. We broke that with a [recent change](https://github.com/iron-fish/ironfish/pull/1288/files#diff-f746f6e76b029871c0e5cc11c3752b1043ac76a0cee18e8eaf8bae23ebdf916cL74) to ironfish reset, so we should do more investigation on that

## Testing Plan
Tested locally

## Breaking Change

```
[] Yes
[X] No
```

### DIFF
--- a/ironfish-cli/src/commands/start.ts
+++ b/ironfish-cli/src/commands/start.ts
@@ -210,6 +210,10 @@ export default class Start extends IronfishCommand {
       await this.firstRun(node)
     }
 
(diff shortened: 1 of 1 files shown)


brekk (approved):
Even I can track this change with almost no familiarity with the code. :shipit:

PR #555, opened 2021-12-03
Update docker-compose configurations

testingPurposes1:
## Summary
This pull request made to make things with Docker setup more clear.

Main changes:

1) Add docker-compose.yml for the users in incentivized testnet and change previous file for the developers -> docker-compose.dev.yml
2) Change version of docker-compose: 3.7 for the devs file and 2.4 for the users file
3) Change Ironfish beta image to main image
3) Add deploy key in the devs file to use Docker in Swarm Mode and/or limit resources/restart behaviour
4) Add memory limitation (default:8192MB) to handle out of memory error
5) Add small usage guide
## Breaking Change
```
[ ] Yes
[x] No
```

PR #1322, opened 2022-05-11
Update docker-compose configuration

etretien:
## Summary
Updated the main (non-dev) `docker-compose.yml` to newer version 3.7, for it to be consistent with the `docker-compose.dev.yml` intended for developers running new networks. The reason here is that I tried to use the dev version to run a public testnet node (being more used to 3.x in general), missed the `bootstrap=''` flag and lost some time trying to figure out why it doesn't sync with the network. Pure my fault, but I thought that some people might do it too if we don't make more it more intuitive.

Also, added `-v` flag to the miner start command for verbose log output.

Documentation was also updated.

## Testing Plan
Tested manually on 4 servers with public testnet nodes. No automated tests due to no changes in application code.

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/tools/Docker.md
+++ b/tools/Docker.md
@@ -2,8 +2,8 @@
 
 There are 2 docker-compose files:
 
- 1. `docker-compose.yml` - Intended for users who wish to start a node, pool, and miner using Docker and latest version of Docker Compose.
- 2. `docker-compose.legacy.yml` - Intended for users who wish to start a node, pool (optional), and miner (optional) using Docker and older versions of Docker Compose.
+ 1. `docker-compose.yml` - Intended for users who wish to start a node, pool, and miner using Docker and newer versions of Docker Compose.
+ 2. `docker-compose.legacy.yml` - Intended for users who wish to start a node, pool, and miner using Docker and older versions of Docker Compose.
  3. `docker-compose.dev.yml` - Intended for developers who wish to run two (or more) nodes with a miner on a new network.
 
 ## Production config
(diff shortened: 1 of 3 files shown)


dguenther:
If you're interested in updating the docs or commands to improve the existing docker-compose files, feel free, but we'd rather leave support for the older 2.4 version -- In the past we had someone add support for that version because (I believe) some Linux distros still ship older versions of Docker (https://github.com/iron-fish/ironfish/pull/555)

etretien:
True, that's why docker official documentation recommends installing from docker repositories instead of OS native ones.
Maybe I could leave both versions and describe more in the docs, so that older version is supported but not recommended?

etretien:
@dguenther I've moved the older version to a separate docker-compose.legacy.yml config. That would incentivise users to use newer versions while keeping compatibility for those who don't want to update. What do you think?
Alternatively, we could older version in docker-compose.yml, and use another name for newer configuration.

Also, I've included the mining pool section to the compose file (made it for my own use and thought it might be useful to others too). Users are free to comment/remove sections that they don't need on a particular machine.

Updated the docs too.

dguenther:
I'd rather not support multiple docker-compose files -- the 2.4 version should work functionally identically, just with different syntax. If you want to update the docs with additional context around how to look up documentation for the 2.4 version, or you'd like to add documentation on how to add flags to the docker compose file, that would be welcome though.

NullSoldier:
Ok so, I have a few thoughts here. I don't want to be the project that keeps their stuff stuck to older versions. I'd like to upgrade things at a reasonable pace, and I think we should ultimately upgrade our docker-compose version at some point. However, I think splitting these into legacy and not legacy doesn't feel great.

I think we should close this PR for now, and pick a point in the future to just upgrade docker-compose. I think if Ubuntu LTS upgrades docker-compose that should be a triggering event for us to do this upgrade here.

Overall, closing PR's for work contributed doesn't feel great because you put in the work, which we thank you for. If you ever have  suggestions for infrastructure you can feel free to message us in discord in [#community-dev](https://discord.com/channels/771503434028941353/831945060777852998)

PR #1323, opened 2022-05-11
Revert "If there is no default account on startup, create one"

danield9tqh:
Reverts iron-fish/ironfish#1300


deekerno (approved):
Happens to the best of us.

PR #1330, opened 2022-05-12
Sync transactions counts transactions in buffer instead of blocks

mat-if:
## Summary

For sync transactions we don't actually care about the number of blocks, but rather the number of transactions. 500 is probably low, but given the current state of the API I'd rather err on the side of caution

## Breaking Change

```
[ ] Yes
[ ] No
```


deekerno (approved):
This seems like a good change.

rohanjadvani (approved):

dguenther (approved):

PR #1339, opened 2022-05-12
Update string.ts

ironfishup:
## Summary
Since the hash string that was used here: md5 was insecure, I changed it to specify that a secure hashing algorithm needs to be added. You may change it to something like SHA-256 or similar. In the commit, I added sha256

## Breaking Change

```
[ ] Yes
[ ] No
```

### DIFF
--- a/ironfish/src/utils/string.ts
+++ b/ironfish/src/utils/string.ts
@@ -20,7 +20,7 @@ function hashToNumber(value: string): number {
  * simple way to hash a string for test or display purposes
  */
 function hash(value: string): Buffer {
-  return Buffer.from(crypto.createHash('md5').update(value).digest('hex'))
+  return Buffer.from(crypto.createHash(SHA256).update(value).digest('hex'))
 }
 
 /**
--- a/ironfish/src/utils/string.ts
+++ b/ironfish/src/utils/string.ts
@@ -20,7 +20,7 @@ function hashToNumber(value: string): number {
  * simple way to hash a string for test or display purposes
  */
 function hash(value: string): Buffer {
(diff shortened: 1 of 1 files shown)


ironfish/src/utils/string.ts
```diff
@@ -20,7 +20,7 @@ function hashToNumber(value: string): number {
  * simple way to hash a string for test or display purposes
  */
 function hash(value: string): Buffer {
-  return Buffer.from(crypto.createHash('md5').update(value).digest('hex'))
+  return Buffer.from(crypto.createHash('sha256').update(value).digest('hex'))
```
dguenther, line 23: There's no need for a cryptographically-secure function here -- it's just trying to turn a string into a number value to index into an array: https://github.com/iron-fish/ironfish/blob/48d9ceff7dd490b3b778a765379ce9b9c6b5b976/ironfish/src/logger/reporters/text.ts#L92

If you'd like to make a change here, I'd suggest bringing over the simple xor from the testnet website: https://github.com/iron-fish/website-testnet/blob/master/utils/graffitiToColor.ts#L14

dguenther, line 23: There's no need for a cryptographically-secure function here -- it's just trying to turn a string into a number value to index into an array: https://github.com/iron-fish/ironfish/blob/48d9ceff7dd490b3b778a765379ce9b9c6b5b976/ironfish/src/logger/reporters/text.ts#L92

If you'd like to make a change here, I'd suggest bringing over the simple xor from the testnet website: https://github.com/iron-fish/website-testnet/blob/master/utils/graffitiToColor.ts#L14

dguenther:
Closing this for now, but feel free to reopen if you make the changes mentioned in the comment!

PR #1351, opened 2022-05-12
Add a try-catch around new block creation

dguenther:
## Summary

Fixes the transaction expiration crash when mining. I added a test for it, but the test takes quite a bit of setup to get working, so I could use a second opinion on it.

Fixes IRO-1876

## Testing Plan

New test should pass.

## Breaking Change

```
[ ] Yes
[ ] No
```

### DIFF
--- a/ironfish/src/rpc/routes/accounts/getPublicKey.test.ts
+++ b/ironfish/src/rpc/routes/accounts/getPublicKey.test.ts
@@ -22,7 +22,7 @@ jest.mock('@ironfish/rust-nodejs', () => {
 })
 
 describe('Route account/getPublicKey', () => {
-  const routeTest = createRouteTest()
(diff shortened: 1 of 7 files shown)


config/jest.config.base.js
```diff
@@ -5,5 +5,5 @@ module.exports = {
   watchPlugins: ["../config/jestWatchPlugin"],
   coverageProvider: "v8",
   coverageReporters: ["text-summary", "json", "clover", "text"],
-  testPathIgnorePatterns: ['.*\\.test\\.slow\\.ts$', '.*\\.test\\.perf\\.ts$'],
+  testPathIgnorePatterns: ['.*\\.test\\.perf\\.ts$'],
```
dguenther, line 8: I need to revert this

mat-if (approved):
what a wonderful test

PR #1357, opened 2022-05-13
Move workerPool saturation check above validation for new transaction gossip

mat-if:
## Breaking Change

```
[ ] Yes
[ ] No
```


leanthebean (approved):

ironfish/src/network/peerNetwork.test.ts
```diff
@@ -290,6 +295,9 @@ describe('PeerNetwork', () => {
           'verifyNewTransaction',
         )
 
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+        expect(node.workerPool.saturated).toEqual(false)
```
mat-if, line 299: This feels pretty goofy, but this test wasn't actually testing the case it's supposed to be, so this feels a little better than the test silently changing and not testing in the event of a future refactor

dguenther (approved):

PR #1387, opened 2022-05-13
Add rolling message counting for peers and banning them if they exceed specific amounts

mat-if:
## Summary

The goal of this is to disallow bad actors from continuing to send arbitrary amounts of messages, valid or not, that could impede performance of your own node.

## Breaking Change

```
[ ] Yes
[ ] No
```


lwisne:
My first instinct is this approach makes me nervous vs a more targeted banning of peers with a large volume of bad transactions. This could create a pretty brutal scenario in which the network gets a high amount of volume, and the response to this is peers start banning each other. I imagine that would end badly.

II would be interested to hear your view on what class of issues we would possibly be ignoring if we just looked at bad transactions for the counter mechanism.

PR #1395, opened 2022-05-14
FEAT: auto-update script

hairtail:
## Summary
Add scripts with which Ironfish node/miner can keep auto-updated.
## Breaking Change

```
[ ] Yes
[ ] No
```


dguenther:
Thanks for the contribution, but we're not currently accepting PRs that add scripts that aren't in use by the development team -- they're hard for us to maintain as we use different environments and may make changes that break them without realizing it.

PR #1448, opened 2022-05-18
Add abstract IronfishClient

hughy:
## Summary

IRO-1940: define IronfishClient, move IronfishRpcClient down one level of abstraction

implementing a TCP client using the native 'net' module will introduce shared
logic with the existing IPC client. the `IronfishMemoryClient` does not have the
same network connection logic that TCP and ICP do, so an additional level of
abstraction is useful for sharing logic between TCP and ICP implementations.

Changes:
- add client.ts, define `IronfishClient`
- move implementation of `IronfishRpcClient` into `IronfishClient`, minus
  abstract functions for connect/close
- change references from `IronfishRpcClient` to `IronfishClient`

## Testing Plan

- existing test suite passes
- local node and pool run successfully

## Breaking Change

```
[ ] Yes
[X] No
```


ironfish/src/sdk.ts
```diff
@@ -146,7 +146,7 @@ export class IronfishSdk {
       config.get('rpcRetryConnect'),
     )
 
-    const clientMemory = new IronfishMemoryClient({ logger })
+    const clientMemory = new IronfishMemoryClient()
```
danield9tqh, line 149: Can we still pass in the logger here?

danield9tqh (approved):

PR #1450, opened 2022-05-18
Refactor MemoryClient

danield9tqh:
## Summary
The [MemoryClient](https://github.com/iron-fish/ironfish/blob/c6a17e98b1f9add2e5727ae3f03676fe14ad0fac/ironfish/src/rpc/clients/memoryClient.ts#L22) is an interface to all the methods you can call on a node like /getStatus , /accounts/balance  etc. It is similar to the TcpClient or IpcClient but it does not connect to a server remotely. Instead it just makes direct function calls to the node. Because of this we want to refactor it to not have all the same functions that an RpcClient might have such as connect . If it is just making direct function calls then it doesn't make sense to connect  to anything. This should simplify the code quite a bit and make it easier to tackle TCP security because we don't have to worry about the MemoryClient functionality which should be unrelated to TCP connections

## Testing Plan
Unit tests + local testing

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/sdk.ts
+++ b/ironfish/src/sdk.ts
@@ -146,7 +146,7 @@ export class IronfishSdk {
       config.get('rpcRetryConnect'),
     )
 
-    const clientMemory = new IronfishMemoryClient({ logger })
+    const clientMemory = new IronfishMemoryClient(logger)
 
     return new IronfishSdk(
       pkg || IronfishPKG,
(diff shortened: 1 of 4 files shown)


ironfish/src/sdk.ts
```diff
@@ -30,7 +30,6 @@ import { NodeUtils } from './utils'
 export class IronfishSdk {
   pkg: Package
   client: IronfishIpcClient
-  clientMemory: IronfishMemoryClient
```
hughy, line 33: This has me wondering whether we need `client` to be a property of the sdk? 🤔

It seems like then we need to keep track of whether that client is connected. Do we have any use for a pre-connection client?

hughy (approved):

PR #1453, opened 2022-05-19
Combine MemoryAdapter and MemoryClient functionality

danield9tqh:
## Summary
The MemoryAdaper is supposed to be analogous to a TcpAdapter and IpcAdapter . All 3 implement the IAdapter  interface that can start and stop the adapter. The main difference of the MemoryAdaper  is that it is not running a server. It does not need to be stopped or started. It is just making function calls directly to the internal Router  code.

This also means that the idea of a client (MemoryClient) server (MemoryAdaper) relationship doesn't make much sense since the client and server are directly connected they should be the same code. For this reason we want to get rid of the MemoryAdapter  and just call Router.route  directly from the MemoryClient . Either that or greatly reduce the code in each and remove unnecessary orchestration.

## Testing Plan
Unit tests + local testing

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/rpc/adapters/memoryAdapter.ts
+++ b/ironfish/src/rpc/adapters/memoryAdapter.ts
@@ -9,7 +9,6 @@ import { Response, ResponseEnded } from '../response'
 import { ALL_API_NAMESPACES, Router } from '../routes'
 import { RpcServer } from '../server'
 import { Stream } from '../stream'
-import { IAdapter } from './adapter'
 import { ResponseError } from './errors'
 
 /**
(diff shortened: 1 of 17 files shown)


ironfish/src/rpc/adapters/memoryAdapter.ts
```diff
@@ -5,55 +5,27 @@ import { Assert } from '../../assert'
...
-    return this.requestStream<TEnd, unknown>(route, data).waitForEnd()
-  }
-
+export class MemoryAdapter {
   /**
    * Makes a request against the routing layer with a given route, and data and returns
    * a response for you to accumulate the streaming results, or wait for a response
    */
-  requestStream<TEnd = unknown, TStream = unknown>(
+  static requestStream<TEnd = unknown, TStream = unknown>(
```
hughy, line 24: `MemoryAdapter` will be a class with just one static method -- should we move `requestStream` into the Client?

danield9tqh: I tried but was running into some errors. Going to investigate before I merge this

danield9tqh: It looks like it causes a circular dependency in the imports. There's not actually a circular dependency or else we'd get the same error here in this code which we don't. But if we move this logic to `MemoryClient` it will cause one just due to file imports. I tried looking into it a bit but I don't think that's something I want to solve in this PR

hughy (approved):

PR #1454, opened 2022-05-19
Add fallback mined block url if terminal doesn't support hyperlinks

vchs1v:
## Summary

Some terminals does not support hyperlink, which is diplayed as a result of miners:mined command. This PR adds a fallback and displays just an explorer url in such case.

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish-cli/src/commands/miners/mined.ts
+++ b/ironfish-cli/src/commands/miners/mined.ts
@@ -12,7 +12,6 @@ import {
 } from '@ironfish/sdk'
 import { CliUx, Flags } from '@oclif/core'
 import readline from 'readline'
-import supportsHyperlinks from 'supports-hyperlinks'
 import { parseNumber } from '../../args'
 import { IronfishCommand } from '../../command'
 import { RemoteFlags } from '../../flags'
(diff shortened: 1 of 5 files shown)


ironfish-cli/src/commands/miners/mined.ts
```diff
@@ -123,10 +124,10 @@ export class MinedCommand extends IronfishCommand {
...
     const amount = MathUtils.round(oreToIron(block.minersFee), 2)
 
-    const link = linkText(
-      `https://explorer.ironfish.network/blocks/${block.hash}`,
-      'view in web',
-    )
+    const explorerLink = `https://explorer.ironfish.network/blocks/${block.hash}`
+    const link = supportsHyperlinks.stdout
+      ? linkText(explorerLink, 'view in web')
+      : explorerLink
```
NullSoldier, line 130: Can you just put this code inside of linkText so that it always falls back to the URL when links are not supported?

vchs1v: I didn't get second part. Why we should add `stdIn` parameter that is default to true?

NullSoldier: If you move it into `linkText()`, which stream will you check support for? So linkText needs to be parameterized now to accept a parameter to indicate which stream to check support for. Though, 99% of the time people will use this module with STDOUT, so it's nice to default that new parameter with `true` so you don't need to configure it most of the time.

NullSoldier: I made the change, merging.

NullSoldier:
I tested manually and audited the new library.

NullSoldier (approved):

ironfish-cli/src/commands/miners/mined.ts
vchs1v: Got it! I was confused by `stdIn`. Thanks for explanation

PR #1460, opened 2022-05-19
Add test for blocks:show command

vchs1v:
## Summary

This PR adds tests for `blocks:show` cli command

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish-cli/src/commands/blocks/show.test.ts
+++ b/ironfish-cli/src/commands/blocks/show.test.ts
@@ -1,0 +1,72 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { GetBlockInfoResponse } from '@ironfish/sdk'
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/blocks/show.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    jest.dontMock('@ironfish/sdk')
+  })
+
+  describe('fetching block content by hash', () => {
+    test
+      .stdout()
+      .command(['blocks:show', 'testHash'])
+      .exit(0)
+      .it('logs block content in json format', (ctx) => {
+        expectCli(ctx.stdout).include(JSON.stringify(responseContent, undefined, '  '))
```
NullSoldier, line 69: This is an ok test, but it does mock a lot, and basically just duplicates the code from the command, in a way. So it's mostly testing that it doesn't diverge from this test file. If it does, someone will have to come in here and update it basically. If we detect the hash from the block in the output, I think that would be fine enough to ensure that it's mostly working. I'll probably merge this for now since it's a definitive addition though.

NullSoldier (approved):

PR #1463, opened 2022-05-20
feat: implements TCP client using native net module

hughy:
## Summary

IRO-1912: our IPC client implementation uses `node-ipc` for either ICP or TCP
connections. We will implement a client and adapter that use TLS, but the
`node-ipc` implementation of TLS will be more difficult to use than writing our
own using the native `net` and `tls` modules. to simplify the future
implementation of TLS these changes implement a TCP client that uses the native
`net` module. A native TCP adapter has already been implemented (IRO-1759).

- creates `IronfishTcpClient` using `net` that extends `IronfishRpcClient`
- moves shared properties and logic from `IronfishIpcClient` up to the RPC client

## Testing Plan

- existing test suite
- local testing of pic and tcp clients

## Breaking Change

```
[ ] Yes
[X] No
```


ironfish/src/rpc/clients/tcpClient.ts
```diff
@@ -0,0 +1,169 @@
...
+  .required()
+
+export class IronfishTcpClient extends IronfishRpcClient {
+  readonly host: string
+  readonly port: number
+  readonly client: net.Socket
+  retryConnect: boolean
+  connectTimeout: SetTimeoutToken | null
+  isConnected = false
+  connectionMode: string | undefined = 'tcp'
```
danield9tqh, line 36: we can probably remove the `undefined` here and also make it readonly

ironfish/src/rpc/clients/tcpClient.ts
```diff
@@ -0,0 +1,169 @@
...
+    data: yup.mixed().required(),
+  })
+  .required()
+
+export class IronfishTcpClient extends IronfishRpcClient {
+  readonly host: string
+  readonly port: number
+  readonly client: net.Socket
+  retryConnect: boolean
+  connectTimeout: SetTimeoutToken | null
```
danield9tqh, line 34: Can we make this private as well?

ironfish/src/rpc/clients/ipcClient.ts
```diff
@@ -38,25 +29,11 @@ export class IronfishIpcClient extends IronfishRpcClient {
   client: IpcClient | null = null
   isConnecting = false
```
danield9tqh, line 30: It doesn't look like we're using this value anywhere. Can we remove it?

danield9tqh (approved):
Left some comments but they will be addressed in future PRs as they are just additional refactoring

ironfish/src/sdk.ts
```diff
@@ -128,20 +129,25 @@ export class IronfishSdk {
...
-          }
-        : {
-            mode: 'ipc',
-            socketPath: config.get('ipcPath'),
-          },
-      logger,
-      config.get('rpcRetryConnect'),
-    )
+    let client: IronfishRpcClient
+    if (config.get('enableNativeRpcTcpAdapter')) {
```
danield9tqh, line 133: Can we check both here
```javascript
config.get('enableNativeRpcTcpAdapter') && config.get('enableRpcTcp')
```

ironfish/src/mining/pool.ts
```diff
@@ -49,8 +49,8 @@ export class MiningPool {
 
   recalculateTargetInterval: SetTimeoutToken | null
 
-  private constructor(options: {
```
hughy, line 52: Bad rebase -- I'll return this to private

ironfish/src/rpc/clients/tcpClient.test.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    client.request(route)
+    expect(client.client.write).toHaveBeenLastCalledWith(expectedMessage)
+  })
+
+  it('should handle message responses', async () => {
+    expect.assertions(1)
+
+    // @ts-ignore
+    const spyOnMessage = jest.spyOn(client, 'onMessage')
```
danield9tqh, line 52: Can we just test the TcpClient.request method here instead of testing the internal (private) methods?

danield9tqh (approved):

PR #1473, opened 2022-05-21
Use yarn frozen-lockfile

kim0:
## Summary
To ensure reproducable and deterministic builds, when performing yarn install, make sure to use the lockfile. Without `--frozen-lockfile`, yarn will update the lockfile rather than using the pinned versions.

## Breaking Change

```
[ ] Yes
[ ] No
```


NullSoldier:
Telling people to run this, doesn't actually make builds less consistent, because the default behaviour of yarn is to use the materialized versions inside the yarn lock file already. The only time `--frozen-lockfile` is useful is if the `package.json` is out of sync with the lockfile. We already use `--frozen-lockfile` in all of our CI and deploy jobs, which would fail the builds if they got out of sync so we know they aren't out of sync.

PR #1482, opened 2022-05-22
Node snapshot blockchain and miner IronFish on Akash Network

Dimokus88:
Hi!
In my proposal, containers for deploying IronFish nodes on AkashNetwork:
1. Deployment of the node for snapshots of the Ironfish network.
- Customizable snapshot times.

2. Deployment of the node and miner Ironfish on Akash power.
- Ready guide in Russian and English.
- flexible rental capacity settings in the deploy.yml file.
- start of the node with access to the mining process within 15 minutes.

Don't hesitate to ask questions!
Graffity: Dimokus
Discord: Dimokus#1032
Telegram: https://t.me/Dimokus


NullSoldier:
First of all, I think this PR is great and boosts the Iron Fish eco system in a way that we want to support.  However, I don't think this should go into https://github.com/iron-fish/ironfish. Instead it should be it's own repository.

We deploy all of our official nodes and infrastructure to AWS, and we have a separate repository that contains all the scripts needed to deploy to AWS because it's host provider specific. I see this as the same thing, where this ties it to just one specific hosting service Akash Network.

I think many users will want to host using Akash Network, but some users may want to use another provider. The example we want to set is for people to create other supporting repositories with these types of tooling and scripts. 

The primary reason for it to be a separate thing as well, is that the core developers are not going to use it. So it's going to be hard for us to maintain that code, and upgrade it as we make breaking changes when we don't use it to begin with. It's better for the users that use this, to maintain it separately as a product. We can even link to it in our community resources.

Dimokus88:
https://github.com/Dimokus88/ironfish
I have created a repository for this project and I support it myself, you don’t have to worry about it =) I would also be grateful if you rate my work in additional points in the general table =)

PR #1483, opened 2022-05-22
Add: TLS. Pin version. Integrity

kim0:
## Summary
The 'integrity' attribute allows for the browser to verify that externally hosted files (for example from a CDN) are delivered without unexpected manipulation. Without this attribute, if an attacker can modify the externally hosted resource, this could lead to XSS and other types of attacks. To prevent this, include the base64-encoded cryptographic hash of the resource (file) you’re telling the browser to fetch in the 'integrity' attribute for all externally hosted files.

## Breaking Change

```
[ ] Yes
[ ] No
```

### DIFF
--- a/ironfish-graph-explorer/src/index.html
+++ b/ironfish-graph-explorer/src/index.html
@@ -1,11 +1,11 @@
 <head>
-    <script src="http://unpkg.com/lodash"></script>
-    <script src="http://unpkg.com/d3-dsv"></script>
-    <script src="http://unpkg.com/dat.gui"></script>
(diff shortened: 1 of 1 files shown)


NullSoldier:
This is fine to merge, but I have some thoughts about this project as a whole.

This project was a toy project demo I made awhile back to debug the block chain. I think we should probably convert this to a real mono repo setup with real dependencies and bundler and remove all these unpkg deps in general. Also we should convert it to typescript at some point.

NullSoldier (approved):

NullSoldier:
I rebased this on staging, merging.

PR #1484, opened 2022-05-22
Don't run as root. Use no-new-privileges

kim0:
## Summary
Not running docker containers as root is best-practice. Since this is an example for running iron-fish in production, I suggest we add this

Services `node` and `miner` allow for privilege escalation via setuid or setgid binaries. Add `no-new-privileges:true` in 'security_opt' to prevent this.

## Breaking Change

[ ] Yes
[ ] No

### DIFF
--- a/tools/docker-compose.yml
+++ b/tools/docker-compose.yml
@@ -8,6 +8,9 @@ services:
         tty: true
         restart: always
         command: ["start", "--graffiti", "graffitiExample", "--name", "nodeExample"]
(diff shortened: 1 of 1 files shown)


NullSoldier:
We deleted these docker compose files, see this https://github.com/iron-fish/ironfish/pull/1513

NullSoldier:
We moved those files to this REPO, https://github.com/iron-fish/docker-compose

Issue #1419, opened 2022-05-16
Payout percent calculation issue


xbeastx:
### What happened?

It looks like there is a bug in payout amount calculation:
https://github.com/iron-fish/ironfish/blob/d6845e388a5a3fc6911ed1ca1e70571c36f5af91/ironfish/src/mining/poolShares.ts#L122

The default value balancePercentPayout is 10 (10 percents) and it will calculate correct 200 / 10 = 20. But if we will choose another value let's say 50% than 200 / 50 it's not make sense.

### Version

v0.1.34

devzorg:
It should be multiplicated,
something like `const payoutAmount = BigIntUtils.multiply(confirmedBalance, BigIntUtils.divide(this.balancePercentPayout, 100))`

PR #1491, opened 2022-05-24
add flag to override pool balance payout percentage

danield9tqh:
## Summary
Addresses https://github.com/iron-fish/ironfish/issues/1419. This change adds a flag to the `miners:pools:start` command to correctly override `balancePercentPayout` in the config.

Currently the config variable `poolBalancePercentPayout` doesn't represent an actual percentage because we are using it incorrectly. Instead it represents a divisor. So
- `config.poolBalancePercentPayout = 2` would pay out 50%
- `config.poolBalancePercentPayout = 3` would pay out 33%
- `config.poolBalancePercentPayout = 4` would pay out 25%
- etc

This change allows you to run `ironfish miners:pools:start --balancePercentPayout=30` and it will actually pay out 30%. If you do not pass in the `--balancePercentPayout` flag, it will retain it's original functionality. This is temporary to allow people running pools to migrate over to using a correct percentage. The old config variable will be removed or updated to calculate correctly in a future release.

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/mining/poolShares.ts
+++ b/ironfish/src/mining/poolShares.ts
@@ -25,14 +25,16 @@ export class MiningPoolShares {
   private attemptPayoutInterval: number
   private accountName: string
   private balancePercentPayout: bigint
+  private balancePercentPayoutFlag: number | undefined
 
-  constructor(options: {
+  private constructor(options: {
     db: PoolDatabase
     rpc: IronfishIpcClient
     config: Config
     logger?: Logger
     discord?: Discord
     enablePayouts?: boolean
+    balancePercentPayoutFlag?: number
   }) {
     this.db = options.db
     this.rpc = options.rpc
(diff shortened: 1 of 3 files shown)


ironfish/src/mining/poolShares.ts
```diff
@@ -119,7 +124,12 @@ export class MiningPoolShares {
     const balance = await this.rpc.getAccountBalance({ account: this.accountName })
     const confirmedBalance = BigInt(balance.content.confirmed)
 
-    const payoutAmount = BigIntUtils.divide(confirmedBalance, this.balancePercentPayout)
+    let payoutAmount: number
+    if (this.balancePercentPayoutFlag !== undefined) {
```
brekk, line 128: Trivial, but `!this.balancePercentPayoutFlag` is idiomatic

brekk (approved):

ironfish/src/mining/poolShares.ts
danield9tqh: I wanted to possibly account for a percentage of 0. Probably not a real use case but just to be more explicit

PR #1492, opened 2022-05-24
pass down logger in mining code

danield9tqh:
## Summary
We would like to start exporting some of our mining pool logs to a file and/or logging service to better analyze errors. In order to modify all the loggers in the mining code, we need to pass them down from when the logger is created in [command.ts](https://github.com/iron-fish/ironfish/blob/4bc3765015cb4bcfbb9732def5ae36ff8dc9db23/ironfish-cli/src/command.ts#L63) so if we change the output of the logs then all the loggers will change.

## Breaking Change

```
[ ] Yes
[X] No
```


lwisne:
How many different loggers do we use?

My assumption is "few/one", in which case it may be nice to have a global logger manager and we just grab the one we want when we need it instead of having to pass around. 

Either way, I think you identified an issue with the code as is, which is that it appears to create a new logger every time one is not passed into a function.

danield9tqh:
We just have one logger. I think one advantage with passing it down is that we can subsequently call the `withTag` method ([example](https://github.com/iron-fish/ironfish/blob/staging/ironfish/src/node.ts#L167)). Then the logs will have all the tags of the components they are a part of. this would be beneficial for filtering out logs for a single component in Datadog for example.

hughy (approved):

PR #1497, opened 2022-05-24
Add logs to mining pool

danield9tqh:
## Summary
Adding some logs to the mining pool to figure out why pool is not sending logs for a long time

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/mining/pool.ts
+++ b/ironfish/src/mining/pool.ts
@@ -302,7 +302,8 @@ export class MiningPool {
   }
 
   private recalculateTarget() {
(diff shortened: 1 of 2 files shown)


ironfish/src/mining/pool.ts
```diff
@@ -302,6 +302,7 @@ export class MiningPool {
   }
 
   private recalculateTarget() {
+    this.logger.log('recalculating target')
```
NullSoldier, line 305: set to debug

mat-if (approved):

PR #1498, opened 2022-05-24
fix: removes default createRootLogger from RPC adapters and clients

hughy:
## Summary

requires that a logger be passed to each RPC adapter and client. passing a
logger makes it easier to configure logging.

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/rpc/adapters/ipcAdapter.test.ts
+++ b/ironfish/src/rpc/adapters/ipcAdapter.test.ts
@@ -5,6 +5,7 @@
 /* eslint-disable jest/no-conditional-expect */
 import os from 'os'
 import * as yup from 'yup'
+import { createRootLogger } from '../../logger'
 import { IronfishSdk } from '../../sdk'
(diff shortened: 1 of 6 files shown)


danield9tqh (approved):

NullSoldier:
I'm a little confused by this PR. I know it merged but I've read the code a few times and don't fully understand the benefit of the change. It does however break a pattern we use everywhere else in the code, so I can imagine running into this and inadvertently reverting this change to make it consistent back to the way the rest of the code base generally does it.

hughy:
> I know it merged but I've read the code a few times and don't fully understand the benefit of the change. It does deviate from a pattern we use everywhere else in the code, to default the logger in the component. I can imagine someone running into this and inadvertently reverting this change to make it consistent back to the way the rest of the code base generally does it.

I think that this may be a question of the desirability and tradeoffs of having the default loggers ⚖️ 

For the sake of consistency I agree that it may be better to revert these changes before someone runs into something unexpected

PR #1505, opened 2022-05-25
Fix incorrectly kicking peers if we have the TX

NullSoldier:
## Summary

There is a bug where users are being kicked from connections if they
rebroadcast transactions to us that we already have. The issue is that
wallest are programmed to rebroadcast transactions if we don't accept
them up until their expiry time. However, acceptTransaction() returns
false if we already have the transaction. We also punish any user
for if acceptTransaction() returns false. You can see the issue
here.

## Testing Plan
Update tests

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/network/peerNetwork.ts
+++ b/ironfish/src/network/peerNetwork.ts
@@ -701,6 +701,10 @@ export class PeerNetwork {
       message.message.transaction,
     )
 
+    if (this.node.memPool.exists(verifiedTransaction.hash())) {
+      return true
+    }
+
     const count = this.badMessageCounter.get(message.peerIdentity)
     if (await this.node.memPool.acceptTransaction(verifiedTransaction)) {
       await this.node.accounts.syncTransaction(verifiedTransaction, {})
--- a/ironfish/src/memPool/memPool.ts
+++ b/ironfish/src/memPool/memPool.ts
@@ -82,7 +82,7 @@ export class MemPool {
   async acceptTransaction(transaction: Transaction): Promise<boolean> {
     const hash = transaction.hash()
 
-    if (this.transactions.has(hash)) {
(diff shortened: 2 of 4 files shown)


ironfish/src/network/peerNetwork.ts
```diff
@@ -701,6 +701,10 @@ export class PeerNetwork {
       message.message.transaction,
     )
 
+    if (this.node.memPool.exists(verifiedTransaction.hash())) {
+      return true
+    }
```
NullSoldier, line 706: If we already have it, then the transaction is valid and should be gossiped out to all of our peers. This is the real fix.

lwisne: But wouldn't we have gossiped the tx the first time we saw it?

And then those nodes will receive the transaction again, and with this change, gossip again, etc.

NullSoldier: 1. You don't know if your peers have seen it or not, because your peers can have changed. That's the idea, we're helping some peer with a valid transaction get their peer out into the network
2. The messages won't get gossiped in a loop because we're dealing with the same message nonce. Each gossip message has a unique nonce that's checked against a bloom filter. If we ever see the same nonce again, we drop the message entirely. You can see that logic in these two places. https://github.com/iron-fish/ironfish/blob/master/ironfish/src/network/peerNetwork.ts#L313 and https://github.com/iron-fish/ironfish/blob/master/ironfish/src/network/peerNetwork.ts#L455
3. We expect and need to have transactions gossiped out again through our wallet rebroadcasting for network and transaction reliability. You can see that here https://github.com/iron-fish/ironfish/blob/master/ironfish/src/account/accounts.ts#L790

mat-if (approved):

PR #1506, opened 2022-05-25
Add ability to log to JSON in the console

danield9tqh:
## Summary
For a better way to parse the output of logs we want the ability to log to JSON in the console.

This PR also includes changes to log lines to make parsing logs safer. It will be easier to log in different formats if the logging interface is more well defined. This PR changes the Logger from taking (...args: any[]) to taking (message: string, ...args: []Record<string, string | number>). This will make it easier for us to convert our logs to JSON

## Testing Plan
Testing locally and unit tests

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/logger/index.ts
+++ b/ironfish/src/logger/index.ts
@@ -8,7 +8,13 @@ import { parseLogLevelConfig } from './logLevelParser'
 import { ConsoleReporter } from './reporters/console'
 export * from './reporters/intercept'
 
-export type Logger = Consola
+type Loggable = string | number
+export interface Logger extends Consola {
+  info(message: string, args?: Record<string, Loggable>): void
+  error(message: string, args?: Record<string, Loggable>): void
+  warn(message: string, args?: Record<string, Loggable>): void
+  debug(message: string, args?: Record<string, Loggable>): void
+}
 
 export const ConsoleReporterInstance = new ConsoleReporter()
 
@@ -24,6 +30,13 @@ export const setLogLevelFromConfig = (logLevelConfig: string): void => {
   for (const config of parsedConfig) {
     ConsoleReporterInstance.setLogLevel(config[0], config[1])
   }
+}
+
+/**
+ * @param logToJSON Whether console logs should be in JSON format
+ */
+export const setJSONLoggingFromConfig = (logToJSON: boolean): void => {
+  ConsoleReporterInstance.logToJSON = logToJSON
 }
 
 /**
--- a/ironfish/src/logger/index.ts
+++ b/ironfish/src/logger/index.ts
@@ -8,12 +8,20 @@ import { parseLogLevelConfig } from './logLevelParser'
 import { ConsoleReporter } from './reporters/console'
 export * from './reporters/intercept'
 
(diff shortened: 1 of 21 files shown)


ironfish/src/logger/reporters/console.ts
```diff
@@ -28,14 +28,34 @@ export const loggers: Record<logType, typeof console.log> = {
...
+
+  const toLog = {
+    ...objectArgs[0],
+    level: logObj.level,
+    tag: logObj.tag,
+    date: logObj.date,
+    message: otherArgs.join(' '),
   }
 
+  return JSON.stringify(toLog)
```
dguenther, line 51: You could potentially use `IJSON.stringify` here to avoid any potential issues if you pass in a bigint as an argument to the logger. There may be a try/catch preventing a crash though, I haven't looked to see

dguenther: Or to clarify, I think it'd be nice to use IJSON.stringify here so we don't have to refactor all of these logs -- there's a good chance someone could add a breaking log in the future since I expect most developers won't be running this flag enabled in production

ironfish/src/logger/index.ts
```diff
@@ -8,7 +8,13 @@ import { parseLogLevelConfig } from './logLevelParser'
...
 export * from './reporters/intercept'
 
-export type Logger = Consola
+type Loggable = string | number
+export interface Logger extends Consola {
+  info(message: string, args?: Record<string, Loggable>): void
+  error(message: string, args?: Record<string, Loggable>): void
+  warn(message: string, args?: Record<string, Loggable>): void
+  debug(message: string, args?: Record<string, Loggable>): void
+}
```
hughy, line 17: It doesn't look like there are many places where we haven't conformed to this interface in the past, but is there any concern over making the logging interface more restrictive to use?

danield9tqh: Yea just judging from our past usage I think this is not too big of a deal and it would force people to think about if it's really worth it to have something less structured in the logs. I'll add some comments here so that there is some explanation as to why it is restrictive. If we need to make it less restrictive in the future I think we can cross that bridge when we come to it

ironfish/src/rpc/adapters/tcpAdapter.ts
```diff
@@ -104,7 +104,7 @@ export class TcpAdapter implements IAdapter {
     const reqMap = new Map<string, Request>()
     this.pending.set(connId, { sock: socket, reqs: reqMap })
     socket.on('data', (data) => {
-      this.onClientData(socket, data, reqMap).catch((err) => this.logger.log(err))
+      this.onClientData(socket, data, reqMap).catch((err) => this.logger.error(err))
```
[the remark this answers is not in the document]
danield9tqh: Yea `err` here is an `any` value. i'm guessing there's going to be a few other places where this breaks but typescript doesn't catch it. Fortunately it doesn't actually break anything besides the type system. It will still get logged to JSON even if it doesn't conform to the interface

hughy (approved):

PR #1508, opened 2022-05-26
feat: implements secure adapter and client for TLS over TCP

hughy:
## Summary
extends TcpAdapter and IronfishTcpClient to establish TLS connection. expects
the node's private key and a self-signed cert to be located in the node's data
dir. the client is configured to allow the unauthorized, self-signed cert from
the node.

includes changes:
- refactors IronfishTcpClient and tests to make it possible to extend the class
  for TLS connections -- tls.Socket is a subclass of net.Socket, but doesn't
  implement its own 'connect' method, so the connection flow doesn't quite
  match the prior implementation
- adds config options for tlsKey, tlsCert, and enableRpcTls

it might be helpful/necessary to provide a script and documentation of how to
generate a key and cert for the node to use TLS. we'll also need to make some
changes to the Dockerfile to get a key and cert in the right place in the
containerized application.

## Testing Plan

manual testing with local nodes, cli commands

## Breaking Change

```
[ ] Yes
[X] No
```


ironfish/src/rpc/adapters/secureTcpAdapter.ts
```diff
@@ -0,0 +1,38 @@
...
+    super(host, port, logger, namespaces)
+    this.nodeKeyPath = nodeKeyPath
+    this.nodeCertPath = nodeCertPath
+  }
+
+  protected createServer(): net.Server {
+    const options = {
+      host: this.host,
+      port: this.port,
+      key: fs.readFileSync(this.nodeKeyPath),
```
danield9tqh, line 32: Can we do this `fs.readFileSync` maybe in the SDK and pass it down?

ironfish/src/rpc/clients/secureTcpClient.ts
```diff
@@ -0,0 +1,61 @@
...
+        } else if (ErrorUtils.isNoEntityError(error)) {
+          reject(new ConnectionRefusedError())
+        } else {
+          reject(error)
+        }
+      }
+
+      const options = {
+        // allows self-signed cert from the server
+        ca: [this.nodeCert],
```
hughy, line 47: This will change to `rejectUnauthorized: false` to allow self-signed certs without verifying certs on the client

PR #1518, opened 2022-05-26
GetUnspentNotes: de-couple spawning worker tasks from waiting for them to finish

mat-if:
## Summary

Before:
image

After:
image

## Breaking Change

```
[ ] Yes
[ ] No
```

### DIFF
--- a/ironfish/src/account/accounts.ts
+++ b/ironfish/src/account/accounts.ts
@@ -616,12 +616,18 @@ export class Accounts {
   ): Promise<ReadonlyArray<{ hash: string; note: Note; index: number | null }>> {
     const unspentNotes = []
 
+    const notePromises = []
     for (const transactionMapValue of this.transactionMap.values()) {
-      const result = await this.workerPool.getUnspentNotes(
-        transactionMapValue.transaction.serialize(),
-        [account.incomingViewKey],
+      notePromises.push(
+        this.workerPool.getUnspentNotes(transactionMapValue.transaction.serialize(), [
+          account.incomingViewKey,
+        ]),
       )
+    }
 
+    const results = await Promise.all(notePromises)
+
+    for (const result of results) {
       for (const note of result.notes) {
         const map = this.noteToNullifier.get(note.hash)
 


ironfish/src/account/accounts.ts
```diff
@@ -616,12 +616,18 @@ export class Accounts {
...
     for (const transactionMapValue of this.transactionMap.values()) {
-      const result = await this.workerPool.getUnspentNotes(
-        transactionMapValue.transaction.serialize(),
-        [account.incomingViewKey],
+      notePromises.push(
+        this.workerPool.getUnspentNotes(transactionMapValue.transaction.serialize(), [
+          account.incomingViewKey,
+        ]),
       )
+    }
```
brekk, line 626: Any reason this for loop couldn't be a `.map`?

```
const notePromises = this.transactionMap.values.map(
  x => this.workerPool.getUnspentNotes(
    x.transaction.serialize(), [account.incomingViewKey]
  )
)
```

brekk (approved):
Not a blocker but I prefer `.map`

ironfish/src/account/accounts.ts
mat-if: my js-fu is disgraceful 😢

mat-if: Oh, jk this is a generator, not an iterator

ironfish/src/account/accounts.ts
```diff
@@ -616,12 +616,18 @@ export class Accounts {
...
 
+    const notePromises = []
     for (const transactionMapValue of this.transactionMap.values()) {
-      const result = await this.workerPool.getUnspentNotes(
-        transactionMapValue.transaction.serialize(),
-        [account.incomingViewKey],
+      notePromises.push(
+        this.workerPool.getUnspentNotes(transactionMapValue.transaction.serialize(), [
+          account.incomingViewKey,
+        ]),
```
hughy, line 624: Does the order in which these promises are resolved matter?

mat-if: It shouldn't - all this is really doing is passing the transaction off to the worker threads to try decrypting the notes to find any that you own (can spend). It's being kicked off to the worker threads to keep the main event loop from being blocked by the decrypt code which takes a decent amount of time.

Since a note can only be spent once, order should be unimportant because we then immediately check if the decrypted notes have been spent from our note-nullifier map: https://github.com/iron-fish/ironfish/pull/1518/files#diff-975265306b3138d4067a44c1f74a98ef80249217a05ac91dff96236b482c429bR632

The actual decrypt note function: https://github.com/iron-fish/ironfish/blob/master/ironfish-rust-nodejs/src/structs/note_encrypted.rs#L80-L94

hughy (approved):

deekerno (approved):

ironfish/src/account/accounts.ts
```diff
@@ -616,12 +616,18 @@ export class Accounts {
...
       )
+    }
+
+    const results = await Promise.all(notePromises)
 
+    for (const result of results) {
```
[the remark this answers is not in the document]
mat-if: Starving the worker pool is potentially an issue here, although these jobs are fairly fast. If you'd feel more comfortable, we could do this in batches: x transactions at a time, await, process, x more, etc. For what it's worth, running 2 accounts:balance on a loop on my test chain isn't enough to back up my worker pool, although it does kick off ~2k jobs.

A potential medium-term fix could be to create a job to batch these so you're kicking off less overall jobs, spending less overhead on creating jobs, etc. I'm not sure if that's worth the effort given we're on verge of beginning the bigger accounts work.

I do agree that this isn't the long-term fix, for sure. The goal here is just a quick-win, whereas the better, bigger fix will likely come with the accounts stuff in the following weeks,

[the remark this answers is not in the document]
lwisne: On the topic of worker pool starvation (not the actual logic change Jason mentioned):

Does our workerPool not have a queueing system so you can call workerPool as many times as needed and it will just adjust to the resources it has? Seems bad that a client of the workerPool would have to think about starvation at all.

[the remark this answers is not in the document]
lwisne: My reading of the code is that we are still doing the same amount of worker tasks though? We are just artificially rate limiting by awaiting the result of each.

The idea that you have to call await to "protect" the workerPool seems pretty fundamentally misaligned with how I would expect a workerPool to function. We should just dump work on it, and if it can't handle the work we optimize how the pool works instead of having the workerPool client code figure out ways to gatekeep/rate limit.

Generally, I'm a bit surprised at the idea that workerPool usage feels like a bottleneck. I would expect us to have to be pretty finely optimized before running a task on a thread becomes a material amount of overhead.

mat-if:
Per discussion with Jason: this has potential risks of causing issues with people running nodes on VPS with only 1 or 2 worker threads due to RTC messages needing to box/unbox via worker jobs, transaction gossip, etc. So, current plan is to table this for now, and depending on how long it looks like it will be before we get some accounts refactors in, I'll consider bringing this back with some better logic to avoid causing issues with lower power nodes.

PR #1526, opened 2022-05-27
feat: adds script for generating key and cert for tls

hughy:
## Summary
starting a tcp server over TLS requires a private key and a public cert for the
server. this script executes the steps required to generate a private key and a
self-signed certificate for the server.

- uses default private key path of 'node-key.pem'
- uses default certificate path of 'node-cert.pem'
- uses default cert subject of '/O=Iron Fish', where 'O' refers to
  'organization'
  
this follows from the changes implemented for IRO-1498. the script generates the key and cert used to establish secure connections between the secure TCP adapter and secure client.

moving the key and cert into the node's data directory isn't within the purview
of this script.

## Breaking Change

```
[ ] Yes
[ ] No
```

### DIFF
--- a/ironfish-cli/scripts/generate-tls-keys.sh
+++ b/ironfish-cli/scripts/generate-tls-keys.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+if [ ! command -v openssl &> /dev/null ]; then
+  echo "openssl is not installed but is required"
+  exit 1
+fi
+
+keyPath=${keyPath:-"node-key.pem"}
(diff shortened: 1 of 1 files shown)


danield9tqh:
I'm ok with adding this to the repo but there are also some other options:
- Create separate repo / gist for the script and reference it somewhere
- Create an actual ironfish  command with heroku-cli. One downside is that this command depends on openssl but we already depend on openssl and doesn't seem like that's being removed anytime soon. If we add the `ironfish generate-certificate` command or something we'd probably want that to move the certificate to the correct dataDir as well

PR #1532, opened 2022-05-28
graph-exploreruse: loading js assets use encrypted connection

Altonhe:
## Summary
ironfish graph explorer is using unpkg as a content delivery network for javascript assets, since unpkg is support https, we should use a security connection here.

## Breaking Change

```
[ ] Yes
[ √ ] No
```


NullSoldier:
This is already completed, you are using a stale branch off master. Please see what index.html looks like here, https://github.com/iron-fish/ironfish/blob/staging/ironfish-graph-explorer/src/index.html

PR #1537, opened 2022-05-29
fix: transaction errors & account balance issues

wd021:
## Summary
Many bugs have been reported relating to tx failures and account balance problems. I've hit these myself quite consistently.

While debugging, I found that https://github.com/iron-fish/ironfish/blob/master/ironfish/src/account/accounts.ts#L714-L740 in `accounts/createTransaction` is causing the problem. This wouldn't update successfully, the tx would then fail with `Insufficient funds`, and querying the account balance afterward would show incorrect values. 

`nullifier.toString('hex')` should be `unspentNote.hash` 👇 
```
          // Update our map so this doesn't happen again
          const noteMapValue = this.noteToNullifier.get(nullifier.toString('hex'))
          if (noteMapValue) {
            this.logger.debug(`Unspent note has index ${String(noteMapValue.noteIndex)}`)
            await this.updateNoteToNullifierMap(nullifier.toString('hex'), {
              ...noteMapValue,
              spent: true,
            })
          }
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/account/accounts.ts
+++ b/ironfish/src/account/accounts.ts
@@ -734,10 +734,10 @@ export class Accounts {
           )
 
           // Update our map so this doesn't happen again
-          const noteMapValue = this.noteToNullifier.get(nullifier.toString('hex'))
+          const noteMapValue = this.noteToNullifier.get(unspentNote.hash)
           if (noteMapValue) {
             this.logger.debug(`Unspent note has index ${String(noteMapValue.noteIndex)}`)
-            await this.updateNoteToNullifierMap(nullifier.toString('hex'), {
+            await this.updateNoteToNullifierMap(unspentNote.hash, {
               ...noteMapValue,
               spent: true,
             })


lwisne:
How would we recreate the error that this fixes?

Ideally, we could get this as an automated test, but if you can explain that would be helpful as well.

wd021:
If you send enough txs one after another, you will hit the `unspentNote.note.value() > BigInt(0)` condition. I haven't dug in to figure out the exact upstream conditions that result in this. As the code comment indicates
```
// Double-check that the nullifier for the note isn't in the tree already
// This would indicate a bug in the account transaction stores
```
so I suspect it has to do with account transaction stores.

But at least we can fix the map update when we do hit this

ironfish/src/account/accounts.ts
```diff
@@ -734,10 +734,10 @@ export class Accounts {
           )
 
           // Update our map so this doesn't happen again
-          const noteMapValue = this.noteToNullifier.get(nullifier.toString('hex'))
+          const noteMapValue = this.noteToNullifier.get(unspentNote.hash)
           if (noteMapValue) {
             this.logger.debug(`Unspent note has index ${String(noteMapValue.noteIndex)}`)
-            await this.updateNoteToNullifierMap(nullifier.toString('hex'), {
+            await this.updateNoteToNullifierMap(unspentNote.hash, {
```
NullSoldier, line 740: Great find, thanks for making this fix.

NullSoldier (approved):

NullSoldier:
Thanks for this fix. I want to make some notes here because we need more work done in this space. Overall, I this code was added as a workaround to a larger problem where we were trying to double spend notes that were already spent. Ideally this entire code block should be deleted.

I think the main issue is that we don't have a lock around getting unspent notes, and marking them as spent. It makes sense because @wd021 also says it happens when creating transactions in a loop.

If we also look at the modifications to accounts that hpool made, they also add a lock around `Accounts.pay()` and also cache unspent notes in memory as an optimization. Maybe when we refactor some of this these issues can be fixed with the above solutions.

PR #1545, opened 2022-05-31
Removed unreachable code

hamza-siddiq:
## Summary
ironfish-cli/src/commands/swim.ts: Removed unreachable code.

## Breaking Change

```
[ ] Yes
[ ✓] No
```
graffiti: ironfishup


mat-if:
I'm going to echo the same comment as I left here: https://github.com/iron-fish/ironfish/pull/1546#issuecomment-1142350014

A better PR would be a fix to our lint checker which runs on all of our pull requests such that it would catch this lint issue

PR #1549, opened 2022-05-31
Added the u flag

hamza-siddiq:
## Summary
Added the u flag with regular expressions, which enables correct handling of UTF-16 surrogate pairs, and ensures the correct behavior of regex character ranges.

## Breaking Change

```
[ ] Yes
[ ✓] No
```
graffiti: ironfishup

### DIFF
--- a/ironfish/src/utils/memo.ts
+++ b/ironfish/src/utils/memo.ts
@@ -3,7 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 function toHuman(memo: string): string {
-  return memo.replace(/[\u0000-\u001F\u007F-\u009F]/g, '').trim()
+  return memo.replace(/[\u0000-\u001F\u007F-\u009F]/gu, '').trim()
 }
 
 export const MemoUtils = {
--- a/ironfish/src/utils/graffiti.ts
+++ b/ironfish/src/utils/graffiti.ts
@@ -13,7 +13,7 @@ function fromString(graffiti: string): Buffer {
 function toHuman(graffiti: Buffer): string {
   return graffiti
     .toString('utf8')
-    .replace(/[\u0000-\u001F\u007F-\u009F]/g, '')
+    .replace(/[\u0000-\u001F\u007F-\u009F]/gu, '')
     .trim()
 }
 


hamza-siddiq:
I would recommend having the u flag, rather than removing it. JavaScript regular expressions are tolerant of syntax errors. For example, /\w{1, 2/ is a syntax error, but JavaScript doesn't throw the error. It matches strings such as "a{1, 2" instead. Such a recovering logic is defined in Annex B of Javascript specification.The u flag disables the recovering logic Annex B of the Javascript specification. As a result, you can find errors early.

ironfish/src/utils/memo.ts
```diff
@@ -3,7 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 function toHuman(memo: string): string {
-  return memo.replace(/[\u0000-\u001F\u007F-\u009F]/g, '').trim()
+  return memo.replace(/[\u0000-\u001F\u007F-\u009F]/gu, '').trim()
 }
```
NullSoldier, line 7: Doesn't this cause it to break, because `\u001F\u007F` will now be interpreted as a single surrogate pair? This needs a test.

NullSoldier, line 7: Doesn't this cause it to break, because `\u001F\u007F` will now be interpreted as a single surrogate pair? This needs a test.

NullSoldier:
I think the PR means well, but the PR author hasn't given evidence that this fixes or improves an existing problem. I'm going to close it but I'm also fine if they reopen this with some tests that demonstrate it's value. I don't really buy that this is a necessary change. The next developer to add a regex is going to forget to do this because javascript developers don't really use this because they rarely match unicode points in a regex pattern itself. In the examples above, we aren't even matching multi byte unicode characters therefore this doesn't affect it.

1. The problem listed is hypothetical and isn't present in the places that were changed
2. If the fix is necessary, then there should be tests to demonstrate that if the necessary code is removed something is broken
3. The example given implies that no tests were written as if this is the only thing that can find a typo in a regex.
4. An experienced developer will come through, and see that this isn't doing anything and remove it. It is generally correct to remove unused code.
5. The fix is not clear what it's fixing, and if you look around at general regex usage you don't really see people using this flag when they aren't using unicode code points, and yet all those regex's still work fine because they don't suffer from unicode issues

PR #1550, opened 2022-05-31
Comparisons fixed

hamza-siddiq:
## Summary
Compared to null with a type-checking operator in multiple files. Comparing to null without a type-checking operator will evaluate to true when comparing to not just a null, but to an undefined value as well.

## Breaking Change

```
[ ] Yes
[ ✓] No
```
graffiti: ironfishup

### DIFF
--- a/ironfish-cli/src/commands/accounts/pay.ts
+++ b/ironfish-cli/src/commands/accounts/pay.ts
@@ -77,7 +77,7 @@ export class Pay extends IronfishCommand {
       this.exit(1)
     }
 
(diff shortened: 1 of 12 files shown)


mat-if:
This is the idiomatic way of checking for null and undefined values. If you look closely at the types of the objects that you're changing these checks on, you'll see that many of them _are_ expected to be `undefined` and not `null`. This change breaks most of the places you've touched. 

In the future, please make sure you are testing your code changes before you put them up.

PR #1552, opened 2022-05-31
Sending transactions in version 0.1.35

OlegBorodiy:
Graffiti alexnode

A dedicated server was used for testing:
CPU AMD EPYC 7502P
Cores (vCores) 32 (64)
Performance per core 2.5 GHz
Total output 80.000
Memory 64 GB DDR4
Hard disks 2 x 2,000 GB SATA
Connection 10,000 Mbit

The miner and node were installed on the same server. The time of sending transactions when working on all threads reached up to 2780 seconds. The situation is similar with VPS servers.


mat-if:
I'm assuming you meant to file an issue and not a pull request? https://github.com/iron-fish/ironfish/issues

If this was meant to be a pull request, I think you've got the branches set incorrectly.

PR #1553, opened 2022-05-31
pool: limit number of concurrent tcp connections per ip

rpanic:
## Summary

Introduce a tracker which counts the number of active connections to the pool per external IP-address. If the maximum number is reached, decline future transactions. 

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/mining/stratum/stratumServer.ts
+++ b/ironfish/src/mining/stratum/stratumServer.ts
@@ -334,7 +334,7 @@ export class StratumServer {
     }
 
     const connectionsByIp = this.openConnections.get(socket.remoteAddress) ?? 0
-    if (connectionsByIp >= this.maxOpenConnections) {
+    if (this.maxOpenConnections > 0 && connectionsByIp >= this.maxOpenConnections) {
       return false
     }
 
--- a/ironfish/src/mining/stratum/stratumServer.ts
+++ b/ironfish/src/mining/stratum/stratumServer.ts
@@ -103,38 +103,21 @@ export class StratumServer {
     this.send(client, 'mining.wait_for_work')
   }
 
-  addToConnectionCount(remoteAddress: string, addition: number): number {
-    this.openConnections.set(
-      remoteAddress,
-      (this.openConnections.get(remoteAddress) ?? 0) + addition,
-    )
-    return this.openConnections.get(remoteAddress) ?? 0
-  }
-
   private onConnection(socket: net.Socket): void {
-    if (!socket.remoteAddress) {
+    if (!this.isSocketAllowed(socket)) {
       socket.destroy()
       return
     }
 
     const client = StratumServerClient.accept(socket, this.nextMinerId++)
 
-    if ((this.openConnections.get(client.remoteAddress) ?? 0) + 1 > this.maxOpenConnections) {
-      //Set timeout for disconnect in order to prevent spamming, then close the connection
-      setTimeout(() => {
-        client.close()
-      }, 10000)
-      return
-    }
-
-    this.addToConnectionCount(client.remoteAddress, 1)
+    this.addConnectionCount(client)
 
     socket.on('data', (data: Buffer) => {
       this.onData(client, data).catch((e) => this.onError(client, e))
     })
 
     socket.on('close', () => this.onDisconnect(client))
-
     socket.on('error', (e) => this.onError(client, e))
 
     this.logger.debug(`Client ${client.id} connected: ${client.remoteAddress}`)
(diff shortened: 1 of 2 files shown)


ironfish/src/mining/stratum/stratumServer.ts
```diff
@@ -127,9 +132,27 @@ export class StratumServer {
...
+
+    if((this.openConnections.get(remoteAddress) ?? 0) + 1 > this.maxOpenConnections){
+      //Set timeout for disconnect in order to prevent spamming, then close the connection
+      setTimeout(() => {
+        client.close()
+      }, 10000)
+      return
+    }
+
+    this.addToConnectionCount(remoteAddress, 1)
```
mat-if, line 154: A few more suggestions:
1. We probably don't want to add a `remoteAddress` to the map if it isn't a falsy value (undefined, etc), otherwise we run the risk of flagging connections that are valid but not returning proper values for whatever reason. If this still ends up being an issue, we probably want to look into better ways to bucket connections.
2. Just some minor readability improvements, white space, splitting some things into vars etc.
3. Instead of doing the `+ 1 >`, we can just do `>=`
4. Adding the client to the `badClient` list so we don't send them messages also, this also prevents a possible issue where a miner connects, thinks everything is good, and then gets disconnected 10 seconds later. This would surface more dramatically, they would never fully connect and get broadcasts (although I think they'd still get the initial work/subscribed message now that I'm thinking about it - possible future improvement).

```suggestion
    const remoteAddress = socket.remoteAddress
    
    if (remoteAddress) {
      const connectionCount = this.openConnections.get(remoteAddress) ?? 0
      if (connectionCount >= this.maxOpenConnections) {
        // Add the client to the bad list so we don't waste resources sending them messages
        this.addBadClient(client)
        // Set timeout for disconnect in order to prevent spamming, then close the connection
        setTimeout(() => {
          this.clients.delete(client.id)
          client.close()
        }, 10000)
        
        return
      }
      
      this.addToConnectionCount(remoteAddress)
    }
```

ironfish/src/mining/stratum/stratumServer.ts
```diff
@@ -158,6 +181,7 @@ export class StratumServer {
     this.logger.debug(`Client ${client.id} disconnected  (${this.clients.size - 1} total)`)
     this.clients.delete(client.id)
     client.close()
+    this.addToConnectionCount(client.socket.remoteAddress ?? "", -1)
```
mat-if, line 184: Suggestion related to the above change of splitting these fns out

```suggestion
    if (client.socket.remoteAddress) {
      this.removeFromConnectionCount(client.socket.remoteAddress)
    }
```

ironfish/src/mining/stratum/stratumServer.ts
```diff
@@ -245,6 +269,7 @@ export class StratumServer {
     client.socket.removeAllListeners()
     client.close()
     this.clients.delete(client.id)
+    this.addToConnectionCount(client.socket.remoteAddress ?? "", -1)
```
mat-if, line 272: Suggestion related to the above change of splitting these fns out

```suggestion
    if (client.socket.remoteAddress) {
      this.removeFromConnectionCount(client.socket.remoteAddress)
    }
```

mat-if (changes requested):
Interesting, thanks for this PR! I'm always a big fan of defensive code 😄  I hadn't really considered doing this, but I haven't paid too much attention to specific connections. Are you finding this is an issue? One possible unintended consequence here is that it could block out people who just genuinely have a lot of mining machines. 

My suggestion would be to make the default max set to `0`, and if it's set to 0, it would disable any limit, so it's configurable for those who want it, but won't cause potential issues for those who are unaware of it.

One other note, I noticed some things that our linter/formatter will catch - for this, I highly recommend installing an eslint-prettier plugin for your code editor. If you use VSCode, I use these 2:
https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint
https://marketplace.visualstudio.com/items?itemName=rvest.vs-code-prettier-eslint
and set it to format on save in the settings as shown in the picture below. It's not always my favorite style, but it's the style our linter enforces, plus I just love not having to think about formatting ever again.
image

ironfish/src/mining/stratum/stratumServer.ts
```diff
@@ -63,8 +63,11 @@ export class StratumServer {
   readonly port: number
   readonly host: string
 
+  readonly maxOpenConnections: number
```
NullSoldier, line 66: You can have more than this number of open connections. This doesn't match the behaviour. In your config it's named, "maxConnectionsPerIp", but here it seems to indicate a global max.

mat-if (approved):

NullSoldier:
I addressed the feedback and I'm merging the PR.

PR #1555, opened 2022-05-31
Reset chainProcessor hash if index head not found in database

deekerno:
## Summary
The head hash of `MinedBlockIndexer`'s meta table should be reset instead of throwing an error, if its head hash in is not found in the chain.

## Testing Plan
Reproduction, as per @NullSoldier:
1. Start the node and sync blocks
2. Delete the `databases` folder
3. Start the node

Ensured that the node now starts properly and logs an error message.

## Breaking Change

```
[ ] Yes
[x] No
```


mat-if (approved):
🙌

PR #1558, opened 2022-06-01
Add way to ad-hoc save logs to local database

danield9tqh:
## Summary
We need way to easily and flexibly get visibility into stats about the node. How many transaction messages are we receiving from nodes? Are there duplicate gossip messages? Are certain peers sending more than others?

We have a few different ways that we keep track of node stats currently (like [Meter](https://github.com/iron-fish/ironfish/blob/staging/ironfish/src/metrics/meter.ts)). But these stats are stored in memory and are not accessible to other systems without creating RPC endpoints. They also aren't as flexible as a SQL database.

For gaining visibility into gossip messages, I would like to create an easy way to save ad-hoc events to a database (like new transaction/block message) to a database to be able to analyze them more in-depth. Ideally this approach will be helpful for future use-cases as well

Here is an example of how a log line might be modified to use this database.
Original log line:
```typescript
    this.logger.info('INCOMING_TRANSACTION', {
      hash,
      inMemPool,
      peer
    })
```

Modified log line:
```typescript
    this.logger.info('INCOMING_TRANSACTION', {
      hash,
      inMemPool,
      peer,
      dbTable: 'transactions',
    })
```

This will create a database when the node starts up, a new table `transactions` and insert all the log lines into this table. The database table will be in the `.ironfish` data directory

You also need to add the dbReporter to the current logger
```typescript
const dbReporter = await createNewLogsDBReporter(config.dataDir)
logger = logger.addReporter(dbReporter)
```
## Testing Plan
Local testing

## Breaking Change

```
[ ] Yes
[ ] No
```

### DIFF
--- a/ironfish/src/logger/index.ts
+++ b/ironfish/src/logger/index.ts
@@ -6,7 +6,7 @@ import type { Consola } from 'consola'
 import consola, { LogLevel } from 'consola'
 import { parseLogLevelConfig } from './logLevelParser'
 import { ConsoleReporter } from './reporters/console'
-export * from './reporters/intercept'
+export * from './reporters'
 
 /**
  * This interface tries to enforce more structured logs while still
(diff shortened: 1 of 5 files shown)


NullSoldier:
This seems like a pretty unique way to ingest, parse, and debug logs that I haven't seen before. What's the advantage over this new way instead of doing what most people do which is just use text / json logs or ingest them into one of hundreds of log services and use those log indexing tools? Most people use elastic search and kibana rather than sqlite because you can structure the metadata.

danield9tqh:
The advantage of using SQL instead of Kabana/Elastic Search for my use case is that I wanted to `join` on different events. Log services just aren't very good at that I don't think or doing complex queries for that matter. 

As far as adding a different service for ingesting logs, I agree. I think that's probably the best long term solution. One reason to have a solution in the node itself though might be if we want to keep track of more complex stats for the node to show real time in a wallet UI for example. In that wallet UI I could see potentially needing to store stuff in another DB.

The reason I didn't add a separate service for this change is mostly for speed and developer convenience. If we had to run another service I think it would be too cumbersome for anyone to use. Just adding a line `dbTable: '...'` to the log seemed like a very simple thing for developers to do so that this might actually get used.

I also made sure to not add ANY code related to this inside the node itself. It's all contained in a `ConsolaReporter` object. Basically this means that all the logs are treated as JSON up until the very last step where we decide whether to send them to stdout, a file or this database. This pattern should make it very easy to move these into separate services in the future. I'm definitely aware/weary of adding too much logic to the node itself which is why I made the changes last week to try to standardize our log structures and optionally format them to JSON so that we could add external services that ingest them in the future.

danield9tqh:
After discussion we decided to add this to a separate project / repo as an external log parser. Adding it to this repo it might go unused and turn into dead code

PR #1560, opened 2022-06-02
Remove redundant `sendTo` function

danield9tqh:
## Summary
Remove redundant `sendTo` function

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/network/peers/peerManager.ts
+++ b/ironfish/src/network/peers/peerManager.ts
@@ -750,15 +750,6 @@ export class PeerManager {
   }
 
   /**
-   * Send a message to a peer, dropping the message if unable.
-   * @param peer The peer identity to send a message to.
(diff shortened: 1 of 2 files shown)


hughy (approved):

ironfish/src/network/peers/peerManager.ts
```diff
@@ -749,15 +749,6 @@ export class PeerManager {
...
   }
 
-  /**
-   * Send a message to a peer, dropping the message if unable.
-   * @param peer The peer identity to send a message to.
-   * @param message The message to send.
-   */
-  sendTo(peer: Peer, message: NetworkMessage): Connection | null {
-    return peer.send(message)
-  }
```
NullSoldier, line 759: We might have to bring this back later, but yeah it's not used right now.

PR #1569, opened 2022-06-02
feat: adds a message buffer to the stratum client

hughy:
## Summary

protects against reading partial messages from the TCP socket stream.

stores any partial messages read from the socket in the message buffer and
concatenates the next read onto the message buffer.

background:
tcp sockets are modeled on streams of data, and the data that the server writes
to the socket may be broken apart during transmission. the reader is guaranteed
to receive data in the same order that the writer sends it, but not necessarily
in a contiguous chunk.

- uses last index of message delimiter ('\n') to identify beginning of any
  partial messages
- clears message buffer on socket disconnect

## Testing Plan

the attached patch reproduces the error when running a pool and miner locally. the changes here resolve the error

[stratum_tcp_test.patch.txt](https://github.com/iron-fish/ironfish/files/8828041/stratum_tcp_test.patch.txt)

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/mining/stratum/stratumClient.ts
+++ b/ironfish/src/mining/stratum/stratumClient.ts
@@ -33,6 +33,7 @@ export class StratumClient {
   private connectWarned: boolean
   private connectTimeout: SetTimeoutToken | null
   private nextMessageId: number
+  private messageBuffer = ''
 
   private readonly publicAddress: string
 
@@ -146,6 +147,7 @@ export class StratumClient {
 
   private onDisconnect = (): void => {
     this.connected = false
+    this.messageBuffer = ''
     this.socket.off('error', this.onError)
     this.socket.off('close', this.onDisconnect)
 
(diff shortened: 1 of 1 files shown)


mat-if:
I'm wondering if `setNoDelay` is the real fix here, I was trying to find a manual flush/drain function when we were looking at this earlier, I didn't see `setNoDelay`. In mining, especially with faster block times, latency is the biggest enemy. It doesn't make sense for our use of sockets to be delayed, and we're not sending _that_ many messages back and forth that I'd expect us to run into issues by sending them too frequently.

mat-if (approved):
Doing some more testing, I think that the buffering is out of our control due to the nature of TCP. Tested this by setting a breakpoint on the miner and letting it sit while the pool adjusted the work a couple of times. This still led to buffering even with no delay, which makes sense. If there are lag spikes, we expect there to be some buffering and re-sending. So while we may be able to benefit a bit from `setNoDelay`, it won't help in the cases of lag spikes or dropped packets or whatever.

dguenther (approved):
Nice fix!

PR #1571, opened 2022-06-03
feat(ironfish): Check for minimum confirmed blocks when fetching balance

rohanjadvani:
## Summary

When fetching unspent notes and computing balance for an account, check if enough blocks have been confirmed.

## Testing Plan

Added unit test.

## Breaking Change

```
[ ] Yes
[x] No
```


ironfish/src/account/accounts.ts
```diff
@@ -630,10 +642,21 @@ export class Accounts {
         }
 
         if (!map.spent) {
+          let confirmedMinimumBlocks = false
+
+          const { blockHash } = transactionMapValue
```
NullSoldier, line 647: nit: do we need to create more variables in this scope, can we just use `transactionMapValue.blockHash`?

ironfish/src/account/accounts.ts
```diff
@@ -630,10 +642,21 @@ export class Accounts {
...
+          if (blockHash) {
+            const block = await this.chain.getBlock(Buffer.from(blockHash, 'hex'))
```
NullSoldier, line 649: We don't use the transactions so we can just get the header here instead of the whole block.

ironfish/src/account/accounts.ts
```diff
@@ -630,10 +642,21 @@ export class Accounts {
...
+            Assert.isNotNull(block)
+            const confirmations = this.chain.head.sequence - block.header.sequence
+            confirmedMinimumBlocks = confirmations >= minimumBlockConfirmations
```
NullSoldier, line 652: Isn't this code not correct, as we would consider it confirmed even if it's on a fork?

ironfish/src/account/accounts.ts
NullSoldier: We're having really bad perf problems right now with getBalance(), and I think this might make it significantly worse if we're loading every block for ever note that's unspent. I think we should consider this, and test it using the bank wallet to see the difference.

ironfish/src/account/accounts.ts
```diff
@@ -630,10 +642,21 @@ export class Accounts {
         }
 
         if (!map.spent) {
+          let confirmedMinimumBlocks = false
```
NullSoldier, line 645: nit: can we just call this something simple like confirmed?

ironfish/src/account/accounts.ts
rohanjadvani: Good call, updated.

ironfish/src/account/accounts.ts
rohanjadvani: Works for me, updated.

ironfish/src/account/accounts.ts
rohanjadvani: Yeah, this could be considered confirmed if on a fork. However, the idea here was choosing a minimum threshold where we have some guarantee of low reversibility after some confirmations. There seems to be similar logic for other projects, where confirmations are checked against a similar threshold when getting [spendable inputs](https://github.com/zcash/zcash/blob/master/src/wallet/wallet.cpp#L2395) or computing [account balance](https://github.com/zcash/zcash/blob/master/src/wallet/wallet.cpp#L4810). These pieces of logic also seem to use the [active chain's height](https://github.com/zcash/zcash/blob/master/src/wallet/wallet.cpp#L6688).

NullSoldier: This looks different than zcash, where their confirmations are calculated based on 
 depth in the main chain. I also don't think we should consider TX on forks confirmed, because that money is not spendable. Any transactions we create with those notes are going to be considered invalid, even to us.

 - https://github.com/zcash/zcash/blob/master/src/wallet/wallet.cpp#L2393
 - https://github.com/zcash/zcash/blob/master/src/wallet/wallet.cpp#L6672-L6676

NullSoldier:
I think this is going to affect a performance sensitive part of the code base, which is already causing significant performance issues in production. Should we create a perf test that checks wallet balance calculation with a certain amount of notes + transactions after this merges?

ironfish/src/account/accounts.ts
rohanjadvani: Agreed, I am not a fan of `transactionMapValue` either. I refactored the loop to use `blockHash` and `transaction` directly [here](https://github.com/iron-fish/ironfish/pull/1571/files#diff-975265306b3138d4067a44c1f74a98ef80249217a05ac91dff96236b482c429bR631).

ironfish/src/account/accounts.ts
rohanjadvani: Added a main chain [check](https://github.com/iron-fish/ironfish/pull/1571/files#diff-975265306b3138d4067a44c1f74a98ef80249217a05ac91dff96236b482c429bR649-R653).

rohanjadvani:
> I think this is going to affect a performance sensitive part of the code base, which is already causing significant performance issues in production. Should we create a perf test that checks wallet balance calculation with a certain amount of notes + transactions after this merges?

That sounds good. Added a ticket for tracking [here](https://linear.app/ironfish/issue/IRO-2123/add-performance-test-for-checking-account-balance).

ironfish/src/account/accounts.test.ts
```diff
@@ -97,4 +97,61 @@ describe('Accounts', () => {
     invalidTxEntry = nodeA.accounts['transactionMap'].get(invalidTx.hash())
     expect(invalidTxEntry?.submittedSequence).toEqual(blockB2.header.sequence)
   }, 120000)
+
+  describe('getBalance', () => {
```
mat-if, line 101: Should we move this to `accounts.tests.slow.ts` given it's got a 60s timeout? I know we haven't made the cleanest distinction between the 2, or if it's even a distinction we care to keep.

mat-if (approved):
Looks good. The perf improvements coming up next should negate any slowdown this introduces 🙌

ironfish/src/account/accounts.test.ts
rohanjadvani: That's a good point. It seems like some normal tests include the timeout, so I think we should decide on some rough guideline to follow. I can move this test over, though.

rohanjadvani: Turns out the timeout wasn't needed ¯\_(ツ)_/¯

PR #1575, opened 2022-06-04
feat: adds a global flag for enabling TLS in RPC

hughy:
## Summary

- adds 'rpc.tcp.tls' flag, with a 'no-rpc.tcp.tls' option. uses 'rpc.tcp.tls'
rather than just 'rpc.tls' since the TLS connection only applies to TCP. adds
the flag as enabling, rather than disabling, TLS so that the logic matches the
existing flags
- enables TLS for TCP connections by default
- removes 'enableNativeRpcTcpAdapter' option

## Breaking Change

```
[ ] Yes
[X] No
```


ironfish-cli/src/command.ts
```diff
@@ -127,6 +130,11 @@ export abstract class IronfishCommand extends Command {
       configOverrides.rpcTcpSecure = rpcTcpSecureFlag
     }
 
+    const rpcConnectTlsFlag = getFlag(flags, RpcUseTlsFlagKey)
```
NullSoldier, line 133: nit: This is small issue but it doesn't match the flag name like all the other flags here

NullSoldier (approved):
It's good besides some name consistency stuff

PR #1578, opened 2022-06-05
Updated node to version 18.2.0

hamza-siddiq:
## Summary
Updated node to the latest stable version 18.2.0. The previous version 16.13, which was outdated by 26 versions.
## Testing Plan
None needed
## Breaking Change
None

```
[ ] Yes
[✓ ] No
```
graffiti: ironfishup


drunkensiberia:
I don't think that there is need to update node to new version. First of all v16 is stable LTS version, and many of Ironfish structure depending on v16. So you can't just change this in Dockerfile.

PR #1581, opened 2022-06-06
Add TcpAdapter.started from IpcAdapter

NullSoldier:
## Summary
I'm trying to bring parity to the two of them

## Testing Plan
Run adapter and user commands

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/rpc/adapters/tcpAdapter.ts
+++ b/ironfish/src/rpc/adapters/tcpAdapter.ts
@@ -47,6 +47,7 @@ export class TcpAdapter implements IAdapter {
   router: Router | null = null
   namespaces: ApiNamespace[]
 
+  started = false
   pending = new Map<string, { sock: net.Socket; reqs: Map<string, Request> }>()
 
   constructor(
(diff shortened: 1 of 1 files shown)


danield9tqh:
Do we just use this to decide whether to execute the `stop` method or not? If that's the only use case maybe we can just make the stop method work whichever state it's called in. It might already work regardless. If it hasn't started yet we shouldn't have anything in the `pending` list and we also check to see if the server exists before trying to close it

NullSoldier:
I'm just trying to make them parallel, because we avoided switching over to the native TCPAdapter in the last deploy because we were seeing errors right? I want to make them parallel in terms of behaviour, because we already trust the IPCAdapter. After that we can start factoring things out and added more general reliability and fixes.

I think your suggestion is reasonable, but this is guaranteed to be a bit safer. If you do it your way, you need to add a lock as well so any async logic doesn't execute in parallel.

hughy (approved):
@danield9tqh may be right that this isn't strictly necessary for `stop` to work as expected, but I support the idea of making IpcAdapter and TpcAdapter consistent in their implementations, and then abstracting.

PR #1584, opened 2022-06-06
Start wrapping clients in Client abstraction

NullSoldier:
## Summary

I noticed the contents of pending weren't pending requests, but
clients. I started tightening that abstraction by making it an
abstraction, similar to StratumServer.

## Testing Plan

Run TCP adapter and run commands

## Breaking Change

```
[ ] Yes
[x] No
```


ironfish/src/rpc/adapters/tcpAdapter.ts
```diff
@@ -39,6 +40,12 @@ export const IncomingNodeIpcSchema: yup.ObjectSchema<IncomingNodeIpc> = yup
   })
   .required()
 
+type TcpAdapterClient = {
```
danield9tqh, line 43: nit: maybe name Connection instead of Client but either seems fine

ironfish/src/rpc/adapters/tcpAdapter.ts
```diff
@@ -110,33 +117,45 @@ export class TcpAdapter implements IAdapter {
...
+  onClientError(client: TcpAdapterClient, error: unknown): void {
+    this.logger.debug(`${this.host}:${this.port} has error: ${ErrorUtils.renderError(error)}`)
+  }
+
   async onClientData(
-    socket: net.Socket,
+    client: TcpAdapterClient,
     data: Buffer,
-    reqMap: Map<string, Request>,
+    pending: Map<string, Request>,
```
hughy, line 152: I don't think `pending` is needed here now that the requests are part of the `TcpAdapterClient` instance

ironfish/src/rpc/adapters/tcpAdapter.ts
NullSoldier: Isn't the client.socket the connection? What do you generally consider a client? If this only represents a connection, is there no client / user concept?

ironfish/src/rpc/adapters/tcpAdapter.ts
NullSoldier: This code is really hairy, I wasn't sure if I wanted to clean it up in this PR or wait til the next one. I'm down to do it now, as long as you are down to see more code change in this PR.

ironfish/src/rpc/adapters/tcpAdapter.ts
```diff
@@ -110,33 +117,45 @@ export class TcpAdapter implements IAdapter {
...
   ): Promise<void> {
     const dataString = data.toString('utf8').trim()
     const result = await YupUtils.tryValidate(IncomingNodeIpcSchema, dataString)
 
     if (result.error) {
-      this.emitResponse(socket, this.constructUnmountedAdapter(), reqMap)
+      this.emitResponse(client.socket, this.constructUnmountedAdapter(), pending)
```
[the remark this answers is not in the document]
NullSoldier: I'm down to do that in the current PR, I was waiting but I'll make that change.

ironfish/src/rpc/adapters/tcpAdapter.ts
NullSoldier: DONE

ironfish/src/rpc/adapters/tcpAdapter.ts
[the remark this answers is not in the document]
NullSoldier: DONE

PR #1586, opened 2022-06-06
Add transaction hash to pool payouts table

danield9tqh:
## Summary
Currently pool payouts don't record which transaction hash they were completed with. This is causing some issues when trying to debug how the payout was completed. We are also running into an issue where some transaction hashes cannot be found and there is no way to work backwards to tell which payouts they were associated with

## Testing Plan
Local testing

DB Fiddle testing the queries: https://www.db-fiddle.com/f/hk54YMjkve6A2QiwvsdBrT/2

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/mining/poolShares.ts
+++ b/ironfish/src/mining/poolShares.ts
@@ -4,6 +4,7 @@
 import { Config } from '../fileStores/config'
 import { Logger } from '../logger'
 import { IronfishRpcClient } from '../rpc/clients/rpcClient'
+import { ErrorUtils } from '../utils'
 import { BigIntUtils } from '../utils/bigint'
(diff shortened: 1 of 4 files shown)


ironfish/src/mining/poolShares.ts
```diff
@@ -185,7 +186,7 @@ export class MiningPoolShares {
         shareCounts.totalShares,
       )
     } catch (e) {
-      this.logger.error('There was an error with the transaction', e)
+      this.logger.error(`There was an error with the transaction ${ErrorUtils.renderError(e)}`)
```
danield9tqh, line 189: this is an unrelated change but just updating this here anyway

hughy (approved):

PR #1589, opened 2022-06-07
Factor out socket RPC code into SocketAdapter

NullSoldier:
## Summary

This is the start of de-duplicating some of our net.Socket based RPC
adapoter code. It creates a new SocketAdapter and renames all the
underlying types from IPC to socket adapter based type names. It also
makes TcpAdapter and TlsAdapter both use it.

 - Added SocketAdapter
 - Renamed all the types to SocketAdapter related names
 - Updated TcpClient to use these new types
 - Moved certs to certs/ folder
 - Renamed SecureTcpAdapter to TlsAdapter

## Testing Plan

Run various RPC tests manually.
 - ironfish stop
 - ironfish status
 - ironfish status -f
 - Disconnect the client in `status -f`
 - Shut down the sever and bring it back up using `status -f`

## Breaking Change

```
[ ] Yes
[x] No
```


ironfish/src/rpc/adapters/socketAdapter/socketAdapter.ts
```diff
@@ -0,0 +1,281 @@
...
+        this.emitResponse(client, this.constructMessage(message.mid, status, data), requestId)
+      },
+      (data: unknown) => {
+        this.emitStream(client, this.constructStream(message.mid, data))
+      },
+    )
+    client.requests.set(requestId, request)
+
+    if (this.router == null) {
+      this.emitResponse(client, this.constructMalformedRequest(data))
```
hughy, line 176: This one seems like an unmounted adapter rather than a malformed request

ironfish/src/rpc/adapters/socketAdapter/socketAdapter.ts
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  onClientError(client: SocketClient, error: unknown): void {
+    this.logger.debug(`${this.host}:${this.port} has error: ${ErrorUtils.renderError(error)}`)
+  }
+
+  async onClientData(client: SocketClient, data: Buffer): Promise<void> {
+    const [parsed, error] = JSONUtils.tryParse(data.toString('utf8').trim())
+    if (error) {
+      this.emitResponse(client, this.constructUnmountedAdapter())
```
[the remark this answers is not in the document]
NullSoldier: Yes, this actually an issue with the existing TCP adapter, and I understand why it's that way. I haven't decided on a fix for it yet.

ironfish/src/rpc/adapters/socketAdapter/socketAdapter.ts
NullSoldier: Changed

PR #1592, opened 2022-06-07
fix: throws error if tcp client connection fails

hughy:
## Summary

the connect method returns a promise, but does not propagate any connection
errors from connectClient, so that promise always resolves even if connection
fails.

these changes throw the error from connectClient if connection is not retried.

## Testing Plan

1. Run `yarn start status -f --rpc.tcp`
2. See no output
3. Start a node
4. Continue to see no output

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/rpc/clients/tcpClient.ts
+++ b/ironfish/src/rpc/clients/tcpClient.ts
@@ -36,20 +36,21 @@ export class IronfishTcpClient extends IronfishRpcClient {
   }
 
   async connect(): Promise<void> {
-    const connected = await this.connectClient()
-      .then(() => true)
-      .catch(() => false)
+    let connected = false
+    await this.connectClient()
+      .then(() => (connected = true))
+      .catch((error) => {
+        if (!this.retryConnect) {
+          this.logger.error(`Failed to connect to ${String(this.host)}:${String(this.port)}`)
+          throw error
+        }
+      })
 
     if (!connected) {
-      if (this.retryConnect) {
-        this.logger.warn(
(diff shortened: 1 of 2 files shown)


ironfish/src/rpc/clients/tcpClient.ts
```diff
@@ -36,20 +36,21 @@ export class IronfishTcpClient extends IronfishRpcClient {
...
     if (!connected) {
-      if (this.retryConnect) {
-        this.logger.warn(
-          `Failed to connect to ${String(this.host)}:${String(this.port)}, retrying`,
-        )
-        this.connectTimeout = setTimeout(() => void this.connect(), CONNECT_RETRY_MS)
-        return
-      }
-
-      this.logger.warn(`Failed to connect to ${String(this.host)}:${String(this.port)}`)
```
NullSoldier, line 52: Does the code need to change? Should we just throw at line 53? I think it's hard for me to understand why the code structure got more complex without an explanation.

hughy: We could throw at line 53, but I think we'd want to throw the error that we catch from `connectClient`. I thought that this was simpler than what I came up with for storing that error and propagating it, but I could give it another pass

NullSoldier: This looks good, thanks for the explanation.

NullSoldier:
This is a good fix, thanks for looking into it and getting it fixed.

NullSoldier (approved):

Issue #1593, opened 2022-06-07
acounts:import bug


wd021:
### What happened?

the import command has a `rescan` flag. if you pass --no-rescan, it still ends up scanning.. 

### Version

0.1.36

### Graffiti

wd021

PR #1594, opened 2022-06-07
enable --no-rescan option on account important

wd021:
## Summary
Fixing https://github.com/iron-fish/ironfish/issues/1593. 

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/rpc/routes/accounts/importAccount.ts
+++ b/ironfish/src/rpc/routes/accounts/importAccount.ts
@@ -47,7 +47,10 @@ router.register<typeof ImportAccountRequestSchema, ImportAccountResponse>(
   ImportAccountRequestSchema,
   async (request, node): Promise<void> => {
     const account = await node.accounts.importAccount(request.data.account)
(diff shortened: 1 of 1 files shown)


ironfish/src/rpc/routes/accounts/importAccount.ts
```diff
@@ -47,7 +47,10 @@ router.register<typeof ImportAccountRequestSchema, ImportAccountResponse>(
   ImportAccountRequestSchema,
   async (request, node): Promise<void> => {
     const account = await node.accounts.importAccount(request.data.account)
-    void node.accounts.startScanTransactionsFor(account)
+
+    if (request.data.rescan !== false) {
```
NullSoldier, line 51: We generally don't add equality checks to things that are already bools

```suggestion
    if (request.data.rescan) {
```

NullSoldier:
Great fix, just one suggestion.

NullSoldier:
Thanks! Looks good!

PR #1607, opened 2022-06-09
Add a round robin queue to the worker pool

mat-if:
## Summary

Prior to this change, the worker pool could receive a large number of jobs for one job type, which would effectively block other job types from executing until they are finished. Ideally, consumers of a job queue should not have to worry about stalling other job types, so this creates a simple round robin queue, which just iterates over the types each time a job is executed to ensure they are all executed in a somewhat timely fashion.

Closes IRO-2095

## Breaking Change

```
[ ] Yes
[ ] No
```

### DIFF
--- a/ironfish/src/workerPool/pool.ts
+++ b/ironfish/src/workerPool/pool.ts
@@ -13,6 +13,7 @@ import { Transaction } from '../primitives/transaction'
 import { Metric } from '../telemetry/interfaces/metric'
 import { WorkerMessageStats } from './interfaces/workerMessageStats'
 import { Job } from './job'
+import { RoundRobinQueue } from './roundrobinqueue'
 import { BoxMessageRequest, BoxMessageResponse } from './tasks/boxMessage'
 import { CreateMinersFeeRequest, CreateMinersFeeResponse } from './tasks/createMinersFee'
 import { CreateTransactionRequest, CreateTransactionResponse } from './tasks/createTransaction'
@@ -37,7 +38,7 @@ export class WorkerPool {
   readonly numWorkers: number
   readonly logger: Logger
 
-  queue: Array<Job> = []
+  queue = new RoundRobinQueue()
   workers: Array<Worker> = []
   started = false
   completed = 0
@@ -117,9 +118,11 @@ export class WorkerPool {
     const queue = this.queue
 
     this.workers = []
-    this.queue = []
 
     queue.forEach((j) => j.abort())
+
+    queue.clearAll()
+
     await Promise.all(workers.map((w) => w.stop()))
   }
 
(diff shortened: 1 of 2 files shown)


ironfish/src/workerPool/roundrobinqueue.ts
```diff
@@ -0,0 +1,95 @@
...
+      }
+    }
+  }
+
+  // Add a job to that type's queue
+  enqueue(type: WorkerMessageType, job: Job): void {
+    const typeQueue = this.queueMap.get(type)
+
+    if (!typeQueue) {
+      return
```
dguenther, line 56: I think it'd be fine to throw if the queue doesn't exist -- I don't think there's a point where we'd be trying to queue up a job that doesn't have a mapping, so it'd probably indicate a bug in our code

ironfish/src/workerPool/roundrobinqueue.ts
```diff
@@ -0,0 +1,95 @@
...
+      if (!queue.length) {
+        continue
+      }
+
+      const nextJob = queue.shift()
+      if (!nextJob) {
+        continue
+      }
+
+      this.lastMapIndex = index
```
dguenther, line 85: I might be getting a bit confused, but is this index relative to the reshuffled array rather than the original queueMap?

I feel like it might be a little easier to read if you work with the original queueEntries array, then loop with `i = 1; i <= queueEntries.length; i++`, generating the index as `(lastMapIndex + i) % queueEntries.length`

ironfish/src/workerPool/pool.ts
```diff
@@ -117,9 +118,11 @@ export class WorkerPool {
     const queue = this.queue
 
     this.workers = []
-    this.queue = []
 
     queue.forEach((j) => j.abort())
+
+    queue.clearAll()
```
dguenther, line 124: I think if you wanted to make an `abortAll` function and replace the above `forEach` with that, it could be handy. The queue could also drop all of the jobs when it calls abort on them.

The clearAll is okay in this situation, but I'm concerned that if someone happens to call it without cleaning up the promises (calling abort), anything awaiting on the job's promise could get stuck forever. Or in other words, I don't think we'd ever want to clear the queue without also aborting the jobs, so it'd make sense to combine them.

dguenther (changes requested):

ironfish/src/workerPool/pool.ts
mat-if: Good call. Initially was just trying to map 1:1 with our current usage of the queue, but that does make for a more brittle API. 👍

dguenther (approved):

PR #1610, opened 2022-06-10
fix: restore enableNativeRpcTcpAdapter option, add option for client

hughy:
## Summary

we've recently made several changes in the rpc layer:
- native TCP adapter
- native TCP client
- TLS adapter/client

it may be useful to be able to test these features separately. this commit
restores an option for enabling the native TCP adapter and introduces an option
for enabling the native TCP client so that we may enable and test the adapter
and client separately.

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/fileStores/config.ts
+++ b/ironfish/src/fileStores/config.ts
@@ -46,6 +46,13 @@ export type ConfigOptions = {
   getFundsApi: string
   ipcPath: string
   /**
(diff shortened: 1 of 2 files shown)


NullSoldier:
If we deploy the adapters and they are stable, will we need this anymore?

hughy:
> If we deploy the adapters and they are stable, will we need this anymore?

Nope, this won't be necessary if the new adapters and clients are stable!

PR #1614, opened 2022-06-10
feat: adds message buffering to rpc tcp sockets

hughy:
## Summary
a tcp socket is modeled as a stream of bytes. we use tcp sockets to send and
receive json-formatted messages, but there is no guarantee that a tcp segment
will contain a full message or messages. it's possible, especially with larger
messages, that we read data from the socket in the middle of a message.
attempting to parse such partial messages results in json parsing errors.

- implements a MessageBuffer class to encapsulate the logic required
for buffering partial messages read from a socket

- adds a MessageBuffer to TcpClient
- adds a MessageBuffer to each SocketClient
- updates 'data' event listeners to use MessageBuffers
- adds MESSAGE_DELIMITER to protocol definitions

## Testing Plan
- `yarn test` tests pass
- local testing running node and commands over tcp
- local testing running node and pool

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/rpc/clients/tcpClient.ts
+++ b/ironfish/src/rpc/clients/tcpClient.ts
@@ -5,11 +5,15 @@ import net from 'net'
 import { Assert } from '../../assert'
 import { createRootLogger, Logger } from '../../logger'
 import { ErrorUtils, SetTimeoutToken, YupUtils } from '../../utils'
-import { ServerSocketRpc, ServerSocketRpcSchema } from '../adapters/socketAdapter/protocol'
+import {
+  MESSAGE_DELIMITER,
+  ServerSocketRpc,
+  ServerSocketRpcSchema,
+} from '../adapters/socketAdapter/protocol'
+import { MessageBuffer } from '../messageBuffer'
 import { ConnectionLostError, ConnectionRefusedError } from './errors'
 import { IronfishRpcClient, RpcClientConnectionInfo } from './rpcClient'
 
-const NODE_IPC_DELIMITER = '\f'
 const CONNECT_RETRY_MS = 2000
 
 export class IronfishTcpClient extends IronfishRpcClient {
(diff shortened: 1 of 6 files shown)


ironfish/src/rpc/clients/tcpClient.ts
```diff
@@ -89,6 +95,7 @@ export class IronfishTcpClient extends IronfishRpcClient {
   close(): void {
     this.client?.end()
 
+    this.messageBuffer.clear()
```
NullSoldier, line 98: This shouldn't do anything, but I know why it's here.

ironfish/src/rpc/messageBuffer.ts
```diff
@@ -0,0 +1,35 @@
...
+    // buffer contains no full messages
+    if (lastDelimiterIndex === -1) {
+      return []
+    }
+
+    const messages = this.buffer.substring(0, lastDelimiterIndex).trim().split(this.delimiter)
+    this.buffer = this.buffer.substring(lastDelimiterIndex + 1)
+    return messages
+  }
+}
```
NullSoldier, line 35: We should make this use all bytes later probably, lots of buffer copies going on here. I don't think we need to optimize it now, though.

NullSoldier (approved):

NullSoldier:
This because it does fix a bug, but I can't get anything to crash without this fix. So this may not be the issue with the RPC that's causing problems in production.

NullSoldier (approved):

NullSoldier:
Ok, so this won't crash without this, but any partial messages will get dropped and ignored. I tested this by adding this to the getStatus command using `ironfish status -f` then logged socket and various data.

```
foo: Buffer.alloc(20 * 1024 * 1024, 'hello').toString('utf-8')
```

I think we were seeing a crash on prod, so that will probably still be happening.

PR #1619, opened 2022-06-13
Remove rpcRetryConnect from TCP adapters

NullSoldier:
## Summary

This feature was not working to begin with, and not well supported. I
will remove the rest from the IPC client once we delete the IPC client
and adapter soon.

## Testing Plan
Ran the node and RPC layer

## Breaking Change

```
[ ] Yes
[x] No
```


hughy (approved):
We could also remove `retryConnect` from the IPC client, but perhaps we're better off waiting until we're ready for a broader IPC refactor

PR #1623, opened 2022-06-13
feat: implements abstract base class for discord, lark

hughy:
## Summary

the implementations for the discord and lark webhooks are nearly identical.
defining a shared base class for message text makes it easier to change the
format and/or output of messages in one place to apply to both webhooks.

## Breaking Change

```
[ ] Yes
[X] No
```


danield9tqh:
I think the real duplication is having to make duplicate calls like `this.lark?.poolPayoutSuccess` and `this.discord?.poolPayoutSuccess`. Can we change that also in this PR?

NullSoldier (approved):

NullSoldier (approved):

PR #1624, opened 2022-06-13
PeerNetwork should sync transactions regardless of mempool accepting it

mat-if:
## Summary
This fixes issues where accounts can become out of sync when used on multiple machines, primarily due to creating transactions on all of them.

Closes IRO-2158

## Breaking Change

```
[ ] Yes
[ ] No
```


ironfish/src/network/peerNetwork.ts
```diff
@@ -702,8 +702,9 @@ export class PeerNetwork {
       return true
     }
 
+    await this.node.accounts.syncTransaction(verifiedTransaction, {})
```
NullSoldier, line 705: I'm not 100% sure if this is the right order or not, so it might be good to get an explanation on this over a call. I'll just post a bullet point of my thoughts.

 - In all other places syncTransaction is used, pre validation occurs. Since were not doing validation anymore, does this mean someone can craft invalid transactions and poison our wallet? It seems like we would want to do validation.
 -  We are still using the mem pool to gate the TX here, is that fine? If were not synced, and if the mem pool already has it.

NullSoldier:
I tested this manually, and my mem pool seems to grow and work correctly. I'll be curious to watch this one on the next production deploy.

NullSoldier (approved):

PR #1634, opened 2022-06-15
feat: adds links to block explorer in webhook messages

hughy:
## Summary
- hyperlinks block header hash hex string and transaction hash hex string to
  block explorer
- changes message in transaction success to indicate transaction created, but
  pending
## Testing Plan
created discord server and webhook, tested messages with local pool
## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/mining/webhooks/webhookNotifier.ts
+++ b/ironfish/src/mining/webhooks/webhookNotifier.ts
@@ -83,6 +83,6 @@ export abstract class WebhookNotifier {
       return `\`${hashHex}\``
     }
 
-    return `[${hashHex}](${explorerUrl + hashHex})`
+    return `${explorerUrl + hashHex}`
   }
 }
--- a/ironfish/src/mining/webhooks/webhookNotifier.ts
+++ b/ironfish/src/mining/webhooks/webhookNotifier.ts
@@ -7,6 +7,8 @@ import { createRootLogger, Logger } from '../../logger'
 import { displayIronAmountWithCurrency, ErrorUtils, oreToIron } from '../../utils'
 import { FileUtils } from '../../utils/file'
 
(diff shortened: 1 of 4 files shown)


deekerno (approved):

ironfish/src/mining/webhooks/webhookNotifier.ts
```diff
@@ -7,6 +7,8 @@ import { createRootLogger, Logger } from '../../logger'
 import { displayIronAmountWithCurrency, ErrorUtils, oreToIron } from '../../utils'
 import { FileUtils } from '../../utils/file'
 
+const BLOCK_EXPLORER_URL = 'https://explorer.ironfish.network'
```
NullSoldier, line 10: Can we open an other PR after this to run these types of things from the config? Remember that some users have custom block explorers and need to customize these settings to point at their custom infrastructure. It's probably best to pass this down into the constructor from the config when the notifiers / pools are created.

ironfish/src/mining/webhooks/webhookNotifier.ts
```diff
@@ -62,4 +77,12 @@ export abstract class WebhookNotifier {
       `Error while sending payout transaction: ${ErrorUtils.renderError(error, true)}`,
     )
   }
+
+  private renderHashHex(hashHex: string, explorerUrl: string | null): string {
+    if (explorerUrl == null) {
+      return `\`${hashHex}\``
+    }
+
+    return `[${hashHex}](${explorerUrl + hashHex})`
```
NullSoldier, line 86: Does this work in discord? Where does this work?

hughy: Works with Discord, but not with Lark. I'll change it to use just the URL

NullSoldier (approved):

PR #1643, opened 2022-06-16
feat: adds webhook messages at intervals

hughy:
## Summary

defines methods for fetching pool status information and emits status information to webhooks at interval

- sends message to webhooks before attempting to send transaction
- changes message in transaction success to indicate transaction created, but
  pending
- adds config option to specify status message interval length. setting to 0
  disables status messages
- outputs status of mining pool, including hash rate, miners, shares

## Testing Plan

- created webhooks for discord and lark servers for testing, ran node, pool, and miner locally

## Breaking Change

```
[ ] Yes
[X] No
```


hughy:
The resulting status messages look like this:
image

ironfish/src/mining/pool.ts
```diff
@@ -134,6 +137,15 @@ export class MiningPool {
 
     this.logger.info('Connecting to node...')
     this.rpc.onClose.on(this.onDisconnectRpc)
+
+    const statusInterval = this.config.get('poolStatusNotificationInterval') ?? 0
```
NullSoldier, line 141: It should produce a warning about this null coelesence right? Since the config is typed as a number?

ironfish/src/mining/pool.ts
```diff
@@ -50,6 +50,8 @@ export class MiningPool {
 
   recalculateTargetInterval: SetTimeoutToken | null
 
+  private notifyStatusInterval: SetTimeoutToken | null
```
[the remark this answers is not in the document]
NullSoldier: Looks like the above one might be wrong too

ironfish/src/mining/pool.ts
```diff
@@ -390,4 +406,26 @@ export class MiningPool {
...
+    }
+  }
+}
+
+export type MiningPoolStatus = {
+  name: string
+  hashRate: number
+  miners: number
+  sharesPending: number
 }
```
NullSoldier, line 431: I usually inline these, but this is fine too as long as we don't have too many of these. Their only purpose is to match sure the types match on the other side but typescript will do that for us so it tends to just gunk up files when there are too many of them.

ironfish/src/mining/pool.ts
```diff
@@ -390,4 +406,26 @@ export class MiningPool {
...
+
+  async getStatus(): Promise<MiningPoolStatus> {
+    return {
+      name: this.name,
+      hashRate: await this.estimateHashRate(),
+      miners: this.stratum.clients.size,
+      sharesPending: await this.shares.sharesPendingPayout(),
+    }
+  }
+}
```
NullSoldier, line 424: You should put these 2 awaits in a Promise.all() to speed this up.

```suggestion
  async getStatus(): Promise<MiningPoolStatus> {
    const [hashRate, sharesPending] = await Promise.all([
        await this.estimateHashRate(),
        await this.shares.sharesPendingPayout(),
    ])

    return {
      name: this.name,
      hashRate: hashRate,
      miners: this.stratum.clients.size,
      sharesPending: sharesPending,
    }
  }
}
```

ironfish/src/mining/webhooks/webhookNotifier.ts
```diff
@@ -78,6 +79,31 @@ export abstract class WebhookNotifier {
...
+  }
+
+  poolStatus(status: MiningPoolStatus): void {
+    this.sendText(
+      `Status for mining pool '${status.name}':\n\tHashrate: ${FileUtils.formatHashRate(
+        status.hashRate,
+      )}/s\n\tMiners: ${status.miners}\n\tShares pending: ${status.sharesPending}`,
+    )
+  }
+
```
NullSoldier, line 106: Hopefully this can be deleted later in favor of a poll mechanic so it doesn't spam notifications.

ironfish/src/mining/webhooks/webhookNotifier.ts
```diff
@@ -78,6 +79,31 @@ export abstract class WebhookNotifier {
...
+    this.sendText(
+      `Creating payout of ${totalShareCount} shares to ${
+        receives.length
+      } users for ${displayIronAmountWithCurrency(
+        Number(oreToIron(Number(total.toString()))),
+        false,
+      )}(${payoutId})`,
+    )
+  }
```
NullSoldier, line 97: Hopefully this can be deleted later too when the pool is more reliable, so we can only have 1 notification per transaction created.

ironfish/src/mining/pool.ts
hughy: I think that I had put this here because of a warning 🧐 , but it's not needed

ironfish/src/mining/pool.ts
hughy: I'm going to leave this here for now -- it will probably make its way into a message schema in followup changes that add a polling for status

PR #1646, opened 2022-06-17
change metrics meter to exponential weighted moving average

wd021:
## Summary
moving our metrics monitor from using a rolling average to exponentially-weighted moving average

## Testing Plan
run node and observe metrics
- yarn start status -f
- yarn start workers:status -f

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/assert.ts
+++ b/ironfish/src/assert.ts
@@ -39,6 +39,12 @@ export class Assert {
     }
   }
 
+  static isGreaterThan(a: number, b: number, message?: string): void {
+    if (a <= b) {
+      throw new Error(message || `Expected ${String(a)} to be greater than ${String(b)}`)
+    }
+  }
+
   static isNever(x: never, message?: string): never {
     throw new Error(message || `Expected value to be never: ${String(x)}`)
   }
--- a/ironfish/src/metrics/ewmaAverage.ts
+++ b/ironfish/src/metrics/ewmaAverage.ts
@@ -1,0 +1,54 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { Assert } from '../assert'
+
+/**
+ * Utility to compute the exponential weighted moving average
+ *
+ * inspired by https://github.com/shaka-project/shaka-player/blob/main/lib/abr/ewma.js
+ */
+export class EwmaAverage {
+  /**
+   * Current average
+   */
+  average = 0
+
+  // larger values of alpha expire historical data more slowly
+  private _alpha = 0
+  private _estimate = 0
+  private _totalWeight = 0
+
+  constructor(halflife: number) {
+    Assert.isGreaterThan(halflife, 0)
+    this._alpha = Math.exp(Math.log(0.5) / halflife)
+    this.reset()
+  }
+
+  /**
+   * Add a sample {value}, {weight}
+   * weight = time delta
+   */
+  add(value: number, weight: number): void {
+    const adjAlpha = Math.pow(this._alpha, weight)
+    const newEstimate = value * (1 - adjAlpha) + adjAlpha * this._estimate
+
+    if (!isNaN(newEstimate)) {
+      this._estimate = newEstimate
+      this._totalWeight += weight
+    }
+
+    // compute average
+    const zeroFactor = 1 - Math.pow(this._alpha, this._totalWeight)
+    this.average = this._estimate / zeroFactor
+  }
+
+  /**
+   * Resets the rolling average (equivalent to 0 samples taken so far)
+   */
+  reset(): void {
+    this._estimate = 0
+    this._totalWeight = 0
+  }
+}
--- a/ironfish/src/metrics/ewmaAverage.ts
+++ b/ironfish/src/metrics/ewmaAverage.ts
@@ -7,7 +7,7 @@ import { Assert } from '../assert'
 /**
  * Utility to compute the exponential weighted moving average
  *
- * inspired by https://github.com/shaka-project/shaka-player/blob/main/lib/abr/ewma.js
+ * inspired by https://github.com/shaka-project/shaka-player/blob/eaadb52627f0a0347390b201866585cce91fe9d0/lib/abr/ewma.js
  */
 export class EwmaAverage {
   /**
@@ -20,6 +20,9 @@ export class EwmaAverage {
   private _estimate = 0
   private _totalWeight = 0
 
+  /**
+   * halflife is the time decay that holds half of the estimate value
+   */
   constructor(halflife: number) {
     Assert.isGreaterThan(halflife, 0)
     this._alpha = Math.exp(Math.log(0.5) / halflife)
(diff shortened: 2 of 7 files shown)


NullSoldier:
Can you provide more of an explanation of why this is a good change, rather than a description of the changes themselves? It's really helpful for those of us without context.

wd021:
absolutely. in this case, I'm implementing the code comment left in `meter.ts` -> `@TODO: Move RollingAverages to exponentially-weighted moving average (EWMA)`. 

I think the reasoning is that we want metrics to be weighted heavier towards recency as opposed to a more even distribution.  I know this is common in trading charts and it's better for spotting price/trend movements. 

but yea, perhaps @dguenther can chime in as well, since he's the author of the original file.

NullSoldier:
Ok great, thanks for the explanation. I'll review soon.

ironfish/src/metrics/meter.ts
```diff
@@ -13,27 +13,36 @@ import { RollingAverage } from './rollingAverage'
...
-  private _rate1s: RollingAverage
-  private _rate5s: RollingAverage
-  private _rate1m: RollingAverage
-  private _rate5m: RollingAverage
-  private _average: RollingAverage
+  private _rate1s: EwmaAverage
+  private _rate5s: EwmaAverage
+  private _rate1m: EwmaAverage
+  private _rate5m: EwmaAverage
+  private _average: EwmaAverage
```
NullSoldier, line 23: this one is not a time series value, but a mathematical average, so should it be EwmaAverage, or RollingAverage?

ironfish/src/metrics/meter.ts
```diff
@@ -106,10 +115,10 @@ export class Meter {
...
     const rateSec = elapsedMs === 0 ? 0 : (this._count / elapsedMs) * 1000
 
-    this._rate1s.add(rateSec)
-    this._rate5s.add(rateSec)
-    this._rate1m.add(rateSec)
-    this._rate5m.add(rateSec)
+    this._rate1s.add(rateSec, this.METER_TIME_INTERVALS_MS.rate1s / this._intervalMs)
+    this._rate5s.add(rateSec, this.METER_TIME_INTERVALS_MS.rate5s / this._intervalMs)
+    this._rate1m.add(rateSec, this.METER_TIME_INTERVALS_MS.rate1m / this._intervalMs)
+    this._rate5m.add(rateSec, this.METER_TIME_INTERVALS_MS.rate5m / this._intervalMs)
```
NullSoldier, line 121: It took me some time to read various sources, and a few implementations of EWMA but I'm not sure this part is correct. I looked in the source code of `ewma_bandwidth_estimator` where this code came from in the shaka library, and they use this a bit different to adjust the alpha for non evenly spaced points.

1. The half-life in the constructor should be in millisecond units. Shaka uses seconds because they keep all their data in seconds consistently. https://github.com/shaka-project/shaka-player/blob/0335b2af2efea6ceda83e536e12094e4cc942a25/lib/abr/ewma_bandwidth_estimator.js#L22-L34
2. Weight is used to adjust the alpha, from unevenly spaced data points which EWMA requires. Shaka just uses elapsed time only, regardless of the EWMA period / half life. https://github.com/shaka-project/shaka-player/blob/0335b2af2efea6ceda83e536e12094e4cc942a25/lib/abr/ewma_bandwidth_estimator.js#L97

ironfish/src/metrics/ewmaAverage.ts
```diff
@@ -0,0 +1,54 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { Assert } from '../assert'
+
+/**
+ * Utility to compute the exponential weighted moving average
+ *
+ * inspired by https://github.com/shaka-project/shaka-player/blob/main/lib/abr/ewma.js
```
NullSoldier, line 10: Is this code have a compatible license? We need to check that as well. If so, link to the git commit with the hash specifically rather than master, since master can change.

NullSoldier (changes requested):

ironfish/src/metrics/meter.ts
```diff
@@ -13,27 +13,36 @@ import { RollingAverage } from './rollingAverage'
...
-    this._rate1s = new RollingAverage(1000 / this._intervalMs)
-    this._rate5s = new RollingAverage(5000 / this._intervalMs)
-    this._rate1m = new RollingAverage((1 * 60 * 1000) / this._intervalMs)
-    this._rate5m = new RollingAverage((5 * 60 * 1000) / this._intervalMs)
-    this._average = new RollingAverage(100)
+    this._rate1s = new EwmaAverage(this._halfLife)
+    this._rate5s = new EwmaAverage(this._halfLife)
+    this._rate1m = new EwmaAverage(this._halfLife)
+    this._rate5m = new EwmaAverage(this._halfLife)
+    this._average = new EwmaAverage(this._halfLife)
```
NullSoldier, line 45: I'm not sure this is correct, I commented about this here. https://github.com/iron-fish/ironfish/pull/1646#discussion_r908798362

ironfish/src/metrics/meter.ts
wd021: 1 - Hmm, not sure why the half-life unit should be in ms. We're also delineating our metrics in s, like Shaka. We record values as `rateSec`. 
 
2 - Yea, Shaka uses different half-life periods but a consistent weight value when inserting data. I was doing it the other way around (same half-life, but weight adj). I've reversed this now.

ironfish/src/metrics/ewmaAverage.ts
wd021: Apache license, i've replaced with the git commit

wd021:
Just to expand on https://github.com/iron-fish/ironfish/pull/1646#discussion_r908798362 👇 

Right now the Meter class has `_rate` & `_average` values and it's not obvious from the code that you cannot use both of these in one instance. 

When using `_rate`, we use `add(count: number)` to inc a counter. ex: we call `this.speed.add(1)` when addBlock gets called in the syncer class.

When using `_average`, we use `add(count: number)` to add distinct values. ex: we call `this.addSpeed.add(addTime)` when connect gets called in the blockchain class.

Unless I'm missing something, i feel like this needs a bit of refactoring for clarity. Also, if this is the case, then our usage of Meter for p2p traffic stats is incorrect.

NullSoldier:
I pushed up a small commit to rename `Exponential Moving Average Average` to `Exponential Moving Average`

ironfish/src/assert.ts
```diff
@@ -39,6 +39,12 @@ export class Assert {
     }
   }
 
+  static isGreaterThan(a: number, b: number, message?: string): void {
+    if (a <= b) {
```
NullSoldier, line 43: this also fails if it's equal to but the method name doesn't match that. Also the method checks if something is greater than something else but there is no greater than used in the function. The logic is inverted.

NullSoldier:
I think @wd021 is right, we shouldn't be mixing use cases for rates and averages. I think I'll put up a FIP for the changes here for my recommendation on what to do. for now I'll just revert the average only back to a rolling average which is what `addSpeed` uses.

NullSoldier (approved):

PR #1647, opened 2022-06-17
small test script cleanup

wd021:
## Summary
- `ironfish/src/memPool/memPool.test.ts` doesn't need to loop through block txs to get the miner's fee. use the first block.
- code comment in `ironfish/src/genesis/genesis.test.slow.ts` is not needed. should've been removed with https://github.com/iron-fish/ironfish/commit/5beeedc2f5cd149da04379aed8c92bd48f52741a.

## Testing Plan
run yarn test

## Breaking Change

```
[ ] Yes
[x] No
```


NullSoldier (approved):
Nice cleanup! Yeah we used to not guarantee that this was transaction 0.

PR #1648, opened 2022-06-17
decouples StratumClient from MiningPoolMiner

hughy:
## Summary

background:
the StratumClient and MiningPoolMiner are tightly coupled -- an instance of one
type holds a reference to an instance of the other type. this coupling makes it
difficult to use StratumClient without a MiningPoolMiner. a developer might want
to do this if they were to implement their own miner, or if they wanted to use
the StratumClient to receive messages outside of a mining context (e.g., to get
status information from the pool).

- removes miner property from StratumClient
- modifies StratumClient to emit message bodies after validating

## Testing Plan

tested with local mining pool and miner

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/mining/stratum/stratumClient.ts
+++ b/ironfish/src/mining/stratum/stratumClient.ts
@@ -39,6 +39,11 @@ export class StratumClient {
 
   private readonly publicAddress: string
 
+  readonly onSubscribed = new Event<[MiningSubscribedMessage]>()
+  readonly onSetTarget = new Event<[MiningSetTargetMessage]>()
+  readonly onNotify = new Event<[MiningNotifyMessage]>()
+  readonly onWaitForWork = new Event<[MiningWaitForWorkMessage]>()
+
   constructor(options: { publicAddress: string; host: string; port: number; logger: Logger }) {
     this.host = options.host
     this.port = options.port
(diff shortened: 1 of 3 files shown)


ironfish/src/mining/stratum/stratumClient.ts
```diff
@@ -61,14 +56,14 @@ export class StratumClient {
...
-  start(): void {
+  async start(): Promise<void> {
     if (this.started) {
       return
     }
 
     this.started = true
     this.logger.info('Connecting to pool...')
-    void this.startConnecting()
+    await this.startConnecting()
```
NullSoldier, line 66: This should be voided, right? Why is this being awaited now?

ironfish/src/mining/stratum/stratumClient.ts
```diff
@@ -161,6 +155,11 @@ export class StratumClient {
     this.logger.error(`Stratum Error ${ErrorUtils.renderError(error)}`)
   }
 
+  readonly onSubscribed = new Event<[MiningSubscribedMessage]>()
+  readonly onSetTarget = new Event<[MiningSetTargetMessage]>()
+  readonly onNotify = new Event<[MiningNotifyMessage]>()
+  readonly onWaitForWork = new Event<[MiningWaitForWorkMessage]>()
```
NullSoldier, line 161: Fields usually go at the top of a class in our files.

ironfish/src/mining/stratum/stratumClient.ts
hughy: The change to await blocks the client from being started until it's finished connecting.

The StratumClient currently sends a 'subscribe' message automatically when it finishes connecting, but these changes have the miner making that call. It can only send that message once the client has connected, so it blocks until that's happened

NullSoldier: Right, but start() is not correlated to being connected. This just starts the connecting loop. What if it connects on the second time, instead of the first time? Start() will resolve but you won't be connected.

If things need to respond to becoming connected, then it should respond to an onConnected event that's fired rather than rely on start() which won't guarantee a connection.

hughy: Good point, I didn't think through how this would resolve with connection retries

NullSoldier (approved):

NullSoldier:
Tested manually, looks good!

PR #1653, opened 2022-06-21
Add http support for RPC

hpool-dev:
## Summary
Add http support for RPC

## Testing Plan
Run node with `--rpc.tcp` and access `http://localhost:$rpcTcpPort+1`. Example `curl http://localhost:8021/node/getStatus`.

## Breaking Change

```
[ ] Yes
[x] No
```


hpool-dev:
I think the current RPC interface of TCP and IPC is not very convenient to call with go,java and etc. This implementation HTTP , and streaming by websocket.

NullSoldier:
This should go in their own adapters. I can move this code, or you can, if you have time.

hpool-dev:
> This should go in their own adapters. I can move this code, or you can, if you have time.

Glad to have this feature endorsed. We will refactor the code.

NullSoldier:
Oh yeah, this is actually a very important feature that's on our roadmap. We're super glad to have this contribution.

NullSoldier:
I'm going to comment on these FIPs, but yes please file a FIP separately for JSON RPC 2.0.

hpool-dev:
> Ok, I think we should port all of our adapters to use JSON RPC, I'm just not sure how it should work for our streaming responses. I'll have to look at the JSON RPC 2.0 spec. I'll comment on the FIP. Obviously an HTTP adapter won't support streaming responses.

I recommend steaming to use the websocket, which is no problem in my PR.

It should be better if it can support JSON RPC 2.0, many other famous chains are using it.

NullSoldier:
I'm gonna close this PR, but I'm gonna finish and port the HTTP portion here. When we finish it we'll give you some PR credit since my work is based off of yours. https://github.com/iron-fish/ironfish/pull/2423

PR #1655, opened 2022-06-21
Add chain snapshot export functionality

deekerno:
## Summary
Adds chain snapshot export functionality. This is done by implementing a new command `service:snapshot` that interacts with a new RPC `snapshotChain` route. The desired blocks are grouped into chunks of a defined size and serialized into buffers. These buffers are written to disk and compressed into an archive, which is then uploaded to a storage bucket along with a `manifest.json` file that contains snapshot metadata. Snapshot ingestion will be put up for review in a separate PR.

Note that there is currently undiagnosed slowness when attempting to export a snapshot while a full node is running. Snapshot export is quite efficient when a full node is not running.

## Testing Plan
Added test. Verified export and upload works properly by creating a test S3 bucket and uploading a test chain to it.

## Breaking Change

```
[ ] Yes
[x] No
```


ironfish/src/rpc/routes/chain/snapshotChain.ts
```diff
@@ -0,0 +1,107 @@
...
+)
+
+function serializeChunk(blocks: Buffer[]): Buffer {
+  let sizeOfBuffers = 0
+  for (const block of blocks) {
+    sizeOfBuffers += bufio.sizeVarBytes(block)
+  }
+  const totalSize = 8 + sizeOfBuffers
+
+  const bw = bufio.write(totalSize)
```
lwisne, line 99: more descriptive var name here would be nice

ironfish/src/rpc/routes/chain/snapshotChain.ts
```diff
@@ -0,0 +1,107 @@
...
+          if (blocks.length >= maxBlocksPerChunk) {
+            const buffer = serializeChunk(blocks)
+            blocks = []
+            request.stream({ start, stop, seq: i, buffer })
+          }
+        }
+      }
+    }
+
+    if (blocks.length) {
```
lwisne, line 83: nit - `blocks.length > 0` here would be slightly clearer in terms of what you want

ironfish/src/rpc/routes/chain/snapshotChain.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    let blocks: Buffer[] = []
+
+    for (let i = start; i <= stop; ++i) {
+      const blockHeader = await node.chain.getHeaderAtSequence(i)
+      if (blockHeader) {
+        const block = await node.chain.getBlock(blockHeader)
+        if (block) {
+          const serializedBlock = node.chain.strategy.blockSerde.serialize(block)
+          const bw = bufio.write(getBlockSize(serializedBlock))
```
lwisne, line 70: more descriptive name here

ironfish-cli/src/commands/service/snapshot.ts
```diff
@@ -0,0 +1,213 @@
...
+      }
+    }
+    Assert.isNotUndefined(exportDir)
+
+    const blockExportPath = this.sdk.fileSystem.join(exportDir, 'blocks')
+    await this.sdk.fileSystem.mkdir(blockExportPath, { recursive: true })
+
+    this.log('Connecting to node...')
+
+    const client = await this.sdk.connectRpc()
```
lwisne, line 91: Have you tried passing `forceLocal` here so you will not attempt to connect to a currently running node? I think it then will crap out if you are trying to spin up a second node, and should help avoid the slowness issue.

ironfish-cli/src/commands/service/snapshot.ts
```diff
@@ -0,0 +1,213 @@
...
+    CliUx.ux.action.stop(`done`)
+
+    const blockHeight = stop
+
+    CliUx.ux.action.start(`Uploading to ${bucket}`)
+    await this.uploadToBucket(snapshotPath, bucket, 'application/x-compressed-tar')
+    CliUx.ux.action.stop(`done`)
+
+    await fsAsync.writeFile(
+      path.join(exportDir, 'manifest.json'),
```
lwisne, line 148: Is it always best to have one huge snapshot? This would suggest so. Seems possible that the manifest should have flexibility to point to multiple files that together make up the current snapshot.

ironfish-cli/src/commands/service/snapshot.ts
```diff
@@ -0,0 +1,213 @@
...
+      JSON.stringify({
+        block_height: blockHeight,
+        checksum,
+        file_name: snapshotFileName,
+        file_size: fileSize,
+        timestamp,
+      }),
+    )
+
+    CliUx.ux.action.start(`Uploading latest snapshot information to ${bucket}`)
```
lwisne, line 158: If the write gives an error code for some reason, this would just go ahead and update the manifest and create an invalid state. May be good here to have an `writefile(...).then(...)` in which the then does the manifest upload

ironfish-cli/src/commands/service/snapshot.ts
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  uploadToBucket(dest: string, host: string, contentType: string): Promise<number | null> {
+    return new Promise<number | null>((resolve, reject) => {
+      const date = new Date().toISOString()
+      const file = path.basename(dest)
+      const acl = 'bucket-owner-full-control'
+
+      const process = spawn(
+        `curl`,
```
lwisne, line 189: Invoking a command line process from within code is generally an antipattern. We use `axios` for this kind of thing elsewhere, would recommend that. Or more directly you can use `node-libcurl` if axios doesn't work (but it should!)

ironfish/src/rpc/routes/chain/snapshotChain.test.ts
```diff
@@ -0,0 +1,59 @@
...
+import { Assert } from '../../../assert'
+import { getBlockSize, writeBlock } from '../../../network/utils/block'
+import { useMinerBlockFixture } from '../../../testUtilities'
+import { createRouteTest } from '../../../testUtilities/routeTest'
+import { SnapshotChainStreamResponse } from './snapshotChain'
+
+describe('Route chain/snapshotChainStream', () => {
+  const routeTest = createRouteTest()
+
+  it('correctly returns a serialized list of blocks', async () => {
```
lwisne, line 14: you don't test the params - start, stop, maxBlocks

ironfish-cli/src/commands/service/snapshot.ts
deekerno: I wouldn't say it's better one way or the other; it's just the implementation that I spec'd out. I'm open to creating separate archives so that a user's node will only download the subset of the chain they are missing.

ironfish-cli/src/commands/service/snapshot.ts
```diff
@@ -0,0 +1,213 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
lwisne, line 1: Need a test for this file. It's complicated with a number of parameters, and specifically you need to really ensure the manifest file system works properly

ironfish-cli/src/commands/service/snapshot.ts
lwisne: But generally - we allow start/stop parameters in the actual command, but then have a manifest setup that ignores those params. 

For instance, I'm seeing that any user could corrupt all imports by simply running the export command with a start param other than 1. This would run, upload to the bucket, then set manifest to point to that new (incomplete) export.

ironfish-cli/src/commands/service/snapshot.ts
lwisne: actually, even better than both of these is the AWS node SDK

https://stackabuse.com/uploading-files-to-aws-s3-with-node-js/

ironfish-cli/src/commands/service/snapshot.ts
deekerno: Sure, I see your point. I think it's safe to assume that a snapshot should be a complete representation of the main chain, and if we choose to do multiple snapshot files that are described in the manifest, then I think it's also safe to do away with the start and stop parameters as the import command should handle retrieval of the files that contain the latest un-added block. What do you think?

ironfish-cli/src/commands/service/snapshot.ts
deekerno: If we're moving away from hardcoding a `curl`call that includes AWS-specific data, is it prudent to use an SDK that ties users to that service directly?

ironfish-cli/src/commands/service/snapshot.ts
lwisne: I think I agree, though will have to see the implementation as there is a bit of room for ambiguity here.

ironfish-cli/src/commands/service/snapshot.ts
lwisne: Well, the parameter for the CLI command is `bucket` and it seems every assumption is made that this is an AWS bucket, so I would think using the AWS SDK is fine.

And to clarify - my comment about the curl command wasn't due to AWS tie-in, it was because you generally shouldn't be invoking OS-level commands from code.

deekerno: Sure, I can use the SDK there.

Agreed on the second, just opening up for alternatives there.

ironfish-cli/src/commands/service/snapshot.ts
```diff
@@ -0,0 +1,213 @@
...
+    },
+  ]
+
+  async start(): Promise<void> {
+    const { flags, args } = await this.parse(CreateSnapshot)
+
+    const bucket = (flags.bucket || process.env.IRONFISH_SNAPSHOT_BUCKET || '').trim()
+    if (!bucket) {
+      this.log(
+        `Cannot upload snapshot without bucket URL. You must set IRONFISH_SNAPSHOT_BUCKET or pass --bucket flag.`,
```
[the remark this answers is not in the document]
deekerno: Yes. I think it's fair to believe that if a user supplies a path, then they may not intend to upload a snapshot. Or given our other discussion about using the AWS SDK, they may intend to upload a snapshot to a different host instead. In that case, they could supply a path for the export and then manually upload those files to a different host. So the change would be to only export if the path is present, and to attempt to upload the snapshot to an S3 bucket if it is not present. Does that seem reasonable?

ironfish/src/rpc/routes/chain/snapshotChain.test.ts
deekerno: Removing the start and stop parameters as stated in the other feedback will make this unnecessary.

ironfish-cli/src/commands/service/snapshot.ts
deekerno: I changed the command to use the `forceLocal` flag and the command now waits for the node to close before exporting a snapshot.

ironfish-cli/src/commands/service/snapshot.ts
deekerno: @lwisne: After thinking about it and tinkering with some implementations, I'm not sure that moving to an SDK is the best move here. If we include it as part of the command, then every user who pulls down our software is going to have to download the entire AWS SDK and all of its dependencies. Furthermore, I think it forces users of this command to use AWS and that may not be a choice they're willing to make.

Additionally, for our use case, AWS requires multiple requests to be made to open a multipart file upload session as they now require it for large files. If we keep `curl` or even move to alternatives (e.g. `node-libcurl`, `axios`) for uploading to S3, hard-coding AWS' multiple request requirement may prove cumbersome for other users who may want to use a different host.

As a alternative option, @NullSoldier mentioned that we could leverage GitHub actions to upload the file(s) and its manifest to AWS, if we so choose. That would reduce some of the complexity of the command (allowing for easier unit testing) and it would allow us to potentially move this command to `chain:export` instead, which would address your concerns with it being hidden.

ironfish-cli/src/commands/service/snapshot.ts
```diff
@@ -0,0 +1,211 @@
...
+import fsAsync from 'fs/promises'
+import os from 'os'
+import path from 'path'
+import { parseNumber } from '../../args'
+import { IronfishCommand } from '../../command'
+import { RemoteFlags } from '../../flags'
+import { ProgressBar } from '../../types'
+
+export default class CreateSnapshot extends IronfishCommand {
+  static hidden = true
```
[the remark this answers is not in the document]
deekerno: I've made some comments [here](https://github.com/iron-fish/ironfish/pull/1655#discussion_r905567173) that partially address this concern.

ironfish-cli/src/commands/service/snapshot.ts
lwisne: Can you explain the Github action methodology a bit more? Not familiar

deekerno: We can define a workflow to deploy artifacts to S3. See [this page](https://dev.to/johnkevinlosito/deploy-static-website-to-s3-using-github-actions-4a0e) for an example. I'm envisioning an AWS instance exporting a snapshot and returning the path of the snapshot, and the GitHub action instructing the instance to send it to S3 through use of the AWS SDK. It doesn't even have to be a GitHub action, really; it could be a separate script that leverages the SDK.

In any case, I'm not sure that keeping the upload operation as a part of the command is the best idea. The most user-facing part of this project is arguably the snapshot import; many more people will use that than the export command. Reducing the scope of the export command to just _exporting_ allows for more choice in what to do with the resultant file; one could decide to just use this command as a backup solution or use it to get an archive that can place on a non-AWS host, e.g. GCP, Yandex, etc.

lwisne: This is definitely something we can do, but IMO these options represent scope creep/overengineering vs the original spec. You had export uploading to a S3 bucket, and only an S3 bucket, and I think sticking to that will be a perfectly solid way to go about it for v1.

And if we are going to upload to an S3 bucket, using Amazon's well-vetted tools to do so seems to be the most robust route. Jason told me the bit of cURL code in `backup.ts` is broken going forward, because the file is too big and needs to be handled in some other fashion in order to do a S3 upload. This is exactly the kind of issue that using the SDK would likely prevent, in addition to my current comment about handling error states gracefully.

I'm just having a lot of trouble seeing the problem we are avoiding here by not choosing the simple, obvious solution of uploading a file to S3 via Amazon's S3 SDK.

deekerno: Fair enough, I'll use the SDK then.

ironfish-cli/src/commands/service/snapshot.ts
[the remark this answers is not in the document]
deekerno: In light of the discussion in the comment thread linked above, my opinion is to keep this command as it is currently: a hidden service command. We can move to replace `chain:export` with this functionality should we feel that revisiting it is necessary.

wd021:
@deekerno @lwisne just to chime in on adding the AWS SDK here. It's quite a big package and if we're only using the S3 upload function, we should probably use their [modular version](https://github.com/aws/aws-sdk-js-v3), which allows us to add S3 without all the other services.

`aws-sdk` -> `@aws-sdk/client-s3` ([docs](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/index.html))

deekerno:
> @deekerno @lwisne just to chime in on adding the AWS SDK here. It's quite a big package and if we're only using the S3 upload function, we should probably use their [modular version](https://github.com/aws/aws-sdk-js-v3), which allows us to add S3 without all the other services.
> 
> 
> 
> `aws-sdk` -> `@aws-sdk/client-s3` ([docs](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/index.html))

Thanks @wd021! I still need to do some testing to make sure all of the flow and error handling works, so I hadn't yet moved to minimize the imports, but it's definitely the next step.

yarn.lock
```diff
@@ -2509,6 +2509,21 @@ atob@^2.1.2:
...
+    events "1.1.1"
+    ieee754 "1.1.13"
+    jmespath "0.16.0"
+    querystring "0.2.0"
+    sax "1.2.1"
+    url "0.10.3"
+    uuid "8.0.0"
+    xml2js "0.4.19"
+
 aws-sdk@^2.1069.0:
```
lwisne, line 2527: hmmm..it appears we already install aws?

ironfish-cli/package.json
```diff
@@ -26,6 +26,7 @@
     "@oclif/test": "2.1.0",
     "@types/blessed": "0.1.17",
     "@types/node": "16.11.1",
+    "@types/tar": "6.1.1",
```
lwisne, line 29: you don't check in the resulting .lock file here

ironfish/src/rpc/routes/chain/snapshotChain.test.ts
```diff
@@ -0,0 +1,52 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import bufio from 'bufio'
+import { Assert } from '../../../assert'
+import { getBlockSize, writeBlock } from '../../../network/utils/block'
+import { useMinerBlockFixture } from '../../../testUtilities'
+import { createRouteTest } from '../../../testUtilities/routeTest'
+import { SnapshotChainStreamResponse } from './snapshotChain'
+
+describe('Route chain/snapshotChainStream', () => {
```
lwisne, line 11: This needs a number of tests.

You want to handle all the obvious failure cases and parameters. For example

- `start` and `end` could not be implemented at all and the tests would pass
- are forks handled?
- does `maxBlocksPerChunk` do anything?

ironfish-cli/src/commands/service/snapshot.test.ts
```diff
@@ -0,0 +1,122 @@
...
+        stat: jest.fn(() => mockStats),
+      }
+    })
+  })
+
+  afterAll(() => {
+    jest.dontMock('@ironfish/sdk')
+  })
+
+  describe('exports a snapshot of the chain and uploads it', () => {
```
lwisne, line 113: Part of the reason you're having trouble writing tests is you aren't thinking of them in a compartmentalized way.

If you have CLI command A that calls function B, C , and D, you generally want to unit test B, C, and D and prove that they fulfill their contract in various conditions. The unit testing allows you to relatively easily vary params and such.

Then you integration test A fully, which tends to be the hardest part because you have to do things like corral output streams and mock file systems.

The most useful and flexible tests tend to be the unit tests. You are focusing in both cases here on integration tests.

ironfish-cli/src/commands/service/snapshot.ts
```diff
@@ -0,0 +1,201 @@
...
+      required: false,
+      description: 'S3 bucket to upload snapshot to',
+    }),
+    path: Flags.string({
+      char: 'p',
+      parse: (input: string): Promise<string> => Promise.resolve(input.trim()),
+      required: false,
+      description: 'The path where the snapshot should be saved',
+    }),
+    maxBlocksPerChunk: Flags.integer({
```
lwisne, line 35: this doesn't appear to do anything.

ironfish-cli/src/commands/service/snapshot.ts
```diff
@@ -0,0 +1,201 @@
...
+
+  static flags = {
+    ...RemoteFlags,
+    bucket: Flags.string({
+      char: 'b',
+      parse: (input: string) => Promise.resolve(input.trim()),
+      required: false,
+      description: 'S3 bucket to upload snapshot to',
+    }),
+    path: Flags.string({
```
lwisne, line 29: would prefer that you test setting this.

ironfish/src/rpc/routes/chain/__fixtures__/snapshotChain.test.ts.fixture
```diff
@@ -0,0 +1,98 @@
...
+      },
+      "transactions": [
+        {
+          "type": "Buffer",
+          "data": "base64:AAAAAAAAAAABAAAAAAAAAAAAAAAAAAAAAAAAAI1hVaR9gEEyL+H3c5fy2qxTj2TZzgqhIwxKo4YREV65lS1aemw+9r9TmtcfvpOT/6FEMEd1pRD+aHiqieZaCu0CGNxQEpJNQ+Zi4R0lcrDxXNk6iixRw2a4VXQEhD6RaxBNfWO+CyI743TuEhUunfNNadPwmitGgOcMLnEhZNKXOiOXSNyCNPxIwHT9v6kLOYi5pNRlKt0xtA/IC8YyI4BnsaElimaOguSWdsQ9wEqRymbV2QQwtxMtGtO1OjSBiiLaNmG1l1gEapVUls0vl3j3UqXdnj/m6zM7x12E5BHFbWUvClHA2YCdc1Eo+2ayQ5fel/kGr1f00FkYGGrgqiOG4pdtPWXypZa/+oezCe51me/3hgM6yzyaOgI7xCgtjR2gbaY4V2mc4LcV83FV1Pk43HOPXfHXyg7HBDDjE4B8iq7xbZsJrNN/K6LD/o7L00Ea5FLNxwwg04Rb1ye9sQoWIWB1SBybOxbKD+8EtCoyk+21PFW3kNr9EDjO+Eg+/X1TFUJlYW5zdGFsayBub3RlIGVuY3J5cHRpb24gbWluZXIga2V5MDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwunjsG0P4ID5/BLklsD1svyHWgr3V7Fg44yVIsAR/DUui87d4XnDJgZldJA7sUUva4jC+FbLtj9gdLEcbAOuVAw=="
+        }
+      ]
+    }
+  ],
+  "Route chain/snapshotChainStream correctly returns a block buffer": [
```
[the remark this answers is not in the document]
deekerno: Hmm, it does seem that the fixture needs to be regenerated. I'll put that up in a new commit.

ironfish/src/rpc/routes/chain/snapshotChain.test.ts
deekerno: The schema has been adjusted, and a new test that explicitly tests `maxBlocksPerChunk` has been added.

Regarding forks, this route will only return blocks on the main chain by way of `Blockchain.getHeaderAtSequence()`; forks would be returned if we used the alternative `Blockchain.getHeadersAtSequence()`.

yarn.lock
wd021: who knew aws-sdk is a dependecy of `oclif`.. on further inspection, `oclif` install size is 100mb+. any appetite for migrating to `commander.js`?

ironfish/src/rpc/routes/chain/snapshotChain.ts
```diff
@@ -0,0 +1,101 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import bufio from 'bufio'
+import * as yup from 'yup'
+import { Assert } from '../../../assert'
+import { getBlockSize, writeBlock } from '../../../network/utils/block'
+import { BlockchainUtils } from '../../../utils/blockchain'
+import { ApiNamespace, router } from '../router'
+
+const MAX_BLOCKS_PER_SNAPSHOT_CHUNK = 1000
```
danield9tqh, line 11: I feel like blocks could be super variable in size. Would it make more sense to limit based on bytes? Doesn't seem like an urgent change since hopefully we will have consistent block sizes for the rest of testnet but maybe a follow up change to do

ironfish/src/rpc/routes/chain/snapshotChain.ts
```diff
@@ -0,0 +1,101 @@
...
+  yup
+    .object({
+      start: yup.number().defined(),
+      stop: yup.number().defined(),
+      seq: yup.number().optional(),
+      buffer: yup.mixed<Buffer>().optional(),
+    })
+    .defined()
+
+router.register<typeof SnapshotChainStreamRequestSchema, SnapshotChainStreamResponse>(
```
danield9tqh, line 43: Maybe a future cleanup could be combining this with `exportChainStream`. The only difference seems to be the serialization format. But they're different enough that makes sense to have them separate for this project Just a thought going forward

ironfish/src/rpc/routes/chain/snapshotChain.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    it('correctly returns a serialized list of blocks', async () => {
+      const { chain, strategy } = routeTest
+      await chain.open()
+      strategy.disableMiningReward()
+
+      const genesis = await chain.getBlock(chain.genesis)
+      Assert.isNotNull(genesis)
+
+      const blockA1 = await useMinerBlockFixture(chain, 2)
+      await expect(chain).toAddBlock(blockA1)
```
danield9tqh, line 24: Do we need these expects here? Seems like this is just testing the fixture helpers

yarn.lock
dguenther: Probably not right now... it would be nice if it significantly shrinks the bundle size, but right now we're focused on keeping our team working as quickly as possible leading up to mainnet, and switching CLI libraries would require some ramp-up time and could introduce bugs on various platforms.

ironfish/src/rpc/routes/chain/snapshotChain.test.ts
```diff
@@ -0,0 +1,116 @@
...
+describe('Route chain/snapshotChainStream', () => {
+  const routeTest = createRouteTest()
+
+  describe('with no maxBlocksPerChunk parameter', () => {
+    it('correctly returns a serialized list of blocks', async () => {
+      const { chain, strategy } = routeTest
+      await chain.open()
+      strategy.disableMiningReward()
+
+      const genesis = await chain.getBlock(chain.genesis)
```
danield9tqh, line 20: Same for this, seems like it is just testing our test utils

ironfish/src/rpc/routes/chain/snapshotChain.test.ts
danield9tqh: Didn't realize `expect(chain).toAddBlock(blockA1)` is actually adding the block so that one makes sense

ironfish/src/rpc/routes/chain/snapshotChain.ts
deekerno: I think that's a sensible change. I think it will be easy to add once the main function is included in the node and we see how it's used realistically.

ironfish/src/rpc/routes/chain/snapshotChain.test.ts
deekerno: Yes, you're correct. Additionally, we do this in a fair number of other test suites as well, so I kept this in line with the others.

ironfish/src/rpc/routes/chain/snapshotChain.ts
deekerno: Agreed, Jason and I had talked about this out-of-band and we'll probably change it at some point and put them together, but we'll wait for now as there are some users that do actively use `exportChainStream`.

ironfish-cli/src/commands/service/snapshot.ts
```diff
@@ -0,0 +1,201 @@
...
+
+    const bucket = (flags.bucket || process.env.IRONFISH_SNAPSHOT_BUCKET || '').trim()
+
+    let exportDir
+
+    if (flags.path) {
+      exportDir = this.sdk.fileSystem.resolve(flags.path)
+    } else {
+      try {
+        exportDir = await fsAsync.mkdtemp(`${os.tmpdir()}${path.sep}`)
```
dguenther, line 56: It might be worth investigating whether it's possible to write large files into the os's tmp directory -- considering the snapshots can be 10s of gigabytes, they might cap out the temp directory on linux/windows. I'm not too familiar with limitations of temp directories though

ironfish/package.json
```diff
@@ -93,4 +93,4 @@
     "url": "https://github.com/iron-fish/ironfish/issues"
   },
   "homepage": "https://ironfish.network"
-}
+}
```
dguenther, line 96: This change can probably be removed

dguenther:
I'm done with comments here -- once the outstanding comments have been addressed, I think this is ready to merge.

ironfish-cli/src/commands/service/snapshot.ts
```diff
@@ -0,0 +1,201 @@
...
+    const progress = CliUx.ux.progress({
+      format: 'Retrieving blocks: [{bar}] {value}/{total} {percentage}% | ETA: {eta}s',
+    }) as ProgressBar
+
+    progress.start(stop - start + 1, 0)
+
+    for await (const result of response.contentStream()) {
+      if (result.buffer && result.seq) {
+        const blockFilePath = this.sdk.fileSystem.join(blockExportPath, `${result.seq}`)
+        await fsAsync.writeFile(blockFilePath, Buffer.from(result.buffer))
```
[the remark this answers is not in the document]
deekerno: In both cases, I can generate a UUID and use that to ensure that the block folders don't collide; does that sound good?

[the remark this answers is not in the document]
dguenther: That answers the first question. My second question around cleaning up the `blocks` directory is more about disk space usage -- the export process dumps out 10s-GB of intermediate block files, then zips them up into another 10s-GB archive. It makes sense to keep the archive around, since users will want to do something with it, but are the intermediate block files still needed?

ironfish/src/rpc/routes/chain/snapshotChain.test.ts
dguenther: I think this is fine enough to leave as-is for this PR.

ironfish/src/rpc/routes/chain/snapshotChain.ts
dguenther: This is a good idea for the future, but yeah, since we have existing consumers I don't think we have to worry about it for this PR.

ironfish/package.json
deekerno: Done.

ironfish-cli/src/commands/service/snapshot.ts
deekerno: After researching this for a few minutes, there doesn't seem to be any file size limitation for temp directories for any major OS. Additionally, the temp directory is limited by either the overall hard disk space (Windows) or by the partition where it resides (Unix-based/Linux).

ironfish-cli/package.json
```diff
@@ -108,4 +111,4 @@
     "url": "https://github.com/iron-fish/ironfish/issues"
   },
   "homepage": "https://ironfish.network"
-}
+}
```
dguenther, line 114: Ah, found one more of these newline changes

ironfish-cli/src/commands/service/snapshot.ts
dguenther: Okay, sounds good, let's try this out then and we can follow up if we get any inbound issues from users. This should mostly be used by service operators like us, anyway

ironfish/package.json
```diff
@@ -20,6 +20,7 @@
     "@ironfish/bfilter": "0.0.1",
     "@ironfish/rust-nodejs": "0.1.6",
     "@napi-rs/blake-hash": "1.3.1",
+    "aws-sdk": "2.1163.0",
```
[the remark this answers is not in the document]
dguenther: Looks like this was updated to use the modular import.

yarn.lock
dguenther: Adding a follow-up to this thread as well, the PR was updated to use the modular aws-sdk import.

ironfish-cli/src/commands/service/snapshot.ts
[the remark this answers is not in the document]
deekerno: No, those files are necessarily needed after the fact as they're contained in the archive. I'll add functionality to remove them.

dguenther (approved):

PR #1659, opened 2022-06-22
Add stratum server peer management

NullSoldier:
## Summary

This adds a new StratumPeers system, which tracks bad clients and bans them. It also manages ban timers, and punishment timers. This also ports the old badClients concept into a shadow ban, which is needed because we don't reject peers with bad versions yet.

## Testing Plan
Run pool, and connect with malicious client.

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/ironfish/src/mining/stratum/stratumPeers.ts
+++ b/ironfish/src/mining/stratum/stratumPeers.ts
@@ -1,0 +1,158 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import net from 'net'
+import { Event } from '../../event'
+import { Config } from '../../fileStores/config'
+import { createRootLogger, Logger } from '../../logger'
+import { SetTimeoutToken } from '../../utils'
+import { StratumServerClient } from './stratumServerClient'
+
+const TWO_HOURS_MS = 2 * 60 * 60 * 1000
+const PEERS_TICK_MS = 10000
+
+export class StratumPeers {
+  readonly logger: Logger
+  readonly maxConnectionsByIp: number
+
+  readonly onBanned = new Event<[StratumServerClient]>()
+
+  banCount = 0
+
+  protected readonly connectionsByIp: Map<string, number> = new Map()
+  protected readonly bannedByIp: Map<string, number> = new Map()
+  protected readonly scoreByIp: Map<string, number> = new Map()
+  protected readonly shadowBans: Set<number> = new Set()
+  protected eventLoopTimeout: SetTimeoutToken | null = null
+
+  constructor(options: { maxConnectionsPerIp?: number; config: Config; logger?: Logger }) {
+    this.logger = options.logger ?? createRootLogger()
+
+    this.maxConnectionsByIp =
+      options.maxConnectionsPerIp ?? options.config.get('poolMaxConnectionsPerIp')
+  }
+
+  start(): void {
+    this.eventLoop()
+  }
+
+  stop(): void {
+    if (this.eventLoopTimeout) {
+      clearTimeout(this.eventLoopTimeout)
+    }
+  }
+
+  punish(client: StratumServerClient, reason?: string, amount = 1): void {
+    if (this.isBanned(client.socket)) {
+      return
+    }
+
+    let banScore = this.scoreByIp.get(client.remoteAddress) ?? 0
+    banScore += amount
+
+    if (banScore < 10) {
+      this.scoreByIp.set(client.remoteAddress, banScore)
+      return
+    }
+
+    this.ban(client, reason)
+    this.scoreByIp.delete(client.remoteAddress)
+  }
+
+  shadowBan(client: StratumServerClient): void {
+    this.shadowBans.add(client.id)
+  }
+
+  ban(client: StratumServerClient, reason = 'unknown', until?: number): void {
+    const existing = this.bannedByIp.get(client.remoteAddress) ?? 0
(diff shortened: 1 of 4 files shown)


ironfish/src/mining/stratum/stratumServer.ts
```diff
@@ -171,36 +165,28 @@ export class StratumServer {
           const body = await YupUtils.tryValidate(MiningSubscribeSchema, header.result.body)
 
           if (body.error) {
-            this.addBadClient(client)
-            continue
-            // throw new ClientMessageMalformedError(client, body.error, header.result.method)
+            this.peers.ban(client)
```
mat-if, line 168: This introduces an interesting edge-case. Let's say we implement some change, including a version bump. If it leads to a validation failure, we will ban them for the max amount, since this occurs before the version check below this, which only bans for 15 minutes. Same edge case exists for the ban on L157.

ironfish/src/mining/stratum/stratumPeers.ts
```diff
@@ -0,0 +1,158 @@
...
+    if (score > 0) {
+      this.scoreByIp.set(remoteAddress, score)
+    } else {
+      this.scoreByIp.delete(remoteAddress)
+    }
+  }
+
+  eventLoop(): void {
+    for (const remoteAddress of this.scoreByIp.keys()) {
+      this.unpunish(remoteAddress)
```
mat-if, line 153: Should we include a similar check to remove expired bans? I think we're missing a mechanism to prune the ban list currently.

ironfish/src/mining/stratum/stratumPeers.ts
```diff
@@ -0,0 +1,158 @@
...
+    if (banScore < 10) {
+      this.scoreByIp.set(client.remoteAddress, banScore)
+      return
+    }
+
+    this.ban(client, reason)
+    this.scoreByIp.delete(client.remoteAddress)
+  }
+
+  shadowBan(client: StratumServerClient): void {
```
mat-if, line 63: This PR removes all occurrences of actually shadowbanning anyone. Do we intend on bringing back the functionality of a shadowban?

ironfish/src/mining/stratum/stratumPeers.ts
```diff
@@ -0,0 +1,158 @@
...
+const TWO_HOURS_MS = 2 * 60 * 60 * 1000
+const PEERS_TICK_MS = 10000
+
+export class StratumPeers {
+  readonly logger: Logger
+  readonly maxConnectionsByIp: number
+
+  readonly onBanned = new Event<[StratumServerClient]>()
+
+  banCount = 0
```
mat-if, line 21: This is very much a preference nit, so up to you, but I think a getter is probably a little less prone to bugs. If we prune the expired bans, we can just do a real simple `this.bannedByIp.size`

ironfish/src/mining/stratum/stratumServer.ts
```diff
@@ -23,28 +23,30 @@ import {
   StratumMessage,
   StratumMessageSchema,
 } from './messages'
+import { StratumPeers } from './stratumPeers'
 import { StratumServerClient } from './stratumServerClient'
-import { STRATUM_VERSION_PROTOCOL_MIN } from './version'
+import { STRATUM_VERSION_PROTOCOL, STRATUM_VERSION_PROTOCOL_MIN } from './version'
+
+const FIFTEEN_MINUTES = 15 * 60 * 1000
```
mat-if, line 30: Very nit, so up to you, but I dig the `_MS` suffix for clarity on the consts in the other file.

ironfish/src/mining/stratum/stratumPeers.ts
NullSoldier: It could be expensive, I think you need a priority queue to do it efficiently. I could just do it inefficiently though.

ironfish/src/mining/stratum/stratumPeers.ts
NullSoldier: I think we only ever intend to shadow ban based on version now, and otherwise outright ban them. That should be ok. What do you think?

ironfish/src/mining/stratum/stratumPeers.ts
NullSoldier: This a count of bans, not people banned.

ironfish/src/mining/stratum/stratumServer.ts
NullSoldier: Fixed

ironfish/src/mining/stratum/stratumServer.ts
```diff
@@ -127,26 +127,19 @@ export class StratumServer {
...
   getClientCount(): number {
-    let count = 0
-    for (const client of this.clients.keys()) {
-      if (this.badClients.has(client)) {
-        continue
-      }
-      count += 1
-    }
-    return count
+    return this.clients.size
```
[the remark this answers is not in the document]
NullSoldier: Fixed comment

ironfish/src/mining/stratum/stratumServer.ts
NullSoldier: I think that's an edge case with the handshake message. The developer needs to make all changes to the handshake message subscribe backwards compatible.

ironfish/src/mining/stratum/stratumPeers.ts
NullSoldier: Oh I see, some how the shadow ban call got lost in a rebase.

mat-if (approved):

ironfish/src/mining/stratum/stratumPeers.ts
NullSoldier: I have an idea for this I'll do it another PR.

PR #1668, opened 2022-06-23
Replaced match with if else

hamza-siddiq:
## Summary
Since we're checking for true false values here, the if else statement makes more sense

## Testing Plan
None needed

## Breaking Change
```
[ ] Yes
[✓ ] No
```
graffiti: ironfishup

### DIFF
--- a/ironfish-rust/src/test_util.rs
+++ b/ironfish-rust/src/test_util.rs
@@ -20,9 +20,10 @@ pub(crate) fn make_fake_witness(note: &Note) -> Witness {
 
     let mut witness_auth_path = vec![];
     for _ in 0..TREE_DEPTH {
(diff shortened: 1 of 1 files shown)


NullSoldier:
First, we welcome external contributions and are happy that you are contributing to the project. However, this is a frivolous change, and frivolous changes don't come without a cost.

This doesn't change any behavior, it only introduces stylistic change to match your preference without offering any explanation as to why it's a good idea. You also don't offer any guidance on how other developers should follow this pattern, if indeed it's a better pattern to follow. What if someone comes through and decides they ternaries, and changes to their preference? What should we do then?

That is not to say that I don't welcome cleanups, but this "cleanup" has no substance, or explanation, if that's even what it's to be called. I'm closing this PR.

PR #1671, opened 2022-06-23
Removed unnecessary negation

hamza-siddiq:
## Summary
Removed unnecessary negation in if-else statements, to improve code readability

## Testing Plan
None needed
## Breaking Change

```
[ ] Yes
[ ✓] No
```
graffiti: ironfishup

### DIFF
--- a/ironfish-rust/src/keys/mod.rs
+++ b/ironfish-rust/src/keys/mod.rs
@@ -124,6 +124,7 @@ impl<'a> SaplingKey {
                     Self::new(byte_arr)
                 } else {
                     Err(errors::SaplingKeyError::InvalidPaymentAddress)
+                }
             }
         }
     }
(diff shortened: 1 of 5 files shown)


NullSoldier:
I'm going to repost my comment from https://github.com/iron-fish/ironfish/pull/1668 but please read it..

```
First, we welcome external contributions and are happy that you are contributing to the project. However, this is a frivolous change, and frivolous changes don't come without a cost.

This doesn't change any behavior, it only introduces stylistic change to match your preference without offering any explanation as to why it's a good idea. You also don't offer any guidance on how other developers should follow this pattern, if indeed it's a better pattern to follow. What if someone comes through and decides they like ternaries, and changes this to their preference? What should we do then?

That is not to say that I don't welcome cleanups, but this "cleanup" has no substance, or explanation, if that's even what it's to be called. I'm closing this PR.
```

hamza-siddiq:
Thanks @NullSoldier . I'll keep that in mind for the future :)

PR #1674, opened 2022-06-24
WIP: Add stratum client disconnect

NullSoldier:
## Summary

Sends messages to stratum clients with disconnect reasons so they know not to try to reconnect.

## Testing Plan
Run stratum server and clients, and ban users.

## Breaking Change

```
[ ] Yes
[x] No
```


ironfish/src/mining/stratum/stratumServer.ts
```diff
@@ -179,19 +184,22 @@ export class StratumServer {
...
+            })
             return
           }
 
           client.publicAddress = body.result.publicAddress
           client.subscribed = true
 
           if (!isValidPublicAddress(client.publicAddress)) {
-            this.peers.ban(client, `Invalid public address: ${client.publicAddress}`)
+            this.peers.ban(client, {
```
mat-if, line 200: A ban for this, and some of the other cases feels a bit extreme. It's something easily fixable, and very possible to just typo/copy-paste error. Granted, we have guards on the miner in this case, but if someone is building a different miner and typoes, they now have to wait 15 minutes to try again.

I wonder if a shadow-ban and a disconnect reason would be a better way to deal with a case like this

mat-if (commented):
Couple things I noticed:
- If you get banned for an 'other' reason, say, bad public address, it seems to side-skirt the ban logic a bit. It doesn't get the reason from the server, only an 'unknown', and it will just keep re-trying on the normal 5s reconnect timer. It seems like we could surface the message string to give miners a clue as to why they're being banned and have it back off appropriately
- If you restart your miner, the server won't resend the ban notice, but rather just disconnects so the miner again goes into the normal 5s reconnect loop

ironfish/src/mining/stratum/stratumServer.ts
NullSoldier: That's a good point, there could be a change here to a less harsh ban but let me bring one more thing up.

The only way this ban can happen is if someone is side stepping the normal client, because this validation happens on the client as well. So if the valid-ness definition diverges, it can also happen. In that case they must be running some custom protocol and are violating protocol.

ironfish/src/mining/stratum/messages.ts
```diff
@@ -43,6 +52,14 @@ export const StratumMessageSchema: yup.ObjectSchema<StratumMessage> = yup
   })
   .required()
 
+export const MiningDisconnectMessageSchema: yup.ObjectSchema<MiningDisconnectMessage> = yup
+  .object({
+    reason: yup.string().optional(),
+    versionExpected: yup.number().optional(),
+    bannedUntil: yup.number().optional(),
+  })
```
mat-if, line 60: Not sure if intentional, but on L17, we add `message` to the `MiningDisconnectMessage`, but we don't add it to the schema here. I think we want to add it here, then we can display it on the client's side right?

mat-if (approved):

PR #1683, opened 2022-06-27
Track known block hashes on Peers

dguenther:
## Summary

Adds a knownBlockHashes LRU to peers. This isn't particularly useful with the current block propagation model, but will be used for the next one.

One potentially dangerous area is the gossip was changed to add the block hashes to all knownPeers and determine whether to rebroadcast the message based on the presence of the hash on a peer, rather than looking at the peer's knownPeers directly.

This doesn't cause any problems in testing, and I think the only potential edge cases would be if a peer is rapidly connecting and disconnecting, causing the LRU to get reset, but a block shouldn't get rebroadcast for an extended duration regardless.

Fixes IRO-2226

## Testing Plan

Added a test for gossip, and tested gossip between a network of 3 peers.

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/network/peerNetwork.test.ts
+++ b/ironfish/src/network/peerNetwork.test.ts
@@ -10,6 +10,7 @@ import http from 'http'
 import net from 'net'
 import ws from 'ws'
 import { Assert } from '../assert'
+import { makeBlockAfter } from '../testUtilities/helpers/blockchain'
 import {
   mockChain,
   mockNode,
@@ -17,6 +18,7 @@ import {
   mockTransaction,
   mockWorkerPool,
 } from '../testUtilities/mocks'
+import { createNodeTest } from '../testUtilities/nodeTest'
 import { DisconnectingMessage } from './messages/disconnecting'
 import { NewBlockMessage } from './messages/newBlock'
 import { NewTransactionMessage } from './messages/newTransaction'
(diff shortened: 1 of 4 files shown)


ironfish/src/network/peers/peer.ts
```diff
@@ -190,6 +192,16 @@ export class Peer {
 
   loggedMessages: Array<LoggedMessage> = []
 
+  /**
+   * Blocks that have been sent or received from this peer. Value is set to true if the block was received
+   * from the peer, and false if the block was sent to the peer.
+   */
+  readonly knownBlockHashes: LRU<Buffer, boolean> = new LRU<Buffer, boolean>(
```
NullSoldier, line 199: There is an alias type we commonly use around if you want to use it, or not. `BlockHash`

ironfish/src/network/peers/peer.ts
```diff
@@ -190,6 +192,16 @@ export class Peer {
...
+    1024,
```
NullSoldier, line 200: since we don't provide getSize() this is how many units. So this should be about `1024 hashes * 32 bytes * 50 peers` which will be about 1.6 MiB.

ironfish/src/network/peerNetwork.ts
```diff
@@ -311,6 +311,27 @@ export class PeerNetwork {
...
+    const block = this.strategy.blockSerde.deserialize(message.block)
+
+    for (const peer of this.peerManager.getConnectedPeers()) {
+      // Don't send the block to peers who already know about it
+      if (peer.knownBlockHashes.has(block.header.hash)) {
+        continue
+      }
+
+      if (peer.send(message)) {
+        peer.knownBlockHashes.set(block.header.hash, false)
```
NullSoldier, line 330: I know we don't technically use this but you use false here, and true in another place. Though if it is true, and someone forgets to use `has()` it'll still work.

NullSoldier:
This PR looks pretty good to me honestly. I'm gonna test test this in production.

ironfish/src/syncer.ts
```diff
@@ -484,6 +484,11 @@ export class Syncer {
 
     const { added, block } = await this.addBlock(peer, newBlock)
 
+    peer.knownBlockHashes.set(block.header.hash, true)
+    for (const knownPeer of peer.knownPeers.values()) {
+      knownPeer.knownBlockHashes.set(block.header.hash, true)
```
danield9tqh, line 489: Should we add this to known peers as well? Especially for blocks it seems like if some of these peers are still syncing it's possible they haven't see this block hash yet. I was not planning on doing this for transactions

ironfish/src/network/peerNetwork.ts
NullSoldier: aaah, I see. In that case it makes sense. I would ignore my comment them. If you want to create an LRUSet I think that would be fine, but otherwise this seems resolved.

NullSoldier (approved):

ironfish/src/network/peerNetwork.test.ts
```diff
@@ -198,6 +200,43 @@ describe('PeerNetwork', () => {
...
+        peer1.knownPeers.set(peer2.getIdentityOrThrow(), peer2)
+        peer2.knownPeers.set(peer1.getIdentityOrThrow(), peer1)
+
+        const newBlockMessage = new NewBlockMessage(strategy.blockSerde.serialize(blockA1))
+
+        const peer1Send = jest.spyOn(peer1, 'send')
+        const peer2Send = jest.spyOn(peer2, 'send')
+        const peer3Send = jest.spyOn(peer3, 'send')
+
+        await peerNetwork['handleGossipMessage'](peer1, newBlockMessage)
```
lwisne, line 229: curious - is there a reason we call the method this way as opposed to `peerNetwork.handleGossipMessage`?

ironfish/src/network/peerNetwork.ts
lwisne: I would agree that it's misleading at first glance that `peer.knownBlockHashes.get(x)` returning false would mean it IS in the set, but it was received vs sent.

For a more memory-friendly but verbose implementation maybe could use an enum that has states like `RECEIVED` and `SENT`?

dguenther: Oh, I like that idea -- let me see if I can make that work with the LRUMap.

ironfish/src/network/peerNetwork.test.ts
dguenther: It's a private method, so we can't call it directly, but typescript lets us sneak around the privacy this way. I'll double check and see if there's a public method we can call.

PR #1688, opened 2022-06-27
Change defaultAccountName in db to defaultAccountId

mat-if:
## Summary

Part 1 of 2 for changing the meta store to the new format: rename `defaultAccountName` to `defaultAccountId` and associated changes.

These changes were just to make the internal logic more consistent around using the id. I think another improvement to consider is making the functions on `accounts` all work off id instead of name, but need to think through that more and didn't want to gum up this PR too much

## Breaking Change

```
[ ] Yes
[ ] No
```

### DIFF
--- a/ironfish/src/account/accounts.ts
+++ b/ironfish/src/account/accounts.ts
@@ -34,10 +34,6 @@ type SyncTransactionParams =
   | Record<string, never>
 
 export class Accounts {
-  readonly onDefaultAccountChange = new Event<
-    [account: Account | null, oldAccount: Account | null]
-  >()
-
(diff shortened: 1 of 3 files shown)


ironfish/src/account/accounts.ts
```diff
@@ -34,10 +34,6 @@ type SyncTransactionParams =
   | Record<string, never>
 
 export class Accounts {
-  readonly onDefaultAccountChange = new Event<
```
mat-if, line 37: This isn't consumed anywhere so removing it. I think this was only used by the old mining director code

rohanjadvani (approved):

PR #1691, opened 2022-06-27
Telemetry for network bandwidth usage per message type

danield9tqh:
## Summary
As part of transaction gossip we want to see how much bandwidth we are using to send and received transaction messages. We could combine this work by just adding telemetry metrics for every message type. 

## Testing Plan
Tested on local node

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/network/peers/connections/webSocketConnection.ts
+++ b/ironfish/src/network/peers/connections/webSocketConnection.ts
@@ -87,9 +87,9 @@ export class WebSocketConnection extends Connection {
       }
 
       let message
+      const byteCount = event.data.byteLength
       try {
         message = parseNetworkMessage(event.data)
-        const byteCount = event.data.byteLength
         this.metrics?.p2p_InboundTraffic.add(byteCount)
         this.metrics?.p2p_InboundTraffic_WS.add(byteCount)
       } catch (error) {
@@ -101,6 +101,8 @@ export class WebSocketConnection extends Connection {
         this.close(new NetworkError(message))
         return
       }
+
+      this.metrics?.p2p_InboundTrafficByMessage.get(message.type)?.add(byteCount)
 
       if (this.shouldLogMessageType(message.type)) {
         this.logger.debug(
@@ -136,6 +138,7 @@ export class WebSocketConnection extends Connection {
     const byteCount = data.byteLength
     this.metrics?.p2p_OutboundTraffic.add(byteCount)
     this.metrics?.p2p_OutboundTraffic_WS.add(byteCount)
+    this.metrics?.p2p_OutboundTrafficByMessage.get(message.type)?.add(byteCount)
 
     return true
   }
--- a/ironfish/src/network/peers/connections/webSocketConnection.ts
+++ b/ironfish/src/network/peers/connections/webSocketConnection.ts
@@ -86,12 +86,15 @@ export class WebSocketConnection extends Connection {
         return
       }
 
+      this.metrics?.p2p_InboundTraffic.add(event.data.byteLength)
+      this.metrics?.p2p_InboundTraffic_WS.add(event.data.byteLength)
+
       let message
-      const byteCount = event.data.byteLength
+
       try {
         message = parseNetworkMessage(event.data)
-        this.metrics?.p2p_InboundTraffic.add(byteCount)
-        this.metrics?.p2p_InboundTraffic_WS.add(byteCount)
+
+        this.metrics?.p2p_InboundTrafficByMessage.get(message.type)?.add(event.data.byteLength)
       } catch (error) {
         // TODO: any socket that sends invalid messages should probably
         // be punished with some kind of "downgrade" event. This should
@@ -101,8 +104,6 @@ export class WebSocketConnection extends Connection {
         this.close(new NetworkError(message))
         return
       }
-
-      this.metrics?.p2p_InboundTrafficByMessage.get(message.type)?.add(byteCount)
 
       if (this.shouldLogMessageType(message.type)) {
         this.logger.debug(
--- a/ironfish/src/network/types.ts
+++ b/ironfish/src/network/types.ts
@@ -20,12 +20,6 @@ export enum NetworkMessageType {
   SignalRequest = 12,
 }
 
-// Enums contain bi-directional mappings so filter out duplicate enum keys
-// This is used for some cases where we need to do some action for each message type
-export const NetworkMessageTypeList = Object.keys(NetworkMessageType)
-  .map((k) => parseInt(k))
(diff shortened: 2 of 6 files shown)


ironfish/src/telemetry/telemetry.ts
```diff
@@ -131,12 +164,12 @@ export class Telemetry {
         {
           name: 'inbound_traffic',
           type: 'float',
-          value: this.metrics.p2p_InboundTraffic.rate1s,
```
danield9tqh, line 134: I'm not sure why this was `rate1s` before but it seems to make more sense to send the 5m rolling average if we are sending stats every 5 minutes

ironfish/src/network/peers/connections/webSocketConnection.ts
```diff
@@ -102,6 +102,8 @@ export class WebSocketConnection extends Connection {
         return
       }
 
+      this.metrics?.p2p_InboundTrafficByMessage.get(message.type)?.add(byteCount)
```
NullSoldier, line 105: any reason not to put these with the others so all the same code is together?

ironfish/src/network/types.ts
```diff
@@ -20,6 +20,12 @@ export enum NetworkMessageType {
   SignalRequest = 12,
 }
 
+// Enums contain bi-directional mappings so filter out duplicate enum keys
+// This is used for some cases where we need to do some action for each message type
+export const NetworkMessageTypeList = Object.keys(NetworkMessageType)
+  .map((k) => parseInt(k))
+  .filter((k) => !isNaN(k))
```
NullSoldier, line 27: Do you want to move this into an IntEnumUtils similar to our StrEnumUtils? It would be nice if this code was reusable for IntEnums in case anyone else needed this too. https://github.com/iron-fish/ironfish/blob/master/ironfish/src/utils/enums.ts

ironfish/src/telemetry/telemetry.ts
NullSoldier: Seems fine to change

ironfish/src/telemetry/telemetry.ts
```diff
@@ -114,9 +115,41 @@ export class Telemetry {
...
+      return value
+        ? [
+            {
+              name: 'inbound_traffic_' + NetworkMessageType[messageType].toLowerCase(),
+              type: 'float',
+              value: value,
+            },
+          ]
+        : []
+    })
```
NullSoldier, line 129: What is your thinking behind filtering out empty meters? Is this related to P2P connectedness?

ironfish/src/network/types.ts
danield9tqh: Yea that's a good idea. I'll do that

ironfish/src/telemetry/telemetry.ts
danield9tqh: Just for the type system. For some reason the `Assert.isNotNull(this.metrics)` was not guaranteeing that `this.metrics` was not null inside the lambda function. Not sure why. I didn't really look into why but I can investigate a little

ironfish/src/network/peers/connections/webSocketConnection.ts
danield9tqh: Nope I'll combine. Also was wondering if we should record `p2p_InboundTraffic` even if the message is not parsed correctly. Don't need to change it now but something to consider

ironfish/src/network/peers/connections/webSocketConnection.ts
```diff
@@ -87,9 +87,9 @@ export class WebSocketConnection extends Connection {
       }
 
       let message
+      const byteCount = event.data.byteLength
```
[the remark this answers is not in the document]
danield9tqh: Yea I'll clean up this as well moving the 3 calls together below

ironfish/src/network/peers/connections/webSocketConnection.ts
```diff
@@ -86,12 +86,15 @@ export class WebSocketConnection extends Connection {
         return
       }
 
+      this.metrics?.p2p_InboundTraffic.add(event.data.byteLength)
```
danield9tqh, line 89: moving these up here so that we record inbound traffic even if the message fails to parse. We were doing this for WebRTC already

ironfish/src/telemetry/telemetry.ts
```diff
@@ -76,7 +77,9 @@ export class Telemetry {
     void this.flushLoop()
 
     if (this.metrics) {
-      void this.metricsLoop()
+      this.metricsInterval = setTimeout(() => {
```
danield9tqh, line 80: Adding this change so that we don't log metrics on initial node startup. This would cause (and was already causing) nodes to send messages with network traffic = 0 because there hasn't been any messages yet. This seems like it would lead to inaccurate data. We want to allow some collection time before messages are sent

ironfish/src/telemetry/telemetry.ts
danield9tqh: found a better way to do this by just looping through the list of `Meter`s

ironfish/src/telemetry/telemetry.ts
```diff
@@ -114,9 +117,35 @@ export class Telemetry {
...
+
     this.submit({
       measurement: 'node_stats',
       timestamp: new Date(),
+      tags: [
+        {
+          name: 'synced',
+          value: this.chain.synced.toString(),
+        },
+      ],
```
NullSoldier, line 148: this will partition all the fields into 2 tables, I think this should be a measurement right?

NullSoldier, line 148: Will this be ok for stats, now that were segmenting stats between synced and non synced nodes?

ironfish/src/telemetry/telemetry.ts
NullSoldier: This is mostly just personal thought, but usually if you need a comment like this, it would be better to encode that explanation into logic. Maybe just check that we don't submit those metrics until a certain amount of time has passed and were connected to the network or something?

NullSoldier (approved):

NullSoldier (approved):

ironfish/src/telemetry/telemetry.ts
danield9tqh: It looks like we are using a `pivot` for [that query ](https://github.com/iron-fish/ironfish-telemetry/blob/3fc1b082f3f168ffe90881d28df2253c78e0965c/grafana/dashboards/dashboards.json#L475) so I think this should not affect the queries. It could affect performance but there's only 2 values so I feel like it should be relatively minor

PR #1694, opened 2022-06-28
Add witness hash method to transactions

danield9tqh:
## Summary
For P2P transaction gossip we need a hash of the transaction that commits to the witness (signature). We will use this hash to record which transactions have been gossiped to which peers. Zcash's [ZIP-239](https://zips.z.cash/zip-0239#motivation) provides a good explanation of this need and how using a hash that does not commit to the witness can be a security flaw for transaction gossip

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/primitives/transaction.ts
+++ b/ironfish/src/primitives/transaction.ts
@@ -3,6 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 import { TransactionPosted } from '@ironfish/rust-nodejs'
+import { blake3 } from '@napi-rs/blake-hash'
 import bufio from 'bufio'
 import { Serde } from '../serde'
 import { NoteEncrypted } from './noteEncrypted'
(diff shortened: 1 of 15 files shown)


ironfish/src/primitives/transaction.ts
```diff
@@ -182,6 +183,14 @@ export class Transaction {
     return this.withReference((t) => t.hash())
   }
 
+  /**
+   * Genereate the hash of a transaction that includes the witness (signature) data.
+   * Used for cases where a signature needs to be commited to in the hash like P2P transaction gossip
+   */
+  witnessHash(): TransactionHash {
```
NullSoldier, line 190: This isn't a hash of the witness, this is a hash of the posted transaction.

NullSoldier: Seems fine now.

NullSoldier (approved):
Looks good. I think the next step here is evaluating if we want to switch over to use the hash() instead of the unsigned hash everywhere.

PR #1704, opened 2022-06-29
adds cli command to get pool status

hughy:
## Summary

background: pool users (and pool operators) often have questions about whether
the pool is running correctly, what their aggregate hashrate is, how many shares
they have submitted, etc. these changes add a cli command for fetching status
information from a running pool process.

- uses a StratumClient to get status information from a mining pool
- defines messages for MiningPoolStatus
- adds getStatus and onStatus to StratumClient

- removes publicAddress property from StratumClient: client only needs the
  address for 'subscribe' messages, so miner can pass its address when it calls
  subscribe.

- adds follow option to pool status command

## Testing Plan

local testing with node, pool, miner

## Breaking Change

```
[ ] Yes
```


ironfish/src/mining/pool.ts
```diff
@@ -421,26 +422,45 @@ export class MiningPool {
...
 
-export type MiningPoolStatus = {
-  name: string
-  hashRate: number
-  miners: number
-  sharesPending: number
-  banCount: number
+    if (publicAddress) {
+      let addressMiners = 0
+      this.stratum.clients.forEach((client) => {
```
NullSoldier, line 448: It's small, but do we need to create a closure here for each status call instead of just using a for loop and iterator from the map.values()? Normally I think it's what ever but there may be a lot of people connected that are using this.

ironfish/src/mining/pool.ts
```diff
@@ -421,26 +422,45 @@ export class MiningPool {
...
+        }
+      })
+
+      return {
+        ...status,
+        addressStatus: {
+          publicAddress: publicAddress,
+          hashRate: await this.estimateHashRate(publicAddress),
+          miners: addressMiners,
+          sharesPending: await this.shares.sharesPendingPayout(publicAddress),
```
NullSoldier, line 460: can we do these awaits inside the top promise.all?

ironfish/src/mining/poolDatabase/database.ts
```diff
@@ -71,6 +71,17 @@ export class PoolDatabase {
...
   }
 
+  async getAddressSharesCountForPayout(publicAddress: string): Promise<number> {
+    const result = await this.db.get<{ count: number }>(
+      'SELECT COUNT(*) AS count from share WHERE payoutId IS NULL and publicAddress = ?',
+      publicAddress,
+    )
+    if (result == null) {
+      return 0
+    }
```
NullSoldier, line 81: Can the sqlite driver really return null here?

ironfish/src/mining/poolDatabase/database.ts
```diff
@@ -120,6 +131,18 @@ export class PoolDatabase {
...
+
+  async addressShareCountSince(publicAddress: string, timestamp: number): Promise<number> {
+    const result = await this.db.get<{ count: number }>(
+      "SELECT COUNT(id) AS count FROM share WHERE publicAddress = ? AND createdAt > datetime(?, 'unixepoch')",
+      publicAddress,
+      timestamp,
+    )
+    if (result == null) {
+      return 0
+    }
```
NullSoldier, line 143: Can the sqlite driver really return null here?

ironfish/src/mining/poolDatabase/database.ts
```diff
@@ -71,6 +71,17 @@ export class PoolDatabase {
     return result.count
   }
 
+  async getAddressSharesCountForPayout(publicAddress: string): Promise<number> {
+    const result = await this.db.get<{ count: number }>(
+      'SELECT COUNT(*) AS count from share WHERE payoutId IS NULL and publicAddress = ?',
```
NullSoldier, line 76: These aren't indexed, should we add an index?

ironfish/src/mining/poolShares.ts
```diff
@@ -206,13 +206,16 @@ export class MiningPoolShares {
...
   }
 
-  private async recentShareCount(): Promise<number> {
+  private async recentShareCount(publicAddress?: string): Promise<number> {
     const timestamp = Math.floor(new Date().getTime() / 1000) - this.recentShareCutoff
 
+    if (publicAddress) {
+      return await this.db.addressShareCountSince(publicAddress, timestamp)
+    }
     return await this.db.shareCountSince(timestamp)
```
NullSoldier, line 219: nit: We're mixing two patterns here, which is fine but it's just interesting. All these functions were parameterized with an optional arg, but all the DB functions were duplicated.

ironfish/src/mining/stratum/messages.ts
```diff
@@ -44,6 +44,24 @@ export type MiningNotifyMessage = {
   header: string
 }
 
+export type MiningGetStatusMessage = {
+  publicAddress?: string
+}
```
NullSoldier, line 49: Idk if we want to make this similar to MiningDisconnectMessage when we have no required arguments, or just force all messages to have a payload. I'm fine either way.

```suggestion
export type MiningGetStatusMessage = {
  publicAddress?: string
} | undefined
```

NullSoldier:
This is a pretty good PR. I wonder if at some point we should add RPC style messages to this down the road. It's nice to keep the protocol simple for now, though.

ironfish/src/mining/pool.ts
hughy: Good call, I think what I really wanted here was something like a Python generator expression, but I didn't think about closures.

ironfish/src/mining/poolDatabase/database.ts
hughy: I think `if (result === undefined)` is more correct here. The sqlite driver will return undefined if the result set is empty. That shouldn't happen with this query in practice, but I suppose it's possible.

NullSoldier: Ok, if it's possible then I'm curious what the conditions are for a count(*) query to be empty. I can't get it to happen even when the table is empty.

ironfish/src/mining/stratum/stratumServer.ts
```diff
@@ -247,6 +249,24 @@ export class StratumServer {
           break
         }
 
+        case 'mining.get_status': {
+          const body = await YupUtils.tryValidate(MiningGetStatusSchema, header.result.body)
+
+          if (body.error) {
+            this.peers.ban(client, {
+              message: body.error.message,
+            })
```
NullSoldier, line 258: ah there is a tricky bug here with fall through.

NullSoldier (approved):
I'm approving, the PR looks great except for one last bug.

PR #1707, opened 2022-06-29
Store headHash per account to allow for syncing accounts

mat-if:
## Summary

Allows future improvements like syncing accounts individually and maintaining state separately

Closes IRO-2223

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/account/accounts.ts
+++ b/ironfish/src/account/accounts.ts
@@ -1251,7 +1251,6 @@ export class Accounts {
     await this.db.setAccount(account)
   }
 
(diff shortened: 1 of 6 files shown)


ironfish/src/account/accounts.ts
```diff
@@ -1239,6 +1251,34 @@ export class Accounts {
     await this.db.setAccount(account)
   }
 
+  /// Find and return the earliest occurring head hash in the account head hashes map
```
rohanjadvani, line 1254: We can remove this comment. The method name is self descriptive.

rohanjadvani (approved):

PR #1708, opened 2022-06-29
feat(ironfish): Move note decryption when syncing transactions to worker pool

rohanjadvani:
## Summary

We currently decrypt notes in the main thread when syncing transactions in the accounts module. This code change migrates this processing to the worker pool.

Note: We return a `serializedNote` from the worker job. This is currently unused in `staging`, but is part of the other broader accounts changes.

## Testing Plan

Created unit tests for serialization; existing unit tests should cover logic.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/account/accounts.ts
+++ b/ironfish/src/account/accounts.ts
@@ -17,6 +17,7 @@ import { ValidationError } from '../rpc/adapters/errors'
 import { IDatabaseTransaction } from '../storage'
 import { PromiseResolve, PromiseUtils, SetTimeoutToken } from '../utils'
 import { WorkerPool } from '../workerPool'
+import { DecryptNoteOptions } from '../workerPool/tasks/decryptNotes'
 import { UnspentNote } from '../workerPool/tasks/getUnspentNotes'
 import { Account } from './account'
 import { AccountDefaults, AccountsDB } from './accountsdb'
@@ -354,7 +355,7 @@ export class Accounts {
     }>
   > {
     const accounts = this.listAccounts()
-    const notes = new Array<{
+    const decryptedNotes = new Array<{
       noteIndex: number | null
       nullifier: string | null
       merkleHash: string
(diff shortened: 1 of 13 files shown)


ironfish/src/account/accounts.ts
```diff
@@ -374,39 +376,21 @@ export class Accounts {
...
-            })
-          }
-          continue
+        const decryptedNote = await this.workerPool.decryptNotes(
+          note.serialize(),
+          account.incomingViewKey,
+          account.outgoingViewKey,
+          account.spendingKey,
+          currentNoteIndex,
+        )
```
NullSoldier, line 385: I think this code is REALLY clean and easy to read.

```suggestion
        )

```

ironfish/src/workerPool/tasks/decryptNotes.ts
```diff
@@ -0,0 +1,242 @@
...
+      if (this.note.nullifier) {
+        size += 32
+      }
+    }
+
+    return size
+  }
+}
+
+export class DecryptNotesTask extends WorkerTask {
```
NullSoldier, line 183: This is plural but only takes a single note

ironfish/src/common/constants.ts
```diff
@@ -0,0 +1,6 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+export const ENCRYPTED_NOTE_LENGTH = 32 + 32 + 32 + 83 + 16 + 64 + 16
+export const KEY_LENGTH = 32
+export const NOTE_LENGTH = 43 + 8 + 32 + 32
```
NullSoldier, line 6: I think these constants should live near the primitives that use them. `ENCRYPTED_NOTE_LENGTH` and `NOTE_LENGTH` should live in primitives where these numbers are derived from. `KEY_LENGTH` doesn't make sense because if I see this I have no idea which key this is the length of in our entire code base.

ironfish/src/workerPool/tasks/decryptNotes.ts
```diff
@@ -0,0 +1,242 @@
...
+    const bw = bufio.write(this.getSize())
+
+    const hasDecryptedNote = Number(!!this.note)
+    bw.writeU8(hasDecryptedNote)
+
+    if (this.note) {
+      let flags = 0
+      flags |= Number(!!this.note.index) << 0
+      flags |= Number(!!this.note.nullifier) << 1
+      flags |= Number(this.note.forSpender)
```
mat-if, line 110:
```suggestion
      flags |= Number(this.note.forSpender) << 2
```

mat-if (approved):
sorry i thought i approved this hours ago 😬

ironfish/src/common/constants.ts
rohanjadvani: Moved

ironfish/src/account/accounts.ts
```diff
@@ -374,39 +376,21 @@ export class Accounts {
...
-            notes.push({
-              noteIndex: currentNoteIndex,
-              forSpender: true,
-              merkleHash: note.merkleHash().toString('hex'),
-              nullifier: null,
-              account: account,
-            })
-          }
-          continue
+        const decryptedNote = await this.workerPool.decryptNotes(
```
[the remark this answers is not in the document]
rohanjadvani: Yep it seems to work faster. Updated to batch process.

ironfish/src/workerPool/tasks/decryptNotes.ts
rohanjadvani: Good catch, fixed

ironfish/src/workerPool/tasks/decryptNotes.ts
rohanjadvani: Keeping as is since this now takes multiple payloads for decryption

NullSoldier (approved):

NullSoldier:
I'm really looking forward to the perf boost this gives us.

mat-if (approved):

Issue #1677, opened 2022-06-26
[FIP] Ironfish auth system for Rpc call


hairtail:
### Description

Support different level of permission for Ironfish Rpc
- Read: Read node state, no private data.
- Write: Write to local store / chain, and read permissions.
- Sign: Use private keys stored in accounts for signing, read and write permissions.
- Admin: Manage permissions, read, write and sign permissions.

NullSoldier:
I'm curious about this, because it's an important feature that we need.

Instead of Read / Write / Sign /Admin should we just do this based on RPC routes / namespaces? You could just have preset configurations called Read, Write, Sign, Admin where each one is just a set of namespaces / requests that we decide makes sense, but users could also provision keys with their own set of requests.

for instance `config/getConfig` is read only, but could reveal sensitive information. I would only mount this request and make it available to keys with `Admin` level even though it's a read only api.

hairtail:
Cool, what you have mentioned above is kind of easy way to implement auth system. If everything is ok, I will try to work on it.

hairtail:
Working on auth sys with [Bearer token](https://oauth.net/2/bearer-tokens/). Let me know if there's some update or anything wrong.

hughy:
RPC auth tokens were added in #2224

PR #1709, opened 2022-06-30
Fix: add account route for rpctcp

hairtail:
## Summary
Mount account based route for rpctcp call.
## Testing Plan
Local test passed.
## Breaking Change

```
[ ] Yes
[✓] No
```

### DIFF
--- a/ironfish/src/sdk.ts
+++ b/ironfish/src/sdk.ts
@@ -220,6 +220,7 @@ export class IronfishSdk {
 
     if (this.config.get('enableRpcTcp')) {
       const namespaces = [
+        ApiNamespace.account,
         ApiNamespace.chain,
         ApiNamespace.event,
         ApiNamespace.faucet,


ironfish/src/sdk.ts
```diff
@@ -220,6 +220,7 @@ export class IronfishSdk {
 
     if (this.config.get('enableRpcTcp')) {
       const namespaces = [
+        ApiNamespace.account,
```
NullSoldier, line 223: This is breaking security for RPC based calls to the TCP adapter. If you want this you should set `rpcTcpSecure` flag to true. You can also use the global flag `--rpc.tcp.secure`. Look at line 236 below in your own PR. https://github.com/iron-fish/ironfish/pull/1709/files#diff-c6255a87960ceb1747bb6bfadb135b4219244345cc96dc33c760fbad5c2122dcR237. You can also use the flag `

hairtail: Disagree. Transaction based routes are supported for tcp adapter, so real funds can be sent to other external account. So it's not a problem for account based routes. By the way, I have opened an FIP for Ironfish permission control system and work on it now. So this security problem would be fixed with auth instead of route control. [Auth System FIP](https://github.com/iron-fish/ironfish/issues/1677)

PR #1712, opened 2022-06-30
Fix chain:export progress bar

vchs1v:
## Summary
Currently, block sequence is displayed as a number of processed blocks in a progress bar. This can lead to incorrect number if user passed a block to start from:
176679308-d2fb7dbd-2c6c-4d76-814a-1236b1cda3fa

This PR fixes this issue:
176679357-65d6b040-d1b4-4826-aa3c-bce9bf0d7a12

## Testing Plan
Tested locally

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/chain/export.ts
+++ b/ironfish-cli/src/commands/chain/export.ts
@@ -67,7 +67,7 @@ export default class Export extends IronfishCommand {
 
     for await (const result of stream.contentStream()) {
       results.push(result.block)
-      progress.increment()
+      progress.update((result.block?.seq || 0) - start + 1) // +1 because start is inclusive
     }
 
     progress.stop()
--- a/ironfish-cli/src/commands/chain/export.ts
+++ b/ironfish-cli/src/commands/chain/export.ts
@@ -67,7 +67,7 @@ export default class Export extends IronfishCommand {
 
     for await (const result of stream.contentStream()) {
       results.push(result.block)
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/chain/export.ts
```diff
@@ -67,7 +67,7 @@ export default class Export extends IronfishCommand {
 
     for await (const result of stream.contentStream()) {
       results.push(result.block)
-      progress.update(result.block?.seq || 0)
+      progress.increment()
```
NullSoldier, line 70: Unfortunately this doesn't work. I think this needs to be  `stop - (result.block?.seq || 0)`.

The reason is that export will export all blocks at each sequence, including forks. That means you will get multiple blocks at sequence 10, which will increment progress for every block at sequence 10, so the total progress will exceed the `stop` value. You can see that here, https://github.com/iron-fish/ironfish/blob/master/ironfish/src/rpc/routes/chain/exportChain.ts#L75-L79

If we didn't export forks, what you wrote would be fine.

vchs1v: Thank you for the explanation.

I have updated PR considering forks. Now progress will be calculated relying on a `start` position and current block sequence.

NullSoldier (approved):

NullSoldier:
Looks good, thanks for the contribution.

PR #1720, opened 2022-06-30
Upload peer lists to telemetry

danield9tqh:
## Summary
We want to be able to visualize network topology. For this we will need to add peer lists to telemetry. We will upload peer lists from all peers every ~5 minutes

## Testing Plan
Local testing

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/metrics/metricsMonitor.ts
+++ b/ironfish/src/metrics/metricsMonitor.ts
@@ -4,6 +4,7 @@
 
 import os from 'os'
 import { createRootLogger, Logger } from '../logger'
+import { Identity } from '../network'
 import { NetworkMessageType } from '../network/types'
 import { NumberEnumUtils, SetIntervalToken } from '../utils'
 import { Gauge } from './gauge'
@@ -23,6 +24,9 @@ export class MetricsMonitor {
   readonly p2p_InboundTrafficByMessage: Map<NetworkMessageType, Meter> = new Map()
   readonly p2p_OutboundTrafficByMessage: Map<NetworkMessageType, Meter> = new Map()
   readonly p2p_PeersCount: Gauge
+
+  // Elements of this map are managed by Peer and PeerNetwork
+  p2p_OutboundMessagesByPeer: Map<Identity, Meter> = new Map()
 
   readonly heapTotal: Gauge
   readonly heapUsed: Gauge
--- a/ironfish/src/mining/manager.ts
+++ b/ironfish/src/mining/manager.ts
@@ -11,7 +11,6 @@ import { IronfishNode } from '../node'
 import { Block } from '../primitives/block'
 import { Transaction } from '../primitives/transaction'
 import { BlockTemplateSerde, SerializedBlockTemplate } from '../serde'
-import { Telemetry } from '../telemetry'
 import { AsyncUtils } from '../utils/async'
 import { GraffitiUtils } from '../utils/graffiti'
 
@@ -30,23 +29,16 @@ export class MiningManager {
   private readonly chain: Blockchain
   private readonly memPool: MemPool
   private readonly node: IronfishNode
-  private readonly telemetry: Telemetry
 
   blocksMined = 0
   minersConnected = 0
 
   readonly onNewBlock = new Event<[Block]>()
 
-  constructor(options: {
-    chain: Blockchain
-    node: IronfishNode
-    memPool: MemPool
-    telemetry: Telemetry
-  }) {
+  constructor(options: { chain: Blockchain; node: IronfishNode; memPool: MemPool }) {
     this.node = options.node
     this.memPool = options.memPool
     this.chain = options.chain
-    this.telemetry = options.telemetry
   }
 
   /**
@@ -194,7 +186,6 @@ export class MiningManager {
 
     this.blocksMined++
     this.onNewBlock.emit(block)
-    this.telemetry.submitBlockMined(block)
 
     return MINED_RESULT.SUCCESS
   }
--- a/ironfish/src/network/peers/peer.ts
+++ b/ironfish/src/network/peers/peer.ts
@@ -4,6 +4,7 @@
 import colors from 'colors/safe'
 import { Event } from '../../event'
 import { createRootLogger, Logger } from '../../logger'
+import { MetricsMonitor } from '../../metrics'
 import { ErrorUtils } from '../../utils'
 import { Identity } from '../identity'
 import { DisconnectingReason } from '../messages/disconnecting'
@@ -63,6 +64,8 @@ export type PeerState =
 export class Peer {
   readonly pendingRPCMax: number
   readonly logger: Logger
+
+  metrics?: MetricsMonitor
 
   /**
    * The current state of the peer.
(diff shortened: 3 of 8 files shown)


ironfish/src/metrics/metricsMonitor.ts
```diff
@@ -80,6 +83,15 @@ export class MetricsMonitor {
     return meter
   }
 
+  recordP2PMessage(peerId: Identity): void {
+    if (this.p2p_OutboundTrafficByPeer.has(peerId)) {
+      this.p2p_OutboundTrafficByPeer.get(peerId)?.add(1)
+    } else {
+      this.p2p_OutboundTrafficByPeer.set(peerId, this.addMeter())
+      this.recordP2PMessage(peerId)
+    }
```
NullSoldier, line 92: This is a bit of strange recursion. We could simplify this here

```suggestion
    const meter = this.p2p_OutboundTrafficByPeer.get(peerId)

    if(!meter) {
        meter = this.addMeter()
        this.p2p_OutboundTrafficByPeer.set(peerId, meter)
    }

    meter.add(1)
```

ironfish/src/metrics/metricsMonitor.ts
```diff
@@ -80,6 +83,15 @@ export class MetricsMonitor {
     return meter
   }
 
+  recordP2PMessage(peerId: Identity): void {
```
NullSoldier, line 86: This function is gonna cause a big performance and memory leak by adding a new meter for every connected peer because these are not cleaned up and stopped.

ironfish/src/metrics/metricsMonitor.ts
```diff
@@ -21,6 +22,8 @@ export class MetricsMonitor {
   readonly p2p_OutboundTraffic_WebRTC: Meter
   readonly p2p_PeersCount: Gauge
 
+  readonly p2p_OutboundTrafficByPeer: Map<Identity, Meter> = new Map()
```
NullSoldier, line 25: nit: This is messages, not bandwidth right? Can we name this something to indicate its not bandwidth like our other ones. Maybe `OutboundMessagesByPeer`

ironfish/src/telemetry/telemetry.ts
```diff
@@ -114,6 +114,26 @@ export class Telemetry {
   private metricsLoop(): void {
     Assert.isNotNull(this.metrics)
 
+    const peers = [...this.metrics.p2p_OutboundTrafficByPeer.entries()]
+    peers.forEach(([peer, meter]) => {
```
NullSoldier, line 118: Any reason not to just loop over the map itself using the iterator? This makes a copy, and then generates a function versus just using the iterator + for loop

ironfish/src/network/peers/peer.ts
```diff
@@ -211,6 +213,7 @@ export class Peer {
 
   constructor(
     identity: Identity | null,
+    metrics: MetricsMonitor,
```
NullSoldier, line 216: Can this go into the options object below?

NullSoldier:
Is this going to be stored in influx, or postgres? I assume influx since were sending this up with telemetry.

NullSoldier:
I think think this should clean up these meters when peers disconnect. I'm a little worried about our app having all this debugging that no one uses after a time, or our app spending a ton of bandwidth sending a lot of data up in bursts every 5 minutes. I think it's fine in the beginning, but we're gonna have to audit all this logging and telemetry and trim it down for performance reasons.

ironfish/src/metrics/metricsMonitor.ts
danield9tqh: Yea fair. I added the recursion when I was doing a two level deep map by peerID and message type. But with just a one level deep map it doesn't help much

ironfish/src/metrics/metricsMonitor.ts
danield9tqh: Yep I can do that 👍🏻

ironfish/src/telemetry/telemetry.ts
danield9tqh: Yep, I think this was an artifact of playing around with some other map types. I'll change it to just use the iterator

ironfish/src/metrics/metricsMonitor.ts
danield9tqh: Good catch, I'll clean them up

danield9tqh: Since we add them by using `addMeter` function I think these all will be cleaned up actually. The `Map<Identity, Meter>` will not be deleted or anything but that should be taken care of by garbage collection? Is there something else you think we should do in cleanup

ironfish/src/network/peers/peer.ts
danield9tqh: yep, added. I also am removing this `= {}` at the end of the options parameter. It looks like what that's saying is that if no argument is passed in for options then we default to an empty object? I don't see that being used anywhere and it also just seems like a strange thing to do. Hopefully I am not missing anything there though

ironfish/src/network/peers/peer.ts
```diff
@@ -416,6 +422,28 @@ export class Peer {
...
+    }
+  }
+
+  private disposeMessageMeter() {
+    if (this.state.identity && this.metrics) {
+      this.metrics.p2p_OutboundMessagesByPeer.get(this.state.identity)?.stop()
+      this.metrics.p2p_OutboundMessagesByPeer.delete(this.state.identity)
+    }
+  }
+
```
NullSoldier, line 446: Should we move these below the public methods like other classes?

ironfish/src/node.ts
```diff
@@ -93,13 +91,12 @@ export class IronfishNode {
...
     this.strategy = strategy
     this.metrics = metrics
-    this.miningManager = new MiningManager({ chain, memPool, node: this, telemetry })
+    this.miningManager = new MiningManager({ chain, memPool, node: this })
     this.memPool = memPool
     this.workerPool = workerPool
     this.rpc = new RpcServer(this)
     this.logger = logger
     this.pkg = pkg
-    this.telemetry = telemetry
```
NullSoldier, line 102: Is there a reason telemetry is being removed here? It's a core system, and all core systems are accessible from the node to our node and SDK consumers.

ironfish/src/mining/manager.ts
```diff
@@ -194,7 +186,6 @@ export class MiningManager {
 
     this.blocksMined++
     this.onNewBlock.emit(block)
-    this.telemetry.submitBlockMined(block)
```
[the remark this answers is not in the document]
danield9tqh: I did this in this PR because we need to be able to access `peerNetwork.localPeer.publicIdentity` in telemetry. So I had to instantiate telemetry after the peerNetwork which meant after the miningManager (since peerNetwork is dependent on mining manager). I think this is probably better as well so we don't have miningManager dependent on telemetry

ironfish/src/node.ts
```diff
@@ -124,11 +121,29 @@ export class IronfishNode {
       logger: logger,
     })
 
+    this.telemetry = new Telemetry({
+      chain,
+      logger,
+      config,
+      metrics,
+      workerPool,
+      localPeerIdentity: this.peerNetwork.localPeer.publicIdentity,
```
NullSoldier, line 130: Do you not want to pass the peer network in here like the other systems?

danield9tqh: Would it be possible in the future for `peerNetwork.localPeer.publicIdentity` to change? Maybe if a new identity is generated? If so yea maybe passing the whole peerNetwork would be better. I just didn't pass it to limit dependencies

NullSoldier (approved):

ironfish/src/node.ts
NullSoldier: To me were copying some state from the concept it belongs to. I don't know why a telemetry has a localPeerIdentity, but if it was getting the state from the authority when it wanted to submit it, I would understand that instinctivly.

ironfish/src/mining/manager.ts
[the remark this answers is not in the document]
NullSoldier: That sounds reasonable.

PR #1729, opened 2022-07-05
FEAT: more miner status info

hairtail:
## Summary
Add miner name for each mining server/PC to distinguish different physical machines. For example, if I run 2 mining server on official pool and one of them is down, for now I have to check the mining log for each server. With this PR, I can realize which one is down immediately with miners:status command.
## Testing Plan
Local test.
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/miners/pools/status.ts
+++ b/ironfish-cli/src/commands/miners/pools/status.ts
@@ -60,6 +60,7 @@ export class PoolStatus extends IronfishCommand {
     const stratum = new StratumClient({
       host: host,
       port: port,
+      clientName: 'default',
       logger: createRootLogger(),
     })
 
@@ -109,7 +110,8 @@ export class PoolStatus extends IronfishCommand {
 
     if (status.addressStatus) {
       result += `\nMining status for address '${status.addressStatus.publicAddress}':\n`
-      result += `Miners:                ${status.addressStatus.miners}\n`
+      result += `Number of miners:      ${status.addressStatus.miners}\n`
+      result += `Active miners:         ${status.addressStatus.activeMiners}\n`
       result += `Hashrate:              ${FileUtils.formatHashRate(
         status.addressStatus.hashRate,
       )}\n`
--- a/ironfish-cli/src/commands/miners/pools/status.ts
+++ b/ironfish-cli/src/commands/miners/pools/status.ts
@@ -60,7 +60,6 @@ export class PoolStatus extends IronfishCommand {
     const stratum = new StratumClient({
       host: host,
       port: port,
-      clientName: 'default',
       logger: createRootLogger(),
     })
 
@@ -111,7 +110,7 @@ export class PoolStatus extends IronfishCommand {
     if (status.addressStatus) {
       result += `\nMining status for address '${status.addressStatus.publicAddress}':\n`
       result += `Number of miners:      ${status.addressStatus.miners}\n`
-      result += `Active miners:         ${status.addressStatus.activeMiners}\n`
+      result += `Connected miners:      ${status.addressStatus.connectedMiners.join(', ')}\n`
       result += `Hashrate:              ${FileUtils.formatHashRate(
         status.addressStatus.hashRate,
       )}\n`
--- a/ironfish-cli/src/commands/miners/start.ts
+++ b/ironfish-cli/src/commands/miners/start.ts
@@ -35,9 +35,7 @@ export class Miner extends IronfishCommand {
     }),
     name: Flags.string({
       char: 'n',
-      default: 'default',
-      description:
-        'the miner name distinguish different machines such as iron1. Server hostname is recommended',
+      description: 'the miner name distinguishes different miners',
     }),
     address: Flags.string({
       char: 'a',
@@ -75,8 +73,6 @@ export class Miner extends IronfishCommand {
         this.error('The given public address is not valid, please provide a valid one.')
       }
 
-      const minerName = flags.name
-
       let host = this.sdk.config.get('poolHost')
       let port = this.sdk.config.get('poolPort')
 
@@ -93,8 +89,9 @@ export class Miner extends IronfishCommand {
         }
       }
 
+      const nameInfo = flags.name ? ` with name ${flags.name}` : ''
       this.log(
-        `Starting to mine with public address: ${flags.address} at pool ${host}:${port} with name ${minerName}`,
+        `Starting to mine with public address: ${flags.address} at pool ${host}:${port}${nameInfo}`,
       )
 
       const miner = new MiningPoolMiner({
@@ -104,7 +101,7 @@ export class Miner extends IronfishCommand {
         batchSize,
         host: host,
         port: port,
-        minerName,
+        name: flags.name,
       })
 
       miner.start()
--- a/ironfish-cli/src/commands/miners/start.ts
+++ b/ironfish-cli/src/commands/miners/start.ts
@@ -33,6 +33,12 @@ export class Miner extends IronfishCommand {
       char: 'p',
       description: 'the host and port of the mining pool to connect to such as 92.191.17.232',
     }),
+    name: Flags.string({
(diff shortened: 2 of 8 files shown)


ironfish/src/mining/stratum/messages.ts
```diff
@@ -137,6 +140,7 @@ export const MiningStatusSchema: yup.ObjectSchema<MiningStatusMessage> = yup
     addressStatus: yup
       .object({
         publicAddress: yup.string().required(),
+        activeMiners: yup.string().required(),
```
hughy, line 143: could this be an array of strings instead? then it wouldn't be necessary to stringify the list below

hughy (changes requested):
i like the idea of being able to identify which miners are connected. i left a few comments in the code, but have a few more general comments:

- this will only be useful if every miner is given a unique name. could we use `StratumServerClient.remoteAddress` instead of adding a name to each miner? that will be the remote IP address of the connected client
- the connected clients aren't necessarily 'active' -- they might not be submitting shares. it might be better to think of these as 'connected' miners. there's more work to do to determine which clients are actively mining

ironfish-cli/src/commands/miners/start.ts
```diff
@@ -94,6 +104,7 @@ export class Miner extends IronfishCommand {
         batchSize,
         host: host,
         port: port,
+        minerName,
```
NullSoldier, line 107: We don't need to make a new variable when we only want to pass the flag through here

```suggestion
        minerName: flags.name,
```

ironfish-cli/src/commands/miners/start.ts
```diff
@@ -33,6 +33,12 @@ export class Miner extends IronfishCommand {
       char: 'p',
       description: 'the host and port of the mining pool to connect to such as 92.191.17.232',
     }),
+    name: Flags.string({
+      char: 'n',
+      default: 'default',
+      description:
+        'the miner name distinguish different machines such as iron1. Server hostname is recommended',
```
NullSoldier, line 40: we don't need to make a recommendation, let the user do what they want.

```suggestion
        'the miner name distinguishes different miners',
```

ironfish-cli/src/commands/miners/pools/status.ts
```diff
@@ -109,7 +110,8 @@ export class PoolStatus extends IronfishCommand {
 
     if (status.addressStatus) {
       result += `\nMining status for address '${status.addressStatus.publicAddress}':\n`
-      result += `Miners:                ${status.addressStatus.miners}\n`
+      result += `Number of miners:      ${status.addressStatus.miners}\n`
+      result += `Active miners:         ${status.addressStatus.activeMiners}\n`
```
NullSoldier, line 114: these are not necessarily active miners, just connected miners. I would expect an active miner to be someone with submitted shares. As @hughy said below, I assume you want this to be miners tied to your public key.

ironfish/src/mining/poolMiner.ts
```diff
@@ -35,6 +35,7 @@ export class MiningPoolMiner {
     publicAddress: string
     host: string
     port: number
+    minerName: string
```
NullSoldier, line 38: If the variable belongs to a thing, you don't need to prefix it with the name of that thing. For example, `miner.minerName` is redundant.

ironfish/src/mining/stratum/stratumClient.ts
```diff
@@ -33,6 +33,7 @@ export class StratumClient {
   readonly socket: net.Socket
   readonly host: string
   readonly port: number
+  readonly clientName: string
```
NullSoldier, line 36: The client prefix is redundent here as `client.clientName`

ironfish/src/mining/stratum/messages.ts
NullSoldier: Yeah, you can do that in YUP like this...
```suggestion
        activeMiners: yup.array(yup.string().required()).required(),
```

hairtail:
> i like the idea of being able to identify which miners are connected. i left a few comments in the code, but have a few more general comments:
> 
> * this will only be useful if every miner is given a unique name. could we use `StratumServerClient.remoteAddress` instead of adding a name to each miner? that will be the remote IP address of the connected client
> * the connected clients aren't necessarily 'active' -- they might not be submitting shares. it might be better to think of these as 'connected' miners. there's more work to do to determine which clients are actively mining

Thanks for your time. I will check all your comments and refactor this PR.
> * It's hard to distinguish different miners with remoteAddress(IP). Most users run some (>1) miners within one external IP address. So it's necessary for users to specify unique name for different miners like most eth mining pools do.
> * Yeah. There's some work to do and I have mentioned it (like data struct refactor or so).

ironfish-cli/src/commands/miners/start.ts
```diff
@@ -69,6 +73,10 @@ export class Miner extends IronfishCommand {
         this.error('The given public address is not valid, please provide a valid one.')
       }
 
+      if (flags.name == null) {
+        this.error('Specify your miner name distinguishes different miners.')
+      }
```
NullSoldier, line 78: This should not be required. Let's delete this.

ironfish/src/mining/poolMiner.ts
```diff
@@ -51,7 +52,7 @@ export class MiningPoolMiner {
       port: options.port,
       logger: options.logger,
     })
-    this.stratum.onConnected.on(() => this.stratum.subscribe(this.publicAddress))
+    this.stratum.onConnected.on(() => this.stratum.subscribe(this.publicAddress, options.name))
```
NullSoldier, line 55: nit: since we save the `publicAddress` on the miner can we save the name on here too for consistency? So you can use `miner.name` as well?

ironfish/src/mining/stratum/messages.ts
```diff
@@ -20,6 +20,7 @@ export type MiningDisconnectMessage =
 
 export type MiningSubscribeMessage = {
   version?: number
+  name: string
```
NullSoldier, line 23: This should not be required or will require a stratum bump. Miner names should be an opt in feature.

```suggestion
  name?: string
```

ironfish/src/mining/stratum/messages.ts
```diff
@@ -109,6 +111,7 @@ export const MiningWaitForWorkSchema: yup.MixedSchema<MiningWaitForWorkMessage>
 export const MiningSubscribeSchema: yup.ObjectSchema<MiningSubscribeMessage> = yup
   .object({
     version: yup.number().optional(),
+    name: yup.string().required(),
```
NullSoldier, line 114: should not be required

ironfish-cli/src/commands/miners/pools/status.ts
```diff
@@ -109,7 +109,10 @@ export class PoolStatus extends IronfishCommand {
 
     if (status.addressStatus) {
       result += `\nMining status for address '${status.addressStatus.publicAddress}':\n`
-      result += `Miners:                ${status.addressStatus.miners}\n`
+      result += `Number of miners:      ${status.addressStatus.miners}\n`
+      result += `Connected miners:      ${JSON.stringify(
+        status.addressStatus.connectedMiners,
+      )}\n`
```
NullSoldier, line 115: You shouldn't serialize this back to JSON, just use this to format this array of strings. status.addressStatus.connectedMiners.join(', ')

NullSoldier:
This is pretty close, thanks for putting in the work to address the comments!

hairtail:
> This is pretty close, just one more comment!
> 
> Thanks for putting in the work to address the comments!
> 
> I can't wait to get this merged in.

Updated. There is still one message struct should be refactored. Need you help, [Here](https://github.com/iron-fish/ironfish/pull/1729/files#:~:text=array(yup.string().required()).required()%2C).

NullSoldier (approved):

hughy (approved):

PR #1730, opened 2022-07-05
Load latest hash and track which accounts are synced

mat-if:
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/account/account.ts
+++ b/ironfish/src/account/account.ts
@@ -16,7 +16,6 @@ export class Account {
   readonly outgoingViewKey: string
   publicAddress: string
   rescan: number | null
-  upToDate: boolean
 
   constructor(id: string, serializedAccount: AccountsValue) {
     this.id = id
@@ -26,7 +25,6 @@ export class Account {
     this.outgoingViewKey = serializedAccount.outgoingViewKey
     this.publicAddress = serializedAccount.publicAddress
     this.rescan = serializedAccount.rescan
-    this.upToDate = true
 
     const prefixHash = new MurmurHash3(this.spendingKey, 1)
       .hash(this.incomingViewKey)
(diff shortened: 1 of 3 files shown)


ironfish/src/account/accounts.ts
```diff
@@ -1297,6 +1303,75 @@ export class Accounts {
...
+    for (const account of this.accounts.values()) {
+      const headHash = this.headHashes.get(account.id)
+
+      if (!headHash) {
+        continue
+      }
+
+      const header = await this.chain.getHeader(Buffer.from(headHash, 'hex'))
+
+      if (!header) {
```
rohanjadvani, line 1317: I think we should assert this is defined - that should be an invariant

ironfish/src/account/account.ts
```diff
@@ -16,6 +16,7 @@ export class Account {
   readonly outgoingViewKey: string
   publicAddress: string
   rescan: number | null
+  upToDate: boolean
```
rohanjadvani, line 19: I feel weird about this value living in an account. An account on its own wouldn't know if its up to date. What do you think about having a check in `accounts.ts`, something like ` function isUpToDate(account: Account): boolean`?

ironfish/src/account/accounts.ts
```diff
@@ -1297,6 +1303,75 @@ export class Accounts {
...
+  }
+
+  updateAccountSyncStatus(): void {
+    const latestHash = this.chainProcessor.hash
+
+    for (const account of this.accounts.values()) {
+      const name = account.displayName
+      // If there is no loaded head hash, we assume all accounts should be updated
+      // as the chain is processed
+      if (!latestHash) {
```
rohanjadvani, line 1340: If I have one account up to date with a chain processor, import a new account (without the chain processor syncing more blocks, so the account doesn't have any notes stored), and run this method, will the `upToDate` be true?

PR #1731, opened 2022-07-05
Add snapshot import functionality

deekerno:
## Summary
This PR adds snapshot import functionality to the node. It is the other half of the major work needed for the snapshot project; the first half can be found in #1655.

## Testing Plan
Added new test to RPC route; manual testing shows that the command is working as intended.

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/chain/importChain.ts
```diff
@@ -0,0 +1,75 @@
...
+      request.end({ headSeq: node.chain.head.sequence })
+    }
+
+    const blocks = request.data?.blocks
+    if (blocks) {
+      const reader = bufio.read(blocks, true)
+      const deserializedBlocks = deserializeChunk(node, reader)
+
+      for (const block of deserializedBlocks) {
+        await node.chain.addBlock(block)
```
dguenther, line 55: We should handle the return value of `addBlock` -- if it returns `added: false` with an error, we should log that out. It'd be nice to propagate the error back through to the CLI command, but I think that could be a follow-up.

dguenther (changes requested):

ironfish-cli/src/commands/chain/import.ts
```diff
@@ -0,0 +1,161 @@
...
+          `This snapshot (${
+            manifest.file_name
+          }) contains the Iron Fish blockchain up to block ${
+            manifest.block_height
+          }. The size of the latest snapshot file is ${FileUtils.formatFileSize(
+            manifest.file_size,
+          )}`,
+        )
+
+        const confirm = await CliUx.ux.confirm('Do you wish to continue (Y/N)?')
```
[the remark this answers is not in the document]
deekerno: Added.

ironfish-cli/src/commands/chain/import.ts
```diff
@@ -0,0 +1,161 @@
...
+    files.sort((a, b) => Number(a) - Number(b))
+
+    const client = await this.sdk.connectRpc()
+    const status = await client.getChainInfo()
+    let headSeq = Number(status.content.currentBlockIdentifier.index)
+
+    for (const file of files) {
+      if (headSeq > Number(file)) {
+        continue
+      }
```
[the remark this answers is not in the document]
deekerno: Adjusted.

ironfish/src/rpc/routes/chain/importChain.ts
deekerno: Added. I chose to log through the `logger` object on `node` in line with other RPC routes.

ironfish-cli/src/commands/chain/import.ts
```diff
@@ -0,0 +1,161 @@
...
+    const { flags } = await this.parse(ImportSnapshot)
+
+    let snapshotPath
+    const tempDir = path.join(os.tmpdir(), uuid())
+    await fsAsync.mkdir(tempDir, { recursive: true })
+
+    if (flags.path) {
+      snapshotPath = flags.path
+    } else {
+      const bucket = (flags.bucket || DEFAULT_SNAPSHOT_BUCKET || '').trim()
```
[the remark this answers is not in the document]
deekerno: Sure, I think a good change would be to change the export operation to use the config as well; service providers could always change it for their own bucket. What do you think?

ironfish/src/rpc/routes/chain/importChain.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+  },
+)
+
+function deserializeChunk(node: IronfishNode, reader: bufio.BufferReader): Block[] {
+  const blocks: Block[] = []
+  const blocksLength = reader.readU64()
+  for (let i = 0; i < blocksLength; i++) {
+    const blockBuffer = reader.readVarBytes()
+    const blockReader = bufio.read(blockBuffer, true)
```
dguenther, line 76: Is it necessary to read this into a separate Buffer first? I think you might just be able to call `readBlock(reader)`

ironfish-cli/src/commands/chain/import.ts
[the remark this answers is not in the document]
dguenther: Do you mean you'd convert the export operation to use the same config setting as import, and get rid of the environment variable? That sounds good to me 👍

ironfish/src/rpc/routes/chain/importChain.ts
deekerno: In the export step, each of the blocks is written as a var byte string, so a reader is instantiated with the contents of that variable length string. I did clean it up a bit, however.

dguenther: Ah, I should have code reviewed that better in the export PR. This is fine

ironfish-cli/src/commands/service/snapshot.ts
```diff
@@ -73,7 +74,7 @@ export default class CreateSnapshot extends IronfishCommand {
   async start(): Promise<void> {
     const { flags } = await this.parse(CreateSnapshot)
 
-    const bucket = (flags.bucket || process.env.IRONFISH_SNAPSHOT_BUCKET || '').trim()
+    const bucketUrl = (flags.bucketUrl || DEFAULT_SNAPSHOT_BUCKET_URL || '').trim()
```
dguenther, line 77: Does this make it impossible to not upload to a bucket? It looks like the default would get filled in if no bucket flag is added.

ironfish-cli/src/commands/chain/import.ts
```diff
@@ -0,0 +1,161 @@
...
+    let snapshotPath
+    const tempDir = path.join(os.tmpdir(), uuid())
+    await fsAsync.mkdir(tempDir, { recursive: true })
+
+    if (flags.path) {
+      snapshotPath = flags.path
+    } else {
+      const bucket = (flags.bucket || DEFAULT_SNAPSHOT_BUCKET_URL || '').trim()
+      if (!bucket) {
+        this.log(`Cannot download snapshot without bucket URL`)
```
dguenther, line 54: Should this exit here? I'm not sure what'll happen if it continues with an empty-string bucket

ironfish-cli/src/commands/chain/import.ts
```diff
@@ -0,0 +1,161 @@
...
+      const status = await client.getChainInfo()
+
+      const manifest = await axios
+        .get<{
+          checksum: string
+          file_name: string
+          file_size: number
+          timestamp: number
+          block_height: number
+        }>(`https://${bucket}/manifest.json`)
```
dguenther, line 67: It'd be nice to extract the type here to another file, and share it with the `snapshot` command, so it at least avoids any potential consistency issues between creating the manifest file and parsing it.

ironfish-cli/src/commands/chain/import.ts
```diff
@@ -0,0 +1,161 @@
...
+        barIncompleteChar: '\u2591',
+        format: 'Downloading snapshot: [{bar}] {value}% | ETA: {eta}s',
+      }) as ProgressBar
+
+      bar.start()
+
+      const hasher = crypto.createHash('sha256')
+      const writer = snapshotFile.createWriteStream()
+
+      await axios({
```
dguenther, line 105: This is using both `.then` and `await` -- I think it could be rewritten to use only await, like `const response = await axios({`

ironfish-cli/src/commands/chain/import.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    const client = await this.sdk.connectRpc()
+
+    for (const file of files) {
+      const blocks = await fsAsync.readFile(path.join(blockExportPath, file))
+      await client.importSnapshot({ blocks })
+    }
+  }
+
+  async unzip(source: string, dest: string, excludes: string[] = []): Promise<void> {
```
dguenther, line 146: It looks like the excludes param isn't used, can we get rid of it?

ironfish/src/rpc/routes/chain/importChain.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export const ImportSnapshotRequestSchema: yup.ObjectSchema<ImportSnapshotRequest> = yup
+  .object({
+    chunkStart: yup.number().nullable().optional(),
+    chunkEnd: yup.number().nullable().optional(),
+    blocks: yup.mixed<Buffer>().nullable().optional(),
+  })
+  .optional()
+
+export const ImportSnapshotSchema: yup.ObjectSchema<ImportSnapshotResponse> = yup
```
dguenther, line 32: Is this one used somewhere?

dguenther: Oooh, I see, the other routes do this too. I think the name should just be `ImportSnapshotResponseSchema`

ironfish-cli/src/commands/chain/import.ts
deekerno: I've added an exit; the import command should only provide two pathways: one where the snapshot is already downloaded and takes a path, and the other should download the snapshot if a path isn't given.

ironfish-cli/src/commands/service/snapshot.ts
deekerno: I've added a boolean `upload` flag to ensure that uploading isn't done unintentionally.

dguenther (approved):

PR #1734, opened 2022-07-05
Log transaction propagation to telemetry

danield9tqh:
## Summary
Outlined in the [transaction gossip proposal](https://coda.io/d/Gossip_d7HSvA-Pxcz/Transaction-Gossip-Proposal_suP5z#_luhjb). We want to be able to measure transaction propagation speed similar to how we do with block propagation. We should add this to telemetry

## Testing Plan
Local testing

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/peerNetwork.ts
+++ b/ironfish/src/network/peerNetwork.ts
@@ -15,6 +15,7 @@ import { MetricsMonitor } from '../metrics'
 import { IronfishNode } from '../node'
 import { IronfishPKG } from '../package'
 import { Platform } from '../platform'
+import { Transaction } from '../primitives'
 import { SerializedBlock } from '../primitives/block'
 import { BlockHeader } from '../primitives/blockheader'
 import { Strategy } from '../strategy'
@@ -81,6 +82,7 @@ export class PeerNetwork {
   readonly localPeer: LocalPeer
   readonly peerManager: PeerManager
   readonly onIsReadyChanged = new Event<[boolean]>()
+  readonly onTransactionAccepted = new Event<[transaction: Transaction, received: Date]>()
 
   private started = false
   private readonly minPeers: number
@@ -726,6 +728,8 @@ export class PeerNetwork {
   private async onNewTransaction(
     message: IncomingPeerMessage<NewTransactionMessage>,
   ): Promise<boolean> {
+    const received = new Date()
+
     if (!this.enableSyncing) {
       return false
     }
@@ -771,6 +775,7 @@ export class PeerNetwork {
     }
 
     if (await this.node.memPool.acceptTransaction(transaction, false)) {
+      this.onTransactionAccepted.emit(transaction, received)
       return true
     }
 
--- a/ironfish/src/node.ts
+++ b/ironfish/src/node.ts
@@ -139,6 +139,10 @@ export class IronfishNode {
       this.telemetry.submitBlockMined(block)
     })
 
+    this.peerNetwork.onTransactionAccepted.on((transaction, received) => {
+      this.telemetry.submitNewTransactionSeen(transaction, received)
+    })
+
     this.syncer = new Syncer({
       chain,
       metrics,
--- a/ironfish/src/telemetry/telemetry.ts
+++ b/ironfish/src/telemetry/telemetry.ts
@@ -8,7 +8,9 @@ import { createRootLogger, Logger } from '../logger'
 import { MetricsMonitor } from '../metrics'
 import { Identity } from '../network'
 import { NetworkMessageType } from '../network/types'
+import { Transaction } from '../primitives'
 import { Block } from '../primitives/block'
+import { TransactionHash } from '../primitives/transaction'
 import { GraffitiUtils, renderError, SetIntervalToken } from '../utils'
 import { WorkerPool } from '../workerPool'
 import { Field } from './interfaces/field'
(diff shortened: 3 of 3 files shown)


ironfish/src/network/peerNetwork.ts
```diff
@@ -677,6 +679,8 @@ export class PeerNetwork {
   private async onNewTransaction(
     message: IncomingPeerMessage<NewTransactionMessage>,
   ): Promise<boolean> {
+    const received = new Date(Date.now())
```
NullSoldier, line 682: A new date defaults to the current time

```suggestion
    const received = new Date()
```

ironfish/src/network/peerNetwork.ts
```diff
@@ -710,6 +714,8 @@ export class PeerNetwork {
       return false
     }
 
+    this.onTransactionReceived.emit(transaction, received)
```
NullSoldier, line 717: What would happen if someone attached a slow event handler? Should we have this after the mem pool processes it?

ironfish/src/telemetry/telemetry.ts
```diff
@@ -328,4 +329,39 @@ export class Telemetry {
       ],
     })
   }
+
+  submitNewTransactionSeen(transaction: Transaction, seenAt: Date): void {
+    if (this.shouldLogTransaction(transaction)) {
```
NullSoldier, line 334: we could get rid of this indentation if we just early return like the other telemetry functions

ironfish/src/network/peerNetwork.ts
danield9tqh: I thought handlers execute async? Or does it have some affect on the timing?

ironfish/src/telemetry/telemetry.ts
danield9tqh: yea I'll do this

ironfish/src/telemetry/telemetry.ts
```diff
@@ -328,4 +329,39 @@ export class Telemetry {
...
+      })
+    }
+  }
+
+  /*
+   * We don't want to log all transaction propagation because there are too many
+   * In this way we can only log propagation for a percentage of transactions
+   */
+  private shouldLogTransaction(t: Transaction) {
+    return t.hash().readDoubleBE() % 10000 === 0
```
NullSoldier, line 365: Is this 0.01% of transactions? 3 per block, 180 per hour?

ironfish/src/network/peerNetwork.ts
NullSoldier: Javascript is single threaded, so what ever happens in that handler will take over the event loop until all handlers are finished.

ironfish/src/node.ts
```diff
@@ -139,6 +139,10 @@ export class IronfishNode {
       this.telemetry.submitBlockMined(block)
     })
 
+    this.peerNetwork.onTransactionReceived.on((transaction, date) => {
+      this.telemetry.submitNewTransactionSeen(transaction, date)
```
NullSoldier, line 143: this is a small nit but in the call site you call this `received` which makes it clear what this is, but here it's a date. It's more clear at the call site that this is the received at date. I like you original name more.

ironfish/src/telemetry/telemetry.ts
danield9tqh: With our given throughput (~200 / block) I think this would be about ~80 per hour. Then I'm sure there are some transactions that just never make it to blocks too so maybe slightly more than that

danield9tqh: That's 15k per 7 days (our influx retention period) so this seems reasonable given our current cardinality of 117k

ironfish/src/node.ts
danield9tqh: agreed

ironfish/src/node.ts
```diff
@@ -139,6 +139,10 @@ export class IronfishNode {
       this.telemetry.submitBlockMined(block)
     })
 
+    this.peerNetwork.onTransactionSeen.on((transaction, recevied) => {
+      this.telemetry.submitNewTransactionSeen(transaction, recevied)
```
dguenther, line 143:
```suggestion
    this.peerNetwork.onTransactionSeen.on((transaction, received) => {
      this.telemetry.submitNewTransactionSeen(transaction, received)
```

ironfish/src/network/peerNetwork.ts
```diff
@@ -725,6 +729,8 @@ export class PeerNetwork {
       return true
     }
 
+    this.onTransactionSeen.emit(transaction, received)
```
dguenther, line 732: It looks like now this will only emit if we don't early-return in this function (so if the mempool accepts the transaction, this won't fire). Should this event/telemetry only be emitting when a transaction is accepted by the mempool? I'm not sure what your intent is for when to log telemetry here.

dguenther (changes requested):

ironfish/src/network/peerNetwork.ts
```diff
@@ -78,6 +79,7 @@ export class PeerNetwork {
   readonly localPeer: LocalPeer
   readonly peerManager: PeerManager
   readonly onIsReadyChanged = new Event<[boolean]>()
+  readonly onTransactionSeen = new Event<[Transaction, Date]>()
```
NullSoldier, line 82: You can name these if you want, in the case that it's not clear from the event name what the parameter is.

```suggestion
  readonly onTransactionSeen = new Event<[Transaction, received: Date]>()
```

ironfish/src/network/peerNetwork.ts
NullSoldier, line 732: This is creating a bug, to move this here, you need to refactor the above logic.

ironfish/src/network/peerNetwork.ts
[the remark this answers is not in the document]
danield9tqh: ok changed it back

ironfish/src/node.ts
danield9tqh: I need a spellchecker

NullSoldier (approved):

PR #1739, opened 2022-07-06
Allow nullable head hash entries

mat-if:
## Summary

While working on tracking which accounts were updated, I realized having this map as a sort-of inbetween state of having some accounts, but not unscanned/imported accounts was causing much more complicated logic than just allowing nullable values in the map.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/account/accounts.ts
+++ b/ironfish/src/account/accounts.ts
@@ -50,7 +50,7 @@ export class Accounts {
   >()
   protected readonly decryptedNotes = new Map<string, Readonly<DecryptedNotesValue>>()
   protected readonly nullifierToNote = new Map<string, string>()
-  protected readonly headHashes = new Map<string, string>()
+  protected readonly headHashes = new Map<string, string | null>()
 
   protected readonly accounts = new Map<string, Account>()
   readonly db: AccountsDB
@@ -108,7 +108,7 @@ export class Accounts {
         })
       }
 
-      await this.updateHeadHash(header.hash)
+      await this.updateHeadHashes(header.hash)
     })
 
     this.chainProcessor.onRemove.on(async (header) => {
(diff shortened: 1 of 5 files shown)


ironfish/src/account/accounts.ts
```diff
@@ -324,23 +324,27 @@ export class Accounts {
...
-        await this.db.removeHeadHash(account)
-      }
+      await this.updateHeadHash(account, headHash)
     }
   }
 
+  async updateHeadHash(account: Account, headHash: Buffer | null): Promise<void> {
+    const hash = headHash ? headHash.toString('hex') : null
+
+    this.headHashes.set(account.id, hash)
```
mat-if, line 336: Funny enough, we were never updating the `Accounts` `this.headHash(es)`. Never needed it, I guess, but now that we're adding more logic around this, it's going to come in handy to keep up to date.

ironfish/src/account/accountsdb.ts
```diff
@@ -174,17 +175,7 @@ export class AccountsDB {
     await this.headHashes.clear()
   }
 
-  async replaceHeadHashes(map: Map<string, string>): Promise<void> {
```
mat-if, line 177: Wasn't sure we'd ever use this, so removing. Can re-add if we find a need for it.

ironfish/src/storage/database/encoding.ts
```diff
@@ -50,6 +50,26 @@ export class StringHashEncoding implements IDatabaseEncoding<string> {
   }
 }
 
+export class NullableStringEncoding implements IDatabaseEncoding<string | null> {
```
mat-if, line 53: This is pretty much a copy-paste of the `MetaValueEncoding`. Whether we want to hard-code the meta db values to use this new encoding, or allowing it to have the freedom to be extended as needed can be a discussion for a later time, I guess, unless anyone feels strongly enough about it.

rohanjadvani (approved):

PR #1746, opened 2022-07-07
refactor(ironfish): Encapsulate decrypted notes inside account

rohanjadvani:
## Summary

Move decrypted note into account for encapsulation

## Testing Plan

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/start.test.ts
+++ b/ironfish-cli/src/commands/start.test.ts
@@ -88,19 +88,15 @@ describe('start command', () => {
     const accounts = {
       accountExists: jest.fn(),
       getDefaultAccount: jest.fn(),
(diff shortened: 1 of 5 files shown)


mat-if (approved):
This PR made me realize that we need to enforce some uniqueness on accounts so we can stop looping on some of these once we've found the account associated with the spend, etc to minimize the amount of work we do

rohanjadvani:
> This PR made me realize that we need to enforce some uniqueness on accounts so we can stop looping on some of these once we've found the account associated with the spend, etc to minimize the amount of work we do

Yep for sure. That'll be coming once I migrate each account data structure from the top level wallet.

PR #1747, opened 2022-07-07
Create network message for querying for transactions based on hash

danield9tqh:
## Summary
Outlined in the [transaction gossip proposal](https://coda.io/d/Gossip_d7HSvA-Pxcz/Transaction-Gossip_sulxZ#_luNdn). We should create a new network message that allows nodes to query for transactions by hash. This would be in response to receiving a new gossiped hash from a node.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/messages/pooledTransactionsRequest.test.ts
+++ b/ironfish/src/network/messages/pooledTransactionsRequest.test.ts
@@ -1,0 +1,19 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { blake3 } from '@napi-rs/blake-hash'
+import { v4 as uuid } from 'uuid'
+import { PooledTrasactionsRequest } from './pooledTransactionsRequest'
+
+describe('PooledTrasactionsRequestMessage', () => {
+  it('serializes the object into a buffer and deserializes to the original object', () => {
+    const rpcId = 53242
+    const hashes = [...Array(10)].map((_) => blake3(uuid()))
+
+    const message = new PooledTrasactionsRequest(hashes, rpcId)
+
+    const buffer = message.serialize()
+    const deserializedMessage = PooledTrasactionsRequest.deserialize(buffer, rpcId)
+    expect(deserializedMessage).toEqual(message)
+  })
+})
--- a/ironfish/src/network/messages/pooledTransactionsRequest.test.ts
+++ b/ironfish/src/network/messages/pooledTransactionsRequest.test.ts
@@ -3,17 +3,17 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { blake3 } from '@napi-rs/blake-hash'
 import { v4 as uuid } from 'uuid'
-import { PooledTrasactionsRequest } from './pooledTransactionsRequest'
+import { PooledTransactionsRequest } from './pooledTransactionsRequest'
 
-describe('PooledTrasactionsRequestMessage', () => {
+describe('PooledTransactionsRequest', () => {
   it('serializes the object into a buffer and deserializes to the original object', () => {
     const rpcId = 53242
     const hashes = [...Array(10)].map((_) => blake3(uuid()))
 
-    const message = new PooledTrasactionsRequest(hashes, rpcId)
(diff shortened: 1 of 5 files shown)


ironfish/src/network/messages/pooledTransactionsRequest.ts
```diff
@@ -0,0 +1,51 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import bufio from 'bufio'
+import { TransactionHash } from '../../primitives/transaction'
+import { NetworkMessageType } from '../types'
+import { Direction, RpcNetworkMessage } from './rpcNetworkMessage'
+
+export class PooledTrasactionsRequest extends RpcNetworkMessage {
```
dguenther, line 9: Should be "transactions", there are a couple other places that will need to be refactored to make this work

```suggestion
export class PooledTransactionsRequest extends RpcNetworkMessage {
```

ironfish/src/network/messages/pooledTransactionsRequest.test.ts
```diff
@@ -0,0 +1,19 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { blake3 } from '@napi-rs/blake-hash'
+import { v4 as uuid } from 'uuid'
+import { PooledTrasactionsRequest } from './pooledTransactionsRequest'
+
+describe('PooledTrasactionsRequestMessage', () => {
+  it('serializes the object into a buffer and deserializes to the original object', () => {
+    const rpcId = 53242
+    const hashes = [...Array(10)].map((_) => blake3(uuid()))
```
dguenther, line 11: I'm guessing this test runs in like 100ms or so? If it's slower than that, we could probably just skip the hashing and create a 32-byte buffer directly, otherwise I think it's fine.

dguenther (changes requested):

ironfish/src/network/messages/pooledTransactionsRequest.test.ts
danield9tqh: Yea I'll create a 32 byte buffer

ironfish/src/network/messages/pooledTransactionsRequest.ts
danield9tqh: good catch

ironfish/src/network/messages/pooledTransactionsRequest.test.ts
danield9tqh: Did some testing on my computer, with a list of size 10k it only took 15ms to do all the uuid and hashes. Seems to be pretty fast

ironfish/src/network/types.ts
```diff
@@ -18,7 +18,11 @@ export enum NetworkMessageType {
   PeerListRequest = 10,
   Signal = 11,
   SignalRequest = 12,
-  NewBlockHashes = 13,
+  PooledTransactionsRequest = 13,
+  PooledTransactionsResponse = 14,
+  NewPooledTransactionHashes = 15,
+  NewTransactionV2 = 16,
+  NewBlockHashes = 17,
```
dguenther, line 25: You'll probably have to resolve this with the messages I just merged to staging -- I don't care who gets which numbers 😆

ironfish/src/network/messages/pooledTransactionsRequest.ts
```diff
@@ -0,0 +1,51 @@
...
+    for (const hash of this.transactionHashes) {
+      bw.writeHash(hash)
+    }
+
+    return bw.render()
+  }
+
+  static deserialize(buffer: Buffer, rpcId: number): PooledTransactionsRequest {
+    const reader = bufio.read(buffer, true)
+    const transactionHashesLength = reader.readU16()
```
dguenther, line 31: I think we discussed potentially swapping these over to varInts

dguenther (approved):
This looks good to go once the comments are addressed 👍

ironfish/src/network/types.ts
danield9tqh: damn thought I'd get in there first. shouldn't have approved your PR haha

dguenther: haha, that was your fatal mistake 😆

PR #1761, opened 2022-07-11
Update accountsToRemove map in minedBlocksIndexer

mat-if:
## Breaking Change

```
[ ] Yes
```


deekerno (approved):
Pesky filter return.

PR #1762, opened 2022-07-11
fix(ironfish): Add mutex for sending transactions

rohanjadvani:
## Summary

Currently there is no synchronization around creating transactions for an account. It's possible for two calls for creating transaction to grab the same note.

## Testing Plan

Manual testing

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/account/accounts.ts
+++ b/ironfish/src/account/accounts.ts
@@ -11,6 +11,7 @@ import { Config } from '../fileStores'
 import { createRootLogger, Logger } from '../logger'
 import { MemPool } from '../memPool'
 import { NoteWitness } from '../merkletree/witness'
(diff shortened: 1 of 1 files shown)


mat-if (approved):
Seems to work as expected. I'd like to see it so we can early return if it's locked, since that seems more correct, like a "Transaction already in progress" error, but that's kind of out of the scope of this PR since it'd require making some maybe-not-trivial changes to `Mutex`

NullSoldier:
I reviewed with https://github.com/iron-fish/ironfish/pull/1762/files?diff=split&w=1

NullSoldier:
I tested it manually, this works for me.

NullSoldier (approved):

PR #1763, opened 2022-07-11
Properly catch and surface leveldb corruption errors

NullSoldier:
## Summary
We were treating all errors as OpenErrors which are thought to be thrown
because the database is locked. It turns out that LevelDB also throws an
OpenError if the DB is corrupt as well. These were being surfaced
specifically as locked errors which are not correct.

## Testing Plan
*Node already in use test*
1. Run a node
2. Run another node to see database already being used error

*Database corrupt test*
1. Delete a random file in your leveldb directory
2. Run your node

## Breaking Change

```
[ ] Yes
```


ironfish/src/storage/levelup/database.ts
```diff
@@ -64,7 +68,20 @@ export class LevelupDatabase extends Database {
       const opened = levelup(this.db, (error?: unknown) => {
         if (error) {
           if (error instanceof levelErrors.OpenError) {
-            reject(new DatabaseIsLockedError(error.message))
+
+            // Here we coerce leveldb specific errors into ironfish storage
+            // layer errors. We need to do message descrimination because the
+            // leveldb JS wrapper does not provide a way to descriminate on the
```
rohanjadvani, line 74:
```suggestion
            // leveldb JS wrapper does not provide a way to discriminate on the
```

ironfish/src/storage/levelup/database.ts
```diff
@@ -64,7 +68,20 @@ export class LevelupDatabase extends Database {
       const opened = levelup(this.db, (error?: unknown) => {
         if (error) {
           if (error instanceof levelErrors.OpenError) {
-            reject(new DatabaseIsLockedError(error.message))
+
+            // Here we coerce leveldb specific errors into ironfish storage
+            // layer errors. We need to do message descrimination because the
```
rohanjadvani, line 73:
```suggestion
            // layer errors. We need to do message discrimination because the
```

rohanjadvani (approved):

PR #1764, opened 2022-07-11
Simple lint to help point out when forgetting to assign filter, etc

mat-if:
## Summary

It's too easy to forget to assign `.filter` etc, let's add a little help. This is probably a _little_ jank, as this could trigger on methods that are correct in this fashion, say if we add some `filter` fn to a random class.  If needed, we can extend this in the future to check for types via TS, so we only check `Array.filter`, for example.

## Breaking Change

```
[ ] Yes
```


rohanjadvani (approved):

mat-if:
Closing in favor of #1773 - I missed this when I was looking yesterday.

PR #1767, opened 2022-07-11
feat(ironfish): Only sync transactions for accounts that can decrypt notes

rohanjadvani:
## Summary

* Only return mapping of account IDs -> decrypted notes if notes are available to remove unnecessary `syncTransaction` calls
* Check if transactions need to be updated before hitting the DB

## Testing Plan

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/account/account.ts
+++ b/ironfish/src/account/account.ts
@@ -162,14 +162,12 @@ export class Account {
     const blockHash = 'blockHash' in params ? params.blockHash : null
     let submittedSequence = 'submittedSequence' in params ? params.submittedSequence : null
 
-    let shouldUpdateTransaction = true
     const record = this.transactions.get(transactionHash)
     if (record) {
       submittedSequence = record.submittedSequence
-      shouldUpdateTransaction = !this.transactionValuesAreEqual(record, transaction, blockHash)
     }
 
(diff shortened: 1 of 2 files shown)


ironfish/src/account/account.ts
```diff
@@ -162,21 +162,37 @@ export class Account {
     const blockHash = 'blockHash' in params ? params.blockHash : null
     let submittedSequence = 'submittedSequence' in params ? params.submittedSequence : null
 
+    let shouldUpdateTransaction = true
     const record = this.transactions.get(transactionHash)
     if (record) {
       submittedSequence = record.submittedSequence
+      shouldUpdateTransaction = !this.transactionValuesAreEqual(record, transaction, blockHash)
```
danield9tqh, line 169: nit: Seems like we could simplify code a bit by just 
```typescript
if (!record.transaction.equals(transaction) || !(record.blockHash === blockHash)) {
  await this.updateTransaction(
    transactionHash,
    { transaction, blockHash, submittedSequence },
    tx,
  )
}
```

rohanjadvani: Yep sure, I can update it

mat-if (approved):

PR #1768, opened 2022-07-11
scanTransactions updates progress incrementally

mat-if:
## Summary

Still technically updates all accounts in the syncTransaction call, but now updates the outdated accounts so that we can save progress if we stop it for any reason

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/account/accounts.ts
+++ b/ironfish/src/account/accounts.ts
@@ -510,10 +510,8 @@ export class Accounts {
         scan.onTransaction.emit(sequence)
       }
 
-      if (blockHeader.sequence % 20 === 0) {
-        for (const account of outdatedAccounts) {
-          await this.updateHeadHash(account, blockHeader.hash)
-        }
+      for (const account of outdatedAccounts) {
+        await this.updateHeadHash(account, blockHeader.hash)
       }
     }
 
--- a/ironfish/src/account/accounts.ts
+++ b/ironfish/src/account/accounts.ts
@@ -466,41 +466,55 @@ export class Accounts {
     const scan = new ScanState()
     this.scan = scan
 
-    // If were updating the account head we need to wait until its finished
+    // If we are updating the account head, we need to wait until its finished
     // but setting this.scan is our lock so updating the head doesn't run again
     await this.updateHeadState?.wait()
 
-    const accountHeadHash = this.chainProcessor.hash
+    const startHash = await this.getEarliestHeadHash()
+    const endHash = this.chainProcessor.hash
 
-    const scanStart = await this.getEarliestHeadHash()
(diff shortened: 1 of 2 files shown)


ironfish/src/blockchain/blockchain.ts
```diff
@@ -1120,19 +1120,15 @@ export class Blockchain {
   }
 
   /**
-   * Iterates through transactions, starting from fromHash or the genesis block,
+   * Iterates through block headers, starting from fromHash or the genesis block,
    * to toHash or the heaviest head.
    */
-  async *iterateTransactions(
+  async *iterateBlockHeaders(
```
mat-if, line 1126: Changed this to just iterate blockHeaders since this function was only used by scanTransactions, and I wanted to iterate blockHeaders so we could track the progress without lots of weird logic by guessing on the sequence returned by `iterateBlockTransactions`

ironfish/src/account/accounts.ts
```diff
@@ -466,41 +466,55 @@ export class Accounts {
...
+        scan.onTransaction.emit(sequence)
       }
 
-      await this.syncTransaction(transaction, { blockHash, initialNoteIndex: initialNoteIndex })
-      scan.onTransaction.emit(sequence)
+      if (blockHeader.sequence % 20 === 0) {
+        for (const account of outdatedAccounts) {
+          await this.updateHeadHash(account, blockHeader.hash)
+        }
+      }
```
mat-if, line 517: This is the only real change here, besides splitting out the blocks/txns iteration, so we now update the head hashes every 20 blocks

rohanjadvani: What's the thought behind every 20 blocks?

ironfish/src/account/accounts.ts
```diff
@@ -466,41 +466,55 @@ export class Accounts {
...
+        await this.syncTransaction(transaction, {
+          blockHash,
+          initialNoteIndex: initialNoteIndex,
+        })
+        scan.onTransaction.emit(sequence)
       }
 
-      await this.syncTransaction(transaction, { blockHash, initialNoteIndex: initialNoteIndex })
-      scan.onTransaction.emit(sequence)
+      if (blockHeader.sequence % 20 === 0) {
```
NullSoldier, line 513: I think this will cause bugs by not updating in the last 20 blocks, which can have mismatched note state in a fork situation. Maybe we should talk through this.

ironfish/src/account/accounts.ts
mat-if: Initially thought it'd be a bit less overhead, but as Jason pointed out, that's not an ideal approach, so changed it to update every block

rohanjadvani (approved):

PR #1771, opened 2022-07-12
Make blockchain iteration return buffer hashes

NullSoldier:
## Summary
I need this in buffer form for another PR I'm going to open up.  Also, we should stop hexifying hashes around the code base unless were rendering them for serialization or for display.

## Testing Plan
Run tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -1129,7 +1129,13 @@ export class Blockchain {
     tx?: IDatabaseTransaction,
     reachable = true,
   ): AsyncGenerator<
-    { transaction: Transaction; initialNoteIndex: number; sequence: number; blockHash: string },
+    {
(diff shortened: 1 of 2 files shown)


danield9tqh (approved):

ironfish/src/blockchain/blockchain.ts
```diff
@@ -1182,8 +1194,9 @@ export class Blockchain {
       yield {
         transaction,
         initialNoteIndex: noteIndex,
-        blockHash: header.hash.toString('hex'),
+        blockHash: header.hash,
         sequence: header.sequence,
+        previousBlockHash: header.previousBlockHash,
```
dguenther, line 1199: I'm guessing this is related to an upcoming change? I think it's fine if it merges, but doesn't seem directly used or related to this PR currently

PR #1772, opened 2022-07-12
Abort rescan if aborted after waiting for head

NullSoldier:
## Summary

When were rescanning, we start a scan, then wait for the lock on the
updateHead() method. However, this may take a long time. It's possible
the head updating was aborted because the node is shutting down. The
issue is that we don't also check if the scan was aborted as soon as
the updateHead() lock is released.

This means we could start scanning when the node is shutting down which
will prevent it from shutting down properly.

## Testing Plan
1. Have a lot of blocks to scan
2. `yarn start  accounts:rescan --reset --local`
3. `yarn start start`
4. Close node in the middle of the reset
5. Start node, it'll start updating the head right away but it will take a long time
6. `yarn start  accounts:rescan --reset`
    * This should get stuck at waiting for scan to start
7. `ironfish stop`

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/account/accounts.ts
+++ b/ironfish/src/account/accounts.ts
@@ -135,11 +135,12 @@ export class Accounts {
       return
     }
 
-    this.updateHeadState = new ScanState()
+    const scan = new ScanState()
+    this.updateHeadState = scan
 
     try {
       const { hashChanged } = await this.chainProcessor.update({
-        signal: this.updateHeadState.abortController.signal,
+        signal: scan.abortController.signal,
       })
 
       if (hashChanged) {
@@ -148,7 +149,7 @@ export class Accounts {
         )
       }
     } finally {
-      this.updateHeadState.signalComplete()
+      scan.signalComplete()
       this.updateHeadState = null
     }
   }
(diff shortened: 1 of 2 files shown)


ironfish/src/account/accounts.ts
```diff
@@ -135,11 +135,12 @@ export class Accounts {
       return
     }
 
-    this.updateHeadState = new ScanState()
+    const scan = new ScanState()
```
NullSoldier, line 138: We cannot reference this.updateHeadState because it can be cleared by shutdown() and this causes a deadlock.

ironfish/src/account/accounts.ts
```diff
@@ -603,6 +598,12 @@ export class Accounts {
 
     const accountHeadHash = this.chainProcessor.hash
 
+    if (scan.isAborted) {
+      scan.signalComplete()
+      this.scan = null
+      return
+    }
```
NullSoldier, line 605: Once we're done waiting on the updateHead() lock, we need to check if we should abort, instead of start iterating the chain. This could happen during a node shutdown.

ironfish/src/node.ts
```diff
@@ -335,10 +335,12 @@ export class IronfishNode {
       this.rpc.stop(),
       this.telemetry.stop(),
       this.metrics.stop(),
-      this.workerPool.stop(),
       this.minedBlocksIndexer.stop(),
     ])
 
+    // Do after to avoid unhandled error from aborted jobs
+    await Promise.allSettled([this.workerPool.stop()])
```
NullSoldier, line 342: This is a bug because stopping the worker pool aborts jobs, which almost no job actually handles. By doing this before all the services are shut down, it cause the process to crash MOST times because of the unhandled JobAborted error. It was preventing clean shutdown from working correctly.

mat-if (approved):

PR #1776, opened 2022-07-13
Improve accounts:rescan CLI command UX

NullSoldier:
## Summary
`ironfish accounts:rescan --local --reset`
```
Scanning for transactions
Asking node to start scanning... done
Scanning Blocks: [----------------------------------------] 363/75956 0% 40/sec | 19m 16s
```
## Testing Plan
Run the CLI command

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/accounts/rescanAccount.ts
+++ b/ironfish/src/rpc/routes/accounts/rescanAccount.ts
@@ -35,8 +35,7 @@ router.register<typeof RescanAccountRequestSchema, RescanAccountResponse>(
 
     if (!scan) {
       if (node.accounts.updateHeadState) {
-        void node.accounts.updateHeadState.abort()
-        await node.accounts.updateHeadState.wait()
+        await node.accounts.updateHeadState.abort()
       }
 
       if (request.data.reset) {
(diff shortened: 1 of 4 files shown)


hughy (approved):

ironfish/src/rpc/routes/accounts/rescanAccount.ts
```diff
@@ -33,6 +34,11 @@ router.register<typeof RescanAccountRequestSchema, RescanAccountResponse>(
     }
 
     if (!scan) {
+      if (node.accounts.updateHeadState) {
+        void node.accounts.updateHeadState.abort()
+        await node.accounts.updateHeadState.wait()
+      }
```
NullSoldier, line 40: If you are running an update head, this will abort abort it so we can start rescanning. A nice usability improvement.

ironfish/src/rpc/routes/accounts/rescanAccount.ts
```diff
@@ -33,6 +34,11 @@ router.register<typeof RescanAccountRequestSchema, RescanAccountResponse>(
     }
 
     if (!scan) {
+      if (node.accounts.updateHeadState) {
+        void node.accounts.updateHeadState.abort()
+        await node.accounts.updateHeadState.wait()
```
NullSoldier, line 39: this shouldn't be necessary, deleting.

NullSoldier: Fixed

mat-if (approved):
Minor nit on how it looks on complete which is a little funky, not sure if its worth handling, but otherwise looks good
image

NullSoldier:
Fixed

PR #1783, opened 2022-07-13
Only scan transactions for accounts where it is relevant

mat-if:
## Summary

**Builds on #1775**

Scans transactions for accounts that are specifically relevant, so we aren't scanning transactions for accounts that don't need it, and don't "set back" account progress

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/account/accounts.ts
+++ b/ironfish/src/account/accounts.ts
@@ -483,16 +483,28 @@ export class Accounts {
     const startHash = await this.getEarliestHeadHash()
     const endHash = this.chainProcessor.hash
 
-    const outdatedAccounts = this.listAccounts().filter((a) => !this.isAccountUpToDate(a))
+    const accounts: Array<Account> = []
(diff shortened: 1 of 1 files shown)


ironfish/src/account/accounts.ts
```diff
@@ -517,14 +529,31 @@ export class Accounts {
...
 
-      for (const account of outdatedAccounts) {
+      for (const account of accounts) {
         await this.updateHeadHash(account, blockHeader.hash)
       }
+
+      const hashHex = blockHeader.hash.toString('hex')
+
+      let i = 0
+      while (i < remainingAccounts.length) {
```
rohanjadvani, line 544: nit: I'm not a huge fan of mutating objects (specifically deletion) while iterating over them using the index + length. Do you think this can be refactored to a `for` loop? I think it can become more readable

rohanjadvani (approved):

PR #1793, opened 2022-07-15
Add Database Migration System

NullSoldier:
## Summary
This adds a new database migration that forces users to run (potentially) long running migration operations on their node before the node will start. This does have an impact on user experience so I'll document that below.

I've repurposed the database version to now be used by the migration system, and no longer used by the database layer itself. Migrations are only meant to migrate one single database, because we can't guarantee atomic transactions over multiple leveldb instances anyway. Because of that, you should never write a migration that requires atomic writes between multiple databases.

Here is a summary of changes.

 - Added new command `migrations:status` to print out migrations and if they are run or not
 - Added new command `migrations:start` to apply all migrations that have not run
 - Added new command `migrations:revert` to run the backwards command off the last migration (mostly for developing migrations)
 - Added new `--upgrade` flag to `ironfish start` to run transactions before the node starts
 - Added a new config `databaseMigrate` to configure whether we should run migrations on `ironfish start`
 - Removed migration logic from database holders around the codebase
 - Added migrations to bring our 3 databses into the migration system and test the impact of user experience

**TODO**
 - [x] Fix tests to migrate database
 - [x] Fix first run experience so it automatically runs migrations silently on an empty DB
 - [x] Fix `ironfish debug` when the DB is not upgraded

## Testing Plan
This should be tested thoroughly by the team by running migrations on fresh, and existing databases.

## Breaking Change

```
[x] Yes
```


ironfish-cli/src/commands/migrations/revert.ts
```diff
@@ -0,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { IronfishCommand } from '../../command'
+import { ConfigFlag, ConfigFlagKey, DataDirFlag, DataDirFlagKey } from '../../flags'
+
+export class RevertCommand extends IronfishCommand {
+  static description = `Run migrations`
```
deekerno, line 8: Should this be "Revert migrations"?

deekerno (changes requested):

ironfish-cli/src/commands/migrations/revert.ts
NullSoldier: Fixed

ironfish-cli/src/commands/migrations/status.ts
```diff
@@ -0,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { IronfishCommand } from '../../command'
+import { ConfigFlag, ConfigFlagKey, DataDirFlag, DataDirFlagKey } from '../../flags'
+
+export class StatusCommand extends IronfishCommand {
+  static description = `Run migrations`
```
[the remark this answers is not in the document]
NullSoldier: Fixed

ironfish/src/merkletree/merkletree.ts
```diff
@@ -176,10 +160,20 @@ export class MerkleTree<
...
-      throw new Error(`No counts found in tree ${this.name} for type ${countType}`)
+      if (countType === 'Leaves') {
+        count = 0
+      } else if (countType === 'Nodes') {
+        count = 1
+      } else {
+        Assert.isNever(countType)
+      }
+
+      await this.counter.put(countType, count)
```
danield9tqh, line 174: should we pass in transaction here?

ironfish/src/migrations/migration.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { IronfishNode } from '../node'
+import { IDatabase, IDatabaseTransaction } from '../storage'
+
+export abstract class Migration {
+  id = 0
+  abstract name: string
+
+  init(): Migration {
+    this.id = Number(this.name.split('-')[0])
```
danield9tqh, line 13: Could we just have 2 fields for this so we don't have to do a split?

danield9tqh: Or have `abstract id: number` as well? Any reason for not doing that?

ironfish/src/merkletree/merkletree.ts
NullSoldier: good find, fixed.

ironfish/src/migrations/migration.ts
NullSoldier: > Or have abstract id: number as well? Any reason for not doing that?

You only declare abstract when you want to force the child to implement it. In this case it's an ID managed by the migration system.

> Could we just have 2 fields for this so we don't have to do a split?

Can you explain what issue you have withi splitting the string? If there's 2 fields, it's 2 fields you need to update in the migration. This gets the name / id from the field.

The alternative could be getting the ID / name from the class but then it won't match the file. Other migration systems. generate it from the file, but we can't exactly do that here.

danield9tqh: should've added 'nit'. Just thought it would provide better type safety to have 2 fields instead of splitting a string. But I see you're getting it from the filename so that makes sense

ironfish-cli/src/command.ts
```diff
@@ -77,6 +78,9 @@ export abstract class IronfishCommand extends Command {
         this.log(error.codeMessage)
       } else if (error instanceof RpcConnectionError) {
         this.log(`Cannot connect to your node, start your node first.`)
+      } else if (error instanceof DatabaseVersionError) {
+        this.log(error.message)
+        this.exit(1)
```
NullSoldier, line 83: This is special casing these errors for better UX because we don't want the stack.

ironfish/src/migrations/migrator.ts
```diff
@@ -0,0 +1,165 @@
...
+      try {
+        await db.open()
+
+        await db.transaction(async (tx) => {
+          await migration.forward(this.node, db, tx)
+          await db.putVersion(migration.id, tx)
+        })
+      } catch (e) {
+        writeOut?.(` ERROR\n`)
+        this.logger.error(ErrorUtils.renderError(e, true))
```
rohanjadvani, line 126: Should this be:

```suggestion
        logger?.error(ErrorUtils.renderError(e, true))
```

Or will we always log it out because it's an error?

rohanjadvani (approved):

ironfish/src/migrations/migrator.ts
NullSoldier: Always log during a crash.

deekerno (approved):

PR #1795, opened 2022-07-16
Add handler for GetBlockTransactionsRequest

dguenther:
## Summary

Adds a handler for GetBlockTransactionsRequest, which responds with the transactions on a given block corresponding to the differentially-encoded indexes.

Note that this message isn't sent in the block propagation code yet.

## Testing Plan

Added tests for the message handler.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/peerNetwork.ts
+++ b/ironfish/src/network/peerNetwork.ts
@@ -73,6 +73,8 @@ import { WebSocketServer } from './webSocketServer'
 const GOSSIP_FILTER_SIZE = 100000
 const GOSSIP_FILTER_FP_RATE = 0.000001
 
+const MAX_GET_BLOCK_TRANSACTIONS_DEPTH = 10
+
 type RpcRequest = {
   resolve: (value: IncomingPeerMessage<RpcNetworkMessage>) => void
   reject: (e: unknown) => void
@@ -561,7 +563,7 @@ export class PeerNetwork {
         } else if (rpcMessage instanceof PooledTransactionsRequest) {
           responseMessage = this.onPooledTransactionsRequest(rpcMessage, rpcId)
         } else if (rpcMessage instanceof GetBlockTransactionsRequest) {
-          responseMessage = this.onGetBlockTransactionsRequest(rpcMessage)
+          responseMessage = await this.onGetBlockTransactionsRequest(peer, rpcMessage)
         } else {
           throw new Error(`Invalid rpc message type: '${rpcMessage.type}'`)
         }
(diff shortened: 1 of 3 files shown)


ironfish/src/network/peerNetwork.ts
```diff
@@ -718,11 +720,73 @@ export class PeerNetwork {
...
+    }
+
+    const transactions = []
+    let currentIndex = 0
+    for (const transactionIndex of message.transactionIndexes) {
+      if (transactionIndex < 0) {
+        const errorMessage = `Requested negative transaction index`
+        peer.punish(BAN_SCORE.MAX, errorMessage)
+        throw new CannotSatisfyRequestError(errorMessage)
+      }
```
dguenther, line 775: This shouldn't practically ever happen, since bufio's `readVarint` only allows for unsigned integers, but it's not easy to propagate that assumption through the code.

ironfish/src/network/peerNetwork.ts
```diff
@@ -718,11 +720,73 @@ export class PeerNetwork {
...
+
+      currentIndex += transactionIndex
```
danield9tqh, line 777: kind of confused as to what this is doing. Is this like an index offset?

ironfish/src/network/peerNetwork.ts
```diff
@@ -718,11 +720,73 @@ export class PeerNetwork {
...
+        throw new CannotSatisfyRequestError(
+          `Peer requested transactions for block ${message.blockHash.toString(
+            'hex',
+          )} with sequence ${header.sequence} while chain head is at sequence ${
+            this.chain.head.sequence
+          }`,
+        )
+      }
+
+      const block = await this.chain.getBlock(header, tx)
```
danield9tqh, line 748: Any reason we can't just call `getBlock` directly with the hash? Is it because we want to check the sequence first before we query the whole block?

danield9tqh:
Looks good, only thing that worries me a bit is the banning mechanism. I think ultimately we probably do want to move towards a reputation system where we just order the peers by reputation and drop peers with the lowest reputation when we reach peer limits (and never ban anyone). I think it may be better to not ban for now until we implement that system. None of these ban cases should happen if the code works as expected but if we have a bug I think it could be better to keep accepting faulty requests than to start banning peers. wdyt?

ironfish/src/network/peerNetwork.ts
dguenther: It's an offset from the previous transaction in the list, referred to as a differentially-encoded index. It's the difference between the previous index and the current index, minus one. Pretty simple to compute, and increases the likelihood that indexes can be encoded in a single byte

ironfish/src/network/peerNetwork.ts
dguenther: Yep -- The block body is relatively hefty compared to the header, so it's worth doing validation on the header first before fetching the rest of the block. It ends up being the same number of database reads, but split up into separate fetches for the header and body.

It might not have been clear, but calling `getBlock` with the hash fetches both the header and the body, and calling `getBlock` with the header fetches only the body.

ironfish/src/network/peerNetwork.ts
danield9tqh: ok that makes sense. was just a little confusing without context

ironfish/src/network/peerNetwork.ts
danield9tqh: 👍🏻

PR #1803, opened 2022-07-19
Create transaction fetcher and gossip transactions by hash

danield9tqh:
## Summary
When receiving a new transaction hash from another peer the node needs to fetch the full transaction. The fetcher defines logic for fetching the transaction from one or multiple peers as well as resolving transactions if it hears a gossip message of the full transaction. More is described in the [Transaction Gossip Proposal](https://coda.io/d/Gossip_d7HSvA-Pxcz)

## Testing Plan
Unit tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/peerNetwork.ts
+++ b/ironfish/src/network/peerNetwork.ts
@@ -412,14 +412,9 @@ export class PeerNetwork {
         continue
       }
 
-      let sent
-      if (isUpgraded(peer)) {
-        sent = peer.send(newTransactionMessage)
-      } else {
-        sent = peer.send(message)
-      }
+      const messageToSend = isUpgraded(peer) ? newTransactionMessage : message
 
-      if (sent) {
+      if (peer.send(messageToSend)) {
         this.markKnowsTransaction(hash, peer.state.identity)
       }
     }
--- a/ironfish/src/network/peerNetwork.ts
+++ b/ironfish/src/network/peerNetwork.ts
@@ -21,7 +21,7 @@ import { Transaction } from '../primitives'
 import { BlockSerde, SerializedBlock } from '../primitives/block'
 import { BlockHeader, BlockHeaderSerde } from '../primitives/blockheader'
 import { SerializedTransaction, TransactionHash } from '../primitives/transaction'
-import { ArrayUtils, AsyncUtils, ErrorUtils } from '../utils'
+import { ArrayUtils, ErrorUtils } from '../utils'
 import { Identity, PrivateIdentity } from './identity'
 import { CannotSatisfyRequest } from './messages/cannotSatisfyRequest'
 import { DisconnectingMessage, DisconnectingReason } from './messages/disconnecting'
@@ -357,6 +357,7 @@ export class PeerNetwork {
     await this.peerManager.stop()
     this.webSocketServer?.close()
     this.updateIsReady()
+    this.transactionFetcher.stop()
   }
 
   /**
@@ -581,8 +582,9 @@ export class PeerNetwork {
     } else if (message instanceof NewPooledTransactionHashes) {
       this.handleNewPooledTransactionHashes(peer, message)
     } else if (message instanceof NewTransactionV2Message) {
-      const hash = new Transaction(message.transaction).hash()
-      const gossipMessage = new NewTransactionMessage(message.transaction, hash)
+      // Set the nonce to the hash of the transaction for older peers
+      const nonce = Buffer.alloc(16, new Transaction(message.transaction).hash())
+      const gossipMessage = new NewTransactionMessage(message.transaction, nonce)
       await this.onNewTransaction(peer, gossipMessage)
     } else {
       throw new Error(
@@ -672,13 +674,8 @@ export class PeerNetwork {
 
       if (rpcMessage instanceof PooledTransactionsResponse) {
         for (const serializedTransaction of rpcMessage.transactions) {
-          const transaction = new Transaction(serializedTransaction)
-          this.transactionFetcher.receivedTransaction(transaction.hash())
-
           const gossipMessage = new NewTransactionMessage(serializedTransaction)
           await this.onNewTransaction(peer, gossipMessage)
-
-          this.transactionFetcher.removeTransaction(transaction.hash())
         }
       }
     }
@@ -1009,15 +1006,20 @@ export class PeerNetwork {
     const hash = new Transaction(message.transaction).hash()
     peer.state.identity && this.markKnowsTransaction(hash, peer.state.identity)
 
+    // Let the fetcher know that a transaction was received and we no longer have to query it
+    this.transactionFetcher.receivedTransaction(hash)
+
     for (const [_, knownPeer] of peer.knownPeers) {
       knownPeer.state.identity && this.markKnowsTransaction(hash, knownPeer.state.identity)
     }
 
     if (!this.shouldProcessTransactions()) {
+      this.transactionFetcher.removeTransaction(hash)
       return false
     }
 
     if (this.alreadyHaveTransaction(hash)) {
+      this.transactionFetcher.removeTransaction(hash)
       return false
     }
 
@@ -1034,6 +1036,7 @@ export class PeerNetwork {
       this.logger.debug(
         `Invalid transaction '${transaction.unsignedHash().toString('hex')}': ${reason}`,
       )
+      this.transactionFetcher.removeTransaction(hash)
       return false
     }
 
@@ -1047,9 +1050,11 @@ export class PeerNetwork {
 
     if (this.node.memPool.exists(transaction.hash())) {
       this.broadcastTransaction(message)
+      this.transactionFetcher.removeTransaction(hash)
       return true
     }
 
+    this.transactionFetcher.removeTransaction(hash)
     return false
   }
 }
--- a/ironfish/src/network/peerNetwork.ts
+++ b/ironfish/src/network/peerNetwork.ts
@@ -992,18 +992,12 @@ export class PeerNetwork {
   private async onNewTransaction(peer: Peer, message: NewTransactionMessage): Promise<boolean> {
     const received = new Date()
 
-    // Mark the peer as knowing about the transaction as well as their known peers
-    // since they probably sent it to them. We will remove the known peers once we start
-    // gossiping message based on hash
+    // Mark the peer as knowing about the transaction
     const hash = new Transaction(message.transaction).hash()
     peer.state.identity && this.markKnowsTransaction(hash, peer.state.identity)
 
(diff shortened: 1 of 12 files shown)


ironfish/src/network/transactionFetcher.ts
```diff
@@ -0,0 +1,181 @@
...
+    }
+    peer.onStateChanged.on(onPeerDisconnect)
+
+    const clearDisconnectHandler = () => {
+      peer.onStateChanged.off(onPeerDisconnect)
+    }
+
+    const timeout = setTimeout(() => {
+      this.fetchFromNextSource(hash)
+    }, 5000)
```
dguenther, line 129: Just noting that a lot of logic here is similar to the RPC messaging system -- I think it's okay as-is, so I'm not sure it's worth refactoring to reuse that code, but it's pretty identical but with a different timeout

ironfish/src/network/transactionFetcher.ts
```diff
@@ -0,0 +1,181 @@
...
+    currentState && this.cleanupCallbacks(currentState)
+
+    // Get the next source
+    const peer = this.popSource(hash)
+    if (!peer) {
+      this.pending.delete(hash)
+      return
+    }
+
+    const timeout = setTimeout(() => this.sendRequest(hash, peer), 500)
```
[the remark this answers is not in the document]
dguenther: Also, I think this will add a delay when fetchFromNextSource is called on subsequent peers -- I think it's fine to move on to the next peer immediately if the first peer fails to respond or errors out in some way

ironfish/src/network/transactionFetcher.ts
```diff
@@ -0,0 +1,181 @@
...
+    if (state.action === 'IN_FLIGHT') {
+      clearTimeout(state.timeout)
+      state.clearDisconnectHandler()
+    } else if (state.action === 'REQUEST_SCHEDULED') {
+      clearTimeout(state.timeout)
+    }
+  }
+
+  private isResolved(hash: TransactionHash): boolean {
+    // TODO: Also check transactions recently added to the chain
```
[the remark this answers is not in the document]
danield9tqh: Yea LRU cache is what I was thinking. ETH has a cache in the mempool but honestly might make more sense to have the cache in `Blockchain.ts`. If we want to cache transactions every time we call [`onConnectBlock`](https://github.com/iron-fish/ironfish/blob/564cee5817b4bd6967dcf8371c86c6ad74fa6868/ironfish/src/blockchain/blockchain.ts#L103) then I feel like it would make more sense to be in the chain since theoretically we could have a transaction that makes it to the chain but was never seen by the mempool and we'd still want to mark that as resolved

ironfish/src/network/transactionFetcher.ts
danield9tqh: I know, I was trying to figure out a way to combine them. Especially since we try to resolve all RPC Responses in the PeerNetwork and this one would kind of be bypassing that. The main problem I was running into was that if a Transaction comes in while the request is out then we want to clear the handlers (timeout and peer onStateChanged). If we used the RPC messaging logic those would kind of be hidden from us. I'll try to think about it a little more to see if I can come up with a good solution

ironfish/src/network/transactionFetcher.ts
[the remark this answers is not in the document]
danield9tqh: Yea this is a good idea. I'll rework it to select from random peers and to not have the extra delay when re-querying

ironfish/src/network/transactionFetcher.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  private popSource(hash: TransactionHash): Peer | undefined {
+    const sources = this.sources.get(hash)
+
+    if (!sources) {
+      return undefined
+    }
+
+    let nextSource
+    for (const source of sources.values()) {
```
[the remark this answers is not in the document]
danield9tqh: Seems similar to the below comment but I agree. That could be useful

ironfish/src/network/transactionFetcher.ts
```diff
@@ -0,0 +1,181 @@
...
+    this.sources.set(hash, sources)
+
+    if (!currentState) {
+      this.fetchFromNextSource(hash)
+    }
+  }
+
+  /* Called when a transaction is received and confirmed from the network
+   * either in a block or in a gossiped transaction request */
+  resolve(transaction: Transaction): void {
```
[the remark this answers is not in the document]
danield9tqh: yea that was one of the ~5 names this function had throughout development haha

ironfish/src/network/transactionFetcher.ts
```diff
@@ -0,0 +1,210 @@
...
+    this.memPool = memPool
+    this.chain = chain
+
+    this.chain.onConnectBlock.on((block) => {
+      this.onConnectBlock(block)
+    })
+  }
+
+  /**
+   * Called when a new transaction hash is received from the newtork
```
dguenther, line 60:
```suggestion
   * Called when a new transaction hash is received from the network
```

ironfish/src/network/transactionFetcher.ts
```diff
@@ -0,0 +1,181 @@
...
+  // Set of peers that also may be able to fetch the transaction
+  private readonly sources = new BufferMap<Set<Peer>>()
+
+  private readonly memPool: MemPool
+
+  constructor(memPool: MemPool) {
+    this.memPool = memPool
+  }
+
+  /* Called when a new transaction hash is received from the newtork
```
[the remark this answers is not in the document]
dguenther: Resolving this, since the autosuggestion doesn't seem like it'll work

ironfish/src/network/peerNetwork.ts
```diff
@@ -359,17 +372,54 @@ export class PeerNetwork {
...
+      }
+
+      let sent
+      if (isUpgraded(peer)) {
+        sent = peer.send(newTransactionMessage)
+      } else {
+        sent = peer.send(message)
+      }
+
+      if (sent) {
```
dguenther, line 422: I think we could simplify this down a bit and make it more like the hash-sending code by doing this:

```suggestion
      const messageToSend = isUpgraded(peer) ? newTransactionMessage : message
      if (peer.send(messageToSend)) {
```

ironfish/src/network/peerNetwork.ts
```diff
@@ -589,14 +646,29 @@ export class PeerNetwork {
...
+          const hash = new Transaction(serializedTransaction).hash()
+          this.markKnowsTransaction(hash, peer.state.identity)
+        }
+      }
     } else {
       const request = this.requests.get(rpcId)
       if (request) {
         request.resolve({ peerIdentity, message: rpcMessage })
-      } else {
-        this.logger.debug(`Dropping response to unknown request ${rpcId}`)
```
dguenther, line 599: Maybe we could put this back and add it to an `else` after the `if (rpcMessage instanceof PooledTransactionsResponse) {` call? I don't feel too strongly about it though

ironfish/src/network/peerNetwork.ts
```diff
@@ -896,21 +963,49 @@ export class PeerNetwork {
...
-      this.syncedTransactionHashes.has(hash) &&
-      this.node.memPool.exists(hash) &&
-      peersToSendTo.length === 0
-    ) {
+     * When we receive a new transaction we want to test if we have already processed it yet
+     * meaning we have it in the mempool or we have it on a block. */
+
+    let peersToSendTo = 0
+    for (const _ of this.connectedPeersWithoutTransaction(hash)) {
+      peersToSendTo++
```
dguenther, line 982: If you want to make maximum use of generator, you could break out of this after hitting 1 peer, I think

ironfish/src/network/peerNetwork.ts
```diff
@@ -896,21 +963,49 @@ export class PeerNetwork {
...
+
+    // Mark the peer as knowing about the transaction
+    const hash = new Transaction(message.transaction).hash()
+    peer.state.identity && this.markKnowsTransaction(hash, peer.state.identity)
+
+    // Let the fetcher know that a transaction was received and we no longer have to query it
+    this.transactionFetcher.receivedTransaction(hash)
+
+    if (!this.shouldProcessTransactions()) {
+      this.transactionFetcher.removeTransaction(hash)
```
dguenther, line 1003: I wonder if we should make a `wrapper` function that calls receivedTransaction, onNewTransaction, then removeTransaction. I think having to remember to call removeTransaction before every return is a bit prone to error

ironfish/src/network/transactionFetcher.ts
```diff
@@ -0,0 +1,202 @@
...
+ * hashes to full transactions from the network. It operates as a state machine. Each transaction
+ * hash has it's own state which changes when peers are queried, requests timeout or a
+ * full transaction is received from the network
+ */
+export class TransactionFetcher {
+  // State of the current requests for each transaction
+  private readonly pending = new BufferMap<TransactionState>()
+
+  // Set of peers that also may be able to fetch the transaction
+  private readonly sources = new BufferMap<Set<Peer>>()
```
dguenther, line 38: I think we should store peerIdentities rather than peers here, to avoid preventing Peers from getting garbage collected.

ironfish/src/network/peerNetwork.ts
danield9tqh: Nice yea I like it

ironfish/src/network/peerNetwork.ts
danield9tqh: I feel like RPC handling needs to be revisited so I think it's uncertain if we'll want this. I tried adding it and it would need to be 
```
} else if (!request) {
   this.logger.debug(`Dropping response to unknown request ${rpcId}`)
}
```

and it just seems kinda confusing

ironfish/src/network/peerNetwork.ts
```diff
@@ -896,59 +960,72 @@ export class PeerNetwork {
...
+      peersToSendTo = true
+      break
+    }
 
-    // Validate the transaction, so that the account and mempool do not receive
-    // an invalid transaction, and we do not gossip.
-    const { valid, reason } = await this.chain.verifier.verifyTransactionNoncontextual(
-      transaction,
+    return (
+      (this.node.memPool.exists(hash) || this.recentlyAddedToChain.has(hash)) && !peersToSendTo
```
dguenther, line 984:
```suggestion
      this.recentlyAddedToChain.has(hash) || (this.node.memPool.exists(hash) && !peersToSendTo)
```

dguenther (approved):
This looks good to go! We can hold off merging this until we're ready to do the release we intend to put it in, since it's fairly major.

PR #1811, opened 2022-07-19
feat(rust,rust-nodejs): Add asset type to note

rohanjadvani:
## Summary

* Adds `asset_type` field to `Note` struct
* Exposes asset identifier in argument when posting transaction

## Testing Plan

N/A

## Breaking Change

```
[ ] Yes
```


mat-if (approved):
:shipit:

PR #1813, opened 2022-07-20
fix(ironfish): Remove `submittedSequence` check in `accounts:transactions`

rohanjadvani:
## Summary

If you submit transactions that make it onto the main chain on an account and import that account (in a separate directory, machine, etc.), your transactions will show as `pending` since `submittedSequence` was never populated for these records. That value is only set for locally initiated transactions. We should update the check to use `blockHash`. This PR also updates the status to include a `forked` status.

Screen Shot 2022-07-19 at 8 09 39 PM

cc @wd021 

## Testing Plan

Manual test

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/account/accounts.ts
+++ b/ironfish/src/account/accounts.ts
@@ -1087,8 +1087,8 @@ export class Accounts {
     await this.scanTransactions()
   }
 
-  getTransactions(account: Account): {
-    transactions: {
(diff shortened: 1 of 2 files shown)


wd021:
@rohanjadvani thanks, good catch 🙏

NullSoldier (approved):

PR #1815, opened 2022-07-20
Remove unnecessary type aliases

mat-if:
## Summary

These were useful when these types all had generic arguments, but since those were removed, we no longer need these type aliases. They only serve to confuse rust-analyzer when it does imports now.

## Breaking Change

```
[ ] Yes
```


rohanjadvani (approved):

dguenther (approved):
Looks like there are some conflicts from the removed wasm

PR #1819, opened 2022-07-20
Wallet 2.0 Refactor

NullSoldier:
## Summary

This introduces Wallet 2.0 to the code base.

The biggest improvement is the re-write of the schemas to store post decrypted notes. This means that things like creating a transaction and balance calculation will be must faster because they can use a pre-calculated set of decrypted notes and not have to slowly decrypt them again.

It also introduces our first major migration, which migrates the old wallet from Wallet 1.0 to Wallet 2.0.

## Testing Plan
Basic Test
1. Run migration on existing database
2. Check your balance
3. Create some transactions

Advanced test
1. Create a new network
2. Mine some blocks and test transactions between multiple counts
3. Test reorganizations move notes off the main chain too

## Breaking Change

```
[x] Yes
```
Added a migration to migrates to the new wallet format. It can potentially run for a long time on a large wallet.


ironfish/src/account/accounts.ts
```diff
@@ -1309,14 +1134,87 @@ export class Accounts {
...
+  async getLatestHeadHash(): Promise<Buffer | null> {
+    let latestHeader = null
+
+    for (const headHash of this.headHashes.values()) {
+      if (!headHash) {
+        continue
+      }
+
+      const header = await this.chain.getHeader(Buffer.from(headHash, 'hex'))
+      Assert.isNotNull(header, `getLatestHeadHash: No header found for ${headHash}`)
```
NullSoldier, line 1174: This will crash if the chain DB is deleted, or were on a fork that's pruned.

hughy (approved):

PR #1820, opened 2022-07-20
Add default trait to memo for better memo ergonomics

mat-if:
## Summary

- Add derived `Default` trait
- Change usage of `Memo([0; 32])` to `Memo::default()`

There looks like some other cleanup we can do, specifically around serializing/deserializing, but this is a good easy win

## Breaking Change

```
[ ] Yes
```


danield9tqh:
📝

PR #1823, opened 2022-07-20
[WIP] Remove nonce proposal

danield9tqh:
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/account/accounts.ts
+++ b/ironfish/src/account/accounts.ts
@@ -13,7 +13,7 @@ import { MemPool } from '../memPool'
 import { NoteWitness } from '../merkletree/witness'
 import { Mutex } from '../mutex'
 import { Note } from '../primitives/note'
-import { Transaction } from '../primitives/transaction'
+import { Transaction, TransactionHash } from '../primitives/transaction'
 import { ValidationError } from '../rpc/adapters/errors'
 import { IDatabaseTransaction } from '../storage'
 import { PromiseResolve, PromiseUtils, SetTimeoutToken } from '../utils'
@@ -429,6 +429,11 @@ export class Accounts {
     }
 
     return decryptedNotes
+  }
+
+  async transactionIsSynced(hash: TransactionHash): Promise<boolean> {
+    // TODO: Implement
+    return Promise.resolve(false)
   }
 
   /**
--- a/ironfish/src/network/peerNetwork.ts
+++ b/ironfish/src/network/peerNetwork.ts
@@ -517,21 +517,26 @@ export class PeerNetwork {
     peer: Peer,
     gossipMessage: GossipNetworkMessage,
   ): Promise<void> {
-    if (!this.seenGossipFilter.added(gossipMessage.nonce)) {
-      return
-    }
-
     const peerIdentity = peer.getIdentityOrThrow()
 
     if (gossipMessage instanceof NewBlockMessage) {
+      if (!this.seenGossipFilter.added(gossipMessage.nonce)) {
+        return
+      }
       const gossip = await this.onNewBlock({ peerIdentity, message: gossipMessage })
 
       if (gossip) {
         this.broadcastBlock(gossipMessage)
       }
     } else if (gossipMessage instanceof NewTransactionMessage) {
+      if (!this.seenGossipFilter.added(gossipMessage.nonce)) {
+        return
+      }
       await this.onNewTransaction({ peerIdentity, message: gossipMessage })
     } else {
+      if (!this.seenGossipFilter.added(gossipMessage.nonce)) {
+        return
+      }
       throw new Error(`Invalid gossip message type: '${gossipMessage.type}'`)
     }
   }
--- a/ironfish/src/network/peerNetwork.ts
+++ b/ironfish/src/network/peerNetwork.ts
@@ -349,28 +349,34 @@ export class PeerNetwork {
     }
   }
 
-  private broadcastTransaction(
-    gossipMessage: NewTransactionMessage,
-    peerIdentity: string,
-  ): void {
+  private getPeersToSendTo(peerIdentity: string): ReadonlyArray<Peer> {
     const peersConnections =
       this.peerManager.identifiedPeers.get(peerIdentity)?.knownPeers || new Map<string, Peer>()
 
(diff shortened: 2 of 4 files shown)


ironfish/src/network/peerNetwork.ts
```diff
@@ -840,6 +854,18 @@ export class PeerNetwork {
       return false
     }
 
+    const hash = new Transaction(message.message.transaction).hash()
+
+    const peersToSendTo = this.getPeersToSendTo(message.peerIdentity, hash)
+
+    if (
```
danield9tqh, line 861: This is the important change here. It replaces the gossip nonce filter for transactions. The nonce filter guards against 3 actions:
1. We don't sync a transaction gossip message to accounts twice (this includes the expensive operation of verifying and validating the transaction)
2. We don't try to add the same gossip message to our mempool twice (this also includes verifying and validating the transaction)
3. We don't send the same gossip message to our peers twice

This check is basically guarding against all those things so that we can remove the checking of the gossip nonce for transactions

ironfish/src/network/peers/peerManager.ts
```diff
@@ -741,15 +741,6 @@ export class PeerManager {
     return !!peer.state.identity && this.banned.has(peer.state.identity)
   }
 
-  /**
-   * Send a message to all connected peers.
-   */
-  broadcast(message: NetworkMessage): void {
```
danield9tqh, line 747: This is trivial, just moving this logic to the `PeerNetwork` so that we can mark peers as having seen transaction hashes

ironfish/src/network/peerNetwork.ts
```diff
@@ -317,16 +323,6 @@ export class PeerNetwork {
     this.updateIsReady()
   }
 
-  /**
-   * Send the message to all connected peers with the expectation that they
-   * will forward it to their other peers. The goal is for everyone to
-   * receive the message.
-   */
-  private gossip(message: GossipNetworkMessage): void {
```
danield9tqh, line 325: This is trivial, just moving this logic to the above so we can mark peers as having seen transaction hashes

ironfish/src/network/peerNetwork.ts
```diff
@@ -348,29 +344,42 @@ export class PeerNetwork {
     }
   }
 
-  private broadcastTransaction(
-    gossipMessage: NewTransactionMessage,
-    peerIdentity: string,
-  ): void {
+  private getPeersToSendTo(peerIdentity: string, hash: TransactionHash): ReadonlyArray<Peer> {
```
danield9tqh, line 347: This is refactoring out which peers to send the gossip message to. The only logic change is filtering out peers that have already seen the transaction hash. I put this into a separate function so that we can check if there are any peers that need the transaction before actually doing the intensive work of verifying and validating the transaction

ironfish/src/network/peerNetwork.ts
```diff
@@ -518,21 +527,26 @@ export class PeerNetwork {
...
     gossipMessage: GossipNetworkMessage,
   ): Promise<void> {
-    if (!this.seenGossipFilter.added(gossipMessage.nonce)) {
-      return
-    }
-
     const peerIdentity = peer.getIdentityOrThrow()
 
     if (gossipMessage instanceof NewBlockMessage) {
+      if (!this.seenGossipFilter.added(gossipMessage.nonce)) {
```
danield9tqh, line 533: This is just continuing to use the `seenGossipFilter` for blocks. I didn't want to also refactor block logic for this proposal. In this new logic `seenGossipFilter` will only contain block nonces

ironfish/src/network/peerNetwork.ts
NullSoldier: What's the trade offs of doing this? We might starve the network of deposits because rebroadcasting doesn't work? You basically get 1 shot to broadcast your deposit and thats it right?

ironfish/src/account/accounts.ts
```diff
@@ -431,6 +431,11 @@ export class Accounts {
     return decryptedNotes
   }
 
+  async transactionIsSynced(hash: TransactionHash): Promise<boolean> {
```
danield9tqh, line 434: [UNKNOWN] This will have to be implemented in a fast way as it replaces checking the `seenGossipFilter` so it will probably get called a lot

ironfish/src/network/peerNetwork.ts
danield9tqh: The `await this.node.accounts.transactionIsSynced(hash)` is currently not implemented. That is one thing I did not know how to do. Worst case scenario we could keep a in memory filter in the `PeerNetwork` for transaction hashes that have already been synced

danield9tqh: Yea that's right. This was the proposed change because both ETH and BTC are implemented this way. We do have a couple guarantees. A peer is not getting the transaction a second time if
1. We already successfully sent them the transaction
2. They sent us the transaction

I think this is good reason not to send it to that peer again but if you think we should allow sending a transaction to those peers again then I'm open to talking about it

dguenther: Bitcoin and Ethereum both filter outgoing transactions based on whether the peer has already seen the transaction, so I think filtering out re-sends on a per-peer basis is okay:

https://github.com/ethereum/devp2p/blob/master/caps/eth.md#transaction-exchange
https://github.com/bitcoin/bitcoin/blob/d1e42659bbdd8da170542d8c638242cd94f71a7d/src/net_processing.cpp#L1898

This change removes the global seenGossipFilter, which fixes the issue where if one peer receives a nonce it has already seen, it won't continue rebroadcasting to its peers. However, we still have to account for the currently-deployed version of Iron Fish. So I believe if we want to do a soft rollout, we wouldn't want to deploy a version that switches over to the new Transaction propagation method until enough people have upgraded to the code in this PR. Does that seem right?

danield9tqh: Yea you’re right we can’t switch over until enough people have upgraded. I think the main reason for that is that we can’t have that situation where you have to broadcast to a legacy peer and don’t have the original nonce. That would create a bunch of new nonces

danield9tqh:
Closing for https://github.com/iron-fish/ironfish/pull/1829

PR #1824, opened 2022-07-20
feat(rust): Create miners fee transaction

rohanjadvani:
## Summary

Create new implementation for miners fee transactions

## Testing Plan

Unit tests

## Breaking Change

```
[ ] Yes
```


mat-if (approved):
:shipit:

Issue #1528, opened 2022-05-27
After Server Reboot !


Masternode24:
### What happened?

```
May 28 01:29:35 Ubuntu-2004-focal-64-minimal-hwe yarn[3597]: $ yarn build && yarn start:js start
May 28 01:29:35 Ubuntu-2004-focal-64-minimal-hwe yarn[3609]: $ tsc -b
May 28 01:29:35 Ubuntu-2004-focal-64-minimal-hwe yarn[3628]: $ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node bin/run start
May 28 01:29:35 Ubuntu-2004-focal-64-minimal-hwe yarn[3647]:                                            
May 28 01:29:35 Ubuntu-2004-focal-64-minimal-hwe yarn[3647]: ::::::::::          :::::::::::::::::
May 28 01:29:35 Ubuntu-2004-focal-64-minimal-hwe yarn[3647]: ::::::::::::       :::::::::::::::::::
May 28 01:29:35 Ubuntu-2004-focal-64-minimal-hwe yarn[3647]: :::::::::::::     :::::::::::::::::::::
May 28 01:29:35 Ubuntu-2004-focal-64-minimal-hwe yarn[3647]: ::::::::::::::   ::::::::::::::::::::::::
May 28 01:29:35 Ubuntu-2004-focal-64-minimal-hwe yarn[3647]:  ::::::::::::: ::::::::::        :::::::::
May 28 01:29:35 Ubuntu-2004-focal-64-minimal-hwe yarn[3647]:    :::::::::::::::::::::          ::::::::::
May 28 01:29:35 Ubuntu-2004-focal-64-minimal-hwe yarn[3647]:    :::::::::::::::::::::          :::::::::
May 28 01:29:35 Ubuntu-2004-focal-64-minimal-hwe yarn[3647]:  ::::::::::::: ::::::::::        :::::::::
May 28 01:29:35 Ubuntu-2004-focal-64-minimal-hwe yarn[3647]: ::::::::::::::   ::::::::::::::::::::::::
May 28 01:29:35 Ubuntu-2004-focal-64-minimal-hwe yarn[3647]: :::::::::::::     :::::::::::::::::::::
May 28 01:29:35 Ubuntu-2004-focal-64-minimal-hwe yarn[3647]: ::::::::::::       :::::::::::::::::::
[...]

Issue #1467, opened 2022-05-20
"Expected value not to be null" when start node


ffteng46:
### What happened?

when i execute ./ironfish start in the direction of ironfish/ironfish-cli/bin,the thread crashed immediately and give this error message.
The operating sequence is ./ironfish stop,then ./ironfish start. This crash happend after ./ironfish start.

### Version

0.1.35 

### Debug output

```shell
Iron Fish version            0.1.35 @ src
Iron Fish library            0.0.13 @ src
Operating system             Linux x64
CPU model(s)                 AMD EPYC 7542 32-Core Processor
CPU threads                  64
RAM total                    1003.90 GiB
Node version                 v16.15.0
ironfish in PATH             false
Telemetry enabled            false
Accounts head hash           69e263e931fa1a2a4b0437a8eff79ffb7a353b6384a7aeac9f90ac12ae4811ef
Accounts head in chain       true
Accounts head sequence       1
```

### Relevant log output

```shell
WebSocket server started at :::9033
/home/cephd/ironfish/ironfish/build/src/assert.js:23
            throw new Error(message || `Expected value not to be null`);
                  ^

Error: Chain processor head not found in chain: 0000000000002e08d23c72d18535426ce32a95b545cebb4c7f33198db2445bbd
    at Function.isNotNull (/home/cephd/ironfish/ironfish/build/src/assert.js:23:19)
    at ChainProcessor.update (/home/cephd/ironfish/ironfish/build/src/chainProcessor.js:54:25)
    at async MinedBlocksIndexer.updateHead (/home/cephd/ironfish/ironfish/build/src/indexers/minedBlocksIndexer.js:174:33)
    at async MinedBlocksIndexer.eventLoop (/home/cephd/ironfish/ironfish/build/src/indexers/minedBlocksIndexer.js:168:9)
```

### Graffiti

luol-tech-node9

deekerno:
Does this issue still persist for you?

Issue #1447, opened 2022-05-18
Stratum Error Error: Expected value not to be null


CooperNic:
### What happened?

When I started _ironfish miners:start_ command the error message occurred:
Stratum Error Error: Expected value not to be null
    at Function.isNotNull (/usr/src/app/node_modules/@ironfish/sdk/build/src/assert.js:23:19)
    at MiningPoolMiner.newWork (/usr/src/app/node_modules/@ironfish/sdk/build/src/mining/poolMiner.js:71:25)
    at StratumClient.onData (/usr/src/app/node_modules/@ironfish/sdk/build/src/mining/stratum/stratumClient.js:144:32)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
Waiting for graffiti from pool...

### Version

O.1.35

### Debug output

```shell
Iron Fish version       0.1.35 @ 4bc3765
Iron Fish library       0.0.13 @ 4bc3765
Operating system        Linux x64
CPU model(s)            AMD Ryzen 5 3600 6-Core Processor
CPU threads             12
RAM total               62.78 GiB
Node version            v16.13.0
ironfish in PATH        true
Telemetry enabled       true
```

### Relevant log output

```shell
~$ docker run --rm --tty --interactive --name iron_fish_miner --network host --volume $HOME/.ironfish:/root/.ironfish ghcr.io/iron-fish/ironfish:latest miners:start -p pool.ironfish.network -a ac1bf****
Starting to mine with public address: ac1bf**** at pool 54.90.34.95:9034
Connecting to pool...
Successfully connected to pool
Listening to pool for new work
Waiting for graffiti from pool...
Waiting for graffiti from pool...
Waiting for graffiti from pool...
Waiting for graffiti from pool...
Waiting for graffiti from pool...
Waiting for graffiti from pool...
Stratum Error Error: Expected value not to be null
    at Function.isNotNull (/usr/src/app/node_modules/@ironfish/sdk/build/src/assert.js:23:19)
    at MiningPoolMiner.newWork (/usr/src/app/node_modules/@ironfish/sdk/build/src/mining/poolMiner.js:71:25)
    at StratumClient.onData (/usr/src/app/node_modules/@ironfish/sdk/build/src/mining/stratum/stratumClient.js:144:32)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
Waiting for graffiti from pool...
Hashrate... ⣽ 36.12 MH/s (36128152)
```

### Graffiti

[...]

Issue #1438, opened 2022-05-18
pool stratum error


QustomQure:
### What happened?

Stratum data somehow had null
![Screenshot (33)](https://user-images.githubusercontent.com/43998465/169024016-5238c5d1-8958-4c06-9ff8-598dd81e0a45.png)

### Version

0.1.34

### Debug output

```shell
PS C:\> ironfish debug
Database in use, skipping output that requires database.
Stop the node and run the debug command again to show full output.

Iron Fish version       0.1.34 @ e0f5e11
Iron Fish library       0.0.12 @ e0f5e11
Operating system        Windows_NT x64
CPU model(s)            AMD Ryzen 9 5950X 16-Core Processor
CPU threads             32
RAM total               63.92 GiB
Node version            v16.13.2
ironfish in PATH        true
Telemetry enabled       true
```

### Relevant log output

```shell
PS C:\Windows\System32> ironfish miners:start -p pool.ironfish.network -a 1e3c2ca8bba2e56e1a14b578f43abb08b6d28d8ac97a413a0c118efd2e39f1172d142568bc12f915414f16 -t 24
Starting to mine with public address: 1e3c2ca8bba2e56e1a14b578f43abb08b6d28d8ac97a413a0c118efd2e39f1172d142568bc12f915414f16 at pool 54.90.34.95:9034
Connecting to pool...
Successfully connected to pool
Listening to pool for new work
Waiting for graffiti from pool...
Stratum Error Error: Expected value not to be null
    at Function.isNotNull (C:\Users\fr0ze\AppData\Roaming\npm\node_modules\ironfish\node_modules\@ironfish\sdk\build\src\assert.js:23:19)
    at MiningPoolMiner.newWork (C:\Users\fr0ze\AppData\Roaming\npm\node_modules\ironfish\node_modules\@ironfish\sdk\build\src\mining\poolMiner.js:71:25)
    at StratumClient.onData (C:\Users\fr0ze\AppData\Roaming\npm\node_modules\ironfish\node_modules\@ironfish\sdk\build\src\mining\stratum\stratumClient.js:144:32)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
Waiting for graffiti from pool...
Found share: 0000000019788b67 6196 90.09 MH/s
```

### Graffiti

Qualia

lwisne:
This was happening when we were having issues with the pool. Should be fixed now.

PR #1828, opened 2022-07-21
Fixed bugs 1826, 1528, 1467, 1447, 1438, 1413: "Expected value not to be null"

### LINKED ISSUE #1826: Error when adding new block 103268 from l2sPq2O: Error: Expected value not to be null
### Description

I start ironfish node server and mining 
Ironfish start
Ironfish miners:start
20min later, the following error code appears:

image
graffiti: ironfishup

hamza-siddiq:
## Summary
Closes #1826, #1528, #1467, #1447, #1438, #1413 

## Testing Plan
Unit Tested, and then ran the network for a few hours. Tested successfully

## Breaking Change

No
```
[ ] Yes
```
graffiti: ironfishup

### DIFF
--- a/ironfish/src/syncer.ts
+++ b/ironfish/src/syncer.ts
@@ -442,7 +442,9 @@ export class Syncer {
 
       if (!this.loader) {
         this.logger.info(`Syncing orphan chain from ${peer.displayName}`)
-        this.startSync(peer)
+        if (peer.sequence !== null) {
+          this.startSync(peer)
+        }
       } else {
         this.logger.info(`Sync already in progress from ${this.loader.displayName}`)
       }


rohanjadvani:
How does this change fix the null exceptions in the linked tickets? Is this check related to those errors?

ironfish/src/syncer.ts
```diff
@@ -442,7 +442,9 @@ export class Syncer {
 
       if (!this.loader) {
         this.logger.info(`Syncing orphan chain from ${peer.displayName}`)
-        this.startSync(peer)
+        if (peer.sequence !== null) {
```
NullSoldier, line 445: We should always have a peer sequence if the peer is identified. This should never happen right? Are we trying to sync from peers that are not identified?

hamza-siddiq:
@

> How does this change fix the null exceptions in the linked tickets? Is this check related to those errors?
It checks to see if the peer sequence is not null, and then proceeds. This fixes the above bugs.

ironfish/src/syncer.ts
hamza-siddiq: Yes, you should always have a peer sequence if the peer is identified. This should never happen. You are also trying to sync from peers that are not identified. Other than fixing the bug like this, there is no other method. Or to fix this bug, you’ll have to to remake hash algorithm.

danield9tqh:
I am going to close this PR since it does not actually fix any of the issues mentioned

This does not fix https://github.com/iron-fish/ironfish/issues/1528 or https://github.com/iron-fish/ironfish/issues/1467 because those bugs are caused by the chain head being null, not the peer sequence.

It is completely unrelated to https://github.com/iron-fish/ironfish/issues/1447 and https://github.com/iron-fish/ironfish/issues/1438 because those happen when running a mining pool NOT a node.

https://github.com/iron-fish/ironfish/issues/1413 and https://github.com/iron-fish/ironfish/issues/1826 are the only two that are possibly related but this does not fix those either. 
1. These issues could be caused by the peer sequence being null, but they could also be caused by `peer.work` or `chain.head` being null as well, it is unclear
2. Even if they are caused by the peer's sequence being null, this will not fix the problem. It will only cover up the error and not surface it to the user. I think we would rather have the error surface to the user than cover it up.

PR #1834, opened 2022-07-21
bump version to 0.0.17 and 0.1.39

danield9tqh:
## Breaking Change

```
[ ] Yes
```


NullSoldier:
Don't we need to bump rust? I think there were some rust changes going in?

danield9tqh:
yep, updated Rust

dguenther (approved):

PR #1836, opened 2022-07-22
feat(ironfish): Speed up confirmed balance for accounts

rohanjadvani:
## Summary

* Fix accounts and only store decryptable notes that each account owns
* Add `sequence` to transaction store
* Make `transactionHash` non-null for decrypted notes
* Store collection of note hashes that are not on the chain (i.e. corresponding transaction does not have a block hash)
* Store mapping of sequences to decrypted note hashes
* Subtract unconfirmed sequence range of unspent note values and non-chain note values for confirmed balance

## Testing Plan

Covered by existing tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/account/account.ts
+++ b/ironfish/src/account/account.ts
@@ -37,7 +37,6 @@ export class Account {
   readonly incomingViewKey: string
   readonly outgoingViewKey: string
   publicAddress: string
-  rescan: number | null
 
   constructor({
     id,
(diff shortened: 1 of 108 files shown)


ironfish/src/account/account.ts
```diff
@@ -22,10 +22,14 @@ export class Account {
     Readonly<{
       transaction: Transaction
       blockHash: string | null
+      sequence: number | null
```
NullSoldier, line 25: should this be blockSequence to match that it's the sequence that came from the block its on?

NullSoldier (approved):

PR #1838, opened 2022-07-22
Add migration for Wallet 2.0

NullSoldier:
## Summary

This migration will upgrade the wallet schema from 11 to 13, and prevent needing a database reset.

The migration freezes the stores for version 11, and 13 into the data folder for migration for use by the migration. You need to freeze the stores, and cannot use the stores in the wallet because they could change with a new migration introduced after this.

I've outlined all of the changes occurring in this migration here, https://github.com/iron-fish/ironfish/blob/account-migration/ironfish/src/migrations/data/013-wallet-2/readme.md#summary

## Testing Plan

1. Run migration on a wallet and check the balance afterwards to see if it's the same
2. Check the migration has run on a wallet with multiple accounts, and that it works

## Breaking Change

```
[x] Yes
```

### DIFF
--- a/ironfish/src/migrations/data/013-wallet-2.ts
+++ b/ironfish/src/migrations/data/013-wallet-2.ts
@@ -24,8 +24,7 @@ type Stores = {
 export class Migration013 extends Migration {
   path = __filename
 
-  async prepare(node: IronfishNode): Promise<IDatabase> {
-    await node.files.mkdir(node.accounts.db.location, { recursive: true })
+  prepare(node: IronfishNode): IDatabase {
     return createDB({ location: node.accounts.db.location })
   }
 
@@ -37,8 +36,11 @@ export class Migration013 extends Migration {
   ): Promise<void> {
     const startTotal = BenchUtils.startSegment()
 
+    const chainDb = createDB({ location: node.config.chainDatabasePath })
+    await chainDb.open()
+
     const stores: Stores = {
-      old: loadOldStores(db),
+      old: loadOldStores(db, chainDb),
       new: loadNewStores(db),
     }
 
@@ -121,6 +123,7 @@ export class Migration013 extends Migration {
     await stores.old.noteToNullifier.clear(tx)
     logger.debug('\t' + BenchUtils.renderSegment(BenchUtils.endSegment(start)))
 
+    await chainDb.close()
     logger.debug(BenchUtils.renderSegment(BenchUtils.endSegment(startTotal)))
   }
 
--- a/ironfish/src/migrations/data/013-wallet-2.ts
+++ b/ironfish/src/migrations/data/013-wallet-2.ts
@@ -74,7 +74,7 @@ export class Migration013 extends Migration {
     )
     logger.debug('\t' + BenchUtils.renderSegment(BenchUtils.endSegment(start)))
 
-    logger.debug('Migrating: balanaces')
+    logger.debug('Migrating: balances')
     start = BenchUtils.startSegment()
     await this.migrateBalances(stores.new.balances, unconfirmedBalances, accounts, tx, logger)
     logger.debug('\t' + BenchUtils.renderSegment(BenchUtils.endSegment(start)))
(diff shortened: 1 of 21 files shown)


mat-if (commented):
Reviewed the "easy" bits (headhash, meta, accounts). Going to dig into the notes/transaction pieces now.

ironfish/src/migrations/data/013-wallet-2.ts
```diff
@@ -0,0 +1,354 @@
...
+    logger: Logger,
+  ): Promise<{ account: DatabaseStoreValue<NewStores['accounts']>; id: string }[]> {
+    const accounts = []
+
+    for await (const [accountName, accountValue] of accountsStoreOld.getAllIter(tx)) {
+      const accountId = uuid()
+
+      logger.debug(`\tAssigned account id ${accountName}: ${accountId}`)
+
+      await accountsStoreNew.put(uuid(), accountValue, tx)
```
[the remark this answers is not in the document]
NullSoldier: Nice bug, this would have lead to all data relying on account id being broken.

ironfish/src/migrations/data/013-wallet-2/readme.md
```diff
@@ -0,0 +1,56 @@
...
+  1. Migrate `balances`
+    1. Use unconfirmed balances calculated from migrating decrypted notes to index into new `balances` store
+  1. Migrate `headHashes`
+    1. Index new headHashes store for each account using Meta.headHash
+  1. Migrate `meta`
+  1. Delete store `noteToNullifier`
+  1. Delete store `nullifierToNote`
+
+# Potential issues:
+  - What happens if we cant decrypt a note?
```
rohanjadvani, line 51:
```suggestion
  - What happens if we can't decrypt a note?
```

ironfish/src/migrations/data/013-wallet-2/readme.md
```diff
@@ -0,0 +1,56 @@
...
+    - We drop it.
+    - Probably was decrypted previously with an account that hasd been since removed.
```
rohanjadvani, line 53:
```suggestion
    - Probably was decrypted previously with an account that has been since removed.
```

ironfish/src/migrations/data/013-wallet-2/readme.md
```diff
@@ -0,0 +1,56 @@
...
+    - Worse case they can rescan
```
rohanjadvani, line 54:
```suggestion
    - Worst case they can rescan
```

ironfish/src/migrations/data/013-wallet-2/new/transactions.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export type TransactionsStore = IDatabaseStore<{
+  key: Buffer
+  value: TransactionsValue
+}>
+
+export interface TransactionsValue {
+  transaction: Buffer
+  blockHash: string | null
+  submittedSequence: number | null
```
rohanjadvani, line 15: Should this have the `sequence`?

ironfish/src/migrations/data/013-wallet-2.ts
```diff
@@ -0,0 +1,354 @@
...
+  async migrateNullifierToNoteHash(
+    nullifierToNoteOld: Stores['old']['nullifierToNote'],
+    nullifierToNoteHashNew: Stores['new']['nullifierToNoteHash'],
+    tx: IDatabaseTransaction,
+    logger: Logger,
+  ): Promise<void> {
+    let count = 0
+
+    for await (const [nullifier, noteHash] of nullifierToNoteOld.getAllIter(tx)) {
+      logger.debug(`\tMigrating note's nullifier: ${HashUtils.renderHashHex(noteHash)}`)
```
mat-if, line 174: If we remove this line, the log output is pretty concise. It's a shame we don't have 1 extra level of verbosity, would be great for stuff like this. Curious what you think about removing this one, I understand this probably has/had some use, so if we want to keep it, that's cool too.
```suggestion
```

Log output is fixed size, even with >100k notes 
image

ironfish/src/migrations/data/013-wallet-2/new/transactions.ts
NullSoldier: Added, fixed.

ironfish/src/migrations/data/013-wallet-2.ts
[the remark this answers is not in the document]
hughy: It looks like this was reverted

mat-if (approved):

hughy (approved):

PR #1841, opened 2022-07-22
Remove most of Strategy

mat-if:
## Summary

There's probably a lot of ways to clean this up a bit better, but this removes most of the strategy spider-web.

## Breaking Change

```
[ ] Yes
```


ironfish/src/blockchain/blockchain.ts
```diff
@@ -956,7 +956,7 @@ export class Blockchain {
       if (index < noteCount) {
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         const oldNote = (await this.notes.get(index, tx))!
-        if (!this.strategy.noteSerde.equals(note, oldNote)) {
+        if (!note.serialize().equals(oldNote.serialize())) {
```
mat-if, line 959: Move this to `.equals` function on the note class

ironfish/src/indexers/minedBlocksIndexer.test.ts
```diff
@@ -164,8 +165,9 @@ describe('MinedBlockIndexer', () => {
 
     it('returns all mined blocks with scanForks flag included', async () => {
       const { node: nodeA, strategy } = await nodeTest.createSetup()
```
NullSoldier, line 167: Personal note to come back and figure out why this has to be disabled now

ironfish/src/primitives/block.test.ts
```diff
@@ -8,10 +8,11 @@ import {
 } from '../testUtilities/fixtures'
 import { makeBlockAfter } from '../testUtilities/helpers/blockchain'
 import { createNodeTest } from '../testUtilities/nodeTest'
-import { SerializedBlock } from './block'
+import { BlockSerde, SerializedBlock } from './block'
 
 describe('Block', () => {
   const nodeTest = createNodeTest()
+  const blockSerde = BlockSerde
```
NullSoldier, line 15: Is there some mocking reason we can't just use `BlockSerde` everywhere in this file?

ironfish/src/primitives/block.ts
```diff
@@ -70,7 +68,7 @@ export class Block {
   }
 
   equals(block: Block): boolean {
-    return block === this || this.header.strategy.blockSerde.equals(this, block)
+    return block === this || BlockSerde.equals(this, block)
```
NullSoldier, line 71: one day this should be inlined and Serde.equals should be deleted.

ironfish/src/primitives/block.ts
```diff
@@ -98,15 +96,9 @@ export type SerializedBlock = {
...
-export class BlockSerde implements Serde<Block, SerializedBlock> {
-  blockHeaderSerde: BlockHeaderSerde
-
-  constructor(readonly strategy: Strategy) {
-    this.blockHeaderSerde = new BlockHeaderSerde(strategy)
-  }
-
-  equals(block1: Block, block2: Block): boolean {
-    if (!this.blockHeaderSerde.equals(block1.header, block2.header)) {
+export class BlockSerde {
```
NullSoldier, line 99: this part is a bit odd because we're mixing serializer patterns now I guess. It looks like a few others were already made static, and serdeInstances are the other ones that are still instances.

ironfish/src/blockchain/blockchain.ts
NullSoldier: Yeah I expect that to just be checking the note hashes for equality

NullSoldier:
This PR seems pretty good. I'll test it manually.

ironfish/src/primitives/block.test.ts
mat-if: No - relic from before I made them static. Will update

ironfish/src/primitives/block.ts
mat-if: That's a good question. I think there's one more step to be done here, probably removing these "Serde" classes that no longer implement the `Serde` interface, and moving the methods to their respective classes. The remaining ones actually use the `Serde` generic bits for the merkle tree stuff which is something that should be a separate PR.

NullSoldier:
I tested this manually, seems good.

NullSoldier (approved):

PR #1842, opened 2022-07-22
WIP: Create transfer and miners fee transactions in JS

rohanjadvani:
## Breaking Change

```
[ ] Yes
```


ironfish/src/consensus/verifier.ts
```diff
@@ -134,15 +135,13 @@ export class Verifier {
...
-      const transaction = new Transaction(serializedTransaction)
+      const transaction = parseTransaction(data)
 
-      // Transaction is lazily deserialized, so we use takeReference()
+      // Transaction is lazily deserialized, so we use withReference()
       // to force deserialization errors here
-      transaction.takeReference()
-
-      return transaction
+      return transaction.withReference(() => transaction)
```
mat-if, line 144: Oh interesting.. I wonder if this has any implications (or if this _had_ unintended implications)

mat-if (commented):
NICE

rohanjadvani:
@mat-if I'm going to close this for now with our current redesign. Will re-open if I feel like this is still a good starting point.

PR #1845, opened 2022-07-22
Try to keep models intact

NullSoldier:
## Summary

Reduced destructuring to keep fields that belong to models on those
models. In one example were hiding a reference to an existing variable
because of a destructure.

## Testing Plan
-

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/account/account.ts
+++ b/ironfish/src/account/account.ts
@@ -138,14 +138,13 @@ export class Account {
   }> {
     const unspentNotes = []
 
-    for (const [hash, { noteIndex, serializedNote, spent, transactionHash }] of this
-      .decryptedNotes) {
-      if (!spent) {
+    for (const [hash, decryptedNote] of this.decryptedNotes) {
+      if (!decryptedNote.spent) {
         unspentNotes.push({
           hash,
-          index: noteIndex,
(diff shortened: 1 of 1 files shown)


ironfish/src/account/account.ts
```diff
@@ -138,14 +138,13 @@ export class Account {
...
+    for (const [hash, decryptedNote] of this.decryptedNotes) {
+      if (!decryptedNote.spent) {
         unspentNotes.push({
           hash,
-          index: noteIndex,
-          note: new Note(serializedNote),
-          transactionHash,
+          index: decryptedNote.noteIndex,
+          note: new Note(decryptedNote.serializedNote),
+          transactionHash: decryptedNote.transactionHash,
```
NullSoldier, line 147: This occluded the `transactionHash` above. Did we mean to use this one, or did we mean to use the one passed in?

rohanjadvani: I don't think there's anything passed in. It looks like that might be from the return type.

rohanjadvani (approved):

ironfish/src/account/account.ts
NullSoldier: oh true, opps.

PR #1855, opened 2022-07-24
Display options extend from CliUx.ux.table

hairtail:
## Summary
Support display options extend from CliUx.ux.table.
## Testing Plan
Local test.
## Breaking Change

```
[ ] Yes
[✓] No
```


dguenther (approved):
Looks good, thanks!

PR #1859, opened 2022-07-25
Add handler for GetCompactBlockRequest

dguenther:
## Summary

Adds a handler for the GetCompactBlockRequest message. The message isn't in use anywhere yet. This also includes a function added to Block to build a compact block from itself.

## Testing Plan

Added tests for the handler.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/utils/block.ts
+++ b/ironfish/src/network/utils/block.ts
@@ -1,9 +1,13 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import bufio from 'bufio'
+import bufio, { sizeVarBytes, sizeVarint } from 'bufio'
 import { Assert } from '../../assert'
-import { SerializedBlock } from '../../primitives/block'
+import {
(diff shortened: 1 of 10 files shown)


ironfish/src/network/utils/block.ts
```diff
@@ -43,6 +47,26 @@ export function writeBlock(
   return bw
 }
 
+export function writeCompactBlock(
```
danield9tqh, line 50: maybe another reason for composable serialization strategy 😅

ironfish/src/primitives/block.ts
```diff
@@ -78,6 +78,24 @@ export class Block {
     Assert.isNotUndefined(tx, 'Block has no miners fee')
     return tx
   }
+
+  toCompactBlock(): SerializedCompactBlock {
+    const header = new BlockHeaderSerde(this.header.strategy).serialize(this.header)
```
danield9tqh, line 83: I think we were getting rid of the the serialization strategies. Might have to change this to just `BlockHeaderSerde.serialize`

dguenther: Right on, fixed this up after rebasing

danield9tqh (approved):

Issue #1858, opened 2022-07-25
FIP: Add node name and graffiti to `status` and `debug`


mat-if:
### Description

This would be very helpful for users to confirm these are set, and also helpful for us to debug things quicker when issues are reported. I think adding them right under `Node`/`Iron Fish library` is probably the best spot for them.

Both fields come from the [`Config`](https://github.com/iron-fish/ironfish/blob/master/ironfish/src/fileStores/config.ts):
- Node name: `config.get('nodeName')`
- Graffiti: `config.get('blockGraffiti')`

Entry points:
- [Status command](https://github.com/iron-fish/ironfish/blob/master/ironfish-cli/src/commands/status.ts)
- [Debug command](https://github.com/iron-fish/ironfish/blob/master/ironfish-cli/src/commands/debug.ts)

For the status command, it probably makes sense to include it in the RPC endpoint payload. The debug command is a bit less clean in it's design since it's mostly a dev tool for when things break, so grabbing them directly from `node.config` (`node.config.get('nodeName')`) etc is fine.

PR #1862, opened 2022-07-25
Add nodeName and blockGraffiti to debug command

drunkensiberia:
## Summary
FIP: #1858

Adding nodeName and blockGraffiti information to debug command via sdk.config.get()

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/debug.ts
+++ b/ironfish-cli/src/commands/debug.ts
@@ -50,6 +50,9 @@ export default class Debug extends IronfishCommand {
 
     const telemetryEnabled = this.sdk.config.get('enableTelemetry').toString()
 
+    const nodeName = this.sdk.config.get('nodeName').toString()
+    const blockGraffiti = this.sdk.config.get('blockGraffiti').toString()
+
     let cmdInPath: boolean
     try {
       execSync('ironfish --help', { stdio: 'ignore' })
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/debug.ts
```diff
@@ -69,6 +72,8 @@ export default class Debug extends IronfishCommand {
       ['Node version', `${process.version}`],
       ['ironfish in PATH', `${cmdInPath.toString()}`],
       ['Telemetry enabled', `${telemetryEnabled}`],
+      ['Node Name', `${nodeName ? nodeName : 'NONE'}`],
+      ['Block Graffiti', `${blockGraffiti ? blockGraffiti : 'NONE'}`],
```
wd021, line 76: not sure we need this check here. `NONE` is a valid node name and graffiti. would just print the string.

nit: lowercase n and g :) just to keep it consistent

NullSoldier: It's probably fine to just print out the block graffiti and leave it blank.

NullSoldier (approved):

NullSoldier:
It looks like the tests are failing.

ironfish-cli/src/commands/debug.ts
drunkensiberia: @wd021 I removed this checking, thanks for comment :)

drunkensiberia:
Example:
image

Example with empty graffiti and nodeName:
image

PR #1863, opened 2022-07-25
Add nodeName and blockGraffiti to status command

drunkensiberia:
## Summary
FIP: #1858

Adding nodeName and blockGraffiti information to status command via RPC payload.
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/status.ts
+++ b/ironfish-cli/src/commands/status.ts
@@ -92,6 +92,12 @@ function renderStatus(content: GetStatusResponse): string {
     telemetryStatus += ` - ${content.telemetry.submitted} <- ${content.telemetry.pending} pending`
   }
 
+  const nodeName = `${content.node.nodeName ? content.node.nodeName : 'NONE'}`
+
+  const blockGraffiti = `${
+    content.miningDirector.blockGraffiti ? content.miningDirector.blockGraffiti : 'NONE'
+  }`
+
   const peerNetworkStatus = `${
     content.peerNetwork.isReady ? 'CONNECTED' : 'WAITING'
   } - In: ${FileUtils.formatFileSize(
@@ -135,6 +141,8 @@ function renderStatus(content: GetStatusResponse): string {
   return `
 Version              ${content.node.version} @ ${content.node.git}
 Node                 ${nodeStatus}
+Node Name            ${nodeName}
+Block Graffiti       ${blockGraffiti}
 Memory               ${memoryStatus}
 P2P Network          ${peerNetworkStatus}
 Mining               ${miningDirectorStatus}
--- a/ironfish-cli/src/commands/status.ts
+++ b/ironfish-cli/src/commands/status.ts
@@ -92,11 +92,9 @@ function renderStatus(content: GetStatusResponse): string {
     telemetryStatus += ` - ${content.telemetry.submitted} <- ${content.telemetry.pending} pending`
   }
 
-  const nodeName = `${content.node.nodeName ? content.node.nodeName : 'NONE'}`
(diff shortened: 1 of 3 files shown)


NullSoldier:
Can you paste the new example output for status so we can see what it looks like in the PR description?

ironfish/src/rpc/routes/node/getStatus.ts
```diff
@@ -211,6 +216,7 @@ function getStatus(node: IronfishNode): GetStatusResponse {
       status: 'started',
       miners: node.miningManager.minersConnected,
       blocks: node.miningManager.blocksMined,
+      blockGraffiti: node.config.config.blockGraffiti,
```
NullSoldier, line 219: `config.config` is not meant to be used. Use config.ge() instead like this:

```js
config.get('blockGraffiti')
```

ironfish/src/rpc/routes/node/getStatus.ts
```diff
@@ -198,6 +202,7 @@ function getStatus(node: IronfishNode): GetStatusResponse {
       status: node.started ? 'started' : 'stopped',
       version: node.pkg.version,
       git: node.pkg.git,
+      nodeName: node.config.config.nodeName,
```
NullSoldier, line 205: `config.config` is not meant to be used. Use config.ge() instead like this:

```js
config.get('nodeName')
```

drunkensiberia:
@NullSoldier 

Example with populated nodeName and graffiti:
image


Example with empty nodeName and blockGraffiti:
image

ironfish/src/rpc/routes/node/getStatus.ts
drunkensiberia: thanks, updated

ironfish-cli/src/commands/status.ts
```diff
@@ -135,6 +141,8 @@ function renderStatus(content: GetStatusResponse): string {
   return `
 Version              ${content.node.version} @ ${content.node.git}
 Node                 ${nodeStatus}
+Node name            ${nodeName}
+Block graffiti       ${blockGraffiti}
```
drunkensiberia, line 145: Maybe we need to change this to:
> Node Name
> Block Graffiti
Does it look better than lowercase n,g?

ironfish/src/rpc/routes/node/getStatus.ts
```diff
@@ -211,6 +216,7 @@ function getStatus(node: IronfishNode): GetStatusResponse {
       status: 'started',
       miners: node.miningManager.minersConnected,
       blocks: node.miningManager.blocksMined,
+      blockGraffiti: node.config.get('blockGraffiti').toString(),
```
NullSoldier, line 219: Isn't this already a string?

ironfish/src/rpc/routes/node/getStatus.ts
```diff
@@ -198,6 +202,7 @@ function getStatus(node: IronfishNode): GetStatusResponse {
       status: node.started ? 'started' : 'stopped',
       version: node.pkg.version,
       git: node.pkg.git,
+      nodeName: node.config.get('nodeName').toString(),
```
NullSoldier, line 205: Isn't this already a string?

drunkensiberia: Yeah they both are strings, that toString() is not necessary here

NullSoldier (approved):

NullSoldier:
I just noticed that unicode is broken in the blessed box in status, we need to enable it in the screen using this https://github.com/chjj/blessed/issues/377

PR #1866, opened 2022-07-26
Chore: Add doc for blockchain.ts functions

GavPierce:
## Summary
This PR  documents function parameters and usage in blockchain.ts

## Breaking Change
Not a breaking change.

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -324,6 +324,12 @@ export class Blockchain {
     await this.db.close()
   }
 
+  /**
+   * This function will take a block and if valid will connect it to the chain.
+   *
+   * @param block  The block to add to the chain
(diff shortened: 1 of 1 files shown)


danield9tqh:
Hey @GavPierce thanks for the contribution. Unfortunately I think it's subjective whether this improve the code or not. Adding documentation for something like `@param block The block to add to the chain` seems redundant since it can be easily determined from the function signature. Similarly with the other params. It also adds more lines to the file which could be a downside when developers are trying to skim through the file.

I am going to close this PR for that reason. For acceptable contributions I would focus on more objective improvements like bug fixes or feature additions. We are going to be working on adding more Github issues with the tag [`good first issue`](https://github.com/iron-fish/ironfish/contribute) that contributors can pick up and solve. In the meantime you can also go through the open issues on Github. Thanks!

PR #1867, opened 2022-07-26
Batch verify transactions

mat-if:
## Summary

This includes a new worker task to verify multiple transactions at a time, which allow us to batch the spend and receipt proof verification. This was a new feature added in bellman 0.11 or 0.12, which lets us speed up the computation time required for verifying blocks.

This will improve speeds in the following scenarios:
- Block syncing (both actual syncing and when new blocks are gossiped)
- Block template verification (when miners find a block)

Here are some benchmarks:

Note that "batch w/ multicore" was a test using `verify_multicore` on the batch verifiers instead of the plain `verify`. This uses an internal-to-bellman `rayon` threadpool. I opted to not include this, as the speed-up is minor and I suspect it might cause issues on machines with less than 8 threads, although I don't have any handy to test on.

```
-- Sync first 1000 blocks
- M1
staging           : ~1m52s
rust upgrades only: ~1m46s
batch 10 notes    : ~1m41s (10% faster)
batch w/ multicore: ~1m35s
- 2015 Quad-core
staging           : ~4m22s
batch 10 notes    : ~3m25s (22% faster)
batch w/ multicore: ~3m22s

-- Sync first 2000 blocks
- M1
staging           : ~8m15s
rust upgrades only: ~7m46s
batch 10 notes    : ~6m46s (18% faster)
batch w/ multicore: ~6m31s
- 2015 Quad-core
staging           : ~19m15s
batch 10 notes    : ~14m12s (27% faster)
batch w/ multicore: ~14m00s
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/transaction/mod.rs
+++ b/ironfish-rust/src/transaction/mod.rs
@@ -630,14 +630,16 @@ pub fn batch_verify_transactions(
     }
 
     if spend_verifier
-        .verify(&mut OsRng, &sapling.spend_params.vk)
+        // .verify(&mut OsRng, &sapling.spend_params.vk)
+        .verify_multicore(&sapling.spend_params.vk)
         .is_err()
     {
         return Err(SaplingProofError::VerificationFailed.into());
     }
 
     if receipt_verifier
-        .verify(&mut OsRng, &sapling.receipt_params.vk)
+        // .verify(&mut OsRng, &sapling.receipt_params.vk)
+        .verify_multicore(&sapling.receipt_params.vk)
         .is_err()
     {
         return Err(SaplingProofError::VerificationFailed.into());
--- a/ironfish-rust/src/transaction/mod.rs
+++ b/ironfish-rust/src/transaction/mod.rs
@@ -12,7 +12,9 @@ use super::{
     witness::WitnessTrait,
     Sapling,
 };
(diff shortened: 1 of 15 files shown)


ironfish-rust/src/transaction/mod.rs
```diff
@@ -422,19 +424,40 @@ impl Transaction {
         // guarantee they are part of this transaction, unmodified.
         let mut binding_verification_key = ExtendedPoint::identity();
 
+        // Batch proof verifiers
+        let mut spends_verifier = Verifier::<Bls12>::new();
```
dguenther, line 428: I don't think it has to be in this PR, but I think it would be nice to replace this function with a call to batch_verify_transactions if possible, so that we only have one instance of the security-sensitive verification logic.

dguenther (approved):
Left a few minor comments, but looks good overall!

ironfish-rust/src/transaction/mod.rs
mat-if: Agreed. I'll have to play with this one a bit since I'll have to change some ownership stuff around that wasn't immediately obvious

mat-if:
Closes IRO-2359

Created another ticket for the follow-up code de-dupe

PR #1872, opened 2022-07-27
Move box/unbox message into Rust

mat-if:
## Summary

Replaces `tweetnacl` with napi bindings to a similar Rust API. Perf testing shows its 10-30x faster, with tasks generally taking <0.05ms, which is faster than it takes to send a job to the worker pool, so this should be an overall slight performance increase.

Closes IRO-2448

## Breaking Change

```
[ ] Yes
```


ironfish-rust/Cargo.toml
```diff
@@ -14,19 +14,21 @@ default = []
...
 byteorder = "1.4.3"
+crypto_box = { version = "0.7.2", features = ["std"] }
 ff = "0.12.0"
 group = "0.12.0"
 jubjub = "0.9.0"
 lazy_static = "1.4.0"
 rand = "0.8.5"
 rust-crypto-wasm = "0.3.1" # in favor of rust-crypto as this one is wasm friendly
-tiny-bip39 = "1.0.0"
+tiny-bip39 = "0.8"
```
mat-if, line 31: We were on `0.8` until today when I merged the rust dep upgrades, but `1.0` has a version conflict with some library deeeeeeeep in `crypto_box` so rolling back for now.

dguenther: Oof, that's inconvenient.

NullSoldier:
Since we're moving this into the main thread, can you convince me that this won't block the event loop now compared to it being off the main thread? Some performance numbers would suffice. What's your thinking here?

mat-if:
> Since we're moving this into the main thread, can you convince me that this won't block the event loop now compared to it being off the main thread? Some performance numbers would suffice. What's your thinking here?

Ah, I meant to include the perf screenshot that kicked this off. Derek mentioned that he was curious about box/unbox in rust, so I ran some tests. This is 25-30x faster, taking roughly 100 _micro_seconds, which is substantially less time than it even takes to serialize the data to the worker pool (as seen by comparing the direct node function call and the worker pool node call)
image

mat-if:
notes to self:

speed comparison of "newKeyPair":

NODE: 7.387
RUST: 0.088

NODE: 0.67
RUST: 0.048

NODE: 0.597
RUST: 0.044

NODE: 0.606
RUST: 0.043

NODE: 0.661
RUST: 0.046

NODE: 1.397
RUST: 0.047

NODE: 0.593
RUST: 0.04

NODE: 0.594
RUST: 0.039

NODE: 0.584
RUST: 0.04

NODE: 0.593
RUST: 0.046

ironfish-rust/src/errors.rs
```diff
@@ -8,6 +8,20 @@ use std::io;
...
+}
+
+impl fmt::Display for IronfishError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+impl Error for IronfishError {}
+
```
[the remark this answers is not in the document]
dguenther: Yeah, the current error handling setup is kind of inconvenient. Not too concerned with making a better one as part of this PR

ironfish/src/network/peers/localPeer.ts
```diff
@@ -89,11 +86,18 @@ export class LocalPeer {
...
-    return this.workerPool.unboxMessage(boxedMessage, nonce, sender, this.privateIdentity)
+  ): { message: string | null } {
+    let message
+    try {
+      message = unboxMessage(boxedMessage, nonce, sender, this.privateIdentity.secretKey)
+    } catch {
+      message = null
+    }
+
+    return { message }
```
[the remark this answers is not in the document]
dguenther: I don't know that the previous approach was correct, but I think it's fine to maintain the behavior here and fix it in a different PR if necessary.

ironfish-rust/src/lib.rs
```diff
@@ -29,7 +31,10 @@ pub use {
     spending::{SpendParams, SpendProof},
     transaction::{ProposedTransaction, Transaction},
 };
-pub mod sapling_bls12;
+
+// Re-exports for napi
+pub use base64;
+pub use hex;
```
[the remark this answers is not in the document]
dguenther: I don't feel strongly about keeping the `napi` crate at 0 deps, if these are only used by the napi crate.

ironfish/src/network/messages/gossipNetworkMessage.ts
```diff
@@ -13,7 +13,7 @@ export abstract class GossipNetworkMessage extends NetworkMessage {
   constructor(type: NetworkMessageType, nonce?: Buffer) {
     super(type)
 
-    this.nonce = nonce ?? Buffer.from(tweetnacl.randomBytes(16))
+    this.nonce = nonce ?? Buffer.from(randomBytes(16))
```
dguenther, line 16: I don't think this has to change, but this could probably just use `import { randomBytes } from 'crypto'` to reduce the API surface of the Rust code a bit -- tweetnacl was just calling that internally anyway.

ironfish/package.json
```diff
@@ -18,7 +18,7 @@
   ],
   "dependencies": {
     "@ironfish/bfilter": "0.0.1",
-    "@ironfish/rust-nodejs": "0.1.9",
+    "@ironfish/rust-nodejs": "0.1.10",
```
dguenther, line 21: Should the version bump for this be in a separate PR? Or should it bump the package.json of ironfish-rust-nodejs in this PR as well

mat-if: Oh, this is a fix that I merged into staging already that I duplicated here to keep testing: #1966

ironfish-rust-nodejs/src/nacl.rs
```diff
@@ -0,0 +1,128 @@
...
+        let secret_key = bytes_to_secret_key(bytes);
+
+        Ok(BoxKeyPair {
+            public_key: secret_key.public_key().as_bytes().to_vec(),
+            secret_key: secret_key.as_bytes().to_vec(),
+        })
+    }
+
+    #[napi(getter)]
+    pub fn public_key(&self) -> Uint8Array {
```
dguenther, line 53: Would it work to return `Buffer` everywhere around here instead of `Uint8Array`? I think most of our code is returning those right now.

ironfish-rust-nodejs/src/nacl.rs
```diff
@@ -0,0 +1,128 @@
...
+    nonce: String,
+    sender_public_key: String,
+    recipient_secret_key: Uint8Array,
+) -> Result<String> {
+    let decoded_sender = base64::decode(sender_public_key)
+        .map_err(|_| Error::from_reason("Unable to decode sender public key".to_owned()))?;
+
+    let sender: [u8; 32] = decoded_sender
+        .try_into()
+        .map_err(|_| Error::from_reason("Unable to convert sender pubic key".to_owned()))?;
```
dguenther, line 113:
```suggestion
        .map_err(|_| Error::from_reason("Unable to convert sender public key".to_owned()))?;
```

ironfish-rust-nodejs/src/nacl.rs
mat-if: Yeah. I went with `Uint8Array` originally since that's what `tweetnacl` was returning, but I like just using `Buffer` since it's our standard anyway.

ironfish/src/network/messages/gossipNetworkMessage.ts
mat-if: Ok but hear me out. This is 1.5 - 2x faster than javascript's crypto randomBytes

(Granted, we're talking about 2-4 microseconds vs 3-7 microseconds, so y'know, world's greatest over-optimzation)

dguenther (approved):

PR #1874, opened 2022-07-28
changes snapshots to zip database from node datadir

hughy:
## Summary

- updates service:snapshot to zip the entire 'databases/<database>' directory
  and upload to s3 for the configured database
- updates chain:import to overwrite database in datadir with downloaded
  snapshot
- fixes downloads to ensure that file is downloaded and hashed completely
- changes both commands to use local flags instead of remote since this method
  requires the node to be local
- updates unit tests for changes to logic

both taking a snapshot and importing require the node to be stopped.

## Testing Plan

- ran `service:snapshot` on databases big and small
- ran `chain:import` to sync new node datases
- updates to unit tests

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/chain/import.ts
```diff
@@ -55,20 +57,15 @@ export default class ImportSnapshot extends IronfishCommand {
...
       if (!flags.confirm) {
         this.log(
-          `This snapshot (${
-            manifest.file_name
-          }) contains the Iron Fish blockchain up to block ${
-            manifest.block_height
-          }. The size of the latest snapshot file is ${FileUtils.formatFileSize(
-            manifest.file_size,
-          )}`,
+          `This snapshot (${manifest.file_name}) contains the Iron Fish blockchain up to block ${manifest.block_height}. The size of the latest snapshot file is ${fileSize}`,
```
NullSoldier, line 68: I know it was this way before, so this is just a note for more work, we should rename height -> sequence in the manifest and the logs. We call it sequence in our code base and our logs. We can do that in another PR.

ironfish-cli/src/commands/chain/import.ts
```diff
@@ -123,17 +122,17 @@ export default class ImportSnapshot extends IronfishCommand {
...
+    const chainDatabasePath = this.sdk.fileSystem.resolve(this.sdk.config.chainDatabasePath)
 
-    for (const file of files) {
-      const blocks = await fsAsync.readFile(path.join(blockExportPath, file))
-      await client.importSnapshot({ blocks })
-    }
+    CliUx.ux.action.start(
+      `Copying snapshot from ${snapshotDatabasePath} to ${chainDatabasePath}`,
+    )
+    await fsAsync.cp(snapshotDatabasePath, chainDatabasePath, { recursive: true, force: true })
```
NullSoldier, line 134: Should we do a move instead of a copy? That way only the disk inodes need to change and not to copy over 30 GB or more of files?

ironfish-cli/src/commands/chain/import.ts
```diff
@@ -87,31 +84,33 @@ export default class ImportSnapshot extends IronfishCommand {
...
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
+      response.data.on('data', (chunk: { length: number }) => {
+        downloaded += chunk.length
+        bar.update(downloaded, { downloadedSize: FileUtils.formatFileSize(downloaded) })
+      })
       // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
       response.data.pipe(writer)
-      hasher.update(response.data)
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
+      response.data.pipe(hasher)
```
NullSoldier, line 112: Does this work to pipe this this way, or is the response stream consumed by the first pipe?

ironfish-cli/src/commands/chain/import.ts
```diff
@@ -87,31 +84,33 @@ export default class ImportSnapshot extends IronfishCommand {
...
-          const percentage = Math.floor((progressEvent.loaded / progressEvent.total) * 100)
-          bar.update(percentage)
-        },
       })
 
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
+      response.data.on('data', (chunk: { length: number }) => {
+        downloaded += chunk.length
+        bar.update(downloaded, { downloadedSize: FileUtils.formatFileSize(downloaded) })
+      })
```
NullSoldier, line 108: We need to listen to error too and handle that from the request, especially since this is a large file.

ironfish-cli/src/commands/service/snapshot.ts
```diff
@@ -108,48 +93,23 @@ export default class CreateSnapshot extends IronfishCommand {
...
-      if (result.buffer && result.seq) {
-        const blockFilePath = this.sdk.fileSystem.join(blockExportPath, `${result.seq}`)
-        await fsAsync.writeFile(blockFilePath, Buffer.from(result.buffer))
-        progress.update(result.seq || 0)
-      }
-    }
-
-    progress.stop()
+    const chainInfoResponse = await client.getChainInfo()
+    const blockHeight = Number(chainInfoResponse.content.currentBlockIdentifier.index)
```
NullSoldier, line 103: This shouldn't need to be casted to a number.. look into this

ironfish-cli/src/commands/service/snapshot.ts
```diff
@@ -108,48 +93,23 @@ export default class CreateSnapshot extends IronfishCommand {
...
 
     const timestamp = Date.now()
 
     const snapshotFileName = `ironfish_snapshot_${timestamp}.tar.gz`
```
NullSoldier, line 107: Do we need to have multiple snapshots here? What if we have just 1 path and delete the one that exists if it already exists. You'll need to clean up all the snapshots otherwise but having 1 will reduce the need for that.

ironfish-cli/src/commands/service/snapshot.ts
```diff
@@ -108,48 +93,23 @@ export default class CreateSnapshot extends IronfishCommand {
...
     const snapshotPath = this.sdk.fileSystem.join(exportDir, snapshotFileName)
 
-    this.log(`Zipping\n    SRC ${blockExportPath}\n    DST ${snapshotPath}\n`)
-    CliUx.ux.action.start(`Zipping ${blockExportPath}`)
-    await this.zipDir(blockExportPath, snapshotPath)
+    this.log(`Zipping\n    SRC ${chainDatabasePath}\n    DST ${snapshotPath}\n`)
+    CliUx.ux.action.start(`Zipping ${chainDatabasePath}`)
+    await this.zipDir(chainDatabasePath, snapshotPath, [databaseLockPath])
```
NullSoldier, line 112: It's not safe to exclude the lock file. That should be unnecessary. For this we'll need to open the DB to ensure that no one else has opened it. The reason is, that it's not safe to zip a leveldb database if writes are inflight. It's almost guaranteed to become corrupted if that's the case. You can use waitForOpen() and then close the node afterwards to guarantee we have exclusive access to the node databases.

ironfish-cli/src/commands/chain/import.ts
hughy: It works this way! I also wondered about this, but checksums match on import

ironfish-cli/src/commands/chain/import.ts
hughy: Yeah a move would be better -- faster to rename and doesn't leave an extra 30GB sitting around

ironfish-cli/src/commands/chain/import.ts
hughy: Updated to sequence

ironfish-cli/src/commands/service/snapshot.ts
hughy: Changed to always use the same default filename with an option to set a different one

ironfish-cli/src/commands/chain/import.ts
```diff
@@ -55,20 +63,15 @@ export default class ImportSnapshot extends IronfishCommand {
         this.exit(1)
```
NullSoldier, line 63: ^ up here were writing to a new tmp directory each attempt

ironfish-cli/src/commands/service/snapshot.ts
```diff
@@ -71,20 +76,13 @@ export default class CreateSnapshot extends IronfishCommand {
...
-        : Number(process.env.MAX_BLOCKS_PER_SNAPSHOT_CHUNK),
-      description: 'The max number of blocks per file in the zipped snapshot',
-    }),
   }
 
   async start(): Promise<void> {
     const { flags } = await this.parse(CreateSnapshot)
 
     const bucketUrl = (flags.bucketUrl || DEFAULT_SNAPSHOT_BUCKET_URL || '').trim()
+    const snapshotFileName = (flags.snapshotFileName || DEFAULT_SNAPSHOT_FILE_NAME).trim()
```
NullSoldier, line 85: I think we can get rid of this config, and just use a name here so on disk it has a static name, but in the bucket the snapshot should probably use the timestamp like it was originally.

NullSoldier (approved):

NullSoldier:
We should write the database version in the manifest and check that after downloading the manifest so we don't download a database that is newer than our code base. You can check that version here, if were less than this https://github.com/iron-fish/ironfish/blob/master/ironfish/src/blockchain/blockchain.ts#L59

ironfish-cli/src/commands/chain/import.ts
```diff
@@ -87,31 +90,39 @@ export default class ImportSnapshot extends IronfishCommand {
       const bar = CliUx.ux.progress({
         barCompleteChar: '\u2588',
         barIncompleteChar: '\u2591',
-        format: 'Downloading snapshot: [{bar}] {value}% | ETA: {eta}s',
+        format:
+          'Downloading snapshot: [{bar}] {percentage}% | {downloadedSize} / {fileSize} | ETA: {eta}s',
```
NullSoldier, line 94: We can use a meter like rescan, and get a better estimate and render it using renderEstimate()

ironfish/src/blockchain/blockchain.ts
```diff
@@ -139,6 +139,10 @@ export class Blockchain {
     this._genesis = newGenesis
   }
 
+  get databaseVersion(): number {
+    return DATABASE_VERSION
```
NullSoldier, line 143: So this isn't exactly the DB version, which is important to point out. We don't need this setter because you can use `this.db.getVersion()`. So just to be clear, this is the software DB version, but the `db.getVersion()` is the real DB version that it's migrated to. The difference is, if you haven't started your node and run migrations then this will report a different DB version than the actual DB version of the database.

ironfish/src/rpc/routes/chain/importChain.ts
```diff
@@ -74,7 +75,7 @@ function deserializeChunk(node: IronfishNode, reader: bufio.BufferReader): Block
   for (let i = 0; i < blocksLength; i++) {
     const blockReader = bufio.read(reader.readVarBytes(), true)
     const serializedBlock = readBlock(blockReader)
-    const block = node.chain.strategy.blockSerde.deserialize(serializedBlock)
+    const block = BlockSerde.deserialize(serializedBlock)
```
NullSoldier, line 78: I think this RPC should be deleted now, right? It can happen after this PR.

ironfish/src/rpc/routes/chain/importChain.test.ts
```diff
@@ -19,7 +20,7 @@ describe('Route chain/importSnapshot', () => {
     Assert.isNotNull(genesis)
 
     const blockA1 = await useMinerBlockFixture(chain, 2)
-    const serializedBlockA1 = strategy.blockSerde.serialize(blockA1)
+    const serializedBlockA1 = BlockSerde.serialize(blockA1)
     const bw = bufio.write(getBlockSize(serializedBlockA1))
     const blockA1Buffer = writeBlock(bw, serializedBlockA1).render()
```
NullSoldier, line 26: I think this RPC should be deleted now, right? It can happen after this PR.

ironfish/src/rpc/routes/chain/snapshotChain.test.ts
```diff
@@ -24,7 +25,7 @@ describe('Route chain/snapshotChainStream', () => {
       await expect(chain).toAddBlock(blockA1)
       expect(blockA1.transactions.length).toBe(1)
 
-      const serializedBlockA1 = strategy.blockSerde.serialize(blockA1)
+      const serializedBlockA1 = BlockSerde.serialize(blockA1)
       const bw2 = bufio.write(getBlockSize(serializedBlockA1))
```
NullSoldier, line 29: I think this RPC should be deleted now, right? It can happen after this PR.

ironfish/src/utils/snapshot.ts
```diff
@@ -2,9 +2,10 @@
...
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 export type SnapshotManifest = {
-  block_height: number
+  block_sequence: number
   checksum: string
   file_name: string
   file_size: number
   timestamp: number
+  database_version: number
 }
```
NullSoldier, line 11: This should go in the CLI in export since it's not used by anyone other than the CLI.

ironfish/src/rpc/routes/chain/importChain.ts
hughy: I'll delete these rpc routes and unneeded snapshot code in a subsequent PR

NullSoldier (approved):

NullSoldier (approved):

PR #1876, opened 2022-07-29
Enable fullUnicode rendering for blessed

drunkensiberia:
## Summary
This will enable rendering of full-width unciode characters in terminals which replaced this characters to '??', '?', ''
I turned on this options in 2 places of code, status command (with possible issue in node name and graffiti) and peers list comand with -fenas flag (here we can get issue with node name).

## Testing Plan
Local testing

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):

NullSoldier:
Thanks for this PR, my terminal displays my graffiti now properly! I use emojis in my graffiti to cause havoc and find bugs. :)

PR #1880, opened 2022-07-29
Add IRONFISH_EXPOSE_GC to expose garbage collector

NullSoldier:
## Summary

Exposing the garbage collector is a useful thing for developers. We can
start taking advantage of it in new memory bench marking tools that can
do certain things if the GC is exposed.

## Testing Plan
1. Run debug command with yarn
2. Run debug command with ./bin/ironfish with `IRONFISH_EXPOSE_GC` as `true`

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/bin/ironfish
+++ b/ironfish-cli/bin/ironfish
@@ -41,5 +41,10 @@ else
   if [ "$DEBUG" == "*" ]; then
     echoerr IRONFISH_BINPATH="$IRONFISH_BINPATH" "$NODE" "$DIR/run" "$@"
   fi
-  "$NODE" "$DIR/run" "$@"
+
+  if [ "$IRONFISH_EXPOSE_GC" == "true" ]; then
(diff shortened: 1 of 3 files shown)


ironfish-cli/bin/ironfish
```diff
@@ -41,5 +41,10 @@ else
...
     echoerr IRONFISH_BINPATH="$IRONFISH_BINPATH" "$NODE" "$DIR/run" "$@"
   fi
-  "$NODE" "$DIR/run" "$@"
+
+  if [ "$IRONFISH_EXPOSE_GC" == "true" ]; then
+    "$NODE" "--expose-gc" "$DIR/run" "$@"
+  else
+    "$NODE" "$DIR/run" "$@"
+  fi
 fi
```
NullSoldier, line 50: This is is a generated file, these changes will be lost if we re-generate the oclif project. However, it's fine if this is lost.

hughy (approved):

Issue #1877, opened 2022-07-29
UX issue with accounts:transactions


lwisne:
### What happened?

The`accounts:transactions` command has a slightly misleading display. 

Once a transaction is added to a block and the block is created, there is a block confirmation time (currently 12 blocks) in which you cannot yet spend new notes created in the block and its transactions. This protects against people spending notes that may end up in a fork.

Given this, **we need one more state for a transaction in which we display that the tx is "confirming".** This means it is waiting for enough blocks to process such that the the new notes created by the tx will now be spendable. Without this, you will see "confirmed" transactions that are not yet applied to your "available to spend" balance.

### Graffiti

lwisne

hughy:
Transaction statuses modified in #1882, #1883, #2264 to include additional statuses

PR #1882, opened 2022-07-30
Add confirming status for getTransactions

hairtail:
## Summary
Issue https://github.com/iron-fish/ironfish/issues/1877
## Testing Plan
Local test.
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/account/accounts.ts
+++ b/ironfish/src/account/accounts.ts
@@ -731,8 +731,11 @@ export class Accounts {
           Assert.isNotNull(header)
           const main = await this.chain.isHeadChain(header)
           if (main) {
-            const confirmations = this.chain.head.sequence - header.sequence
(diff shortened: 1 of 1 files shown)


ironfish/src/account/accounts.ts
```diff
@@ -1122,8 +1128,10 @@ export class Accounts {
           const header = await this.chain.getHeader(Buffer.from(blockHash, 'hex'))
           Assert.isNotNull(header)
           const main = await this.chain.isHeadChain(header)
-          if (main) {
+          if (main && header.sequence + minimumBlockConfirmations < heaviestHead.sequence) {
```
dguenther, line 1131: I think there's an off-by-one compared to how confirmations are calculated elsewhere. I'll fix it up quick since this has been sitting around for a while!

dguenther (approved):

PR #1884, opened 2022-07-30
Disallow importing accounts with the same keys

mat-if:
## Summary

There isn't a valid reason to have the same account imported multiple times, and it is a drain of overall node performance since each note needs to be decrypted twice for no good reason. This also allows further optimizations with decrypting notes such that we could stop trying to decrypt a note for accounts once we've found one match, which can be a sizable performance increase for nodes with multiple accounts.

Closes IRO-2378

## Breaking Change

```
[ ] Yes
```


ironfish/src/migrations/data/013-accounts-dedupe.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    for (const duplicateAccount of duplicateAccounts) {
+      await node.accounts.removeAccount(duplicateAccount.name)
+    }
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  async backward(): Promise<void> {}
+}
```
NullSoldier, line 37: I assume the reason you want to do this is to make the data consistent with the variant right? That way no one ever tries to map on the spend key? I think it's probably ok to allow this for now and get this change out without the migration since it already improves things for users.

mat-if: Well, we just shouldn't allow it because like you're alluding to, they can get out of sync and just cause issues.

But I'm mostly concerned about adding the ability to stop trying to decrypt a note as soon as one account has decrypted it. This should lead to fairly substantial speed increases for users with multiple accounts, but we can't do that until we can guarantee account uniqueness.

NullSoldier: Ok, what about this...

1. Merge the change on its own because I think it's a good improvement
2. Merge in the migration along with wallet 2.0 so users run it anyway when running wallet 2.0 migrations and they don't get hit with more migrations
3. Open PR with optimization now that we know wallets are unique per spend key.

rohanjadvani (approved):

NullSoldier:
LGTM

PR #1891, opened 2022-08-01

hamza-siddiq:
## Summary
The default timeout value of jest is set to only 5 seconds. This
period seems insufficient for some tests that apparently take
longer to execute. This PR changes the timeout to 32 seconds,
which should suffice for the current tests. An improvement would be
to make this value get set dynamically according to the specs of the
computer running the tests.

## Breaking Change

No
```
[ ] Yes
```
graffiti: ironfishup


hamza-siddiq:
Ironfish/cli

PR #1894, opened 2022-08-01
Add DatabaseStoreValue and DatabaseStoreKey

NullSoldier:
## Summary

These types let you extract the key and value from an IDatabaseStore
type. How to use:
```typescript
const store = db.addStore<{ key: string, value: AccountsValue }>(...)

type key = DatabaseStoreKey<typeof store> // string
type value = DatabaseStoreValue<typeof store> // value
```

## Testing Plan
Types only, no tests.

## Breaking Change

```
[ ] Yes
```


hughy (approved):
🧙‍♂️

NullSoldier:
🪄

PR #1895, opened 2022-08-01

hamza-siddiq:
## Summary
The default timeout value of jest is set to only 5 seconds. This
period seems insufficient for some tests that apparently take
longer to execute. This PR changes the timeout to 32 seconds,
which should suffice for the current tests. An improvement would be
to make this value get set dynamically according to the specs of the
computer running the tests.

A previous PR https://github.com/iron-fish/ironfish/pull/1891 was closed. I mixed up branches there, so had to make a new branch, and then open the request again. 

[danield9tqh](https://github.com/danield9tqh) in the previous PR asked which tests are timing out?
The tests that are timing out are Ironfish/cli

## Testing Plan
Tested passed

## Breaking Change
No
```
[ ] Yes
```
graffiti: ironfishup

### DIFF
--- a/config/jest.config.base.js
+++ b/config/jest.config.base.js
@@ -6,4 +6,9 @@ module.exports = {
   coverageProvider: "v8",
   coverageReporters: ["text-summary", "json", "clover", "text"],
   testPathIgnorePatterns: ['.*\\.test\\.slow\\.ts$', '.*\\.test\\.perf\\.ts$'],
+  
+  // set timeout to a higher value than the default (5000), in order to avoid 
+  // timing out when running tests. 32 seconds should suffice for the existing
+  // tests.
+  testTimeout: 32000
 };


danield9tqh:
Can you list specific tests that are failing the time limit. `Ironfish/cli` includes a ton of tests. That doesn't narrow it down much

danield9tqh:
Also it's good that we have a time limit for tests. It can help us catch slow tests. I think if tests are failing because of timeout on your machine you can increase the timeout but we should probably keep the timeout for when tests run on Github PR validation and things like that

hamza-siddiq:
@danield9tqh i'll get back to you in a bit. Please don't close this for another 48 hours

lwisne:
FYI here - unless it is failing the official test suite run on github, the only fix we will accept for a test timing out is a code optimization that makes the test in question run faster.

NullSoldier:
The reason the tests are timing out is because our CLI tests are broken and they are using / starting up an actual node using your data at ~/.ironfish. They are timing out because it's spending time connecting to peers, or loading your database, or other things.  We need to delete the CLI tests, or do something with them to fix this issue. 

See our contributing guidelines here, https://github.com/iron-fish/ironfish/blob/master/CONTRIBUTING.md#tests-for-cli-commands

hamza-siddiq:
Thanks @NullSoldier . That is very helpful. Working on this now, will have an update soon

hamza-siddiq:
Still working on this. I'll close this for now, and reopen it later

hamza-siddiq:
So turns out that the way I solved it is really the recommended way of solving such issues.
IMG-20220808-WA0015
This is from the official documentation of "jest" (the framework used to write unit-tests)
Here is the documentation, if you'd like to have a look at it: 
https://jest-bot.github.io/jest/docs/troubleshooting.html

PR #1905, opened 2022-08-02
Add re-usable build-and-test step for rust-nodejs package

mat-if:
## Summary

This allows us to manually run the build and testing of the various host platforms, so we can find build issues before we're mid-deploy

https://docs.github.com/en/actions/using-workflows/reusing-workflows

I don't think the manually-callable "Build ironfish/rust-nodejs NPM Packages" will show up until it reaches master, so this won't be useful for the next release, but after that, we should be able to manually run that job before we merge STAGING -> MASTER if there are rust changes, etc.

Example of the re-usable piece successfully calling the job including the artifacts: https://github.com/iron-fish/ironfish/runs/7622752359?check_suite_focus=true You can click "List packages" to see that it successfully finds the artifacts. I will delete/archive this job after this merges so that it doesn't sit in the actions history for eternity.

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):

.github/workflows/build-ironfish-rust-nodejs.yml
```diff
@@ -0,0 +1,188 @@
+name: Build @ironfish/rust-nodejs NPM Packages
+
+env:
+  DEBUG: 'napi:*'
+  MACOSX_DEPLOYMENT_TARGET: '10.13'
+
+on:
+  workflow_dispatch:
```
NullSoldier, line 8: 👍

PR #1910, opened 2022-08-02
Add locks to all functions that modify transactions/notes

mat-if:
## Summary

You're gonna wanna look at this without whitespace, it's almost entirely whitespace changes: https://github.com/iron-fish/ironfish/pull/1910/files?w=1

Goal here is to add a lock around anything that might modify the transactions/notes so that we can guarantee there's only one writer at any given time.

Closes IRO-2446

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):
This looks good to me, it'll get tested manually here for dead locks when it hits wallet 2.0.

mat-if:
Going to close this in favor of #2046 for now. May end up needing it, will bring it back if so

PR #1913, opened 2022-08-02
Fix test for duplicate key import

NullSoldier:
## Summary

This is now failing on this branch because accounts are in memory.

## Testing Plan
None

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/account/accounts.test.ts
+++ b/ironfish/src/account/accounts.test.ts
@@ -329,9 +329,10 @@ describe('Accounts', () => {
 
       expect(node.accounts.accountExists(account.name)).toEqual(true)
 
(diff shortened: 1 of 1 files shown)


mat-if (approved):

ironfish/src/account/accounts.test.ts
```diff
@@ -329,9 +329,10 @@ describe('Accounts', () => {
 
       expect(node.accounts.accountExists(account.name)).toEqual(true)
 
-      account.name = 'Different name'
```
NullSoldier, line 332: this was modifying the wallet in memory account from the fixture restoration

PR #1915, opened 2022-08-02
Add messages to asserts in accounts

mat-if:
## Summary

Add a handful of messages for asserts in accounts, so that we can tell where an error is occurring easier.

## Breaking Change

```
[ ] Yes
```


ironfish/src/account/accounts.ts
```diff
@@ -428,7 +428,7 @@ export class Accounts {
       for (const [accountId, decryptedNotes] of decryptedNotesByAccountId) {
         await this.db.database.transaction(async (tx) => {
           const account = this.accounts.get(accountId)
```
NullSoldier, line 430: Random note for later but It's possible that on a high volume account, accounts could mismatch this.accounts by the time decryption is done here causing this crash if you remove the high volume account.

NullSoldier (approved):

PR #1918, opened 2022-08-02
Add total memory and OS CPU core count to startup telemetry

deekerno:
## Summary
Address IRO-2362 and IRO-2363. Adds stats for total memory and OS CPU core count.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/telemetry/telemetry.ts
+++ b/ironfish/src/telemetry/telemetry.ts
@@ -281,9 +281,18 @@ export class Telemetry {
   }
 
   submitNodeStarted(): void {
+    let fields: Field[] = [{ name: 'online', type: 'boolean', value: true }]
+
+    if (this.metrics) {
+      fields = fields.concat([
+        { name: 'cpu_cores', type: 'integer', value: this.metrics.cpuCores },
+        { name: 'memory_total', type: 'integer', value: this.metrics.memTotal },
+      ])
+    }
+
     this.submit({
       measurement: 'node_started',
-      fields: [{ name: 'online', type: 'boolean', value: true }],
+      fields,
       timestamp: new Date(),
     })
   }
--- a/ironfish/src/metrics/metricsMonitor.ts
+++ b/ironfish/src/metrics/metricsMonitor.ts
@@ -37,6 +37,8 @@ export class MetricsMonitor {
   readonly memTotal: number
   readonly heapMax: number
 
+  readonly cpuCores: number
+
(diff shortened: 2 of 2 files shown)


ironfish/src/telemetry/telemetry.ts
```diff
@@ -281,11 +281,24 @@ export class Telemetry {
...
+    if (this.metrics) {
+      this.submit({
+        measurement: 'node_started',
+        fields: [
+          { name: 'cpu_cores', type: 'integer', value: this.metrics.cpuCores },
+          { name: 'total_mem', type: 'integer', value: this.metrics.memTotal },
+        ],
+        timestamp,
+      })
+    }
```
NullSoldier, line 301: Can we use the existing telemetry above instead of duplicating it?

deekerno: Fixed.

ironfish/src/telemetry/telemetry.ts
```diff
@@ -281,9 +281,18 @@ export class Telemetry {
   }
 
   submitNodeStarted(): void {
+    const fields: Field[] = [{ name: 'online', type: 'boolean', value: true }]
+
+    if (this.metrics) {
+      fields.concat([
```
NullSoldier, line 287: You should double check what this method is doing to see if it does what you want.

ironfish/src/telemetry/telemetry.ts
```diff
@@ -281,9 +281,18 @@ export class Telemetry {
   }
 
   submitNodeStarted(): void {
+    let fields: Field[] = [{ name: 'online', type: 'boolean', value: true }]
+
+    if (this.metrics) {
+      fields = fields.concat([
+        { name: 'cpu_cores', type: 'integer', value: this.metrics.cpuCores },
+        { name: 'total_mem', type: 'integer', value: this.metrics.memTotal },
```
rohanjadvani, line 289: nit: Kind of weird it's `total_mem` and the metrics field is `memTotal`. What do you think about `memory_total` to be explicit / consistent?

rohanjadvani (commented):
What is the testing plan for this? Can we fill that out or add a screenshot to show this works as intended?

ironfish/src/telemetry/telemetry.ts
NullSoldier: Yes please, I agree with Rohan.

NullSoldier (approved):

PR #1931, opened 2022-08-04
Convert all camel case commands to snake case

NullSoldier:
## Summary

Convert all camel case commands to snake case. We should not use commands like `chain:resetBlocks` it should always be `chain:reset-blocks`. This is the standard git uses, and the one we will use too. However, you should avoid using snake case at all cost and always try to user a single word if you absolutely can even if it's less descriptive. 

## Testing Plan

No testing needed.

## Breaking Change

```
[ ] Yes
```

No breaking changes because all commands were hidden, but they had aliases added anyway.

### DIFF
--- a/ironfish-cli/src/commands/chain/readd-block.ts
+++ b/ironfish-cli/src/commands/chain/readd-block.ts
@@ -6,6 +6,8 @@ import { IronfishCommand } from '../../command'
 import { LocalFlags } from '../../flags'
 
 export default class ReAddBlock extends IronfishCommand {
+  static aliases = ['chain:readdblock']
+
   static description =
     'Remove and readd a block on the chain if it has no other blocks after it'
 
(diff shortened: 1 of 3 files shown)


hughy (approved):
🐫 🐍

hughy:
apparently this is sometimes called kebab-case 🍢

ironfish-cli/src/commands/chain/readd-block.ts
```diff
@@ -6,6 +6,8 @@ import { IronfishCommand } from '../../command'
 import { LocalFlags } from '../../flags'
 
 export default class ReAddBlock extends IronfishCommand {
+  static aliases = ['chain:readdblock']
```
lwisne, line 9: There's an extra d here

lwisne:
> apparently this is sometimes called kebab-case 🍢

Yeah, think that's a more accurate description - snake case is generally defined as using underscores.

PR #1932, opened 2022-08-04
Fix backup command by using snapshot upload

NullSoldier:
## Summary

Backup does not work because files are too big to not use multipart
uploads. This fixes a key command by factoring out the upload code to be
used by backup.

## Testing Plan

Run backup with a public bucket under https://s3.console.aws.amazon.com/s3/buckets

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/service/snapshot.ts
+++ b/ironfish-cli/src/commands/service/snapshot.ts
@@ -1,26 +1,17 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import {
-  AbortMultipartUploadCommand,
-  CompleteMultipartUploadCommand,
-  CreateMultipartUploadCommand,
-  S3Client,
-  UploadPartCommand,
-} from '@aws-sdk/client-s3'
-import { Assert, FileUtils, NodeUtils } from '@ironfish/sdk'
+import { S3Client } from '@aws-sdk/client-s3'
+import { FileUtils, NodeUtils } from '@ironfish/sdk'
 import { CliUx, Flags } from '@oclif/core'
 import crypto from 'crypto'
 import fsAsync from 'fs/promises'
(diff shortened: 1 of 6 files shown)


ironfish-cli/src/commands/service/snapshot.ts
```diff
@@ -1,26 +1,17 @@
...
 import { CliUx, Flags } from '@oclif/core'
 import crypto from 'crypto'
 import fsAsync from 'fs/promises'
 import path from 'path'
-import tar from 'tar'
 import { IronfishCommand } from '../../command'
 import { LocalFlags } from '../../flags'
-import { DEFAULT_SNAPSHOT_BUCKET, SnapshotManifest } from '../../utils'
+import { DEFAULT_SNAPSHOT_BUCKET, S3Utils, SnapshotManifest } from '../../utils'
+import { TarUtils } from '../../utils/tar'
```
hughy, line 13: nit - this could be from `../../utils` too

ironfish-cli/src/utils/tar.ts
```diff
@@ -0,0 +1,39 @@
...
+        } else {
+          return true
+        }
+      },
+    },
+    [sourceFile],
+  )
+}
+
+function unzipTar(source: string, dest: string): Promise<number | null> {
```
hughy, line 31: we could also implement this using the `'tar'` module --`importSnapshot` implements unzip that way

hughy (approved):

ironfish-cli/src/commands/service/snapshot.ts
NullSoldier: In Javascript, if you can import from direct files it actually helps a lot with circular dependency issues which cause modules to become undefined. It's not a hard rule I have, and I'm fine with developers doing either but I'll generally try to import from the file it's from.

ironfish-cli/src/utils/tar.ts
NullSoldier: Yeah that would require code change, I was just moving code around to re-use it. I hope someone else will do that.

PR #1939, opened 2022-08-04
Handle errors better

hamza-siddiq:
## Summary
Feature request:
#1940 
The problem that I have encountered as I dig deeper into the code was that most error messages aren't really handled well. When people encounter error messages they need to be more specific and handled better. I believe this is very important. This PR does exactly that.

1. Add explaination field to `VerifyError' class
This commit adds a field (explaination) that can be used to clarify the error with a detailed message. This commit also makes the instances of Verify errors convertible to string so they can be printed conveniently in logs

2. Add an error to indicate an empty MerkeTree error

3. Properly handle and report errors
This commit handles errors properly, rather than silently ignore them. This commit also changes error type `ERROR' to `UNKNOWN_ERROR' to better signal that the source of the error is unknown (in the case of the inability to identify the error's cause).

4. Use correct error type instead of the generic ERROR

5. Raise the correct error types depending on the error condition

6. Make sure to pass explaination message to both error cases

7. Simplify message returning

8. Update expected error reasons

## Testing Plan
Tested successfully

## Breaking Change

No
```
[ ] Yes
```
graffiti: ironfishup


hamza-siddiq:
Fails because lint, due to unchanged code. Just some syntax issues in your code. You may change it if you like. Other than that, everything looks fine.

lwisne:
Please create or link to an issue this solves, as described in the [new contributor guidelines](https://github.com/iron-fish/ironfish/blob/b0b91940c51009b2b00f5e58e27be89c9a5c1e39/CONTRIBUTING.md).

hamza-siddiq:
@lwise I forgot to make a feature request. Would go ahead and do that right now, because this is that. Would highly recommend this PR

hamza-siddiq:
@lwisne you commented in the issue https://github.com/iron-fish/ironfish/issues/1940 that it's broad, but it's not a big deal since I worked on it myself. I think the PR would help the codebase

lwisne:
This will be changing certain CLI info displays. Can you post before/after?

hamza-siddiq:
@lwisne working on it right now

hamza-siddiq:
Below, I explain how would this PR change how errors are reported in the affected functions:

1. Function `verifySpend` always ignored any error thrown in the functions called within it. The image below shows the erroneous try/catch block:

   image
  
    After the fix, the try/catch block now looks like:

   image

2. Function `pastRoot` used to throw a generic Error (which then got silently discarded):

   image

   After the fix, the correct error reason is returned depending on the actual error condition. A helper lambda/arrow function<sup>(1)</sup> is used to reduce the repetition of error handling code. The check for tree emptiness<sup>(2)</sup> is extracted into an independent check (to return the correct error reason, `EMPTY_MERKLE_TREE`). The remaining spend arguments validity checks<sup>(3)</sup> now return the correct error reason.

   image

3. If `verfiy()` fails, the printed error reason would be "Verfiy_transaction" (`VerificationResultReason.VERIFY_TRANSACTION`) instead of just "Error" (`VerificationResultReason.ERROR`).

   image

4. If `verfiyTransactions()` fails, the printed error reason would be "Verfiy_transaction" 
(`VerificationResultReason.VERIFY_TRANSACTION`) instead of just "Error" (`VerificationResultReason.ERROR`).

   image

ironfish/src/blockchain/blockchain.ts
```diff
@@ -1315,11 +1315,27 @@ export class Blockchain {
 export class VerifyError extends Error {
   reason: VerificationResultReason
   score: number
+  explanation: string | undefined
 
-  constructor(reason: VerificationResultReason, score = 0) {
+  constructor(reason: VerificationResultReason, score = 0, explanation?: string) {
```
danield9tqh, line 1320: I think it's a little redundant adding an explanation parameter when we already have a `reason`. I think a better solution would be to add some documentation of what each reason means

ironfish/src/merkletree/merkletree.ts
```diff
@@ -453,8 +455,22 @@ export class MerkleTree<
     return this.db.withTransaction(tx, async (tx) => {
       const leafCount = await this.getCount('Leaves', tx)
 
-      if (leafCount === 0 || pastSize > leafCount || pastSize === 0) {
-        throw new Error(`Unable to get past size ${pastSize} for tree with ${leafCount} nodes`)
+      Assert.isFalse(leafCount < 0)
```
danield9tqh, line 458: This function `pastRoot` is used in other places besides just verification. That's why we don't throw a `VerifyError` and we throw a generic error instead. `VerificationResultReason.INVALID_SPEND` also does not make sense in this function because this function is used for blocks as well, not just spends.

danield9tqh:
This is not a bad thought for an error message. We actually have a PR open that adds a similar error message. Take a look at it [here](https://github.com/iron-fish/ironfish/pull/2037).

While I appreciate this contribution I would recommend against so-called 'refactoring' PRs. This PR does not change any functionality (besides the CLI output) so it is much more difficult to judge whether it is beneficial or not. If a PR fixes an actual issue then that is easily verifiable. Members of the core team will do refactoring PRs but that's because we have context of how the code is going to evolve in the future. It's much more difficult for external contributors to refactor because they lack that context. This PR for example: if it had been merged in then the changes would have just been lost when we changed the functionality in the [previous PR]((https://github.com/iron-fish/ironfish/pull/2037)) I mentioned.

Additionally it is not correct to put these errors in the `pastRoot` function because `pastRoot` should be agnostic to spends. For example we use it in [`rootHash`](https://github.com/iron-fish/ironfish/blob/5441d5953ea728aa2274f1a8b6a5c9a7cee382bf/ironfish/src/merkletree/merkletree.ts#L511) which is unrelated to spends. If we threw an `VerificationResultReason.INVALID_SPEND` when calling `rootHash` that would not make any sense. So I would also caution against opening PRs that are not fully tested. If there are too many PRs that have these type of issues then we start to lose confidence in the contributor.

We appreciate the contributions but PLEASE try to be mindful of these things as it takes the team a lot of time to review contributions.

ironfish/src/blockchain/blockchain.ts
hamza-siddiq: I added it as an **optional** parameter that can be used in the case of having some additional information that can be attached to the error.

Here is an example usecase:
image

ironfish/src/merkletree/merkletree.ts
hamza-siddiq: I think the solution is not to just use a generic error that never gets caught. If you believe that is not viable to use `VerifyError` then a different (more related?) error type should be used), and that error type should be caught.
And about the use of `INVALID_SPEND`, I couldn't find an error reason from the existing ones that better described the error. Perhaps an additional error reason could be added (i.e., INVALID_TREE_INDEX)?

PR #1953, opened 2022-08-05
Remove 0 metrics from telemetry

danield9tqh:
## Summary
When no network messages have been sent for a certain message type, we don't want to send those messages to telemetry

## Breaking Change

```
[ ] Yes
```


dguenther:
I think we can handle this on the server-side instead if we want to filter these out.

Issue #1792, opened 2022-07-15
error running "ironfish debug"


msa6867:
### What happened?

Entered "ironfish debug" in CLI with node running.  In the past this has never been in issue and the return was to provide what state could be while running the node (which was less state than would be returned when node is stopped).  Now it returns an error as follows:
C:\Users\Mark>ironfish debug
    Error: Database is not open. Call IDatabase.open() first

(If this is the new expected behavior please leave comment and disregard.)

(debug output below is from previous query when node was stopped)

### Version

v0.1.38

### Debug output

```shell
Iron Fish version            0.1.38 @ 85dc82e
Iron Fish library            0.0.16 @ 85dc82e
Operating system             Windows_NT x64
CPU model(s)                 AMD Ryzen 9 5950X 16-Core Processor
CPU threads                  32
RAM total                    31.92 GiB
Node version                 v16.15.0
ironfish in PATH             true
Telemetry enabled            true
Accounts head hash           000000000009a99e6deee713542042dd21e70ae404231fd779428c1387386f08
Accounts head in chain       true
Accounts head sequence       93006
```

### Relevant log output

```shell
C:\Users\Mark>ironfish debug
    Error: Database is not open. Call IDatabase.open() first
```

### Graffiti

msa6867

lwisne:
Do you still get this behavior?

msa6867:
I am away from my setup until sunday.  I will answer when I get back.  When I left on vacation last Wed, the behavior persisted; that was the last I checked

msa6867:
Problem persists

PR #1958, opened 2022-08-05
Fixed a bug: error running "ironfish debug"

hamza-siddiq:
## Summary
The issue https://github.com/iron-fish/ironfish/issues/1792 reported a bug where \`ironfish debug\' command had an exception message shown. After investigating the code, it turned out that not all exceptions were handled by the catch block, which left the `dbOpen' flag variable set to true (when the database may not be actually open). This commit also contains a refactoring of the output collection functions.

## Testing Plan
Before Bug reporter:
> C:\Users\Mark>ironfish debug
Error: Database is not open. Call IDatabase.open() first

Since the issue happened to the bug reporter while running a node, I ran the node while testing.

Before Mine:

> yarn start debug
yarn run v1.22.15
$ yarn build && yarn start:js debug
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc bin/run debug
Database in use, skipping output that requires database.
Stop the node and run the debug command again to show full output.

> Iron Fish version               0.1.44 @ src
Iron Fish library               0.0.21 @ src
Operating system                Linux x64
CPU model(s)                    AMD Ryzen 5 3600 6-Core Processor
CPU threads                     12
RAM total                       62.72 GiB
Heap total                      4.01 GiB
Node version                    v16.16.0
ironfish in PATH                true
Garbage Collector Exposed       true
Telemetry enabled               false
Node name                       
Block graffiti                  ironfishup
Done in 1.81s.

After mine:

> yarn start debug
yarn run v1.22.15
$ yarn build && yarn start:js debug
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc bin/run debug
Database in use, skipping output that requires database.
Stop the node and run the debug command again to show full output.

> Iron Fish version               0.1.44 @ src
Iron Fish library               0.0.21 @ src
Operating system                Linux x64
CPU model(s)                    AMD Ryzen 5 3600 6-Core Processor
CPU threads                     12
RAM total                       62.72 GiB
Heap total                      4.01 GiB
Node version                    v16.16.0
ironfish in PATH                true
Garbage Collector Exposed       true
Telemetry enabled               false
Node name                       
Block graffiti                  ironfishup
Done in 1.81s.

## Breaking Change
No
```
[ ] Yes
```
graffiti: ironfishup


lwisne:
Please show output before and after that you tested this.

hamza-siddiq:
Thanks @lwisne i'll work on that

hamza-siddiq:
@lwisne I have edited the description, and posted the output requested

lwisne:
You don't appear to be fixing the error in the linked bug?

hamza-siddiq:
@lwisne . That's because the bug doesn't happen on my computer as you can see in my before output, most probably doesn't happens on yours as well when you run start debug, hence, why it's most probably still not verified. But it could happen in some cases, as reported by the bug reporter.

Since not all exceptions are handled in the catch block in the before code, this code should care of it. In my case though, the error never happens in both the before or after case, so it never gets caught. But if the error ever happens, and when the error happens in cases like the bug reporter, the catch block should now take care of it. The testing shows that when it runs on my or your computer, it should work fine, when there is no exception to handle.

Is there a way to make the bug appear so the catch block runs. I can't really think of a way, but if you can, please let me know, and I'll help with the testing there as well?

lwisne:
I'm going to save you the trouble - this does not fix bug 1792, or really anything at all.

You literally posted the same output twice as your before/after.

Closing, and please consider people's time when opening up these kinds of spurious reviews.

hamza-siddiq:
I posted that output a while ago. I'll post the new output in a bit, and you'll know what I mean

hamza-siddiq:
@lwisne you are right. I had opened this issue before our discussion, and I noticed you're getting a bit frustrated seeing what i had done. I'm working on easier but efficient issues since we had talked. My apologies man

Issue #1379, opened 2022-05-13
Usage of "@" in account name - inconsistent result and error


johnsgeorgecet:
### What happened?

When @ is used in creating new account name, in one scenario it says unrecognzied token, but in another scenario it is accepted and account is created

PS C:\Users\johnsge> ironfish accounts:create @
At line:1 char:26
+ ironfish accounts:create @
+                          ~
Unrecognized token in source text.
    + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : UnrecognizedToken

PS C:\Users\johnsge> ironfish accounts:create @test
Enter the name of the account: @test
Creating account @test
Account @test created with public address ec3714c92eef191de9b90676709199045b066f2b66960077a7a8b33cf42de90298981f0aef69464bbdc424
Run "ironfish accounts:use @test" to set the account as default
PS C:\Users\johnsge> ironfish accounts:create @1223233
Enter the name of the account: @
Creating account @
Account @ created with public address 12e33c3163f364d9d84f430fb949f2dd30a6c6c0ecdc7e2521bbf058d1fcc3b604b26ef30b37c9cdb48f1a
Run "ironfish accounts:use @" to set the account as default
PS C:\Users\johnsge> ironfish accounts:create @#$%$%#$%
At line:1 char:26
+ ironfish accounts:create @#$%$%#$%
+                          ~
Unrecognized token in source text.
    + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : UnrecognizedToken

PS C:\Users\johnsge>

### Version

0.1.33

### Graffiti

johnsgeorge

sunsweet733:
I haven't this error
http://joxi.ru/p27BK7ptZvpy6m

What is your operating system?

sunsweet733:
What client are you using to connect to the server?

johnsgeorgecet:
THe pasted output is from Windows, npm install. command ran from powershell.
Let me try in mac OS as well

johnsgeorgecet:
Do you mean '@' is a legit input for an address? 

Let me include a few cases how an input with @ behave

case 1 - @ followed by alphanumeric eg. @abc 
-------
PS C:\Users\johnsge> ironfish accounts:create @abc
Enter the name of the account: @abc
Creating account @abc
[...]

PR #1960, opened 2022-08-06
Input validation for account name

hamza-siddiq:
## Summary
Fixes: https://github.com/iron-fish/ironfish/issues/1379

The commit validates account name in the following commands:
accounts:address
accounts:balance
accounts:create
accounts:export
accounts:notes
accounts:pay
accounts:remove
accounts:transactions
accounts:use
deposit
deposit-all
depositAll

The account name can only contain english letters, english numbers and underscores.

The regular expression used in this commit could be changed to meet other requirements.

graffiti: ironfishup

## Testing Plan
Tested successfully

## Breaking Change
No
```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/command.ts
+++ b/ironfish-cli/src/command.ts
@@ -193,3 +193,9 @@ function getFlag(flags: unknown, flag: FLAGS): unknown | null {
       (flags as any)[flag]
     : null
 }
+
+export abstract class InputValidator {
+  public static readonly accountName = new RegExp('[^a-z^A-Z^0-9^_]')
+
+  // Other input validation goes here
+}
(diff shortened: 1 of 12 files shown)


wd021:
I'm not sure this is the solution we want for the original bug, i don't think we ned to limit account names to `[^a-z^A-Z^0-9^_]`. Also, would use a const instead of rewriting it in every file and why is yarn.lock being updated?

hamza-siddiq:
@wd021 Thanks. Would they want the account name to just exclude special characters and if yes what characters exactly?

wd021:
i'll let someone from the team comment on this, but I'm not sure we should remove all special chars

danield9tqh:
@hamza-siddiq I've close the original issue because we don't want to enforce validation of specific account names. We just treat them as strings. You can see the [comment](https://github.com/iron-fish/ironfish/issues/1379#issuecomment-1213518550) I left on the original issue. Sorry for the confusion. I'm going to close this PR too since the validation is not needed

PR #1964, opened 2022-08-06
Dont snapshot resolve after reject

NullSoldier:
## Summary

There is a simple issue here in that during an error, resolve() is
called after reject() is called. This is because 'close' is ALWAYS
called no matter what if the stream finishes, or there is an error.

You can verify this if you add console logs to the writer handlers in
the staging branch.

## Testing Plan

1. Add console logs to writer handers
2. Cause an error to happen and see that 'error' is fired then 'close' is fired
3. This causes reject() then resolve() to trigger

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/chain/download.ts
+++ b/ironfish-cli/src/commands/chain/download.ts
@@ -149,13 +149,20 @@ export default class Download extends IronfishCommand {
       })
 
       await new Promise<void>((resolve, reject) => {
-        writer.on('error', (e) => {
+        const onWriterError = (e: unknown) => {
+          writer.removeListener('close', onWriterClose)
+          writer.removeListener('onWriterError', onWriterError)
           reject(e)
-        })
+        }
 
-        writer.on('close', () => {
+        const onWriterClose = () => {
+          writer.removeListener('close', onWriterClose)
+          writer.removeListener('onWriterError', onWriterError)
           resolve()
-        })
+        }
+
+        writer.on('error', onWriterError)
+        writer.on('close', onWriterClose)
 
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/chain/download.ts
```diff
@@ -149,13 +149,20 @@ export default class Download extends IronfishCommand {
       })
 
       await new Promise<void>((resolve, reject) => {
-        writer.on('error', (e) => {
+        const onWriterError = (e: unknown) => {
+          writer.removeListener('close', onWriterClose)
+          writer.removeListener('onWriterError', onWriterError)
```
hughy, line 154: Should this be `writer.removeListener('error' . . .`?

ironfish-cli/src/commands/chain/download.ts
```diff
@@ -149,13 +149,20 @@ export default class Download extends IronfishCommand {
...
           reject(e)
-        })
+        }
 
-        writer.on('close', () => {
+        const onWriterClose = () => {
+          writer.removeListener('close', onWriterClose)
+          writer.removeListener('onWriterError', onWriterError)
           resolve()
```
hughy, line 161: Would we be better off by removing `onWriterClose` altogether and writing 
```
response.data.on('end', () => {
  writer.close()
  resolve()
}
```
?

hughy (approved):

ironfish-cli/src/commands/chain/download.ts
NullSoldier: I think so, yeah.

Issue #1968, opened 2022-08-06
Feature: Cleaning up test timeouts


wd021:
### Description

Noticed we have some tests with large timeouts that seem arbitrary. For example, `accounts.test.slow.ts` contains 60000 & 600000 ms timeouts (not sure if was a typo).

In a lot of other test files, i see hardcoded timeouts that aren't inline with how long tests actually take.

Would be nice to clean this up and use, test -> default, test:slow -> 60s, test:perf -> 60s, which is sufficient for all our tests.

mat-if:
Fixed and merged #1970 🙌

PR #1970, opened 2022-08-06
Cleanup test timeout values

wd021:
## Summary
Fix for https://github.com/iron-fish/ironfish/issues/1968

using these default timeouts
`test` -> 5s (default)
`test:slow` -> 60s
`test:perf` -> 60s

deleting all unnecessary hardcoded timeout values

## Testing Plan
- run yarn test, yarn test:slow, yarn test:pref in `/ironfish`

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/workerPool/job.test.slow.ts
+++ b/ironfish/src/workerPool/job.test.slow.ts
@@ -20,7 +20,7 @@ describe('Worker Pool', () => {
     expect(minersFee.serialize()).toBeInstanceOf(Buffer)
 
     expect(workerPool.completed).toBe(1)
-  }, 60000)
+  })
 
   it('verify', async () => {
     const { workerPool } = nodeTest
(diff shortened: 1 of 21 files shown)


ironfish/src/workerPool/job.test.slow.ts
```diff
@@ -36,5 +36,28 @@ describe('Worker Pool', () => {
 
     expect(result.valid).toBe(true)
     expect(workerPool.completed).toBe(1)
-  }, 60000)
+  })
+
+  it('boxMessage and unboxMessage', async () => {
```
danield9tqh, line 41: Sorry, did a rebase but accidentally added this test back in 😬

wd021: no worries, fixed

mat-if (approved):
We might need to tweak these a bit when we need to regenerate test fixtures, but this is good cleanup

Issue #1400, opened 2022-05-14
Please add start and final balance from 'ironfish deposit --confirm'


pgulyaev88:
### Description

Please add start and final balance from 'ironfish deposit --confirm'

hughy:
Closed by #1975

PR #1975, opened 2022-08-08
Add transaction details to deposit and deposit --confirm

hamza-siddiq:
## Summary
Fixes: #1400

When the confirm flag was used with the deposit command, there were no details about the transaction like the balance before the deposit, the transaction fee, the amount to bo be sent, the remaining balance after the deposit and what graffiti the memo will contain.

**Update:**
After some discussion, the before and after balances were added to both the commands: deposit, as well as deposit --confirm. The deposit command also displays the before balance, before a transaction gets confirmed. Some text alignment changes were also made. Previous code in the file was also cleaned up, e.g the confirmedBalance was changed to iron once, and the feeInIron was made use of instead of fee, for calculations with IRON_TO_FEE, rather than doing the conversions several times.

Below, you can see the before and after of the deposit as well as deposit --confirm command.

Before deposit:
Before deposit

After deposit:
After deposit

Before deposit --confirm:
Before deposit --confirm

After deposit -- confirm:
After deposit --confirm

## Testing Plan
Tested successfully

## Breaking Change
No

```
[ ] Yes
```
graffiti: ironfishup


ironfish-cli/src/commands/deposit.ts
```diff
@@ -122,6 +122,12 @@ The memo will contain the graffiti "${graffiti}".
         this.log('Transaction aborted.')
         this.exit(0)
       }
+    } else {
+      this.log(`
+Your balance is ${confirmedBalance} and you are about to send ${displayAmount} plus a transaction fee of ${displayFee} to the Iron Fish deposit account.
+Your remaining balance after this transaction will be ${displayNewBalance}.
+The memo will contain the graffiti "${graffiti}".
+      `)
```
wd021, line 130: no need to repeat this code in both if and else condition

hamza-siddiq: It isn't repeated. Look closer

hamza-siddiq: Repeating is fine, since we need to display the transaction transaction details before and after the transaction. But the wording is not. Just a wording mistake. I'll correct it

hamza-siddiq: It's because the if condition is when the confirm flag is not used, and the else when the confirm flag is used. The bug is that when the confirm flag is used, there is no output to show any details regarding the account balance

wd021: i know, i'm just saying both conditions are printing 👇 

```
you are about to send ${displayAmount} plus a transaction fee of ${displayFee} to the Iron Fish deposit account.
Your remaining balance after this transaction will be ${displayNewBalance}.
The memo will contain the graffiti "${graffiti}".
```

hamza-siddiq: @wd021 I updated the code

ironfish-cli/src/commands/deposit.ts
```diff
@@ -108,15 +108,16 @@ export default class Bank extends IronfishCommand {
...
-      this.log(`
-You are about to send ${displayAmount} plus a transaction fee of ${displayFee} to the Iron Fish deposit account.
+
+    this.log(`
+Your balance is ${confirmedBalance} and you are about to send ${displayAmount} plus a transaction fee of ${displayFee} to the Iron Fish deposit account.
 Your remaining balance after this transaction will be ${displayNewBalance}.
 The memo will contain the graffiti "${graffiti}".
 
 * This action is NOT reversible *
       `)
```
hughy, line 118:
```suggestion
    `)
```
tiny nit -- the indentation of the closing parentheses no longer matches

hughy (approved):
Looks good, thanks!

ironfish-cli/src/commands/deposit.ts
```diff
@@ -108,15 +108,16 @@ export default class Bank extends IronfishCommand {
     const displayAmount = displayIronAmountWithCurrency(IRON_TO_SEND, true)
     const displayFee = displayIronAmountWithCurrency(feeInIron, true)
     const displayNewBalance = displayIronAmountWithCurrency(newBalance, true)
-    if (!flags.confirm) {
-      this.log(`
-You are about to send ${displayAmount} plus a transaction fee of ${displayFee} to the Iron Fish deposit account.
+
+    this.log(`
+Your balance is ${confirmedBalance} and you are about to send ${displayAmount} plus a transaction fee of ${displayFee} to the Iron Fish deposit account.
```
hughy, line 113: One other suggestion! It would be nice if `confirmedBalance` was formatted the same way that the other values are ☝️

hughy (changes requested):

ironfish-cli/src/commands/deposit.ts
hamza-siddiq: @hughy done. Corrected the format too!

hughy: @hamza-siddiq , I meant that the other values use `displayIronAmountWithCurrency`

That way a confirmed balance of $IRON 1 will be displayed as `$IRON 1.00000000 ($ORE 100,000,000)` instead of `100000000`

hamza-siddiq: Aah, I see. I'll work on it

hamza-siddiq: @hughy done. Thanks for the suggestion. It should be good now

danield9tqh:
I don't think it really makes sense to have the same text for both cases. Phrases like `you are about to send` `* This action is NOT reversible *` don't make sense if the send is automatic. I think we should just add the before and after balance after the transaction has been created [here](https://github.com/iron-fish/ironfish/blob/5441d5953ea728aa2274f1a8b6a5c9a7cee382bf/ironfish-cli/src/commands/deposit.ts#L174). Sorry for going back and forth on this change request

hamza-siddiq:
No worries @danield9tqh . That's a really good suggestion. I'll fix it up.

ironfish-cli/src/commands/deposit.ts
```diff
@@ -167,12 +170,12 @@ The memo will contain the graffiti "${graffiti}".
...
+Depositing ${displayAmount} from ${
         transaction.fromAccountName
       }
 Transaction Hash: ${transaction.hash}
-Transaction fee: ${displayIronAmountWithCurrency(feeInIron, true)}
-
+Transaction fee: ${displayFee}
+New Balance: ${displayNewBalance}
 Find the transaction on https://explorer.ironfish.network/transaction/${
         transaction.hash
```
lwisne, line 180: clean up the lint - looks like you may have something off in your tab/space settings

ironfish-cli/src/commands/deposit.ts
```diff
@@ -167,12 +170,12 @@ The memo will contain the graffiti "${graffiti}".
...
       const transaction = result.content
       this.log(`
-Depositing ${displayIronAmountWithCurrency(IRON_TO_SEND, true)} from ${
+Depositing ${displayAmount} from ${
         transaction.fromAccountName
       }
 Transaction Hash: ${transaction.hash}
-Transaction fee: ${displayIronAmountWithCurrency(feeInIron, true)}
-
+Transaction fee: ${displayFee}
```
lwisne, line 177: using these vars that were already set is a nice cleanup

hamza-siddiq: Thank you @lwisne :)

ironfish-cli/src/commands/deposit.ts
hamza-siddiq: @lwisne done. Cleaned it up!

lwisne (approved):

lwisne:
Approved! 

As mentioned with other PRs, try to always give before and after command output when changing CLI commands. Makes the approval process much easier and faster.

hamza-siddiq:
@lwisne I went ahead and added the before and after of the CLI as well. I'll keep all of your teams suggestions in mind. They have been very helpful 👍

hughy (approved):

ironfish-cli/src/commands/deposit.ts
```diff
@@ -105,11 +105,14 @@ export default class Bank extends IronfishCommand {
 
     const newBalance = oreToIron(confirmedBalance - ironToOre(IRON_TO_SEND) - fee)
 
+    const displayConfirmedBalance = displayIronAmountWithCurrency(confirmedBalance, true)
```
hughy, line 108: This PR looks good, but there's one more thing to be wary of here -- `confirmedBalance` is in $ORE, so this will lead to incorrect output:
```
Your balance is $IRON 100,000,000.00000000 ($ORE 10,000,000,000,000,000).
You are about to send $IRON 0.10000000 ($ORE 10,000,000) plus a transaction fee of $IRON 0.00000001 ($ORE 1) to the Iron Fish deposit account.
Your remaining balance after this transaction will be $IRON 0.89999999 ($ORE 89,999,999).
The memo will contain the graffiti "test".
```

lwisne (changes requested):
We'll wait on that after output given Hugh's note.

ironfish-cli/src/commands/deposit.ts
hughy: Nice! I see your latest change just fixed this!

hamza-siddiq:
Guys don't merge it yet, i'm testing the final change, just come back in like 15 mins

lwisne:
I won't be merging until we get that output. I assume it says WIP because it doesn't work as advertised yet.

hamza-siddiq:
@lwisne it did. I just wanted to clean somethings up. I just added an everything to the description. Please have a look at the update which has everything

lwisne (approved):

lwisne (approved):

lwisne:
Merging in now.

Final piece of feedback: Throwing in an additional couple commits after everyone looks through the code, calling them "cleanup", but having them actually change the logic gating the deposit block, is SUPER dangerous. Please refrain from doing something like that in the future - commit messages should be descriptive and help the reviewers know where to look for possible significant changes.

PR #1981, opened 2022-08-08
add telemetry for transaction creation

danield9tqh:
## Summary
In order to track transaction propagation we want to know when the transaction was created. Right now we log transaction propagation but not transaction creation. This means that in a certain timeframe (3hr) our best estimate for when to start looking at propagation time is the first event we see for each transaction. This is not always accurate and we could be starting in the middle of the timeframe for some transactions. Knowing when the transaction was actually first created would be useful for these queries.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/telemetry/telemetry.ts
+++ b/ironfish/src/telemetry/telemetry.ts
@@ -354,6 +354,37 @@ export class Telemetry {
     })
   }
 
+  submitNewTransactionCreated(transaction: Transaction, seenAt: Date): void {
+    const hash = transaction.hash()
+
+    if (!this.shouldSubmitTransaction(hash)) {
+      return
+    }
+
+    this.submit({
+      measurement: 'transaction_created',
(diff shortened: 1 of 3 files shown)


dguenther (approved):

ironfish/src/telemetry/telemetry.ts
```diff
@@ -354,6 +354,37 @@ export class Telemetry {
...
+      fields: [
+        {
+          name: 'notes',
+          type: 'integer',
+          value: transaction.notesLength(),
+        },
+        {
+          name: 'spends',
+          type: 'integer',
+          value: transaction.spendsLength(),
```
NullSoldier, line 382: If we submit these, should we remove these points for `transaction_propagation`?

danield9tqh: Yea I think we can. I will do that

PR #1982, opened 2022-08-08

hamza-siddiq:
## Summary
The default timeout value of jest is set to only 5 seconds. This period seems insufficient for some tests that apparently take
longer to execute. This PR changes the timeout to 32 seconds, which should suffice for the current tests. An improvement would be to make this value get set dynamically according to the specs of the computer running the tests.

This is a continuation of a previous PR: https://github.com/iron-fish/ironfish/pull/1895 . For some reason I could not reopen it. Come back here after having a look at that. 

So turns out that the way I solved it is really the recommended way of solving such issues:

IMG-20220808-WA0015
This is from the official documentation of "jest" (the framework used to write unit-tests)
Here is the documentation, if you'd like to have a look at it:
https://jest-bot.github.io/jest/docs/troubleshooting.html

Let me know if this solves the issue. If not, i'll try to work on another way

## Testing Plan
Tested successfully

## Breaking Change
No
```
[ ] Yes
```
graffiti: ironfishup

### DIFF
--- a/config/jest.config.base.js
+++ b/config/jest.config.base.js
@@ -6,4 +6,9 @@ module.exports = {
   coverageProvider: "v8",
   coverageReporters: ["text-summary", "json", "clover", "text"],
   testPathIgnorePatterns: ['.*\\.test\\.slow\\.ts$', '.*\\.test\\.perf\\.ts$'],
+
+  // set timeout to a higher value than the default (5000), in order to avoid 
+  // timing out when running tests. 32 seconds should suffice for the existing
+  // tests.
+  testTimeout: 32000
 };


wd021:
[This](https://github.com/iron-fish/ironfish/pull/1895#issuecomment-1201942527) is the reason tests are timing out on your local machine. As @danield9tqh [mentioned](https://github.com/iron-fish/ironfish/pull/1895#issuecomment-1201789088), we should probably not change the timeouts globally.

hamza-siddiq:
Thanks @wd021 wo

hamza-siddiq:
Thanks @wd021. But I do know that. The documentation has a different recommendation though. So wanted to double check that with the team

danield9tqh:
Yea that documentation makes sense if we have tests that are long running that we want to keep. In this case though the long running tests are not tests that we want to be long running, they are long running because we haven't gotten around to making them better yet. The real fix is to make those specific tests faster or remove them. If we increase the timeout for all tests globally we will miss other cases where tests are running too long. I will close this PR for that reason

PR #1999, opened 2022-08-09
Migrate transaction sequence for Wallet 2.0

NullSoldier:
## Summary
Migrate TransactionValue.sequence which we need to calculate from the Blockchain data so I'm freezing the headers store here.

## Testing Plan
Run the migration and check for crashes.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/migrations/data/013-wallet-2.ts
+++ b/ironfish/src/migrations/data/013-wallet-2.ts
@@ -11,7 +11,7 @@ import { Transaction } from '../../primitives'
 import { NoteEncrypted } from '../../primitives/noteEncrypted'
 import { DatabaseStoreValue, IDatabase, IDatabaseTransaction } from '../../storage'
 import { createDB } from '../../storage/utils'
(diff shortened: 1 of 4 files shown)


ironfish/src/migrations/data/013-wallet-2.ts
```diff
@@ -242,17 +255,69 @@ export class Migration013 extends Migration {
...
+        const header = await headersStoreOld.get(blockHash)
+
+        if (header) {
+          migrated.sequence = header.header.sequence
+          countMigrated++
+        } else {
+          logger.debug(
+            `\tDropping TX ${transactionHashHex}: block not found ${transactionValue.blockHash}`,
+          )
+          await transactionsStoreOld.del(transactionHash, tx)
```
hughy, line 293: Doesn't look like this is needed -- all transactions get deleted from `transactionsStoreOld` below

hughy (approved):

PR #2000, opened 2022-08-09
add node start time to telemetry

danield9tqh:
## Summary
For things like memory usage we want to know how long nodes have been running for. 

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/telemetry/telemetry.ts
+++ b/ironfish/src/telemetry/telemetry.ts
@@ -33,6 +33,7 @@ export class Telemetry {
   private readonly localPeerIdentity: Identity
 
   private started: boolean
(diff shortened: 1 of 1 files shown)


mat-if:
Should we use a time delta or something? Taking the system time is unreliable, it can be set to anything so we can't necessarily assume how long a system has been running

dguenther (approved):

danield9tqh:
Closing because we are going to use the [node started](https://github.com/iron-fish/ironfish/blob/8c6d2dd03ca9154aea996b5854f518301560935b/ironfish/src/telemetry/telemetry.ts#L283) log to determine this instead

PR #2001, opened 2022-08-09
Allow submitting telemetry without fields

danield9tqh:
## Summary
There was a bug introduced in https://github.com/iron-fish/ironfish/pull/1981. When trying to submit without fields we throw an error. However we should check to see if fields are empty after we have concatenated default fields.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/telemetry/telemetry.ts
+++ b/ironfish/src/telemetry/telemetry.ts
@@ -228,16 +228,16 @@ export class Telemetry {
       return
     }
 
-    if (metric.fields.length === 0) {
-      throw new Error('Cannot submit metrics without fields')
-    }
-
     let tags = this.defaultTags
     if (metric.tags) {
       tags = tags.concat(metric.tags)
     }
 
     const fields = this.defaultFields.concat(metric.fields)
+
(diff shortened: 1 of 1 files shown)


ironfish/src/telemetry/telemetry.ts
```diff
@@ -228,17 +228,17 @@ export class Telemetry {
...
     let tags = this.defaultTags
     if (metric.tags) {
       tags = tags.concat(metric.tags)
     }
 
     const fields = this.defaultFields.concat(metric.fields)
 
+    if (fields.length === 0) {
+      throw new Error('Cannot submit metrics without fields')
+    }
```
NullSoldier, line 240: Is there a measurement we want to send the default fields up without any other fields?

danield9tqh: Yes, the [`transaction_propagation`](https://github.com/iron-fish/ironfish/blob/8c6d2dd03ca9154aea996b5854f518301560935b/ironfish/src/telemetry/telemetry.ts#L404)

dguenther (approved):

PR #2002, opened 2022-08-09
Add the ability to exclude the chain DB in backup

NullSoldier:
## Summary

This let's you only upload accounts if you want. This also fixes matching the excludes as patterns which is how the built in CLI excludes works.

## Testing Plan

Run `ironfish backup --no-mined --no-chain --acounts`

## Breaking Change

```
[ ] Yes
```


mat-if (approved):

hughy (approved):
beyond the scope of this PR I think, but it might be useful in the future to exclude files like `internal.json` and `hosts.json` during backup

NullSoldier:
You might want those files if you want to reproduce an issue someone is running into fully.

Issue #1997, opened 2022-08-09
We don't show expiration sequence by default in accounts:transactions


lwisne:
### What happened?

Add this info.

### Version

0.1.43

danield9tqh:
Fixed in https://github.com/iron-fish/ironfish/pull/2005

PR #2005, opened 2022-08-09
Fix accounts:transactions to show expiration by default

hamza-siddiq:
## Summary
The expiration was set to extended which requires the accounts:transactions 
command to be flaged with -x to show expiration sequence.

Fixes: https://github.com/iron-fish/ironfish/issues/1997

## Testing Plan
Tested successfully

## Breaking Change
No
```
[ ] Yes
```
graffiti: ironfishup


danield9tqh:
Looks good thanks!

danield9tqh (approved):

Issue #1879, opened 2022-07-29
Feature: Dynamic fees on deposit + pay


lwisne:
### Description

The `deposit` command defaults to 1 $ORE as the fee. However, miners are increasingly not picking up txs that only have 1 $ORE fee. You have to set this a bit higher.

Ideally, the default fee would have a value that adjusts based on recent transactions. So if you look at the last x hours of transactions, pick a fee in the 20th percentile or so. This would help ensure people who just use the deposit without knowing fee mechanics would see their txs go through.

wd021:
As a first step, https://github.com/iron-fish/ironfish/pull/1977 adds a new `ironfish fees` command to get a better look into fees for the last N blocks. 

Next, some changes to deposit + pay -> https://github.com/iron-fish/ironfish/issues/1978

PR #2009, opened 2022-08-10
Add support for dynamic fees

wd021:
## Summary
Add route to support dynamic fees for https://github.com/iron-fish/ironfish/issues/1879. Adding a fees command in the CLI, where you can get the 25,50,75 percentile distributions of transaction fees for the last N blocks.

Examples
```
wd021@DWs-MacBook-Pro ironfish-cli % yarn start fees 
Fee distribution for last 10 blocks: (128433 - 128443)
p25: 36 ORE
p50: 40 ORE
p75: 40 ORE
✨  Done in 1.45s.
```
```
wd021@DWs-MacBook-Pro ironfish-cli % yarn start fees 100
Fee distribution for last 100 blocks: (128343 - 128443)
p25: 20 ORE
p50: 36 ORE
p75: 40 ORE
✨  Done in 1.62s.
note: working off account-migration branch
```
## Testing Plan
- test new command yarn start fees
- yarn test and check new tests

## Breaking Change

```
[ ] Yes
```


ironfish/src/utils/percentile.ts
```diff
@@ -0,0 +1,49 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
danield9tqh, line 1: This looks like a lot of sorting logic as well. If this is not going to be used in any performance critical paths then maybe just something like this would suffice and be easier to maintain
```typescript
function percentile(data: number[], percentile: number):  number | null {
	const sorted = data.slice()
	sorted.sort()
	const k = Math.ceil((data.length * percentile) / 100) - 1
	return data[k]
}
```

I think for getting multiple percentiles in a row it could be faster to sort the whole thing upfront too before getting the kth element (if performance is an issue). Don't think that's necessary though

lwisne: definitely agree here. For simplicity and optimization, would prefer to see an approach in which you just sort all the numbers, calculate the 25/50/75 percentile indices, then just pick from your sorted array. 

Wouldn't really even need its own util, and could be a function in the main RPC code.

lwisne (commented):
Really excited to have this! Have a few logic tightening pieces you could clean up.

ironfish/src/utils/percentile.ts
wd021: yea good point, shouldn't be sorting multiple times. removed util and simplified this function.

ironfish/src/rpc/routes/fees/getFees.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    const latestBlockHeader = node.chain.latest
+    let latestBlock = await node.chain.getBlock(latestBlockHeader.hash)
+    Assert.isNotNull(latestBlock, 'No block found')
+
+    const fees: number[] = []
+    const endBlock = latestBlockHeader.sequence
+    const startBlock = endBlock - numOfBlocks
+
+    latestBlock.transactions.forEach((transaction) => {
```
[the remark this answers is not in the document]
wd021: yup, have tightened this up.

wd021:
cleaned up

ironfish/src/rpc/routes/fees/getFees.ts
```diff
@@ -0,0 +1,85 @@
...
+  } else {
+    return fees[base]
+  }
+}
+
+router.register<typeof GetFeesRequestSchema, GetFeesResponse>(
+  `${ApiNamespace.fees}/getFees`,
+  GetFeesRequestSchema,
+  async (request, node): Promise<void> => {
+    const numOfBlocks = request.data.numOfBlocks - 1
```
lwisne, line 49: nit - not a huge fan of this, as it creates a misleadingly named var.

You use this var for `startBlock` below, and have a `<=` in the loop. 

Instead, you can do a `-1` for startBlock and a `<` for the loop.

lwisne (approved):

ironfish/src/rpc/routes/fees/getFees.ts
wd021: makes sense, Assert needs to be `-1` too

PR #2017, opened 2022-08-10
Add a start:debug command for inspecting the node process

mat-if:
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/package.json
+++ b/ironfish-cli/package.json
@@ -45,6 +45,7 @@
     "start:dev": "node start",
     "start": "yarn build && yarn start:js",
     "start:js": "cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc bin/run",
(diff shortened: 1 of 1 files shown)


ironfish-cli/package.json
```diff
@@ -45,6 +45,7 @@
     "start:dev": "node start",
     "start": "yarn build && yarn start:js",
     "start:js": "cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc bin/run",
+    "start:debug": "yarn build && cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --inspect bin/run",
```
dguenther, line 48: Do you want the GC exposed on this too? I'm not sure if it's automatically exposed as part of --inspect

dguenther (approved):

ironfish-cli/package.json
mat-if: I suspect it isn't, but might come in handy for debugging! Added

PR #2024, opened 2022-08-10
updates wallet.getNotes to read from account's decryptedNotes

hughy:
## Summary

each account stores decrypted notes that it owns. leveraging this store saves
some decryption time when collecting notes for the 'accounts:notes' command.
it's still necessary to iterate over transactions to decrypt notes where the
account is the spender since these are not stored on the account.

## Testing Plan

- run `accounts:notes > old.txt`
- make/checkout changes
- run `accounts:notes > new.txt`
- compare output with `diff old.txt new.txt`

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/accounts/getNotes.ts
+++ b/ironfish/src/rpc/routes/accounts/getNotes.ts
@@ -10,10 +10,10 @@ export type GetAccountNotesRequest = { account?: string }
 export type GetAccountNotesResponse = {
   account: string
   notes: {
-    spender: boolean
     amount: number
     memo: string
     noteTxHash: string
+    spent: boolean
   }[]
 }
 
@@ -30,10 +30,10 @@ export const GetAccountNotesResponseSchema: yup.ObjectSchema<GetAccountNotesResp
       .array(
         yup
           .object({
-            spender: yup.boolean().defined(),
             amount: yup.number().defined(),
             memo: yup.string().trim().defined(),
             noteTxHash: yup.string().defined(),
+            spent: yup.boolean().defined(),
           })
           .defined(),
       )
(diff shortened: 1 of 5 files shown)


ironfish/src/wallet/account.ts
```diff
@@ -131,6 +131,27 @@ export class Account {
...
+    amount: number
+    memo: string
+    noteTxHash: string
+    spent: boolean
+  }[] {
+    const notes = []
+
+    for (const decryptedNote of this.decryptedNotes.values()) {
+      const note = new Note(decryptedNote.serializedNote)
+      notes.push({
```
NullSoldier, line 144:
```suggestion

      notes.push({
```

ironfish/src/wallet/account.ts
```diff
@@ -131,6 +131,27 @@ export class Account {
...
+        amount: Number(note.value()),
```
NullSoldier, line 145: this should match getUnspentNotes()

NullSoldier (approved):

ironfish/src/rpc/routes/accounts/getNotes.ts
```diff
@@ -46,7 +46,18 @@ router.register<typeof GetAccountNotesRequestSchema, GetAccountNotesResponse>(
...
-    const { notes } = node.accounts.getNotes(account)
-    request.end({ account: account.displayName, notes })
+    const notes = account.getNotes()
+    const responseNotes = []
+
+    for (const note of notes) {
+      responseNotes.push({
+        amount: Number(note.note.value()),
+        memo: note.note.memo().replace(/\x00/g, ''),
+        noteTxHash: note.transactionHash.toString('hex'),
```
NullSoldier, line 56: In the future this can be noteTxHash

PR #2027, opened 2022-08-10
Call build at top level when invoking the app

mat-if:
## Summary

We often run into issues where trying to (re-)start the app doesn't re-build all of the apps. Triggering the top level workspace build, we invoke build on all of the apps as expected. This will fix the issues with, example, rust changes not getting picked up when trying to start the app

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/package.json
+++ b/ironfish-cli/package.json
@@ -43,7 +43,7 @@
     "lint": "tsc -b && eslint --ext .ts,.tsx,.js,.jsx src/",
     "lint:fix": "tsc -b && eslint --ext .ts,.tsx,.js,.jsx src/ --fix",
     "start:dev": "node start",
-    "start": "yarn build && yarn start:js",
+    "start": "yarn --cwd ../ build && yarn start:js",
     "start:js": "cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc bin/run",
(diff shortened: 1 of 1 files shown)


ironfish-cli/package.json
```diff
@@ -43,7 +43,7 @@
     "lint": "tsc -b && eslint --ext .ts,.tsx,.js,.jsx src/",
     "lint:fix": "tsc -b && eslint --ext .ts,.tsx,.js,.jsx src/ --fix",
     "start:dev": "node start",
-    "start": "yarn build && yarn start:js",
+    "start": "yarn --cwd ../ build && yarn start:js",
```
NullSoldier, line 46: Man this is really nice for fixing end user problems, but it does make running it a lot slower. I'll have to edit this out when developing myself. I think a lot of nodes-guru people who run this for all their commands will also get a lot slower.

mat-if: That is a trade-off, for sure. I don't think I re-run commands often enough that the slowdown will have much affect on me, but I suppose everyone's workflow is different. I'm fine not merging this if we prefer.

mat-if:
I'm going to close this for now, I think the issue isn't running yarn from root, but rather some bug with the way the rust-nodejs repo is working with yarn

Issue #1924, opened 2022-08-03
Validate config files


lwisne:
### What happened?

See comment below:

https://github.com/iron-fish/ironfish/issues/1675#issuecomment-1166271085_

It appears we don't validate config file settings, which can lead to odd issues (or maybe even a security hole if the settings are used in certain ways).

neuronull:
Hello! I am interested in addressing this.

aditya-manit:
Looks like this is already been addressed, we should close this issue I think

Issue #1675, opened 2022-06-25
Error: listen EACCES: permission denied  /home/ubuntu/.ironfish/ironfish.ipc


juniormayhe:
### Description

After installing, setting up and starting ironfish, it throws an error (personal info suppressed with ***):
```
Version       0.1.36 @ 6dfc736
Node Name     ***
Graffiti      ***
Peer Identity ***
Peer Agent    ironfish/0.1.36/6dfc736
Peer Port     9033
Bootstrap     test.bn1.ironfish.network

WebSocket server started at :::9033
    Error: listen EACCES: permission denied  /home/ubuntu/.ironfish/ironfish.ipc
    Code: EACCES
```

The ironfish.ipc file or reference doesn't exist in that folder, all other files are there
```
total 32
drwxrwxrwx 5 ubuntu ubuntu 4096 Jun 25 11:51 .
drwxr-x--- 7 ubuntu ubuntu 4096 Jun 25 11:51 ..
drwxrwxrwx 4 ubuntu ubuntu 4096 Jun 25 11:32 accounts
-rwxrwxrwx 1 ubuntu ubuntu 1472 Jun 25 11:51 config.json
drwxrwxrwx 3 ubuntu ubuntu 4096 Jun 25 11:32 databases
-rw-rw-r-- 1 ubuntu ubuntu    2 Jun 25 11:32 hosts.json
drwxrwxrwx 3 ubuntu ubuntu 4096 Jun 25 11:32 indexes
-rwxrwxrwx 1 ubuntu ubuntu  181 Jun 25 11:51 internal.json
```
it should show something like
```
drwxrwxr-x 5 junior junior 4096 Jun 23 14:46 .
drwxr-x--- 8 junior junior 4096 Jun 23 14:45 ..
drwxrwxr-x 3 junior junior 4096 Jun 15 07:41 accounts
-rw-rw-r-- 1 junior junior 1463 Jun  6 11:30 config.json
drwxrwxr-x 3 junior junior 4096 Jun 15 07:41 databases
-rw-rw-r-- 1 junior junior 6821 Jun 23 20:06 hosts.json
drwxrwxr-x 3 junior junior 4096 Jun 15 07:41 indexes
-rw-rw-r-- 1 junior junior  181 Jun 23 14:46 internal.json
srwxrwxr-x 1 junior junior    0 Jun 23 14:46 ironfish.ipc
```

current settings:
```
[...]

PR #2028, opened 2022-08-10
Validate config file settings

hamza-siddiq:
## Summary

Fixes: https://github.com/iron-fish/ironfish/issues/1924
This commit adds validation to ConfigOptions and enables for additional
validations

## Testing Plan
Tested successfully

## Breaking Change
No
```
[ ] Yes
```
graffiti: ironfishup


ironfish/src/fileStores/config.ts
```diff
@@ -240,8 +240,66 @@ export type ConfigOptions = {
...
+    enableRpcTcp: yup.boolean(),
+    enableRpcTls: yup.boolean(),
+    enableSyncing: yup.boolean(),
+    enableTelemetry: yup.boolean(),
+    enableMetrics: yup.boolean(),
+    getFundsApi: yup.string(),
+    ipcPath: yup.string(),
+    miningForce: yup.boolean(),
+    logPeerMessages: yup.boolean(),
+    // Already got a pareser to validate
```
danield9tqh, line 261: nit: typo `pareser`

ironfish/src/fileStores/config.ts
```diff
@@ -240,8 +240,66 @@ export type ConfigOptions = {
...
+    logLevel: yup.string(),
+    logPrefix: yup.string().matches(RegExp('^([^%]*%(time|level|tag)%[^%]*)*$')),
```
danield9tqh, line 263: I think we should probably not require this regex here. If this regex gets out of sync with our logic to parse the logPrefix then it could prevent people from running their node. I think `yup.string()` is probably enough here

danield9tqh:
I think this PR is useful but it doesn't fix the original issue that was opened. The original issue references that the [`ipcPath` config had a leading space](https://github.com/iron-fish/ironfish/issues/1675#issuecomment-1166271085) and so was invalid. This just checks if `ipcPath` is a string and so it will still pass the validation.

ironfish/src/fileStores/config.ts
hamza-siddiq: I think the benefit justifies the use of a regex. Plus, it's only one place to change should the logic evolve in the future.

lwisne:
Agreed with Daniel - this change will require maintenance and doesn't even fix the issue called out in https://github.com/iron-fish/ironfish/issues/1675.  Going to close.

PR #2031, opened 2022-08-10
Nullifier to note to buffer

NullSoldier:
## Summary
This is the first step in removing hex strings from wallet 2.0. We don't ever want to store hex strings.

## Testing Plan
Run wallet 2.0 migrations and check the balance.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/database/accountsdb.ts
+++ b/ironfish/src/wallet/database/accountsdb.ts
@@ -8,8 +8,8 @@ import { FileSystem } from '../../fileSystems'
 import { Transaction } from '../../primitives/transaction'
 import {
   BigIntLEEncoding,
-  BufferEncoding,
   BUFFER_ENCODING,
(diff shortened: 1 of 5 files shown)


ironfish/src/wallet/database/accountsdb.ts
```diff
@@ -9,6 +9,7 @@ import { Transaction } from '../../primitives/transaction'
 import {
   BigIntLEEncoding,
   BUFFER_ENCODING,
```
hughy, line 11: `BUFFER_ENCODING` is an instance of `BufferEncoding`. Could we do without it?

hughy (approved):

ironfish/src/wallet/database/accountsdb.ts
NullSoldier: Removed

PR #2032, opened 2022-08-11
loads decryptedNotes without computing unconfirmed balance

hughy:
## Summary

unconfirmed balance is updated every time the decrypted notes store is updated,
so it isn't necessary to recompute unconfirmed balance.

calculating the balance on load has the benefit of ensuring that the balance
matches the decrypted notes store when the node starts, but we read the balance
from the database each time we use it rather than tracking it in memory, so
we're already relying on updating the balance in tandem with the notes.

## Testing Plan

- imported account on `staging`, ran `accounts:balance`
- checked out `feature/no-balance-on-load`, ran wallet 2.0 migration
- ran `accounts:balance`, verified balances unchanged following migration

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/database/accountsdb.ts
+++ b/ironfish/src/wallet/database/accountsdb.ts
@@ -321,15 +321,14 @@ export class AccountsDB {
     })
   }
 
-  async *loadDecryptedNotes(): AsyncGenerator<{
-    hash: string
-    decryptedNote: DecryptedNoteValue
-  }> {
+  async loadDecryptedNotes(
+    map: Map<
+      string,
+      { nullifierHash: string | null; noteIndex: number | null; spent: boolean }
+    >,
+  ): Promise<void> {
     for await (const [hash, decryptedNote] of this.decryptedNotes.getAllIter()) {
-      yield {
-        hash,
-        decryptedNote,
-      }
+      map.set(hash, decryptedNote)
     }
   }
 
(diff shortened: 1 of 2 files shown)


ironfish/src/wallet/database/accountsdb.ts
```diff
@@ -321,15 +321,14 @@ export class AccountsDB {
...
   }
 
-  async *loadDecryptedNotes(): AsyncGenerator<{
-    hash: string
-    decryptedNote: DecryptedNoteValue
-  }> {
+  async loadDecryptedNotes(
+    map: Map<
+      string,
+      { nullifierHash: string | null; noteIndex: number | null; spent: boolean }
```
NullSoldier, line 327: Should this be `DecryptedNoteValue`?

hughy:
`account.reset` currently sets unconfirmed balance to 0 in the database, so recalculating on load may be necessary

ironfish/src/wallet/database/accountsdb.ts
```diff
@@ -321,15 +321,14 @@ export class AccountsDB {
...
+    >,
+  ): Promise<void> {
     for await (const [hash, decryptedNote] of this.decryptedNotes.getAllIter()) {
-      yield {
-        hash,
-        decryptedNote,
-      }
+      map.set(hash, decryptedNote)
```
hughy, line 331: We need to filter decrypted notes by account ID when adding to the map

hughy:
Closing this -- I think it's fine to recalculate the balance when loading the decrypted notes since we have to iterate over all of the notes during load anyway.

A different change, later on, would be to avoid holding the decrypted notes in memory

PR #2036, opened 2022-08-11
show mempool memory usage in status command

danield9tqh:
## Summary
This can be helpful for determining how much memory the mempool is using. We are importing a lot of transactions now and we may want to give users the option of limiting their mempool size. The `sizeBytes` function will help with that. Also adding it to the status command we can watch out for memory hungry mempools

## Testing Plan
Local testing and unit tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/memPool/memPool.ts
+++ b/ironfish/src/memPool/memPool.ts
@@ -18,7 +18,11 @@ interface MempoolEntry {
 
 export class MemPool {
   private readonly transactions = new BufferMap<Transaction>()
+  /* Keep track of number of bytes stored in the transaction map */
+  private transactionsBytes = 0
   private readonly nullifiers = new BufferMap<Buffer>()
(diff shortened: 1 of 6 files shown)


ironfish/src/memPool/memPool.ts
```diff
@@ -54,6 +58,11 @@ export class MemPool {
     return this.transactions.size
   }
 
+  sizeBytes(): number {
+    const queueSize = this.size() * 32 // estimate the queue size
```
hughy, line 62: Should this be `this.queue.size`?

danield9tqh: Yea I will change it. I thought it was a generator for some reasons and didn't have a size but it does

hughy (approved):

dguenther (approved):

PR #2037, opened 2022-08-11
Drop transactions if nullifiers already exist in the chain

dguenther:
## Summary

Changes onNewTransaction to drop transactions that have nullifiers already in the chain. On my machine, this reduced the mempool to less than 1/10 the size, from >3k transactions to <200 transactions.

It could be useful to have more complex behavior that keeps these double-spent transactions around in case of forking, to potentially allow miners replace transactions with higher-fee transactions using the same spend. We could also consider adding a confirmation window to discarding the transactions, but in the general case where the chain is progressing forward, it should be safe to discard them.

## Testing Plan

Added a test to verify the behavior.

## Breaking Change

```
[ ] Yes
```


mat-if (approved):
👏

ironfish/src/network/peerNetwork.ts
```diff
@@ -1008,6 +1008,23 @@ export class PeerNetwork {
       // Force lazy deserialization of the transaction as a first sanity check
       const transaction = this.chain.verifier.verifyNewTransaction(message.transaction)
 
+      // Next, check the spends are valid and not already in the chain
+      const spendVerificationResult = await this.chain.verifier.verifyTransactionSpends(
```
danield9tqh, line 1012: It looks like this will return `VerificationResultReason.ERROR` (from `pasRoot`) if the spend size is greater than the current merkle tree size. I think this means that we are seeing a transaction before we're seeing the block of the note it was created in. I was running this and seeing a lot of these errors. Although there were a lot of double spends, this error was the majority. 

Does it make sense to ignore those transactions? They would not currently be valid for the node but if we eventually catch up to the block they were created from they could become valid. Maybe we want those in the mempool. For compact blocks this would help especially

danield9tqh (changes requested):

ironfish/src/consensus/verifier.ts
```diff
@@ -148,29 +148,53 @@ export class Verifier {
...
+      return { valid: false, reason: VerificationResultReason.DESERIALIZATION }
+    }
+
+    // Next, check the spends are valid and not already in the chain
+    const reason = await this.chain.db.withTransaction(null, async (tx) => {
+      const nullifierSize = await this.chain.nullifiers.size(tx)
+      const notesSize = await this.chain.notes.size(tx)
+
+      for (const spend of transaction.spends()) {
+        if (await this.chain.nullifiers.contained(spend.nullifier, nullifierSize, tx)) {
```
danield9tqh, line 170: I feel like we could probably reuse `verifySpend` here since it's a lot of the same logic. Maybe we can add a new verification failure for the spend.size being less than and return it in `verifySpend` then just ignore those errors. Wdyt?

ironfish/src/network/peerNetwork.ts
```diff
@@ -998,34 +998,33 @@ export class PeerNetwork {
...
       if (!valid) {
         Assert.isNotUndefined(reason)
-        this.logger.debug(
-          `Invalid transaction '${transaction.unsignedHash().toString('hex')}': ${reason}`,
-        )
+        // Logging hash because unsignedHash is slow
+        this.logger.debug(`Invalid transaction '${hash.toString('hex')}': ${reason}`)
+        // TODO: It might be more accurate to have a separate map of invalid transactions,
+        // but it accomplishes the same outcome of discarding the transaction in the future.
+        this.recentlyAddedToChain.set(hash, true)
```
danield9tqh, line 1018: I think we should probably remove this just to keep the LRU consistent with transactions that are in the chain and not generally invalid transactions

danield9tqh (approved):

PR #2038, opened 2022-08-11
upgrades accounts:transactions to use wallet 2.0 stores

hughy:
## Summary

- removes 'getTransactionsWithMetadata`: uses 'getTransactions' and moves
  parsing logic to rpc route
- uses nullifierToNoteHash to determine if an account created a transaction: if
  the account has the nullifier for for any of the transaction spends, then the
  account created the transaction
- outputs all transactions in transactions store: the account does not track
  transactions for which it did not spend or receive notes
- uses node chain to determine transaction status

- moves getTransactionStatus to utils

- removes 'getTransaction' from wallet.ts

- moves parsing of transaction for 'accounts:transactions -t' to rpc route
- re-uses getTransactionStatus util function
- changes transaction notes output in 'accounts:transactions' to match
  'accounts:notes': removes 'spender' column, only shows owned notes, adds
  'spent' column

## Testing Plan

local testing of cli commands, starting with diff to compare output before/after changes

## Breaking Change

```
[ ] Yes
```


hughy:
This removes outgoing 'spender' notes from the output. We may want to add those back in (same with `accounts:notes`), but we'd reconsider whether to store them in an account's decrypted notes store before

mat-if (approved):

mat-if:
> This removes outgoing 'spender' notes from the output. We may want to add those back in (same with `accounts:notes`), and we may want to add them to the account's decrypted notes store.

Oh yeah, I missed this. I think having the "spender" notes is pretty useful, otherwise you have no way of really seeing your 'bank statement' so to speak. We can add it back in a followup PR though, this is a nice cleanup.

hughy:
> > This removes outgoing 'spender' notes from the output. We may want to add those back in (same with `accounts:notes`), and we may want to add them to the account's decrypted notes store.
> 
> Oh yeah, I missed this. I think having the "spender" notes is pretty useful, otherwise you have no way of really seeing your 'bank statement' so to speak. We can add it back in a followup PR though, this is a nice cleanup.

I'm going to put the 'spender' notes back into `accounts:transactions` and `accounts:notes` in a separate issue: https://linear.app/ironfish/issue/IRO-2473/add-spender-notes-to-output-of-transactions-and-notes-commands

PR #2045, opened 2022-08-15
Add block fetcher and compact block propagation

dguenther:
## Summary

This is unfinished, but will add handlers for new messages added for the new block propagation method, plus a fetcher for retrieving compact blocks and transactions for compact blocks.

Fixes IRO-2229
Fixes IRO-2277
Fixes IRO-2278
Fixes IRO-2280
Fixes IRO-2283
Fixes IRO-2284
Fixes IRO-2318

## Testing Plan

* [x] New tests should be added for the new message handlers
* [x] Smoke test with at least 4 clients

## Breaking Change

This should be backward-compatible with the previous method of block propagation.

```
[ ] Yes
```


ironfish/src/network/peerNetwork.ts
```diff
@@ -657,24 +719,278 @@ export class PeerNetwork {
...
+      }
+    }
+
+    return { ok: true, transactions }
   }
 
-  private handleNewBlockV2Message(peer: Peer, message: NewBlockV2Message) {
-    this.logger.debug(`Received unimplemented message ${message.type}`)
+  private async onNewBlockTransactions(peer: Peer, message: GetBlockTransactionsResponse) {
+    const block = this.blockFetcher.receivedBlockTransactions(message)
```
danield9tqh, line 882: I think maybe we should check `alreadyHaveBlock` here too

ironfish/src/network/peerNetwork.ts
```diff
@@ -657,24 +719,278 @@ export class PeerNetwork {
...
+
+    // verify the header
+    const verifyHeaderResult = this.chain.verifier.verifyBlockHeader(header)
+    if (!verifyHeaderResult.valid) {
+      this.chain.addInvalid(header, verifyHeaderResult.reason ?? VerificationResultReason.ERROR)
+      this.blockFetcher.removeBlock(header.hash)
+      return
+    }
+
+    if ((await this.alreadyHaveBlock(header.hash)) || this.chain.isInvalid(header) !== null) {
```
danield9tqh, line 922: This `chain.isInvalid` sometimes adds blocks to invalid. But we also check if it's invalid in `alreadyHaveBlock`. I feel like we could split up the checking and the marking of invalid so it's more explicit here

ironfish/src/network/peerNetwork.ts
```diff
@@ -921,37 +1241,99 @@ export class PeerNetwork {
...
+    this.blockFetcher.receivedFullBlock(block)
+
+    // Re-gossip the full block
+    const serializedBlock = BlockSerde.serialize(block)
+    const newBlockMessage = new NewBlockMessage(
+      serializedBlock,
+      nonce ?? Buffer.alloc(16, block.header.hash),
+    )
+
+    this.broadcastBlock(newBlockMessage)
```
danield9tqh, line 1302: Is there a reason why we broadcast full blocks and compact blocks here but then broadcast the hashes later on?

dguenther: Yeah, it's basically a balance between optimistic propagation of the block and fully validating the block body + merkle root. Similar to ethereum, but we send compact blocks instead: https://github.com/ethereum/devp2p/blob/master/caps/eth.md#block-propagation

ironfish/src/network/peerNetwork.ts
dguenther: I'll see if I can refactor the isInvalid call to take either a header or a hash, and then propagate that through.

ironfish/src/network/peerNetwork.ts
dguenther: I don't think it's necessary, unless there's an edge case I'm not thinking of. `receivedBlockTransactions` will return null for the block and drop the message if the block state isn't `TRANSACTION_REQUEST_IN_FLIGHT`, and we should be changing the state to `PROCESSING_FULL_BLOCK` in `onNewFullBlock`, or removing the block if it enters the chain, is marked as invalid, or is marked as an orphan.

Actually, this made me realize we don't remove blocks from the fetcher when marking them as an orphan, so I'll do that

ironfish/src/network/peerNetwork.ts
danield9tqh: alright yea that makes sense

ironfish/src/network/peerNetwork.ts
danield9tqh: Yea i was thinking it might make sense to not change the state of invalid in a function that is just called `isInvalid` but I think it is pretty safe here. We know that the input hashes to a certain header so it seems impossible that we'd accidentally mark something as invalid that is valid

ironfish/src/telemetry/telemetry.ts
```diff
@@ -18,7 +18,7 @@ import { Metric } from './interfaces/metric'
 import { Tag } from './interfaces/tag'
 
 export class Telemetry {
-  private readonly FLUSH_INTERVAL = 5 * 60 * 1000
+  private readonly FLUSH_INTERVAL = 1 * 60 * 1000
```
dguenther, line 21: @danield9tqh Does this need to be reverted? I'm guessing this was added in debugging due to the message being blocked at the API level

danield9tqh: Yea thought I did revert that hmm

dguenther: I pushed a revert of this

danield9tqh (approved):

PR #2046, opened 2022-08-15
Add db transactions for all wallet and account operations

mat-if:
## Summary

Add levelup db transactions to all account db operations

This should help stop issues with the data stores becoming out of sync, by explicitly using database transactions anywhere multiple operations happen in the same function

`withTransaction` is a great helper that allows us to either create a transaction or use an existing one. This helps a lot when certain operations on their own might need a transaction, but should also be a part of a transaction at a higher level

Closes IRO-2446

## Breaking Change

```
[ ] Yes
```


hughy (approved):
The only place I could find where we're interacting with the database _without_ a transaction is in `Account.load()` and `loadDecryptedNotesAndBalance`, but I can't think of a scenario where we'd get into trouble without one

ironfish/src/wallet/account.ts
```diff
@@ -118,17 +118,19 @@ export class Account {
...
+    })
   }
 
-  async reset(): Promise<void> {
+  async reset(tx?: IDatabaseTransaction): Promise<void> {
     this.decryptedNotes.clear()
     this.nullifierToNoteHash.clear()
     this.transactions.clear()
-    await this.saveUnconfirmedBalance(BigInt(0))
+    await this.saveUnconfirmedBalance(BigInt(0), tx)
```
NullSoldier, line 133: Is this a separate bug that no save is called here?

NullSoldier (approved):

PR #2048, opened 2022-08-15
defines BufferUtils.toHuman, replaces versions in GraffitiUtils and MemoUtils

hughy:
## Summary

we have several places in our codebase that convert buffers into human-readable
strings using the same logic.

- replaces hardcoded transformation of note memos
- refactors shared logic from GraffitiUtils.toHuman and MemoUtils.toHuman into a
  single function
- changes Note.memo() to convert buffer directly to human-readable string when
  returning

Closes IRO-2475

## Breaking Change

```
[ ] Yes
```


mat-if (approved):

NullSoldier:
At some point we should not render any of these in the RPC. That's some tech debt we need to fix.

hughy:
> At some point we should not render any of these in the RPC. That's some tech debt we need to fix.

Do you mean that we should send buffers over RPC and render on the recipient side?

PR #2051, opened 2022-08-16
Improving pay and deposit commands

wd021:
## Summary
1 - `deposit` and `deposit-all` by default will now use dynamic fee (p25 of last 100 blocks) instead of 1 ORE
```
wd021@DWs-MacBook-Pro ironfish-cli % yarn start deposit-all
You are about to deposit all your $IRON to the Iron Fish deposit account. Each transaction will use a fee of $IRON 0.00000013 ($ORE 13). The memos will contain the graffiti "wd021".
```
```
wd021@DWs-MacBook-Pro ironfish-cli % yarn start deposit

Your balance is $IRON 2.18146315 ($ORE 218,146,315).

You are about to send $IRON 0.10000000 ($ORE 10,000,000) plus a transaction fee of $IRON 0.00000013 ($ORE 13) to the Iron Fish deposit account.
Your remaining balance after this transaction will be $IRON 2.08146302 ($ORE 208,146,302).
The memo will contain the graffiti "wd021".
```

2 - `accounts:pay` will now display dynamic fee, max possible spend in prompt.
```
wd021@DWs-MacBook-Pro ironfish-cli % yarn start accounts:pay

Enter the fee amount in $IRON (min: 0.00000001, dynamic: 0.00000013): 0.00000013
Enter the amount in $IRON (balance available: 2.18146315, max: 2.18146302): 2.18146302

```

## Testing Plan
- test `deposit` `deposit-all` and `accounts:pay` commands

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/accounts/pay.ts
+++ b/ironfish-cli/src/commands/accounts/pay.ts
@@ -3,6 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 import {
+  displayIronAmount,
   displayIronAmountWithCurrency,
   ironToOre,
   isValidAmount,
@@ -77,15 +78,46 @@ export class Pay extends IronfishCommand {
       this.exit(1)
     }
 
-    if (amount == null || Number.isNaN(amount)) {
-      const response = await client.getAccountBalance({ account: from })
+    if (fee == null || Number.isNaN(fee)) {
+      let dynamicFee: string | null
+
+      try {
+        // fees p25 of last 100 blocks
+        dynamicFee = displayIronAmount(
+          oreToIron((await client.getFees({ numOfBlocks: 100 })).content.p25),
+        )
+      } catch {
+        dynamicFee = null
+      }
 
       const input = Number(
         await CliUx.ux.prompt(
-          `Enter the amount in $IRON (balance available: ${displayIronAmountWithCurrency(
+          `Enter the fee amount in $IRON (min: 0.00000001${
+            dynamicFee ? `, dynamic: ${dynamicFee}` : ''
+          })`,
+          {
+            required: true,
+          },
+        ),
+      )
+
+      if (Number.isNaN(input)) {
+        this.error(`A valid fee amount is required`)
+      }
+
+      fee = input
+    }
+
+    if (amount == null || Number.isNaN(amount)) {
+      const response = await client.getAccountBalance({ account: from })
+
+      const maxToSend = oreToIron(Number(response.content.confirmed)) - fee
+
+      const input = Number(
+        await CliUx.ux.prompt(
+          `Enter the amount in $IRON (balance available: ${displayIronAmount(
             oreToIron(Number(response.content.confirmed)),
-            false,
-          )})`,
+          )}, max: ${displayIronAmount(maxToSend)})`,
           {
             required: true,
           },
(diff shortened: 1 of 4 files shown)


ironfish-cli/src/commands/accounts/pay.ts
```diff
@@ -77,41 +78,57 @@ export class Pay extends IronfishCommand {
...
+        dynamicFee = null
+      }
 
       const input = Number(
         await CliUx.ux.prompt(
-          `Enter the amount in $IRON (balance available: ${displayIronAmountWithCurrency(
-            oreToIron(Number(response.content.confirmed)),
-            false,
-          )})`,
+          `Enter the fee amount in $IRON (min: 0.00000001${
```
lwisne, line 95: Any particular reason you're switching order here? I can imagine this may break some scripts.

wd021: i think its useful to have the `max` amount payable in the 'amount' prompt and that has to come after the fee has been inputted. just easier UX for users to not have to do the extra mental math.

ironfish-cli/src/commands/deposit-all.ts
```diff
@@ -216,6 +228,7 @@ export default class DepositAll extends IronfishCommand {
 
   private async verifyCanSend(
```
lwisne, line 229: can we take this opportunity to move this entire function into `utils/currency.ts` and only have it one time for both`deposit` and `depositAll`? As far as I can tell, the code is identical.

ironfish-cli/src/commands/accounts/pay.ts
lwisne: I see the argument here, but I think I would prefer to leave as is. Changing the order of params will likely break some scripts/workflows and not sure it's worth it.

ironfish-cli/src/commands/deposit.ts
```diff
@@ -28,9 +28,8 @@ export default class Bank extends IronfishCommand {
 
   static flags = {
     ...RemoteFlags,
-    fee: Flags.integer({
+    fee: Flags.string({
```
lwisne, line 31: why this change? It will remove some type-checking by the CLI command parser.

wd021: wanted to sync the same types across different files. accounts/pay.ts uses string for the fee flag. I'll revert this, leave this refactoring for another task.

ironfish-cli/src/commands/deposit-all.ts
wd021: will do

ironfish-cli/src/utils/currency.ts
```diff
@@ -0,0 +1,80 @@
...
+  RpcClient,
+  WebApi,
+} from '@ironfish/sdk'
+
+const REGISTER_URL = 'https://testnet.ironfish.network/signup'
+
+export async function verifyCanSend(
+  client: RpcClient,
+  api: WebApi,
+  flags: Record<string, unknown>,
```
lwisne, line 18: A good general tip is to abstract away CLI flags as soon as possible and instead use the values in the structure.

In this case, just having a param `expirationSequenceDelta` and passing that in would be both easier to read and more compartmentalized.

wd021: yea agree, overlooked it.

lwisne (approved):
Thanks for doing this - really think this one will be a HUGE boost in usability!

Issue #2052, opened 2022-08-16
MinersFeeTransaction of forked block always displays pending status 


hairtail:
### What happened?
MinersFeeTransaction of forked block displays pending status.
 [Check transaction on explorer here](https://explorer.ironfish.network/transaction/a30d7a7ece8adde7b007286e944cfd06d2c402b2dd70864a0634e54169f9e43c)
 [Check block on explorer here](https://explorer.ironfish.network/blocks/00000000000125254209525ec49d4a5a348b54575262363802ce3d0b127f81e9) 
```
Transaction: a30d7a7ece8adde7b007286e944cfd06d2c402b2dd70864a0634e54169f9e43c
Status: pending
Miner Fee: ✔
Fee ($ORE): -2000005000
Spends: 0

---Notes---

 Spender Amount ($IRON) Memo
 ─────── ────────────── ────
 x       20.00005
```
```
 Status  Creator Hash                                                       Miner Fee Fee ($ORE)  Notes Spends Expiration
 ─────── ─────── ────────────────────────────────────────────────────────── ───────── ─────────── ───── ────── ──────────
 pending x       15445addcd7c6a5f929c86550fc215e3f87edd7f2fd26a1bbb958f60b… ✔         -2000028666 1     0      0
 pending x       1cee87f922b2351dd001671764b76cfce4bbf259e5993eea9aa9ca4d1… ✔         -2000032166 1     0      0
 pending x       379424053a34c28f9d0ddf258e84ae30e145a37fe432a5dfe63faeb12… ✔         -2000012666 1     0      0
 pending x       4740a53e7f7bb0d5c92818fcba1e0135c8382f9c9564ac7d27cf86551… ✔         -2000046666 1     0      0
 pending x       4b0fbfdb843f3b3954036ffa94ec77ad588045a90dfc2620111b69ae6… ✔         -2000025555 1     0      0
 pending x       4b7770efa26ef09e1940d2e8a5dc1c29a309a00dce333881bf6d3b410… ✔         -2000022671 1     0      0
 pending x       6dd1f0ce7a589f007f90c0e7a305c7474447e38c8fc2668e88be6cd74… ✔         -2000020000 1     0      0
 pending x       70dddfc5ce408b967542c72a17e1cd29412a5875c070b25acb6edc7d5… ✔         -2000000000 1     0      0
[...]

PR #2053, opened 2022-08-16
MinersFeeTransaction status of forked block

hairtail:
## Summary
Fix issue https://github.com/iron-fish/ironfish/issues/2052
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/account/accounts.ts
+++ b/ironfish/src/account/accounts.ts
@@ -1150,6 +1150,8 @@ export class Accounts {
           } else {
             status = 'forked'
           }
+        } else if (transaction.isMinersFee()) {
+          status = 'forked'
         }
 
         transactions.push({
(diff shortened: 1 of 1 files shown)


hairtail:
It's strange that all transactions of forked block should be removed once chain reorganized, add all transactions are supposed to have a blockhash. But I find that minersFeeTransactions of forked block do not have blockhash item. Then this kind of transactions are marked as pending status while they should be in forked status. This is a simple fix to update transaction status. We should figure out the reason leads to the issue mentioned in this PR.

dguenther:
If the `blockHash` field is set to a block hash, it indicates that the transaction is part of the heaviest chain, also known as the main chain or the head chain, and the associated funds are spendable. I think this unnecessary check may have already existed in our code, but if `blockHash` is set, we should assume that the transaction is part of the main chain, and it's not necessary to check `isHeadChain`.

I think it's correct to say that a transaction is `forked` if it's a miner's fee, so this change makes sense -- the transaction will never be reattached to another block. However, if a transaction has a blockHash but isn't part of the main chain, that should be a bug in our Accounts code.

dguenther (approved):

PR #2056, opened 2022-08-16
Clean up routing and ensure different invocations of the node use the…

### LINKED ISSUE #2049: MemoryClient and IpcClient do not share router namespaces
### What happened?

If you add a namespace to our RPCs, the `MemoryClient` picks this up, as it automatically pulls in all namespaces.

However, the `RpcIpcClient` acts off a static list which is supposed to contain every namespace. Fix this so they are in sync and you can't make a new route that works in memory but not in IPC.

Screenshot - I stopped my node between commands. You can see how it fails while the node is running, due to having a different list of valid routes.

Screen Shot 2022-08-15 at 6 01 41 PM

### Version

0.1.44

### Debug output

_No response_

### Relevant log output

_No response_

### Graffiti

_No response_

lwisne:
… same namespaces. 

## Summary
Fixes #2049.

## Testing Plan
Added a number of automated tests to catch the issue if it comes up again.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/adapters/ipcAdapter.test.ts
+++ b/ironfish/src/rpc/adapters/ipcAdapter.test.ts
@@ -124,7 +124,7 @@ describe('IpcAdapter', () => {
 
   it('handles all RPC namespaces', async () => {
     const allowedNamespaces = ALL_API_NAMESPACES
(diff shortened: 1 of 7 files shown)


hughy (approved):
Looks good to me pending lint fixes!

PR #2058, opened 2022-08-16
Force pay and deposit commands to connect to a running node

mat-if:
## Summary

These commands currently will spin up a memory client if a node isn't running, but given the node needs to be connected to a network, it doesn't really make sense to do that. Instead, they should only try connecting to a running node.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/accounts/pay.ts
+++ b/ironfish-cli/src/commands/accounts/pay.ts
@@ -66,7 +66,8 @@ export class Pay extends IronfishCommand {
     const expirationSequence = flags.expirationSequence
     const memo = flags.memo || ''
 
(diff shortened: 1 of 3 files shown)


ironfish-cli/src/commands/accounts/pay.ts
```diff
@@ -66,7 +66,8 @@ export class Pay extends IronfishCommand {
     const expirationSequence = flags.expirationSequence
     const memo = flags.memo || ''
 
-    const client = await this.sdk.connectRpc()
```
hughy, line 69: These could also change to `const client = await this.sdk.connectRpc(false, true)` to force a remote RPC connection

hughy (approved):

PR #2062, opened 2022-08-17
updates account to store note hashes as buffers

hughy:
## Summary

- changes decryptedNotes key encoding to BufferEncoding
- changes decryptedNotes storage of nullifier from string to buffer
- updates wallet-2 migration for decryptedNotes

- changes noteHashesBySequence and nonChainNoteHashes to use BufferSets to store
  note hashes

## Testing Plan

- ran migrations on partially rescanned pool account
- compared output of `accounts:balance`, `accounts:notes`, `accounts:transactions` between current branch and `accounts-performance` branch

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/account.ts
+++ b/ironfish/src/wallet/account.ts
@@ -465,7 +465,7 @@ export class Account {
     for (let i = unconfirmedSequenceStart; i < headSequence; i++) {
       const noteHashes = this.noteHashesBySequence.get(i)
       if (noteHashes) {
-        for (const hash of noteHashes.toKeys()) {
+        for (const hash of noteHashes) {
           const note = this.decryptedNotes.get(hash)
           Assert.isNotUndefined(note)
           if (!note.spent) {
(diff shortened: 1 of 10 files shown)


ironfish/src/wallet/account.ts
```diff
@@ -458,7 +458,7 @@ export class Account {
       }
     }
 
-    for (const noteHash of this.nonChainNoteHashes.values()) {
+    for (const noteHash of this.nonChainNoteHashes.toKeys()) {
```
NullSoldier: This is an existing bug in the typedef files I noticed. I wondered why this doesn't use the iterator. It looks like the typedef file in our repo doesn't match the actual javascript library.

 - https://github.com/iron-fish/ironfish/blob/master/ironfish/src/typedefs/buffer-map.d.ts#L38
 - https://github.com/chjj/buffer-map/blob/master/lib/buffer-map.js#L151-L153

NullSoldier (approved):

NullSoldier (approved):

PR #2063, opened 2022-08-17
Convert transactionValue.blockHash to Buffer

NullSoldier:
## Breaking Change

```
[ ] Yes
```


hughy (approved):
looks good pending lint

PR #2064, opened 2022-08-17
Updated tests: Fixed a bug: error running "ironfish debug"

hamza-siddiq:
## Summary
Fixes: https://github.com/iron-fish/ironfish/issues/1792

So a previous PR was closed where I did not do the tests properly. My sincere apologies for not being thorough with my tests, before opening a PR. This PR adds in the proper tests, and proper explanations.

In the above issue, the bug is that when ironfish debug is run, it only returns the error: "Error: Database is not open. Call IDatabase.open() first". This PR fixes that.

I have also added explanations infront of the code in comments which should make the problem and changes easy to understand.

Old code:
IMG-20220818-WA0033

New code:
IMG-20220818-WA0034

## Testing Plan

Before:
If any error occurs other than DatabaseIsLockedError, DatabaseOpenError, this would be the output of the debug command:
IMG-20220818-WA0029

After:
IMG-20220818-WA0031

## Breaking Change

No
```
[ ] Yes
```


lwisne:
This doesn't fix the issue stated - it just changes the format of the error output.

Closing.

hamza-siddiq:
Well, it does.
Before the fix, any error other than `DatabaseIsLockedError` and `DatabaseOpenError` would crash the command, and prevent prevent it from showing the basic output that does not require an open database. This is obviously a bug that not only does it impact the reliability of the CLI, but also degrades the user experience.  

#### The fix makes sure of the following:
*  The basic output will be shown **regardless** of the error type.
* Catch **every** possible error that might occur.
* Remove the reliance on a flag variable (which is a bad practice)
* Remove redundancy in output printing code

lwisne:
The bug https://github.com/iron-fish/ironfish/issues/1792 is not about the error format.

It's about the fact that the user was running `ironfish debug` and it worked, and now it gives an error. 

You are not addressing their concerns at all.

hamza-siddiq:
The fix is NOT about the format either. If you carefully read the changes and my last reply, you would realize the cause of the bug, and the solution that I applied.
The cause of the bug is that the `catch` block has `if` statements that check only of `DatabaseIsLockedError` and `DatabaseOpenError` errors and set the flag variable `dbOpen` accordingly. If an error of a type other than the types I mentioned earlier was thrown by the `try` block, the `dbOpen` flag variable would still hold the value `true`, which is obviously not right.

PR #2068, opened 2022-08-18
uses unsignedHash in CLI output for transactions, notes

hughy:
## Summary

the block explorer displays unsignedHash from transactions. if we show the
signed hash in CLI output then users won't be able to match the output to the
block explorer.

## Testing Plan

ran `accounts:notes`, `accounts:transactions`, and verified that hashes were found in block explorer

## Breaking Change

```
[ ] Yes
```


hughy:
https://github.com/iron-fish/ironfish/pull/2069 also makes this change, along with updating the storage of transactions by hash

PR #2070, opened 2022-08-18
retries 'rm' on chain database to address EBUSY errors

hughy:
## Summary

Windows users may run into EBUSY when running chain:download as the command
removes their existing chain database. increase the maxRetries from 0 to 10 to
try to get around this.

- sets maxRetries to 10
- users 'force' to ignore errors when the path does not exist instead of
  checking for the path first

## Breaking Change

```
[ ] Yes
```


NullSoldier:
This looks good, but I wonder if this is happening also because the node is running.

NullSoldier (approved):

PR #2072, opened 2022-08-19
use current block for newBlock

hpool-dev:
## Summary
There may be `chain.head` modifications during the call to `createNewBlockTemplate` resulting in incorrect data in the new block template. An `invalid target` exception may occur on call `submitBlock`.

## Testing Plan
No.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -864,6 +864,7 @@ export class Blockchain {
     userTransactions: Transaction[],
     minersFee: Transaction,
     graffiti?: Buffer,
+    headBlock?: Block
   ): Promise<Block> {
     const transactions = [minersFee, ...userTransactions]
     return await this.db.transaction(async (tx) => {
@@ -880,7 +881,7 @@ export class Blockchain {
         previousSequence = 0
         target = Target.maxTarget()
       } else {
-        const heaviestHead = this.head
+        const heaviestHead = headBlock ? headBlock.header : this.head
         if (
           originalNoteSize !== heaviestHead.noteCommitment.size ||
           originalNullifierSize !== heaviestHead.nullifierCommitment.size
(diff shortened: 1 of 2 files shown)


ironfish/src/blockchain/blockchain.ts
```diff
@@ -880,7 +881,7 @@ export class Blockchain {
         previousSequence = 0
         target = Target.maxTarget()
       } else {
-        const heaviestHead = this.head
+        const heaviestHead = headBlock ? headBlock.header : this.head
```
hpool-dev, line 884: Probably `head` was modified again before calling this code.

dguenther: Do you have a way to reproduce this issue consistently? I think it would be better to fix the issue by fetching `head` from the database instead of using `this.head`, like this:

```
        const head = await this.meta.get('head', tx)
        Assert.isNotUndefined(head)
        const heaviestHead = await this.getHeader(head, tx)
        Assert.isNotNull(heaviestHead)
```
Could you verify if that change fixes the issue as well?

dguenther: Actually, can you provide more logs from the error you're receiving?

It looks like that error would have come from `isValidAgainstPrevious`, which gets called while adding the block to the chain, and at that point it would have already verified that the `previousBlockHash` is correct. I'm not sure how in `newBlock` the previousBlockHash could be set to `heaviestHead.hash`, but then the target generated by `calculateTarget` is incorrect.

NullSoldier: This function can only ever calculate a block on the current head inside the lock, and no other block. The reason is because the note commitments need to match, but if the head changes and the note commitments are calculated they'll be wrong and not based on the passed in block. You could guarantee it by locking the chain DB when generating the block template, but I think it's better to probably return the head it's based on in the template result.

NullSoldier:
Closing this PR as there is no activity.

PR #2073, opened 2022-08-19
Add error case in migration for nullifierToNote

NullSoldier:
## Summary
this adds an error case to the migration in case someone tries to migrate a corrupt wallet.

## Testing Plan
Run this with a corrupt wallet, and without.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/migrations/data/013-wallet-2.ts
+++ b/ironfish/src/migrations/data/013-wallet-2.ts
@@ -123,6 +123,11 @@ export class Migration013 extends Migration {
     await stores.old.noteToNullifier.clear(tx)
     logger.debug('\t' + BenchUtils.renderSegment(BenchUtils.endSegment(start)))
 
+    logger.debug('Migrating: Checking nullifierToNote')
+    start = BenchUtils.startSegment()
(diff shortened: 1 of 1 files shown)


ironfish/src/migrations/data/013-wallet-2.ts
```diff
@@ -426,6 +431,24 @@ export class Migration013 extends Migration {
...
+      const hasNote = await stores.new.decryptedNotes.has(noteHash, tx)
+
+      if (!hasNote) {
+        missing++
+      }
+    }
+
+    if (missing) {
+      throw new Error(
+        `Your wallet is corrupt and missing ${missing} notes for nullifiers. You should delete your accounts database and run this again.`,
```
hughy, line 448: Should we provide any guidance here on how to delete or reset the accounts database?

hughy (approved):

ironfish/src/migrations/data/013-wallet-2.ts
NullSoldier: Fixed

PR #2075, opened 2022-08-21
Estimate fee from mempool

hairtail:
## Summary
Estimate fee from mempool for better fee experience.
## Testing Plan
Local test.
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/package.json
+++ b/ironfish-cli/package.json
@@ -94,6 +94,9 @@
       "config": {
         "description": "Show and edit the node configuration"
       },
+      "fees": {
+        "description": "Get the fee status"
+      },
       "faucet": {
         "description": "Get coins to start using Iron Fish"
       },
--- a/ironfish-cli/src/commands/fees/estimateFees.ts
+++ b/ironfish-cli/src/commands/fees/estimateFees.ts
@@ -1,0 +1,48 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { IronfishCommand } from '../../command'
+import { RemoteFlags } from '../../flags'
+
+const DEFAULT_SPEED = 1
+
+export class EstimateFeesCommand extends IronfishCommand {
+  static description = `Estimate fee for send transaction`
+
+  static flags = {
+    ...RemoteFlags,
+  }
+
+  static args = [
+    {
+      name: 'speed',
+      parse: (input: string): Promise<string> => Promise.resolve(input.trim()),
+      required: false,
+      description: 'expected number of blocks for your transaction to be mined',
+    },
+  ]
+
+  async start(): Promise<void> {
+    const { args } = await this.parse(EstimateFeesCommand)
+    const speed = args.speed as number | null
+
+    const targetConfirmSpeed = speed || DEFAULT_SPEED
+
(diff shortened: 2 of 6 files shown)


NullSoldier:
I think we should probably look at deprecating the existing fee model that uses the block chain that was added recently and use this. Most fee estimation models use the mem pool and not the block chain as far as I can tell. However, given that I think we should not add this PR alongside the existing set of code, but change it to use this.

wd021:
Some thoughts on fee estimation - 

1 - [bitcoin](https://bitcoincore.org/en/doc/0.16.0/rpc/util/estimatesmartfee/) and [ethereum](https://web3js.readthedocs.io/en/v1.5.0/web3-eth.html#getfeehistory) all use prev blocks for fee estimation. bitcoin actually uses a combo..

**The problems i see with using mempool are**

1. every node has a different mempool
2. attack vector to flood the mempool with txs way below current avg fee to break the 'fee estimator'
3. what happens when the mempool is empty and you want to submit a tx right after prev block, where most of the mempool fills after your tx
4. massive variance. can't abstract from the spikes and lulls.
5. can't estimate for fast vs. slow situation. sometimes i want tx in the next block, sometimes I'm ok to wait x blocks.

One issue with relying solely on using previous block is that miners could game this by stuffing their own block with high fee txs and artificially increase fees. The attack becomes way more costly if mempool data is also included in the algorithm, which is why bitcoin core includes both. 

For now, i think what we have is sufficient. Probably look to upgrade down the road and include both. just my 2c.

wd021:
Some thoughts on fee estimation - 

1 - [bitcoin](https://bitcoincore.org/en/doc/0.16.0/rpc/util/estimatesmartfee/) and [ethereum](https://web3js.readthedocs.io/en/v1.5.0/web3-eth.html#getfeehistory) all use prev blocks for fee estimation. bitcoin actually uses a combo..

**The problems i see with using mempool are**

1. every node has a different mempool
2. attack vector to flood the mempool with txs way below current avg fee to break the 'fee estimator'
3. what happens when the mempool is empty and you want to submit a tx right after prev block, where most of the mempool fills after your tx
4. massive variance. can't abstract from the spikes and lulls.
5. can't estimate for fast vs. slow situation. sometimes i want tx in the next block, sometimes I'm ok to wait x blocks.

One issue with relying solely on using previous block is that miners could game this by stuffing their own block with high fee txs and artificially increase fees. The attack becomes way more costly if mempool data is also included in the algorithm, which is why bitcoin core includes both. 

For now, i think what we have is sufficient. Probably look to upgrade down the road and include both. just my 2c.

wd021:
metamask uses prev block, not the mempool. see here: https://metamask.io/1559. "MetaMask will initially set this amount based on the previous block’s history."

hairtail:
> metamask uses prev block, not the mempool. see here: https://metamask.io/1559. "MetaMask will initially set this amount based on the previous block’s history."

Get

hughy:
We're working on some changes to fee estimation that we hope to ship in the next few weeks.

The contributions here for estimating fees from the mempool and the earlier PR that added fee estimation from recent blocks are both very helpful, and the discussion on this PR was informative as we researched our approach to fee estimation!

At a high level, the approach that we're taking in the upcoming fee estimation changes is to use fees from recent blocks for estimation, but to filter out transactions that aren't in the mempool. This filtering step is to prevent the fee attack from miners stuffing blocks with high fee transactions.

Thank you both @hairtail and @wd021 !

PR #2100, opened 2022-08-24
Remove toThrowErrorInstance

NullSoldier:
## Summary

You can pass a constructable to the build in jest toThrowError() so this
is not needed. I didn't know this when I wrote it, or it wasn't
available then.

## Breaking Change

```
[ ] Yes
```


ironfish/src/testUtilities/nodeTest.ts
```diff
@@ -1,6 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import './matchers'
```
NullSoldier, line 4: So all of our matchers are available in any test. This is an orthagonal change but I snuck it in here.

rohanjadvani, line 4: Is this needed?

rohanjadvani: Nevermind, saw your above comment

rohanjadvani (approved):

PR #2103, opened 2022-08-24
deposit-all should not exit if balance changes

NullSoldier:
## Summary

Before if balance changes while creating a transaction, it would crash
before.

## Testing Plan

Throw a `NotEnoughFundsError` inside of sendTransaction to trigger this.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/deposit-all.ts
+++ b/ironfish-cli/src/commands/deposit-all.ts
@@ -232,6 +232,7 @@ export default class DepositAll extends IronfishCommand {
           ) {
             // Our balance changed while trying to create a payout, ignore this error
             await PromiseUtils.sleep(30000)
+            continue
           }
 
           screen.destroy()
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/deposit-all.ts
```diff
@@ -224,6 +226,14 @@ export default class DepositAll extends IronfishCommand {
...
           txs.push(transaction)
         } catch (error) {
+          if (
+            error instanceof RpcRequestError &&
+            error.code === ERROR_CODES.INSUFFICIENT_BALANCE
+          ) {
+            // Our balance changed while trying to create a payout, ignore this error
+            await PromiseUtils.sleep(30000)
+          }
+
```
hughy, line 236: Does this need a `continue` or an `else` block? It looks like this will sleep 30s but then throw the error

NullSoldier: Good catch, thank you I forgot to bring that over from my other branch.

hughy (approved):

PR #2105, opened 2022-08-24
filters nullifiers and transactions by account on load

hughy:
## Summary

fixes transaction expiration with multiple accounts

expiring transactions on a wallet with multiple accounts may fail with `Error:
nullifierToNote mappings must have a corresponding decryptedNote`.

expiring transactions deletes the transaction from all accounts in the wallet.
each account loads all transactions and all nullifiers from the database, but
loads only the decrypted notes that it owns. the error occurs when a transaction
uses a nullifier to try to load a decrypted note that it does not own.

the solution suggested here is for each account to load only the nullifiers that
correspond to decrypted notes that the account owns. along with filtering
nullifiers by account it makes sense to filter transactions by account.

- adds test of expiring transactions on node with multiple accounts
- builds nullifierToNoteHash mapping while loading notes
- reduces the nullifiers tracked by an account to only those that map to a
  decrypted note that belongs to that account.
- filters transactions by account to only those transactions where an account
  either owns a note that was spent in the transaction or a note output by the
  transaction.
- adds tests of loading transactions and nullifiers by account

- renames Accounts.removeTransaction to deleteTransaction to match Account
- refactors Account.deleteTransaction
        - renames merkleHash to noteHash
        - removes hash parameter, reads hash from transaction parameter

## Testing Plan

- adds unit tests
- run a local node with multiple accounts, send a transaction and have it expire without error

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/account.ts
+++ b/ironfish/src/wallet/account.ts
@@ -96,12 +96,6 @@ export class Account {
   }
 
   async load(): Promise<void> {
-    await this.accountsDb.loadNullifierToNoteHash(this.nullifierToNoteHash)
-    await this.accountsDb.loadTransactions(this.transactions)
-    await this.loadDecryptedNotesAndBalance()
-  }
-
-  private async loadDecryptedNotesAndBalance(): Promise<void> {
     let unconfirmedBalance = BigInt(0)
 
     for await (const { hash, decryptedNote } of this.accountsDb.loadDecryptedNotes()) {
@@ -112,7 +106,34 @@ export class Account {
           unconfirmedBalance += new Note(decryptedNote.serializedNote).value()
         }
 
-        this.saveDecryptedNoteSequence(decryptedNote.transactionHash, hash)
+        const nullifierHash = decryptedNote.nullifierHash
+        if (nullifierHash) {
+          this.nullifierToNoteHash.set(nullifierHash, hash)
+        }
+
+        const transactionHash = decryptedNote.transactionHash
+        const transaction = await this.accountsDb.loadTransaction(transactionHash)
+        Assert.isNotUndefined(
+          transaction,
+          `Transaction undefined for '${transactionHash.toString('hex')}'`,
+        )
+
+        this.transactions.set(transactionHash, transaction)
+
+        this.saveDecryptedNoteSequence(transactionHash, hash)
+      }
+    }
+
+    for await (const { hash, transaction } of this.accountsDb.loadTransactions()) {
+      if (this.transactions.has(hash)) {
+        continue
+      }
+
+      for (const spend of transaction.transaction.spends()) {
+        if (this.nullifierToNoteHash.has(spend.nullifier)) {
+          this.transactions.set(hash, transaction)
+          break
+        }
       }
     }
 
--- a/ironfish/src/wallet/account.ts
+++ b/ironfish/src/wallet/account.ts
@@ -99,29 +99,31 @@ export class Account {
     let unconfirmedBalance = BigInt(0)
 
     for await (const { hash, decryptedNote } of this.accountsDb.loadDecryptedNotes()) {
-      if (decryptedNote.accountId === this.id) {
-        this.decryptedNotes.set(hash, decryptedNote)
-
-        if (!decryptedNote.spent) {
-          unconfirmedBalance += new Note(decryptedNote.serializedNote).value()
-        }
-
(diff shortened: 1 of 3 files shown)


ironfish/src/wallet/accounts.test.slow.ts
```diff
@@ -125,6 +125,88 @@ describe('Accounts', () => {
     })
   }, 600000)
 
+  it('Loads only the transactions that an account owns a note or spend nullifier for', async () => {
```
NullSoldier, line 128: This test only takes 300-400ms, can we move it to the non slow file?

```text
> tsc -b && tsc -b tsconfig.test.json && cross-env TEST_INIT_RUST=true jest --testMatch "**/accounts.test.slow.ts" --testPathIgnorePatterns

 PASS   @ironfish/sdk  src/wallet/accounts.test.slow.ts (13.04 s)
  Accounts
    ✓ Returns the correct balance when an account receives a miners fee (327 ms)
    ✓ Saves and restores transactions from accounts db (263 ms)
    ✓ Loads only the transactions that an account owns a note or spend nullifier for (191 ms)
    ✓ Loads only the nullifiers that an account owns a note for (206 ms)
    ✓ Lowers the balance after using pay to spend a note (1600 ms)
    ✓ Creates valid transactions when the worker pool is enabled (1622 ms)
    ✓ creates valid transactions with multiple outputs (2080 ms)
    ✓ throws a ValidationError with an invalid expiration sequence (51 ms)
    ✓ Expires transactions when calling expireTransactions (1683 ms)
    ✓ Expires transactions when calling expireTransactions with restarts (1897 ms)
    ✓ Counts notes correctly when a block has transactions not used by any account (1357 ms)
    ✓ Removes notes when rolling back a fork (158 ms)
    ✓ Keeps spends created by the node when rolling back a fork (214 ms)
    ✓ Undoes spends created by another node when rolling back a fork (207 ms)

Test Suites: 1 passed, 1 total
Tests:       14 passed, 14 total
Snapshots:   0 total
Time:        13.078 s
Ran all test suites.
✨  Done in 15.27s.
```

NullSoldier, line 128: Just noting that this test takes 300-400ms.

ironfish/src/wallet/accounts.test.slow.ts
```diff
@@ -125,6 +125,88 @@ describe('Accounts', () => {
...
+    const accountB = await node.accounts.createAccount('B', true)
+
+    // Load account data from db
+    await node.accounts.loadAccountsFromDb()
+
+    // Account A should have one nullifier
+    expect(accountA['nullifierToNoteHash'].size).toEqual(1)
+
+    // Account B should have zero nullifiers
+    expect(accountB['nullifierToNoteHash'].size).toEqual(0)
```
NullSoldier, line 207: If we put these lines before `loadAccountsFromDb` it'll still pass. So you can delete `loadAccountsFromDb` and the test will still pass. How do we know we've prevented the bug?

hughy: This test fails on `accounts-performance` where `loadAccountsFromDb` is loading nullifiers into both accounts.

I'll update the test to clear the in-memory stores for both accounts before loading so that it's clearer that the load is changing for both accounts

ironfish/src/wallet/account.ts
```diff
@@ -112,7 +106,34 @@ export class Account {
           unconfirmedBalance += new Note(decryptedNote.serializedNote).value()
         }
 
-        this.saveDecryptedNoteSequence(decryptedNote.transactionHash, hash)
+        const nullifierHash = decryptedNote.nullifierHash
```
NullSoldier, line 109: this is opinionated nit but can we early return on the node id and just unindent this whole section?

ironfish/src/wallet/database/accountsdb.ts
```diff
@@ -275,25 +275,46 @@ export class AccountsDB {
...
+    tx?: IDatabaseTransaction,
+  ): Promise<
+    | {
+        transaction: Transaction
+        blockHash: Buffer | null
+        sequence: number | null
+        submittedSequence: number | null
       }
-    })
+    | undefined
```
NullSoldier, line 308: Let's return null here, like in most of blockchain.

NullSoldier (approved):

PR #2106, opened 2022-08-24
separates clearing databases from replacing keys

hughy:
## Summary

the `replace*` methods in the wallet database clear all keys for a datastore
before setting the key-value pairs from a map passed to the method. however,
each map might only have key-value pairs for a single account in the wallet.
this results in clearing all decrypted notes from the database whenever an
account saves its decrypted notes map.

- adds test of saving and reloading with multiple accounts
- moves `clear` out of `replace*` methods into a separate method for each
  datastore
- uses `clear` in `Accounts.reset` instead of using `saveAccountsToDb` to clear
  keys from the database

## Testing Plan

- adds unit test to verify decryptedNotes aren't cleared by save from another account

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/database/accountsdb.ts
+++ b/ironfish/src/wallet/database/accountsdb.ts
@@ -214,9 +214,7 @@ export class AccountsDB {
   }
 
   async removeHeadHashes(tx?: IDatabaseTransaction): Promise<void> {
-    await this.database.withTransaction(tx, async (tx) => {
-      await this.headHashes.clear(tx)
-    })
+    await this.headHashes.clear(tx)
   }
 
   async *loadHeadHashes(
@@ -253,9 +251,7 @@ export class AccountsDB {
   }
 
   async clearTransactions(tx?: IDatabaseTransaction): Promise<void> {
-    await this.database.withTransaction(tx, async (tx) => {
-      await this.transactions.clear(tx)
-    })
+    await this.transactions.clear(tx)
   }
 
   async replaceTransactions(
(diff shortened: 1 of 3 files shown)


ironfish/src/wallet/database/accountsdb.ts
```diff
@@ -252,6 +252,12 @@ export class AccountsDB {
     })
   }
 
+  async clearTransactions(tx?: IDatabaseTransaction): Promise<void> {
+    await this.database.withTransaction(tx, async (tx) => {
```
NullSoldier, line 256: we don't need this withTransaction, we can just pass this down directly.

ironfish/src/wallet/database/accountsdb.ts
```diff
@@ -312,13 +316,17 @@ export class AccountsDB {
     })
   }
 
+  async clearNullifierToNoteHash(tx?: IDatabaseTransaction): Promise<void> {
+    await this.database.withTransaction(tx, async (tx) => {
```
NullSoldier, line 320: again, I don't think we need to start a transaction.

ironfish/src/wallet/wallet.ts
```diff
@@ -294,7 +294,9 @@ export class Accounts {
     await this.db.database.transaction(async (tx) => {
       await this.resetAccounts(tx)
       this.chainProcessor.hash = null
-      await this.saveAccountsToDb(tx)
+      await this.db.clearDecryptedNotes(tx)
+      await this.db.clearNullifierToNoteHash(tx)
+      await this.db.clearTransactions(tx)
```
NullSoldier, line 299: I think these 3 lines should probably go into resetAccounts. My intuition is that I ask myself, "What is resetAccounts responsibility? I don't think reset accounts should ever be called without these 3 lines accompanying it, so that's why I think they should be together.

hughy: I think the only use case for resetting the in-memory account storage without clearing the database is in tests

I'll update

NullSoldier (approved):

PR #2113, opened 2022-08-25
deserializes transaction on TransactionValue

hughy:
## Summary

changes the `transaction` field of `TransactionValue` to an instance of the
`Transaction` primitive instead of a serialized transaction buffer. avoids
defining a separate anonymous type elsewhere in wallet and avoids using spread
operator to coax TransactionValue into that type.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/database/transactionValue.test.ts
+++ b/ironfish/src/wallet/database/transactionValue.test.ts
@@ -1,22 +1,17 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { Transaction } from '../../primitives'
-import { createNodeTest, useAccountFixture } from '../../testUtilities'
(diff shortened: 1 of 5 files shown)


ironfish/src/wallet/database/transactionValue.test.ts
```diff
@@ -1,15 +1,25 @@
...
+import { createNodeTest, useAccountFixture } from '../../testUtilities'
 import { TransactionValue, TransactionValueEncoding } from './transactionValue'
 
 describe('TransactionValueEncoding', () => {
+  const nodeTest = createNodeTest()
+  let transaction: Transaction
+
+  beforeEach(async () => {
+    const account = await useAccountFixture(nodeTest.accounts)
+    transaction = await nodeTest.strategy.createMinersFee(BigInt(0), 1, account.spendingKey)
```
NullSoldier, line 14: It's better to move these to each test that uses it, and use a useMinersTxFixture()

NullSoldier (approved):

NullSoldier (approved):

PR #2123, opened 2022-08-26
Remove duplicated commands from ironfish-cli.

krompets:
Currently in help we see two commands deposit-all and depositAll which doing the same things.
```
  deposit-all  Deposit $IRON for testnet points
  depositAll   Deposit $IRON for testnet points
```
In the previous tasks were decided to use kebab-case instead of camelCase in commands.
After this change we will have only one command:
```
deposit-all  Deposit all $IRON for testnet points
```

## Breaking Change

```
[ No ] 
```

### DIFF
--- a/ironfish-cli/src/commands/deposit-all.ts
+++ b/ironfish-cli/src/commands/deposit-all.ts
@@ -25,8 +25,7 @@ const REGISTER_URL = 'https://testnet.ironfish.network/signup'
 const IRON_TO_SEND = 0.1
 
 export default class DepositAll extends IronfishCommand {
-  static aliases = ['depositAll']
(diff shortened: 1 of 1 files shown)


NullSoldier:
This is removing backwards compatibility without offering a strategy for it moving forward. Are you implying we never offer backwards compatibility when we rename commands, or just that it's been long enough since we changed the command?

krompets:
@NullSoldier I just was a little bit confused when saw two commands which doing the same. Also I saw that we want to avoid using camelCase. But I did not think about backwards compatibility. So this change can produce requests from users who uses "depositAll" instead of "deposit-all" command. Maybe we can write in announcement that "depositAll command deprecated please use deposit-all instead"?

PR #2127, opened 2022-08-28
Add blockInfo for getTransaction

hairtail:
## Summary
Add blockInfo for getTransaction.
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/accounts/transaction.ts
+++ b/ironfish-cli/src/commands/accounts/transaction.ts
@@ -47,6 +47,10 @@ export class TransactionCommand extends IronfishCommand {
     this.log(`Status: ${response.content.transaction.status}`)
     this.log(`Miner Fee: ${response.content.transaction.isMinersFee ? `✔` : `x`}`)
     this.log(`Fee: ${CurrencyUtils.renderIron(response.content.transaction.fee, true)}`)
+    if (response.content.transaction.blockHash && response.content.transaction.blockSequence) {
+      this.log(`Block Hash: ${response.content.transaction.blockHash}`)
+      this.log(`Block Sequence: ${response.content.transaction.blockSequence}`)
+    }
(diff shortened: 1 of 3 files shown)


NullSoldier:
Were not accepting changes to wallet 1.0 at this time until wallet 2.0 merges.

hairtail:
After read some documentation about ethereum rpc api`eth_getTransactionByHash`, I think it's better to add `hash` `blockHash` `blockSequence` in `getTransaction`.

ironfish-cli/src/commands/accounts/transaction.ts
```diff
@@ -47,6 +47,10 @@ export class TransactionCommand extends IronfishCommand {
     this.log(`Status: ${response.content.transaction.status}`)
     this.log(`Miner Fee: ${response.content.transaction.isMinersFee ? `✔` : `x`}`)
     this.log(`Fee: ${CurrencyUtils.renderIron(response.content.transaction.fee, true)}`)
+    if (response.content.transaction.blockHash && response.content.transaction.blockSequence) {
+      this.log(`Block Hash: ${response.content.transaction.blockHash}`)
+      this.log(`Block Sequence: ${response.content.transaction.blockSequence}`)
+    }
```
hughy, line 53: It might be helpful to display these rows even if the values are null. Perhaps with some indication that they're null because the transaction hasn't been included in a block?

hughy:
This looks good! You may need to rebase onto the latest `staging` branch to fix the failing CI check.

ironfish-cli/src/commands/accounts/transaction.ts
hairtail: I think this is unnecessary because we have `response.content.transaction.status` to indicate whether the transaction is `pending` `unconfirmed` `expired` etc.

hughy (approved):

PR #2128, opened 2022-08-28
Add from sequence for accounts:rescan

hairtail:
## Breaking Change

```
[ ] Yes
```


NullSoldier:
Were not accepting any wallet code right now, as its making the rebases on wallet 2.0 very difficult. Once wallet 2.0 merges we'll consider these types of PRS.

NullSoldier:
Can you write an explanation under which a user would use this, and why it wouldn't be automatically done for them?

hairtail:
> Can you write an explanation under which a user would use this, and why it wouldn't be automatically done for them?

Example: account my_account is created at sequence 100000, so blocks before 100000 is unnecessary to scan.

PR #2131, opened 2022-08-29
Move transaction to protectedNamespaces for rpc.tcp

hairtail:
## Summary
Transaction route should be protected since IRON token can be sent via transaction/sendTransaction directly.
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/router.ts
+++ b/ironfish/src/rpc/routes/router.ts
@@ -25,7 +25,11 @@ export enum ApiNamespace {
   rpc = 'rpc',
 }
 
-export const API_NAMESPACES_PROTECTED = [ApiNamespace.account, ApiNamespace.config]
+export const API_NAMESPACES_PROTECTED = [
+  ApiNamespace.account,
+  ApiNamespace.config,
+  ApiNamespace.transaction,
+]
 export const ALL_API_NAMESPACES = StrEnumUtils.getValues(ApiNamespace)
 
 export type RouteHandler<TRequest = unknown, TResponse = unknown> = (


ironfish/src/sdk.ts
```diff
@@ -204,7 +204,11 @@ export class IronfishSdk {
     }
 
     if (this.config.get('enableRpcTcp')) {
-      const protectedNamespaces = [ApiNamespace.account, ApiNamespace.config]
+      const protectedNamespaces = [
+        ApiNamespace.account,
+        ApiNamespace.config,
+        ApiNamespace.transaction,
+      ]
```
NullSoldier, line 211: This seems good, we'll have to put out an announcement to the community regarding this they are likely to no longer be able to send tranactions.

hairtail: If i want to use `transaction/sendTransaction` over tcp, i would first use `accounts/getBalance`. 
Since `accounts` routes are protected before, there is no necessary to put out an announcement for `sendTransaction`.

NullSoldier: Good point.

NullSoldier (approved):

NullSoldier:
Can you rebase this? We moved the protected namespaces next to ALL_API_NAMESPACES now.

hairtail:
> Can you rebase this? We moved the protected namespaces next to ALL_API_NAMESPACES now.

Sure things

NullSoldier:
We're going to close this in favor of https://github.com/iron-fish/ironfish/pull/2224 and deprecate protected namespaces entirely. We'll still give you points for this PR.

PR #2139, opened 2022-08-30
Store all test data in the repo

NullSoldier:
## Summary

This was taking up all the system storage, and not getting cleaned up.
This now stores all test node data directories inside of /testdbs with
our other data. This directory gets deleted at the start of each run.

## Testing Plan

Run tests, and check ./testdbs

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/jest.setup.js
+++ b/ironfish-cli/jest.setup.js
@@ -1,9 +1,15 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 const fs = require('fs')
+const path = require('path')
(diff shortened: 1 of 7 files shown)


mat-if:
This seems to leave more data in the testdb folder than the current implementation does. On staging, when running yarn test, it will leave behind 1 folder in testdbs, whereas on this branch, I get hundreds. Consequent runs will clean up on both. I'm curious if you're seeing different behavior or if I'm misunderstanding the goal here.

Staging: 
image

This branch (there's way too many to try and fit):
image

NullSoldier:
the way it used to be done, is to just use your nodes tmp directory, but I found that folder never gets cleaned up by mac in reality and it causes my disk to fill up. This contains all the test garbage to the repo where it can get cleaned up easily.

mat-if (approved):

PR #2140, opened 2022-08-30
Add key ranges to all range based DB ops

NullSoldier:
## Summary

This now supports key ranges for all range based database operations. That means you can use any operations like clear(), getAllKeys() filtered to a subset of rows.

This also adds a new storage utility file to help generate buffer key range prefixes easily.

## Testing Plan

Run the new written tests, and run the node.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/storage/database.test.ts
+++ b/ironfish/src/storage/database.test.ts
@@ -16,6 +16,7 @@ import {
   StringEncoding,
   TransactionWrongDatabaseError,
 } from './database'
+import { StorageUtils } from './database/utils'
 import { LevelupDatabase, LevelupStore } from './levelup'
 
 type FooValue = {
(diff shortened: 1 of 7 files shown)


ironfish/src/storage/database.test.ts
```diff
@@ -678,6 +719,12 @@ describe('Database', () => {
     })
 
     it('should not find entries with an off-by-one prefix and empty key', async () => {
+      const testPrefixKeyStore = db.addStore<BazSchema>({
+        name: 'PrefixKey',
+        keyEncoding: new BufferEncoding(),
+        valueEncoding: new StringEncoding(),
+      })
```
NullSoldier, line 726: Just moved this here since no one else should use this anyway, as indicated by the comment.

PR #2143, opened 2022-08-30
Remove unnecessary transaction verification

dguenther:
## Summary

Transactions have to be turned into Rust transactions when we run verification in the worker pool anyway, so we may as well not do it on the main thread (since it's fairly slow)

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/consensus/verifier.ts
+++ b/ironfish/src/consensus/verifier.ts
@@ -187,6 +187,17 @@ export class Verifier {
    * the mempool and rebroadcasted to the network.
    */
   async verifyNewTransaction(transaction: Transaction): Promise<VerificationResult> {
+    let verificationResult
+    try {
+      verificationResult = await this.workerPool.verify(transaction)
+    } catch {
+      verificationResult = { valid: false, reason: VerificationResultReason.VERIFY_TRANSACTION }
+    }
+
+    if (!verificationResult.valid) {
+      return verificationResult
+    }
+
     // Check the spends are valid and not already in the chain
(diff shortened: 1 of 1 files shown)


mat-if (approved):

ironfish/src/consensus/verifier.ts
```diff
@@ -187,16 +187,7 @@ export class Verifier {
...
-      // Transaction is lazily deserialized, so we use takeReference()
-      // to force deserialization errors here
-      transaction.takeReference()
-      transaction.returnReference()
-    } catch (e) {
-      return { valid: false, reason: VerificationResultReason.DESERIALIZATION }
-    }
-
-    // Next, check the spends are valid and not already in the chain
+    // Check the spends are valid and not already in the chain
```
NullSoldier, line 190: This will let crashes through potentially to through the rest of the code base. If you think this is acceptable, then I'm fine merging it but I feel like removing validation doesn't feel really safe.

dguenther: That's incorrect -- the same verification is being performed, but removing a duplicate step (constructing a TransactionPosted on the main thread, then constructing a TransactionPosted in the worker pool).

However, this change might allow invalid data to go to spend verification. I can't imagine a scenario where the original transaction would be parsed without crashing in the `Transaction` constructor, then end up crashing in spend verification -- spend verification is mostly just comparing the equality of buffers. But regardless, I moved the worker pool verification before it.

PR #2149, opened 2022-08-31
Hide inspect port in use error

NullSoldier:
## Summary

This will use a randomly assigned port if the default inspect port is in use. Otherwise you get a message like this `Starting inspector on 127.0.0.1:9229 failed: address already in use` printed.

## Testing Plan

1. Run `yarn start start`
2. Run `yarn start debug`
3. See that `Starting inspector on 127.0.0.1:9229 failed: address already in use` is not printed

## Breaking Change

```
[ ] Yes
```


mat-if:
If we randomize the port, we need a way to surface that port. Debugger tools know to look for the default one, but not random ones. I haven't seen a way to figure out the port after-the-fact yet, although I've only begun looking

mat-if (approved):

PR #2155, opened 2022-08-31
Make each TcpAdapter test use an OS assigned port

NullSoldier:
## Summary

Make each TcpAdapter test use an OS assigned port

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/adapters/tcpAdapter.test.ts
+++ b/ironfish/src/rpc/adapters/tcpAdapter.test.ts
@@ -5,65 +5,77 @@
 /* eslint-disable jest/no-conditional-expect */
 import os from 'os'
 import * as yup from 'yup'
+import { Assert } from '../../assert'
 import { IronfishSdk } from '../../sdk'
-import { RpcRequestError, RpcSocketClient } from '../clients'
-import { ALL_API_NAMESPACES, ApiNamespace } from '../routes'
+import { RpcRequestError } from '../clients'
+import { RpcTcpClient } from '../clients/tcpClient'
+import { ALL_API_NAMESPACES } from '../routes'
 import { ERROR_CODES, ValidationError } from './errors'
 import { RpcTcpAdapter } from './tcpAdapter'
 
 describe('TcpAdapter', () => {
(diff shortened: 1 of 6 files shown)


ironfish/src/rpc/adapters/tcpAdapter.test.ts
```diff
@@ -3,64 +3,63 @@
...
+    },
+  })
 
-    tcp = node.rpc.adapters[0] as RpcTcpAdapter
+  let adapter: RpcTcpAdapter
+  let client: RpcTcpClient
 
-    client = sdk.client
+  beforeEach(() => {
+    Assert.isInstanceOf(nodeTest.node.rpc.adapters[0], RpcTcpAdapter)
```
lwisne, line 29: very small nit, but I would have flipped this line with the one below and asserted against adapter

ironfish/src/rpc/adapters/tcpAdapter.test.ts
```diff
@@ -3,64 +3,64 @@
...
-    tcp.router?.register('foo/bar', yup.string(), (request) => {
+    adapter.router?.register('foo/bar', yup.string(), (request) => {
       request.end(request.data)
     })
 
-    await tcp.start()
+    await adapter.start()
+
+    Assert.isNotNull(adapter.addressPort)
+    client = new RpcTcpClient('localhost', adapter.addressPort)
```
lwisne, line 46: Its' a little funky that you call this constructor instead of using `sdk.client`, as the latter is used by a lot of code and structured differently. However, I don't see an easy alternative as it may be cludgy to pass the port stuff into `sdk`

lwisne (approved):
shall nit, otherwise looks great!

ironfish/src/rpc/adapters/tcpAdapter.test.ts
NullSoldier: You can't do that in typescript because you can't assign RpcClient to RpcTcpClient, and then if you change the variable to RpcClient you'll need to copy this assert to every test that uses it.

PR #2157, opened 2022-09-01
Remove duplicate type

hairtail:
## Breaking Change

```
[ ] Yes
```


mat-if (approved):
Good catch. I'm a little surprised the typescript linter doesn't catch this

PR #2158, opened 2022-09-01
Ignore /target dir in build script, minor dockerfile improvements

mat-if:
## Summary

Takes my local docker image from like 1.3gb to 450mb

Closes IRO-2555

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/Dockerfile
+++ b/ironfish-cli/Dockerfile
@@ -1,13 +1,17 @@
 FROM node:16.15.1-bullseye as build
 ENV PATH="/root/.cargo/bin:${PATH}"
 
-COPY ./ ./
-
 RUN \
     apt-get update && \
-    apt-get install jq rsync gcc cmake python -y && \
-    curl https://sh.rustup.rs -sSf | sh -s -- -y && \
-    ./ironfish-cli/scripts/build.sh
+    apt-get install jq rsync gcc cmake python -y
+
+RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
+
+WORKDIR ./app
+
+COPY ./ ./
+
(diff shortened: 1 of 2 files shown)


ironfish-cli/Dockerfile
```diff
@@ -1,13 +1,17 @@
...
 ENV PATH="/root/.cargo/bin:${PATH}"
 
-COPY ./ ./
-
 RUN \
     apt-get update && \
-    apt-get install jq rsync gcc cmake python -y && \
-    curl https://sh.rustup.rs -sSf | sh -s -- -y && \
-    ./ironfish-cli/scripts/build.sh
+    apt-get install jq rsync gcc cmake python -y
```
mat-if, line 6: Splitting these into individual run steps lets us use docker layer caching so we don't have to re-install these and rustup every time

ironfish-cli/Dockerfile
```diff
@@ -1,13 +1,17 @@
...
+
+RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
+
+WORKDIR ./app
```
mat-if, line 10: This is just a minor tweak, made it a little cleaner when debugging, so I don't have to go "what is home and where did it come from" since it was building at / previously

ironfish-cli/Dockerfile
NullSoldier: Yeah. We can do this because this step in the multi stage build is dropped. If this was in the final image, docker recommends to reduce layers by combining them to avoid packages writing to the same files in older layers. https://docs.docker.com/develop/dev-best-practices/#how-to-keep-your-images-small

NullSoldier (approved):

PR #2162, opened 2022-09-03
Access control for rpc request over tcp

hairtail:
## Summary
Add auth system for protected rpc requests over tcp. With this PR, protected routes can be visited with a specific token.
## Testing Plan
Local test completed.
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/fileStores/config.ts
+++ b/ironfish/src/fileStores/config.ts
@@ -21,7 +21,7 @@ export const DEFAULT_MINER_BATCH_SIZE = 25000
 export const DEFAULT_EXPLORER_BLOCKS_URL = 'https://explorer.ironfish.network/blocks/'
 export const DEFAULT_EXPLORER_TRANSACTIONS_URL =
   'https://explorer.ironfish.network/transaction/'
-export const DEFAULT_TCP_AUTH_TOKEN = 'ironfish-testnet'
+export const DEFAULT_RPC_AUTH_TOKEN = 'ironfish-testnet'
 
 // Pool defaults
 export const DEFAULT_POOL_NAME = 'Iron Fish Pool'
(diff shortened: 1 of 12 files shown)


ironfish/src/fileStores/config.ts
```diff
@@ -330,6 +336,7 @@ export class Config extends KeyStore<ConfigOptions> {
       jsonLogs: false,
       explorerBlocksUrl: DEFAULT_EXPLORER_BLOCKS_URL,
       explorerTransactionsUrl: DEFAULT_EXPLORER_TRANSACTIONS_URL,
+      authToken: DEFAULT_TCP_AUTH_TOKEN,
```
NullSoldier, line 339: We should name this `rpcAuthToken`

NullSoldier: The name of this default doesn't match the config.

hairtail: Done.

NullSoldier:
Were going to credit you for this PR but were gonna close this in favor of https://github.com/iron-fish/ironfish/pull/2224 which we were already starting. Thank you for this PR because it's serving as the base for their PR.

hairtail:
> Were going to give you testnet credit for this PR, but were gonna close this in favor of #2224. We already started it internally, and want to do it slightly differently. Mostly we want to
> 
> 1. Have authentication live in RpcServer and allow any adapter to use it
> 2. Authenticate over the whole adapter and delete Protected Namespaces (similar to bitcoin)

Sorry, i can't get the idea why we need auth over the whole adapter. There are some public rpc interface like `chain/getBlock, node/getStatus, fee/getFees`. These kind of routes are unnecessary to verify the auth.

PR #2166, opened 2022-09-06
Set proper Error type and message on Insufficient funds case

Skakruk:
## Summary

DepositAll command is catching an error with the error code `INSUFFICIENT_BALANCE`, but `createTransaction` was throwing a default Error, without that code, which leads the command to shut down.

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):

NullSoldier:
Thanks for this!

PR #2167, opened 2022-09-06
Remove support for old transaction gossip message

danield9tqh:
## Summary
This is cleanup related to the transaction gossip changes. We had a message type NewTransactionMessage  that was used in the previous gossip implementation. We kept backwards compatibility for it to allow people to upgrade slowly. Now we are at >90% of the network upgraded so feel that we are ok to remove support.

## Testing Plan
Unit tests and running node on network

## Breaking Change

```
[ ] Yes
```


NullSoldier:
This is marked as not a backwards compatible breaking change, but it is right?

ironfish/src/network/version.ts
```diff
@@ -3,4 +3,4 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 export const VERSION_PROTOCOL = 18
-export const VERSION_PROTOCOL_MIN = 15
+export const VERSION_PROTOCOL_MIN = 17
```
NullSoldier, line 6: We bump this 2 because we did 2 versions of network protocol that are not incompatible?

danield9tqh: Looks like we bumped it once (15 -> 16) when we replaced the [gossip nonce filter with a transaction hash filter](https://github.com/iron-fish/ironfish/pull/1865) and bumped it again for [starting to gossip transactions based on hash](https://github.com/iron-fish/ironfish/pull/1803)

danield9tqh: So we did two bumps for the transaction changes. We may have been thinking we needed to bump the min version at the same time we made the change to start gossiping transactions by hash but we didn't end up doing that

NullSoldier:
I think the PR looks pretty good but can we mark it as non backwards compatible?

danield9tqh:
Closing this for now since I've seen a few potential bugs running it locally and moving on to some more urgent bug fixes. Will revisit later

Issue #2092, opened 2022-08-23
Feature: Move all list commands into root oclif categories


NullSoldier:
### Description

Currently we use `list` as the standard for commands that list an item. For eaxmple: `peers:list`, `accounts:list`. Instead, we should make the root oclif topic the list itself. To do this, we should move the list commands to index.ts in each topic. It's important to note we should have aliases back to their old command for now. Generally we support aliases for a month or two before we delete those aliases.

https://linear.app/ironfish/issue/IRO-2565/move-all-list-commands-into-root-oclif-categories

PR #2171, opened 2022-09-12
Move commands migrations:status and config:show to their root commands

mirayadav:
## Summary

Feature #2092, move additional commands (`migrations:status` and `config:show`) to their respective root commands (`migrations` and `config`). Leave in original commands as aliases for now (to be deleted eventually).

## Testing Plan
- Tested both `migrations` and `config` commands to ensure they perform the same operation as `migrations:status` and `config:show`.

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):

NullSoldier:
Looks good, thanks!

Issue #2147, opened 2022-08-31
Blockchain "synced" is too sticky in Ironfish Status.  Does not clear when sync is lost 


crProductGuy:
### What happened?

When a node loses sync after achieving sync, `ironfish status -f` "Blockchain" line with "Synced" state should say "NOT SYNCED", but in fact it still says Synced even while the Syncer line in Status shows SYNCING.

Something is preventing the Blockchain status line from discovering the true current status. 

I have observed this multiple times when sync is lost for one reason or another, for example very slow block add times, loss of network connection, inadvertent pausing the node in CMD window, etc. 

I have also seen it just after restarting the node (dedicated node, no miners) on my laptop, when the node had been a few blocks behind sync when I stopped the node purposely (to make it choose new better peers).

On almost all other occasions, stopping the node and restarting it will put it the Blockchain line back into NOT SYNCED correctly.  

Why is this bad?

It's misleading to the user: one line (syncer) says SYNCING yncing, while the Blockchain line says Synced.  They can't both be right.

The current meaning of Blockchain SYNCED is more like "node has been sync'd sometime in the past, but is not guaranteed to be sync'd now, unless Syncer says Idle".

2 Error depictions:
![Syncing with long 65 Sec block add time, Blockchain still says Synced 2022-08-26 065422](https://user-images.githubusercontent.com/95465010/187586170-cde609b6-a0da-4e11-9ab4-953ef5139ee0.png)

![Blockchain Falsely Synced while syncing after fresh node start 2022-08-29 180448](https://user-images.githubusercontent.com/95465010/187586338-3904a7bb-088f-49f1-bbd8-f70dc23e0591.png)

Correct operation while syncing after node restart:
![Restart sped up block re-syncing, correctly shows Not Synced in Blockchain 2022-08-26 070558](https://user-images.githubusercontent.com/95465010/187586427-a42990fa-65fe-40b5-8b3d-b6c3751542c9.png)

Correctly shows Blockchain Synced when it is, and Syncer is idle.

![Blockchain Synced, Syncer Idle  All good   2022-08-26 083347](https://user-images.githubusercontent.com/95465010/187586575-7a6a8575-d4d1-4b38-86a7-c4498b33d977.png)

How to reproduce:

Easiest is to disconnect the WLAN or LAN connection to the node for several minutes,  so it loses sync.   Reconnect LAN and observe `ironfish status` showing Syncer: Syncing, and Blockchain should still say Synced incorrectly.

### Version

[...]

PR #2176, opened 2022-09-13
Add time since HEAD timestamp to status command

danield9tqh:
## Summary
Addressing [this issue](https://github.com/iron-fish/ironfish/issues/2147) about the SYNCED  field on the status command being confusing. Changing the output of that status to show the time since the head block (e.g. 30s, 10m etc). I think that would be more informative that just saying chain is SYNCED which we determine by checking if the chain is less than 5 hours behind

## Testing Plan
Local tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/status.ts
+++ b/ironfish-cli/src/commands/status.ts
@@ -108,9 +108,11 @@ function renderStatus(content: GetNodeStatusResponse, debugOutput: boolean): str
     content.peerNetwork.peers
   }`
 
-  const blockchainStatus = `HEAD ${content.blockchain.head}, behind: ${TimeUtils.renderSpan(
-    Date.now() - content.blockchain.headTimestamp,
-  )} (${content.blockchain.synced ? 'SYNCED' : 'NOT SYNCED'})`
+  const blockchainStatus = `HEAD @ ${
+    content.blockchain.head
+  }, Since HEAD: ${TimeUtils.renderSpan(Date.now() - content.blockchain.headTimestamp)} (${
+    content.blockchain.synced ? 'SYNCED' : 'NOT SYNCED'
+  })`
 
   let miningDirectorStatus = `${content.miningDirector.status.toUpperCase()} - ${
     content.miningDirector.miners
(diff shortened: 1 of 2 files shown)


ironfish/src/rpc/routes/node/getStatus.ts
```diff
@@ -129,6 +130,7 @@ export const GetStatusResponseSchema: yup.ObjectSchema<GetNodeStatusResponse> =
       .object({
         synced: yup.boolean().defined(),
         head: yup.string().defined(),
+        head_timestamp: yup.number().defined(),
```
NullSoldier, line 133: This convention is a totally new naming convention

ironfish-cli/src/commands/status.ts
```diff
@@ -108,9 +108,9 @@ function renderStatus(content: GetNodeStatusResponse, debugOutput: boolean): str
     content.peerNetwork.peers
   }`
 
-  const blockchainStatus = `${content.blockchain.synced ? 'SYNCED' : 'NOT SYNCED'} @ HEAD ${
-    content.blockchain.head
-  }`
+  const blockchainStatus = `HEAD ${content.blockchain.head}, behind: ${TimeUtils.renderSpan(
```
NullSoldier, line 111: I also don't know exactly how this will answer any user's questions exactly but I'm willing to experiment to see what will happen. I imagine these will come up

 - Users will file bug reports because status always says the word "behind" even when it's fully caught up.
 - It says I'm behind 5 hours but it says I'm synced

danield9tqh: Yea I thought of those questions coming up as well. Especially the wording of `behind` but didn't think of anything better. One solution would just be only add `SYNCED` `NOT SYNCED` when using `--all`. I think HEAD is still fine because that hash and  behind time are describing the `HEAD` block

ironfish/src/rpc/routes/node/getStatus.ts
danield9tqh: Yea sorry.. fixed this

ironfish-cli/src/commands/status.ts
danield9tqh: Ok updated this a bit to be slightly less confusing

NullSoldier (approved):

PR #2180, opened 2022-09-14
Cleanup Sapling references

mat-if:
## Summary

We pass this around in a lot of places. These were probably more relevant before the librustzcash upgrades, but now I think it's cleaner to just use them at the specific call-sites since there's very few of them.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/transaction/mod.rs
+++ b/ironfish-rust/src/transaction/mod.rs
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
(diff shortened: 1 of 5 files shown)


rohanjadvani (approved):

ironfish-rust/src/transaction/mod.rs
```diff
@@ -334,15 +329,18 @@ impl ProposedTransaction {
     }
 }
 
+impl Default for ProposedTransaction {
```
mat-if, line 332: Not sure why clippy didn't complain about this before, but rust-analyzer can auto-generate this for you, nice.

dguenther (approved):

PR #2181, opened 2022-09-14
refactor(rust): Move note content and preimage building to util file

rohanjadvani:
## Summary

There's duplicated logic when computing the pre-image from asset info or building note contents for hashing. This change moves the repeated code into a utils file. I spent too long debugging a test and it was a minor difference between the two blocks of code.

## Testing Plan

Run this unit test for sanity:
```sh

running 1 test
test proofs::circuit::mint_asset::test::test_create_mint_asset_and_spend_circuit ... ok
```

The mint constraints are broken; I will address that in a subsequent PR.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/proofs/circuit/create_asset.rs
+++ b/ironfish-rust/src/proofs/circuit/create_asset.rs
@@ -1,18 +1,17 @@
-use std::slice;
-
 use bellman::{
     gadgets::{blake2s, boolean},
     Circuit,
 };
-use zcash_primitives::constants::{GH_FIRST_BLOCK, VALUE_COMMITMENT_GENERATOR_PERSONALIZATION};
+use zcash_primitives::constants::VALUE_COMMITMENT_GENERATOR_PERSONALIZATION;
 use zcash_proofs::{
     circuit::{ecc, pedersen_hash},
     constants::NOTE_COMMITMENT_RANDOMNESS_GENERATOR,
 };
 
-use crate::primitives::{asset_type::AssetInfo, constants::ASSET_IDENTIFIER_PERSONALIZATION};
-
-use super::sapling::slice_into_boolean_vec_le;
+use crate::{
+    primitives::{asset_type::AssetInfo, constants::ASSET_IDENTIFIER_PERSONALIZATION},
+    proofs::circuit::util::hash_asset_info_to_preimage,
+};
 
 pub struct CreateAsset {
     pub asset_info: Option<AssetInfo>,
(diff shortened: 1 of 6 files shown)


ironfish-rust/src/proofs/circuit/create_asset.rs
```diff
@@ -26,56 +25,7 @@ impl Circuit<bls12_381::Scalar> for CreateAsset {
...
-            self.asset_info
-                .as_ref()
-                .and_then(|i| slice::from_ref(i.nonce()).into()),
-            8,
-        )?;
-
-        assert_eq!(nonce_bits.len(), 8);
-
-        combined_preimage.extend(nonce_bits);
+        let combined_preimage = hash_asset_info_to_preimage(cs, self.asset_info)?;
```
mat-if, line 28: We should probably use a namespace every time we pass the constraint system into a function, to avoid possible duplicate witness names. By adding this extra namespace, it'll change from like
`../booleanize first block` to `../asset info preimage/booleanize first block`

```suggestion
        let combined_preimage = hash_asset_info_to_preimage(cs.namespace(|| "asset info preimage"), self.asset_info)?;
```

ironfish-rust/src/proofs/circuit/mint_asset.rs
```diff
@@ -171,29 +136,14 @@ impl Circuit<bls12_381::Scalar> for MintAsset {
...
-            cs.namespace(|| "asset_generator"),
-            asset_type
-                .as_ref()
-                .and_then(|at| at.asset_generator().into()),
-        )?;
 
-        let value_commitment_generator = ecc::EdwardsPoint::witness(
-            cs.namespace(|| "value commitment generator"),
-            asset_type
+        let public_address_bits = slice_into_boolean_vec_le(
```
mat-if, line 140: One downside here is that we're duplicating work which means more constraints and a slower circuit - I guess we'll have to find the right balance. I was thinking about adding an extra param to the preimage helper fn like `(..., public_address_bits: Option<Vec<Boolean>>)` but that makes things quite a bit more complex. We will probably need to think about this a bit more

ironfish-rust/src/proofs/circuit/mint_asset.rs
```diff
@@ -51,42 +51,7 @@ impl Circuit<bls12_381::Scalar> for MintAsset {
...
-        )?;
-
-        let nonce_bits = slice_into_boolean_vec_le(
-            cs.namespace(|| "booleanize asset info nonce"),
-            self.asset_info
-                .as_ref()
-                .and_then(|i| slice::from_ref(i.nonce()).into()),
-            8,
-        )?;
+        let identifier_commitment_contents = hash_asset_info_to_preimage(cs, self.asset_info)?;
```
mat-if, line 54:
```suggestion
        let identifier_commitment_contents = hash_asset_info_to_preimage(cs.namespace(|| "asset info preimage"), self.asset_info)?;
```

ironfish-rust/src/proofs/circuit/mint_asset.rs
```diff
@@ -317,50 +262,8 @@ impl Circuit<bls12_381::Scalar> for MintAsset {
...
-
-        assert_eq!(
-            note_contents.len(),
-            256 + // asset_generator
-            64 + // value
-            256 + // g_d
-            256 // pk_d
-        );
+        let (note_contents, _) =
+            build_note_contents(cs, asset_type, self.value_commitment, g_d, pk_d)?;
```
mat-if, line 266:
```suggestion
            build_note_contents(cs.namespace(|| "note contents preimage"), asset_type, self.value_commitment, g_d, pk_d)?;
```

ironfish-rust/src/proofs/circuit/spend.rs
```diff
@@ -154,73 +153,10 @@ impl Circuit<bls12_381::Scalar> for Spend {
...
-
-        assert_eq!(
-            note_contents.len(),
-            256 + // asset_generator
-            64 + // value
-            256 + // g_d
-            256 // pk_d
-        );
+        let (note_contents, value_num) =
+            build_note_contents(cs, self.asset_type, self.value_commitment, g_d, pk_d)?;
```
mat-if, line 159:
```suggestion
            build_note_contents(cs.namespace(|| "note contents preimage"), self.asset_type, self.value_commitment, g_d, pk_d)?;
```

mat-if: I imagine the changes to this file also mean we'll need to re-generate parameters for the Spend circuit

mat-if (approved):

ironfish-rust/src/proofs/circuit/mint_asset.rs
rohanjadvani: Yeah, I added a TODO for now. I think we'll revisit this with upcoming refactor so I'll defer to then.

Issue #2148, opened 2022-08-31
The pool crashes when reconnecting


NullSoldier:
### What happened?

When the pool was reconnecting to the node because the node shut down, the pool crashed.

### Version

0.1.45 @ src

### Debug output

```shell
Iron Fish version               0.1.45 @ src
Iron Fish library               0.0.22 @ src
Operating system                Darwin arm64
CPU model(s)                    Apple M1 Pro
CPU threads                     10
RAM total                       16.00 GiB
Heap total                      3.85 GiB
Node version                    v16.15.1
ironfish in PATH                true
Garbage Collector Exposed       true
Telemetry enabled               true
Node name                       JASON💪💎🚀
Block graffiti                  JASON💪💎🚀
Accounts head hash              000000000003d2a4842ba88c82168f0a3d6a79d8fa38b76f0bbf90853f6e08c9
Accounts head in chain          true
Accounts head sequence          27729
```

### Relevant log output

```shell
Disconnected from node unexpectedly. Reconnecting.

/Users/jasonspafford/Projects/fish/ironfish/src/rpc/clients/ipcClient.ts:139
      request.reject(new RpcConnectionLostError(request.type))
                     ^

Error: account/getBalance
    at Client.RpcIpcClient.onDisconnect (/Users/jasonspafford/Projects/fish/ironfish/src/rpc/clients/ipcClient.ts:139:22)
    at Client.emit (/Users/jasonspafford/Projects/fish/node_modules/event-pubsub/es5.js:74:21)
    at Socket.connectionClosed (/Users/jasonspafford/Projects/fish/node_modules/node-ipc/dao/client.js:175:24)
    at Socket.emit (node:events:527:28)
    at Socket.emit (node:domain:475:12)
    at Pipe.<anonymous> (node:net:709:12)
```

### Graffiti

JASON💪💎🚀

PR #2183, opened 2022-09-14
Fix RPC stream kills process on disconnect

danield9tqh:
## Summary
Reported [here](https://github.com/iron-fish/ironfish/issues/2148). When a server disconnects from a client during a streaming request it causes the client to crash. This is because the handling of the disconnect error (`.catch()`) is not attached until after `reject` is called on the request. This causes the `reject` to be treated as an unhandled promise reject. This default behavior for this in node is to crash the process. This fix adds a `.catch` statement so that even if the error is ignored it still will not crash the process

## Testing Plan
Local testing

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/response.ts
+++ b/ironfish/src/rpc/response.ts
@@ -49,8 +49,10 @@ export class RpcResponse<TEnd = unknown, TStream = unknown> {
    * not propagated
    */
   async *contentStream(ignoreClose = true): AsyncGenerator<TStream, void> {
-    this.promise.catch((e) => {
-      throw e
+    this.promise.catch(() => {
+      // In the streaming case the error is piped through the stream instead
+      // and we handle it there (below). The same error is piped through this promise
+      // but since we are already handling it in the stream we can ignore this one
     })
 
     if (this.timeout) {
--- a/ironfish/src/rpc/response.ts
+++ b/ironfish/src/rpc/response.ts
@@ -34,6 +34,12 @@ export class RpcResponse<TEnd = unknown, TStream = unknown> {
     timeout: SetTimeoutToken | null,
   ) {
     this.promise = promise
+    this.promise.catch(() => {
+      // Eat the exception so there isn't an unhandled exception
+      // error that crashes the process. You can still read/catch
+      // the exception using waitForEnd().catch() or try..catch
+      // around contentStream() for non-RpcConnectionLostError errors
+    })
     this.stream = stream
     this.timeout = timeout
   }
@@ -43,20 +49,25 @@ export class RpcResponse<TEnd = unknown, TStream = unknown> {
     return this as RpcResponseEnded<TEnd>
   }
 
-  async *contentStream(ignoreClose = true): AsyncGenerator<TStream, void> {
+  /*
+   * Returns a generator of stream results. If a disconnect error occurs during
+   * the streaming request it just causes the generator to end, the error is
+   * not propagated
+   */
+  async *contentStream(): AsyncGenerator<TStream, void> {
     if (this.timeout) {
       clearTimeout(this.timeout)
     }
 
-    for await (const value of this.stream) {
-      yield value
-    }
-
-    await this.promise.catch((e) => {
-      if (e instanceof RpcConnectionLostError && ignoreClose) {
+    try {
+      for await (const value of this.stream) {
+        yield value
+      }
+    } catch (e) {
+      if (e instanceof RpcConnectionLostError) {
         return
       }
       throw e
-    })
+    }
   }
 }
--- a/ironfish/src/rpc/response.ts
+++ b/ironfish/src/rpc/response.ts
@@ -34,12 +34,6 @@ export class RpcResponse<TEnd = unknown, TStream = unknown> {
     timeout: SetTimeoutToken | null,
   ) {
     this.promise = promise
-    this.promise.catch(() => {
-      // Eat the exception so there isn't an unhandled exception
-      // error that crashes the process. You can still read/catch
-      // the exception using waitForEnd().catch() or try..catch
-      // around contentStream() for non-RpcConnectionLostError errors
-    })
     this.stream = stream
     this.timeout = timeout
   }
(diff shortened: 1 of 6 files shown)


ironfish/src/rpc/response.ts
```diff
@@ -34,6 +34,12 @@ export class RpcResponse<TEnd = unknown, TStream = unknown> {
     timeout: SetTimeoutToken | null,
   ) {
     this.promise = promise
+    this.promise.catch(() => {
+      // Eat the exception so there isn't an unhandled exception
+      // error that crashes the process. You can still read/catch
+      // the exception using waitForEnd().catch() or try..catch
+      // around contentStream() for non-RpcConnectionLostError errors
+    })
```
NullSoldier, line 42: Are we sure we want this exception to be eaten for all requests? If you start a request and don't wait for it to end do we not want those to bubble as an unhandled rejection?

ironfish/src/rpc/response.ts
```diff
@@ -43,20 +49,25 @@ export class RpcResponse<TEnd = unknown, TStream = unknown> {
     return this as RpcResponseEnded<TEnd>
   }
 
-  async *contentStream(ignoreClose = true): AsyncGenerator<TStream, void> {
+  /*
+   * Returns a generator of stream results. If a disconnect error occurs during
+   * the streaming request it just causes the generator to end, the error is
+   * not propagated
+   */
+  async *contentStream(): AsyncGenerator<TStream, void> {
```
NullSoldier, line 57: Do we want to reduce this functionality for users? Before this, we let users decide before if they care if a disconnect occurs or not. Some people may want to handle the disconnect case themselves. Now it's not possible to know that a response stream was terminated because of the connection ending.

ironfish/src/rpc/stream.ts
```diff
@@ -1,32 +1,66 @@
...
+ *
+ * Because Stream implements AsyncIterable it can also be used in for...in loops
+ */
 export class Stream<T> implements AsyncIterable<T> {
   buffer: T[] = []
-  waiting: PromiseResolve<IteratorResult<T>>[] = []
+  waiting: [PromiseResolve<IteratorResult<T>>, PromiseReject][] = []
+
+  // Stream can either be open or closed
+  // It can be closed because of an error or just closed normally
```
NullSoldier, line 35: I think this whole comment doesn't seem useful to me personally

```suggestion
```

ironfish/src/rpc/stream.ts
```diff
@@ -36,12 +70,16 @@ export class Stream<T> implements AsyncIterable<T> {
       return Promise.resolve({ done: false, value: value as T })
     }
 
+    if (this.error) {
+      return Promise.reject(this.error)
+    }
```
NullSoldier, line 75: With this, it's possible to read a valid value from a stream that's been ended with an error. Is that okay? That's potentially very strange behaviour that a stream with closed = true can still yield values.

ironfish/src/rpc/stream.ts
```diff
@@ -1,32 +1,66 @@
...
+ * data will be queued up in the `buffer` list
+ *
+ * stream.write(1) -> [1]                            data is queued up in the buffer
+ * stream.write(2) -> [1, 2]
+ * stream.write(3) -> [1, 2, 3]
+ * stream.next()   -> [2, 3]                         data is consumed from the buffer and resolved to next()
+ * stream.next()   -> [3]
+ *
+ * Because Stream implements AsyncIterable it can also be used in for...in loops
+ */
```
NullSoldier, line 29: I think large documentation like this is a bit strange. Do you need an explanation on how to use something as simple as this?

```
interface Stream<T> {
    write()
    next()
    close()
}
```

I read the docs but they kind of just re-describe how an async stream is implemented but that's already clear because the class is called Stream<T>. Is it because this is called Stream<T> and not AsyncStream<T>?

NullSoldier:
I just pushed up a fix the PR.

ironfish/src/rpc/stream.ts
danield9tqh: Yea it was pretty confusing for me initially. Now that I understand it it makes sense but I figured this would help when we come back to it or someone new comes in. It's a simple interface but internal workings I don't think are clear just by looking at the interface. Like what happens when write and next are not called in step. I also think it's fine to provide a comment because this seems like a pretty reusable component that's unlikely to change going forward. If it was just a one-off pattern I wouldn't provide a comment.

ironfish/src/rpc/stream.ts
danield9tqh: Will change this thanks for noticing

ironfish/src/rpc/stream.ts
danield9tqh: Ok I can remove this comment

ironfish/src/rpc/response.ts
danield9tqh: It wasn't used anywhere which is why I removed it. + when things are in the potential refactor path I think it makes it easier if all sugar coating is kind of stripped away before so it's easier to refactor. But yea I can add it back in just so we don't forget about it

ironfish/src/rpc/stream.ts
```diff
@@ -1,36 +1,72 @@
...
       return
     }
 
     this.buffer.push(value)
   }
 
-  close(): void {
+  close(e?: unknown): void {
     this.closed = true
+    this.error = e
```
NullSoldier, line 54: We just need to clear the buffer here too

NullSoldier (approved):

NullSoldier:
Looks good, when this merges we'll should deploy this to AWS to test it out.

ironfish/src/rpc/stream.ts
danield9tqh: @NullSoldier Actually I just changed this so that `next()` still returns values if there are values in the buffer. Only after the buffer is closed will it start returning errors (nor null if it was closed without an error). I think this makes more sense conceptually and that's how it was working before as well

Issue #2185, opened 2022-09-15
[ironfish-rust]Duplicate Randomnesses are calculated


hairtail:
### What happened?

Source code here, [mine_batch](https://github.com/iron-fish/ironfish/blob/master/ironfish-rust/src/mining/mine.rs).
Example: 
```
thread_count = 32
step_size = 32
batch_size = 100
search space of each threads should be:
1st: (0, 100), (100, 200)...  
2nd: (1, 101), (101, 201)...
...
32th: (31, 131), (131, 231)...
Randomnesses during first round for each threads should be:
1st: [0, 32, 64, 96]
2nd: [1, 33, 65, 97]
4th: [4, 36, 68, 100]
5th: [5, 37, 69, 101]
Randomnesses during second round for each threads should be:
1st: [100, 132, 164, 196]
2nd: [101, 133, 165, 197]
[100, 101, ...] are calculated twice.
```

### Version

0.1.46 @ src

### Graffiti

miningKing

mat-if:
Thanks for reporting this - definitely a bug. https://github.com/iron-fish/ironfish/pull/2187 should fix it

hairtail:
With #2187, the example above still exists.

scuwan:
Perhaps it's better to split the search space this way: 
subspace_size = u64::MAX/thread_count;
[subspace_size * 0.. subspace_size*1)... [subspace_size * id.. subspace_SIZE * (id +1)]... [subspace_size * (thread_count - 1)..subspace_size * thread_count)

mat-if:
> With #2187, the example above still exists.

I just did some more testing, and I don't think it was. By default, the range is not inclusive, so while it would show an end that overlapped, that number would never get executed in the first loop. However, I've since refactored it to make it more clear by making it inclusive and a smaller `end` reduced by `step_size`.

mat-if:
> Perhaps it's better to split the search space this way: subspace_size = u64::MAX/thread_count; [subspace_size * 0.. subspace_size*1)... [subspace_size * id.. subspace_size * (id +1)]... [subspace_size * (thread_count - 1)..subspace_size * thread_count)

While simpler at this level, that creates a hard loop that will never stop to check if it's received a new command. It's probably possible to accomplish the same task, but it would lead to more complex code I think since you now have to figure out a way to manage thread state from the main thread

scuwan:
I found this bug because my mine pool was receiving duplicate shares from miners.

scuwan:
[...]

PR #2187, opened 2022-09-15
Fix overlapping values for mining randomness

mat-if:
## Summary

Closes https://github.com/iron-fish/ironfish/issues/2185

Previously, the miner would check values at the batch size boundaries more than once, leading to wasted work. This change prevents that from happening by making the end value less than the start of the next batch size

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/mining/mine.rs
+++ b/ironfish-rust/src/mining/mine.rs
@@ -86,43 +86,66 @@ mod test {
         let header_bytes_base = &mut (0..128).collect::<Vec<u8>>();
         let target = &[0u8; 32];
         let mut start = 0;
-        let batch_size = 12;
-        let step_size = 2;
+        let batch_size = 10;
+        let step_size = 3;
+        // This is a helper var to hold the value of the start of the next batch basically
+        // Batch 1 should only test i values between start (0) and batch_size - 1 (9)
+        // Batch 2 should only test i values between 10 and 19, etc.
+        let mut expected_end = batch_size;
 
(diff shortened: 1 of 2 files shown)


ironfish-rust/src/mining/mine.rs
```diff
@@ -77,6 +81,50 @@ mod test {
         assert_eq!(result.unwrap(), 43);
     }
 
+    #[test]
+    fn test_mine_batch_step_size() {
```
dguenther, line 85: This test seems to pass on the old code as well -- I might be misunderstanding which issue it's fixing, but could you modify the test so it fails on the old code?

dguenther:
I made a quick spreadsheet to demonstrate the issue + this fix. In the current code, we:
* Initialize step_size to the number of threads
* Initialize batch_size to some large number
* Give each thread a unique starting value from 0 to (step_size - 1)
* Each thread hashes values from start to start + batch_size - 1, in increments of step_size
* When a thread completes a batch, increase its start by batch_size

The below example demonstrates this. You can see that some values are hashed twice:

image

The change here is to update the following:

* Each thread hashes values from start to start + batch_size - step_size, in increments of step_size

This addresses the duplicate hashing, but does skip some values in the process.

image


I believe an alternative that would include all values would be the following:
* Each thread hashes values from start to start + batch_size (inclusive), in increments of step_size
* When a thread completes a batch, increase its start by batch_size + step_size - (batch_size % step_size)

This is a bit less easy to test, however. For example:
image

dguenther (approved):

PR #2188, opened 2022-09-15
evict transactions cache from account

hughy:
## Summary

reads transactions from database instead of holding all transactions in memory.

- converts all reads of transactions to use database
- removes test of loading transactions into cache
- refactors saveDecryptedNoteSequence to take a note hash and a sequence instead of loading a transaction

## Testing Plan

- updated unit tests
- ran `accounts:transactions` on migrated pool account

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/account.ts
+++ b/ironfish/src/wallet/account.ts
@@ -179,6 +179,7 @@ export class Account {
       }
 
       this.decryptedNotes.set(noteHash, note)
+      await this.accountsDb.saveDecryptedNote(this, noteHash, note, tx)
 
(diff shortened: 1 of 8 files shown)


ironfish/src/wallet/account.ts
```diff
@@ -172,6 +166,7 @@ export class Account {
   ): Promise<void> {
     await this.accountsDb.database.withTransaction(tx, async (tx) => {
       const existingNote = this.decryptedNotes.get(noteHash)
+
```
NullSoldier, line 169: 😍

NullSoldier (approved):

PR #2189, opened 2022-09-15
Implement Rust Rolling Bloom Filter with xxhash

mat-if:
## Summary

This is around 10x faster than using the pure JS RollingFilter/murmurhash

Converting from JS to Rust gave us a speedup in the realm of 5x, and converting murmur hash to xxhash3 was another ~5x (Thanks @dguenther!). Adding in napi bindings and such brought the finally speed-up tally to around 10x, with a test of BloomFilter size 1 million, FP rate of 0.0000001 and 5 million `.add`/`.test` calls

This is largely based off of bitcoin's Rolling Bloom Filter implementation: https://github.com/bitcoin/bitcoin/blob/master/src/common/bloom.cpp

Closes IRO-2589

## Breaking Change

```
[ ] Yes
```


ironfish/src/network/peerNetwork.ts
```diff
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-import { RollingFilter } from '@ironfish/bfilter'
```
dguenther, line 5: We can pull this out of the package.json now, and probably delete the typedefs for bfilter.d.ts too.

dguenther (approved):

PR #2191, opened 2022-09-15
pass sequence through to updateDecryptedNote

hughy:
## Summary

avoids loading transaction from database for each note

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/account.ts
+++ b/ironfish/src/wallet/account.ts
@@ -139,6 +139,7 @@ export class Account {
   async updateDecryptedNote(
     noteHash: Buffer,
     note: Readonly<DecryptedNoteValue>,
+    sequence: number | null,
     tx?: IDatabaseTransaction,
   ): Promise<void> {
     await this.accountsDb.database.withTransaction(tx, async (tx) => {
@@ -157,8 +158,7 @@ export class Account {
 
       await this.accountsDb.saveDecryptedNote(this, noteHash, note, tx)
 
-      const transaction = await this.getTransaction(note.transactionHash, tx)
-      this.saveDecryptedNoteSequence(noteHash, transaction?.sequence ?? null)
+      this.saveDecryptedNoteSequence(noteHash, sequence)
     })
   }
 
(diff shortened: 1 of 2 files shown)


ironfish/src/wallet/account.ts
```diff
@@ -157,8 +158,7 @@ export class Account {
 
       await this.accountsDb.saveDecryptedNote(this, noteHash, note, tx)
 
-      const transaction = await this.getTransaction(note.transactionHash, tx)
-      this.saveDecryptedNoteSequence(noteHash, transaction?.sequence ?? null)
+      this.saveDecryptedNoteSequence(noteHash, sequence)
```
NullSoldier, line 161: this does break, beause the wrong sequence is being passed in some places. I haven't tracked it down yet but the test failure is an accurate new bug that's created.

ironfish/src/wallet/account.ts
```diff
@@ -205,8 +205,8 @@ export class Account {
       }
 
       const isRemovingTransaction = submittedSequence === null && blockHash === null
-      await this.bulkUpdateDecryptedNotes(transactionHash, decryptedNotes, tx)
-      await this.processTransactionSpends(transaction, isRemovingTransaction, tx)
+      await this.bulkUpdateDecryptedNotes(transactionHash, sequence, decryptedNotes, tx)
+      await this.processTransactionSpends(transaction, sequence, isRemovingTransaction, tx)
```
NullSoldier, line 209: I think in some cases, this is clearing out the sequence causing notes to get moved off chain. If you look at line 200, it only happens if the transaction should update but if the transaction hasn't changed and no sequence is passed then will this clear it?

ironfish/src/wallet/wallet.ts
```diff
@@ -701,10 +701,14 @@ export class Accounts {
...
-                ...noteMapValue,
-                spent: true,
-              })
+              await sender.updateDecryptedNote(
+                unspentNote.hash,
+                {
+                  ...noteMapValue,
+                  spent: true,
+                },
+                null,
```
NullSoldier, line 710: This will break any notes that are being fixed up. It'll be marked as spent but it'll be set to an off chain note.

hughy: You're right, we need to look up the transaction for the note here to get the right sequence.

hughy:
I'm wondering if this is the right approach after all since we still need to read transactions from the database to get the sequence in multiple cases.

NullSoldier:
Ok, let's close this for now.

PR #2210, opened 2022-09-19
Move account in memory caches to accounts DB

NullSoldier:
## Summary

This PR will remove `sequenceToNoteHashes` and `nonChainNoteHashes`. These two are being moved to database stores that will be fetched from using range queries.

## Testing Plan

 - Added new tests
 - Rerun migration

## Breaking Change

```
[x] Yes
```

### DIFF
--- a/ironfish/src/wallet/database/accountsdb.ts
+++ b/ironfish/src/wallet/database/accountsdb.ts
@@ -10,14 +10,18 @@ import { Nullifier } from '../../primitives/nullifier'
 import { TransactionHash } from '../../primitives/transaction'
 import {
   BigIntLEEncoding,
+  BUFFER_ENCODING,
   BufferEncoding,
   IDatabase,
   IDatabaseStore,
   IDatabaseTransaction,
+  NULL_ENCODING,
   NullableBufferEncoding,
   PrefixEncoding,
   StringEncoding,
+  U32_ENCODING,
 } from '../../storage'
+import { StorageUtils } from '../../storage/database/utils'
 import { createDB } from '../../storage/utils'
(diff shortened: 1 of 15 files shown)


ironfish/src/wallet/database/accountsdb.ts
```diff
@@ -65,6 +69,16 @@ export class AccountsDB {
     value: Buffer
   }>
 
+  sequenceToNoteHash: IDatabaseStore<{
+    key: [Account['prefix'], [number, TransactionHash]]
```
hughy, line 73: `TransactionHash` should be some kind of note hash type or Buffer

ironfish/src/wallet/database/accountsdb.ts
```diff
@@ -123,6 +137,22 @@ export class AccountsDB {
...
+      keyEncoding: new PrefixEncoding(
+        new PrefixEncoding(new BufferEncoding(), U32_ENCODING, 32),
+        new BufferEncoding(),
+        4,
+      ),
+      valueEncoding: NULL_ENCODING,
+    })
+
+    this.nonChainNoteHashes = this.database.addStore({
+      name: 's',
```
hughy, line 151: Duplicate name from above

hughy:
Do we need to update the migration for these new stores?

ironfish/src/wallet/database/accountsdb.ts
NullSoldier: fixed

ironfish/src/wallet/database/accountsdb.ts
NullSoldier: fixed

PR #2211, opened 2022-09-19
Add service:forks command to submit forks count to telemetry API

mirayadav:
## Summary

Add a new service command `service:forks` to submit a `forks_count` metric to the telemetry API so that we can create a corresponding Influx dashboard. Emit an `onBlockGossipReceived` event for new compact blocks and orphan blocks to make the forks count calculation more accurate. Update `onGossipResponse` to only include the block header data.

## Testing Plan

Tested the command locally with an Influx notebook.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/chain/forks.ts
+++ b/ironfish-cli/src/commands/chain/forks.ts
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { GraffitiUtils, PromiseUtils, TARGET_BLOCK_TIME_IN_SECONDS } from '@ironfish/sdk'
-import { RpcBlock } from '@ironfish/sdk'
+import { RpcBlockHeader } from '@ironfish/sdk'
 import blessed from 'blessed'
 import { IronfishCommand } from '../../command'
 import { RemoteFlags } from '../../flags'
@@ -23,7 +23,7 @@ export default class ForksCommand extends IronfishCommand {
     let connected = false
     const forks = new Map<
       string,
-      { block: RpcBlock; time: number; mined: number; old?: boolean }
+      { header: RpcBlockHeader; time: number; mined: number; old?: boolean }
     >()
 
     const screen = blessed.screen({ smartCSR: true })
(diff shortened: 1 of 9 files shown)


ironfish-cli/src/commands/service/forks.ts
```diff
@@ -0,0 +1,124 @@
...
+            measurement: 'forks_count',
+            timestamp: new Date(),
+            fields: [
+              {
+                name: 'forks',
+                type: 'integer',
+                value: count,
+              },
+            ],
+            tags: [{ name: 'version', value: '0.1.24' }],
```
rohanjadvani, line 89: Can we use the default tag set [here](https://github.com/iron-fish/ironfish/blob/master/ironfish/src/node.ts#L116)?

ironfish-cli/src/commands/service/forks.ts
```diff
@@ -0,0 +1,124 @@
...
+          },
+        ],
+        graffiti: 'forks_count_metric',
```
rohanjadvani, line 92: `graffiti` is set on the node as a configuration, so I don't think we should hard-code this value. We can update the `submitTelemetry` method to omit this field since it is [optional](https://github.com/iron-fish/ironfish-api/blob/master/src/telemetry/dto/write-telemetry-points.dto.ts#L73-L75) in our API.

ironfish-cli/src/commands/service/forks.ts
mirayadav: Updated the version tag to use the current ironfish-cli package version

ironfish-cli/src/commands/service/forks.ts
mirayadav: Updated the `submitTelemetry` method to make the graffiti parameter optional

ironfish/src/webApi.ts
```diff
@@ -210,7 +210,8 @@ export class WebApi {
     return response.data
   }
 
-  async submitTelemetry(payload: { points: Metric[]; graffiti: string }): Promise<void> {
+  async submitTelemetry(points: Metric[], graffiti?: string): Promise<void> {
```
rohanjadvani, line 213: Does it work if we leave the function signature mostly the same, but just make the field optional? Seems cleaner and would remove the `graffiti` undefined check below.

```suggestion
  async submitTelemetry(payload: { points: Metric[]; graffiti?: string }): Promise<void> {
```

mirayadav: Yup, changed it to just making the field optional. (I tried this yesterday and it didn't work but I realized this was due to another bug which is now fixed.)

NullSoldier (approved):

ironfish-cli/src/commands/chain/forks.ts
```diff
@@ -93,16 +93,16 @@ export default class ForksCommand extends IronfishCommand {
...
       const mined = prev ? prev.mined + 1 : 0
 
       if (prev) {
         prev.old = true
-        forks.set(block.previousBlockHash, prev)
+        forks.set(header.previousBlockHash, prev)
       }
 
-      forks.set(block.hash, { block: block, time: Date.now(), mined: mined })
+      forks.set(header.hash, { header: header, time: Date.now(), mined: mined })
```
rohanjadvani, line 105:
```suggestion
      forks.set(header.hash, { header, time: Date.now(), mined })
```

rohanjadvani (approved):
One minor nit. Otherwise looks great

PR #2213, opened 2022-09-19
Add chain:benchmark command for testing add block time

dguenther:
## Summary

Adds a `chain:benchmark` command that copies blocks from the current node to a new node to benchmark the time it takes to add blocks.

## Breaking Change

```
[ ] Yes
```


mat-if (approved):
Nice this is way easier than `rm -rf ~/.benchmark; cp ~/.benchmark-base ~/.benchmark; fish start` that I usually do

NullSoldier (approved):

Issue #2215, opened 2022-09-20
[ShallowCopy] RecalculateTarget changes miningRequestBlocks of current miningRequestId


hairtail:
### What happened?

[SourceCodeHere](https://github.com/iron-fish/ironfish/blob/master/ironfish/src/mining/pool.ts#:~:text=const%20latestBlock%20%3D%20this.miningRequestBlocks.get(this.nextMiningRequestId%20%2D%201))
```
const latestBlock = this.miningRequestBlocks.get(this.nextMiningRequestId - 1)
latestBlock.header.target = BigIntUtils.toBytesBE(newTarget.asBigInt(), 32).toString('hex')
latestBlock.header.timestamp = newTime.getTime()
// timeA
// target & timestamp of this.miningRequestBlocks.get(this.nextMiningRequestId - 1) changes
this.distributeNewBlock(latestBlock)
// timeB
```
Valid shares submitted between timeA and timeB are marked as Invalid.

### Version

0.1.47 @ src

### Graffiti

miningKing

mat-if:
Fixed in #2216 🙌 

Thanks!

PR #2216, opened 2022-09-20
Fix: recalculateTraget bug caused by shallow copy

hairtail:
## Summary
Fix https://github.com/iron-fish/ironfish/issues/2215
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/mining/pool.ts
+++ b/ironfish/src/mining/pool.ts
@@ -337,7 +337,11 @@ export class MiningPool {
     Assert.isNotNull(this.currentHeadTimestamp)
     Assert.isNotNull(this.currentHeadDifficulty)
 
-    const latestBlock = this.miningRequestBlocks.get(this.nextMiningRequestId - 1)
+    const currentBlock = this.miningRequestBlocks.get(this.nextMiningRequestId - 1)
+    Assert.isNotNull(currentBlock)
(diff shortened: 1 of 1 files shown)


ironfish/src/mining/pool.ts
```diff
@@ -337,7 +337,9 @@ export class MiningPool {
     Assert.isNotNull(this.currentHeadTimestamp)
     Assert.isNotNull(this.currentHeadDifficulty)
 
-    const latestBlock = this.miningRequestBlocks.get(this.nextMiningRequestId - 1)
+    const currentBlock = this.miningRequestBlocks.get(this.nextMiningRequestId - 1)
+    const latestBlock = JSON.parse(JSON.stringify(currentBlock)) as typeof currentBlock
```
mat-if, line 341: What do you think about the approach used here? https://github.com/iron-fish/ironfish/blob/master/ironfish/src/mining/pool.ts#L223-L224

```suggestion
    const latestBlock = Object.assign({}, currentBlock)
    latestBlock.header = Object.assign({}, currentBlock.header)
```

I suspect the JSON stringify and parse is slower than `Object.assign`, but it also makes me a bit nervous as it could introduce subtle serialization bugs

hairtail: Updated.

mat-if:
Thanks!

PR #2223, opened 2022-09-20
Have DecryptedNoteValue encoder handle Note

NullSoldier:
## Summary
We're just doing the same here that we did to TransactionValue already.

## Testing Plan
Run migration again

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/migrations/data/013-wallet-2/new/decryptedNotes.ts
+++ b/ironfish/src/migrations/data/013-wallet-2/new/decryptedNotes.ts
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import bufio from 'bufio'
+import { Note, NOTE_LENGTH } from '../../../../primitives/note'
 import { IDatabaseEncoding, IDatabaseStore } from '../../../../storage'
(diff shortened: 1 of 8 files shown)


ironfish/src/migrations/data/013-wallet-2/new/decryptedNotes.ts
```diff
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import bufio from 'bufio'
+import { Note, NOTE_LENGTH } from '../../../../primitives/note'
 import { IDatabaseEncoding, IDatabaseStore } from '../../../../storage'
 
 export const NOTE_SIZE = 43 + 8 + 32 + 32
```
hughy, line 8: Can we delete `NOTE_SIZE` here too?

hughy (approved):
Looks good to me pending updates to tests that depend on `DecryptedNoteValue.serializedNote`

ironfish/src/migrations/data/013-wallet-2/new/decryptedNotes.ts
NullSoldier: Deleted

PR #2224, opened 2022-09-20
Add basic auth to RPC

ygao76:
## Summary
https://linear.app/ironfish/issue/IRO-1496/add-basic-authentication-to-rpc-layer
- Add basic auth to RPC
- Store the auth token in internal.json config file
- Add auth field in RpcRequest
- TcpClient and TlsClient will send request with auth token
- Read token from file in SocketAdaptor
- Create a new rpc.auth flag to pass auth token in cli

More details on this coda page https://coda.io/d/_dJrIPvc0s6j/RPC-Auth_suc7i#_luDqc
## Testing Plan
unit tests
## Breaking Change

```
[x ] Yes
```
@iron-fish/engineering @NullSoldier @danield9tqh @hughy

### DIFF
--- a/ironfish-cli/src/command.ts
+++ b/ironfish-cli/src/command.ts
@@ -112,7 +112,7 @@ export abstract class IronfishCommand extends Command {
     const configFlag = getFlag(flags, ConfigFlagKey)
 
     const configOverrides: Partial<ConfigOptions> = {}
-    const internalConfigOverrides: Partial<InternalOptions> = {}
+    const internalOverrides: Partial<InternalOptions> = {}
 
     const databaseNameFlag = getFlag(flags, DatabaseFlagKey)
     if (typeof databaseNameFlag === 'string' && databaseNameFlag !== DatabaseFlag.default) {
@@ -159,12 +159,13 @@ export abstract class IronfishCommand extends Command {
 
     const rpcAuthFlag = getFlag(flags, RpcAuthFlagKey)
     if (typeof rpcAuthFlag === 'string') {
-      internalConfigOverrides.rpcAuthToken = rpcAuthFlag
+      internalOverrides.rpcAuthToken = rpcAuthFlag
     }
 
     this.sdk = await IronfishSdk.init({
       pkg: IronfishCliPKG,
       configOverrides: configOverrides,
+      internalOverrides: internalOverrides,
       configName: typeof configFlag === 'string' ? configFlag : undefined,
       dataDir: typeof dataDirFlag === 'string' ? dataDirFlag : undefined,
       logger: this.logger,
--- a/ironfish-cli/src/command.ts
+++ b/ironfish-cli/src/command.ts
@@ -18,7 +18,6 @@ import {
   DatabaseFlag,
   DatabaseFlagKey,
   DataDirFlagKey,
-  RpcAuthFlag,
   RpcAuthFlagKey,
   RpcTcpHostFlagKey,
   RpcTcpPortFlagKey,
--- a/ironfish-cli/src/command.ts
+++ b/ironfish-cli/src/command.ts
@@ -17,6 +17,8 @@ import {
   DatabaseFlag,
   DatabaseFlagKey,
   DataDirFlagKey,
+  RpcAuthFlag,
+  RpcAuthFlagKey,
   RpcTcpHostFlagKey,
   RpcTcpPortFlagKey,
   RpcTcpSecureFlag,
@@ -46,6 +48,7 @@ export type FLAGS =
   | typeof RpcTcpSecureFlagKey
   | typeof RpcTcpTlsFlagKey
   | typeof VerboseFlagKey
+  | typeof RpcAuthFlagKey
 
 export abstract class IronfishCommand extends Command {
   // Yes, this is disabling the type system but any code
@@ -151,6 +154,11 @@ export abstract class IronfishCommand extends Command {
     const verboseFlag = getFlag(flags, VerboseFlagKey)
     if (typeof verboseFlag === 'boolean' && verboseFlag !== VerboseFlag.default) {
       configOverrides.logLevel = '*:verbose'
+    }
+
+    const rpcAuthFlag = getFlag(flags, RpcAuthFlagKey)
+    if (typeof rpcAuthFlag === 'string') {
+      configOverrides.rpcAuthToken = rpcAuthFlag
     }
 
     this.sdk = await IronfishSdk.init({
--- a/ironfish-cli/src/command.ts
+++ b/ironfish-cli/src/command.ts
@@ -6,6 +6,7 @@ import {
   createRootLogger,
   DatabaseVersionError,
   ErrorUtils,
+  InternalOptions,
   IronfishSdk,
   Logger,
   RpcConnectionError,
@@ -112,6 +113,7 @@ export abstract class IronfishCommand extends Command {
     const configFlag = getFlag(flags, ConfigFlagKey)
 
     const configOverrides: Partial<ConfigOptions> = {}
+    const internalConfigOverrides: Partial<InternalOptions> = {}
 
     const databaseNameFlag = getFlag(flags, DatabaseFlagKey)
     if (typeof databaseNameFlag === 'string' && databaseNameFlag !== DatabaseFlag.default) {
@@ -158,7 +160,7 @@ export abstract class IronfishCommand extends Command {
 
     const rpcAuthFlag = getFlag(flags, RpcAuthFlagKey)
     if (typeof rpcAuthFlag === 'string') {
-      configOverrides.rpcAuthToken = rpcAuthFlag
+      internalConfigOverrides.rpcAuthToken = rpcAuthFlag
     }
 
     this.sdk = await IronfishSdk.init({
--- a/ironfish/src/fileStores/config.ts
+++ b/ironfish/src/fileStores/config.ts
@@ -106,6 +106,7 @@ export type ConfigOptions = {
   rpcTcpSecure: boolean
   tlsKeyPath: string
   tlsCertPath: string
+  rpcAuthTokenPath: string
   /**
    * The maximum number of peers we can be connected to at a time. Past this number,
    * new connections will be rejected.
@@ -304,6 +305,7 @@ export class Config extends KeyStore<ConfigOptions> {
       rpcTcpSecure: false,
       tlsKeyPath: files.resolve(files.join(dataDir, 'certs', 'node-key.pem')),
       tlsCertPath: files.resolve(files.join(dataDir, 'certs', 'node-cert.pem')),
+      rpcAuthTokenPath: files.resolve(files.join(dataDir, 'certs', 'rpc-auth-token.txt')),
       maxPeers: 50,
       minimumBlockConfirmations: 12,
       minPeers: 1,
--- a/ironfish/src/fileStores/config.ts
+++ b/ironfish/src/fileStores/config.ts
@@ -1,6 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { uniqueId } from 'lodash'
 import * as yup from 'yup'
 import { FileSystem } from '../fileSystems'
 import { KeyStore } from './keyStore'
@@ -106,7 +107,7 @@ export type ConfigOptions = {
   rpcTcpSecure: boolean
   tlsKeyPath: string
   tlsCertPath: string
-  rpcAuthTokenPath: string
+  rpcAuthToken: string
   /**
    * The maximum number of peers we can be connected to at a time. Past this number,
    * new connections will be rejected.
@@ -305,7 +306,7 @@ export class Config extends KeyStore<ConfigOptions> {
       rpcTcpSecure: false,
       tlsKeyPath: files.resolve(files.join(dataDir, 'certs', 'node-key.pem')),
       tlsCertPath: files.resolve(files.join(dataDir, 'certs', 'node-cert.pem')),
-      rpcAuthTokenPath: files.resolve(files.join(dataDir, 'certs', 'rpc-auth-token.txt')),
+      rpcAuthToken: '',
       maxPeers: 50,
       minimumBlockConfirmations: 12,
       minPeers: 1,
--- a/ironfish/src/fileStores/config.ts
+++ b/ironfish/src/fileStores/config.ts
@@ -1,7 +1,6 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { uniqueId } from 'lodash'
 import * as yup from 'yup'
 import { FileSystem } from '../fileSystems'
 import { KeyStore } from './keyStore'
@@ -107,7 +106,6 @@ export type ConfigOptions = {
   rpcTcpSecure: boolean
   tlsKeyPath: string
   tlsCertPath: string
-  rpcAuthToken: string
   /**
    * The maximum number of peers we can be connected to at a time. Past this number,
    * new connections will be rejected.
@@ -306,7 +304,6 @@ export class Config extends KeyStore<ConfigOptions> {
       rpcTcpSecure: false,
       tlsKeyPath: files.resolve(files.join(dataDir, 'certs', 'node-key.pem')),
       tlsCertPath: files.resolve(files.join(dataDir, 'certs', 'node-cert.pem')),
-      rpcAuthToken: '',
       maxPeers: 50,
       minimumBlockConfirmations: 12,
       minPeers: 1,
--- a/ironfish/src/rpc/adapters/errors.ts
+++ b/ironfish/src/rpc/adapters/errors.ts
@@ -9,6 +9,7 @@ export enum ERROR_CODES {
   ROUTE_NOT_FOUND = 'route-not-found',
   VALIDATION = 'validation',
   INSUFFICIENT_BALANCE = 'insufficient-balance',
+  UNAUTHENTICATION = 'unauthentication'
 }
 
 /**
--- a/ironfish/src/rpc/adapters/errors.ts
+++ b/ironfish/src/rpc/adapters/errors.ts
@@ -9,7 +9,7 @@ export enum ERROR_CODES {
   ROUTE_NOT_FOUND = 'route-not-found',
   VALIDATION = 'validation',
   INSUFFICIENT_BALANCE = 'insufficient-balance',
-  UNAUTHENTICATION = 'unauthentication',
+  UNAUTHENTICATED = 'unauthenticated',
 }
 
 /**
--- a/ironfish/src/rpc/adapters/errors.ts
+++ b/ironfish/src/rpc/adapters/errors.ts
@@ -9,7 +9,7 @@ export enum ERROR_CODES {
   ROUTE_NOT_FOUND = 'route-not-found',
   VALIDATION = 'validation',
   INSUFFICIENT_BALANCE = 'insufficient-balance',
-  UNAUTHENTICATION = 'unauthentication'
+  UNAUTHENTICATION = 'unauthentication',
 }
 
 /**
--- a/ironfish/src/rpc/adapters/ipcAdapter.test.ts
+++ b/ironfish/src/rpc/adapters/ipcAdapter.test.ts
@@ -26,6 +26,9 @@ describe('IpcAdapter', () => {
         enableRpc: false,
         enableRpcIpc: false,
       },
+      internalConfigOverrides: {
+        rpcAuthToken: 'test token',
+      },
     })
 
     ipc = new RpcIpcAdapter(ALL_API_NAMESPACES, {
@@ -78,7 +81,7 @@ describe('IpcAdapter', () => {
     await ipc.start()
     await client.connect()
 
-    const response = client.request('foo/bar', 'test token')
+    const response = client.request('foo/bar')
     expect((await response.contentStream().next()).value).toBe('hello 1')
     expect((await response.contentStream().next()).value).toBe('hello 2')
 
@@ -96,7 +99,7 @@ describe('IpcAdapter', () => {
     await ipc.start()
     await client.connect()
 
-    const next = client.request('foo/bar', 'test token').contentStream().next()
+    const next = client.request('foo/bar').contentStream().next()
 
     client.close()
     waitResolve()
@@ -113,7 +116,7 @@ describe('IpcAdapter', () => {
     await ipc.start()
     await client.connect()
 
-    const response = client.request('foo/bar', 'test token')
+    const response = client.request('foo/bar')
 
     await expect(response.waitForEnd()).rejects.toThrowError(RpcRequestError)
     await expect(response.waitForEnd()).rejects.toMatchObject({
@@ -134,7 +137,7 @@ describe('IpcAdapter', () => {
     await ipc.start()
     await client.connect()
 
-    const response = client.request('foo/bar', 'test token', body)
+    const response = client.request('foo/bar', body)
 
     await expect(response.waitForEnd()).rejects.toThrowError(RpcRequestError)
     await expect(response.waitForEnd()).rejects.toMatchObject({
--- a/ironfish/src/rpc/adapters/ipcAdapter.test.ts
+++ b/ironfish/src/rpc/adapters/ipcAdapter.test.ts
@@ -12,135 +12,135 @@ import { ALL_API_NAMESPACES } from '../routes/router'
 import { ERROR_CODES, ValidationError } from './errors'
 import { RpcIpcAdapter } from './ipcAdapter'
 
-describe('IpcAdapter', () => {
-  let ipc: RpcIpcAdapter
-  let sdk: IronfishSdk
-  let client: RpcIpcClient
+// describe('IpcAdapter', () => {
+//   let ipc: RpcIpcAdapter
+//   let sdk: IronfishSdk
+//   let client: RpcIpcClient
 
-  beforeEach(async () => {
-    const dataDir = os.tmpdir()
+//   beforeEach(async () => {
+//     const dataDir = os.tmpdir()
 
-    sdk = await IronfishSdk.init({
-      dataDir,
-      configOverrides: {
-        enableRpc: false,
-        enableRpcIpc: false,
-      },
-    })
+//     sdk = await IronfishSdk.init({
+//       dataDir,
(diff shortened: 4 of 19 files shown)


ironfish-cli/src/commands/chain/benchmark.ts
```diff
@@ -0,0 +1,104 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
```
ygao76, line 2: Not sure how did this file got included in my changes

ironfish/src/rpc/adapters/socketAdapter/protocol.ts
```diff
@@ -45,6 +46,7 @@ export const ClientSocketRpcSchema: yup.ObjectSchema<ClientSocketRpc> = yup
       .object({
         mid: yup.number().required(),
         type: yup.string().required(),
+        auth: yup.string().required(),
```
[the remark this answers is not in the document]
hairtail: For publicNameSpace like 'chain/getBlock', it's unnecessary to check auth. Auth is only needed for protectedNameSpace. So it should be an optional attribute. This is my opinion https://github.com/iron-fish/ironfish/pull/2162/

ironfish/src/rpc/adapters/ipcAdapter.test.ts
```diff
@@ -78,7 +78,7 @@ describe('IpcAdapter', () => {
     await ipc.start()
     await client.connect()
 
-    const response = client.request('foo/bar')
+    const response = client.request('foo/bar', 'test token')
```
NullSoldier, line 81: Can we inject this into the client and have that auto insert this into each request instead of having to pass it into each request here?

ironfish/src/rpc/adapters/socketAdapter/protocol.ts
NullSoldier, line 49: Will non authed sockets always pass this with an empty string?

ironfish/src/rpc/adapters/socketAdapter/socketAdapter.ts
```diff
@@ -35,6 +36,8 @@ export abstract class RpcSocketAdapter implements IRpcAdapter {
   server: net.Server | null = null
   router: Router | null = null
   namespaces: ApiNamespace[]
+  rpcAuthTokenPath: string
+  fileSystem: FileSystem
```
NullSoldier, line 40: I think we should move this auth level up to the RpcServer, so we don't have to do this in every adapter we want to implement auth for.

ironfish/src/fileStores/config.ts
```diff
@@ -304,6 +305,7 @@ export class Config extends KeyStore<ConfigOptions> {
       rpcTcpSecure: false,
       tlsKeyPath: files.resolve(files.join(dataDir, 'certs', 'node-key.pem')),
       tlsCertPath: files.resolve(files.join(dataDir, 'certs', 'node-cert.pem')),
+      rpcAuthTokenPath: files.resolve(files.join(dataDir, 'certs', 'rpc-auth-token.txt')),
```
NullSoldier, line 308: I see you want to store this in it's own file, versus using one of our existing config stores. Is there some reason why we don't use internal.json or config.json?

ironfish/src/rpc/clients/tcpClient.ts
```diff
@@ -70,13 +75,14 @@ export class RpcTcpClient extends RpcSocketClient {
     }
   }
 
-  protected send(messageId: number, route: string, data: unknown): void {
+  protected send(messageId: number, route: string, authToken: string, data: unknown): void {
```
NullSoldier, line 78: Would be nice if we didn't have to pass this every time

ironfish/src/rpc/adapters/socketAdapter/socketAdapter.ts
```diff
@@ -193,6 +200,22 @@ export abstract class RpcSocketAdapter implements IRpcAdapter {
...
+        this.logger.debug(`Missing RPC Auth token files at ${this.rpcAuthTokenPath}.`)
+        this.emitResponse(client, this.constructUnauthenticatedRequest())
+        return
+      }
+
+      const rpcAuthToken = await this.fileSystem.readFile(this.rpcAuthTokenPath)
+      if (requestAuthToken !== rpcAuthToken) {
+        this.emitResponse(client, this.constructUnauthenticatedRequest())
+        return
+      }
```
NullSoldier, line 217: If we move this into RpcServer, we wouldn't have to duplicate this for each adapter we want to do this.

ironfish/src/rpc/clients/ipcClient.ts
```diff
@@ -24,10 +24,11 @@ export class RpcIpcClient extends RpcSocketClient {
 
   constructor(
     connection: Partial<RpcClientConnectionInfo> = {},
+    authToken: string,
```
NullSoldier, line 27: I don't think we want authentication in the IPC protocol right?

ironfish/src/rpc/adapters/socketAdapter/protocol.ts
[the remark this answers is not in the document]
ygao76: Update the `auth` to notRequired so its backward compatible.

ironfish/src/fileStores/config.ts
ygao76: Updated. store the token in internal.json

ironfish/src/rpc/clients/ipcClient.ts
ygao76: removed

ironfish/src/rpc/adapters/tcpAdapter.test.ts
```diff
@@ -131,4 +134,30 @@ describe('TcpAdapter', () => {
       codeMessage: expect.stringContaining('this must be defined'),
     })
   })
+
+  it('should authenticate', async () => {
```
hughy, line 138: It would be good to add a test that confirms that authentication matches when the tokens match

ironfish/src/rpc/server.ts
```diff
@@ -73,4 +76,21 @@ export class RpcServer {
       this._startPromise = promise
     }
   }
+
+  /** Authenticate the RPC request */
+  authenticate(requestAuthToken: string | undefined): boolean {
+    if (requestAuthToken) {
```
hughy, line 82: Should this be `if (!requestAuthToken)`? It looks like `authenticate` will always return false

ironfish/src/rpc/adapters/tlsAdapter.ts
```diff
@@ -37,6 +42,15 @@ export class RpcTlsAdapter extends RpcSocketAdapter {
...
     const nodeCertExists = await this.fileSystem.exists(this.nodeCertPath)
+    const rpcAuthToken = this.node.internal.get('rpcAuthToken')
+
+    if (!rpcAuthToken || rpcAuthToken === '') {
+      this.logger.debug(
+        `Missing RPC Auth token in internal.json config. Automatically generating auth token.`,
+      )
+      this.node.internal.set('rpcAuthToken', uuid())
+      await this.node.internal.save()
+    }
```
hughy, line 53: This is implemented in both the TlsAdapter and below in the SDK; do we need both?

Passing the node into the adapter creates a circular reference

ironfish/src/sdk.ts
```diff
@@ -133,14 +146,37 @@ export class IronfishSdk {
     }
 
     let client: RpcSocketClient
+    let rpcAuthToken = internal.get('rpcAuthToken')
```
hughy, line 149: I'm worried that automatically generating the token for the client could lead to some confusion.

If I pull down these changes on a remote node, then it will automatically generate a token for the adapter, but when I try to connect with a local client, then it will automatically generate a different mismatched token.

ironfish/src/rpc/adapters/errors.ts
```diff
@@ -9,6 +9,7 @@ export enum ERROR_CODES {
   ROUTE_NOT_FOUND = 'route-not-found',
   VALIDATION = 'validation',
   INSUFFICIENT_BALANCE = 'insufficient-balance',
+  UNAUTHENTICATION = 'unauthentication',
```
hughy, line 12: nit - `unauthentication` looks odd. Maybe `unauthenticated` or `unauthorized`?

ironfish/src/rpc/clients/socketClient.ts
```diff
@@ -34,10 +35,21 @@ export abstract class RpcSocketClient extends RpcClient {
   abstract client: IpcClient | net.Socket | null
   abstract isConnected: boolean
   abstract connection: Partial<RpcClientConnectionInfo>
+  rpcAuthToken: string
```
hughy, line 38: nit - `authToken` since it's a property of `RpcSocketClient`?

ironfish-cli/src/command.ts
```diff
@@ -109,6 +112,7 @@ export abstract class IronfishCommand extends Command {
     const configFlag = getFlag(flags, ConfigFlagKey)
 
     const configOverrides: Partial<ConfigOptions> = {}
+    const internalConfigOverrides: Partial<InternalOptions> = {}
```
NullSoldier, line 115: This says config, but config is separate from internal.

```suggestion
    const internalOverrides: Partial<InternalOptions> = {}
```

ironfish-cli/src/command.ts
```diff
@@ -153,6 +157,11 @@ export abstract class IronfishCommand extends Command {
       configOverrides.logLevel = '*:verbose'
     }
 
+    const rpcAuthFlag = getFlag(flags, RpcAuthFlagKey)
+    if (typeof rpcAuthFlag === 'string') {
```
NullSoldier, line 161: We'll need to check the default here like other flags I think.

ironfish/src/rpc/adapters/errors.ts
NullSoldier: Yes we should copy HTTP here probably which is `Unauthorized`, https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401

ironfish/src/rpc/adapters/ipcAdapter.test.ts
```diff
@@ -26,6 +26,9 @@ describe('IpcAdapter', () => {
         enableRpc: false,
         enableRpcIpc: false,
       },
+      internalConfigOverrides: {
+        rpcAuthToken: 'test token',
```
NullSoldier, line 30: we don't need this since IPC has no auth

ironfish/src/rpc/adapters/socketAdapter/socketAdapter.ts
```diff
@@ -58,11 +60,13 @@ export abstract class RpcSocketAdapter implements IRpcAdapter {
     port: number,
     logger: Logger = createRootLogger(),
     namespaces: ApiNamespace[],
+    enableAuthentication: boolean,
```
NullSoldier, line 63: I don't think we should take this as a required parameter since we default it in the class

ironfish/src/rpc/adapters/socketAdapter/socketAdapter.ts
```diff
@@ -193,6 +197,21 @@ export abstract class RpcSocketAdapter implements IRpcAdapter {
 
       const message = result.result.data
 
+      if (this.router == null || this.router.server == null) {
+        this.emitResponse(client, this.constructUnmountedAdapter())
+        return
+      }
```
NullSoldier, line 203: This has to go back where it was to properly respond to the message and handle it properly.

ironfish/src/rpc/adapters/socketAdapter/socketAdapter.ts
```diff
@@ -193,6 +197,21 @@ export abstract class RpcSocketAdapter implements IRpcAdapter {
 
       const message = result.result.data
 
+      if (this.router == null || this.router.server == null) {
+        this.emitResponse(client, this.constructUnmountedAdapter())
```
NullSoldier, line 201: I was wondering why this moved, but I think I understand why. You saw it doesn't respond or use the request at all, so you moved it above right? Now that this code is moving, it makes me realize that this is a small bug where it will never reject the incoming request on the client side in this case.

ironfish/src/rpc/adapters/socketAdapter/socketAdapter.ts
```diff
@@ -193,6 +197,21 @@ export abstract class RpcSocketAdapter implements IRpcAdapter {
...
+        return
+      }
+
+      // Authentication
```
NullSoldier, line 205:
```suggestion
```

I don't think we need this comment since it's fairly obvious what this is doing.

ironfish/src/rpc/adapters/socketAdapter/socketAdapter.ts
```diff
@@ -316,4 +330,16 @@ export abstract class RpcSocketAdapter implements IRpcAdapter {
...
+    const error = new Error(`Missing or bad authentication`)
+
+    const data: SocketRpcError = {
+      code: ERROR_CODES.UNAUTHENTICATION,
+      message: error.message,
+      stack: error.stack,
+    }
+
+    return this.constructMessage(request.mid, 401, data)
+  }
```
NullSoldier, line 344: This is for another PR, but I think if we just made a new method to construct a response to a message from an error, we could probably reuse it for this case, and the `ResponseError` on line 231.

ironfish/src/rpc/adapters/tcpAdapter.test.ts
```diff
@@ -53,7 +56,7 @@ describe('TcpAdapter', () => {
       request.end(request.data)
     })
 
-    client = new RpcTcpClient('localhost', tcp.addressPort)
+    client = new RpcTcpClient('localhost', tcp.addressPort, 'test token')
```
NullSoldier, line 59: I think we shouldn't pass this in these tests. We should test that we don't need to pass this in the old tests to show that they still pass.

ironfish/src/rpc/adapters/tcpAdapter.test.ts
```diff
@@ -72,7 +75,7 @@ describe('TcpAdapter', () => {
       request.end()
     })
 
-    client = new RpcTcpClient('localhost', tcp.addressPort)
+    client = new RpcTcpClient('localhost', tcp.addressPort, 'test token')
```
NullSoldier, line 78: Here too, remove this.

ironfish/src/rpc/adapters/tcpAdapter.test.ts
NullSoldier: yes we should be able to also make a request here to check that authentication does work too

ironfish/src/rpc/adapters/tlsAdapter.ts
```diff
@@ -4,29 +4,34 @@
...
     port: number,
     fileSystem: FileSystem,
     nodeKeyPath: string,
     nodeCertPath: string,
+    node: IronfishNode,
     logger: Logger = createRootLogger(),
     namespaces: ApiNamespace[],
   ) {
-    super(host, port, logger, namespaces)
+    super(host, port, logger, namespaces, true)
```
NullSoldier, line 30: I don't really know what this true is, it would be a bit cleaner if it just over-rode authentication below on line 35.

ironfish/src/rpc/adapters/tlsAdapter.ts
NullSoldier, line 53: If we add another adapter that uses the auth, and don't mount the TlsAdapter, then authentication will break because this is the only place that generates a token. We should do it in a more general place like the RpcServer because that's the only place that consumes the token.

ironfish/src/rpc/clients/ipcClient.ts
```diff
@@ -23,11 +23,12 @@ export class RpcIpcClient extends RpcSocketClient {
   onError = new Event<[error: unknown]>()
 
   constructor(
+    rpcAuthToken: string,
```
NullSoldier, line 26: ipc client doesn't use auth so we should remove this

ironfish/src/rpc/clients/socketClient.ts
```diff
@@ -34,10 +35,21 @@ export abstract class RpcSocketClient extends RpcClient {
...
+  }
 
   abstract connect(options?: Record<string, unknown>): Promise<void>
   abstract close(): void
-  protected abstract send(messageId: number, route: string, data: unknown): void
+  protected abstract send(
+    messageId: number,
+    route: string,
+    data: unknown,
+    authToken: string,
```
NullSoldier, line 51: I think we should remove this authToken from here and have it passed down through the constructor

ironfish/src/rpc/clients/socketClient.ts
NullSoldier: This should be nullable since it's not required in a client.

ironfish/src/rpc/clients/socketClient.ts
```diff
@@ -34,10 +35,21 @@ export abstract class RpcSocketClient extends RpcClient {
   abstract client: IpcClient | net.Socket | null
   abstract isConnected: boolean
   abstract connection: Partial<RpcClientConnectionInfo>
+  rpcAuthToken: string
+
+  constructor(logger: Logger, rpcAuthToken: string) {
```
NullSoldier, line 40:
```suggestion
  constructor(logger: Logger, authToken?: string) {
```

ironfish/src/rpc/clients/tcpClient.test.ts
```diff
@@ -23,6 +23,7 @@ describe('IronfishTcpClient', () => {
         data: {
           mid: messageId,
           type: route,
+          auth: 'test token',
```
NullSoldier, line 26: I don't think we should pass this in any old test, to ensure they work as is. Especially since it should not be required.

NullSoldier, line 26: we don't have a test to test that this works when auth is undefined, which I think is fine. What do you think?

ironfish/src/rpc/clients/tcpClient.ts
```diff
@@ -23,8 +23,13 @@ export class RpcTcpClient extends RpcSocketClient {
   connection: RpcClientConnectionInfo
   private messageBuffer: MessageBuffer
 
-  constructor(host: string, port: number, logger: Logger = createRootLogger()) {
-    super(logger.withTag('tcpclient'))
+  constructor(
+    host: string,
+    port: number,
+    rpcAuthTokoen: string,
```
NullSoldier, line 29: This is a typo, and I don't think it should be required since auth is not required. I also think we can just name this `authToken` since this is already in the RpcClient.

```suggestion
   authToken: string,
```

ironfish/src/rpc/server.ts
```diff
@@ -73,4 +76,21 @@ export class RpcServer {
...
+      this.logger.debug(`Missing Auth token in RPC request.`)
```
NullSoldier, line 83: we don't want this because it can let an attacker flood our logs

ironfish/src/rpc/server.ts
```diff
@@ -73,4 +76,21 @@ export class RpcServer {
...
+      return false
+    }
+
+    const rpcAuthToken = this.node.internal.get('rpcAuthToken')
+
+    if (rpcAuthToken || rpcAuthToken === '') {
```
NullSoldier, line 89: This will always go into this branch because any non empty string is truthy

ironfish/src/rpc/server.ts
```diff
@@ -73,4 +76,21 @@ export class RpcServer {
...
+      this.logger.debug(`Missing RPC Auth token in internal.json config.`)
```
NullSoldier, line 90: I think this should be a warning. It means they are trying to use an authenticated adapter without configuring their RpcAuthToken

ironfish/src/sdk.ts
```diff
@@ -67,6 +74,7 @@ export class IronfishSdk {
     pkg,
     configName,
     configOverrides,
+    internalConfigOverrides,
```
NullSoldier, line 77: this isn't config, its internal so it should be just `internalOverrides`

```suggestion
    internalOverrides,
```

ironfish/src/sdk.ts
```diff
@@ -133,14 +146,37 @@ export class IronfishSdk {
...
+        client = new RpcTcpClient(
+          config.get('rpcTcpHost'),
+          config.get('rpcTcpPort'),
+          rpcAuthToken,
+          logger,
+        )
       }
     } else {
       client = new RpcIpcClient(
+        rpcAuthToken,
```
NullSoldier, line 179: ipc client does not use token auth, so it should not be passed here.

ironfish/src/rpc/server.ts
```diff
@@ -73,4 +76,21 @@ export class RpcServer {
...
   }
+
+  /** Authenticate the RPC request */
+  authenticate(requestAuthToken: string | undefined): boolean {
+    if (requestAuthToken) {
+      this.logger.debug(`Missing Auth token in RPC request.`)
+      return false
+    }
+
+    const rpcAuthToken = this.node.internal.get('rpcAuthToken')
```
NullSoldier, line 87: Here's a security question, but should we let the token change after the server starts, or should we cache this when the server starts and require you to restart the RpcServer to pick up the new token?

ironfish/src/rpc/clients/socketClient.ts
NullSoldier, line 51: Technically we don't need to pass this through since any inheritors can access it, but this is fine. Though the type here is wrong because it should not be required.

ironfish/src/rpc/adapters/tlsAdapter.ts
ygao76: I will keep the one in SDK and delete this one in TlsAdapter

ironfish/src/sdk.ts
ygao76: If the server and client are not on the same machine, it wont match, which is expected behavior. We want to stop unknown client connect to server. 
The user need to copy paste the internal.json file from server to their local client to pass authentication

ironfish/src/rpc/adapters/errors.ts
ygao76: I will use `unauthenticated` 
HTTP code is technically not accurate, see the post here http://web.archive.org/web/20190904190534/https://www.dirv.me/blog/2011/07/18/understanding-403-forbidden/index.html
`Unauthorized` should be used when server know who are the user of the request, but the user don’t have permission to access this resource.

ironfish/src/rpc/adapters/ipcAdapter.test.ts
ygao76: removed

ironfish/src/rpc/adapters/tlsAdapter.ts
ygao76: If we add another adapter, it should write the token generator code if they use auth token.

ironfish/src/rpc/clients/ipcClient.ts
ygao76: removed

ironfish/src/rpc/clients/tcpClient.test.ts
ygao76: added

ironfish/src/rpc/server.ts
ygao76: I want to keep the debug log for now in case I need to debug. I will clean it once the feature is launched successfully

ironfish/src/rpc/server.ts
ygao76: It should be ```if(!rpcAuthToken)``` updated

ironfish/src/rpc/server.ts
ygao76: If rpc auth token is empty, it's an internal error because we should have automatically generated the token at beginning.
If user use an authenticated adapter, they should not need to configure if they are running server and client on the same machine. 
If they run them on separate machine, they will receive unauthenticated error. They will need to copy the internal.json to client

ironfish/src/rpc/adapters/socketAdapter/socketAdapter.ts
ygao76: I will keep the comment, the logic for onClientData is already very bloated with a lot of if statement. This help the code readability.

ironfish/src/sdk.ts
hughy: > If the server and client are not on the same machine, it wont match, which is expected behavior. We want to stop unknown client connect to server. The user need to copy paste the internal.json file from server to their local client to pass authentication

That's true, but generating a token in the client won't match any server. It will only match if a user later starts a server on the same machine

ygao76: ok, I removed the token generation logic since now the token is optional.

ironfish/src/rpc/adapters/tlsAdapter.ts
NullSoldier: I'm unsure of what you mean, do you mean you'll copy paste this logic into each adapter that needs to ensure that it has a token?

ironfish/src/rpc/clients/socketClient.ts
```diff
@@ -34,10 +35,23 @@ export abstract class RpcSocketClient extends RpcClient {
   abstract client: IpcClient | net.Socket | null
   abstract isConnected: boolean
   abstract connection: Partial<RpcClientConnectionInfo>
+  authToken: string | null = null
+
+  constructor(logger: Logger, authToken?: string) {
+    super(logger)
+    if (authToken) {
+      this.authToken = authToken
+    }
```
NullSoldier, line 44: We usually use this syntax and don't initialize it in the class 

```suggestion
    this.authToken = authToken ?? null
```

ironfish/src/rpc/server.ts
NullSoldier: You mentioned it's an internal error, but you still don't think you want the user to see that error and report the bug as a warning / error?

ironfish/src/sdk.ts
```diff
@@ -133,9 +145,16 @@ export class IronfishSdk {
     }
 
     let client: RpcSocketClient
+    const rpcAuthToken = internal.get('rpcAuthToken')
```
NullSoldier, line 148: any reason we assign this here instead of just passing it in directly like all of our other config values?

ygao76: hmm because currently the ```init()``` does not have internal pass in

ironfish/src/rpc/server.ts
ygao76: There are different error messages. User will see `Missing authetication` or `Bad authentication`

NullSoldier (approved):

PR #2227, opened 2022-09-20
Include hash when iterating DecryptedNotes

NullSoldier:
## Summary

If we just include the hash at the AccountDB level, we don't need to construct these objects further up. We should probably move hash to DecryptedNoteValue later.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/account.ts
+++ b/ironfish/src/wallet/account.ts
@@ -81,10 +81,10 @@ export class Account {
   }
 
   async load(): Promise<void> {
-    for await (const { hash, decryptedNote } of this.accountsDb.loadDecryptedNotes(this)) {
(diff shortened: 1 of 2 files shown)


ironfish/src/wallet/account.ts
```diff
@@ -108,14 +108,8 @@ export class Account {
     transactionHash: Buffer
     spent: boolean
   }> {
```
hughy, line 110: We could change the return type to `AsyncGenerator<DecryptedNoteValue>`

hughy (approved):

ironfish/src/wallet/account.ts
NullSoldier: changed

PR #2229, opened 2022-09-21
Add cpu usage statistics to /status command

danield9tqh:
## Summary
Adds CPU usage to the status command. Calculates the % of CPU the process is using across all cores of the operating system

## Testing Plan
Tested locally

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/metrics/cpuMeter.ts
+++ b/ironfish/src/metrics/cpuMeter.ts
@@ -1,0 +1,103 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import os from 'os'
+import { HRTime } from '../utils'
+import { RollingAverage } from './rollingAverage'
+
+/**
+ * Calculates an exponentially weighted moving average and rolling average
+ * for CPU usage percentage. This keeps track of the % of overall OS CPU the
+ * current process is using. Overall OS CPU is calculated with the sum of total CPU time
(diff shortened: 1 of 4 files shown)


NullSoldier:
This number basically has nothing to do with Iron Fish right? We're just writing all of this code to expose their CPU usage from our app so we can show it in status. We can't ever report this number in telemetry because it won't be indicative of anything related to Iron Fish.

danield9tqh:
Yea maybe percentage is the wrong measure then. I think showing the CPU usage of the process is helpful. Certainly for our own nodes it is helpful to see it in status since CPU percentage correlates with slowdown of other processes. Similar to how we show % of memory used and can correlate that with node slowness

ironfish/src/metrics/cpuMeter.ts
```diff
@@ -0,0 +1,110 @@
...
+      return
+    }
+
+    const elapsedMs = now[0] - this._lastReading.time[0]
+    const elapsedCpuTime = this.totalCpuTime(osCpu) - this.totalCpuTime(this._lastReading.osCpu)
+    const elapsedProcessUserTime = processCpu.user - this._lastReading.processCpu.user
+    const elapsedProcessSysTime = processCpu.system - this._lastReading.processCpu.system
+
+    // process time is in nanoseconds, os time is in milliseconds
+    const percProcessCpu =
```
NullSoldier, line 88: what does perc mean?

danield9tqh: percentage

pasha121 (approved):
it is ok

NullSoldier (approved):

PR #2232, opened 2022-09-21
clean up snapshot download after import

hughy:
## Summary

the snapshot file downloaded from s3 is ~50GB in size. after unzipping the snapshot and moving the database into place the snapshot file remains.

changes `chain:download` to automatically remove the snapshot file after a successful import.

- adds 'cleanup' and 'no-cleanup' options to `chain:download`
- factors download logic into separate method for readability
- adds error message to failed imports with direction on how to retry importing a downloaded snapshot

## Testing Plan

tested importing a snapshot locally with and without cleanup

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/chain/download.ts
+++ b/ironfish-cli/src/commands/chain/download.ts
@@ -50,6 +50,7 @@ export default class Download extends IronfishCommand {
       default: true,
       description: 'Remove downloaded snapshot file after import',
       allowNo: true,
+      hidden: true,
     }),
   }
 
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/chain/download.ts
```diff
@@ -40,11 +40,16 @@ export default class Download extends IronfishCommand {
...
     }),
     confirm: Flags.boolean({
       default: false,
-      description: 'Confirm without asking',
+      description: 'Confirm download without asking',
+    }),
+    cleanup: Flags.boolean({
+      default: true,
+      description: 'Remove downloaded snapshot file after import',
+      allowNo: true,
```
NullSoldier, line 52: should we add this as a hidden flag? Should most users need to use this?

NullSoldier (approved):

ironfish-cli/src/commands/chain/download.ts
hughy: Setting to hidden -- most users shouldn't need to keep snapshots around after import

PR #2237, opened 2022-09-22
unzip snapshot directly to chainDatabasePath

hughy:
## Summary

we previously unzipped the snapshot to the 'temp' directory before moving files to the chain database. unzipping directly to the chain database directory will skip the move step and may circumvent errors users have run into while moving files

- remove contents of chainDatabasePath before unzipping
- ensure that chainDatabasePath exists
- unzip directly into chainDatabasePath

## Testing Plan

local chain:download

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/chain/download.ts
+++ b/ironfish-cli/src/commands/chain/download.ts
@@ -215,21 +215,28 @@ export default class Download extends IronfishCommand {
       }
     }
 
-    const chainDatabasePath = this.sdk.fileSystem.resolve(this.sdk.config.chainDatabasePath)
+    try {
+      const chainDatabasePath = this.sdk.fileSystem.resolve(this.sdk.config.chainDatabasePath)
 
-    // chainDatabasePath must be empty before unzipping snapshot
-    CliUx.ux.action.start(
-      `Removing existing chain data at ${chainDatabasePath} before importing snapshot`,
-    )
-    await fsAsync.rm(chainDatabasePath, { recursive: true, force: true, maxRetries: 10 })
-    CliUx.ux.action.stop('done')
+      // chainDatabasePath must be empty before unzipping snapshot
+      CliUx.ux.action.start(
+        `Removing existing chain data at ${chainDatabasePath} before importing snapshot`,
+      )
+      await fsAsync.rm(chainDatabasePath, { recursive: true, force: true, maxRetries: 10 })
+      CliUx.ux.action.stop('done')
 
-    // ensure that chainDatabasePath exists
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/chain/download.ts
```diff
@@ -215,22 +215,20 @@ export default class Download extends IronfishCommand {
...
-    // chainDatabasePath must be empty before renaming snapshot
     await fsAsync.rm(chainDatabasePath, { recursive: true, force: true, maxRetries: 10 })
-    await fsAsync.rename(snapshotDatabasePath, chainDatabasePath)
+    CliUx.ux.action.stop('done')
+
+    // ensure that chainDatabasePath exists
+    await fsAsync.mkdir(chainDatabasePath, { recursive: true })
+
+    CliUx.ux.action.start(`Unzipping ${snapshotPath} to ${chainDatabasePath}`)
+    await this.unzip(snapshotPath, chainDatabasePath)
```
ygao76, line 231: Do we need to handle errors during unzip?

hughy: It's a good idea to handle errors here however we can. We can provide a message telling users what to do next if there's an error, but I'm not sure what more we can do in error handling here.

ironfish-cli/src/commands/chain/download.ts
```diff
@@ -215,23 +215,28 @@ export default class Download extends IronfishCommand {
...
+
+      // ensure that chainDatabasePath exists
+      await fsAsync.mkdir(chainDatabasePath, { recursive: true })
+
+      CliUx.ux.action.start(`Unzipping ${snapshotPath} to ${chainDatabasePath}`)
+      await this.unzip(snapshotPath, chainDatabasePath)
+      CliUx.ux.action.stop('done')
+    } catch (e) {
+      this.logger.error(
+        `Failed to import snapshot. Run \`ironfish chain:download --path ${snapshotPath}\` to retry import without download`,
```
NullSoldier, line 236: I'm usually pretty wary about adding things like this, because they can confuse users a lot but if you think it's useful I'm ok putting it in. For instance, could we remove this error and just do this automatically for the user? If we see a snapshot at the location just ask if they want to import that instead then remove this?

NullSoldier (approved):

ironfish-cli/src/commands/chain/download.ts
hughy: I see what you mean.

I think I'll remove the message here, but I think it could be tricky to automatically import a snapshot if they have multiple snapshot files lying around

PR #2240, opened 2022-09-23
Return early for addTransaction and deleteTransaction in the mempool

danield9tqh:
## Summary
If the transaction already exists in the mempool we want to return early from `addTransaction`. Similarly we want to return early from `deleteTransaction` if the transaction doesn't exist.

## Testing Plan
Unit tests and local testing

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/memPool/memPool.ts
+++ b/ironfish/src/memPool/memPool.ts
@@ -173,16 +173,14 @@ export class MemPool {
     let addedTransactions = 0
 
     for (const transaction of block.transactions) {
-      if (this.exists(transaction.hash())) {
-        continue
-      }
-
       if (transaction.isMinersFee()) {
         continue
       }
 
-      this.addTransaction(transaction)
-      addedTransactions++
+      const added = this.addTransaction(transaction)
+      if (added) {
+        addedTransactions++
+      }
     }
 
     this.logger.debug(`Added ${addedTransactions} transactions`)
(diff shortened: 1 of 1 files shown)


ironfish/src/memPool/memPool.ts
```diff
@@ -173,30 +173,31 @@ export class MemPool {
...
 
-      this.addTransaction(transaction)
-      addedTransactions++
+      const added = this.addTransaction(transaction)
+      if (added) {
+        addedTransactions++
+      }
     }
 
     this.logger.debug(`Added ${addedTransactions} transactions`)
```
ygao76, line 186: Include the debug log in the `if(added)` block?

ironfish/src/memPool/memPool.ts
```diff
@@ -173,30 +173,31 @@ export class MemPool {
...
-      }
-
       if (transaction.isMinersFee()) {
         continue
       }
 
-      this.addTransaction(transaction)
-      addedTransactions++
+      const added = this.addTransaction(transaction)
+      if (added) {
```
ygao76, line 181: anything we should do if `added` is null ?

NullSoldier (approved):

ironfish/src/memPool/memPool.ts
NullSoldier: Will that message be useful to anyone? It's expected to print that a lot, so it'll print it many many times but there's no actionable thing you can do with that message I think because it's expected.

NullSoldier: That will print 300~ lines every block that's reorged.

PR #2241, opened 2022-09-24
Fix exposing account before TX commits

NullSoldier:
## Summary

Found a bunch of issues where account was put into a map or removed from it before the DB transaction would commit leading to various bugs.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/database/accountsdb.ts
+++ b/ironfish/src/wallet/database/accountsdb.ts
@@ -181,9 +181,10 @@ export class AccountsDB {
     })
   }
 
-  async removeAccount(id: string, tx?: IDatabaseTransaction): Promise<void> {
+  async removeAccount(account: Account, tx?: IDatabaseTransaction): Promise<void> {
     await this.database.withTransaction(tx, async (tx) => {
-      await this.accounts.del(id, tx)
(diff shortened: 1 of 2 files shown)


ironfish/src/wallet/database/accountsdb.ts
```diff
@@ -181,9 +181,10 @@ export class AccountsDB {
     })
   }
 
-  async removeAccount(id: string, tx?: IDatabaseTransaction): Promise<void> {
+  async removeAccount(account: Account, tx?: IDatabaseTransaction): Promise<void> {
     await this.database.withTransaction(tx, async (tx) => {
-      await this.accounts.del(id, tx)
+      await this.accounts.del(account.id, tx)
+      await this.balances.del(account.id, tx)
```
hughy, line 187: Should we clear other account data here?

hughy (approved):

ironfish/src/wallet/database/accountsdb.ts
NullSoldier: Maybe, but that'll also take a looong time and freeze the node on a big wallet. We need to implement time sliced removal where we mark the account as deleted, then have something in the background that cleans it up over time until its all gone.

PR #2245, opened 2022-09-25
FEAT: tls mining server

hairtail:
## Summary
Tls mining channel supported. Miner SRBMiner supports tls mining, which performs better than tcp considering security. This PR supports tls mining server. Next step is tls stratum client.
## Breaking Change

```
[ ] Yes
```


dguenther:
I'd like to get this in, @hairtail  I appreciate the work getting this feature started. I wanted the architecture to be similar to the RPC server, so I felt like it'd be a bit quicker to fix it up myself vs. describing the changes I was looking for in Github comments, hopefully that's okay 👍 

I changed it to only allow either TCP or TLS, which removes the need for extra config for now, but it also removes the ability to run both at once for backward compatibility. We need to think a bit more about how to describe combinations of adapters (for example, if we added an in-memory one) using config settings without adding a lot of them, and I think that effort can be done separately from this PR.

I'll see if I can get some review on this in the next day or two.

hairtail:
> I'd like to get this in, @hairtail I appreciate the work getting this feature started. I wanted the architecture to be similar to the RPC server, so I felt like it'd be a bit quicker to fix it up myself vs. describing the changes I was looking for in Github comments, hopefully that's okay 👍
> 
> I changed it to only allow either TCP or TLS, which removes the need for extra config for now, but it also removes the ability to run both at once for backward compatibility. We need to think a bit more about how to describe combinations of adapters (for example, if we added an in-memory one) using config settings without adding a lot of them, and I think that effort can be done separately from this PR.
> 
> I'll see if I can get some review on this in the next day or two.

Cool, i like this architecture. It makes all things clear.

mat-if (approved):

PR #2246, opened 2022-09-26
Stream output for accounts:notes

mirayadav:
## Summary

Stream responses from `getAccountNotes` so that the `accounts:notes` command can render output progressively.

## Testing Plan

Tested locally by running command with simulated notes.

## Breaking Change

```
[ ] Yes
```

https://linear.app/ironfish/issue/IRO-2573/stream-responses-from-accountsnotes

### DIFF
--- a/ironfish-cli/src/commands/accounts/notes.ts
+++ b/ironfish-cli/src/commands/accounts/notes.ts
@@ -31,14 +31,14 @@ export class NotesCommand extends IronfishCommand {
     const response = client.getAccountNotesStream({ account })
     let firstResponse = true
 
-    for await (const { account: accountResponse, notes } of response.contentStream()) {
+    for await (const { account: accountResponse, note } of response.contentStream()) {
       const noHeader = firstResponse ? false : true
       if (firstResponse) {
         this.log(`\n ${accountResponse} - Account notes\n`)
         firstResponse = false
       }
       CliUx.ux.table(
-        notes,
+        [note],
         {
           isOwner: {
             header: 'Owner',
--- a/ironfish-cli/src/commands/accounts/notes.ts
+++ b/ironfish-cli/src/commands/accounts/notes.ts
@@ -29,13 +29,13 @@ export class NotesCommand extends IronfishCommand {
     const client = await this.sdk.connectRpc()
 
     const response = client.getAccountNotesStream({ account })
-    let first_response = true
+    let firstResponse = true
 
     for await (const { account: accountResponse, notes } of response.contentStream()) {
-      const no_header = first_response ? false : true
-      if (first_response) {
+      const noHeader = firstResponse ? false : true
+      if (firstResponse) {
         this.log(`\n ${accountResponse} - Account notes\n`)
-        first_response = false
+        firstResponse = false
       }
       CliUx.ux.table(
         notes,
@@ -65,7 +65,7 @@ export class NotesCommand extends IronfishCommand {
             },
           },
         },
-        { 'no-header': no_header },
+        { 'no-header': noHeader },
       )
     }
     this.log(`\n`)
--- a/ironfish-cli/src/commands/accounts/notes.ts
+++ b/ironfish-cli/src/commands/accounts/notes.ts
@@ -28,39 +28,46 @@ export class NotesCommand extends IronfishCommand {
 
     const client = await this.sdk.connectRpc()
 
-    const response = await client.getAccountNotes({ account })
+    const response = client.getAccountNotesStream({ account })
+    let first_response = true
 
-    const { account: accountResponse, notes } = response.content
-
-    this.log(`\n ${accountResponse} - Account notes\n`)
-
-    CliUx.ux.table(notes, {
-      isOwner: {
-        header: 'Owner',
-        get: (row) => (row.owner ? `✔` : `x`),
-      },
-      amount: {
-        header: 'Amount ($IRON)',
-        get: (row) => oreToIron(row.amount),
-      },
-      memo: {
-        header: 'Memo',
-      },
-      transactionHash: {
-        header: 'From Transaction',
-      },
-      isSpent: {
-        header: 'Spent',
-        get: (row) => {
-          if (row.spent === undefined) {
-            return '-'
-          } else {
-            return row.spent ? `✔` : `x`
-          }
+    for await (const { account: accountResponse, notes } of response.contentStream()) {
+      const no_header = first_response ? false : true
+      if (first_response) {
+        this.log(`\n ${accountResponse} - Account notes\n`)
+        first_response = false
+      }
+      CliUx.ux.table(
+        notes,
+        {
+          isOwner: {
+            header: 'Owner',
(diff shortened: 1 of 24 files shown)


ironfish-cli/src/commands/accounts/notes.ts
```diff
@@ -28,39 +28,46 @@ export class NotesCommand extends IronfishCommand {
 
     const client = await this.sdk.connectRpc()
 
-    const response = await client.getAccountNotes({ account })
+    const response = client.getAccountNotesStream({ account })
+    let first_response = true
```
rohanjadvani, line 32: We usually use camelCase throughout our codebase. Can you update this variable and `no_header` below?

ironfish-cli/src/commands/accounts/notes.ts
```diff
@@ -28,39 +28,46 @@ export class NotesCommand extends IronfishCommand {
 
     const client = await this.sdk.connectRpc()
 
-    const response = await client.getAccountNotes({ account })
```
rohanjadvani, line 31: Is `getAccountNotes` unused now? If so, do we still need this method?

ironfish/src/rpc/routes/accounts/getNotes.ts
```diff
@@ -50,10 +51,15 @@ router.register<typeof GetAccountNotesRequestSchema, GetAccountNotesResponse>(
     const account = getAccount(node, request.data.account)
 
     const responseNotes = []
+    const stream = request.data?.stream
 
     for await (const { transaction } of account.getTransactions()) {
       const notes = await getTransactionNotes(account, transaction)
-      responseNotes.push(...notes)
+      if (stream) {
+        request.stream({ account: account.displayName, notes: [...notes] })
```
rohanjadvani, line 59: Is `[...notes]` needed, or does `notes` work? 

```suggestion
        request.stream({ account: account.displayName, notes })
```

ironfish-cli/src/commands/accounts/notes.ts
mirayadav: The `getAccountNotes` method is unused now so I could delete the method. However, should I still leave in support for non-streaming calls to the `getAccountNotes` API? Could be helpful for future use cases, but I'm not sure if we would need to call this API anywhere else.

ironfish/src/rpc/routes/accounts/getNotes.ts
mirayadav: `notes` returned from `getTransactionNotes` is readonly, while `notes` in `GetAccountNotesResponse` is mutable so I used `note: [...notes]`. Is there a better way to get around this?

ironfish-cli/src/commands/accounts/notes.ts
rohanjadvani: I don't think we need the non-streaming implementation since notes are expected to grow overtime (unless we supported some filter or pagination). Similar to [exporting mined blocks](https://github.com/iron-fish/ironfish/blob/master/ironfish/src/rpc/routes/mining/exportMined.ts), we just stream all blocks that are returned. We can follow a similar pattern for account notes too.

ironfish/src/rpc/clients/client.ts
```diff
@@ -206,13 +206,13 @@ export abstract class RpcClient {
     ).waitForEnd()
   }
 
-  async getAccountNotes(
-    params: GetAccountNotesRequest = {},
-  ): Promise<RpcResponseEnded<GetAccountNotesResponse>> {
-    return await this.request<GetAccountNotesResponse>(
-      `${ApiNamespace.account}/getAccountNotes`,
+  getAccountNotesStream(
+    params: GetAccountNotesStreamRequest,
```
NullSoldier, line 210: This shouldn't be required

I cannot call `client.getAccountNotesStream()`

ironfish/src/rpc/routes/accounts/getNotes.ts
```diff
@@ -18,44 +18,44 @@ export type GetAccountNotesResponse = {
...
+  GetAccountNotesStreamRequestSchema,
   async (request, node): Promise<void> => {
     const account = getAccount(node, request.data.account)
 
-    const responseNotes = []
-
     for await (const { transaction } of account.getTransactions()) {
       const notes = await getTransactionNotes(account, transaction)
-      responseNotes.push(...notes)
+      request.stream({ account: account.displayName, notes: [...notes] })
```
NullSoldier, line 56: The stream output here should be each note, it looks like it's outputting a stream of chunk of notes per transaction.

hughy, line 56: It's a small bummer to have to stream back the account name each time.

We only need the account name in a response at all when we don't pass an account name to `accounts:notes` and use the default account.

hughy (approved):

ironfish/src/rpc/routes/accounts/getNotes.ts
```diff
@@ -3,59 +3,63 @@
...
-
-    for await (const { transaction } of account.getTransactions()) {
-      const notes = await getTransactionNotes(account, transaction)
-      responseNotes.push(...notes)
+    for await (const { note, spent, transactionHash } of account.getNotes()) {
+      request.stream({
+        account: account.displayName,
+        note: {
+          owner: true,
+          amount: Number(note.value()),
```
NullSoldier, line 55: We cannot cast value() to number, because that's lossy. We have to encode values as strings.

ironfish/src/rpc/routes/accounts/getNotes.ts
```diff
@@ -3,59 +3,63 @@
...
-    const responseNotes = []
-
-    for await (const { transaction } of account.getTransactions()) {
-      const notes = await getTransactionNotes(account, transaction)
-      responseNotes.push(...notes)
+    for await (const { note, spent, transactionHash } of account.getNotes()) {
+      request.stream({
+        account: account.displayName,
+        note: {
+          owner: true,
```
NullSoldier, line 54: we can get rid of this now since this only returns notes from the owner

ironfish/src/rpc/routes/accounts/getNotes.ts
```diff
@@ -3,59 +3,63 @@
...
     const account = getAccount(node, request.data.account)
 
-    const responseNotes = []
-
-    for await (const { transaction } of account.getTransactions()) {
-      const notes = await getTransactionNotes(account, transaction)
-      responseNotes.push(...notes)
+    for await (const { note, spent, transactionHash } of account.getNotes()) {
+      request.stream({
+        account: account.displayName,
```
NullSoldier, line 52: I think we should consider getting rid of this now. I was gonna remove it in the transaction stream.

mirayadav: What about the case where we call getNotes with no account and return the notes from the default account. Wouldn't we want to return the name of the default account in this case?

NullSoldier (approved):

PR #2248, opened 2022-09-26
adds output path option for snapshot download

hughy:
## Summary

some users may want to download snapshots to a location other than datadir/temp. for instance, a user might want to download the snapshot file to a separate mounted disk.

- adds `-o` option for full download path

## Testing Plan

local testing with `chain:download -o`

## Breaking Change

```
[ ] Yes
```


ygao76 (approved):

NullSoldier (approved):

NullSoldier:
Looks good, just needs a rebase.

PR #2251, opened 2022-09-26
adds progress bar on snapshot unzip

hughy:
## Summary

uses 'tar.list' to asynchronously count the number of files in the snapshot archive and the 'onentry' callback of 'tar.extract' to update a progress bar representing the progress on unzipping the snapshot archive.

## Testing Plan

chain_download_unzip_progress

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/chain/download.ts
+++ b/ironfish-cli/src/commands/chain/download.ts
@@ -61,7 +61,6 @@ export default class Download extends IronfishCommand {
     await NodeUtils.waitForOpen(node)
 
     let snapshotPath
-    await fsAsync.mkdir(this.sdk.config.tempDir, { recursive: true })
 
     if (flags.path) {
       snapshotPath = this.sdk.fileSystem.resolve(flags.path)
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/chain/download.ts
```diff
@@ -245,10 +243,43 @@ export default class Download extends IronfishCommand {
...
+    progressBar.start(totalEntries, 0, {
+      speed: '0',
+      estimate: TimeUtils.renderEstimate(0, 0, 0),
+    })
+    speed.start()
+
+    tar.list({
+      file: source,
+      onentry: (_) => progressBar.setTotal(++totalEntries),
+    })
```
NullSoldier, line 267: Is this fast or should we add a CLI status action around this?

NullSoldier (approved):

ironfish-cli/src/commands/chain/download.ts
hughy: It's not that fast -- It takes a couple minutes for the total in the progress bar to stop updating. It slows the unzip down a bit during that time

What would adding a CLI status action around it look like?

PR #2255, opened 2022-09-27
Standardize account flag to match other commands

NullSoldier:
## Summary

This name was wrong, and it doesnt trim the account like the others. Should we pull out some of these so they are shared?

## Breaking Change

```
[ ] Yes
```


hughy (approved):
I think it makes sense to pull this out and standardize -- there are a lot of accounts commands that use this arg

PR #2256, opened 2022-09-27
Move account:transaction to its own command

NullSoldier:
## Summary

Moved most of the logic out of accounts:transactions into its own command. I feel like that command was overloaded and not doing what you expect anyway. I expect when I pass a TX hash for it to display the same output, but filtered on only that transaction. Instead it has a totally alternate flow.

## Testing Plan

Run `ironfish accounts:transaction`

## Breaking Change

```
[x] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/accounts/transaction.ts
+++ b/ironfish-cli/src/commands/accounts/transaction.ts
@@ -6,8 +6,8 @@ import { CliUx } from '@oclif/core'
 import { IronfishCommand } from '../../command'
 import { RemoteFlags } from '../../flags'
 
-export class TransactionsCommand extends IronfishCommand {
-  static description = `Display the account transactions`
+export class TransactionCommand extends IronfishCommand {
+  static description = `Display an account transaction`
 
   static flags = {
     ...RemoteFlags,
@@ -29,7 +29,7 @@ export class TransactionsCommand extends IronfishCommand {
   ]
 
   async start(): Promise<void> {
-    const { args } = await this.parse(TransactionsCommand)
+    const { args } = await this.parse(TransactionCommand)
     const hash = args.hash as string
     const account = args.account as string | undefined
 
--- a/ironfish-cli/src/commands/accounts/transaction.ts
+++ b/ironfish-cli/src/commands/accounts/transaction.ts
@@ -1,0 +1,71 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { CurrencyUtils } from '@ironfish/sdk'
+import { CliUx } from '@oclif/core'
+import { IronfishCommand } from '../../command'
+import { RemoteFlags } from '../../flags'
+
+export class TransactionsCommand extends IronfishCommand {
+  static description = `Display the account transactions`
+
+  static flags = {
+    ...RemoteFlags,
+  }
+
+  static args = [
+    {
+      name: 'hash',
+      parse: (input: string): Promise<string> => Promise.resolve(input.trim()),
+      required: true,
+      description: 'Hash of the transaction',
+    },
+    {
+      name: 'account',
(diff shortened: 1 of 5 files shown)


ironfish-cli/src/commands/accounts/transaction.ts
```diff
@@ -0,0 +1,71 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { CurrencyUtils } from '@ironfish/sdk'
+import { CliUx } from '@oclif/core'
+import { IronfishCommand } from '../../command'
+import { RemoteFlags } from '../../flags'
+
+export class TransactionsCommand extends IronfishCommand {
```
ygao76, line 9: Can you rename to 'TransactionCommand' to keep consistency with the file name

ironfish-cli/src/commands/accounts/transaction.ts
```diff
@@ -0,0 +1,71 @@
...
+  static description = `Display the account transactions`
```
ygao76, line 10: the description is the same as transactions. Need to update it.

NullSoldier: fixed

ironfish-cli/src/commands/accounts/transaction.ts
NullSoldier: fixed

ironfish/src/rpc/routes/accounts/getTransaction.ts
```diff
@@ -3,26 +3,25 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import * as yup from 'yup'
 import { ApiNamespace, router } from '../router'
-import { getAccount, getTransactionNotes, getTransactionStatus } from './utils'
+import { serializeRpcAccountDecryptedNote, serializeRpcAccountTransaction } from './types'
+import { getAccount, getTransactionStatus } from './utils'
 
 export type GetAccountTransactionRequest = { account?: string; hash: string }
 
 export type GetAccountTransactionResponse = {
```
ygao76, line 11: the `GetAccountTransactionResponse` and `GetAccountTransactionRequest` are updated with non backward compatible change, is this okay

ygao76 (dismissed):

ironfish/src/rpc/routes/accounts/getTransaction.ts
```diff
@@ -3,26 +3,25 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import * as yup from 'yup'
 import { ApiNamespace, router } from '../router'
-import { getAccount, getTransactionNotes, getTransactionStatus } from './utils'
+import { serializeRpcAccountDecryptedNote, serializeRpcAccountTransaction } from './types'
+import { getAccount, getTransactionStatus } from './utils'
 
 export type GetAccountTransactionRequest = { account?: string; hash: string }
```
ygao76, line 9: Nit, I noticed this request it has name `hash`, but in the response we referred it as `transactionHash`. It will be good if we keep the naming consistent over our API schema

ygao76: More specifically, in Mira's PR https://github.com/iron-fish/ironfish/pull/2246/files
it is named `transactionHash`. Can we decide on one name and stick with it moving forward.

NullSoldier: I think we should stick with this name then, as if you're in an api called GetX, you should be able to know the hash is to identify X. Any other hash is should be clarified.

hughy (approved):

ironfish/src/rpc/routes/accounts/getTransaction.ts
NullSoldier: I think so, since the new command uses it. We'll have to focus on backwards compatibility at some point.

PR #2257, opened 2022-09-28
changes getTransactions rpc to stream responses

hughy:
## Summary

some accounts (e.g., pool, bank, etc.) have many transactions. loading them all into memory takes a long time, which leads to a bad user experience as the accounts:transactions cli command appears to hang. loading all transactions into memory may also exhaust heap space in extreme cases.

streaming transactions back to the client one at a time mitigates these issues.

- amends RpcAccountTransaction and serializer to include hash, expirationSequence
- changes getAccountTransactions rpc endpoint to streaming
- modifies flow of getTransactions to use serializer
- updates rendering of responses for streaming and changes in serialized response types (e.g., fee change from number to string)

## Testing Plan

accounts_transactions_stream

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/clients/client.ts
```diff
@@ -224,13 +224,13 @@ export abstract class RpcClient {
     ).waitForEnd()
   }
 
-  async getAccountTransactions(
+  getAccountTransactions(
```
NullSoldier, line 227: It looks like we usually use a `Stream` suffix here.

NullSoldier (approved):

Issue #2262, opened 2022-09-28
Mac M1 installaion error


wd021:
### What happened?

Helping someone install ironfish on a M1 mac and ran into an error with prebuilt binaries for `node-datachannel`. Running node v 16.17.1.

### Version

0.1.47

### Debug output

```shell
lindawitters@Lindas-MacBook-Air ~ % npm install -g ironfish
 
npm WARN deprecated @types/minipass@3.3.5: This is a stub types definition. minipass provides its own type definitions, so you do not need this installed.
npm ERR! code 2
npm ERR! path /Users/lindawitters/.nvm/versions/node/v16.17.1/lib/node_modules/ironfish/node_modules/node-datachannel
npm ERR! command failed
npm ERR! command sh /var/folders/pf/38wtgh3s67n4fp5q8n5s1ztr0000gn/T/install-ccfb0790.sh
npm ERR! added 655 packages, and audited 657 packages in 26s
npm ERR! 
npm ERR! 64 packages are looking for funding
npm ERR!   run `npm fund` for details
npm ERR! 
npm ERR! 4 high severity vulnerabilities
npm ERR! 
npm ERR! To address all issues (including breaking changes), run:
npm ERR!   npm audit fix --force
npm ERR! 
npm ERR! Run `npm audit` for details.
npm ERR! 
npm ERR! > node-datachannel@0.3.4 _prebuild
npm ERR! > prebuild --backend cmake-js
npm ERR! 
npm ERR! [
npm ERR!   '/Users/lindawitters/.nvm/versions/node/v16.17.1/bin/node',
npm ERR!   '/Users/lindawitters/.nvm/versions/node/v16.17.1/lib/node_modules/ironfish/node_modules/node-datachannel/node_modules/.bin/cmake-js',
npm ERR!   'rebuild',
npm ERR!   '--runtime-version=16.17.1',
npm ERR!   '--arch=arm64',
npm ERR!   '--runtime=node'
npm ERR! ]
npm ERR! prebuild-install WARN install No prebuilt binaries found (target=16.17.1 runtime=node arch=arm64 libc= platform=darwin)
npm ERR! npm WARN deprecated har-validator@5.1.5: this library is no longer supported
npm ERR! npm WARN deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
[...]

PR #2263, opened 2022-09-28
Bump node datachannel

wd021:
## Summary
Fix https://github.com/iron-fish/ironfish/issues/2262 by bumping node-datachannel to v0.3.6. [Mac M1 binaries were released with v0.3.6](https://github.com/murat-dogan/node-datachannel/issues/120).

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/package.json
+++ b/ironfish/package.json
@@ -35,7 +35,7 @@
     "leveldown": "5.6.0",
     "levelup": "4.4.0",
     "lodash": "4.17.21",
-    "node-datachannel": "0.3.4",
+    "node-datachannel": "0.3.6",
     "node-forge": "1.3.1",
     "node-ipc": "9.1.3",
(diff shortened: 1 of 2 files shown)


dguenther:
The one thing I'd like to double-check on this is that it doesn't try to re-build the node-datachannel binaries from scratch on Windows and Mac (non-M1) -- I feel like I ran into some issues with this when upgrading node-datachannel on a different repository, and it's a bit inconvenient since it requires cmake and openssl to be installed.

dguenther:
Okay, I still have occasional issues where node-datachannel will attempt to build rather than installing the prebuilds on Windows, But this happens to me on the current version (0.3.4) as well, so I think it's fine to upgrade this. Tested connecting to the network and syncing a bit, and seems to work great, thanks!

PR #2264, opened 2022-09-28
moves transaction status into wallet

hughy:
## Summary

determines the status of a trnasaction relative to an account's head hash. matches the status of a transaction for the account to how the notes from that transaction affect balance calculation for the account.

- defines 'getAccountHeadSequence' to fetch the sequence for an account's head hash. pulls logic from 'getBalance'
- uses transaction sequence, account head sequence, and minimum block confirmations to determine transaction status
- removes 'forked' status: sequence is only populated on a transaction when it is added to the main chain, and if there are enough block confirmations we consider a transaction confirmed.
- when account head hash is null, uses 'unknown' status

unrelated cleanup:
- removes unnecessary ternary operator in 'isAccountUpToDate'

## Testing Plan

local testing with `accounts:transactions`

## Breaking Change

```
[ ] Yes
```


hughy:
> when account head hash is null, uses 'unknown' status

alternatively, we could return no results from `accounts:transactions` if the user needs to start a rescan

ironfish/src/rpc/routes/accounts/getTransaction.ts
```diff
@@ -80,12 +80,12 @@ router.register<typeof GetAccountTransactionRequestSchema, GetAccountTransaction
...
-      node,
-      transaction.blockHash,
-      transaction.sequence,
-      transaction.transaction.expirationSequence(),
-    )
+    const headSequence = await node.accounts.getAccountHeadSequence(account)
+    const minimumBlockConfirmations = node.config.get('minimumBlockConfirmations')
+
+    const status = headSequence
+      ? account.getTransactionStatus(transaction, headSequence, minimumBlockConfirmations)
```
NullSoldier, line 87: I have the feeling this code would be a little cleaner if it was on the wallet and not on an account. Then the wallet already knows about all the pre-requisite data.

ironfish/src/wallet/accounts.test.ts
```diff
@@ -341,4 +342,146 @@ describe('Accounts', () => {
...
+
+      const transactionValue = await accountA.getTransaction(transaction.hash())
+      Assert.isNotUndefined(transactionValue)
+
+      const transactionStatus = await node.accounts.getTransactionStatus(
+        accountA,
+        transactionValue,
+      )
+
+      expect(transactionStatus).toEqual('confirmed')
```
ygao76, line 367: should this equal ('unconfirmed')?!

ironfish/src/wallet/wallet.ts
```diff
@@ -888,6 +887,36 @@ export class Accounts {
     }
   }
 
+  async getTransactionStatus(
```
ygao76, line 890: should the transaction status be a enum type instead of free form string?

ironfish/src/wallet/accounts.test.ts
```diff
@@ -341,4 +342,146 @@ describe('Accounts', () => {
       )
     })
   })
+
+  describe('getTransactionStatus', () => {
+    it('should show unconfirmed transactions as unconfirmed', async () => {
```
ygao76, line 347: Can you add a test to test the 'unknown' status

ironfish/src/wallet/accounts.test.ts
hughy: It should! Looks like this test is broken!

ironfish/src/wallet/accounts.test.ts
hughy: I think that would be helpful -- I think we need to standardize transaction statuses across the CLI and our documentation

ironfish/src/wallet/wallet.ts
hughy: I think that would be helpful -- I think we need to standardize transaction statuses across the CLI and our documentation

ironfish/src/rpc/routes/accounts/getTransaction.ts
```diff
@@ -66,36 +66,34 @@ router.register<typeof GetAccountTransactionRequestSchema, GetAccountTransaction
...
     const transactionHash = Buffer.from(request.data.hash, 'hex')
-    const transaction = await account.getTransactionByUnsignedHash(transactionHash)
 
-    if (!transaction) {
-      return request.end({
-        account: account.displayName,
-        transaction: null,
-      })
-    }
+    await node.accounts.db.database.transaction(async (tx) => {
```
NullSoldier, line 70: Do we need this transaction? What atomic writes do we need here?

ironfish/src/rpc/routes/accounts/getTransactions.ts
```diff
@@ -61,19 +62,26 @@ router.register<typeof GetAccountTransactionsRequestSchema, GetAccountTransactio
 
     if (request.data.hash) {
       const hash = Buffer.from(request.data.hash, 'hex')
-      const transaction = await account.getTransactionByUnsignedHash(hash)
 
-      if (transaction) {
-        await streamTransaction(request, node, account, transaction)
-      }
+      await node.accounts.db.database.transaction(async (tx) => {
```
NullSoldier, line 66: I'm not sure we need this transaction exactly. It looks like it'll be fine without it

NullSoldier (approved):

PR #2265, opened 2022-09-28
Check if wallet is fully synced before creating transaction

ygao76:
## Summary

https://linear.app/ironfish/issue/IRO-2484/prevent-creating-transactions-when-the-wallet-is-not-fully-scanned
Check if wallet is fully synced before creating transaction

## Testing Plan
unit test

## Breaking Change

```
[ ] Yes
```
@iron-fish/engineering @hughy @NullSoldier @mirayadav


ironfish/src/wallet/wallet.ts
```diff
@@ -680,6 +681,28 @@ export class Accounts {
     try {
       this.assertHasAccount(sender)
 
+      // check if the chain data is fully synced
+      let headHash = this.headHashes.get(sender.id)
+      headHash = await sender.getHeadHash()
+      if (!headHash) {
+        headHash = this.headHashes.get(sender.id)
+      }
```
hughy, line 689: `this.headHashes.get(sender.id)` and `sender.getHeadHash()` should give us the same value

I'm not sure we need the in-memory cache of head hashes, @NullSoldier ?

ironfish/src/wallet/wallet.ts
```diff
@@ -680,6 +681,28 @@ export class Accounts {
     try {
       this.assertHasAccount(sender)
 
+      // check if the chain data is fully synced
+      let headHash = this.headHashes.get(sender.id)
+      headHash = await sender.getHeadHash()
+      if (!headHash) {
```
danield9tqh, line 687: Is there a reason we check twice? Does `sender.getHeadHash()` do something that would change the result of checking the first time?

ironfish/src/wallet/wallet.ts
```diff
@@ -680,6 +681,28 @@ export class Accounts {
...
+        throw new ResponseError(`No headhash found for this account.`, ERROR_CODES.ERROR, 400)
+      }
+
+      const header = await this.chain.getHeader(headHash)
+      Assert.isNotNull(header, `Missing block header for hash '${headHash.toString('hex')}'`)
+
+      if (header.timestamp.valueOf() < Date.now() - MAX_SYNCED_AGE_MS) {
+        throw new ResponseError(
+          `Your node must be synced with the Iron Fish network to send a transaction. `,
```
danield9tqh, line 700: Maybe we can make this error message specific to the `Accounts` head so it's clear that the account is not synced instead of just the node not being synced

ironfish/src/wallet/wallet.ts
```diff
@@ -680,6 +681,28 @@ export class Accounts {
...
+
+      if (!headHash) {
+        throw new ResponseError(`No headhash found for this account.`, ERROR_CODES.ERROR, 400)
+      }
+
+      const header = await this.chain.getHeader(headHash)
+      Assert.isNotNull(header, `Missing block header for hash '${headHash.toString('hex')}'`)
+
+      if (header.timestamp.valueOf() < Date.now() - MAX_SYNCED_AGE_MS) {
+        throw new ResponseError(
```
hughy, line 699: nit - we use `ValidationError` elsewhere in this file already. It's mostly for convenience, but pre-fills the status and error code

ironfish/src/primitives/transaction.test.slow.ts
```diff
@@ -46,6 +48,46 @@ describe('Accounts', () => {
     expect(transactionB.isMinersFee()).toBe(true)
   })
 
+  it('throw error if account is not fully synced when creating transaction', async () => {
+    const nodeA = nodeTest.node
+
+    // Create an account A
+    const accountA = await useAccountFixture(nodeA.accounts, () =>
```
danield9tqh, line 55: nit: seems like pattern is usually: 
```
const accountA = await useAccountFixture(accounts, 'accountA')
const accountB = await useAccountFixture(accounts, 'accountB')
```
using the string instead of generator function

danield9tqh:
Actually thinking maybe we should move this check to be in the same location as [our check for chain.synced](https://github.com/iron-fish/ironfish/blob/486d4842d4ec5c6bac02e1b8a3356600bb1591ef/ironfish/src/rpc/routes/transactions/sendTransaction.ts#L89). accounts.createTransaction() is used in many places like [testing fixtures](https://github.com/iron-fish/ironfish/blob/486d4842d4ec5c6bac02e1b8a3356600bb1591ef/ironfish/src/testUtilities/fixtures.ts#L227) and we may not want to be checking `Date.now` so deep in the code so that this function remains reusable

hughy (commented):
We might want to implement this on the account similar to what we do for the chain: https://github.com/iron-fish/ironfish/blob/master/ironfish/src/blockchain/blockchain.ts#L1329

ironfish/src/wallet/wallet.ts
ygao76: Do we want to consider this error as Validation error ?  Its not exact validation error since user request/input does not have issue.  but it's kind of fall into the category of blockchain not in valid state for creating transaction.

ygao76:
https://github.com/iron-fish/ironfish/pull/2265#issuecomment-1262927814

I change the fixture function to support header timestamp. The fixture tests should not be an issue now.

ironfish/src/blockchain/blockchain.test.perf.ts
```diff
@@ -35,8 +35,20 @@ describe('Blockchain', () => {
...
+          undefined,
+          nodeA.accounts,
+        )
+        blockB = await useMinerBlockFixture(
+          nodeB.chain,
+          undefined,
+          accountB,
+          undefined,
+          nodeB.accounts,
+        )
```
NullSoldier, line 51: What is changing in this old test? Is there code change here?

ironfish/src/wallet/wallet.ts
```diff
@@ -680,6 +681,27 @@ export class Accounts {
...
+      }
+
+      if (!headHash) {
+        throw new ResponseError(`No headhash found for this account.`, ERROR_CODES.ERROR, 400)
+      }
+
+      const header = await this.chain.getHeader(headHash)
+      Assert.isNotNull(header, `Missing block header for hash '${headHash.toString('hex')}'`)
+
+      if (header.timestamp.valueOf() < Date.now() - MAX_SYNCED_AGE_MS) {
```
NullSoldier, line 697: This should not be based on time, but based on if we processed the chain or not in the chain processor.

We should just check if the account head is scanned up to the chain head. It's safe to create transactions when the chain is not synced, but not safe to create them when we haven't processed the chain.

In that case, it can be 1 single line to check the account head sequence vs the chain head sequence, or just that they match if we want to be more strict.

ygao76: How do we know if we haven't processed the chain when they create the transaction? @NullSoldier

NullSoldier: It's when the account head matches the chain head, which is what isAccountUpToDate does

ironfish/src/wallet/wallet.ts
NullSoldier: agreed, I would like a better message.

NullSoldier: We should just throw a normal `Error` here, and not ResponseError since that's for the RPC layer. I see we merged in a change from the community that is throwing a ValidationError above, which is actually a mistake because that person's code should throw this error instead. https://github.com/iron-fish/ironfish/blob/staging/ironfish/src/wallet/errors.ts#L5. I think it could be a bug.

NullSoldier: I fixed the issue here that I was talking about, https://github.com/iron-fish/ironfish/pull/2289

hughy (approved):

PR #2267, opened 2022-09-28
keeps transactions in accountsdb on expiration

hughy:
## Summary

users sending transactions want to know the status of their transactions and to know if their transactions have expired without being processed. if expired transactions are deleted from the database then there is no record that they were attempted.

changes 'deleteTransaction' to 'expireTransaction' and does not delete expired transactions from the database. still deletes notes from expired transactions from the database since the account doesn't own these notes

wallet will lose any record of expired transactions on reset since there is no history of expired transactions on the chain

## Testing Plan

sent an expired transaction, saw its status in 'accounts:transactions'

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -394,13 +394,9 @@ export class Wallet {
     })
   }
 
-  /**
-   * Deletes a transaction from the transaction map and updates
-   * the related maps.
-   */
-  async deleteTransaction(transaction: Transaction): Promise<void> {
+  async expireTransaction(transaction: Transaction): Promise<void> {
     for (const account of this.accounts.values()) {
-      await account.deleteTransaction(transaction)
+      await account.expireTransaction(transaction)
     }
   }
 
@@ -882,7 +878,7 @@ export class Wallet {
         )
 
         if (isExpired) {
-          await this.deleteTransaction(transaction)
+          await this.expireTransaction(transaction)
         }
       }
     }
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -394,12 +394,6 @@ export class Wallet {
     })
   }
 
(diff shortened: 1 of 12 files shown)


ygao76 (approved):

ironfish/src/wallet/wallet.ts
```diff
@@ -394,13 +394,9 @@ export class Accounts {
     })
   }
 
-  /**
-   * Deletes a transaction from the transaction map and updates
-   * the related maps.
-   */
-  async deleteTransaction(transaction: Transaction): Promise<void> {
+  async expireTransaction(transaction: Transaction): Promise<void> {
```
NullSoldier, line 397: I think we should still have a function to delete transactions. I think we'll need it to solve a problem coming up about TX staying around when they are permanently invalid, especially miners fees.

NullSoldier: Nevermind, ignore this comment I think this is fine.

ironfish/src/wallet/wallet.ts
```diff
@@ -882,7 +878,7 @@ export class Accounts {
         )
 
         if (isExpired) {
-          await this.deleteTransaction(transaction)
+          await this.expireTransaction(transaction)
```
NullSoldier, line 881: This will keep re-expiring the same transactions over and over for eternity We should add have a separate list of TX that need to be checked for expiration that shrinks over time. Maybe a new composite key table that's `[Account.id, expirationSequence, TransactionHash]`,

hughy: Didn't think of that 🤦 

Having a separate list of pending transactions will probably be beneficial for larger wallets anyway to avoid repeatedly iterating over all transactions

ygao76 (approved):

NullSoldier (approved):
The issue is the migration already ran. So anyone that is on staging will not have transactions in the new table.

PR #2269, opened 2022-09-29
Use ResponseError and don't make manual responses

NullSoldier:
## Summary

Review with https://github.com/iron-fish/ironfish/pull/2269/files?diff=split&w=1

This code was constructing response manually, when we already support that through ResponseError handling in the adapter.

This is fixing this comment: https://github.com/iron-fish/ironfish/pull/2224/files#r979296411

## Testing Plan

Covered by existing tests. IT will be sending back the same thing to the client.

## Breaking Change

```
[ ] Yes
```


ygao76 (approved):
Thanks for refactoring the codes ❤️

PR #2270, opened 2022-09-29
Update account deletion to clean up records in stores

mirayadav:
## Summary

Clear records in stores for a deleted account. Adds a store to keep track of IDs of accounts that have yet to be cleaned up and then processes this clean up in the background.

## Testing Plan

Ran `accounts:remove` locally.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -198,17 +198,12 @@ export class Wallet {
       this.headHashes.set(accountId, headHash)
     }
 
-    for await (const { accountId } of this.walletDb.loadAccountIdsToCleanup()) {
-      this.accountIdsToCleanup.push(accountId)
-    }
-
     this.chainProcessor.hash = await this.getLatestHeadHash()
   }
 
   private unload(): void {
     this.accounts.clear()
     this.headHashes.clear()
-    this.accountIdsToCleanup.length = 0
 
     this.defaultAccount = null
     this.chainProcessor.hash = null
@@ -1033,10 +1028,8 @@ export class Wallet {
 
       await this.walletDb.removeAccount(account, tx)
       await this.walletDb.removeHeadHash(account, tx)
-      await this.walletDb.saveAccountIdToCleanup(account.id, tx)
     })
 
-    this.accountIdsToCleanup.push(account.id)
     this.accounts.delete(account.id)
     this.onAccountRemoved.emit(account)
   }
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -1034,6 +1034,10 @@ export class Wallet {
   }
 
   async cleanupDeletedAccounts(): Promise<void> {
+    if (!this.isStarted) {
+      return
+    }
+
     await this.walletDb.cleanupDeletedAccounts(this.eventLoopAbortController.signal)
   }
 
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -1038,6 +1038,10 @@ export class Wallet {
       return
     }
 
+    if (this.scan || this.updateHeadState) {
+      return
+    }
+
     await this.walletDb.cleanupDeletedAccounts(this.eventLoopAbortController.signal)
   }
 
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -197,17 +197,12 @@ export class Wallet {
       this.headHashes.set(accountId, headHash)
     }
 
-    for await (const { accountId } of this.walletDb.loadAccountIdsToCleanup()) {
-      this.accountIdsToCleanup.push(accountId)
-    }
-
     this.chainProcessor.hash = await this.getLatestHeadHash()
   }
 
   private unload(): void {
     this.accounts.clear()
     this.headHashes.clear()
-    this.accountIdsToCleanup.length = 0
 
     this.defaultAccount = null
     this.chainProcessor.hash = null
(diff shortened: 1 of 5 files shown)


ironfish/src/wallet/wallet.ts
```diff
@@ -973,6 +973,11 @@ export class Accounts {
 
       await this.db.removeAccount(account, tx)
       await this.db.removeHeadHash(account, tx)
+      await this.db.clearDecryptedNotes(account, tx)
+      await this.db.clearNullifierToNoteHash(account, tx)
+      await this.db.clearTransactions(account, tx)
+      await this.db.clearSequenceToNoteHash(account, tx)
+      await this.db.clearNonChainNoteHashes(account, tx)
```
NullSoldier, line 980: This is gonna hang my node for hours potentially. We should test this on the bank and the pool.

hughy: Tested this on my pool account (~9k notes) and removed the account in 10 seconds

On my partially-scanned bank account (~4M notes) it took about 30 minutes

NullSoldier:
We're going to change this to a passive delete in the background. This will be fine with multiple accounts because we'll generate a new non deterministic ID per account.

ironfish/src/wallet/wallet.ts
```diff
@@ -56,6 +57,7 @@ export class Accounts {
   protected isStarted = false
   protected isOpen = false
   protected eventLoopTimeout: SetTimeoutToken | null = null
+  protected readonly accountIdsToCleanup: string[] = []
```
NullSoldier, line 60: do we need this in memory cache, or can we just load these from the DB?

ironfish/src/wallet/wallet.ts
```diff
@@ -973,12 +981,31 @@ export class Accounts {
...
+      await this.db.saveAccountIdToCleanup(account.id, tx)
     })
 
+    this.accountIdsToCleanup.push(account.id)
     this.accounts.delete(account.id)
     this.onAccountRemoved.emit(account)
   }
 
+  async cleanupDeletedAccounts(): Promise<void> {
+    setTimeout(() => void this.cleanupDeletedAccounts(), 100000)
```
NullSoldier, line 993: One of these will get queued up every event loop cycle, was that intended?

NullSoldier: Were not consuming the return value of setTimeout, which means we're creating a bug with a hanging co-routine. What happens here during shut down?

ironfish/src/wallet/walletdb/walletdb.ts
```diff
@@ -572,4 +584,38 @@ export class WalletDB {
...
+      this.decryptedNotes,
+    ]
+
+    for (const [accountId] of await this.accountIdsToCleanup.getAll()) {
+      const prefix = calculateAccountPrefix(accountId)
+      const range = StorageUtils.getPrefixKeyRange(prefix)
+
+      for (const store of stores) {
+        for await (const key of store.getAllKeysIter(undefined, range)) {
+          if (signal?.aborted === true || recordsToCleanup === 0) {
```
ygao76, line 609: I am a little confused here. Does the value of signal.aborted change during the execution of this function? 

If the value does not change,  it seems when signal is aborted, no need to call cleanuoDeletedAccounts function. Can we move the condition outside of the function instead of doing it in every for iteration?

NullSoldier: Yes, it can change during the execution of this function. It's an AbortedController which is used to signal to an asynchronous task that it should stop. https://developer.mozilla.org/en-US/docs/Web/API/AbortController

ironfish/src/wallet/walletdb/walletdb.ts
```diff
@@ -572,4 +584,38 @@ export class WalletDB {
...
+            return
+          }
+
+          await store.del(key)
```
ygao76, line 613: does this need try catch in case the del is not successful?

ironfish/src/wallet/wallet.ts
NullSoldier: This has been changed to an incremental cleanup in the background, so this is no longer a problem.

ironfish/src/wallet/walletdb/walletdb.ts
```diff
@@ -173,6 +178,12 @@ export class WalletDB {
       ),
       valueEncoding: NULL_ENCODING,
     })
+
+    this.accountIdsToCleanup = this.db.addStore({
+      name: 'A',
```
ygao76, line 183: why we name it 'A' loll

ironfish/src/wallet/walletdb/walletdb.ts
NullSoldier: What would you expect the catch to do? If If the delete is not successful the program should crash because the leveldb is corrupt and can no longer operate.

ygao76 (approved):

ironfish/src/wallet/walletdb/walletdb.ts
NullSoldier: What would a better name be? all the stores have single letter names. The reason was because the prefix buffers weren't hashed when calculated for the store. Now they are fixed size buffers of 4 bytes to avoid collision automatically but that is 4 extra bytes per row instead of 1 byte per row.

NullSoldier (approved):

ironfish/src/wallet/walletdb/walletdb.ts
ygao76: I just looked up, `del` does not throw any error.

PR #2274, opened 2022-09-29
adds method to get transaction by unsignedHash

hughy:
## Summary

updates getTransaction and getTransactions to use getTransactionByUnsignedHash to load specific transactions from the account.

we need to search by unsignedHash (for now) because we use the unsigned hash for the transaction in the block explorer and cli output. users have no way of finding the other hash.

## Testing Plan

image

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):

NullSoldier:
Thanks for the fix, good eye on this bug.

PR #2275, opened 2022-09-29
Only store peers in CONNECTED or CONNECTING states

dguenther:
## Summary

Opening this as a draft to give some visibility into the work here.

### TODO

* [x] Move connectionRetry out of Peer so we don't have to keep disconnected Peers
* [x] Move disconnectUntil out of Peer so we don't have to keep disconnected Peers
* [x] Measure memory usage before + after
* [x] <s>Cap number of peerCandidates in the peerCandidateMap?</s>
* [x] If memory usage is improved, write a doc + clean it up
* [x] Re-enable fetching Peers on a timer
* [x] <s>Remove tryDispose function</s>
* [x] Fix tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/peers/addressManager.ts
+++ b/ironfish/src/network/peers/addressManager.ts
@@ -4,7 +4,6 @@
 import { HostsStore } from '../../fileStores'
 import { ArrayUtils } from '../../utils'
 import { Peer } from '../peers/peer'
-import { ConnectionDirection, ConnectionType } from './connections'
 import { PeerAddress } from './peerAddress'
 import { PeerManager } from './peerManager'
 
@@ -71,21 +70,14 @@ export class AddressManager {
   }
 
   /**
-   * Persist all currently connected peers and unused peer addresses to disk
+   * Persist all currently connected peers to disk
    */
-  async save(peers: Peer[]): Promise<void> {
+  async save(): Promise<void> {
     // TODO: Ideally, we would like persist peers with whom we've
     // successfully established an outbound Websocket connection at
     // least once.
-    const inUsePeerAddresses: PeerAddress[] = peers.flatMap((peer) => {
-      if (
-        peer.state.type === 'CONNECTED' &&
-        !this.peerManager.getConnectionRetry(
-          peer.state.identity,
-          ConnectionType.WebSocket,
-          ConnectionDirection.Outbound,
-        )?.willNeverRetryConnecting
-      ) {
+    const inUsePeerAddresses: PeerAddress[] = this.peerManager.peers.flatMap((peer) => {
+      if (peer.state.type === 'CONNECTED') {
         return {
           address: peer.address,
           port: peer.port,
--- a/ironfish/src/network/peers/addressManager.ts
+++ b/ironfish/src/network/peers/addressManager.ts
@@ -6,6 +6,7 @@ import { ArrayUtils } from '../../utils'
 import { Peer } from '../peers/peer'
 import { ConnectionDirection, ConnectionType } from './connections'
 import { PeerAddress } from './peerAddress'
+import { PeerManager } from './peerManager'
 
 /**
  * AddressManager stores the necessary data for connecting to new peers
@@ -13,9 +14,11 @@ import { PeerAddress } from './peerAddress'
  */
 export class AddressManager {
   hostsStore: HostsStore
+  peerManager: PeerManager
 
-  constructor(hostsStore: HostsStore) {
+  constructor(hostsStore: HostsStore, peerManager: PeerManager) {
     this.hostsStore = hostsStore
+    this.peerManager = peerManager
   }
 
   get priorConnectedPeerAddresses(): ReadonlyArray<Readonly<PeerAddress>> {
@@ -77,8 +80,11 @@ export class AddressManager {
     const inUsePeerAddresses: PeerAddress[] = peers.flatMap((peer) => {
       if (
         peer.state.type === 'CONNECTED' &&
-        !peer.getConnectionRetry(ConnectionType.WebSocket, ConnectionDirection.Outbound)
-          .willNeverRetryConnecting
+        !this.peerManager.getConnectionRetry(
+          peer.state.identity,
+          ConnectionType.WebSocket,
+          ConnectionDirection.Outbound,
+        )?.willNeverRetryConnecting
       ) {
         return {
           address: peer.address,
--- a/ironfish/src/network/peers/connections/webSocketConnection.ts
+++ b/ironfish/src/network/peers/connections/webSocketConnection.ts
@@ -81,7 +81,7 @@ export class WebSocketConnection extends Connection {
     this.socket.onmessage = (event: MessageEvent) => {
       if (!Buffer.isBuffer(event.data)) {
         const message = 'Received non-buffer message'
-        this.logger.warn(message, event.data)
+        this.logger.debug(message, event.data)
         this.close(new NetworkError(message))
         return
       }
--- a/ironfish/src/network/peers/connections/webSocketConnection.ts
+++ b/ironfish/src/network/peers/connections/webSocketConnection.ts
@@ -81,7 +81,7 @@ export class WebSocketConnection extends Connection {
     this.socket.onmessage = (event: MessageEvent) => {
       if (!Buffer.isBuffer(event.data)) {
         const message = 'Received non-buffer message'
-        this.logger.debug(message, event.data)
+        this.logger.warn(message, event.data)
         this.close(new NetworkError(message))
         return
       }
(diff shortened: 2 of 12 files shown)


ironfish/src/network/peers/connections/webSocketConnection.ts
```diff
@@ -81,7 +81,7 @@ export class WebSocketConnection extends Connection {
     this.socket.onmessage = (event: MessageEvent) => {
       if (!Buffer.isBuffer(event.data)) {
         const message = 'Received non-buffer message'
-        this.logger.debug(message, event.data)
+        this.logger.warn(message, event.data)
```
dguenther, line 84: Revert this

ironfish/src/network/peers/peerManager.ts
```diff
@@ -742,7 +792,6 @@ export class PeerManager {
   }
 
   start(): void {
-    this.requestPeerListHandle = setInterval(() => this.requestPeerList(), 60000)
```
dguenther, line 745: Revert this

ironfish/src/network/peers/peer.ts
```diff
@@ -259,36 +228,40 @@ export class Peer {
...
+   * Returns the existing connection if the peer already has a WebRTC connection.
    * @param connection The WebRTC connection to set
    */
-  setWebRtcConnection(connection: WebRtcConnection): void {
+  setWebRtcConnection(connection: WebRtcConnection): WebRtcConnection | null {
+    let existingConnection = null
     if (this.state.type !== 'DISCONNECTED' && this.state.connections.webRtc) {
-      this.logger.warn('Already have a WebRTC connection, ignoring the new one')
-      return
+      existingConnection = this.state.connections.webRtc
```
danield9tqh, line 237: Is there a reason we've changed this to not early return anymore? Don't we still want to keep the existing connection if we have one?

ironfish/src/network/peers/peer.ts
```diff
@@ -259,36 +228,40 @@ export class Peer {
...
+   * Returns the existing connection if the peer already has a WebSocket connection.
    * @param connection The WebSocket connection to set
    */
-  setWebSocketConnection(connection: WebSocketConnection): void {
+  setWebSocketConnection(connection: WebSocketConnection): WebSocketConnection | null {
+    let existingConnection = null
     if (this.state.type !== 'DISCONNECTED' && this.state.connections.webSocket) {
-      this.logger.debug('Already have a WebSocket connection, ignoring the new one')
-      return
+      existingConnection = this.state.connections.webSocket
```
danield9tqh, line 256: Same here, don't we still want to keep the existing connection if we have one?

ironfish/src/network/peers/peer.ts
```diff
@@ -424,19 +397,6 @@ export class Peer {
 
     this._address = address
     this._port = port
-
-    if (address === null && port === null) {
```
danield9tqh, line 428: It looks like we've removed this logic but we aren't calling `PeerManager.getConnectionRetry.neverRetryConnecting` in place of removing this logic. Is that what we want to do? We call `setWebSocketAddress` in a few places where I don't think this logic has been replaced

ironfish/src/network/testUtilities/helpers.ts
```diff
@@ -156,8 +150,24 @@ export function getSignalingWebRtcPeer(
...
+  pm.peerCandidateMap.set(brokeringPeerIdentity, {
+    address: brokeringPeer.address,
+    port: brokeringPeer.port,
+    neighbors: new Set([peerIdentity]),
+    webRtcRetry: new ConnectionRetry(),
+    websocketRetry: new ConnectionRetry(),
+    peerRequestedDisconnectUntil: null,
+    localRequestedDisconnectUntil: null,
+  })
+  pm.peerCandidateMap.set(peerIdentity, {
```
danield9tqh, line 162: Do we want to reset everything like the `webRtcRetry` and `websocketRetry` here? What if the peer already exists?

ironfish/src/network/testUtilities/helpers.ts
```diff
@@ -156,8 +150,24 @@ export function getSignalingWebRtcPeer(
...
+    address: brokeringPeer.address,
```
danield9tqh, line 163: Should this be setting to `peer.address` and `peer.port`?

ironfish/src/network/peers/addressManager.ts
```diff
@@ -77,8 +80,11 @@ export class AddressManager {
     const inUsePeerAddresses: PeerAddress[] = peers.flatMap((peer) => {
       if (
         peer.state.type === 'CONNECTED' &&
-        !peer.getConnectionRetry(ConnectionType.WebSocket, ConnectionDirection.Outbound)
-          .willNeverRetryConnecting
+        !this.peerManager.getConnectionRetry(
```
danield9tqh, line 83: nit: A little inconsistent that we pass `PeerManager.peers` into this function but get `PeerManager.peerCandidateMap` by passing in the whole `PeerManager` into the constructor. Maybe we could do both one way or another

ironfish/src/network/peers/peerManager.ts
```diff
@@ -178,19 +216,26 @@ export class PeerManager {
...
 
-    // If we're trying to connect to the peer, we don't care about limiting the peer's connections to us
-    peer.localRequestedDisconnectUntil = null
-    peer.localRequestedDisconnectReason = null
+    const alternateIdentity = peer.state.identity ?? peer.getWebSocketAddress()
 
-    // Clear out peerRequestedDisconnect if we passed it
-    peer.peerRequestedDisconnectUntil = null
-    peer.peerRequestedDisconnectReason = null
+    const candidate = this.peerCandidateMap.get(alternateIdentity)
```
danield9tqh, line 221: This is a little confusing that we store a mapping in `peerCandidateMap` by both web socket address and peer identity. Would this ever result in duplicates? I guess it makes sense that we'd need to use web socket address because some of these peers are unidentified but if that's the case then couldn't we use web socket address for all of them and have some separate mapping of peerIdentity to web socket address? Just seems a little brittle to map by both

ironfish/src/network/peers/peerManager.ts
```diff
@@ -165,6 +188,21 @@ export class PeerManager {
...
+        port: url.port,
+        neighbors: new Set(),
+        webRtcRetry: new ConnectionRetry(),
+        websocketRetry: new ConnectionRetry(),
+        localRequestedDisconnectUntil: null,
+        peerRequestedDisconnectUntil: null,
+      })
+    }
+
     peer.isWhitelisted = isWhitelisted
```
danield9tqh, line 206: I think we could probably move `isWhitelisted` onto `ConnectionRetry`. That's the only place that we use it is when we call `failedConnection`

danield9tqh: We decided to move `isWhitelisted` onto `ConnectionRetry`

ironfish/src/network/testUtilities/helpers.ts
danield9tqh: Yes we should, updated

ironfish/src/network/testUtilities/helpers.ts
danield9tqh: This is just used in a test helper and the peer should never exist at this point. Adding an `Assert` to make sure that's true

ironfish/src/network/peers/peer.ts
danield9tqh: We added a new function to differentiate between replacing an existing connection and setting a connection if it does not exist

ironfish/src/network/peers/peer.ts
danield9tqh: We added a new function to differentiate between replacing an existing connection and setting a connection if it does not exist

ironfish/src/network/peers/peerManager.ts
danield9tqh: This is used for the bootstrap node primarily since it does not have an identity. We will consider refactoring connecting to the bootstrap nodes in the future but leave this as is for now

danield9tqh (approved):

ironfish/src/network/peers/peer.ts
```diff
@@ -274,14 +238,37 @@ export class Peer {
...
+    }
+  }
+
   /**
    * Sets a WebSocket connection on the peer, moving it into the CONNECTING state if necessary.
    * Ignores the connection if the peer already has a WebSocket connection.
    * @param connection The WebSocket connection to set
    */
   setWebSocketConnection(connection: WebSocketConnection): void {
     if (this.state.type !== 'DISCONNECTED' && this.state.connections.webSocket) {
```
jowparks, line 270: log line updated

PR #2287, opened 2022-10-03
Constant delete priority queue

danield9tqh:
## Summary
If the mempool is very large, adding blocks takes longer because as part of adding blocks we also [remove transactions in that block from the mempool](https://github.com/iron-fish/ironfish/blob/staging/ironfish/src/memPool/memPool.ts#L145). We also remove expired transactions. Both of these actions require a [O(n) operation to remove transactions](https://github.com/lemire/FastPriorityQueue.js/blob/7b70143f3c826288bd1aeeb2b13484374e003b52/FastPriorityQueue.js#L144) from the priority queue.

Because we do a linear scan of all transactions looking for the one to delete, removing transactions takes on average n/2 calls to [hash.equals()](https://github.com/iron-fish/ironfish/blob/ed3ecac87960440c0e2627e548e50d1491695143/ironfish/src/memPool/memPool.ts#L223). Doing some profiling. with a mempool of size 50k and ~300 transactions per block that’s 150*50000 calls to hash.equals() which on an M1 mac takes around ~2 seconds.

This PR adds a new data structure with constant time delete from a priority queue to improve mempool performance

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/memPool/priorityQueue.ts
+++ b/ironfish/src/memPool/priorityQueue.ts
@@ -10,8 +10,8 @@
 export class PriorityQueue<T> {
   private _items: T[] = []
   private _positions: { [key: string]: number } = {}
-  private _compare: (a: T, b: T) => boolean
-  hash: (item: T) => string
+  readonly compare: (a: T, b: T) => boolean
+  readonly hash: (item: T) => string
 
   /**
    * This data structure is a **Priority Queue** as well as a **Set**. Duplicate entries are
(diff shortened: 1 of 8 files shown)


ironfish/src/memPool/priorityQueue.ts
```diff
@@ -0,0 +1,227 @@
...
+   * This data structure is a **Priority Queue** as well as a **Set**. Duplicate entries are
+   * not allowed. Entries are identified based on the hash function provided
+   *
+   * @param compare should return true if a is higher priority than b, otherwise false
+   * @param hash should return a unique hash of the item as a string. It should and should
+   * be a pure function and very fast (ideally just a field access) because it is called often
+   */
+  constructor(compare: (a: T, b: T) => boolean, hash: (item: T) => string) {
+    this._compare = compare
+    this.hash = hash
```
dguenther, line 26: Could hash be private as well? I'm curious why hash is public and _compare is private

dguenther (approved):

PR #2288, opened 2022-10-04
Fix wallet event loop shutdown

mirayadav:
## Summary

Return from functions running in the wallet event loop during shutdown using an AbortController.

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/wallet.ts
```diff
@@ -861,6 +868,10 @@ export class Wallet {
 
     for (const account of this.accounts.values()) {
       for await (const { transaction } of account.getExpiredTransactions(head.sequence)) {
+        if (this.eventLoopAbortController.signal.aborted) {
```
NullSoldier, line 871: Let's put this per account too

NullSoldier (approved):

PR #2293, opened 2022-10-05
adds rpc:token cli command

hughy:
## Summary

adds a command to retrieve or set the value of the rpcAuthToken key in internal.json

`ironfish rpc:token` prints out the token, if any.

`ironfish rpc:token --token=<value>` sets the token.

Closes https://airtable.com/appIXmGgVqP9QdbCf/pagcUNImP87iLU1r7?FG6K5=recUzR97TjxBboY00

## Testing Plan

image

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/rpc/token.ts
+++ b/ironfish-cli/src/commands/rpc/token.ts
@@ -1,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Flags } from '@oclif/core'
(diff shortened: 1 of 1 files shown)


ygao76 (approved):

ygao76:
From the security standpoint, can malicious people get the rpc token and hijack the tls calll? I assume this call needs auth if its via tls so the auth token is not exposed publicly. Can people get the auth token w/t the auth token?

hughy:
> From the security standpoint, can malicious people get the rpc token and hijack the tls calll? I assume this call needs auth if its via tls so the auth token is not exposed publicly. Can people get the auth token w/t the auth token?

That's a good question! This will be safe because it doesn't make an RPC connection to a node or any TLS calls. It uses the SDK to read/write the local `internal.json` file without interacting with the node.

PR #2298, opened 2022-10-05
Drop notes where their transaction is missing

NullSoldier:
## Summary

User's are hitting this when running migration 13. We can only just drop these TX, similar to if the TX is missing from the cahce. Unsure how user's DB's are getting into this state. It could be left over data from deleted accounts.

## Testing Plan

Run migration on accounts at version 12, and comment out updating the transactionCache so all TX will be missing.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/migrations/data/013-wallet-2.ts
+++ b/ironfish/src/migrations/data/013-wallet-2.ts
@@ -275,7 +275,12 @@ export class Migration013 extends Migration {
     )) {
       const noteHash = Buffer.from(noteHashHex, 'hex')
       const transactionHash = await noteToTransaction.get(noteHash)
-      Assert.isNotUndefined(transactionHash)
+
+      if (!transactionHash) {
+        countMissingTx++
+        speed.add(1)
+        continue
+      }
 
       const transaction = await transactionsCache.get(transactionHash)
 


ironfish/src/migrations/data/013-wallet-2.ts
```diff
@@ -273,9 +274,27 @@ export class Migration013 extends Migration {
...
+      countNoteTotal++
+
+      if (countNote % 1000 === 0) {
+        await cacheMeta.put('countMissingAccount', countMissingAccount)
+        await cacheMeta.put('countMissingTx', countMissingTx)
+        await cacheMeta.put('countNote', countNote)
+        await cacheMeta.put('countNoteTotal', countNoteTotal)
+
+        logger.debug(
+          `Migrated ${countNote} out ${countNoteTotal} notes: ${speed.rate1s.toFixed(2)}/s`,
```
hughy, line 286: tiny nit -- `out of`

hughy (approved):

ironfish/src/migrations/data/013-wallet-2.ts
NullSoldier: Fixed

PR #2299, opened 2022-10-05
Dont materialize unspent notes when creating a tx

NullSoldier:
## Summary

This loaded all TX when creating a note and joinin its related
transaction to get its sequence.

## Testing Plan
Run existing tests, and try creating a transaction on a big account.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/fileStores/config.ts
+++ b/ironfish/src/fileStores/config.ts
@@ -303,7 +303,7 @@ export class Config extends KeyStore<ConfigOptions> {
       tlsKeyPath: files.resolve(files.join(dataDir, 'certs', 'node-key.pem')),
       tlsCertPath: files.resolve(files.join(dataDir, 'certs', 'node-cert.pem')),
       maxPeers: 50,
-      minimumBlockConfirmations: 12,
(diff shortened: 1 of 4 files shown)


ironfish/src/fileStores/config.ts
```diff
@@ -303,7 +303,7 @@ export class Config extends KeyStore<ConfigOptions> {
       tlsKeyPath: files.resolve(files.join(dataDir, 'certs', 'node-key.pem')),
       tlsCertPath: files.resolve(files.join(dataDir, 'certs', 'node-cert.pem')),
       maxPeers: 50,
-      minimumBlockConfirmations: 12,
+      minimumBlockConfirmations: 2,
```
hughy, line 306: Is this an intentional change or for testing?

hughy (approved):

ironfish/src/fileStores/config.ts
NullSoldier: Intentional, but I snuck it into this PR. I should have wrote it in the description.

PR #2302, opened 2022-10-05
Fix transaction and block size in followChainStream

mirayadav:
## Summary

Fix the transaction and block size calculations in followChainStream so that the correct sizes are displayed in the block explorer.

## Testing Plan

Uploaded blocks with the new block size calculation to a local instance of the ironfish API server with the `service:sync` command. Ran the block explorer locally with this API to check that the new block size was displayed.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/chain/followChain.ts
+++ b/ironfish/src/rpc/routes/chain/followChain.ts
@@ -1,9 +1,11 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import bufio from 'bufio'
 import * as yup from 'yup'
 import { Assert } from '../../../assert'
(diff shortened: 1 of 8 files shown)


NullSoldier:
I think this needs a rebase, it has my extra commit to bump the versions in it.

ironfish/src/rpc/routes/chain/followChain.ts
```diff
@@ -117,7 +119,7 @@ router.register<typeof FollowChainStreamRequestSchema, FollowChainStreamResponse
         return transaction.withReference(() => {
           return {
             hash: BlockHashSerdeInstance.serialize(transaction.unsignedHash()),
-            size: Buffer.from(JSON.stringify(transaction.serialize())).byteLength,
+            size: bufio.sizeVarBytes(transaction.serialize()),
```
NullSoldier, line 122: shouldn't this be just the byte length of this, or else it'll have 2 extra bytes?

NullSoldier (approved):

PR #2307, opened 2022-10-07
Show Ipc & socketRpc method list

hairtail:
## Summary
This is a hello world page for rpc users. With this PR, users can get knowledge of all supported rpc methods.
## Breaking Change

```
[ ] Yes
```


NullSoldier:
No other blockchain does this, we just need RPC docs like everyone else.

hairtail:
> No other blockchain does this, we just need RPC docs like everyone else.

Substrate does have this method. But it's better to provide RPC docs. I'll willing to do this.
https://polkadot.js.org/docs/substrate/rpc/#:~:text=Retrieves%20the%20list%20of%20RPC%20methods%20that%20are%20exposed%20by%20the%20node

PR #2308, opened 2022-10-07
Update rpc protect rules with auth

hairtail:
## Summary
1. Move transaction based routes to protected for rpc request over tcp as mentioned before. https://github.com/iron-fish/ironfish/pull/2131
2. Since we have activated auth system for rpc request over tls, we should use all namespace for request over tls and auth system is responsible for rpc security.
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/router.ts
+++ b/ironfish/src/rpc/routes/router.ts
@@ -25,7 +25,11 @@ export enum ApiNamespace {
   rpc = 'rpc',
 }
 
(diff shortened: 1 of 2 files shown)


ygao76:
Hi @hairtail , thanks for putting up this PR. You change looks good, but I will remove the protected namespace and related logics. https://github.com/iron-fish/ironfish/pull/2315

PR #2315, opened 2022-10-07
Remove protected namespace

ygao76:
## Summary
Remove protected namespace and clean up the codes
https://github.com/iron-fish/ironfish/pull/2308
## Testing Plan
unit test 
## Breaking Change

```
[ ] Yes
```
@NullSoldier

### DIFF
--- a/ironfish/src/sdk.test.ts
+++ b/ironfish/src/sdk.test.ts
@@ -151,7 +151,7 @@ describe('IronfishSdk', () => {
       expect(ipc?.namespaces).toEqual(ALL_API_NAMESPACES)
     })
 
-    it('should use all RPC namespaces for TCP with rpcTcpSecure flag', async () => {
+    it('should use all RPC namespaces for TCP', async () => {
       const sdk = await IronfishSdk.init({
         dataDir: os.tmpdir(),
         configOverrides: {
--- a/ironfish/src/sdk.test.ts
+++ b/ironfish/src/sdk.test.ts
@@ -8,7 +8,6 @@ import { IronfishNode } from './node'
 import { Platform } from './platform'
 import {
   ALL_API_NAMESPACES,
-  API_NAMESPACES_PROTECTED,
   RpcClient,
   RpcIpcAdapter,
   RpcMemoryClient,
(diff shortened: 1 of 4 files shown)


ironfish/src/sdk.ts
```diff
@@ -250,7 +242,7 @@ export class IronfishSdk {
             this.config.get('rpcTcpHost'),
             this.config.get('rpcTcpPort'),
             this.logger,
-            namespaces,
+            ALL_API_NAMESPACES,
```
hughy, line 245: Auth is still disabled for the RpcTcpAdapter. Perhaps we should get rid of the adapter altogether?

I think it's a question of whether we want to allow users the option to use TCP without encryption or without authentication

hughy (commented):
It would be nice to get rid of this 'secure' flag!

ironfish/src/sdk.ts
hairtail: I think there are two kind rpc request, public rpc request and private rpc request. For the first one like getBlock, authentication is unnecessary. But our auth system running now checks authentication for all request over tls, but no for request over tcp. And with this PR, all request can be performed over tcp without authentication. My idea in https://github.com/iron-fish/ironfish/pull/2308, use tcp for public request and tls for public + private request. All requests need authentication == no authentication?

NullSoldier:
This PR needs to be rebased it has other peoples commits in it.

NullSoldier:
I changed this PR to run against master now.

ironfish/src/sdk.test.ts
```diff
@@ -152,27 +151,6 @@ describe('IronfishSdk', () => {
...
-      )
-
-      const allowedNamespaces = ALL_API_NAMESPACES.filter(
-        (namespace) => !API_NAMESPACES_PROTECTED.includes(namespace),
-      )
-
-      expect(tcp?.namespaces.sort()).toMatchObject(allowedNamespaces.sort())
-    })
-
     it('should use all RPC namespaces for TCP with rpcTcpSecure flag', async () => {
```
NullSoldier, line 154: This test should be looked it, it refers to a deleted flag.

ironfish/src/sdk.ts
NullSoldier: Got it, you're talking about mounting custom name spaces per adapter. I think that's an interesting concept that we can do but it's outside of this PR.  One thing is that we shouldn't really pre-choose for the user what is mounted and what isn't for usability. My ideas are this.

1. Only mount IPC adapter by default ever
2. Let users enable a public adapter explicitly ```--rpc.tls or --rpc.tcp```
3. Let users manually configure what routes are enabled for each adapter 
   - `ironfish rpc:mount tls node/getStatus`
   - `ironfish rpc:unmount tls node/getStatus`
   - Store this configuration in a file in the data directory

ironfish/src/sdk.test.ts
ygao76: changed the name

ironfish/src/sdk.ts
hairtail: Thanks for your reply. @NullSoldier 
This is complicated. My ideas are this. 
1. All namespace are mounted.
2. Namespace like, node and chain are public, and no authentication needed for this kind of request.
3. Namespace like, account and transaction are private, and authentication needed for this kind of request.
4. Maybe we should refactor our namespace to make all methods of public namespace really public.

NullSoldier: I think there are a lot of instances were nodes were attacked and drained by unsuspecting node runners who didn't realize certain API's were public. I think we should lock everything 100% down if it's public to be safe. Having a public RPC is something that BTC frowned on and removed entirely, read the security section here. https://github.com/bitcoin/bitcoin/blob/master/doc/JSON-RPC-interface.md#security

So we should assume the worst, and only let absolute power users that assume the risk enable these API's publically. Attackers are 100% going to port scan and start making requests against your node in an attempt to DDOS it for sure.

NullSoldier (approved):

Issue #2317, opened 2022-10-09
Feature: Change transaction status 'unconfirmed' 


wd021:
### Description

I was looking into a transaction and noticed that it went from `pending` to `unconfirmed`. At first, i thought the transaction didn't go through until i realized `unconfirmed` is actually confirmed, awaiting block confirmations. 
```
root@dep-b:~/ironfish/ironfish-cli# yarn start accounts:transaction 9b8159dde1d92134d79d1d07359d2fad1d257e68bb5c7fe270e70513046727b0
yarn run v1.22.19
$ yarn build && yarn start:js accounts:transaction 9b8159dde1d92134d79d1d07359d2fad1d257e68bb5c7fe270e70513046727b0
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc --inspect=:0 --inspect-publish-uid=http --enable-source-maps bin/run accounts:transaction 9b8159dde1d92134d79d1d07359d2fad1d257e68bb5c7fe270e70513046727b0
Transaction: 9b8159dde1d92134d79d1d07359d2fad1d257e68bb5c7fe270e70513046727b0
Account: default
Status: pending
Miner Fee: x
Fee: $IRON 0.00000500
Spends Count: 1
Notes Count: 1
Done in 33.82s.
root@dep-b:~/ironfish/ironfish-cli# yarn start accounts:transaction 9b8159dde1d92134d79d1d07359d2fad1d257e68bb5c7fe270e70513046727b0
yarn run v1.22.19

$ yarn build && yarn start:js accounts:transaction 9b8159dde1d92134d79d1d07359d2fad1d257e68bb5c7fe270e70513046727b0
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc --inspect=:0 --inspect-publish-uid=http --enable-source-maps bin/run accounts:transaction 9b8159dde1d92134d79d1d07359d2fad1d257e68bb5c7fe270e70513046727b0
Transaction: 9b8159dde1d92134d79d1d07359d2fad1d257e68bb5c7fe270e70513046727b0
Account: default
Status: unconfirmed
Miner Fee: x
Fee: $IRON 0.00000500
Spends Count: 1
Notes Count: 1
Done in 34.44s.
root@dep-b:~/ironfish/ironfish-cli# 
```

IMO, it makes more sense to rename `unconfirmed` as `confirming`.

PR #2318, opened 2022-10-09
Update transaction statuses text

wd021:
## Summary
Update for https://github.com/iron-fish/ironfish/issues/2317. Changing `unconfirmed` to `confirming` for transaction statuses.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -29,7 +29,7 @@ export enum TransactionStatus {
   CONFIRMED = 'confirmed',
   EXPIRED = 'expired',
   PENDING = 'pending',
-  UNCONFIRMED = 'unconfirmed',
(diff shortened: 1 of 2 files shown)


ironfish/src/wallet/wallet.ts
```diff
@@ -29,7 +29,7 @@ export enum TransactionStatus {
   CONFIRMED = 'confirmed',
   EXPIRED = 'expired',
   PENDING = 'pending',
-  UNCONFIRMED = 'unconfirmed',
+  CONFIRMING = 'confirming',
```
NullSoldier, line 32: I think we should stick with the standard names that BTC and other coins use. Which is unconfirmed / confirmed, and pending some times has a different name. They match the balance names too... pending, unconfirmed, confirmed.

PR #2324, opened 2022-10-10
Add max transaction count into consensus

mirayadav:
## Summary

Add max 301 transactions (300 transactions + 1 miner's fee transaction) verification to consensus.

## Testing Plan

Wrote additional tests in `verifier.test.ts`

## Breaking Change

```
[ ] Yes
```


ironfish/src/consensus/verifier.ts
```diff
@@ -355,6 +359,17 @@ export class Verifier {
...
+      this.chain.consensus.isActive(
+        this.chain.consensus.V2_MAX_TRANSACTIONS,
+        block.header.sequence,
+      )
+    ) {
+      if (block.transactions.length > MAX_TRANSACTIONS_PER_BLOCK) {
+        return { valid: false, reason: VerificationResultReason.MAX_TRANSACTIONS_EXCEEDED }
+      }
+    }
+
```
NullSoldier, line 372: This should be in contextual-less verification

ironfish/src/consensus/consensus.ts
```diff
@@ -65,6 +65,11 @@ export const GRAFFITI_SIZE = 32
  */
 export const IRON_FISH_YEAR_IN_BLOCKS = (365 * 24 * 60 * 60) / TARGET_BLOCK_TIME_IN_SECONDS
 
+/**
+ * Max transactions allowed per block
+ */
+export const MAX_TRANSACTIONS_PER_BLOCK = 300
```
NullSoldier, line 71: This will cause the network to halt, if you look at our block explorer all our blocks have a max of 301.

ironfish/src/consensus/consensus.ts
```diff
@@ -83,5 +96,6 @@ export class TestnetParameters extends ConsensusParameters {
   constructor() {
     super()
     this.V1_DOUBLE_SPEND = 204000
+    this.V2_MAX_TRANSACTIONS = 230000
```
NullSoldier, line 99: We could get away from a hard fork declaration because no one has ever made a block with more than 301 transactions which is our miner manager soft cap.

ironfish/src/consensus/verifier.test.ts
```diff
@@ -436,4 +448,51 @@ describe('Verifier', () => {
       )
     })
   })
+
+  describe('verifyBlockConnect', () => {
```
NullSoldier, line 452: This describe name is wrong now

ironfish/src/consensus/verifier.test.ts
```diff
@@ -436,4 +448,51 @@ describe('Verifier', () => {
...
+        for (const spend of spends) {
+          yield spend
+          yield spend
+        }
+      })
+
+      expect(await chain.verifier.verifyBlockConnect(block)).toEqual({
+        valid: false,
+        reason: VerificationResultReason.DOUBLE_SPEND,
+      })
```
NullSoldier, line 487: I don't think this has anything to do with consensus, because this would always have been caught by our verifier.

NullSoldier (approved):

NullSoldier:
This looks good! I tested this manually. Luckily no one has created any blocks with more than 301 transactions.

PR #2325, opened 2022-10-11
reverts to unzipping and moving snapshot download

hughy:
## Summary

some users have run into errors when downloading snapshots when they do not have enough space for both the snapshot archive and the unzipped snapshot. unzipping directly into place has the unfortunate side-effect of leaving the node in an unrecoverable state if the unzipping operation fails: the download process removes all files from the database before beginning to unzip.

these changes revert to unzipping to a temporary location before moving the snapshot files into the database

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/chain/download.ts
+++ b/ironfish-cli/src/commands/chain/download.ts
@@ -227,6 +227,11 @@ export default class Download extends IronfishCommand {
       }
     }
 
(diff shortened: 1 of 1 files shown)


NullSoldier (approved):

NullSoldier:
Thanks, this should fix a lot of user's issues.

PR #2332, opened 2022-10-11
Create Recent Fee Cache in Queue structure

ygao76:
## Summary
Refactor Fee estimation. https://github.com/iron-fish/ironfish/pull/2075

- Create a queue class
- Create Recent Fee Cache with add, delete and get suggest fee methods

## Testing Plan
unit tests

## Breaking Change

```
[ ] Yes
```
@hughy @NullSoldier @mirayadav

### DIFF
--- a/ironfish/src/wallet/queue.ts
+++ b/ironfish/src/wallet/queue.ts
@@ -49,4 +49,11 @@ export class Queue<T> {
     }
     return this._items[index]
   }
+
+  /**
+   * Get all items in the queue
+   */
+  getAll(): T[] {
+    return this._items
+  }
 }
--- a/ironfish/src/wallet/queue.ts
+++ b/ironfish/src/wallet/queue.ts
@@ -1,0 +1,52 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+export class Queue<T> {
+  private _items: T[] = []
+
+  constructor(private capacity: number = Infinity) {}
+
+  /**
+   * Adds @param item to the queue as the last element
+   */
+  enqueue(item: T): void {
+    if (this.isFull()) {
+      throw Error('Queue has reached max capacity, you cannot add more items')
+    }
+    this._items.push(item)
+  }
+
+  /**
+   * Remove the first element in the queue. If the queue is empty, returns `undefined`
+   */
+  dequeue(): T | undefined {
+    return this._items.shift()
+  }
+
+  /**
+   * return the number of elements in the queue
+   */
+  size(): number {
+    return this._items.length
+  }
+
+  /**
+   * Returns true if the queue size reaches the capacity
+   */
+  isFull(): boolean {
+    return this.capacity === this.size()
+  }
+
+  /**
+   * Get the item in the queue by index
+   */
+  get(index: number): T | undefined {
+    if (index < 0 || index >= this._items.length) {
+      return undefined
+    }
+    return this._items[index]
+  }
+}
--- a/ironfish/src/wallet/recentFeeCache.ts
+++ b/ironfish/src/wallet/recentFeeCache.ts
@@ -1,0 +1,132 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Assert } from '../assert'
+import { Blockchain } from '../blockchain'
+import { createRootLogger, Logger } from '../logger'
+import { PriorityQueue } from '../memPool'
+import { MetricsMonitor } from '../metrics'
+import { Block, Transaction } from '../primitives'
+import { Queue } from './queue'
+
+export class RecentFeeCache {
+  private queue: Queue<Transaction>
+  readonly chain: Blockchain
+  private readonly logger: Logger
+  private readonly metrics: MetricsMonitor
+  private numOfRecentBlocks = 10
+  private numOfTxSamples = 3
+  private defaultSuggestedFee = BigInt(2)
+
+  constructor(
+    chain: Blockchain,
+    recentBlocksNum?: number,
+    txSampleSize?: number,
+    metrics?: MetricsMonitor,
+    logger?: Logger,
+  ) {
+    this.logger = logger || createRootLogger()
+    this.numOfRecentBlocks = recentBlocksNum ?? this.numOfRecentBlocks
+    this.numOfTxSamples = txSampleSize ?? this.numOfTxSamples
+
+    this.queue = new Queue<Transaction>(this.numOfRecentBlocks * this.numOfTxSamples)
+    this.chain = chain
+    this.logger.withTag('recentFeeCache')
+    this.metrics = metrics || new MetricsMonitor({ logger: this.logger })
+  }
+
+  async setUpCache(): Promise<void> {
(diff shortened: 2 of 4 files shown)


ironfish/src/wallet/queue.ts
```diff
@@ -0,0 +1,52 @@
...
+  /**
+   * Get the item in the queue by index
+   */
+  get(index: number): T | undefined {
+    if (index < 0 || index >= this._items.length) {
+      return undefined
+    }
+    return this._items[index]
+  }
+}
```
NullSoldier, line 52: The built in javascript array is a queue, so we can just do that.

ironfish/src/wallet/recentFeeCache.ts
```diff
@@ -0,0 +1,132 @@
...
+  private numOfRecentBlocks = 10
+  private numOfTxSamples = 3
+  private defaultSuggestedFee = BigInt(2)
+
+  constructor(
+    chain: Blockchain,
+    recentBlocksNum?: number,
+    txSampleSize?: number,
+    metrics?: MetricsMonitor,
+    logger?: Logger,
```
NullSoldier, line 26: We can use objects here to match our other top level items, because it gets bad when you need to expand this later and you have optionals.

ironfish/src/wallet/recentFeeCache.ts
```diff
@@ -0,0 +1,132 @@
...
+  ) {
+    this.logger = logger || createRootLogger()
+    this.numOfRecentBlocks = recentBlocksNum ?? this.numOfRecentBlocks
+    this.numOfTxSamples = txSampleSize ?? this.numOfTxSamples
+
+    this.queue = new Queue<Transaction>(this.numOfRecentBlocks * this.numOfTxSamples)
+    this.chain = chain
+    this.logger.withTag('recentFeeCache')
```
NullSoldier, line 34: This doesn't do anything

ironfish/src/wallet/recentFeeCache.ts
```diff
@@ -0,0 +1,132 @@
...
+    if (this.queue.isFull()) {
+      this.deleteOldestTransaction()
+    }
+    try {
+      this.queue.enqueue(transaction)
+      return true
+    } catch {
+      this.logger.error(`Error enqueue transaction to Recent Fee Cache`)
+      return false
+    }
```
NullSoldier, line 72: You don't want to do this type of try/catch in javascript. What error are you handling here? We should remove this try/catch.

ironfish/src/wallet/recentFeeCache.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  getSuggestedFee(percentile: number): bigint {
+    if (this.queue.size() < this.numOfRecentBlocks) {
+      return this.defaultSuggestedFee
+    }
+
+    const transaction = this.queue.get(Math.round(((this.queue.size() - 1) * percentile) / 100))
+    if (!transaction) {
+      return this.defaultSuggestedFee
```
NullSoldier, line 124: This code looks like it should be impossible to trigger

ironfish/src/wallet/recentFeeCache.ts
```diff
@@ -0,0 +1,132 @@
...
+
+      currentBlockHash = currentBlock.header.previousBlockHash
+    }
+  }
+
+  /**
+   * Add recent transactions to fee cache
+   */
+
+  addFee(transaction: Transaction): boolean {
```
NullSoldier, line 62: this is called `addFee` but doesn't take a fee, it takes a transaction. Likewise I think we should have this take a block or else we can't take advantage of bulk optimizations if we want to optimize this. We generally want to process whole blocks of transactions right?

NullSoldier: I think we should have this take a block or else we can't take advantage of bulk optimizations if we want to optimize this. We generally want to process whole blocks of transactions right?

ironfish/src/wallet/recentFeeCache.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    return transactions.reverse()
+  }
+
+  getSuggestedFee(percentile: number): bigint {
+    if (this.queue.size() < this.numOfRecentBlocks) {
+      return this.defaultSuggestedFee
+    }
+
+    const transaction = this.queue.get(Math.round(((this.queue.size() - 1) * percentile) / 100))
```
hughy, line 122: I think we need a separate copy of the queue items sorted by fee -- the queue is storing transactions in insertion order, so the percentile here won't be based on the fee amount

ironfish/src/wallet/recentFeeCache.ts
```diff
@@ -0,0 +1,132 @@
...
+import { Assert } from '../assert'
+import { Blockchain } from '../blockchain'
+import { createRootLogger, Logger } from '../logger'
+import { PriorityQueue } from '../memPool'
+import { MetricsMonitor } from '../metrics'
+import { Block, Transaction } from '../primitives'
+import { Queue } from './queue'
+
+export class RecentFeeCache {
+  private queue: Queue<Transaction>
```
hughy, line 13: Do we need anything from the transaction in the queue besides the fee rate? I think this could be a queue of numbers

ironfish/src/wallet/recentFeeCache.ts
```diff
@@ -0,0 +1,132 @@
...
+  readonly chain: Blockchain
+  private readonly logger: Logger
+  private readonly metrics: MetricsMonitor
```
hughy, line 16: We could do without the MetricsMonitor for now

ironfish/src/wallet/queue.ts
ygao76: This is queue with size, not just an array. There are enqueue logic that encapsulate in the queue class.

ironfish/src/wallet/recentFeeCache.ts
ygao76: I will rename as `addTransaction` or `addToFeeCache`.
I thought about having this function as `addTransactions`, which handle bulk. 
But before we jump into that, here is the code flow im thinking:

When mempool receives a new block:
```
 lowestFeeTransactions = recentFeeCache.getLowestTransactions(newBlock)
 lowestFeeTransactions.forEach((tx)=>{ recentFeeCache.addTransaction(tx)})
```

ygao76: I can have the bulk add which the forEach will be inside the `addTransactions` function, but overall it does not optimize the code or performance much.

ironfish/src/wallet/recentFeeCache.ts
ygao76: You are right 😆  updating

ironfish/src/wallet/recentFeeCache.ts
hughy: On second thought we may need some transaction data (e.g., block hash) to handle removing transactions when a forked block is disconnected from the chain

ygao76: sounds good!

ironfish/src/wallet/recentFeeCache.ts
ygao76: use options as input

ironfish/src/wallet/queue.ts
hughy: There is some logic around the maximum capacity, but since `enqueue` throws an error if the queue is at capacity the client still needs to keep track of whether the queue is full.

I'm not sure if the extra abstraction is needed here

hughy (approved):

ironfish/src/wallet/queue.ts
ygao76: I did not include automatically dequeue when queue is full during enqueuing. I left it for the caller to handle what to do when queue is full.

The queue class store the capacity so it knows when the queue is full. If we have the capacity at recent fee cache class is also doable. But I think this way the recent fee cache class is cleaner.

ygao76:
Merge the PR before I am leaving to unblock other task. Feel free to give comments, I will address them in a separate PR when I come back.

Issue #2326, opened 2022-10-11
`ironfish status` indicates blockchain is "SYNCED" when it is not.


neuronull:
### What happened?

`ironfish status` indicates blockchain is "SYNCED" , however it is not synced (over 12 hours).

```
$ ironfish status                                                                                                                                                        
Version              0.1.49 @ ec57ed3
Node                 STARTED
Node Name            null
Block Graffiti       null
Memory               Heap: 554.89 MiB -> 605.12 MiB / 2.01 GiB (26.9%), RSS: 2.16 GiB (17.4%), Free: 10.06 GiB (19.1%)      
CPU                  Cores: 4, Current: 102.1%
P2P Network          CONNECTED - In: 7.94 KB/s, Out: 3.11 KB/s, peers 48
Mining               STARTED - 0 miners, 0 mined
Mem Pool             Size: 198262 tx, Bytes: 304.89 MiB
Syncer               IDLE - avg time to add block 20941.05 ms
Blockchain           00000000000040e6dd68ab893aa30b230c4499816ff5f96655d68d228f7fac95 (222334), Since HEAD: 12h 35m (SYNCED)
Accounts             0000000000079e6fc5798dfd54fc2d3d814ccc5ed2f4db55e19b944b5e206dac (72069)
Telemetry            STARTED - 30692 <- 207 pending
Workers              STARTED - 51 -> 3 / 3 - 2.93 jobs Δ, 31.76 jobs/s
```

I restarted the node when it was like this for 2 hours, but it did not resolve the issue.
But I restarted the node again now after 12 hours, and it seems to be syncing now.

```
$ ironfish status
Version              0.1.49 @ ec57ed3
Node                 STARTED
Node Name            null
Block Graffiti       null
Memory               Heap: 51.74 MiB -> 121.17 MiB / 2.01 GiB (2.5%), RSS: 834.98 MiB (6.6%), Free: 11.43 GiB (8.1%)
CPU                  Cores: 4, Current: 104.7%
P2P Network          CONNECTED - In: 2.01 KB/s, Out: 2.86 KB/s, peers 40
Mining               STARTED - 0 miners, 0 mined
Mem Pool             Size: 0 tx, Bytes: 0 B
Syncer               IDLE - avg time to add block 4733.2 ms, progress: 99.66%
Blockchain           000000000000553a000b6f8d8914139699317a4b3ab2f43beda290b691c9f3f7 (222354), Since HEAD: 12h 42m (NOT SYNCED)
[...]

PR #2333, opened 2022-10-12
changes updateSynced to update to false

hughy:
## Summary

updateSynced never updates synced to false. if the node is running but falls behind the chain (e.g., due to network connectivity, thread starvation, or other interruption) then synced will not be updated to reflect that the node is out of sync until the node is restarted.

- changes synced to false if chain head timestamp doesn't meet synced criteria
- only emits onSynced if synced changes from false to true

relates to #2326

## Testing Plan

TODO: update tests that rely on `chain.synced`, which depends on time

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -1315,11 +1315,12 @@ export class Blockchain {
   }
 
   private updateSynced(): void {
-    if (this.synced) {
+    if (this.head.timestamp.valueOf() < Date.now() - MAX_SYNCED_AGE_MS) {
+      this.synced = false
       return
     }
 
-    if (this.head.timestamp.valueOf() < Date.now() - MAX_SYNCED_AGE_MS) {
+    if (this.synced) {
       return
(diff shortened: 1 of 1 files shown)


ironfish/src/blockchain/blockchain.ts
```diff
@@ -1315,11 +1315,12 @@ export class Blockchain {
   }
 
   private updateSynced(): void {
-    if (this.synced) {
+    if (this.head.timestamp.valueOf() < Date.now() - MAX_SYNCED_AGE_MS) {
+      this.synced = false
```
NullSoldier, line 1319: This has very bad implications, and basically will cause non dominant miners to not be able to mine, especially when we have issues with getting gossiped blocks right now. I strongly recommend this as it will cause the network to halt in many cases. If you want something to stop in this case, you probably want to instead check if you're connected to the network.

hughy:
Closing this PR -- if we're to change anything here we will need to take a much more nuanced approach

PR #2335, opened 2022-10-12
add getRawTransaction

hairtail:
## Summary
Support getRawTransaction, so that transaction can de decrypted locally. This is helpful since node doesn't keep notes that are spent already now. This changes in node make it difficult for transaction sender to give a proof of that he has sent the transaction.
## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/chain/getRawTransaction.ts
```diff
@@ -0,0 +1,93 @@
...
+    let parentBlock
+    if (block.header.sequence === GENESIS_BLOCK_SEQUENCE) {
+      parentBlock = block
+    } else {
+      parentBlock = await node.chain.getBlock(block.header.previousBlockHash)
+    }
+
+    if (!parentBlock) {
+      throw new ValidationError(`No parent block found`)
+    }
```
hughy, line 61: Why is `parentBlock` required?

ironfish/src/rpc/routes/chain/getRawTransaction.ts
```diff
@@ -0,0 +1,93 @@
...
+    const rawTransaction: GetRawTransactionResponse = {
+      fee: '0',
+      expirationSequence: 0,
+      notesCount: 0,
+      spendsCount: 0,
+      signature: '',
+      notesEncrypted: [],
+    }
+    block.transactions.map((transaction) => {
+      if (transaction.unsignedHash().toString('hex') === request.data.transactionHash) {
```
hughy, line 72: If the transaction that we're looking for isn't found on the block that we're looking for, should we throw an error or return the 'empty' response?

hughy:
This RPC endpoint seems to have a lot of overlap with `getBlock` and `getBlockInfo`. What other data are missing from either of those endpoints that makes a new one necessary?

hairtail:
> This RPC endpoint seems to have a lot of overlap with `getBlock` and `getBlockInfo`. What other data are missing from either of those endpoints that makes a new one necessary?

Both `getBlock` and `getBlockInfo` miss `rawTransaction` data. If I want to decode a transaction that I own locally without an ironfish node, I have to get the `rawTransaction` that contains complete `notes` instead of the `hash` of `notes`. Actually, this is what i'm working on now. I want to provide a tool with which people can verify that a transaction actually happens and he owns it, as a proof of "Tom has sent some token to Amy". 

To do this, 
1. `getRawTransaction` to get complete data. 
2. Use `oreos` for notes decryption, notes decryption has been implemented in the latest version of `oreos`.

ironfish/src/rpc/routes/chain/getRawTransaction.ts
hairtail: This is unnecessary.

ironfish/src/rpc/routes/chain/getRawTransaction.ts
hairtail: Sure things.

hughy:
> Actually, this is what i'm working on now. I want to provide a tool with which people can verify that a transaction actually happens and he owns it, as a proof of "Tom has sent some token to Amy".

That sounds awesome! I suppose I'm wondering if it makes more sense to add the necessary fields to `getBlock` or to add a new RPC endpoint?

hairtail:
> > Actually, this is what i'm working on now. I want to provide a tool with which people can verify that a transaction actually happens and he owns it, as a proof of "Tom has sent some token to Amy".
> 
> That sounds awesome! I suppose I'm wondering if it makes more sense to add the necessary fields to `getBlock` or to add a new RPC endpoint?

I think it's better to add a new endpoint. Leave `getBlock` for block, `getRawTransaction` for transaction. But, yeah, both new and field in `getBlockwork` work.

hughy:
> I think it's better to add a new endpoint. Leave getBlock for block, getRawTransaction for transaction. But, yeah, both new and field in getBlockwork work.

That makes sense! Last comment I have is to maybe use `getTransaction` instead of `Raw` -- there isn't a `getTransaction` endpoint, and I don't think it's clear what "raw" means 😅

hughy (approved):

PR #2337, opened 2022-10-13
memPool updates recent fee cache on connect block

hughy:
## Summary

adds a RecentFeeCache to the memPool. implements onConnectBlock on the RecentFeeCache to add the lowest fees from that block into the fee estimator cache.

uses the memPool to filter trnasactions from the block when updating the fee estimator cache.

- moves recentFeeCache to memPool

fixes and cleanup to recentFeeCache

- excludes minersFee transactions
- fixes sorting of fees by value instead of string representation
- fixes tests to add blocks to chain, simplifies fixture usage, and removes search through fee values
- removes 'cache' from method names

## Testing Plan

adds and updates unit tests

## Breaking Change

```
[ ] Yes
```


ironfish/src/memPool/recentFeeCache.ts
```diff
@@ -113,12 +123,21 @@ export class RecentFeeCache {
     this.queue.getAll().forEach((tx) => {
       fees.push(tx.fee())
     })
-    fees.sort()
-    const fee = fees[Math.round(((this.queue.size() - 1) * percentile) / 100)]
-    return fee
+
+    fees.sort((a, b) => {
```
NullSoldier, line 127: this should do the same thing
```ts
fees.sort((a, b) => a - b)
```

ironfish/src/memPool/recentFeeCache.ts
```diff
@@ -30,15 +30,18 @@ export class RecentFeeCache {
...
     for (let i = 0; i < this.numOfRecentBlocks; i++) {
       const currentBlock = await this.chain.getBlock(currentBlockHash)
       Assert.isNotNull(currentBlock, 'No block found')
 
-      const lowestFeeTransactions = this.getLowestFeeTransactions(currentBlock)
+      const lowestFeeTransactions = this.getLowestFeeTransactions(
+        currentBlock,
+        this.numOfTxSamples,
+      )
       lowestFeeTransactions.forEach((tx) => {
```
NullSoldier, line 45: we can just use a for loop here too, we don't need to create a closure just for this.

NullSoldier (approved):

PR #2346, opened 2022-10-14
unsignedHash as notes.transactionHash

hairtail:
## Summary
Another way to use unsignedHash https://github.com/iron-fish/ironfish/pull/2344
## Breaking Change

```
[ ] Yes
```


hairtail:
> This approach is better, but it's risky to modify the object here in this way -- the notes returned by `getNotes` won't match the notes returned by other methods.
> 
> Since we only need the unsigned hash for rendering in RPC responses, I think that this logic might be a better fit in the RPC endpoint:
> 
> https://github.com/iron-fish/ironfish/blob/ec57ed37ad7b484fb0e3573e57f3fb45bd886e7d/ironfish/src/rpc/routes/accounts/getNotes.ts#L40

Good point, updated.

hughy (approved):
Looks great, thank you!

PR #2349, opened 2022-10-14
Add block download speed to status

danield9tqh:
## Summary
We should know how long it is taking to actually download blocks from peers when syncing. This will help us determine if the syncing slowness is in download or in addBlock function.

## Testing Plan
Local testing

## Breaking Change

```
[ ] Yes
```


ironfish/src/syncer.ts
```diff
@@ -355,12 +358,17 @@ export class Syncer {
         )} (${sequence}) from ${peer.displayName}`,
       )
 
+      const start = Date.now()
```
dguenther, line 361: Maybe use BenchUtils.start for this? It's vaguely more accurate, not that it makes a huge difference, and we're using it elsewhere in syncer

ironfish-cli/src/commands/status.ts
```diff
@@ -84,7 +84,9 @@ function renderStatus(content: GetNodeStatusResponse, debugOutput: boolean): str
   const avgTimeToAddBlock = content.blockSyncer.syncing.blockSpeed
 
   if (content.blockSyncer.status === 'syncing') {
-    blockSyncerStatusDetails.push(`${content.blockSyncer.syncing.speed} blocks per seconds`)
+    blockSyncerStatusDetails.push(
+      `${content.blockSyncer.syncing.speed} blocks added /s, ${content.blockSyncer.syncing.downloadSpeed} blocks downloaded /s`,
```
dguenther, line 88: I'm open to this being an "I'm dense" issue, but the /s isn't immediately obvious to me. I wonder if it might help to remove the space like `blocks added/s` or `blocks added/sec`? I don't feel too strongly about this

image

ironfish/src/metrics/meter.ts
```diff
@@ -37,11 +41,18 @@ export class Meter {
...
     this._rate5s = new EwmAverage(this.METER_TIME_INTERVALS_MS.rate5s / this._intervalMs)
     this._rate1m = new EwmAverage(this.METER_TIME_INTERVALS_MS.rate1m / this._intervalMs)
     this._rate5m = new EwmAverage(this.METER_TIME_INTERVALS_MS.rate5m / this._intervalMs)
+
     this._average = new RollingAverage(1000)
+
+    this._rollingRate1s = new RollingAverage(1000 / this._intervalMs)
+    this._rollingRate5s = new RollingAverage(5000 / this._intervalMs)
+    this._rollingRate1m = new RollingAverage((1 * 60 * 1000) / this._intervalMs)
+    this._rollingRate5m = new RollingAverage((5 * 60 * 1000) / this._intervalMs)
```
dguenther, line 55: Want to use `this.METER_TIME_INTERVALS_MS` for these? It's used right above in the EwmAverage and seems to be the same

danield9tqh: Yea I'll use it. Although not really sure why it was needed in the first place. Just seems like more code indirection. Now that we have two places using it though I guess it helps 🤷🏻

ironfish-cli/src/commands/status.ts
danield9tqh: yea I do like the `blocks added/sec`. I'll change to that

ironfish/src/syncer.ts
danield9tqh: 👍🏻

ironfish/src/metrics/meter.ts
dguenther: Talked about this in a call, I'm okay with getting rid of the variables.

dguenther (approved):

PR #2352, opened 2022-10-14
changes RecentFeeCache to use fee rates

hughy:
## Summary

changing the block size limit from a maximum number of transactions to a maximum size in bytes will prompt miners to prioritize transactions by fee rate ($ORE/byte) instead of fee.

- implements getFeeRate to caculate the fee rate for a given transaction from its fee and its size
- changes FeeEntry to FeeRateEntry such that RecentFeeCache tracks fee rates instead of fees
- changes getSugggestedFee to estimateFeeRate

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/memPool/recentFeeCache.ts
+++ b/ironfish/src/memPool/recentFeeCache.ts
@@ -20,7 +20,7 @@ export class RecentFeeCache {
   private numOfRecentBlocks = 10
   private numOfTxSamples = 3
   private maxQueueLength: number
-  private defaultFeeRate = BigInt(2)
+  private defaultFeeRate = BigInt(1)
 
   constructor(options: {
     chain: Blockchain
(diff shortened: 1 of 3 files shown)


ironfish/src/memPool/recentFeeCache.ts
```diff
@@ -125,25 +120,17 @@ export class RecentFeeCache {
...
-    fees.sort((a, b) => {
-      if (a < b) {
-        return -1
-      } else if (a > b) {
-        return 1
-      } else {
-        return 0
-      }
-    })
+    fees.sort((a, b) => (a < b ? -1 : a > b ? 1 : 0))
```
ygao76, line 133: Can we use ```fees.sort((a, b) => a - b))```

hughy: The compareFn for sort needs to return a number, but these are bigints, so we need to work around it 😢

ygao76 (approved):

ironfish/src/memPool/recentFeeCache.ts
ygao76: I see, so need to convert a-b to number type. Either way looks fine.

PR #2353, opened 2022-10-14
implements fees:estimateFeeRates rpc endpoint

hughy:
## Summary

endpoint takes a priority level (i.e., low, medium, high) in each request and responds with the estimated fee rate.

a request without a priority level fetches the estimated fee rate for all priority levels.

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/fees/estimate.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const EstimateFeeRateRequestSchema: yup.ObjectSchema<EstimateFeeRateRequest> = yup
+  .object({
+    percentile: yup.number().defined(),
+  })
+  .defined()
+
+export const EstimateFeeRateResponseSchema: yup.ObjectSchema<EstimateFeeRateResponse> = yup
+  .object({
+    estimate: yup.number().defined(),
```
NullSoldier, line 18: This is a small nit but the result of "EstimateFeeRate" should be a fee rate right? The fee rate is the estimate.

NullSoldier (approved):

ironfish/src/memPool/feeEstimator.ts
```diff
@@ -141,6 +141,19 @@ export class FeeEstimator {
     return result
   }
 
+  estimateFeeRates(): { priority: PriorityLevel; feeRate: bigint }[] {
```
NullSoldier, line 144:
```suggestion
  /* returns an estimated fee rate as ore/kb */
  estimateFeeRates(): { priority: PriorityLevel; feeRate: bigint }[] {
```

NullSoldier: After this merges.. what do you think about just keying this? Otherwise users are gonna have to iterate the array to get the level they want. Especially since its a finite set.

hughy: I think keying is a better approach. I hadn't really thought about getting a particular level, only about getting all of them.

Since this is new functionality I think I'll make the change here before merging.

NullSoldier (approved):

Issue #2347, opened 2022-10-14
Show Last 10-100 Transactions


airj1012:
### Description

Trying to understand the status of recent transactions. It would be good to see the most recent transactions so you could look at the status without seeing a full list of transactions (ironfish accounts:transactions --show=100)

Similarly, how are the transactions sorted? Doesn't appear to be sequential.

hughy:
I think that something like this would definitely be useful!

Right now the output of transactions in the list is ordered by the database keys of the transactions rather than sequentially.

PR #2355, opened 2022-10-15
get lastest transactions

hairtail:
## Summary
Implementation https://github.com/iron-fish/ironfish/issues/2347
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/accounts/transactions.ts
+++ b/ironfish-cli/src/commands/accounts/transactions.ts
@@ -16,6 +16,9 @@ export class TransactionsCommand extends IronfishCommand {
       char: 't',
       description: 'Transaction hash to get details for',
     }),
+    show: Flags.integer({
+      description: 'Number of latest transactions to get details for',
+    }),
   }
 
   static args = [
@@ -32,7 +35,11 @@ export class TransactionsCommand extends IronfishCommand {
     const account = args.account as string | undefined
 
     const client = await this.sdk.connectRpc()
-    const response = client.getAccountTransactionsStream({ account, hash: flags.hash })
+    const response = client.getAccountTransactionsStream({
+      account,
+      hash: flags.hash,
+      show: flags.show,
+    })
 
     let showHeader = true
 
--- a/ironfish-cli/src/commands/accounts/transactions.ts
+++ b/ironfish-cli/src/commands/accounts/transactions.ts
@@ -16,7 +16,7 @@ export class TransactionsCommand extends IronfishCommand {
       char: 't',
       description: 'Transaction hash to get details for',
     }),
-    show: Flags.integer({
+    limit: Flags.integer({
       description: 'Number of latest transactions to get details for',
     }),
   }
@@ -38,7 +38,7 @@ export class TransactionsCommand extends IronfishCommand {
     const response = client.getAccountTransactionsStream({
       account,
       hash: flags.hash,
-      show: flags.show,
+      limit: flags.limit,
     })
 
     let showHeader = true
--- a/ironfish/src/rpc/routes/accounts/getTransactions.ts
+++ b/ironfish/src/rpc/routes/accounts/getTransactions.ts
@@ -62,7 +62,19 @@ router.register<typeof GetAccountTransactionsRequestSchema, GetAccountTransactio
 
     const headSequence = await node.wallet.getAccountHeadSequence(account)
     const queue = new FastPriorityQueue<TransactionValue>(function (a, b) {
-      return a.transaction.expirationSequence() < b.transaction.expirationSequence()
+      if (a.sequence && b.sequence) {
+        // both a and b are mined on chain, use sequence as sort key
+        return a.sequence < b.sequence
+      } else {
+        // at least one is in pending status, use expirationSequence as sort key
+        if (a.transaction.expirationSequence() && b.transaction.expirationSequence()) {
+          // no minersFee transaction, use expirationSequence as sort key directly
+          return a.transaction.expirationSequence() < b.transaction.expirationSequence()
+        } else {
+          // minersFee transaction without sequence is always latest
+          return b.transaction.expirationSequence() === 0
+        }
+      }
     })
 
     for await (const transaction of account.getTransactions()) {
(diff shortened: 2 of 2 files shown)


ironfish-cli/src/commands/accounts/transactions.ts
```diff
@@ -16,6 +16,9 @@ export class TransactionsCommand extends IronfishCommand {
       char: 't',
       description: 'Transaction hash to get details for',
     }),
+    show: Flags.integer({
+      description: 'Number of latest transactions to get details for',
+    }),
```
hughy, line 21: nit - I'm not sure that `show` is the best description of what this does. Maybe `limit` or `number`?

ironfish/src/rpc/routes/accounts/getTransactions.ts
```diff
@@ -66,12 +69,41 @@ router.register<typeof GetAccountTransactionsRequestSchema, GetAccountTransactio
...
+        queue.add(transaction)
+        if (queue.size > capacity) {
+          queue.poll()
+        }
+      } else {
+        if (request.closed) {
+          break
+        }
+
+        await streamTransaction(request, node, account, transaction, { headSequence })
```
hughy, line 93: The logic here is getting a bit confusing because we're supporting three different ways of fetching transactions.

It might be time to split this up into separate functions for the different use cases of this RPC route

ironfish/src/rpc/routes/accounts/getTransactions.ts
```diff
@@ -66,12 +69,41 @@ router.register<typeof GetAccountTransactionsRequestSchema, GetAccountTransactio
 
     const headSequence = await node.wallet.getAccountHeadSequence(account)
 
+    let capacity = -1
+    let queue = undefined
+    if (request.data.show) {
+      capacity = request.data.show
+      queue = new FastPriorityQueue<TransactionValue>(function (a, b) {
+        return a.transaction.expirationSequence() < b.transaction.expirationSequence()
```
hughy, line 77: This will be a min-heap, so it will give the earliest transactions instead of the latest

I'm also wondering if `expirationSequence` is the best sort key. `TransactionValue` also has `sequence` which is the sequence of the block that contains the transaction (if any)

An `expirationSequence` of `0` also means that the transaction will never expire

hughy (commented):
Thanks for taking this on!

I left a few suggestions that I think could improve this

ironfish/src/rpc/routes/accounts/getTransactions.ts
hairtail: For the 1st issue, since we use `poll` func to remove element when queue is full and new element comes,  the elements remain in the queue after all transactions are handled are the latest in my opinion.
For the 2nd issue, I have tried to use `sequence` as the sort key. But there are some pending transactions who do not have sequence, but they are the latest. So `expirationSequence` is used here now.

ironfish/src/rpc/routes/accounts/getTransactions.ts
```diff
@@ -52,26 +55,49 @@ router.register<typeof GetAccountTransactionsRequestSchema, GetAccountTransactio
...
+          request.data.limit,
+          queue,
+          false,
+          headSequence,
+        )
+      } else {
+        if (request.closed) {
+          break
+        }
+        await handleAllTransactions(request, node, account, transaction, headSequence)
```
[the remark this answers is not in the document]
hairtail: `handleLimitedTransactions`  first pushed all transactions into sorted-queue here with `endStream = false `, then called again with `endStream = true` to write all transactions in queue to socket.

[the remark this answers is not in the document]
hughy: Right, but if we're calling `handleAllTransactions`, then we never insert any transactions into the queue.

We could move all of the queue logic into `handleLimitedTransactions` and have both `handleLimitedTransaction` and `handleAllTransactions` implement separate logic for iterating over transactions.

ironfish/src/rpc/routes/accounts/getTransactions.ts
hairtail: Updated. `sequence` as sort key for transactions already mined, `expirationSequence ` as sort key for transactions already pending and minersFee also handled.

ironfish/src/rpc/routes/accounts/getTransactions.ts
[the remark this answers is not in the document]
hairtail: Got your idea now, I will make some changes later.

ironfish/src/rpc/routes/accounts/getTransactions.ts
```diff
@@ -109,3 +103,74 @@ const streamTransaction = async (
...
+    if (a.sequence && b.sequence) {
+      // both a and b are mined on chain, use sequence as sort key
+      return a.sequence < b.sequence
+    } else {
+      // at least one is in pending status, use expirationSequence as sort key
+      if (a.transaction.expirationSequence() && b.transaction.expirationSequence()) {
+        // no minersFee transaction, use expirationSequence as sort key directly
+        return a.transaction.expirationSequence() < b.transaction.expirationSequence()
+      } else {
+        // minersFee transaction without sequence is always latest
```
hughy, line 139:
```suggestion
        // transactions without expiration are always latest
```
nit -- transactions with expirationSequence === 0 aren't necessarily miners fees

ironfish/src/rpc/routes/accounts/getTransactions.ts
```diff
@@ -109,3 +103,74 @@ const streamTransaction = async (
...
+  headSequence?: number | null,
+): Promise<void> => {
+  const queue = new FastPriorityQueue<TransactionValue>(function (a, b) {
+    if (a.sequence && b.sequence) {
+      // both a and b are mined on chain, use sequence as sort key
+      return a.sequence < b.sequence
+    } else {
+      // at least one is in pending status, use expirationSequence as sort key
+      if (a.transaction.expirationSequence() && b.transaction.expirationSequence()) {
+        // no minersFee transaction, use expirationSequence as sort key directly
```
hughy, line 136:
```suggestion
```
nit -- transactions with expirationSequence === 0 aren't necessarily miners fees

hughy (approved):
Looks great, thank you!

PR #2359, opened 2022-10-16
Validate config file options (`ConfigOptions`)

neuronull:
## Summary

Fixes: https://github.com/iron-fish/ironfish/issues/1924

Adds additional validations to `ConfigOptions` by means of the `yup` API.

## Testing Plan

Added unit tests for configuration schema validations, and applying the default config against the schema.

## Breaking Change

No

## Graffiti

null

### DIFF
--- a/ironfish/src/fileStores/config.ts
+++ b/ironfish/src/fileStores/config.ts
@@ -240,10 +240,6 @@ export type ConfigOptions = {
 // Matches either an empty string, or a string that has no leading or trailing whitespace.
 const reNoWhitespaceBegEnd = /^[^\s]+(\s+[^\s]+)*$|^$/
 
-// config validation error messages
-const cfgErrMsgPathWhiteSpace = 'Path should not contain leading or trailing whitespace.'
-const cfgErrMsgURL = 'Invalid URL.'
-
 // config number value validators
 export const isWholeNumber = yup.number().integer().min(0)
 export const isPort = yup.number().integer().min(1).max(65535)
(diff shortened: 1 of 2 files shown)


ironfish/src/fileStores/config.ts
```diff
@@ -238,9 +237,86 @@ export type ConfigOptions = {
   explorerTransactionsUrl: string
 }
 
+// Matches either an empty string, or a string that has no leading or trailing whitespace.
+const reNoWhitespaceBegEnd = /^[^\s]+(\s+[^\s]+)*$|^$/
+
+// config validation error messages
+const cfgErrMsgPathWhiteSpace = 'Path should not contain leading or trailing whitespace.'
+const cfgErrMsgURL = 'Invalid URL.'
```
hughy, line 245: nit - since these strings are only used once each in `noWhitespaceBegEnd` and `isUrl`, below, I think we could use them in place there without defining separate variables

hughy (commented):
Thanks for picking up this task!

I left one small suggestion

ironfish/src/fileStores/config.ts
neuronull: Good call.
I had it this way before I added `isUrl` and `noWhitespaceBegEnd` , so was using the error messages multiple times in the config schema directly.

neuronull: Addressed in https://github.com/iron-fish/ironfish/pull/2359/commits/4f4a7c0136835c516b5bdf3826a4669b426f88bd.

hughy:
Looks good, just need to fix lint issues.

You can run `yarn lint:fix` to fix automatically where possible.

neuronull:
> Looks good, just need to fix lint issues.
> 
> You can run `yarn lint:fix` to fix automatically where possible.

Ah, good to know. I'm quite new to TS. I think this means I solved them?

```
$ yarn lint 
yarn run v1.22.19
$ tsc -b && tsc -b tsconfig.test.json && eslint --ext .ts,.tsx,.js,.jsx src/
Done in 60.07s.
```

hughy (approved):
Great! Thanks again!

Issue #2363, opened 2022-10-17
Incorrect Calculations of required balance while sending deposit TX with "deposit" command


Ryushka:
### What happened?

When using ironfish deposit command, it tries to calculate sufficient balance with additional 1 ore of fee though -f flag used to determine the fee amount. 
Like: 
Balance 0.10010000
Executing Deposit command: ironfish deposit -f 10000
Gives an error insufficient balance, Required: 0.10010000000000001!
Many zeroes after 0.1001 also concerning, there must be 4...

Though i had enough 0.1Iron + 10000Ore to execute this deposit, had to lower fee by one Ore for transaction to be executed

### Version

0.1.49

### Debug output

```shell
Iron Fish version               0.1.49 @ src
Iron Fish library               0.0.26 @ src
Operating system                Linux x64
CPU model(s)                    Intel(R) Core(TM) i7-10700K CPU @ 3.80GHz
CPU threads                     16
RAM total                       15.52 GiB
Heap total                      4.00 GiB
Node version                    v16.17.1
ironfish in PATH                true
Garbage Collector Exposed       true
Telemetry enabled               true
Node name                       BUN1
Block graffiti                  PaRUSaI
Done in 1.95s.
```

### Relevant log output

```shell
New Balance: $IRON 0.10010000 ($ORE 10,010,000)

Find the transaction on https://explorer.ironfish.network/transaction/522771972a48402f661efafb4f06b55c768d6ce63c79f7c818de10e1fd8388d6 
(it can take a few minutes before the transaction appears in the Explorer)

number of attempts: 107
Insufficient balance: 0.1001. Required: 0.10010000000000001



ironfish accounts:balance --explain
yarn run v1.22.19
$ yarn build && yarn start:js accounts:balance --explain
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc --inspect=:0 --inspect-publish-uid=http --enable-source-maps bin/run accounts:balance --explain
Account: 3d (e8863f2)

Your balance is made of notes on the chain that are safe to spend
Balance: $IRON 0.10010000

2 notes worth 0.08780623 are on the chain within 12 blocks of the head
Unconfirmed: 0.18790623

0 notes worth 0.00000000 are waiting for miners to add them to the chain
Pending: 0.18790623
Done in 2.77s.
```

### Graffiti

PaRUSaI

hughy:
Closed by #2364

PR #2364, opened 2022-10-17
Fix floating point bug

wd021:
## Summary
Fixes https://github.com/iron-fish/ironfish/issues/2363. There's a floating precision bug on [L116 in deposit.ts](https://github.com/iron-fish/ironfish/blob/master/ironfish-cli/src/commands/deposit.ts#L116). We could add a new decimal type, but comparing ORE rather than IRON circumvents this and we already do this in deposit-all.

## Testing Plan
Test `deposit` command. 

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/deposit.ts
+++ b/ironfish-cli/src/commands/deposit.ts
@@ -64,7 +64,6 @@ export default class Bank extends IronfishCommand {
       }
     }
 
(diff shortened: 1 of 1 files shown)


hughy (approved):
Thanks for this!

PR #2365, opened 2022-10-17
Simplify Rust error types

mat-if:
## Summary

Simplify ironfish-rust to single error type and create a small helper for converting to napi errors

The goal here is to make it a little easier to work with errors, spending less time thinking about which type is correct. This also makes it easier to extend functionality (implement From/Into traits, etc).

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust-nodejs/src/lib.rs
+++ b/ironfish-rust-nodejs/src/lib.rs
@@ -2,10 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+use std::fmt::Display;
+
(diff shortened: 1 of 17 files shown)


ironfish-rust-nodejs/src/lib.rs
```diff
@@ -15,6 +16,10 @@ pub mod nacl;
 pub mod rolling_filter;
 pub mod structs;
 
+fn to_napi_err(err: impl Display) -> napi::Error {
```
mat-if, line 19: This is a stop-gap to make it a little easier to convert to napi errors (which are effectively strings passed to the Javascript) until I can think of a cleaner way to deal with it

rohanjadvani (approved):

PR #2366, opened 2022-10-17
feat: Add accounts status cli command

jowparks:
## Summary
Adds the command `ironfish accounts:status` which returns various fields for each account in the wallet.

## Testing Plan
Unit test
+
```
╰─ yarn start accounts:status
yarn run v1.22.19
$ yarn build && yarn start:js accounts:status
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc --inspect=:0 --inspect-publish-uid=http --enable-source-maps bin/run accounts:status
 Account    ID                                   Head Hash                                                        Head In Chain Head Sequence 
 ────────── ──────────────────────────────────── ──────────────────────────────────────────────────────────────── ───────────── ───────────── 
 default    61acc8b1-7f3d-42b0-82a1-39f49d085b16 69e263e931fa1a2a4b0437a8eff79ffb7a353b6384a7aeac9f90ac12ae4811ef true          1             
```
## Breaking Change

```
[ ] Yes
```


jowparks:
Couple of questions, I tried to run test with
1. `yarn test update` and it says there are no updates, not sure why since I added a new test file:
```
yarn test update                                                                                                                                                                              
yarn run v1.22.19
$ tsc -b && tsc -b tsconfig.test.json && jest update
No tests found, exiting with code 1
Run with `--passWithNoTests` to exit with code 0
In /Users/joe/repos/ironfish/ironfish
  2011 files checked.
  testMatch: **/*.test.ts, **/*.test.slow.ts, **/*.test.perf.ts - 117 matches
  testPathIgnorePatterns: .*\.test\.slow\.ts$, .*\.test\.perf\.ts$ - 2001 matches
  testRegex:  - 0 matches
Pattern: update - 0 matches
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

2. How do you all enable automatic linting on file save? I tried enabling:
```
  "editor.codeActionsOnSave": {
    "source.fixAll.eslint": true
  }
```
in VSCode but it doesn't seem to autoreformat.

3. Is the test case sufficient? Should I mock out values for the head hash stuff? If so do you know where I can find an example of how to mock?

ironfish-rust-nodejs/index.d.ts
```diff
@@ -3,6 +3,12 @@
 
 /* auto-generated by NAPI-RS */
 
+export class ExternalObject<T> {
+  readonly '': {
+    readonly '': unique symbol
+    [K: symbol]: T
+  }
+}
```
NullSoldier, line 11: This keeps being generated, can we revert this change and ask the rust devs why this keeps being generated and if its ok to commit?

ironfish/src/rpc/routes/accounts/getStatus.test.ts
```diff
@@ -0,0 +1,39 @@
...
+import { createRouteTest } from '../../../testUtilities/routeTest'
+import { Account } from '../../../wallet'
+
+describe('Route account/status', () => {
+  const routeTest = createRouteTest(true)
+  let account = {} as Account
+
+  beforeAll(async () => {
+    account = await routeTest.node.wallet.createAccount(uuid())
+    await routeTest.node.wallet.setDefaultAccount(account.name)
```
NullSoldier, line 18: You don't need to call this, you can pass the second parameter to createAccount as true.

Also I don't feel strongly but since we have 1 test I would just move this into the test

ironfish/src/rpc/routes/accounts/getStatus.ts
```diff
@@ -0,0 +1,66 @@
...
+  .object({})
+  .defined()
+
+export const GetAccountStatusResponseSchema: yup.ObjectSchema<GetAccountStatusResponse> = yup
+  .object({
+    accounts: yup
+      .array(
+        yup
+          .object({
+            account: yup.string().defined(),
```
NullSoldier, line 29: can we change this to name to match the account model?

NullSoldier:
> 1. `yarn test update` and it says there are no updates, not sure why since I added a new test file:

You ran that inside of our CLI which we have no tests for. The test you added was in the iron fish package.

> 2. How do you all enable automatic linting on file save? I tried enabling:

I'm not sure, what you did should be fine. I don't use that feature personally I just use a shortcut key. I would look in the vs code extension terminal to see if there are any errors.

> 3. Is the test case sufficient? Should I mock out values for the head hash stuff? If so do you know where I can find an example of how to mock?

I think the test is fine, I don't really mock things out. If you wanted to you could write a test where you set an account to a random head hash byte array.

ironfish-rust-nodejs/index.d.ts
jowparks: https://iron-fish.slack.com/archives/C02PF56HESK/p1666109061641279

jowparks:
> You ran that inside of our CLI which we have no tests for. The test you added was in the iron fish package.
I ran that command from `ironfish/ironfish` directory. I know it is the correct dir because when I run with just `yarn test` rather than `yarn test update` it runs the test.

NullSoldier (approved):

PR #2367, opened 2022-10-17
Introduce ironfish-zkp crate, Cargo workspace

mat-if:
## Summary

Built on top of #2365 

Introduces a new crate that will be the primary place to import circuits and other Sapling related primitives. Right now, it's just a bunch of re-exports from librustzcash, but this will change as we begin to introduce multi-asset concepts. Most of the things we used from librustzcash will need to be forked, and this lets us do it in a way that is less disruptive to the codebase.

Also introduces a Cargo workspace, which uses 1 root-level Cargo.lock, Cargo.toml and /target dir. For anyone who has existing /target dirs in ironfish-rust or ironfish-rust-nodejs, those can be deleted, which should save a couple gb of disk space. This also lets us patch crates in one place, instead of having to try and make sure they're manually in sync between each Cargo.toml file.

- [ ] Get all CI stuff hooked up
- [ ] Verify that our napi binding build toolchain still works

## Breaking Change

```
[ ] Yes
```


.github/workflows/rust_ci.yml
```diff
@@ -30,9 +30,6 @@ jobs:
           sharedKey: base
           working-directory: ironfish-rust
 
-      - name: Check for license headers
```
mat-if, line 33: Just re-ordering these to be with the fmt and clippy, so that tests will run even if this fails, making the CI run a little more useful

rohanjadvani (approved):

PR #2368, opened 2022-10-18
Use bigint to calc currency to avoid float precision problem

web3Liuwentong:
## Summary
https://github.com/iron-fish/ironfish/issues/2363 This issue isn't resolved completely.
[There is a problem with the precision of the floating point number in javascript.](https://stackoverflow.com/questions/11695618/dealing-with-float-precision-in-javascript)
But **our code use number to calc the ores & iron, this is unsafe**. 

So I use bigint to represent ores, use string to represent iron, and convert every currency calc to ores calc to avoid the precision problem.

To convert between ore and iron, I use the @ethersproject/bignumber, this is an excellent project for Ethereum, and it's bignumber matches our needs.

And the MAXIMUM_IRON_AMOUNT is way beyond the Number.MAX_SAFE_INTEGER(Such as 9007199254740991 in chrome), so I convert it to the bigint

## Breaking Change

```
[] Yes

```

### DIFF
--- a/ironfish-cli/src/commands/accounts/pay.ts
+++ b/ironfish-cli/src/commands/accounts/pay.ts
@@ -3,10 +3,9 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 import {
-  displayIronAmount,
   displayIronAmountWithCurrency,
   ironToOre,
-  isValidAmount,
+  isValidIronAmount,
   isValidPublicAddress,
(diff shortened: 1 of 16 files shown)


NullSoldier (approved):
I fixed many issues and cleaned up the format of CurrencyUtils. This external library did not do correct local conversion so I removed commify entirely. It was hard coded to use commas.

PR #2369, opened 2022-10-18
Config rules of nodeWorkersMax

hairtail:
## Summary
`nodeWorkersMax = -1` is an easy way to enable all cpu cores for workerPool. Since some miners have already work with this flag for a long time, config rules running now will force them to make some changes. And most of them have no idea about physical and logic cpu cores, it's complicated to enable all cpu cores with `nodeWorkersMax = 10 or 20`. 
## Breaking Change

```
[ ] Yes
```


hughy (approved):
Thanks! Missed this in the config validation!

PR #2370, opened 2022-10-18
forked status of minersFee transaction

hairtail:
## Summary
Status of `minersFee` transaction of forked block is always 'pending' while it should be 'forked'. https://github.com/iron-fish/ironfish/issues/2371
## Breaking Change

```
[ ] Yes
```


NullSoldier:
There is no such thing as a forked transaction. Miners transactions need to be deleted from the wallet once they are disconnected from the main chain. I'm working on putting in a fix right now.

PR #2373, opened 2022-10-18
fixes database compact to use db instance

hughy:
## Summary

compact attempts to call compactRange on the instance of levelup, but it is the db that will be an instance of LevelDOWN. changes the compactRange call to use the db instance.

resolves in the 'else' case of compact. process currently exits if compaction is not feasible.

## Testing Plan

tested with recent changes to service:snapshot #2360 

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):
Ah this is a good find, and a sloppy bug I introduced. Thank you.

PR #2374, opened 2022-10-18
allows users to set datadir using env variable

hughy:
## Summary

updates datadir flag to read from 'IRONFISH_DATA_DIR' env variable. allows users to set this env variable to avoid having to pass '--datadir=...' with every command if they use a non-default datadir.

datadir is somewhat unique as a flag: it can't be set in an Iron Fish config file, since the config file is in the datadir.

## Testing Plan

- set IRONFISH_DATA_DIR, ran commands to see that datadir was set
- ran commands passing --datadir to see that datadir was overridden

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/flags.ts
+++ b/ironfish-cli/src/flags.ts
@@ -45,6 +45,7 @@ export const ConfigFlag = Flags.string({
 export const DataDirFlag = Flags.string({
   default: DEFAULT_DATA_DIR,
   description: 'The path to the data dir',
+  env: 'IRONFISH_DATA_DIR',
 })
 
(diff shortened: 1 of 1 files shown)


jowparks:
Is the order of priority then (highest to lowest)?:
1. Flag
2. env var
3. default

LGTM

jowparks (approved):

hughy:
> Is the order of priority then (highest to lowest)?:

The oclif docs don't say clearly what the priority is (https://oclif.io/docs/flags), but that seems to be it (flag, env, default) after testing

PR #2375, opened 2022-10-18
chore: update readme onboarding

jowparks:
## Summary
Updates some README info based on onboarding

## Testing Plan
N/A

## Breaking Change

N/A
```
[ ] Yes
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,6 @@ The following steps should only be used to install if you are planning on contri
 1. Run `yarn install` from the root directory to install packages (see sections below for OS specific details).
 
 ### Troubleshooting
-1. MacOS:
    - If `yarn install` fails with an error that includes "Failed to build cmake", you may need to first install cmake. For example, on macOS:
 
      1. Run `brew install cmake`, you'll need cmake version 3.15 or higher.
@@ -41,7 +40,6 @@ The following steps should only be used to install if you are planning on contri
      
      1. Run `brew install python`
 
-1. Linux:
    - If `yarn install` fails with an error that includes "Error: not found: make", "make: cc: Command not found", or "make: g++: Command not found", you may need to [install a C/C++ compiler toolchain](https://github.com/nodejs/node-gyp#on-unix).
 
      1. On Ubuntu: `apt install build-essential`
--- a/README.md
+++ b/README.md
@@ -17,14 +17,14 @@ The following steps should only be used to install if you are planning on contri
 1. Install [Node.js 16.x](https://nodejs.org/en/download/)
 1. Install [Rust](https://www.rust-lang.org/learn/get-started).
 1. Install [Yarn](https://classic.yarnpkg.com/en/docs/install).
+1. Run `yarn install` from the root directory to install packages (see sections below for OS specific details).
 1. Windows:
    1. Install [Build Tools for Visual Studio 2019](https://docs.microsoft.com/en-us/visualstudio/releases/2019/history#release-dates-and-build-numbers), including the C++ Build Tools and the Windows 10 SDK, for Rust.
    1. Rust builds also require installing [MinGW-w64 via win-builds](http://win-builds.org/doku.php/download_and_installation_from_windows).
       1. Choose `Native Windows`, `x86_64`, choose an empty directory, and click OK.
       1. On the next screen, click `Process`.
       1. Once it finishes, add the `bin` folder containing `cc` to your PATH environment variable.
-1. Run `yarn install` from the root directory to install packages.
-
+1. MacOS:
    - If `yarn install` fails with an error that includes "Failed to build cmake", you may need to first install cmake. For example, on macOS:
 
      1. Run `brew install cmake`, you'll need cmake version 3.15 or higher.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -76,6 +78,7 @@ If you want to contribute code, you must first read [our contributing guidelines
 - [iron-fish/website-testnet](https://github.com/iron-fish/website-testnet): The repo that powers [testnet.ironfish.network](https://testnet.ironfish.network)
 - [iron-fish/ironfish-api](https://github.com/iron-fish/ironfish-api): The repository that powers most Iron Fish API services.
 - [iron-fish/chain-explorer](https://github.com/iron-fish/chain-explorer): A visual tool to explore the block chain and all of its forks.
+- [iron-fish/block-explorer](https://github.com/iron-fish/block-explorer): V2 of chain explorer.
```
jowparks, line 81: Is this a correct description?

README.md
```diff
@@ -17,32 +17,34 @@ The following steps should only be used to install if you are planning on contri
...
-
+1. MacOS:
    - If `yarn install` fails with an error that includes "Failed to build cmake", you may need to first install cmake. For example, on macOS:
 
      1. Run `brew install cmake`, you'll need cmake version 3.15 or higher.
 
    - If `yarn install` fails with an error that includes "Could NOT find OpenSSL", you may need to first install OpenSSL and add an environment variable. For example, on macOS:
 
      1. Run `brew install openssl`
-     1. Run `` export OPENSSL_ROOT_DIR=`brew --prefix openssl`  ``
```
jowparks, line 35: Not sure which is preferable, I used the new line and it fixed permanently (no zshrc edits)

README.md
NullSoldier: What do you think about, "The repo that powers our block explorer at <link>"? Like the first entry?

README.md
```diff
@@ -17,32 +17,34 @@ The following steps should only be used to install if you are planning on contri
...
+     1. Run `brew link --force openssl`
```
NullSoldier, line 35: I wanted to talk about this change.

`OPENSSL_ROOT_DIR` before is read by the [node-datachannel](https://www.npmjs.com/package/node-datachannel) build scripts only, so were installing openssl in brew, and manually referencing it from brew.

By using link --force, this is including the openssl libs in your default $PATH so it'll get picked up by everything as the default openssl instead of the one included with mac.

In the [documentation](https://docs.brew.sh/How-to-Build-Software-Outside-Homebrew-with-Homebrew-keg-only-Dependencies) for link and kegs it says this,

```
A number of people in this situation are either forcefully linking keg-only tools with brew link --force or moving default system utilities out of the PATH and replacing them with manually created symlinks to the Homebrew-provided tool.


Please do not remove macOS native tools and forcefully replace them with symlinks back to the Homebrew-provided tool. Doing so can and likely will cause significant breakage when attempting to build software.```

What do you think?

File: README.md [hunk 4]
```diff
@@ -17,32 +17,34 @@ The following steps should only be used to install if you are planning on contri
 1. Install [Node.js 16.x](https://nodejs.org/en/download/)
 1. Install [Rust](https://www.rust-lang.org/learn/get-started).
 1. Install [Yarn](https://classic.yarnpkg.com/en/docs/install).
+1. Run `yarn install` from the root directory to install packages (see sections below for OS specific details).
```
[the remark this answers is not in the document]
jowparks: Ahh it may :) My issue was that the OS specific stuff was sort of mixed together, but I also didn't want to move the install command to the bottom. I will think of a better way to format

README.md
jowparks: Yea I think you are right. It might not actually be a problem for OpenSSL since LibreSSL (I think this might be a backward compatibility thing, but not sure I want to go down that rabbit hole), probably best to heir on the side of caution.

Also if this is only used for the build scripts of a dependency, then we shouldn't need it more than in the current install session. My only concern is that the user wipes the `node_modules` and then tries to reinstall later, the install script would break because the export is temporary.

Either way happy to remove.

README.md
```diff
@@ -17,14 +17,16 @@ The following steps should only be used to install if you are planning on contri
...
    1. Install [Build Tools for Visual Studio 2019](https://docs.microsoft.com/en-us/visualstudio/releases/2019/history#release-dates-and-build-numbers), including the C++ Build Tools and the Windows 10 SDK, for Rust.
    1. Rust builds also require installing [MinGW-w64 via win-builds](http://win-builds.org/doku.php/download_and_installation_from_windows).
       1. Choose `Native Windows`, `x86_64`, choose an empty directory, and click OK.
       1. On the next screen, click `Process`.
       1. Once it finishes, add the `bin` folder containing `cc` to your PATH environment variable.
-1. Run `yarn install` from the root directory to install packages.
+1. Run `yarn install` from the root directory to install packages (see sections below for OS specific details).
 
+### Troubleshooting
+1. MacOS:
```
dguenther, line 29: These aren't macOS-specific issues, (you'll encounter missing cmake and openssl on Windows as well), so I don't think we should add OS headers here to the troubleshooting. Otherwise looks good!

jowparks:
This basically amounts to no changes after comments, other than adding block explorer to readme, gonna delete.

PR #2378, opened 2022-10-19
Add V2 consensus upgrade for block size limit

mirayadav:
## Summary

Enforce block size limit of 2 MB.

Added a new consensus upgrade for this limit on block 252000, estimated to occur on October 31, 2022 at 12:30 pm ET. All users will need to upgrade by then.

## Testing Plan

Added tests to `verifier.test.ts` to confirm the block size limit is enforced after consensus upgrade V2.

Can also test by lowering the activation height (`V2_MAX_BLOCK_SIZE`) and max block size (`MAX_BLOCK_SIZE_BYTES`) and trying to sync - all blocks larger than max size set should be marked as invalid.

## Breaking Change

```
[X] Yes
```

### DIFF
--- a/ironfish/src/consensus/verifier.ts
+++ b/ironfish/src/consensus/verifier.ts
@@ -43,7 +43,6 @@ export class Verifier {
     block: Block,
     options: { verifyTarget?: boolean } = { verifyTarget: true },
   ): Promise<VerificationResult> {
-    // Verify the block size
     if (
       this.chain.consensus.isActive(
         this.chain.consensus.V2_MAX_BLOCK_SIZE,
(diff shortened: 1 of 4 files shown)


ironfish/src/consensus/verifier.ts
```diff
@@ -42,6 +43,20 @@ export class Verifier {
     block: Block,
     options: { verifyTarget?: boolean } = { verifyTarget: true },
   ): Promise<VerificationResult> {
+    // Verify the block size
```
NullSoldier, line 46: nit: I think the code is self explanatory

NullSoldier (approved):

PR #2386, opened 2022-10-20
Update Rust to 1.64.0, edition 2021

dguenther:
## Summary

This appeared to compile with no issues. It makes it easier to run Criterion, which only officially supports the previous 3 Rust versions, and staying current is generally good. Rustup should handle automatically downloading the newer version.

## Breaking Change

```
[ ] Yes
```


mat-if (approved):
🙌

PR #2399, opened 2022-10-22
stop wallet rebroadcasting expired transactions

hughy:
## Summary

we rebroadcast any transactions in the wallet that have not been added to a block. however, now that we continue to store expired transactions in the wallet we must ensure that we do not rebroadcast them.

rebroadcasting expired transactions also updates them in the wallet db with the current head as submittedSequence. when this update is made they are added to the wallet db store for pending transaction hashes. they then 're-expire' and their notes are reverted to unspent (even if they have later been spent in another transaction).

## Testing Plan

added unit test

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):

ironfish/src/wallet/wallet.ts
```diff
@@ -839,6 +839,16 @@ export class Wallet {
...
+        // Skip expired transactions
+        if (
+          this.chain.verifier.isExpiredSequence(
+            transaction.expirationSequence(),
+            this.chain.head.sequence,
+          )
+        ) {
+          continue
+        }
+
```
NullSoldier, line 851: we should change this whole method to only loop over the expiring transactions store that was added.

NullSoldier (approved):

NullSoldier (approved):

PR #2400, opened 2022-10-22
Use queue map to store recent fee rates

ygao76:
## Summary
Use a queue map to store the recent fee rates. Each queue corresponding to a certain percentile of the transaction fee rate among all the transactions in the block

The percentile follows the same configuration as MetaMask gas fee estimator https://github.com/MetaMask/controllers/blob/main/src/gas/fetchGasEstimatesViaEthFeeHistory/calculateGasFeeEstimatesForPriorityLevels.ts

## Testing Plan
unit test
## Breaking Change

```
[ ] Yes
```
@hughy


ironfish/src/memPool/feeEstimator.ts
```diff
@@ -46,101 +53,118 @@ export class FeeEstimator {
...
       )
 
-      for (const transaction of lowestFeeTransactions) {
-        if (this.isFull()) {
-          break
+      this.percentiles.forEach((percentile, i) => {
+        const priorityLevel = PERCENTILES_TO_PRIORITY_LEVELS.get(percentile)
+
+        if (priorityLevel) {
+          const queue = this.queues.get(priorityLevel)
```
hughy, line 65: I'm wondering if it makes sense here to iterate over the queues instead of the percentiles?

With the percentiles we need to look up the priority level and the queue and check that each is non-null. We could also include the percentile in the queues data structure so that it's a `Map<PriorityLevel, [Percentile, FeeRateEntry[]]>`

ironfish/src/memPool/feeEstimator.ts
```diff
@@ -46,101 +53,118 @@ export class FeeEstimator {
...
-    }
+    })
+  }
 
-    const transactions: Transaction[] = []
+  private getPercentileFeeRateTransactions(
+    block: Block,
+    percentiles: number[],
+    exclude: (transaction: Transaction) => boolean = (t) => t.isMinersFee(),
+  ): Transaction[] {
```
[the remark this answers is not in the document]
ygao76: updated

ironfish/src/memPool/feeEstimator.ts
ygao76: I updated to iterate over the queues, it has fewer if conditions to check. I also changed the `getPercentileFeeRateEntry` to return a map of percentile and fee rate entry. So its easier to look up the fee rate entry to add to queue.

ironfish/src/memPool/feeEstimator.ts
```diff
@@ -53,105 +64,115 @@ export class FeeEstimator {
...
+      const feeRateEntryList = this.getPercentileFeeRateEntries(
         currentBlock,
-        this.numOfTxSamples,
+        this.percentiles,
       )
 
-      for (const transaction of lowestFeeTransactions) {
-        if (this.isFull()) {
-          break
+      this.queues.forEach((queue, priorityLevel) => {
```
hughy, line 72: We can use a for loop here to avoid creating closures

ironfish/src/memPool/feeEstimator.ts
```diff
@@ -53,105 +64,115 @@ export class FeeEstimator {
...
       (t) => !memPool.exists(t.hash()),
-    )) {
-      if (this.isFull()) {
-        this.queue.shift()
-      }
+    )
 
-      this.queue.push({ feeRate: getFeeRate(transaction), blockHash: block.header.hash })
-    }
+    this.queues.forEach((queue, priorityLevel) => {
```
hughy, line 95: Here too, we can avoid creating closures

ironfish/src/memPool/feeEstimator.ts
```diff
@@ -53,105 +64,115 @@ export class FeeEstimator {
...
+    })
   }
 
   onDisconnectBlock(block: Block): void {
-    while (this.queue.length > 0) {
-      const lastEntry = this.queue[this.queue.length - 1]
-      if (!lastEntry.blockHash.equals(block.header.hash)) {
-        break
-      }
+    this.queues.forEach((queue) => {
```
hughy, line 110: Same re: looping and closures

ironfish/src/memPool/feeEstimator.ts
```diff
@@ -53,105 +64,115 @@ export class FeeEstimator {
...
-        transactions.push(transaction)
-      }
+    if (sortedTransaction.length === 0) {
+      return new Map()
     }
 
-    return transactions.reverse()
+    const result = new Map<Percentile, FeeRateEntry>()
+
+    percentiles.map(
```
hughy, line 138: Same re: looping and closures

ironfish/src/memPool/feeEstimator.ts
```diff
@@ -4,43 +4,54 @@
...
-  private queue: Array<FeeRateEntry>
+  private queues: Map<PriorityLevel, Array<FeeRateEntry>>
   readonly chain: Blockchain
   private wallet: Wallet
   private readonly logger: Logger
   private numOfRecentBlocks = 10
-  private numOfTxSamples = 3
-  private maxQueueLength: number
   private defaultFeeRate = BigInt(1)
+  private readonly percentiles = PRIORITY_LEVEL_PERCENTILES.map((x) => x)
```
hughy, line 43: Do we need `percentiles` or can we use the `PRIORITY_LEVEL_PERCENTILES` constant?

hughy:
In a later PR I think it might be useful for the 10, 20, 30 percentiles to be configurable. What do you think?

ygao76:
> In a later PR I think it might be useful for the 10, 20, 30 percentiles to be configurable. What do you think?

Yes, good idea and I'm also thinking about it. Where should we store the configure? In configure.json?

hughy (approved):
Looks good pending lint warnings!

PR #2407, opened 2022-10-24
Convert some deposit-all numbers to bigints

NullSoldier:
## Summary

I noticed some currency amounts in deposit all were in numbers that should be ore. We should _always_ take strings and parse them into bigints for ore amounts.

```
Balance
Confirmed:   $IRON 0.00021458
Unconfirmed: $IRON 0.00021458
Pending:     $IRON 0.00021458

--- Completed Transactions (0) ---
```

## Testing Plan
I ran the command manually and looked at the UI.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/deposit-all.ts
+++ b/ironfish-cli/src/commands/deposit-all.ts
@@ -4,11 +4,11 @@
 
 import {
   Assert,
-  displayIronAmountWithCurrency,
+  BigIntUtils,
+  CurrencyUtils,
   ERROR_CODES,
   ironToOre,
   MINIMUM_IRON_AMOUNT,
-  oreToIron,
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/commands/deposit-all.ts
```diff
@@ -111,7 +121,7 @@ export default class DepositAll extends IronfishCommand {
       this.client,
       this.api,
       expirationSequenceDelta,
-      fee,
+      Number(fee),
```
NullSoldier, line 124: This is still broken, and all of this needs to be fixed in another PR. verifyCanSend is broken because it's accepting ore but treating it like iron.

jowparks (approved):

PR #2409, opened 2022-10-24
Update mining transaction picking logic to use fee rate instead of fee

mirayadav:
## Summary

- Update mining manager to skip transactions that are larger than max block size
- Update mempool to order transactions by highest fee rate (ore/KB) instead of just fee

## Testing Plan

Updated `orderedTransactions` test in `memPool.test.ts`.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/memPool/memPool.test.ts
+++ b/ironfish/src/memPool/memPool.test.ts
@@ -99,26 +99,38 @@ describe('MemPool', () => {
   describe('orderedTransactions', () => {
     const nodeTest = createNodeTest()
 
-    it('returns transactions from the node mempool sorted by fees', async () => {
+    it('returns transactions from the node mempool sorted by fee rate', async () => {
       const { node } = nodeTest
       const { wallet, memPool } = node
       const accountA = await useAccountFixture(wallet, 'accountA')
       const accountB = await useAccountFixture(wallet, 'accountB')
       const accountC = await useAccountFixture(wallet, 'accountC')
-      const { transaction: transactionA } = await useBlockWithTx(node, accountA, accountB)
-      const { transaction: transactionB } = await useBlockWithTx(node, accountB, accountC)
+      const { block, transaction: transactionA } = await useBlockWithTx(
+        node,
+        accountA,
+        accountB,
+      )
+      const transactionB = block.transactions[0]
       const { transaction: transactionC } = await useBlockWithTx(node, accountC, accountA)
 
-      jest.spyOn(transactionA, 'fee').mockImplementationOnce(() => BigInt(1))
+      const transactionASize = getTransactionSize(transactionA.serialize())
+      const transactionBSize = getTransactionSize(transactionB.serialize())
+      const transactionCSize = getTransactionSize(transactionC.serialize())
+
+      // Miner's fee transaction has a smaller size than the normal transaction fixture
+      Assert.isTrue(transactionBSize < transactionASize)
+      Assert.isEqual(transactionASize, transactionCSize)
+
+      jest.spyOn(transactionA, 'fee').mockImplementationOnce(() => BigInt(4))
       jest.spyOn(transactionB, 'fee').mockImplementationOnce(() => BigInt(4))
-      jest.spyOn(transactionC, 'fee').mockImplementationOnce(() => BigInt(3))
+      jest.spyOn(transactionC, 'fee').mockImplementationOnce(() => BigInt(1))
 
       memPool.acceptTransaction(transactionA)
       memPool.acceptTransaction(transactionB)
       memPool.acceptTransaction(transactionC)
 
       const transactions = Array.from(memPool.orderedTransactions())
-      expect(transactions).toEqual([transactionB, transactionC, transactionA])
+      expect(transactions).toEqual([transactionB, transactionA, transactionC])
     })
 
     it('does not return transactions that have been removed from the mempool', async () => {
--- a/ironfish/src/memPool/memPool.test.ts
+++ b/ironfish/src/memPool/memPool.test.ts
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { Assert } from '../assert'
-import { getEmptyBlockSize, getTransactionSize } from '../network/utils/serializers'
+import { getBlockWithMinersFeeSize, getTransactionSize } from '../network/utils/serializers'
 import { Transaction } from '../primitives'
 import { createNodeTest, useAccountFixture, useBlockWithTx } from '../testUtilities'
 import { getFeeRate } from './feeEstimator'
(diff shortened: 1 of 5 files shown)


ironfish/src/memPool/memPool.ts
```diff
@@ -89,6 +92,12 @@ export class MemPool {
     return this.transactions.has(hash)
   }
 
+  getFeeRate(mempoolEntry: MempoolEntry): bigint {
+    const rate = mempoolEntry.fee / BigInt(Math.round(mempoolEntry.sizeKb))
+
+    return rate > 0 ? rate : BigInt(1)
+  }
```
NullSoldier, line 99: This should be on the memPool entry to optimize sorting since this will get run multiple times in the compare function.

ironfish/src/mining/manager.ts
```diff
@@ -104,8 +104,12 @@ export class MiningManager {
         nullifiers.add(spend.nullifier)
       }
 
-      // Stop adding transactions when the addition would cause the block to exceed the max size
       const transactionSize = getTransactionSize(transaction.serialize())
+      if (transactionSize > this.chain.consensus.MAX_BLOCK_SIZE_BYTES) {
+        continue
+      }
```
NullSoldier, line 110: Should our mem pool just reject these in its acceptFunction?

ironfish/src/memPool/memPool.ts
```diff
@@ -47,10 +48,12 @@ export class MemPool {
...
     this.queue = new PriorityQueue<MempoolEntry>(
       (firstTransaction, secondTransaction) => {
-        if (firstTransaction.fee === secondTransaction.fee) {
+        const firstTransactionFeeRate = this.getFeeRate(firstTransaction)
+        const secondTransactionFeeRate = this.getFeeRate(secondTransaction)
+        if (firstTransactionFeeRate === secondTransactionFeeRate) {
           return firstTransaction.hash.compare(secondTransaction.hash) > 0
         }
-        return firstTransaction.fee > secondTransaction.fee
+        return firstTransactionFeeRate > secondTransactionFeeRate
```
NullSoldier, line 56: It's miner but it would be nice if this was consistent with https://github.com/iron-fish/ironfish/blob/master/ironfish/src/primitives/blockheader.ts#L28-L42

ironfish/src/mining/manager.ts
dguenther: Agree, and the practical limit will be something like MAX_BLOCK_SIZE - block header - miner fee (though it'll fail the below check anyway)

ironfish/src/memPool/memPool.test.ts
```diff
@@ -99,26 +99,38 @@ describe('MemPool', () => {
...
       const accountB = await useAccountFixture(wallet, 'accountB')
       const accountC = await useAccountFixture(wallet, 'accountC')
-      const { transaction: transactionA } = await useBlockWithTx(node, accountA, accountB)
-      const { transaction: transactionB } = await useBlockWithTx(node, accountB, accountC)
+      const { block, transaction: transactionA } = await useBlockWithTx(
+        node,
+        accountA,
+        accountB,
+      )
+      const transactionB = block.transactions[0]
```
dguenther, line 113: I think it'd be worth using the `block.minersFee` helper here -- not a huge deal, but makes it a bit clearer.

ironfish/src/memPool/memPool.ts
mirayadav: Updated to calculate fee rate when creating the memPool entry. Also, now that we calculate the fee rate using the transaction, I just reused the getFeeRate function from feeEstimator.

ironfish/src/mining/manager.ts
mirayadav: Updated to reject these transaction in acceptTransaction and to use (MAX_BLOCK_SIZE  - block header size - miner's fee transaction size - transaction length size) limit. Right now I'm using 562 and 2 bytes directly as the values for the miner's fee transaction size and transaction length size in `manager.ts` and `memPool.ts`, do you guys think it would be better if we had these as exported consts somewhere?

dguenther: Yeah, that's a good idea. Now we'll be doing (blockheader + 562 + 2) in several places as well, so I think it'd be better to have that calculation centralized somewhere.

Another thing that came to mind is that miner's fee transactions aren't guaranteed to be a fixed size. I don't think this needs to change since we don't currently support generating miner transactions with multiple outputs, but just something to keep in mind for the future.

ironfish/src/memPool/memPool.ts
```diff
@@ -123,6 +124,14 @@ export class MemPool {
     const hash = transaction.hash().toString('hex')
     const sequence = transaction.expirationSequence()
 
+    if (
+      getTransactionSize(transaction.serialize()) >
+      this.chain.consensus.MAX_BLOCK_SIZE_BYTES - getBlockHeaderSize() - 562 - 2
```
NullSoldier, line 129: What are these 2 numbers?  It would be nice to write a comment here, this is gonna be hard for someone else to figure out. In that case I'd prefer if this was just a variable as MAX_TRANSACTION_SIZE_BYTES somewhere that had documentation on why its what it is.

This is trying to encode "remaining size after a block header", but that won't work if the block header ever becomes variable size.

ironfish/src/memPool/memPool.test.ts
```diff
@@ -99,26 +99,38 @@ describe('MemPool', () => {
...
       const { transaction: transactionC } = await useBlockWithTx(node, accountC, accountA)
 
-      jest.spyOn(transactionA, 'fee').mockImplementationOnce(() => BigInt(1))
+      const transactionASize = getTransactionSize(transactionA.serialize())
+      const transactionBSize = getTransactionSize(transactionB.serialize())
+      const transactionCSize = getTransactionSize(transactionC.serialize())
+
+      // Miner's fee transaction has a smaller size than the normal transaction fixture
+      Assert.isTrue(transactionBSize < transactionASize)
+      Assert.isEqual(transactionASize, transactionCSize)
```
jowparks, line 122: nit: you could add some assertions here with `getFeeRate`

dguenther: Yeah, I think we could get rid of the `Assert`s and add `expect`s for these after mocking the fee (it'll probably have to use `mockImplementation` rather than `mockImplementationOnce`)

mirayadav: Got rid of `Assert`s and added `expect`s for `getFeeRate`.

ironfish/src/network/utils/serializers.ts
```diff
@@ -114,14 +118,20 @@ export function readBlock(reader: bufio.BufferReader): SerializedBlock {
...
-  size += 2 // transactions length
+  size += BLOCK_TRANSACTIONS_LENGTH_BYTES
   for (const transaction of block.transactions) {
     size += getTransactionSize(transaction)
   }
 
   return size
 }
 
+export function getEmptyBlockSize(): number {
```
dguenther, line 129: I think this should be named `getBlockWithMinersFeeSize` just for a little more clarity, since it's a bit unclear whether a miner's fee transaction is included in an empty block or not. Otherwise this looks good to go, I'm going to pre-approve 👍

dguenther (approved):

ironfish/src/network/utils/serializers.ts
mirayadav: Renamed

PR #2410, opened 2022-10-25
Estimate transaction size

ygao76:
## Summary
Estimate transaction size before post it to the network. This will be used for fee estimation, together with Gas Fee Rate Estimator.

Add serialize function in Transaction in Rust package 

## Testing Plan
unit test 

## Breaking Change

```
[x] Yes
```
@hughy @NullSoldier


hughy (commented):
I think we can estimate the size of the transaction without creating one. A transaction has a few fields of fixed size (spendsLength, notesLength, fee, expirationSequence, signature), a list of spends, and a list of notes. Each spend and each note has fixed size.

https://github.com/iron-fish/ironfish/blob/master/ironfish/src/primitives/transaction.ts#L29

Creating a transaction is slow, and we usually do it off the main thread, so I think we should try to avoid it if we can for the size estimation

hughy:
`transaction.spend` and `transaction.receive` both construct zk proofs, but we'll end up not using those results and recomputing when we create the transaction again.

The number of spends will be the number of notes required to cover the amount sent in the transaction, and the number of notes in the output will be equal to the number of receivers plus one for change to the sender.

There are a couple of edge cases that can affect these numbers:
1. There could be extra spends required to cover the fee
2. There might not be any change coming back to the sender

ironfish/src/memPool/transactionSizeEstimator.ts
```diff
@@ -0,0 +1,61 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { createRootLogger, Logger } from '../logger'
+import { Wallet } from '../wallet'
+import { Account } from '../wallet/account'
+import { NotEnoughFundsError } from '../wallet/errors'
+
+const SPEND_SERIALIZED_SIZE_IN_BYTE_PERCENTILES = 388
+const NOTE_SERIALIZED_SIZE_IN_BYTE_PERCENTILES = 467
```
hughy, line 10: Why percentiles?

ironfish/src/memPool/transactionSizeEstimator.ts
```diff
@@ -0,0 +1,61 @@
...
+        `Insufficient funds: Needed ${amountNeeded.toString()} but have ${amount.toString()}`,
+      )
+    }
+
+    size += notesToSpend.length * SPEND_SERIALIZED_SIZE_IN_BYTE_PERCENTILES
+
+    size += receives.length * NOTE_SERIALIZED_SIZE_IN_BYTE_PERCENTILES
+
+    if (estimateFeeRate) {
+      amountNeeded += estimateFeeRate * BigInt(Math.ceil(size / 1000))
```
hughy, line 48: Should this amount subtract the `amount` returned from `createSpends`, above?

We might have some left over in the last spend that we could use for the fee

ironfish/src/memPool/transactionSizeEstimator.ts
ygao76: typo

ironfish/src/memPool/transactionSizeEstimator.ts
ygao76: Updated, pls take a look if I understand it correct

hughy (approved):

ironfish/src/memPool/feeEstimator.ts
```diff
@@ -139,6 +146,58 @@ export class FeeEstimator {
...
+
+    if (estimateFeeRate) {
+      const additionalAmountNeeded =
+        estimateFeeRate * BigInt(Math.ceil(size / 1000)) - (amount - amountNeeded)
+
+      if (additionalAmountNeeded > 0) {
+        const { notesToSpend: additionalNotesToSpend } = await this.wallet.createSpends(
+          sender,
+          additionalAmountNeeded,
+        )
```
hughy, line 191: @ygao76 I realized there's a problem with this second call to `createSpends` -- it will iterate over the same notes that have already been included in `notesToSpend`

Issue #2412, opened 2022-10-25
Feature:  Show all notes in getTransaction


hairtail:
### Description

`getTransaction` should show all information about a transaction. But we show `incoming notes` only now. In this situation, if i send some token to someone, Tom. I can't show the proof about what i have done in my node with `accounts:transaction`.

hughy:
Implemented in #2413 

Thanks again for the contribution!

PR #2413, opened 2022-10-25
show note for spender in getTransaction

hairtail:
## Summary
Add spender note information for `getTransaction` route. https://github.com/iron-fish/ironfish/issues/2412
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/account.ts
+++ b/ironfish/src/wallet/account.ts
@@ -461,6 +461,19 @@ export class Account {
           hash: noteHash,
         })
       }
+
+      const decryptedNoteForSpender = note.decryptNoteForSpender(this.outgoingViewKey)
+      if (decryptedNoteForSpender) {
+        notes.push({
+          accountId: this.id,
+          note: decryptedNoteForSpender,
+          spent: true,
+          transactionHash: transaction.hash(),
+          index: null,
+          nullifier: null,
+          hash: noteHash,
+        })
+      }
     }
 
     return notes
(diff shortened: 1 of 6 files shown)


ironfish/src/wallet/account.ts
```diff
@@ -461,6 +461,19 @@ export class Account {
           hash: noteHash,
         })
       }
+
+      const decryptedNoteForSpender = note.decryptNoteForSpender(this.outgoingViewKey)
```
hughy, line 465: Decrypting notes is slow, so we're trying to never do decryption on the main thread and to move it to worker tasks instead.

It should also be possible to look up a spent note for an account using the nullifier and the nullifierToNoteHash data store

hughy: You're right, we only store notes that the account owns!

Sorry, I misunderstood the intent here and was thinking of including the notes that are _spent_ in a transaction.

It would be useful to have a way for Tom to verify that he sent a particular note to Amy. We do want to avoid decrypting in the main thread, though, especially for every note one at a time. For a transaction with many output notes, like a pool payout, this would take a very long time.

I think we'll want to either move this decryption to a worker task, or update the wallet to have some storage of output notes from the transactions that an account creates.

hairtail: Note decryption has moved to workerPool in the latest commit.

dguenther:
I tweaked the structure of this a bit, but I still need to test it & make sure tests pass before it's ready for merge.

hairtail:
> I tweaked the structure of this a bit, but I still need to test it & make sure tests pass before it's ready for merge.

Sure things. Thanks for your attention.

dguenther (approved):

PR #2414, opened 2022-10-26
Normalization currency

hairtail:
## Summary
1. Replace `BigInt` & `toString` with `CurrencyUtils.decode` & `CurrencyUtils.encode` 
2. Use `bigint` for fee in `verifyCanSend`
3. Remove unnecessary convert.
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/deposit.ts
+++ b/ironfish-cli/src/commands/deposit.ts
@@ -112,7 +112,7 @@ export default class Bank extends IronfishCommand {
       this.client,
       this.api,
       expirationSequenceDelta,
-      Number(fee),
+      fee,
       graffiti,
     )
     if (!canSend) {
@@ -124,7 +124,7 @@ export default class Bank extends IronfishCommand {
     const balanceResp = await this.client.getAccountBalance({ account: accountName })
     const confirmedBalance = BigInt(balanceResp.content.confirmed)
 
-    if (confirmedBalance < BigInt(fee) + minDepositOre) {
+    if (confirmedBalance < fee + minDepositOre) {
       this.log(
         `Insufficient balance: ${CurrencyUtils.renderIron(
           confirmedBalance,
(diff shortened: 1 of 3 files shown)


ironfish-cli/src/commands/deposit.ts
```diff
@@ -119,10 +119,14 @@ export default class Bank extends IronfishCommand {
...
 
     const balanceResp = await this.client.getAccountBalance({ account: accountName })
-    const confirmedBalance = BigInt(balanceResp.content.confirmed)
-    const requiredBalance = CurrencyUtils.decodeIron(IRON_TO_SEND) + BigInt(fee)
+    const confirmedBalance = CurrencyUtils.decode(balanceResp.content.confirmed)
+    const requiredBalance = CurrencyUtils.decodeIron(IRON_TO_SEND) + fee
     if (confirmedBalance < requiredBalance) {
-      this.log(`Insufficient balance: ${confirmedBalance}. Required: ${requiredBalance}`)
+      this.log(
+        `Insufficient balance: ${CurrencyUtils.encode(
```
NullSoldier, line 126: I think were thinking about a way to make this more clear, but you should use `CurrencyUtils.renderIron()` when displaying to humans.

ironfish-cli/src/commands/deposit.ts
```diff
@@ -119,10 +119,14 @@ export default class Bank extends IronfishCommand {
...
+          confirmedBalance,
+        )}. Required: ${CurrencyUtils.encode(requiredBalance)}`,
```
NullSoldier, line 128: Same here with `CurrencyUtils.renderIron()`

hairtail: fixed

jowparks:
Thanks for pulling in the changes from staging. LGTM

jowparks (approved):

NullSoldier (approved):

PR #2415, opened 2022-10-26
reverses blocks when setting up fee estimator

hughy:
## Summary

when initializing the fee estimator we use transactions from recent blocks without filtering transactions using the mempool. these blocks should be added to the queue such that the most recent block is at the end of the queue rather than the beginning.

- reverses queue when initializing fee estimator
- changes 'setUp' to 'init' to match other modules in the codebase
- makes chain a parameter of init instead of a property of the fee estimator since the chain is only used during init
- calls init within node.start

## Testing Plan

adds unit test to cover initialization with multiple blocks

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/memPool/feeEstimator.ts
+++ b/ironfish/src/memPool/feeEstimator.ts
@@ -62,6 +62,10 @@ export class FeeEstimator {
         this.queue.push({ feeRate: getFeeRate(transaction), blockHash: currentBlockHash })
       }
 
+      if (currentBlockHash.equals(chain.genesis.hash)) {
+        break
+      }
+
       currentBlockHash = currentBlock.header.previousBlockHash
     }
 
--- a/ironfish/src/memPool/feeEstimator.ts
+++ b/ironfish/src/memPool/feeEstimator.ts
@@ -44,6 +44,10 @@ export class FeeEstimator {
   }
 
   async init(chain: Blockchain): Promise<void> {
+    if (chain.isEmpty) {
+      return
+    }
+
     let currentBlockHash = chain.latest.hash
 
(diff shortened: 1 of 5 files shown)


ironfish/src/memPool/feeEstimator.ts
```diff
@@ -41,16 +40,14 @@ export class FeeEstimator {
     this.maxQueueLength = this.numOfRecentBlocks * this.numOfTxSamples
 
     this.queue = []
-    this.chain = options.wallet.chain
```
NullSoldier, line 44: you don't wanna keep this here in the constructor?

ironfish/src/node.ts
```diff
@@ -318,6 +318,8 @@ export class IronfishNode {
     // so we should start it as soon as possible
     this.workerPool.start()
 
+    await this.memPool.feeEstimator.init(this.chain)
```
NullSoldier, line 321: This looks a bit different than the others now. We used to have

- open: open any DB connections and load initial data (should not depend on other systems)
- close: close any db connections and unload any data
- start: start any active event loops and systems that passively run
- start: stop any active event loops and systems that passively run

ironfish/src/memPool/feeEstimator.ts
hughy: I removed `chain` as a property of the fee estimator because it's only needed during `init`. Open to restoring it though!

ironfish/src/node.ts
hughy: Ah it sounds like this would be a better fit in `open`. I'll see what it takes to move it there

PR #2416, opened 2022-10-26
adds service command to send fee rate estimates to telemetry

hughy:
## Summary

periodically fetches fee rate estimates from the estimateFeeRate RPC and submits them to telemetry.

- converts fee rate to iron/kb in order to submit float value instead of bigint

- fetches low, medium, high fee rates from rpc

## Testing Plan

tested with local node, logging estimates instead of submitting to telemetry

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/service/estimate-fee-rates.ts
+++ b/ironfish-cli/src/commands/service/estimate-fee-rates.ts
@@ -1,0 +1,100 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { CurrencyUtils, PromiseUtils } from '@ironfish/sdk'
+import { WebApi } from '@ironfish/sdk'
+import { Flags } from '@oclif/core'
+import { IronfishCommand } from '../../command'
+import { RemoteFlags } from '../../flags'
+import { IronfishCliPKG } from '../../package'
+
+export default class EstimateFees extends IronfishCommand {
+  static hidden = true
+
+  static description = `
+     Meaures fee rate estimates and submits them to telemetry API
+   `
+
+  static flags = {
+    ...RemoteFlags,
+    endpoint: Flags.string({
+      char: 'e',
+      parse: (input: string) => Promise.resolve(input.trim()),
+      required: false,
+      description: 'API host to sync to',
+    }),
+    token: Flags.string({
+      char: 't',
+      parse: (input: string) => Promise.resolve(input.trim()),
+      required: false,
+      description: 'API host token to authenticate with',
+    }),
+    delay: Flags.integer({
+      char: 'd',
+      required: false,
+      default: 60000,
+      description: 'Delay (in ms) to wait before uploading fee rate estimates',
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/service/estimate-fee-rates.ts
```diff
@@ -0,0 +1,104 @@
...
+
+        await api.submitTelemetry({
+          points: [
+            {
+              measurement: 'fee_rate_estimate',
+              timestamp: new Date(),
+              fields: [
+                {
+                  name: `fee_rate_low`,
+                  type: 'integer',
```
ygao76, line 81: will the fee rate has decimal?

ironfish-cli/src/commands/service/estimate-fee-rates.ts
```diff
@@ -0,0 +1,104 @@
...
+import { Flags } from '@oclif/core'
+import { IronfishCommand } from '../../command'
+import { RemoteFlags } from '../../flags'
+import { IronfishCliPKG } from '../../package'
+
+export default class EstimateFees extends IronfishCommand {
+  static hidden = true
+
+  static description = `
+     Meaures fee rate estimates and submits them to telemetry API
```
ygao76, line 15: typo, Measures

ygao76 (approved):
Overall look good to me

ironfish-cli/src/commands/service/estimate-fee-rates.ts
```diff
@@ -0,0 +1,104 @@
...
+                  type: 'integer',
+                  value: feeRateHigh,
+                },
+              ],
+              tags: [{ name: 'version', value: IronfishCliPKG.version }],
+            },
+          ],
+        })
+      }
+
```
jowparks, line 100: Is there a reason why you want to run this via CLI? I just recently used a pattern that sets up a graphile worker on the API to periodically run a job and submit metrics.
There are some changes mixed in for `minAndMaxDeposit` but should be simple to parse out. You might have a good reason for doing this way, just thought I would check :) 

https://github.com/iron-fish/ironfish-api/pull/1053/files

ironfish-cli/src/commands/service/estimate-fee-rates.ts
hughy: The rate we return will always be a bigint in ORE/kb, so no decimal

ironfish-cli/src/commands/service/estimate-fee-rates.ts
hughy: We need to hit a running node's RPC endpoint for the fee estimates.

Now that we have authentication for RPC we could build out RPC calls from the API to a node, but I don't think we've used that pattern before. It seems like it could be lighter-weight than spinning up an EC2 instance to run this!

This will follow the pattern we've used for the syncer and deposit syncer -- a small EC2 instance runs a CLI command that pulls data from a node and sends it to the API.

ironfish-cli/src/commands/service/estimate-fee-rates.ts
hughy, line 15:
```suggestion
     Measures fee rate estimates and submits them to telemetry API
```

PR #2417, opened 2022-10-26
Refactor SpendParams, OutputParams to be builder-style

mat-if:
## Summary

The goal of this PR was to push the heavy logic into the `.post`/`.build` functions, so a transaction can be crafted without a lot of computation, such as generating the proofs. This also required some changes to the `ProposedTransaction`, since it expected the proofs to be built when they are added. By moving all of the logic into `.post`, we were able to de-duplicate some computation, as well as split up the proof-generation, signature creation and spend-signing into cleaner steps within the `.post` function.

## Breaking Change

```
[ ] Yes
```


ironfish-rust/src/spending.rs
```diff
@@ -62,132 +44,138 @@ pub struct SpendParams {
...
     /// contains the root-hash at the time the witness was created and the path
     /// to verify the location of that note in the tree.
-    pub fn new(
-        spender_key: SaplingKey,
-        note: &Note,
-        witness: &dyn WitnessTrait,
-    ) -> Result<SpendParams, IronfishError> {
-        // This is a sanity check; it would be caught in proving the circuit anyway,
-        // but this gives us more information in the event of a failure
-        if !witness.verify(&MerkleNoteHash::new(note.commitment_point())) {
```
mat-if, line 85: I want to specifically call this out - this doesn't exist in the refactor (yet). Re-adding it does one of two things:
- We add it in the constructor, which requires `add_spend` to return a `Result`, and is extra computation during the setup instead of in the build phase
- We add it to the build function which will require keeping track of the witness object passed in from JS, which may lead to needing awkward lifetimes or something. 

Curious to hear others thoughts on this one. Still thinking about it.

ironfish-rust/src/merkle_note.rs
```diff
@@ -48,7 +50,7 @@ pub struct MerkleNote {
     pub(crate) note_commitment: Scalar,
 
     /// Public part of ephemeral diffie-hellman key-pair. See the discussion on
-    /// keys::shared_secret to understand how this is used
+    /// [`shared_secret`] to understand how this is used
```
mat-if, line 53: We can link in docs! This lets us A) go-to-definition in editors and B) would create links in the auto-generated rust docs, if/when we start using that.

ironfish-rust/src/merkle_note.rs
```diff
@@ -93,9 +89,51 @@ impl MerkleNote {
...
+            note,
+            value_commitment,
+            diffie_hellman_keys,
+            note_encryption_keys,
+        )
+    }
+
+    /// Helper function to instantiate a MerkleNote with pre-set
+    /// note_encryption_keys. Should only be used for miners fee transactions.
+    pub(crate) fn new_for_miners_fee(
```
mat-if, line 105: This whole code path feels a bit crufty, but was necessary to set the hard-coded encryption keys before the build step is executed. I think eventually, I'd like to create wrappers specifically for miners fee so we can separate edge-cased logic that exists specifically for miners fees from regular code paths.

ironfish-rust/src/spending.rs
```diff
@@ -62,132 +44,138 @@ pub struct SpendParams {
...
+}
 
-    /// Sign this spend with the stored private key, and return a SpendProof
-    /// suitable for serialization.
-    ///
-    /// Verifies the proof before returning to prevent posting broken
-    /// transactions
-    pub fn post(&self, signature_hash: &[u8; 32]) -> Result<SpendProof, IronfishError> {
-        let private_key = redjubjub::PrivateKey(self.spender_key.spend_authorizing_key);
+pub struct UnsignedSpendProof {
```
mat-if, line 144: This sort of defines the in-between state that exists during the transaction building. Building the transaction requires the proofs to be generated, and signing the spends requires the transaction to be built. It feels cleaner than mashing it all in one step as it is now. This is ultimately what lets us remove a lot of the duplicated code between the spend/output params and spend/output proofs.

ironfish-rust/src/transaction/mod.rs
```diff
@@ -80,8 +75,6 @@ pub struct ProposedTransaction {
 impl ProposedTransaction {
     pub fn new(spender_key: SaplingKey) -> ProposedTransaction {
         ProposedTransaction {
-            binding_signature_key: <jubjub::Fr as Field>::zero(),
```
mat-if, line 83: We're no longer keeping track of these throughout the entire life-cycle. Instead we calculate them during the build phase. This removes the ability to `check_value_consistency` at each point of adding spends/outputs, but it's not something we were really doing anyway.

ironfish-rust/src/transaction/mod.rs
```diff
@@ -91,30 +84,18 @@ impl ProposedTransaction {
...
     /// Spend the note owned by spender_key at the given witness location.
-    pub fn spend(&mut self, note: &Note, witness: &dyn WitnessTrait) -> Result<(), IronfishError> {
-        let proof = SpendParams::new(self.spender_key.clone(), note, witness)?;
-
-        self.binding_signature_key += proof.value_commitment.randomness;
-        self.binding_verification_key += proof.value_commitment();
-
-        self.spends.push(proof);
+    pub fn add_spend(&mut self, note: Note, witness: &dyn WitnessTrait) {
         self.transaction_fee += note.value() as i64;
```
NullSoldier, line 88: every spends value is added to the transaction_fee variable, weird.

ironfish-rust/src/transaction/mod.rs
```diff
@@ -91,30 +84,18 @@ impl ProposedTransaction {
...
 
     /// Spend the note owned by spender_key at the given witness location.
-    pub fn spend(&mut self, note: &Note, witness: &dyn WitnessTrait) -> Result<(), IronfishError> {
-        let proof = SpendParams::new(self.spender_key.clone(), note, witness)?;
-
-        self.binding_signature_key += proof.value_commitment.randomness;
-        self.binding_verification_key += proof.value_commitment();
-
-        self.spends.push(proof);
+    pub fn add_spend(&mut self, note: Note, witness: &dyn WitnessTrait) {
```
NullSoldier, line 87: Later, this will need to take a spend key here right?

ironfish-rust/src/transaction/mod.rs
mat-if: Yeah agreed, it should be called `value_balance` or something. We'll be refactoring that soon when adding the asset identifier as part of value.

rohanjadvani (approved):

PR #2420, opened 2022-10-27
Use RpcSocketAdapter base for RpcIpcAdapter

NullSoldier:
## Summary

This deletes all the IpcAdapter code and uses the SocketAdapter at a base.

## Testing Plan
Start an new adapter and try to connect with the old IpcClient from production, and vise versa.

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/adapters/socketAdapter/socketAdapter.ts
```diff
@@ -80,6 +80,12 @@ export abstract class RpcSocketAdapter implements IRpcAdapter {
     this.inboundTraffic.start()
     this.outboundTraffic.start()
 
+    if (this.listen.path) {
+      await fsAsync.unlink(this.listen.path).catch(() => {
```
danield9tqh, line 84: Yea interesting. I wonder if we'd ever run into problems if a node was started while another node is already running. Right it doesn't get past the DB in use error so we don't have any issues.

danield9tqh (approved):

PR #2422, opened 2022-10-27
Combine all socket clients into RpcSocketClient

NullSoldier:
## Summary

This combines all the clients into 1 implementation of RpcSocketClient.

## Testing Plan
Connect with various clients, including prod client against staging source.

## Breaking Change

```
[ ] Yes
```


hughy (approved):
Awesome!

Do we have any dependencies on `node-ipc` remaining after this?

ironfish/src/rpc/clients/ipcClient.ts
```diff
@@ -1,157 +1,12 @@
...
-    if (connection.mode === 'ipc' && !connection.socketPath) {
-      throw new Error('No IPC socket path given to connect to.')
-    }
-
-    if (connection.mode === 'tcp' && (!connection.host || !connection.port)) {
-      throw new Error('No IPC socket path given to connect to.')
-    }
-
-    const { IPC } = await import('node-ipc')
-    const ipc = new IPC()
```
dguenther, line 51: We should test this on Windows, if you haven't already -- looks like there may be some weirdness around how the IPC paths need to be specified that node-ipc was abstracting over https://nodejs.org/api/net.html#identifying-paths-for-ipc-connections

ironfish/src/rpc/clients/socketClient.ts
```diff
@@ -2,60 +2,53 @@
...
-  | {
-      mode: 'tcp'
-      host: string
-      port: number
-    }
+export type RpcSocketClientConnectionInfo = {
+  path?: string
+  host?: string
+  port?: number
+}
```
dguenther, line 33: I think this is more confusing than the previous RpcClientConnectionInfo type -- on the naming, "path" could be a combination of "host" and "port". And I think the correct usage of RpcClientConnectionInfo is to specify either fields for ipc-mode, or fields for tcp-mode, and the previous type made it clear which fields were relevant to which connection mode.

dguenther:
> Awesome!
> 
> Do we have any dependencies on `node-ipc` remaining after this?

I think the ipcAdapter is the remaining piece!

ironfish/src/rpc/clients/ipcClient.ts
NullSoldier: Fixed.

PR #2425, opened 2022-10-27
refactor(rust,ironfish): Update MerkleNote/NoteEncrypted to OutputDescription

rohanjadvani:
## Breaking Change

```
[ ] Yes
```


ironfish/src/blockchain/blockchain.ts
```diff
@@ -76,10 +76,10 @@ export class Blockchain {
   synced = false
   opened = false
   notes: MerkleTree<
-    NoteEncrypted,
-    NoteEncryptedHash,
-    SerializedNoteEncrypted,
-    SerializedNoteEncryptedHash
+    OutputDescription,
```
rohanjadvani, line 79: I'm open to renaming this to `noteCommitments` if we want to be more explicit.

PR #2427, opened 2022-10-27
moves note/spend size constants to primitives

hughy:
## Summary

moves constants for noteEncrypted and spend serialized byte sizes to primitives modules for NoteEncrypted and Spend.

adds documentation of return value of estimateFeeRate.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/primitives/spend.ts
+++ b/ironfish/src/primitives/spend.ts
@@ -4,6 +4,8 @@
 
 import { Nullifier } from './nullifier'
 
+export const SPEND_SERIALIZED_SIZE_IN_BYTE = 388
+
 export interface Spend {
   nullifier: Nullifier
   commitment: Buffer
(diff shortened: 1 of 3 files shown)


ironfish/src/primitives/spend.ts
```diff
@@ -4,6 +4,8 @@
 
 import { Nullifier } from './nullifier'
 
+export const SPEND_SERIALIZED_SIZE_IN_BYTE = 388
```
hughy, line 7: It feels a bit odd to define this alongside `Spend`, which doesn't contain most of the bytes that are in a spend as serialized on a transaction, but I'm not sure that it fits better anywhere else.

ygao76: Once TransactionBuilder is ready, we should be able to get rid of the use of constants.

ygao76 (approved):

PR #2428, opened 2022-10-27
Estimate fee RPC endpoint

ygao76:
## Summary
RPC for estimate fee. This will return the suggested fee for a proposed transaction.
## Testing Plan
unit test
## Breaking Change

@hughy @NullSoldier 
```
[ ] Yes
```


ironfish/src/rpc/routes/fees/estimateFee.ts
```diff
@@ -0,0 +1,74 @@
...
+  `${ApiNamespace.fees}/estimateFee`,
+  EstimateFeeRequestSchema,
+  async (request, node): Promise<void> => {
+    const account = node.wallet.getAccountByName(request.data.fromAccountName)
+
+    if (!account) {
+      throw new ValidationError(`No account found with name ${request.data.fromAccountName}`)
+    }
+
+    const priority = request.data.priority || 'medium'
```
hughy, line 56: If there's no priority passed in I'm wondering if it's better to default to medium or to fetch estimates for all levels?

When this is used in the CLI `pay` and `deposit` commands, will we make separate requests for each priority level?

hughy (approved):

ironfish/src/rpc/routes/fees/estimateFee.test.ts
```diff
@@ -0,0 +1,59 @@
...
+  beforeAll(async () => {
+    await routeTest.node.wallet.createAccount('existingAccount', true)
+    const node = routeTest.node
+    const { block } = await useBlockWithTx(node, undefined, undefined, true, {
+      fee: 1,
+    })
+    await node.chain.addBlock(block)
+    await node.wallet.updateHead()
+  })
+
```
jowparks, line 20: It doesn't look like there is a test in here for the different priority levels, do they exist already elsewhere?

ygao76: Yeah, the test with different priority level is in [https://github.com/iron-fish/ironfish/ironfish/src/memPool/feeEstimator.test.ts](url)

ironfish/src/rpc/routes/fees/estimateFee.ts
ygao76: updated the interface

ironfish/src/rpc/routes/fees/estimateFee.ts
```diff
@@ -0,0 +1,79 @@
...
+        amount: BigInt(receive.amount),
+        memo: receive.memo,
+      }
+    })
+
+    const feeEstimator = node.memPool.feeEstimator
+
+    const low = await feeEstimator.estimateFee('low', account, receives)
+    const medium = await feeEstimator.estimateFee('medium', account, receives)
+    const high = await feeEstimator.estimateFee('high', account, receives)
```
hughy, line 71: For a followup PR - this will result in iterating over the notes for a transaction three times.

Maybe when TransactionBuilder lands we should explore caching a built transaction between calls to estimateFee and sendTransaction?

hughy (approved):

ironfish/src/rpc/routes/fees/estimateFee.ts
ygao76: sounds good!

PR #2434, opened 2022-10-28
Remove unused element

hairtail:
## Summary
`priority` in request data is never used. 
## Breaking Change

```
[ ] Yes
```


hughy (approved):
Good catch, thanks!

PR #2437, opened 2022-10-28
Add batch hashing for new merkle tree elements

dguenther:
## Summary

Replaces `rehashRightPath` in the merkle tree with `hashTree`, which allows for rehashing more than just the most recent element. Currently, we calculate new hashes for the right path for every element added to the tree, and the pedersen hashes used in our notes merkle tree are fairly slow. We also are only interested in the root hash of the tree after adding all notes to the merkle tree, so the intermediate hashes are unnecessary.

As an example, our blocks can have 300 transactions, and a typical transaction will have 2 notes. I found that the merkle tree performs 6180 hashes when adding 600 notes to the tree, but we should only need to calculate 1 hash per node in the optimal case. So, I'd expect to reduce the time spent in rehashRightPath to roughly 1/10 its current time.

These performance profiles demonstrate the result:

### Before (Adding first 1000 blocks to the chain)

image

### After (Adding first 1000 blocks to the chain)

image

## Testing Plan

Added new tests covering the addBatch function.

## Breaking Change

```
[ ] Yes
```


yarn.lock
```diff
@@ -9828,6 +9828,13 @@ set-blocking@^2.0.0, set-blocking@~2.0.0:
   resolved "https://registry.yarnpkg.com/set-blocking/-/set-blocking-2.0.0.tgz#045f9782d011ae9a6803ddd382b24392b3d890f7"
   integrity sha1-BF+XgtARrppoA93TgrJDkrPYkPc=
 
+set-getter@0.1.1:
```
danield9tqh, line 9831: Can we remove this

dguenther: Yep, removed

ironfish/src/merkletree/merkletree.test.ts
```diff
@@ -146,6 +208,59 @@ describe('Merkle tree', function () {
...
+    await tree.addBatch('jklmnopqrstuvwxyz')
+    await expect(tree).toHaveLeaves(
+      'abcdefghijklmnopqrstuvwxyz',
+      [
+        1, 1, 2, 2, 4, 4, 7, 7, 8, 8, 12, 12, 13, 13, 15, 15, 16, 16, 21, 21, 22, 22, 24, 24,
+        25, 25,
+      ],
+    )
+    await expect(tree).toHaveNodes([
+      [1, Side.Left, 3, '<c|d-0>'],
```
danield9tqh, line 221: I think this format is pretty difficult to read. You think there is a way we could assert something that makes this easier to reason about? Maybe some helper functions to construct the hashes?

ironfish/src/blockchain/blockchain.ts
```diff
@@ -969,56 +971,6 @@ export class Blockchain {
...
-  async addNote(index: number, note: NoteEncrypted, tx?: IDatabaseTransaction): Promise<void> {
-    return this.db.withTransaction(tx, async (tx) => {
-      const noteCount = await this.notes.size(tx)
-
-      // do we have a note at this index already?
-      if (index < noteCount) {
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        const oldNote = (await this.notes.get(index, tx))!
-        if (!note.equals(oldNote)) {
-          const message = `Tried to insert a note, but a different note already there for position ${index}`
```
danield9tqh, line 981: Was trying to think of any cases where this assertion would happen. I'm wondering why this doesn't happen when we truncate the tree when disconnecting a block and then try to add another block. When a fork happens for instance

danield9tqh: Ok I see we were kind of checking the bounds when we did 
```javascript
const noteCount = await this.notes.size(tx)
```

danield9tqh (approved):

PR #2438, opened 2022-10-28
Push V2 consensus upgrade (block size) to activate at block 255000

mirayadav:
## Summary

Added the consensus upgrade for max block limit to block 255000, estimated to occur on November 2, 2022 at 1:30 pm ET. All users will need to upgrade by then. Relevant changes are included in https://github.com/iron-fish/ironfish/pull/2378.

## Breaking Change

```
[X] Yes
```


hughy (approved):
Will this be ~24 hours after release of v0.1.50?

mirayadav:
> Will this be ~24 hours after release of v0.1.50?

Starting a release of v0.1.50 now, so this should be ~48 hours after.

PR #2439, opened 2022-10-28
increases precision of fee calculation

hughy:
## Summary

defers division by 1000 in conversion from (bytes * ore/kb) to ore. rounding transaction size before computing fee leads to imprecise fee calculation.

## Breaking Change

```
[ ] Yes
```


ygao76 (approved):
The change looks good. 
Do we not want to round up transaction size in kb? Because the transaction size is tend to smaller than the actual size. Because we calculate it roughly, and did not count extra receive and expiration sequence

PR #2441, opened 2022-10-28
fixes silent failures in chain:download

hughy:
## Summary

uses 'strict' mode during unzip to treat errors as failures (including ENOSPC errors). warns users that downloading and unzipping will require 2x the size of the snapshot in free disk space.

## Testing Plan

- Filled the disk of my AWS node instance, then tried to unzip another snapshot 😈 

image

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/chain/download.ts
+++ b/ironfish-cli/src/commands/chain/download.ts
@@ -86,10 +86,12 @@ export default class Download extends IronfishCommand {
       }
 
       const fileSize = FileUtils.formatFileSize(manifest.file_size)
+      const spaceRequired = FileUtils.formatFileSize(manifest.file_size * 2)
 
       if (!flags.confirm) {
         const confirm = await CliUx.ux.confirm(
-          `Download ${fileSize} snapshot to update from block ${node.chain.head.sequence} to ${manifest.block_sequence}?` +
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/chain/download.ts
```diff
@@ -86,10 +86,12 @@ export default class Download extends IronfishCommand {
       }
 
       const fileSize = FileUtils.formatFileSize(manifest.file_size)
+      const spaceRequired = FileUtils.formatFileSize(manifest.file_size * 2)
 
       if (!flags.confirm) {
         const confirm = await CliUx.ux.confirm(
-          `Download ${fileSize} snapshot to update from block ${node.chain.head.sequence} to ${manifest.block_sequence}?` +
+          `Download ${fileSize} snapshot to update from block ${node.chain.head.sequence} to ${manifest.block_sequence}? ` +
+            `At least ${spaceRequired} of free disk space is required to download and unzip the snapshot file.` +
```
ygao76, line 94: nit: maybe add a new line for `At least ...`

ygao76 (approved):

ironfish-cli/src/commands/chain/download.ts
hughy: updated!

PR #2442, opened 2022-10-28
Refactor out seed genesis block from addBlock

danield9tqh:
## Summary
Refactor out seeding the genesis block from other add block operations. We have a lot of genesis block specific logic in addBlock that unnecessarily complicates things. Factoring this out into its own function

## Testing Plan
Unit tests and ran a new chain locally for the first ~1000 blocks. Restarted the node and ran again from 1000 - 1200, no errors.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/blockchain.test.ts
+++ b/ironfish/src/blockchain/blockchain.test.ts
@@ -732,17 +732,6 @@ describe('Blockchain', () => {
     })
   })
 
-  it('reject block with null previous hash', async () => {
-    const { node } = await nodeTest.createSetup()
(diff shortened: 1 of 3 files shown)


ironfish/src/blockchain/blockchain.test.ts
```diff
@@ -732,17 +732,6 @@ describe('Blockchain', () => {
     })
   })
 
-  it('reject block with null previous hash', async () => {
```
danield9tqh, line 735: Removing this test because `verifyBlockAdd` can no longer be called without a previous block

PR #2444, opened 2022-10-29
Rpc connect error

hairtail:
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/clients/tlsClient.ts
+++ b/ironfish/src/rpc/clients/tlsClient.ts
@@ -23,9 +23,7 @@ export class RpcTlsClient extends RpcTcpClient {
         client.off('secureConnection', onSecureConnect)
         client.off('error', onError)
 
-        if (ErrorUtils.isConnectRefusedError(error)) {
-          reject(new RpcConnectionRefusedError())
-        } else if (ErrorUtils.isNoEntityError(error)) {
+        if (ErrorUtils.isConnectFailError(error)) {
           reject(new RpcConnectionRefusedError())
         } else {
           reject(error)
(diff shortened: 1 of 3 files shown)


ironfish/src/rpc/clients/tlsClient.ts
```diff
@@ -23,9 +23,7 @@ export class RpcTlsClient extends RpcTcpClient {
         client.off('secureConnection', onSecureConnect)
         client.off('error', onError)
 
-        if (ErrorUtils.isConnectRefusedError(error)) {
-          reject(new RpcConnectionRefusedError())
-        } else if (ErrorUtils.isNoEntityError(error)) {
+        if (ErrorUtils.isConnectFailError(error)) {
           reject(new RpcConnectionRefusedError())
```
hughy, line 27: We have other error types defined alongside `RpcConnectionRefusedError` that might be more appropriate depending on the error.

For example, if `isConnectTimeOutError` or `isConnectResetError`, then it might be better to use [RpcConnectionLostError](https://github.com/iron-fish/ironfish/blob/02939ae309a98205781908fd31e218aee4e62fea/ironfish/src/rpc/clients/errors.ts#L33)

hairtail: Updated.

danield9tqh:
Hey @hairtail thanks for the contribution. Looks like a good change to me. Can you add some description of the problem it solves in the `Summary` section and then we can merge

danield9tqh (approved):

danield9tqh:
On startup the node tries to connect to a node that's already running. If that connection fails it tries to open the local database itself. This PR adds some other errors (`ETIMEDOUT` and `ECONNRESET`) to that failure condition. This way if we receive either of those two errors now we will try to open the local database instead of crashing.

hairtail:
> Hey @hairtail thanks for the contribution! Next time can you add a description of the problem in the `Summary` section

I will do that next time.😅

PR #2445, opened 2022-10-29
Negative transaction amount

hairtail:
## Summary
`sendTransaction` rpc endpoint accepts negative amount and fee now, which should be forbidden.
## Breaking Change

```
[ ] Yes
```


ironfish/src/utils/currency.ts
```diff
@@ -97,6 +97,18 @@ export class CurrencyUtils {
...
   }
+
+  static isValidOre(amount: string): boolean {
+    try {
+      const ore = this.decode(amount)
+      return ore >= MINIMUM_ORE_AMOUNT && ore <= MAXIMUM_ORE_AMOUNT
+    } catch (e) {
+      if (ErrorUtils.isNodeError(e) && e.code === 'NUMERIC_FAULT') {
+        return false
+      }
```
hughy, line 108: Would `decode` throw these kinds of errors? It looks like `BigInt('0.5')` would throw a `RangeError`, which this wouldn't catch.

ironfish/src/rpc/routes/transactions/sendTransaction.ts
```diff
@@ -91,6 +92,24 @@ router.register<typeof SendTransactionRequestSchema, SendTransactionResponse>(
       )
     }
 
+    // Check whether amount and fee are valid or not
+    transaction.receives.map((receive) => {
```
hughy, line 96: nit - we already iterate over `transaction.receives` twice in this function on lines 116 and 135. Maybe we should consolidate that logic to a single loop?

hughy (approved):
Excellent, thank you!

PR #2449, opened 2022-10-31
fix negative sequence

aditya-manit:
## Summary
`ironfish blocks:show 0`
displays nothing
`ironfish blocks:show [anything negative]`
display the block with sequence 1

## Testing Plan
I have setup the repo in local and have tested locally with:
`yarn start blocks:show [any negative number]`

## Breaking Change
No

## Side Note
1. Correct me if I am wrong but negative sequence aren't possible right ? 
2. Also I know this seems like a small PR but the amount of time it took me to setup repo in local and to resolve all errors so that I can was huge, requesting you to consider that while allocating points


hughy:
Providing a negative sequence number as the argument to `blocks:show` searches relative to the head of the chain: `-1` yields the most recent block, `-2` the second most recent, etc.

aditya-manit:
Got it

PR #2452, opened 2022-10-31
add checks on user input

aditya-manit:
## Summary
When running chain:export command, there is no check on user inputs

Example:

If user runs chain:export 1000 20000 but chain or their node is synced till block 200
then only 200th block will be exported, without informing the user that it is not what they requested, or what they request isn't available

or If user runs: chain:export -100
Then from 1 -> last synced block will be exported, without informing the user that their input is wrong

## Testing Plan
Tested locally

## Breaking Change
No

### DIFF
--- a/ironfish-cli/src/commands/chain/export.ts
+++ b/ironfish-cli/src/commands/chain/export.ts
@@ -49,14 +49,23 @@ export default class Export extends IronfishCommand {
 
     const client = await this.sdk.connectRpc()
 
+    const argStart = args.start as number | null
+    const argStop = args.stop as number | null
     const stream = client.exportChainStream({
-      start: args.start as number | null,
-      stop: args.stop as number | null,
+      start: argStart,
+      stop: argStop,
     })
-
     const { start, stop } = await AsyncUtils.first(stream.contentStream())
+    if (
+      (argStart && argStart < Number(GENESIS_BLOCK_SEQUENCE)) ||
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/chain/export.ts
```diff
@@ -49,14 +49,23 @@ export default class Export extends IronfishCommand {
...
-
     const { start, stop } = await AsyncUtils.first(stream.contentStream())
+    if (
+      (argStart && argStart < Number(GENESIS_BLOCK_SEQUENCE)) ||
+      (argStop != null && argStop > stop)
+    ) {
+      this.error(
+        `You can can only export chain from ${Number(GENESIS_BLOCK_SEQUENCE)} -> ${stop}`,
+      )
+      this.exit(1)
```
hughy, line 66: This will break valid uses of the export command: negative values for start or end export blocks relative to the end of the chain.

aditya-manit: Got it, but I guess we may still want some check on the upper end ? For example chain is at 100 and someone requested upto 200, in that case do we want a message to tell the users what the highest they can request ?

If yes I can fix this PR to have only that change

aditya-manit: Okay Sure

aditya-manit: Updated the PR to reflect the same

Issue #2173, opened 2022-09-13
Feature: Wallet restore height


wd021:
We should add block height metadata to accounts to eliminate the unnecessary scanning of genesis to the height at which the wallet was created. Should also add this to export/import functionality.

**New account Structure**
```
name
spendingKey
incomingViewKey
outgoingViewKey
publicAddress
rescan
restoreHeight
```

Additional thought regarding wallet UX:

Block height isn't very user-friendly for mainstream users. I think this feature would be better exposed to users as selecting the date range when wallet was created and in the backend, we use a mapping to identify the height at which scanning starts.

hughy:
Implemented as account 'birthday' in #3719

PR #2453, opened 2022-10-31
Add from sequence for accounts:rescan

hairtail:
## Summary
Same as https://github.com/iron-fish/ironfish/pull/2128
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/accounts/rescan.ts
+++ b/ironfish-cli/src/commands/accounts/rescan.ts
@@ -29,11 +29,14 @@ export class RescanCommand extends IronfishCommand {
       default: false,
       description: 'Force the rescan to not connect via RPC',
     }),
+    from: Flags.integer({
+      description: 'Sequence to start account rescan from',
+    }),
   }
 
   async start(): Promise<void> {
     const { flags } = await this.parse(RescanCommand)
-    const { follow, reset, local } = flags
+    const { follow, reset, local, from } = flags
 
     if (local && !follow) {
       this.error('You cannot pass both --local and --no-follow')
@@ -45,7 +48,7 @@ export class RescanCommand extends IronfishCommand {
       stdout: true,
     })
 
-    const response = client.rescanAccountStream({ reset, follow })
+    const response = client.rescanAccountStream({ reset, follow, from })
 
     const speed = new Meter()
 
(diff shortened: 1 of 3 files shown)


wd021:
I would opt for adding this to [account import](https://github.com/iron-fish/ironfish/issues/2173) and not rescan. More user-friendly imo.

hairtail:
> I would opt for adding this to [account import](https://github.com/iron-fish/ironfish/issues/2173) and not rescan. More user-friendly imo.

Most people use this for reason that they want to start a rescan and they actually know when this account was created, instead of a new account imported.

hairtail:
> > I would opt for adding this to [account import](https://github.com/iron-fish/ironfish/issues/2173) and not rescan. More user-friendly imo.
> 
> Most people use this for reason that they want to start a rescan and they actually know when this account was created, instead of a new account imported.

Clearly, they know when the first transaction happened to this account instead of account creation time.

wd021:
I don't think most users remember the height at which their account was created. By embedding this metadata into the account import/export feature, it removes the need for users to 'know this', it can just be a boolean flag.

hairtail:
> I don't think most users remember the height at which their account was created. By embedding this metadata into the account import/export feature, it removes the need for users to 'know this', it can just be a boolean flag.

...emm, I have already pointed out that "the height at which their account was created" is not important. But "the height at which the first transaction sent to this account" is important especially for miners in mining pool.

wd021:
i'm aware of how a blockchain works. I'm thinking about features with the lense of mainstream usability, not fixating on the 100 power users who can tear apart a node. Most people will end up creating an account with the default wallet and it doesn't inhibit those who don't.

hughy:
In principle I like the idea of limiting the scan, but I think that it would be better to combine the approach with some account property like "restore height" (https://github.com/iron-fish/ironfish/issues/2173) or "scanned from".

I think that without that there's a risk that a user could rescan from the wrong place and leave their wallet in an incorrect state that would be difficult to debug.

dguenther:
I agree that something like `restoreHeight` would be nice -- generally, right now people will be generating their accounts through our wallet. And I also agree that this is a power-user-only feature, and I'd rather not have users filing issues because they started rescanning too recently.

I think the underlying issue here is that rescanning takes too long, which we might also be able to improve, but I think this at least helps out power users until we build a `restoreHeight` feature, and a good compromise could be making the flag hidden to see how often it's useful, so I've made that change.

hughy (approved):

NullSoldier:
@hairtail What graffiti do you want this credited to?

hairtail:
> @hairtail What graffiti do you want this credited to?

https://testnet.ironfish.network/users/14586

PR #2455, opened 2022-10-31
fix: delete deposit commands after phase 2 end

jowparks:
## Summary
Deletes deposit commands, this is to be merged after end of phase 2.

## Breaking Change

```
[ ] Yes
```


lwisne (approved):

NullSoldier (commented):
Will you delete the surrounding code too that's no longer used? min max api, `verifyCanSend`?

PR #2456, opened 2022-10-31
Cache pastRoot function in merkle tree

danield9tqh:
## Summary
When validating transactions, the validator loops through every spend in the transaction and compares the spend’s commitment to the note tree’s hash at the spend’s tree size. To do this, it recalculates the root hash of the note tree at the spend’s tree size. Since our transaction creation uses block-aligned tree sizes for spend commitments, there should be a clustering of common tree sizes in spends. Adding a cache here should greatly reduce the amount of rehashing needed in the merkle tree when adding a block.

## Testing Plan
Local testing and unit tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -644,8 +644,8 @@ export class Blockchain {
     Assert.isNotNull(prev)
 
     await this.saveConnect(block, prev, tx)
-    this.notes.pastRootTxCommited(tx)
     await tx.update()
+    this.notes.pastRootTxCommited(tx)
 
     this.head = block.header
     await this.onConnectBlock.emitAsync(block, tx)
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -644,6 +644,7 @@ export class Blockchain {
     Assert.isNotNull(prev)
 
     await this.saveConnect(block, prev, tx)
+    this.notes.pastRootTxCommited(tx)
     await tx.update()
 
     this.head = block.header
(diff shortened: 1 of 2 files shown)


ironfish/src/merkletree/merkletree.ts
```diff
@@ -469,6 +478,53 @@ export class MerkleTree<
...
+      const local = this.localPastRootCache.get(tx.id)
+      const localResult = local && local.get(pastSize)
+
+      return localResult || this.globalPastRootCache.get(pastSize)
+    }
+
+    return this.globalPastRootCache.get(pastSize)
+  }
+
+  pastRootTxCommited(tx: IDatabaseTransaction): void {
```
dguenther, line 518: Should be `Committed`

ironfish/src/merkletree/merkletree.ts
```diff
@@ -42,6 +44,12 @@ export class MerkleTree<
   readonly leavesIndex: IDatabaseStore<LeavesIndexSchema<H>>
   readonly nodes: IDatabaseStore<NodesSchema<H>>
 
+  private readonly globalPastRootCache: LRU<number, H> = new LRU<number, H>(300 * 60)
+  private readonly localPastRootCache: LRU<number, Map<number, H>> = new LRU<
```
dguenther, line 48: Maybe `transactionPastRootCache`?

ironfish/src/blockchain/blockchain.ts
```diff
@@ -1273,6 +1273,7 @@ export class Blockchain {
     }
 
     await tx.update()
+    this.notes.pastRootTxCommited(tx)
```
dguenther, line 1276: Does this need to go in `reconnect` as well?

dguenther (approved):

PR #2463, opened 2022-10-31
Remove transaction count consensus

mirayadav:
## Summary

Now that we have block size consensus (max block size 2 MB), we can remove transaction count consensus (max transaction count 300).

## Breaking Change

```
[ ] Yes
```


dguenther:
Conflicts on this now, looks like

dguenther (approved):

PR #2469, opened 2022-10-31
Update deposit and deposit-all with new fee estimation

ygao76:
## Summary
Use new fee estimation for 
- deposit
- deposit-all
- fees
Use medium priority level as the fee estimation.
## Testing Plan
local
```
yarn start deposit
yarn run v1.22.19
$ yarn build && yarn start:js deposit
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc --inspect=:0 --inspect-publish-uid=http --enable-source-maps bin/run deposit
Insufficient balance: 0.00000000 IRON.  Fee (0.00000365 IRON) + minimum deposit (0.1 IRON) = total required (0.10000365 IRON)
```
```
yarn start deposit -f 0.1
yarn run v1.22.19
$ yarn build && yarn start:js deposit -f 0.1
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc --inspect=:0 --inspect-publish-uid=http --enable-source-maps bin/run deposit -f 0.1
Error reading the fee value, please enter a valid number.
```

```
yarn start fees
yarn run v1.22.19
$ yarn build && yarn start:js fees
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc --inspect=:0 --inspect-publish-uid=http --enable-source-maps bin/run fees
Fee distribution for last 10 block
percentile 10: 654 ORE
percentile 20: 654 ORE
percentile 30: 654 ORE
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/deposit-all.ts
+++ b/ironfish-cli/src/commands/deposit-all.ts
@@ -99,16 +99,21 @@ export default class DepositAll extends IronfishCommand {
     if (flags.fee) {
       const [parsedFee] = BigIntUtils.tryParse(flags.fee)
 
-      if (parsedFee != null) {
-        fee = parsedFee
+      if (parsedFee == null) {
+        this.log(`Error reading the fee value, please enter a valid number.`)
+        this.exit(0)
+      } else if (parsedFee <= 0n) {
+        this.log(`The minimum fee is ${CurrencyUtils.renderIron(1n, true)}`)
+        this.exit(0)
       }
+
+      fee = parsedFee
     }
 
     const { canSend, errorReason } = await verifyCanSend(
       this.client,
       this.api,
       expirationSequenceDelta,
-      fee || 1n,
       graffiti,
     )
     if (!canSend) {
@@ -197,7 +202,7 @@ export default class DepositAll extends IronfishCommand {
       unconfirmedBalance = CurrencyUtils.decode(balanceResp.content.unconfirmed)
       pendingBalance = CurrencyUtils.decode(balanceResp.content.pending)
 
-      if (fee == null) {
+      if (flags.fee === undefined || fee == null) {
         try {
           const response = await this.client.estimateFee({
             fromAccountName: accountName,
--- a/ironfish-cli/src/commands/deposit-all.ts
+++ b/ironfish-cli/src/commands/deposit-all.ts
@@ -60,26 +60,6 @@ export default class DepositAll extends IronfishCommand {
     this.client = await this.sdk.connectRpc(false, true)
     this.api = new WebApi()
 
-    let fee = null
-
-    if (flags.fee) {
-      const [parsedFee] = BigIntUtils.tryParse(flags.fee)
-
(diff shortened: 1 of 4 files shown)


ironfish-cli/src/commands/fees.ts
```diff
@@ -13,32 +14,28 @@ export class FeeCommand extends IronfishCommand {
...
       this.log(
-        `Fee distribution for last ${numOfBlocks} block${numOfBlocks > 1 ? 's' : ''}: (${
-          response.content.startBlock
-        } - ${response.content.endBlock})\np25: ${response.content.p25} ORE\np50: ${
-          response.content.p50
-        } ORE\np75: ${response.content.p75} ORE`,
+        `Fee distribution for last ${JSON.stringify(numOfBlocks)} block\n` +
+          `percentile ${JSON.stringify(low)}: ${response.content.low || ''} ORE\n` +
+          `percentile ${JSON.stringify(medium)}: ${response.content.medium || ''} ORE\n` +
+          `percentile ${JSON.stringify(high)}: ${response.content.high || ''} ORE\n`,
```
hughy, line 38: nit - the units for the fee rates will be $ORE/kB

ironfish-cli/src/commands/deposit-all.ts
```diff
@@ -208,6 +196,25 @@ export default class DepositAll extends IronfishCommand {
...
+                amount: CurrencyUtils.encode(confirmedBalance),
+                memo: graffiti,
+              },
+            ],
+          })
+          fee = CurrencyUtils.decode(response.content.medium)
+        } catch {
+          fee = 1n
+        }
+      }
```
hughy, line 216: We could move this out of the while loop -- `fee` will always be non-null after the first iteration.

Or we could update to get a new estimate on each loop

ironfish-cli/src/commands/deposit-all.ts
```diff
@@ -114,11 +94,21 @@ export default class DepositAll extends IronfishCommand {
...
+        fee = parsedFee
+      }
+    }
+
     const { canSend, errorReason } = await verifyCanSend(
       this.client,
       this.api,
       expirationSequenceDelta,
-      fee,
+      fee || 1n,
```
hughy, line 111: I think we could remove the `fee` argument from `verifyCanSend` -- checking that the fee is non-negative there seems out-of-place

ygao76: removed fee arugment from verifyCanSend.
Add validation for parsed fee. Check if its valid and > 0

ironfish-cli/src/commands/deposit-all.ts
ygao76: check `flags.fee` instead. If user did not pass in fee value. We will get a new estimate fee on each loop.

PR #2470, opened 2022-10-31
refactor(rust): Minor cleanup and tests for Asset

mat-if:
## Summary

- Introduce str_to_array helper for converting strings to fixed size arrays
- De-duplicate group_hash function
- Add some tests for Asset

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/util.rs
+++ b/ironfish-rust/src/util.rs
@@ -1,0 +1,35 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+use std::cmp;
+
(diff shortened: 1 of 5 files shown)


ironfish-rust/src/util.rs
```diff
@@ -0,0 +1,35 @@
...
+    use super::str_to_array;
+
+    #[test]
+    fn test_str_to_array() {
+        let string_fits = "asdf";
+        let arr: [u8; 8] = str_to_array(string_fits);
+
+        assert_eq!(arr, [97, 115, 100, 102, 0, 0, 0, 0]);
+
+        let string_too_big = "asdfasdf";
```
rohanjadvani, line 30: I'm not sure what the Rust standard is, but would it make sense to have a second test (ex. `test_str_too_long_to_array`)?

rohanjadvani (approved):

rohanjadvani (approved):

PR #2485, opened 2022-11-01
Deprecate get fees api

ygao76:
## Breaking Change

```
[ ] Yes
```
@hughy


ironfish/src/rpc/routes/fees/getFees.test.ts
```diff
@@ -1,55 +0,0 @@
...
-  it('should fail if no block is found', async () => {
-    await expect(
-      routeTest.client.request('fees/getFees', { numOfBlocks: 2 }).waitForEnd(),
-    ).rejects.toThrow('numOfBlocks must be less than the current head sequence')
-  })
-
-  it('ignores miners fee from calculations', async () => {
-    const node = routeTest.node
-
-    const block = await useMinerBlockFixture(node.chain)
```
NullSoldier, line 21: we also need to delete the fixture file too

ygao76: removed

PR #2491, opened 2022-11-01
Parallelize downloading and adding blocks

danield9tqh:
## Summary
Currently we sequentially download a group of blocks (20) and then add them to the chain. If we parallelize this we can start adding blocks while we are downloading the next group of blocks from peers. After some improvements to addBlock such as caching pastRoot and batch adding notes to the merkle tree addBlock time has improved a lot. Now the limitation seems to be more around downloading blocks. We're seeing download speeds anywhere from 0.5 - 20 blocks per second depending on the peer and connection. We are seeing add block times of ~1-2 blocks/s. If we parallelize these we could potentially double syncing speed. 

## Testing Plan
Local testing and unit tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/peerNetwork.ts
+++ b/ironfish/src/network/peerNetwork.ts
@@ -600,7 +600,9 @@ export class PeerNetwork {
     })
   }
 
-  async getBlockHashes(peer: Peer, start: number, limit: number): Promise<Buffer[]> {
+  async getBlockHashes(peer: Peer, start: number, limit: number): Promise<[Buffer[], number]> {
+    const begin = BenchUtils.start()
+
     const message = new GetBlockHashesRequest(start, limit)
     const response = await this.requestFrom(peer, message)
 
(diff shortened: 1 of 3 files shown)


ironfish/src/network/peerNetwork.ts
```diff
@@ -600,30 +600,46 @@ export class PeerNetwork {
...
-    return response.message.hashes
+    return [response.message.hashes, BenchUtils.end(begin)]
   }
 
-  async getBlocks(peer: Peer, start: Buffer, limit: number): Promise<SerializedBlock[]> {
+  async getBlocks(
+    peer: Peer,
+    start: Buffer,
+    limit: number,
+  ): Promise<[SerializedBlock[], number] | Error> {
```
dguenther, line 627: I think the `[SerializedBlock[], number]` would be a little clearer if written as an object instead of an array.

ironfish/src/syncer.ts
```diff
@@ -358,23 +359,34 @@ export class Syncer {
...
-        head,
-        this.blocksPerMessage + 1,
-      )
-
-      const elapsedSeconds = BenchUtils.end(start) / 1000
-      this.downloadSpeed.add((blocks.length + 1) / elapsedSeconds)
+      const blocksResponse = await blocksPromise
+      if (blocksResponse instanceof Error) {
+        throw blocksResponse
+      }
```
dguenther, line 365: I'm not totally sure how this works -- it'd be nice if we could find a different way to solve the issue of logging an error when closing the node, or if not, I think we should add a comment explaining this.

ironfish/src/network/peerNetwork.ts
danield9tqh: yea agreed, changed

dguenther (approved):

Issue #995, opened 2022-02-13
No progress bar from yarn test:coverage:html run in \ironfish


AmberKiso:
### What happened?

On a Windows command prompt.

If I run `yarn test:coverage:html` from a sub-directory of \ironfish, like \ironfish-cli or \src it will display a nice display of the .test.ts files queued and being executed.

However, running directly from \ironfish does not display anything. It does complete though.

Not sure if that's intentional, but the readme.md in \ironfish lists the command without restriction.

### Version

0.1.21

### Relevant log output

```shell
\GitHub\ironfish>yarn test:coverage:html
yarn run v1.22.17
$ lerna run test:coverage:html
lerna notice cli v4.0.0
lerna info versioning independent
lerna info Executing command in 2 packages: "yarn run test:coverage:html"

<<Nothing displayed after this until command finishes and command prompt available>>

lerna ERR! yarn run test:coverage:html exited 1 in 'ironfish'
lerna ERR! yarn run test:coverage:html stdout:
$ tsc -b tsconfig.test.json && jest  --coverage --coverage-reporters html --testPathIgnorePatterns
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
lerna ERR! yarn run test:coverage:html stderr:
PASS ironfish src/primitives/transaction.test.slow.ts (33.461 s)
PASS ironfish src/strategy.test.slow.ts (42.501 s)
FAIL ironfish src/mining/director.test.slow.ts (44.923 s)
  ● Mining director › After mining › discards block if chain changed

    Expected timestamp is in future to be undefined

      10 |   static isUndefined<T>(x: undefined | T, message?: string): asserts x is undefined {
      11 |     if (x !== undefined) {
    > 12 |       throw new Error(message || `Expected ${String(x)} to be undefined`)
         |             ^
      13 |     }
      14 |   }
      15 |

      at Function.isUndefined (src/assert.ts:12:13)
      at Object.makeBlockAfter (src/testUtilities/helpers/blockchain.ts:49:10)
      at Object.<anonymous> (src/mining/director.test.slow.ts:255:23)

<<Long output, saved if you want me to post later>>
```

lwisne:
This persists - if someone wants to make it prettier it's probably a simple config change.

aminazeb:
I am interested to work on it.

danield9tqh:
Hey @AmberKiso. Thanks for the report. Closing this after adding a fix

PR #2492, opened 2022-11-02
stream tests results when running from root directory

danield9tqh:
## Summary
Stream tests results when running from root directory

## Breaking Change

```
[ ] Yes
```


danield9tqh:
Fixing https://github.com/iron-fish/ironfish/issues/995

NullSoldier (approved):

PR #2498, opened 2022-11-02
Remove CurrencyUtils.isValid*()

NullSoldier:
## Summary

These have nothing to do with our currency. They have everything to do with user input validation that is more like isValidFee(). Let's keep our currency conversion code separate from our user input. If any input should be done in CurrencyUtils it's to ensure the values are within valid ranges.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/transactions/sendTransaction.ts
+++ b/ironfish/src/rpc/routes/transactions/sendTransaction.ts
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import * as yup from 'yup'
-import { ArrayUtils, CurrencyUtils } from '../../../utils'
+import { CurrencyUtils } from '../../../utils'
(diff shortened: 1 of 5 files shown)


ironfish/src/rpc/routes/transactions/sendTransaction.ts
```diff
@@ -92,53 +92,38 @@ router.register<typeof SendTransactionRequestSchema, SendTransactionResponse>(
...
-    }
-    let sum = BigInt(transaction.fee)
-    transaction.receives.map((receive) => {
-      if (!CurrencyUtils.isValidOre(receive.amount)) {
-        throw new ValidationError(
-          `Invalid transaction amount, ${receive.amount}`,
-          undefined,
-          ERROR_CODES.VALIDATION,
-        )
+    const receives = transaction.receives.map((receive) => {
```
hughy, line 95: We iterate over receives three separate times; here, calculating the sum, and checking that the amounts are all positive. Could we consolidate into a single loop?

hughy (approved):

PR #2499, opened 2022-11-02
Corrected deposit all check amount logic

web3Liuwentong:
## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/deposit-all.ts
```diff
@@ -227,13 +227,13 @@ export default class DepositAll extends IronfishCommand {
       const sendableOre = confirmedBalance - fee
 
       // terminate condition
-      if (terminate && pendingBalance < sendableOre + fee) {
```
web3Liuwentong, line 230: sendableOre + fee = confirmedBalance due to the upper logic, if pendingBalance < confirmedBalance didn't make sense.
if pendingBalance < fee then destory, it's correct.

PR #2501, opened 2022-11-02
Corrected deposit all check amount logic

web3Liuwentong:
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/deposit-all.ts
+++ b/ironfish-cli/src/commands/deposit-all.ts
@@ -227,13 +227,13 @@ export default class DepositAll extends IronfishCommand {
       const sendableOre = confirmedBalance - fee
 
       // terminate condition
-      if (terminate && pendingBalance < sendableOre + fee) {
+      if (terminate || pendingBalance < fee) {
         screen.destroy()
         process.exit(0)
       }
 
       // send transaction
-      if (confirmedBalance >= sendableOre + fee) {
+      if (sendableOre >= minDepositOre) {
         try {
           const oreToSend = BigIntUtils.min(
             (sendableOre / minDepositOre) * minDepositOre,


ironfish-cli/src/commands/deposit-all.ts
```diff
@@ -227,13 +227,13 @@ export default class DepositAll extends IronfishCommand {
       const sendableOre = confirmedBalance - fee
 
       // terminate condition
-      if (terminate && pendingBalance < sendableOre + fee) {
```
web3Liuwentong, line 230: sendableOre + fee = confirmedBalance due to the upper logic, if pendingBalance < confirmedBalance didn't make sense.
so i change to if pendingBalance < fee then destory

ironfish-cli/src/commands/deposit-all.ts
```diff
@@ -227,13 +227,13 @@ export default class DepositAll extends IronfishCommand {
...
+      if (terminate || pendingBalance < fee) {
         screen.destroy()
         process.exit(0)
       }
 
       // send transaction
-      if (confirmedBalance >= sendableOre + fee) {
```
web3Liuwentong, line 236: confirmedBalance = sendableOre + fee is always true due to line 227.

ironfish-cli/src/commands/deposit-all.ts
```diff
@@ -227,13 +227,13 @@ export default class DepositAll extends IronfishCommand {
...
+      if (sendableOre >= minDepositOre) {
```
hughy, line 236: This change has already been applied in #2502

ironfish-cli/src/commands/deposit-all.ts
```diff
@@ -227,13 +227,13 @@ export default class DepositAll extends IronfishCommand {
       const sendableOre = confirmedBalance - fee
 
       // terminate condition
-      if (terminate && pendingBalance < sendableOre + fee) {
+      if (terminate || pendingBalance < fee) {
```
hughy, line 230: The condition here isn't quite right.

It needs to be `&&`: `terminate` is a command-line flag, so if it is set to true then this change would result in termination after the first deposit.

The check here doesn't account for `minDepositOre`: if the pending balance is less than `minDepositOre + fee` then no more deposits can be made.

NullSoldier:
This code has been deleted.

Issue #2493, opened 2022-11-02
Invalid Transaction Amount


airj1012:
### What happened?

Was running deposit-all.

Had 3 deposits confirmed. Then received the following error.

"Invalid transaction amount"

However the amount should be controlled by the node. I didn't add any extra parameters and 3 deposits had already been completed
![Invalid Transaction](https://user-images.githubusercontent.com/64874325/199374327-83d8785c-e84e-4962-aed6-7a749ec54120.png)

### Version

0.1.51

### Graffiti

airj1012

danield9tqh:
Thanks for the report @airj1012 we've added a fix and it will be out in the next release. In the meantime I don't think there's much of a workaround for the error. But the deposit-all command should still work as intended even though it is getting an error, all IRON should be deposited. Let us know if that's not the case

airj1012:
Bug bounty worthy?

PR #2502, opened 2022-11-02
Deposit all negative transaction amount

danield9tqh:
## Summary
It looks like we were unintentionally attempting deposits with a negative amount. Changing our check to make sure we have enough balance to send a deposit transaction including the fee

## Breaking Change

```
[ ] Yes
```


jowparks:
Thanks for the fix @danield9tqh , looks correct to me. Related to:
https://github.com/iron-fish/ironfish/issues/2493

jowparks (approved):
add >=, then good to go.

PR #2504, opened 2022-11-02
iterates over pending transactions for rebroadcast

hughy:
## Summary

limits the wallet to iterate only over pending transactions in rebroadcastTransactions.

- implements loadPendingTransactions to yield transactions with expiration sequence equal to 0 or greater than the given head sequence
- removes check on expired transactions in rebroadcastTransactions
- updates tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/walletdb/walletdb.ts
+++ b/ironfish/src/wallet/walletdb/walletdb.ts
@@ -553,6 +553,48 @@ export class WalletDB {
     }
   }
 
+  async *loadPendingTransactions(
+    account: Account,
+    headSequence: number,
+    tx?: IDatabaseTransaction,
+  ): AsyncGenerator<TransactionValue> {
+    const encoding = new PrefixEncoding(
+      BUFFER_ENCODING,
+      U32_ENCODING,
+      account.prefix.byteLength,
+    )
+
+    const noExpirationRange = StorageUtils.getPrefixesKeyRange(
+      encoding.serialize([account.prefix, 0]),
+      encoding.serialize([account.prefix, 0]),
+    )
+
+    for await (const [, [, transactionHash]] of this.pendingTransactionHashes.getAllKeysIter(
+      tx,
+      noExpirationRange,
+    )) {
+      const transaction = await this.loadTransaction(account, transactionHash, tx)
+      Assert.isNotUndefined(transaction)
+
+      yield transaction
+    }
+
+    const pendingRange = StorageUtils.getPrefixesKeyRange(
+      encoding.serialize([account.prefix, headSequence + 1]),
+      encoding.serialize([account.prefix, 2 ^ 32]),
+    )
+
+    for await (const [, [, transactionHash]] of this.pendingTransactionHashes.getAllKeysIter(
+      tx,
+      pendingRange,
+    )) {
+      const transaction = await this.loadTransaction(account, transactionHash, tx)
+      Assert.isNotUndefined(transaction)
+
+      yield transaction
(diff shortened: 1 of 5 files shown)


ironfish/src/wallet/walletdb/walletdb.ts
```diff
@@ -553,6 +553,48 @@ export class WalletDB {
     }
   }
 
+  async *loadPendingTransactions(
+    account: Account,
+    headSequence: number,
+    tx?: IDatabaseTransaction,
+  ): AsyncGenerator<TransactionValue> {
+    const encoding = new PrefixEncoding(
```
danield9tqh, line 561: I'm curious why we couldn't do `const encoding = this.pendingTransactionHashes.keyEncoding` here. I see that it's not exactly the same. One is of type `[Buffer, [Buffer, number]]` and the other is `[Buffer, number]`. What's with the extra buffer encoding when we initialize `pendingTransactionHashes` in the constructor?

danield9tqh: I think we could also consider reusing this somehow though now that we are using this encoding in the constructor, `loadPendingTransactions` and `loadExpiredTransactions`

ironfish/src/wallet/walletdb/walletdb.ts
```diff
@@ -553,6 +553,48 @@ export class WalletDB {
...
+      BUFFER_ENCODING,
+      U32_ENCODING,
+      account.prefix.byteLength,
+    )
+
+    const noExpirationRange = StorageUtils.getPrefixesKeyRange(
```
danield9tqh, line 567: This is just supposed to be a single prefix right, not a start and end? We could use [`StorageUtils.getPrefixeKeyRange`](https://github.com/iron-fish/ironfish/blob/bc85577f559d4132857ba26328342e5e14c6b1c4/ironfish/src/storage/database/utils.ts#L16)?

ironfish/src/wallet/walletdb/walletdb.ts
danield9tqh: Ok I see, looks like we store the transaction hash as part of the key and then store nothing in the value. Curious if there's a reason we do that? Not that we can change it

danield9tqh: We could potentially reuse `this.pendingTransactionHashes.keyEncoding` and then just add an empty buffer for the transaction hash range?

hughy: I'll give that a try and see if we can avoid defining a separate prefix

ironfish/src/wallet/walletdb/walletdb.ts
hughy: Yes! This should be just a single prefix

danield9tqh (approved):

ironfish/src/wallet/walletdb/walletdb.ts
hughy: I was able to get this to work by specifying the full range of buffer values with `getPrefixesKeyRange`:

```
const encoding = this.pendingTransactionHashes.keyEncoding

const pendingRange = StorageUtils.getPrefixesKeyRange(
      encoding.serialize([account.prefix, [headSequence + 1, Buffer.alloc(4)]]),
      encoding.serialize([account.prefix, [2 ^ 32, Buffer.alloc(4, 255)]]),
)
```

PR #2506, opened 2022-11-03
Remove support for old gossip messages

danield9tqh:
## Summary
As part of the [gossip project](https://coda.io/d/_dMnayiE39lL/Gossip-Project_sukbV) we removed the concept of a GossipMessage. This task is to clean up reference to these unused old Network messages.

## Testing Plan
Unit tests and local testing

## Breaking Change

```
[ ] Yes
```


ironfish/src/network/peerNetwork.ts
```diff
@@ -632,9 +631,7 @@ export class PeerNetwork {
   ): Promise<void> {
     const { message } = incomingMessage
 
-    if (message instanceof GossipNetworkMessage) {
-      await this.handleGossipMessage(peer, message)
```
dguenther, line 636: If we're no longer going to parse Gossip messages, we should delete them and refactor the code to remove references, though it'll require some changes to `onNewTransaction`+more.

ironfish/src/network/peerNetwork.ts
```diff
@@ -33,7 +33,6 @@ import {
   GetBlockTransactionsResponse,
 } from './messages/getBlockTransactions'
 import { GetCompactBlockRequest, GetCompactBlockResponse } from './messages/getCompactBlock'
-import { GossipNetworkMessage } from './messages/gossipNetworkMessage'
```
dguenther, line 36: Not related to this code change, but we should bump the min protocol version to 18 to indicate we're not supporting the older messages any longer (even though there aren't many people left on versions earlier than v0.1.45)

ironfish/src/network/messageRegistry.ts
```diff
@@ -97,22 +88,6 @@ const parseRpcNetworkMessage = (
...
-
-  switch (type) {
-    case NetworkMessageType.NewBlock:
-      return NewBlockMessage.deserialize(body, nonce)
-    case NetworkMessageType.NewTransaction:
-      return NewTransactionMessage.deserialize(body, nonce)
-    default:
-      throw new Error(`Unknown gossip network message type: ${type}`)
-  }
-}
```
dguenther, line 114: It's not necessarily related to this PR so I'd be okay having it elsewhere, but it did make me think it'd be nice to have a test that sends a buffer with a nonexistent message from a Peer to ensure it doesn't crash somewhere in the PeerNetwork code -- it should just disconnect the Peer.

dguenther (changes requested):

dguenther:
Based on https://stats.ironfish.network/?orgId=1, we have around 120 nodes on version 0.1.41, which will be incompatible, but otherwise I'm not seeing many on 0.1.42-0.1.44. Given we've logged 6500 nodes online, I don't think this will impact a substantial portion of our current users.

ironfish/src/network/peerNetwork.ts
danield9tqh: Done 👍🏻

ironfish/src/network/peerNetwork.ts
danield9tqh: Yea I can make that change in this PR. I was thinking of doing it in a follow up just so the PRs are easier to review. I think they can be logically separated between receiving old messages and sending old messages wdyt?

ironfish/src/network/messageRegistry.ts
danield9tqh: I looked into this a bit and it would require a lot of mocking. We'd have to mock the NodeDataChannel in WebRTCConnection. We could just add a test in `WebRTCConnection` but that wouldn't really test that the node doesn't crash. I can add a task for this though

dguenther: Ah I see, I found the try/catches that handle invalid messages in the WebRTCConnection and WebSocketConnection. It seems clear it won't crash from invalid messages, so I don't think we imminently need a test for this.

If someone ends up writing a test, I'd imagine creating a PeerNetwork, creating a Peer with a mocked websocket connection (we already mock `ws` in the PeerNetwork tests), then calling the `socket.onmessage` function on the mocked websocket connection directly and make sure an exception doesn't happen anywhere.

ironfish/src/network/peerNetwork.ts
dguenther: Sounds good, as long as the rest of the work is in progress (and it sounds like it's almost done) I'm fine with that

dguenther (approved):

PR #2509, opened 2022-11-04
Update account rpc endpoint

wd021:
## Summary
With wallet 2.0 we updated the folder structure from src `/accounts` to `/wallet`. To keep naming conventions in sync, updating the rpc endpoint as well. 

## Testing Plan
- yarn test

## Breaking Change

```
[ ] Yes
```


NullSoldier:
Rebase please, and we can get this in. I think we're also considering changing the CLI to wallet as well, im just not sure yet. It would require a major update to our website too.

wd021:
@NullSoldier done. weird, file diff wasn't showing properly after rebase, had to reset base branch to sync to latest staging.

i think wallet does make more sense. for the website, you mean updating docs and everything? can help do that if you need someone.

NullSoldier (approved):

PR #2510, opened 2022-11-04
Remove support for sending old gossip messages

danield9tqh:
## Summary
Continuation of the cleanup of old gossip message from https://github.com/iron-fish/ironfish/pull/2506

## Testing Plan
Local testing and unit tests

## Breaking Change

```
[ ] Yes
```


ironfish/src/network/peerNetwork.ts
```diff
@@ -1238,13 +1188,11 @@ export class PeerNetwork {
     return new GetCompactBlockResponse(block.toCompactBlock(), message.rpcId)
   }
 
-  private async handleNewBlockMessage(peer: Peer, message: NewBlockMessage) {
+  private async handleNewBlockMessage(peer: Peer, block: Block) {
```
dguenther, line 1191: Maybe `handleRequestedBlock` now, since it's only used in the case where a block is directly requested by the block fetcher

dguenther (commented):
This looks good, I just have a rename suggestion then I'm good with approving 👍

ironfish/src/network/peerNetwork.ts
danield9tqh: yep 👍🏻

dguenther (approved):

PR #2513, opened 2022-11-05
iterate over pending transactions for rebroadcast

hughy:
## Summary

limits the wallet to iterate only over pending transactions in rebroadcastTransactions.

implements loadPendingTransactions to yield transactions with expiration sequence equal to 0 or greater than the given head sequence
    
- removes check on expired transactions in rebroadcastTransactions
- updates tests

NOTE: repeat of #2504, but targeted to staging

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/__fixtures__/account.test.ts.fixture
+++ b/ironfish/src/wallet/__fixtures__/account.test.ts.fixture
@@ -148,5 +148,137 @@
         }
       ]
     }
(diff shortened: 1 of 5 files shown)


lwisne (approved):
I am assuming daniel's review of https://github.com/iron-fish/ironfish/pull/2504 was good :)

PR #2514, opened 2022-11-05
reuses transaction size for different fee rates

hughy:
## Summary

estimates fees for all fee rate priority level with only one call to 'createSpends' to avoid iterating over notes multiple times.

changes 'getPendingTransactionSize' to return both the size and the spender change for a pending transaction without accounting for the fee.

implements 'getPendingTransactionFee' to use the size and spender change to determine the fee based on how big the transaction is and whether more notes or spends are needed to cover the fee and spender change.

implements 'estimateFees' to return estimates for all fee rate priority levels and updates estimateFee RPC endpoint.

accounts for additional note for spender change in transaction sizes.

## Testing Plan

Updated unit tests and ran benchmark script from #2436 . It's only *very slightly* faster than before on my node, which has very few notes.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/memPool/feeEstimator.ts
+++ b/ironfish/src/memPool/feeEstimator.ts
@@ -9,6 +9,7 @@ import { getTransactionSize } from '../network/utils/serializers'
 import { Block, Transaction } from '../primitives'
 import { NOTE_ENCRYPTED_SERIALIZED_SIZE_IN_BYTE } from '../primitives/noteEncrypted'
 import { SPEND_SERIALIZED_SIZE_IN_BYTE } from '../primitives/spend'
+import { BigIntUtils } from '../utils'
 import { Wallet } from '../wallet'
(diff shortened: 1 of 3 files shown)


ironfish/src/memPool/feeEstimator.ts
```diff
@@ -194,22 +207,34 @@ export class FeeEstimator {
...
-        size += additionalSpendsLength
-      }
-    }
+    const spenderChange = amount - amountNeeded
 
-    return size
+    return { size, spenderChange }
+  }
+
+  private getPendingTransactionFee(
```
hughy, line 215: The logic here will be useful in calculating a fee from fee rate in `Wallet.createTransaction`

hughy:
Closing this PR since we plan to make changes that move fee calculation from fee rate to `createTransaction`

PR #2517, opened 2022-11-05
avoids logging unparsed buffers in error message

hughy:
## Summary

nodes sending now-unsupported message types produce parsing errors. the parsing error message includes a stringified buffer:

```
error parsing messageK9j�KG4x,ç7�k������n����b�çi��� . . .
```

removes the event data from the error message to avoid logging the unparsed buffer.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/peers/connections/webSocketConnection.ts
+++ b/ironfish/src/network/peers/connections/webSocketConnection.ts
@@ -99,8 +99,8 @@ export class WebSocketConnection extends Connection {
         // TODO: any socket that sends invalid messages should probably
         // be punished with some kind of "downgrade" event. This should
         // probably happen at a higher layer of abstraction
-        const message = 'error parsing message'
-        this.logger.warn(`${message} ${event.data.toString()}`)
+        const message = 'Error parsing message'
+        this.logger.warn(message)
         this.close(new NetworkError(message))
         return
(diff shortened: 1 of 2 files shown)


ironfish/src/network/peers/connections/webSocketConnection.ts
```diff
@@ -99,8 +99,8 @@ export class WebSocketConnection extends Connection {
         // TODO: any socket that sends invalid messages should probably
         // be punished with some kind of "downgrade" event. This should
         // probably happen at a higher layer of abstraction
-        const message = 'error parsing message'
-        this.logger.warn(`${message} ${event.data.toString()}`)
+        const message = 'Error parsing message'
```
dguenther, line 102: Is this an issue on the WebRTCConnection as well?

hughy: I think we're ok on the WebRTCConnection because the incoming message will be `string | Uint8Array`, but we could make this change there too for consistency?

dguenther: Yep, I think it'd be worth making the same change for consistency at least 👍

dguenther (approved):

Issue #2516, opened 2022-11-05
Node crashes with `Error: nullifierToNote mappings must have a corresponding decryptedNote`


neuronull:
### What happened?

I upgraded to 0.1.51 a few days ago.

Before I upgraded, my node Accounts was not fully synced and I was informed that should be fine.
After I upgraded, the syncing went very quickly, which was great!

Then I issued the `deposit-all` command and that seemed to go ok except now my node is crashing.

This has happened twice so far now.

I installed and upgraded the node on Windows Subsystem for Linux using npm.

```
$ ironfish version
name       ironfish
version    0.1.51
git        67da183
runtime    node/16.17.1
```

### Version

v0.1.51

### Debug output

```shell
$ ironfish debug
Database in use, skipping output that requires database.
Stop the node and run the debug command again to show full output.

Iron Fish version               0.1.51 @ 67da183
Iron Fish library               0.0.28 @ 67da183
Operating system                Linux x64
CPU model(s)                    Intel(R) Core(TM) i5-4690 CPU @ 3.50GHz
CPU threads                     4
RAM total                       12.44 GiB
Heap total                      2.01 GiB
Node version                    v16.17.1
ironfish in PATH                true
Garbage Collector Exposed       false
Telemetry enabled               true
Node name                       null
Block graffiti                  null
```

### Relevant log output

```shell
[00:57:23.613] [info] WebSocket server started at :::9033
[00:57:23.954] [info] Connected to the Iron Fish network
[00:57:54.144] [info] Starting sync from Rob3nxV (ddkissfmnode). work: +76431612140838651, ours: 258598, theirs: 258688
[00:57:54.145] [info] Finding ancestor using linear search on last 3 blocks starting at 00000...6396b (258598) from peer Rob3nxV (ddkissfmnode) at 258688
[00:57:54.331] [info] Found peer Rob3nxV (ddkissfmnode) ancestor 00000...6396b, syncing from 258598 -> 258688 (90) after 1 requests
[00:57:54.331] [info] Requesting 5 blocks starting at 00000...6396b (258598) from Rob3nxV (ddkissfmnode)
[00:57:58.631] [info] Added block seq: 258600, hash: 00000...3e0c9, txs: 101, progress: 99.96%, time: 593.1ms
[00:58:03.647] [info] Requesting 5 blocks starting at 00000...efd4f (258603) from Rob3nxV (ddkissfmnode)
[...]

PR #2519, opened 2022-11-05
deletes nullifier from nullifierToNote if nullifier set to null on note

hughy:
## Summary

it is curently possible for a note to be spent on a fork, but, after reorg, a nullifierToNote mapping is left in the database.

this can result in an error when expiring transactions based on the following sequence:

```
1. transaction A created, mined on a fork
2. transaction B created, spending a note from A
3. reorg from the fork containing A, nullifiers for notes in A set to null (but
not remeoved from nullifierToNote)
4. transaction A expires, removing notes from the database (still not removing
nullifiers from nullifierToNote, since the referential integrity is broken)
5. transaction B expires, but expiration fails because the nullifier from the
spend in B maps to a note that has been deleted
```

fixes the referential integrity issue by deleting nullifierToNote entry during updateDecryptedNote if the note's nullifier changes from non-null to null

resolves cause of #2516

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/__fixtures__/wallet.test.ts.fixture
+++ b/ironfish/src/wallet/__fixtures__/wallet.test.ts.fixture
@@ -1624,5 +1624,181 @@
         }
       ]
     }
+  ],
+  "Accounts should update balances for expired transactions with spends on a fork": [
+    {
+      "id": "620570ee-585c-4e2c-8fb6-86880e5a8f38",
+      "name": "a",
+      "spendingKey": "b2321c9334719dba486a2e59680f845a358c668a0c95bb2441bbac3d492b7cf8",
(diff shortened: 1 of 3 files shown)


hughy:
This may have left some wallets with transactions that they cannot successfully expire. To fix that I think our options are:

1. Implement a `repair` command to remove the nullifiers that don't map to notes that this caused
2. Handle the missing note without failing here: https://github.com/iron-fish/ironfish/blob/c3f0ab6a8a3646f2c6f5b16d6add35a0ca17b6e7/ironfish/src/wallet/account.ts#L352

dguenther (approved):
Confirmed the two new tests fail on the prior version of the code, nice fix! Looks like it needs a rebase then it's good to merge.

Issue #2520, opened 2022-11-07
Negative confirmed balance


hairtail:
### What happened?

```
Your balance is made of notes on the chain that are safe to spend
Balance: $IRON -755.98014174

3 notes worth 1133.97021261 are on the chain within 12 blocks of the head
Unconfirmed: 377.99007087

2 notes worth 2000.00000000 are waiting for miners to add them to the chain
Pending: 2377.99007087
```

### Version

0.1.51 @ src

### Graffiti

miningKing

hughy:
Resolved by #2565

PR #2521, opened 2022-11-07
Handle note status change during getBalance

hairtail:
## Summary
This PR handles note status change during `getBalance`, which should be the reason of https://github.com/iron-fish/ironfish/issues/2520 in my opinion.
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/account.ts
+++ b/ironfish/src/wallet/account.ts
@@ -400,10 +400,12 @@ export class Account {
     const pending = await this.getUnconfirmedBalance(tx)
 
     let unconfirmed = pending
+    const notOnChainNotes: DecryptedNoteValue[] = []
     for await (const note of this.walletDb.loadNotesNotOnChain(this, tx)) {
       if (!note.spent) {
         pendingCount++
         unconfirmed -= note.note.value()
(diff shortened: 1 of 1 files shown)


ironfish/src/wallet/account.ts
```diff
@@ -423,6 +425,11 @@ export class Account {
         tx,
       )) {
         if (!note.spent) {
+          if (notOnChainNotes.includes(note)) {
```
hughy, line 428: This won't fix the bug, but it might cover it up.

It shouldn't be possible for a note to be in _both_ `loadNotesNotOnChain` and `loadNotesInSequenceRange`: if it's not on chain then it doesn't have a sequence, and if it has a sequence then it is on chain.

hughy:
Thanks for looking into this!

I think that this PR correctly identifies where and how the balance is going negative, but I think we need to fix the underlying issue in how we're updating data in the database.

hairtail:
> Thanks for looking into this!
> 
> I think that this PR correctly identifies where and how the balance is going negative, but I think we need to fix the underlying issue in how we're updating data in the database.

Glad to do it.

PR #2526, opened 2022-11-08
feat(rust): add token identifier to asset

mat-if:
## Breaking Change

```
[ ] Yes
```


ironfish-rust/src/assets/asset.rs
```diff
@@ -137,7 +150,7 @@ mod test {
     use super::{Asset, NATIVE_ASSET};
 
     #[test]
-    fn test_new_with_nonce() {
+    fn test_asset_new_with_nonce() {
```
mat-if, line 153: test renames were just for simpler namespacing, so i can do `cargo test test_asset`

rohanjadvani (approved):

PR #2527, opened 2022-11-08
adds blockHash and sequence to decryptedNoteValue

hughy:
## Summary

adds chain state fields to decryptedNotes store and implements data migration to populate these fields.

~~migration does not (yet) remove sequenceToNoteHash or nonChainNoteHashes stores. subsequent changes will modify wallet code to cease using these stores and then, finally, remove them.~~ Computing the balance will be much less efficient without sequenceToNoteHash and nonChainNoteHashes, but having chain state in decryptedNotes will make it easier to keep them in sync

## Testing Plan

```
[11:05:31.374] Applying 1 migrations:
[11:05:31.377]   014-note-sequence
[11:05:31.377] 
[11:05:31.377] Running 014-note-sequence...
[11:05:31.447] Adding chain state data to decryptedNotes...
[11:05:31.452] Migrating decrypted notes for account test...
[11:05:31.453] 	Migrated 0 decrypted notes for account test
[11:05:31.454] Migrating decrypted notes for account default...
[11:05:31.465] 	Migrated 7 decrypted notes for account default
[11:05:31.532] Benchmark - Time: 86.8829ms, Heap: 325.78 KiB, RSS: 1.87 MiB, Mem: 2.19 MiB
[11:05:31.533] Successfully applied 1 migrations
✨  Done in 5.85s.
```

## Breaking Change

```
[X] Yes
```

Requires data migration.

### DIFF
--- a/ironfish/src/wallet/walletdb/decryptedNoteValue.ts
+++ b/ironfish/src/wallet/walletdb/decryptedNoteValue.ts
@@ -14,17 +14,22 @@ export interface DecryptedNoteValue {
   // These fields are populated once the note's transaction is on the main chain
   index: number | null
   nullifier: Buffer | null
+  blockHash: Buffer | null
+  sequence: number | null
 }
 
 export class DecryptedNoteValueEncoding implements IDatabaseEncoding<DecryptedNoteValue> {
   serialize(value: DecryptedNoteValue): Buffer {
-    const { accountId, nullifier, index, note, spent, transactionHash } = value
+    const { accountId, nullifier, index, note, spent, transactionHash, blockHash, sequence } =
+      value
     const bw = bufio.write(this.getSize(value))
 
     let flags = 0
     flags |= Number(!!index) << 0
     flags |= Number(!!nullifier) << 1
     flags |= Number(spent) << 2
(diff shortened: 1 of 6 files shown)


ironfish/src/wallet/walletdb/decryptedNoteValue.ts
```diff
@@ -14,17 +14,22 @@ export interface DecryptedNoteValue {
   // These fields are populated once the note's transaction is on the main chain
   index: number | null
   nullifier: Buffer | null
+  blockHash: Buffer | null
+  sequence: number | null
```
NullSoldier, line 18: should we do blockSequence, or no?

hughy: On one hand, blockSequence is clearer, but, on the other hand, TransactionValue uses sequence

Do we choose clarity or consistency?

hughy: I think I prefer `sequence` since there are only a couple of places in the code where we currently use `blockSequence`, but I don't have a strong preference!

NullSoldier (approved):

PR #2529, opened 2022-11-08
Fee command changes

NullSoldier:
## Summary
 - Dont need to cast the keys
 - If RPC has all fields as optional the whole payload can be optional
 - Dont need to try catch in a command because that happens in the base command

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/fees.ts
+++ b/ironfish-cli/src/commands/fees.ts
@@ -1,11 +1,8 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { ConfigOptions } from '@ironfish/sdk'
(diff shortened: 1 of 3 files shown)


ironfish-cli/src/commands/fees.ts
```diff
@@ -17,32 +14,19 @@ export class FeeCommand extends IronfishCommand {
...
-          `percentile ${JSON.stringify(high)}: ${response.content.high || ''} ORE/kb\n`,
-      )
-    } catch (error: unknown) {
-      if (error instanceof Error) {
-        this.error(error.message)
-      }
+    const low = config.content['feeEstimatorPercentileLow'] || '10'
+    const medium = config.content['feeEstimatorPercentileMedium'] || '20'
+    const high = config.content['feeEstimatorPercentileHigh'] || '30'
+    const numOfBlocks = config.content['feeEstimatorMaxBlockHistory']
```
NullSoldier, line 23: This is always filled in by the default config value

hughy (approved):

PR #2530, opened 2022-11-08
Fix error when peer disconnects while syncing

danield9tqh:
## Summary
We are sometimes getting an error when a peer disconnects that causes the node to crash. This is most likely related to changes related to [parallelizing addBlock and download blocks](https://github.com/iron-fish/ironfish/pull/2491)

## Testing Plan
Unit tests and tested disconnected peer locally

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/syncer.ts
+++ b/ironfish/src/syncer.ts
@@ -356,6 +356,8 @@ export class Syncer {
       this.logger.warn(
         `Error while syncing from ${peer.displayName}: ${ErrorUtils.renderError(e)}`,
       )
+
+      peer.close()
       this.stopSync(peer)
       return {
         blocks: [],
--- a/ironfish/src/syncer.ts
+++ b/ironfish/src/syncer.ts
@@ -341,21 +341,36 @@ export class Syncer {
     }
   }
 
-  async syncBlocks(peer: Peer, head: Buffer, sequence: number): Promise<void> {
-    this.abort(peer)
-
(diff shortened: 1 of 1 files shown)


ironfish/src/syncer.ts
```diff
@@ -341,21 +341,36 @@ export class Syncer {
...
-      `Requesting ${this.blocksPerMessage} blocks starting at ${HashUtils.renderHash(
-        head,
-      )} (${sequence}) from ${peer.displayName}`,
+      `Requesting ${limit - 1} blocks starting at ${HashUtils.renderHash(start)} from ${
+        peer.displayName
+      }`,
     )
 
-    let blocksPromise = this.peerNetwork.getBlocks(peer, head, this.blocksPerMessage + 1)
+    return this.peerNetwork.getBlocks(peer, start, limit).catch((e) => {
```
dguenther, line 355: Should this `.catch` do the same thing as the `.catch` in `syncFrom`? I think it calls `close` on the Peer

dguenther:
Other than the one comment, I'm okay with this 👍

ironfish/src/syncer.ts
danield9tqh: Yea usually this happens because the peer was closed but it could happen because of a timeout or another random error. I can add `peer.close` here too

dguenther (approved):

PR #2535, opened 2022-11-09
uses decrypted note sequence to update indexes

hughy:
## Summary

the addition of the block sequence to decryptedNoteValue makes it possible to update the sequenceToNoteHash and nonChainNoteHashes indexes without loading a transaction. it also allows us to ensure that we delete the sequence to hash mapping if a note is removed from the chain.

- updates bulkUpdateDecryptedNotes to update each note's blockHash and sequence from the transaction
- changes updateDecryptedNote to use the note sequence to update indexes
- move function calls that update nullifierToNoteHash, sequenceToNoteHash, and nonChainNoteHashes into updateDecryptedNote and deleteDecryptedNote. the purpose of this is to try to ensure that whenever we update a decrypted note we also update each of those three indexes
- reuse processTransactionSpends in expireTransaction

- updates wallet.getUnspentNotes to use note sequence instead of loading transaction to check confirmations

- adds test to reproduce negative balance

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/migrations/data/014-note-sequence.ts
+++ b/ironfish/src/migrations/data/014-note-sequence.ts
@@ -1,0 +1,130 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Logger } from '../../logger'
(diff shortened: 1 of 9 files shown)


hughy:
Closing this since we're refactoring a lot of the code that this touches to adopt decryptedNote.sequence

PR #2537, opened 2022-11-09
Update node-datachannel to 0.4.0

dguenther:
## Summary

The broken Windows prebuilt binaries were fixed in this commit: https://github.com/murat-dogan/node-datachannel/commit/d087a7a5868d326e85ff4e08f465ce896e2845cc If someone on Windows has attempted to install this version, the older, nonworking version will be cached. But previously we had shipped 0.3.6, so none of our users should have installed 0.4.0.

As with 0.3.6, this has the benefit of adding support for prebuilt binaries on M1 machines (not requiring OpenSSL + cmake to be installed)

Also switched from calling `peer.close()` to `peer.destroy()`, since it also handles cleaning up callbacks. I ran this overnight to ensure it didn't crash.

## Testing Plan

* [x] Run `yarn` from root on an M1 machine without cmake installed and make sure it installs
* [x] Run `yarn` from root on a Windows machine without cmake installed and make sure it installs

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/package.json
+++ b/ironfish/package.json
@@ -36,7 +36,7 @@
     "leveldown": "5.6.0",
     "levelup": "4.4.0",
     "lodash": "4.17.21",
-    "node-datachannel": "0.3.5",
+    "node-datachannel": "0.4.0",
     "node-forge": "1.3.1",
     "parse-json": "5.2.0",
     "sqlite": "4.0.23",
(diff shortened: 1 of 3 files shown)


danield9tqh (approved):

dguenther:
I'll test this on Windows this evening

dguenther:
Tested on a Windows machine without cmake, this works as expected.

PR #2538, opened 2022-11-09
adds note for spender change to estimated tx size

hughy:
## Summary

getPendingTransactionSize currently only adds notes for receives. if there is change left over after the fee then another note needs to be added to return change to the sender.

- adds note size when spender has change
- does not call createSpends a second time when estimating size because it will return the same notes from the first call
- assumes that only one more spend will cover the fee to avoid recursively calling createSpends
- updates test to avoid hardcoded fee

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/memPool/feeEstimator.ts
+++ b/ironfish/src/memPool/feeEstimator.ts
@@ -226,7 +226,7 @@ export class FeeEstimator {
   private async getPendingTransactionSize(
     sender: Account,
     receives: { publicAddress: string; amount: bigint; memo: string }[],
-    estimateFeeRate?: bigint,
+    feeRate: bigint,
   ): Promise<number> {
     let size = 0
     size += 8 // spends length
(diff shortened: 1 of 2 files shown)


ironfish/src/memPool/feeEstimator.ts
```diff
@@ -226,7 +226,7 @@ export class FeeEstimator {
   private async getPendingTransactionSize(
     sender: Account,
     receives: { publicAddress: string; amount: bigint; memo: string }[],
-    estimateFeeRate?: bigint,
+    feeRate: bigint,
   ): Promise<number> {
     let size = 0
     size += 8 // spends length
```
ygao76, line 232: nit: can you use a constant like Numeric_Value_Length to replace the hardcoded number. and new constants for signature and expiration length too. a good place for these constants are in ironfish/src/primitives/transaction.ts

ygao76 (approved):
Looks good, left one comment.

ironfish/src/memPool/feeEstimator.ts
hughy: That's a good idea! I'll create a separate issue for this; it's tangential to the scope of these changes

PR #2540, opened 2022-11-09
Remove diversifier from public address

ygao76:
## Summary
Use constant generator for g_d when generate public address
Replace payment address with transmission key from public address
Dont use zcash Note, create new cmu calculation function. 
Update note length
Add constant to represent length
Update fixture data

## Testing Plan
unit test

## Breaking Change

```
[ ] Yes
```
@mat-if @rohanjadvani


ironfish-rust/src/keys/public_address.rs
```diff
@@ -179,7 +150,8 @@ impl PublicAddress {
     /// another because `pk_d` is not a name I want to expose in a public
     /// interface.
     pub(crate) fn sapling_payment_address(&self) -> PaymentAddress {
```
ygao76, line 152: do we still need this sapling payment address? should we call public_address when we need a payment address?

ironfish-rust/src/keys/public_address.rs
```diff
@@ -110,10 +94,9 @@ impl PublicAddress {
...
-    pub fn public_address(&self) -> [u8; 43] {
-        let mut result = [0; 43];
-        result[..11].copy_from_slice(&self.diversifier.0);
-        result[11..].copy_from_slice(
+    pub fn public_address(&self) -> [u8; 32] {
+        let mut result = [0; 32];
+        result[..32].copy_from_slice(
             &point_to_bytes(&self.transmission_key)
                 .expect("transmission key should be convertible to bytes"),
         );
```
iamalwaysuncomfortable, line 102: Given memory is expensive to allocate and `point_to_bytes` already allocates a 32-byte array, it is possible to just use the return value of point_to_bytes directly

```suggestion
        let result = point_to_bytes(&self.transmission_key)
                .expect("transmission key should be convertible to bytes")
```

ironfish-rust/src/keys/public_address.rs
```diff
@@ -16,92 +16,76 @@ use std::{convert::TryInto, io};
 
 use super::{IncomingViewKey, SaplingKey};
 
-/// The address to which funds can be sent, stored as a diversifier and public
-/// transmission key. Combining a diversifier with an incoming_viewing_key allows
-/// the creation of multiple public addresses without revealing the viewing key.
-/// This allows the user to have multiple "accounts", or to even have different
-/// payment addresses per transaction.
+/// The generator for public address.
+pub const PUBLIC_KEY_GENERATOR: SubgroupPoint = SubgroupPoint::from_raw_unchecked(
```
iamalwaysuncomfortable, line 20: It's likely best not expose the generator as part of the public api

```suggestion
pub(crate) const PUBLIC_KEY_GENERATOR: SubgroupPoint = SubgroupPoint::from_raw_unchecked(
```

ironfish-rust/src/keys/public_address.rs
```diff
@@ -16,92 +16,76 @@ use std::{convert::TryInto, io};
...
+        0x0daf_03b5_47e2_022b,
+        0x6dad_65e6_2328_d37a,
+    ]),
+    bls12_381::Scalar::from_raw([
+        0x5095_1f1f_eff0_8278,
+        0xf0b7_03d5_3a3e_dd4e,
+        0xca01_f580_9c00_eee2,
+        0x6996_932c_ece1_f4bb,
+    ]),
+);
```
iamalwaysuncomfortable, line 33: Are we %100 sure this is in the correct prime subgroup + is it known that there aren't any attacks  where people try to move this point into a different coset?

iamalwaysuncomfortable (commented):
Couple stylistic comments + question about the generator

ironfish-rust/src/keys/public_address.rs
iamalwaysuncomfortable: Also do we want this to have an error condition instead of a crash? Crashing across boundaries is normally undefined behavior.

ironfish-rust/src/note.rs
```diff
@@ -196,11 +190,10 @@ impl<'a> Note {
...
     pub fn encrypt(&self, shared_secret: &[u8; 32]) -> [u8; ENCRYPTED_NOTE_SIZE + aead::MAC_SIZE] {
         let mut bytes_to_encrypt = [0; ENCRYPTED_NOTE_SIZE];
-        bytes_to_encrypt[..11].copy_from_slice(&self.owner.diversifier.0[..]);
-        bytes_to_encrypt[11..43].clone_from_slice(self.randomness.to_repr().as_ref());
+        bytes_to_encrypt[..32].clone_from_slice(self.randomness.to_repr().as_ref());
 
-        LittleEndian::write_u64_into(&[self.value], &mut bytes_to_encrypt[43..51]);
-        bytes_to_encrypt[51..].copy_from_slice(&self.memo.0[..]);
+        LittleEndian::write_u64_into(&[self.value], &mut bytes_to_encrypt[32..40]);
+        bytes_to_encrypt[40..].copy_from_slice(&self.memo.0[..]);
```
iamalwaysuncomfortable, line 196: I'd like to possibly suggest that there be constants representing byte length values to avoid potential proof errors?

ironfish-rust/src/keys/public_address.rs
ygao76: Remove the generator point value here. Import from zkp package instead.

ironfish-zkp/src/lib.rs
```diff
@@ -7,5 +7,6 @@ pub use zcash_primitives::sapling::{
 };
 
 pub mod proofs {
-    pub use zcash_proofs::circuit::sapling::{Output, Spend};
+    pub use crate::circuits::{output::Output, spend::Spend};
+    pub use crate::constants::PUBLIC_KEY_GENERATOR;
```
mat-if, line 11: This is already accessible from `ironfish_zkp::constants::PUBLIC_KEY_GENERATOR`, so we don't need to duplicate the export here.
```suggestion
```

ironfish-rust/src/note.rs
```diff
@@ -196,25 +195,69 @@ impl<'a> Note {
...
         let mut encrypted_bytes = [0; ENCRYPTED_NOTE_SIZE + aead::MAC_SIZE];
         aead::encrypt(shared_secret, &bytes_to_encrypt, &mut encrypted_bytes);
 
         encrypted_bytes
     }
 
+    /// Computes the note commitment, returning the full point.
+    fn cm_full_point(&self) -> jubjub::SubgroupPoint {
```
mat-if, line 209: What do you think about renaming this? I think it's helpful to use a more clear naming scheme
```suggestion
    fn commitment_full_point(&self) -> jubjub::SubgroupPoint {
```

ironfish-zkp/src/circuits/output.rs
```diff
@@ -206,20 +187,32 @@ mod test {
...
+                    .get_u();
+
+                // let expected_cmu = Some(Note {
+                //     value: value_commitment.value,
+                //     rseed: Rseed::BeforeZip212(commitment_randomness),
+                //     g_d: PUBLIC_KEY_GENERATOR,
+                //     pk_d: payment_address,
+                // })
+                // .expect("should be valid")
+                // .cmu();
```
mat-if, line 215:
```suggestion
```

ironfish-zkp/src/circuits/util.rs
```diff
@@ -132,3 +140,34 @@ where
 
     Ok(value_bits)
 }
+
+/// Computes the note commitment, returning the full point.
+pub fn commitment_full_point(note: Note) -> jubjub::SubgroupPoint {
```
mat-if, line 145: `circuits/util.rs` is for functions used in circuits. While this is used in a test for a circuit, it isn't being used by a circuit directly. What do you think about putting this in like `ironfish-zkp/src/util.rs` or something?

ironfish-rust/src/keys/public_address.rs
```diff
@@ -110,38 +78,24 @@ impl PublicAddress {
 
     /// Retrieve the public address in byte form. It is comprised of the
     /// 11 byte diversifier followed by the 32 byte transmission key.
```
mat-if, line 80:
```suggestion
    /// Retrieve the public address in byte form.
```

ironfish-zkp/src/circuits/output.rs
```diff
@@ -52,33 +52,15 @@ impl Circuit<bls12_381::Scalar> for Output {
 
         // Let's deal with g_d
```
mat-if, line 53:
```suggestion
        // Let's deal with epk
```

ironfish-zkp/src/circuits/spend.rs
```diff
@@ -138,23 +137,15 @@ impl Circuit<bls12_381::Scalar> for Spend {
...
-            Some(ExtendedPoint::from(PUBLIC_KEY_GENERATOR)),
-        )?;
-
-        // Check that g_d is not small order. Technically, this check
-        // is already done in the Output circuit, and this proof ensures
-        // g_d is bound to a product of that check, but for defense in
-        // depth let's check it anyway. It's cheap.
-        g_d.assert_not_small_order(cs.namespace(|| "g_d not small order"))?;
-
         // Compute pk_d = g_d^ivk
```
mat-if, line 140:
```suggestion
        // Compute pk_d
```

ironfish/src/migrations/data/013-wallet-2/old/accounts.ts
```diff
@@ -5,7 +5,7 @@ import bufio from 'bufio'
 import { IDatabaseEncoding, IDatabaseStore } from '../../../../storage'
 
 const KEY_LENGTH = 32
-const PUBLIC_ADDRESS_LENGTH = 43
+const PUBLIC_ADDRESS_LENGTH = 32
```
ygao76, line 9: @NullSoldier Do I need to run migration for wallet db for this change during network reset? If I dont need I will keep this change. If I need to run a migration I will not change this file.

ygao76: cc @mat-if

NullSoldier: You cannot modify old migrations like this or it will break the migration.

ygao76: Ok, I will revert.

ironfish/src/wallet/wallet.test.ts
```diff
@@ -297,35 +297,6 @@ describe('Accounts', () => {
   })
 
   describe('scanTransactions', () => {
-    it('should update head status', async () => {
```
ygao76, line 300: I have to delete these two wallet tests for now. Its failing at line 315. and line 418. Somehow the header hash is not null, the failure def has nothing to do with public address change. I am deleting these temporarily I will revisit once the change merged. Since current changes become hard to manage.

mat-if (approved):

ironfish/src/wallet/wallet.test.ts
hughy: This pr should fix these two tests: https://github.com/iron-fish/ironfish/pull/2609

ironfish/src/workerPool/job.test.slow.ts
```diff
@@ -21,20 +20,4 @@ describe('Worker Pool', () => {
 
     expect(workerPool.completed).toBe(1)
   })
-
-  it('verify', async () => {
-    const { workerPool } = nodeTest
```
ygao76, line 26: Delete this test bc its loading data from wallet db for genesis block, which has old data format. Need a db migration before adding it back.

PR #2542, opened 2022-11-10
Import Output and Spend class from zcash

ygao76:
## Summary
Import Output and Spend class from zcash sapling
## Testing Plan
unit test
## Breaking Change

```
[ ] Yes
```
@mat-if @rohanjadvani


ironfish-zkp/Cargo.toml
```diff
@@ -21,3 +21,5 @@ jubjub = "0.9.0"
 rand = "0.8.5"
 zcash_primitives = "0.7.0"
 zcash_proofs = "0.7.1"
+rand_core = "0.6.3"
```
mat-if, line 24: We can skip this import, since we already import `rand`. Any import that uses this can be replaced. `rand_core::{...` -> `rand::{...`

ironfish-zkp/src/circuits/output.rs
```diff
@@ -0,0 +1,263 @@
...
+        Ok(())
+    }
+}
+
+#[test]
+fn test_output_circuit_with_bls12_381() {
+    use bellman::gadgets::test::*;
+    use ff::Field;
+    use group::Group;
+    use rand_core::{RngCore, SeedableRng};
```
mat-if, line 168:
```suggestion
    use rand::{RngCore, SeedableRng};
```
Then we can remove the `rand_core` dependency from the `Cargo.toml`

mat-if:
We'll need to check in the changes to the `Cargo.lock` file, and we should be able to change the imports in `ironfish-zkp/lib.rs` to use our forked circuits instead of the librustzcash ones now.

ironfish-zkp/src/circuits/spend.rs
```diff
@@ -0,0 +1,326 @@
...
+    circuit::{
+        ecc::{self},
+        pedersen_hash,
+    },
+    constants::{
+        NOTE_COMMITMENT_RANDOMNESS_GENERATOR, NULLIFIER_POSITION_GENERATOR,
+        PROOF_GENERATION_KEY_GENERATOR, SPENDING_KEY_GENERATOR,
+    },
+};
+
```
jowparks, line 25: Is this code a fork of the ZCash definition? Would be nice to include a comment breadcrumb link if it is.

mat-if:
Since this isn't a breaking change, let's merge this into staging

mat-if (approved):

ironfish-zkp/src/circuits/spend.rs
ygao76: Yeah, but they wont stay the same as ZCash class.  I will modify these code in my next PR.

PR #2544, opened 2022-11-10
feat: masp transaction syncer

jowparks:
## Summary
Creates a base syncer, then split the two syncers (masp, and deposit) into separate commands.

The MASP command serialization is currently stubbed out until we update the `primitives` to handle multi-asset.
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/service/sync-masp-transactions.ts
+++ b/ironfish-cli/src/commands/service/sync-masp-transactions.ts
@@ -1,11 +1,24 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { MaspTransactionTypes, PromiseUtils, WebApi } from '@ironfish/sdk'
+import {
+  ApiMaspUpload,
+  Assert,
+  GetTransactionStreamResponse,
+  MaspTransactionTypes,
+  Meter,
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/service/sync-deposit-transactions.ts
```diff
@@ -0,0 +1,30 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { ApiDepositUpload, GetTransactionStreamResponse, WebApi } from '@ironfish/sdk'
+import { SyncTransactions } from './sync-transactions'
+
+export default class SyncMaspTransactions extends SyncTransactions<ApiDepositUpload> {
+  static aliases = ['service:syncTransactions']
```
NullSoldier, line 8: delete alias

ironfish-cli/src/commands/service/sync-masp-transactions.ts
```diff
@@ -0,0 +1,189 @@
...
+    // }
+
+    // const response = this.sdk.client.getTransactionStream({
+    //   incomingViewKey: flags.viewKey,
+    //   head: head,
+    // })
+
+    // const speed = new Meter()
+    // speed.start()
+
```
jowparks, line 106: test send

NullSoldier (approved):

PR #2547, opened 2022-11-10
Fix errors when regenerating fixtures for fee estimator tests

dguenther:
## Summary

Some of the fee estimator tests were failing when regenerating the fixtures. Several of them were due to one of the transactions not being accepted into the mempool due to duplicate spends -- I addressed this by creating two separate accounts to receive/spend tokens.

Another one was due to the estimated fee changing -- I'll follow up on this to double-check the value makes sense.

## Testing Plan

Tests should pass.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/memPool/feeEstimator.test.ts
+++ b/ironfish/src/memPool/feeEstimator.test.ts
@@ -135,7 +135,7 @@ describe('FeeEstimator', () => {
       expect(feeEstimator.size(PRIORITY_LEVELS[1])).toBe(0)
       expect(feeEstimator.size(PRIORITY_LEVELS[2])).toBe(0)
 
-      node.memPool.acceptTransaction(transaction)
+      expect(node.memPool.acceptTransaction(transaction)).toBe(true)
 
       feeEstimator.onConnectBlock(block, node.memPool)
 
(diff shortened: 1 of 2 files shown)


ironfish/src/memPool/feeEstimator.test.ts
```diff
@@ -411,7 +405,7 @@ describe('FeeEstimator', () => {
         },
       ])
 
-      expect(fee).toBe(BigInt(6))
+      expect(fee).toBe(BigInt(9))
```
dguenther, line 408: @hughy or @ygao76 , could you double-check to make sure that this value seems reasonable? I'm not quite sure why it changed in this case, but I'd expect it to be pretty similar to the fees used in `transaction` and `transaction2`, so I think it's correct.

hughy, line 408: This looks right. I think I'd like for this to be less of a magic number and derive it from the fee rate and the size of the transaction, but I think that can happen separately

I think the change in the fee estimate came from different transactions/fees being added to the chain in the old fixtures

hughy (approved):

ironfish/src/memPool/feeEstimator.test.ts
dguenther: Cool, sounds good!

PR #2548, opened 2022-11-10
network-reset: Remove miner's fee from block header 

dguenther:
## Summary

Removes the miner's fee from the block header, since it's unnecessary to commit to.

## Breaking Change

```
[ ] Yes
```


ironfish/src/migrations/data/013-wallet-2/old/headers.ts
```diff
@@ -16,6 +18,229 @@ export type HeaderValue = {
   header: BlockHeader
 }
 
+class PartialBlockHeaderSerde {
```
danield9tqh, line 21: This is just copy and paste of previous one?

dguenther: Yep, needed to keep the migrations working.

PR #2549, opened 2022-11-11
Use block size ratio in fee rate estimator

ygao76:
## Summary
Add block size ratio in fee estimation.

Add a new attribute to store the latest block size and block hash. 

## Testing Plan
unit tests
## Breaking Change

```
[ ] Yes
```


ironfish/src/memPool/feeEstimator.ts
```diff
@@ -128,18 +138,35 @@ export class FeeEstimator {
     return feeRates.sort((a, b) => (a > b ? 1 : -1))
   }
 
-  estimateFeeRates(): { low: bigint; medium: bigint; high: bigint } {
+  private async getLastBlockSizeBytes(): Promise<number> {
```
hughy, line 141: Could we update `latestBlockSize` as blocks come in through `onConnectBlock` and `onDisconnectBlock`?

It would be nice not to have to use the chain to look up blocks

ironfish/src/memPool/feeEstimator.ts
```diff
@@ -152,8 +179,13 @@ export class FeeEstimator {
     }
 
     fees.sort((a, b) => (a < b ? -1 : a > b ? 1 : 0))
+    const lastBlockSize = await this.getLastBlockSizeBytes()
+    const blockSizeRatio = lastBlockSize / this.wallet.chain.consensus.MAX_BLOCK_SIZE_BYTES
```
hughy, line 183: It looks like the ratio here is only calculated from the size of the most recent block. That will be very sensitive to individual blocks -- if we have one empty block, then the fee rate will drop to zero.

ironfish/src/memPool/feeEstimator.ts
```diff
@@ -152,8 +179,13 @@ export class FeeEstimator {
...
+
+    let feeRate = fees[Math.round((queue.length - 1) / 2)]
+    feeRate = BigInt(Math.ceil(Number(feeRate) * blockSizeRatio))
```
hughy, line 186: This won't change the fee rate -- `blockSizeRatio` is between 0 and 1 from the above calculation, so `ceil` will make this equal to `feeRate`.

If we can I think it would be good to avoid converting between number and bigint

hughy (commented):
Could this work with updating the `blockSizeRatio` similar to how ETH updates base fee? That's what I was thinking of from the brainstorming earlier this week

https://github.com/ethereum/go-ethereum/blob/a2b7481b7876439d6e4659b377b96f8c36ec1e03/consensus/misc/eip1559.go#L55

that would maintain a ratio or factor on the FeeEstimator that updates with each block

ironfish/src/memPool/feeEstimator.ts
ygao76: in geth implementation, it also only consider one parent block. If we really have a empty block(due to no transaction) then the next transaction should be the minimal since its extremely low demands.

ironfish/src/memPool/feeEstimator.ts
ygao76: yeah I really dont like the conversion between bigint and number. But the fee rate is in bigint. But the ratio is a double presented in number, so the ratio can't be in bigint. Is. there any way I can avoid the conversion. I can multiple 100 with ratio to make it as a integer, and divide the final result by 100 and convert back to bigint. What do you think?

ironfish/src/memPool/feeEstimator.ts
ygao76: I thought about it, its easy to add in onConnectBlock. But for onDisconnectBlock,  i dont have a new value to update this latestBlockSize. In the end, I still need to keep this lastBlockSize check in case onDisconnect delete the value, and it becomes null.

ygao76:
> Could this work with updating the `blockSizeRatio` similar to how ETH updates base fee? That's what I was thinking of from the brainstorming earlier this week
> 
> https://github.com/ethereum/go-ethereum/blob/a2b7481b7876439d6e4659b377b96f8c36ec1e03/consensus/misc/eip1559.go#L55
> 
> that would maintain a ratio or factor on the FeeEstimator that updates with each block

What do you think the main difference between my change and Eth base fee? I follow the same concept of the eth base fee. We maintain a ratio and updates it with the last block. Is it the way we update this ratio that is different? I replied to your comment about onConnect and disConnect.

ironfish/src/memPool/feeEstimator.ts
hughy: One way to address this might be to keep the block sizes in a queue the way we do with the `FeeRateEntry` queues.

ironfish/src/memPool/feeEstimator.ts
hughy: That sounds like a good approach! We do something a bit like that when estimating the hash rate for the mining pool: https://github.com/iron-fish/ironfish/blob/915e2ee30a823e848bb3cb5801e6194096336bfc/ironfish/src/mining/pool.ts#L411

ironfish/src/memPool/feeEstimator.ts
```diff
@@ -211,8 +266,8 @@ export class FeeEstimator {
     return size
   }
 
-  private isFull(array: FeeRateEntry[]): boolean {
-    return array.length === this.maxBlockHistory
+  private isFull(arrayLength: number): boolean {
+    return arrayLength === this.maxBlockHistory
```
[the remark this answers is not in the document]
ygao76: so this function will work for both FeeRate queues and BlockSize queue.

ironfish/src/memPool/feeEstimator.ts
```diff
@@ -17,19 +17,32 @@ export interface FeeRateEntry {
...
+    high: FeeRateEntry[]
+    blockSize: BlockSizeEntry[]
+  }
   private percentiles: PriorityLevelPercentiles
   private wallet: Wallet
   private readonly logger: Logger
   private maxBlockHistory = 10
   private defaultFeeRate = BigInt(1)
+  private latestBlockSize: number
+  private latestBlockHash: Buffer | null = null
```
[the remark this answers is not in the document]
ygao76: no, dont need this anymore.

ironfish/src/memPool/feeEstimator.ts
```diff
@@ -51,23 +65,34 @@ export class FeeEstimator {
...
           queue.push({ feeRate, blockHash: currentBlock.header.hash })
         }
       }
 
+      // construct block size cache
+      if (!this.isFull(this.queues[BLOCK_SIZE].length)) {
+        const blockSize = currentBlock.transactions.reduce(
+          (a, t) => a + t.serialize().length,
+          0,
+        )
```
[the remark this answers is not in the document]
ygao76: I see, there are overhead data in the block. I will use that function to get block size.

ironfish/src/memPool/feeEstimator.ts
```diff
@@ -152,8 +198,17 @@ export class FeeEstimator {
     }
 
     fees.sort((a, b) => (a < b ? -1 : a > b ? 1 : 0))
+    const averageBlockSize =
+      this.queues[BLOCK_SIZE].reduce((a, b) => a + b.blockSize, 0) /
+      this.queues[BLOCK_SIZE].length
+    const blockSizeRatio = BigInt(
```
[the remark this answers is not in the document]
hughy: Discussed elsewhere -- we should not use block size to _increase_ the fee rate: doing so would allow miners to artificially fill blocks with transactions with fees of 0 or 1 $ORE in order to increase fee rate estimates.

hughy (approved):

PR #2554, opened 2022-11-11
Don't ban peer on sync disconnect

danield9tqh:
## Summary
We were unintentionally banning the peer when they timed out during syncing or otherwise failed a request. This is not the worst thing since we want to disconnect from them anyway but we probably don't want to ban them for this

## Testing Plan
Local testing and unit tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/syncer.ts
+++ b/ironfish/src/syncer.ts
@@ -345,25 +345,25 @@ export class Syncer {
     peer: Peer,
     start: Buffer,
     limit: number,
-  ): Promise<{ blocks: Block[]; time: number }> {
+  ): Promise<{ ok: true; blocks: Block[]; time: number } | { ok: false }> {
     this.logger.info(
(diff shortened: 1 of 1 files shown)


ironfish/src/syncer.ts
```diff
@@ -373,10 +373,17 @@ export class Syncer {
     let blocksPromise = this.getBlocks(peer, head, this.blocksPerMessage + 1)
 
     while (currentHead) {
+      const blocksResult = await blocksPromise
+      if (!blocksResult.ok) {
+        peer.close()
```
dguenther, line 378: Do you want to log the error here like we were doing before? It could be generic too

danield9tqh: Still logging it in the catch statement above 👍🏻

dguenther (approved):

PR #2555, opened 2022-11-11
Handle negative balances in depositAll

danield9tqh:
## Summary
There is a bug where the wallet potentially returns a negative balance. Until we fix this issue we don't want the depositAll command to be failed by it.

## Testing Plan
Local testing

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/deposit-all.ts
+++ b/ironfish-cli/src/commands/deposit-all.ts
@@ -202,6 +202,11 @@ export default class DepositAll extends IronfishCommand {
       unconfirmedBalance = CurrencyUtils.decode(balanceResp.content.unconfirmed)
       pendingBalance = CurrencyUtils.decode(balanceResp.content.pending)
 
+      if (confirmedBalance < 0 || unconfirmedBalance < 0 || pendingBalance < 0) {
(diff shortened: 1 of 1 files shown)


hughy (approved):

ironfish-cli/src/commands/deposit-all.ts
```diff
@@ -202,6 +202,11 @@ export default class DepositAll extends IronfishCommand {
       unconfirmedBalance = CurrencyUtils.decode(balanceResp.content.unconfirmed)
       pendingBalance = CurrencyUtils.decode(balanceResp.content.pending)
 
+      if (confirmedBalance < 0 || unconfirmedBalance < 0 || pendingBalance < 0) {
+        await PromiseUtils.sleep(30000)
+        continue
+      }
```
dguenther, line 208: It seems unusual for any balances to be negative -- Does it fix itself eventually if it continues like this? Otherwise, I wonder if it's a bug in the wallet, and if it's better to exit and tell users to rescan their accounts.

dguenther:
Ah, looks like #2565 addresses this.

PR #2558, opened 2022-11-12
adds accounts:repair command

hughy:
## Summary

adds a cli command intended to fix balance and note indexing issues in the wallet without the need for a full rescan/reset of the wallet.

repairs the balance by recomputing it from 0 and adding all unspent notes. uses the chain to determine if each note has been spent and changes to spent if the note is marked as unspent

repairs sequenceToNoteHashes by ensuring that all notes that are on chain have the sequence of the transaction that created the note. any sequence to note hash mapping that does not have a note is removed.

repairs nonChainNoteHashes by removing any note hashes for notes that are on chain.

repairs nullifierToNoteHash by removing any nullifier that maps to a note that is not in the wallet or maps to a note that is not on the chain.

https://coda.io/d/Engineering-Specs_dMnayiE39lL/accounts-repair_suPC4#_lueQX

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/accounts/repair.ts
+++ b/ironfish-cli/src/commands/accounts/repair.ts
@@ -103,7 +103,7 @@ export default class Repair extends IronfishCommand {
 
         await walletDb.saveDecryptedNote(account, decryptedNoteValue.hash, {
           ...decryptedNoteValue,
-          spent,
+          spent: true,
         })
       } else if (
         !spent &&
@@ -112,6 +112,13 @@ export default class Repair extends IronfishCommand {
           chain.head.sequence,
         )
       ) {
+        if (decryptedNoteValue.spent) {
+          await walletDb.saveDecryptedNote(account, decryptedNoteValue.hash, {
+            ...decryptedNoteValue,
+            spent: false,
+          })
+        }
+
         unconfirmedBalance += decryptedNoteValue.note.value()
       }
     }
@@ -135,7 +142,7 @@ export default class Repair extends IronfishCommand {
     )) {
       const decryptedNoteValue = await account.getDecryptedNote(noteHash)
 
-      if (!decryptedNoteValue) {
+      if (!decryptedNoteValue || !decryptedNoteValue.nullifier) {
         missingNotes++
 
         await walletDb.deleteNullifier(account, nullifier)
@@ -175,6 +182,8 @@ export default class Repair extends IronfishCommand {
         incorrectSequences++
 
         await walletDb.sequenceToNoteHash.del([account.prefix, [sequence, noteHash]])
+
+        await walletDb.setNoteHashSequence(account, noteHash, transactionValue.sequence)
       }
     }
 
--- a/ironfish-cli/src/commands/accounts/repair.ts
+++ b/ironfish-cli/src/commands/accounts/repair.ts
@@ -49,7 +49,7 @@ export default class Repair extends IronfishCommand {
     await this.repairNullifierToNoteHash(account, node.wallet.walletDb)
 
     this.log('Repairing sequenceToNoteHash')
-    await this.repairSequenceToNoteHash(account, node.wallet.walletDb)
+    await this.repairSequenceToNoteHash(account, node.wallet.walletDb, node.chain)
   }
 
   private loadAccount(wallet: Wallet, accountName: string | undefined): Account {
@@ -88,7 +88,13 @@ export default class Repair extends IronfishCommand {
 
         Assert.isNotUndefined(
           transactionValue,
-          'Account has a note but is missing the transaction that it received the note from. Account must be rescanned',
+          'Account has a note but is missing the transaction that it received the note from. Account must be rescanned using `accounts:rescan --reset`',
+        )
+
+        await this.verifyBlockContainsTransaction(
+          decryptedNoteValue.transactionHash,
+          transactionValue.blockHash,
+          chain,
         )
 
         await walletDb.setNoteHashSequence(
@@ -101,7 +107,7 @@ export default class Repair extends IronfishCommand {
         if (!decryptedNoteValue.nullifier) {
           if (transactionValue.sequence) {
             throw new Error(
-              'Transaction marked as on chain, but note missing nullifier. Account must be rescanned.',
+              'Transaction marked as on chain, but note missing nullifier. Account must be rescanned using `accounts:rescan --reset`.',
             )
           }
 
@@ -188,7 +194,11 @@ export default class Repair extends IronfishCommand {
     )
   }
 
-  private async repairSequenceToNoteHash(account: Account, walletDb: WalletDB): Promise<void> {
+  private async repairSequenceToNoteHash(
+    account: Account,
+    walletDb: WalletDB,
+    chain: Blockchain,
+  ): Promise<void> {
     let sequenceMismatches = 0
     let missingNotes = 0
 
@@ -210,7 +220,13 @@ export default class Repair extends IronfishCommand {
 
           Assert.isNotUndefined(
             transactionValue,
-            'Account has a note but is missing the transaction that it received the note from. Account must be rescanned.',
+            'Account has a note but is missing the transaction that it received the note from. Account must be rescanned using `accounts:rescan --reset`.',
+          )
+
+          await this.verifyBlockContainsTransaction(
+            decryptedNoteValue.transactionHash,
+            transactionValue.blockHash,
+            chain,
           )
 
           if (transactionValue.sequence !== sequence) {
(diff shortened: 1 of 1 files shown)


hughy:
Example output:
```
Repairing wallet for account default
Repairing decryptedNotes and balance
        Saving new unconfirmed balance: $IRON 0.01019860
        Repaired 0 decrypted notes incorrectly marked as unspent
        Repaired 0 nullifiers mapped to an incorrect note hash
Repairing nullifierToNote
        Repaired 0 nullifiers that map to notes that are not in the wallet
        Repaired 0 nullifiers that map to notes that are not on chain
Repairing sequenceToNoteHash
        Repaired 0 sequenceToNoteHash mappings with missing notes
        Repaired 0 sequenceToNoteHash mappings with incorrect sequences
✨  Done in 3.78s.

File: ironfish-cli/src/commands/accounts/repair.ts [hunk 1]
```diff
@@ -0,0 +1,232 @@
...
+    await walletDb.db.transaction(async (tx) => {
+      for await (const decryptedNoteValue of account.getNotes()) {
+        const transactionValue = await account.getTransaction(
+          decryptedNoteValue.transactionHash,
+          tx,
+        )
+
+        Assert.isNotUndefined(
+          transactionValue,
+          'Account has a note but is missing the transaction that it received the note from. Account must be rescanned',
```
danield9tqh, line 91: Maybe should give the rescan command here

ironfish-cli/src/commands/accounts/repair.ts
```diff
@@ -0,0 +1,276 @@
...
+        await walletDb.setNoteHashSequence(
+          account,
+          decryptedNoteValue.hash,
+          transactionValue.sequence,
+          tx,
+        )
+
+        if (!decryptedNoteValue.nullifier) {
+          if (transactionValue.sequence) {
+            throw new Error(
```
danield9tqh, line 109: I think in this case we already have verified if the transaction is on the block or not. We could get the nullifier from the transaction and patch it

ironfish-cli/src/commands/accounts/repair.ts
```diff
@@ -0,0 +1,276 @@
...
+          decryptedNoteValue.transactionHash,
+          tx,
+        )
+
+        Assert.isNotUndefined(
+          transactionValue,
+          'Account has a note but is missing the transaction that it received the note from. Account must be rescanned using `accounts:rescan --reset`',
+        )
+
+        await this.verifyBlockContainsTransaction(
```
danield9tqh, line 94: Here I think we may want to do `chain.getHashAtSequence` instead of `chain.getBlock`. Then we can verify if the transaction is actually on the main chain or not

ironfish-cli/src/commands/accounts/repair.ts
```diff
@@ -0,0 +1,276 @@
...
+          if (transactionValue.sequence) {
+            throw new Error(
+              'Transaction marked as on chain, but note missing nullifier. Account must be rescanned using `accounts:rescan --reset`.',
+            )
+          }
+
+          continue
+        }
+
+        const spent = await chain.nullifiers.contains(decryptedNoteValue.nullifier)
```
danield9tqh, line 117: Here I think we may want to do `chain.getHashAtSequence` instead since we were having issues with the nullifier contains. This would depend on running the other repair command correctly

ironfish-cli/src/commands/accounts/repair.ts
```diff
@@ -0,0 +1,276 @@
...
+          'Account has a note but is missing the transaction that it received the note from. Account must be rescanned using `accounts:rescan --reset`',
+        )
+
+        await this.verifyBlockContainsTransaction(
+          decryptedNoteValue.transactionHash,
+          transactionValue.blockHash,
+          chain,
+        )
+
+        await walletDb.setNoteHashSequence(
```
danield9tqh, line 100: We should do this after we've verified that it's on chain, probably by `chain.getHashAtSequence`

ironfish-cli/src/commands/accounts/repair.ts
```diff
@@ -0,0 +1,276 @@
...
+
+          continue
+        }
+
+        const spent = await chain.nullifiers.contains(decryptedNoteValue.nullifier)
+
+        if (spent && !decryptedNoteValue.spent) {
+          noteUnspentMismatches++
+
+          await walletDb.saveDecryptedNote(
```
danield9tqh, line 122: I feel like we could combine some of these operations with walletDB.updateDecryptedNote. Or if we don't trust that function create a similar one. I think we should update `decryptedNote.nullifier`, `decryptedNote.index`, `decryptedNote.spent` and `setNoteHashSequence` all at the same time since they should all be in sync

ironfish-cli/src/commands/accounts/repair.ts
hughy: Would getting the nullifier from the transaction require decrypting the notes again? Should we call `syncTransaction`?

ironfish-cli/src/commands/accounts/repair.ts
hughy: Discussed on a call -- we still need to check `chain.nullifiers` because we don't have a way to determine the transaction or the block that a note was spent on.

ironfish-cli/src/commands/accounts/repair.ts
danield9tqh: Oh yea, that might be a good idea

ironfish-cli/src/commands/accounts/repair.ts
hughy: Updated this method to only ever change notes from unspent to spent. The indexes won't change from that update, so it's more efficient (and maybe safer) to save the note directly.

ironfish-cli/src/commands/accounts/repair.ts
hughy: Updated to limit the scope of repair to only address a few issues, so I think re- `syncTransaction` isn't necessary

ironfish-cli/src/commands/accounts/repair.ts
```diff
@@ -0,0 +1,197 @@
...
+            account,
+            decryptedNoteValue.hash,
+            {
+              ...decryptedNoteValue,
+              spent,
+            },
+            tx,
+          )
+        } else if (!spent) {
+          unconfirmedBalance += decryptedNoteValue.note.value()
```
hughy, line 118: We may also need to check whether a transaction is already expired. I've identified an instance where a wallet seems to have re-synced a transaction that was already expired.

ironfish-cli/src/commands/accounts/repair.ts
```diff
@@ -0,0 +1,203 @@
...
+  private async repairBalance(
+    account: Account,
+    walletDb: WalletDB,
+    chain: Blockchain,
+  ): Promise<void> {
+    let unconfirmedBalance = 0n
+
+    let noteUnspentMismatches = 0
+
+    await walletDb.db.transaction(async (tx) => {
```
NullSoldier, line 81: This transaction will make it so any larger account can't run this command. Do we need this transaction?

hughy: We don't need the transaction -- we should be able to safely update each note and the balance separately.

hughy: Removed

ironfish-cli/src/commands/accounts/repair.ts
```diff
@@ -0,0 +1,183 @@
...
+      }
+
+      const spent = await chain.nullifiers.contains(decryptedNoteValue.nullifier)
+
+      if (spent && !decryptedNoteValue.spent) {
+        noteUnspentMismatches++
+
+        await walletDb.saveDecryptedNote(account, decryptedNoteValue.hash, {
+          ...decryptedNoteValue,
+          spent,
```
NullSoldier, line 106:
```suggestion
          true,
```

ironfish-cli/src/commands/accounts/repair.ts
```diff
@@ -0,0 +1,183 @@
...
+        })
+      } else if (
+        !spent &&
+        !chain.verifier.isExpiredSequence(
+          transactionValue.transaction.expirationSequence(),
+          chain.head.sequence,
+        )
+      ) {
+        unconfirmedBalance += decryptedNoteValue.note.value()
```
NullSoldier, line 115: we don't mark the note as unspent here

ironfish-cli/src/commands/accounts/repair.ts
```diff
@@ -0,0 +1,183 @@
...
+      undefined,
+      account.prefixRange,
+    )) {
+      const decryptedNoteValue = await account.getDecryptedNote(noteHash)
+
+      if (!decryptedNoteValue) {
+        missingNotes++
+
+        await walletDb.deleteNullifier(account, nullifier)
+      }
```
NullSoldier, line 142: Could we also do this if decryptedNoteValue.nullifier is null?

ironfish-cli/src/commands/accounts/repair.ts
```diff
@@ -0,0 +1,183 @@
...
+      Assert.isNotUndefined(
+        transactionValue,
+        `Account has a note but is missing the transaction that it received the note from. ${RESCAN_MESSAGE}`,
+      )
+
+      if (transactionValue.sequence !== sequence) {
+        incorrectSequences++
+
+        await walletDb.sequenceToNoteHash.del([account.prefix, [sequence, noteHash]])
+      }
```
NullSoldier, line 178: should this put it at the correct sequence?

NullSoldier (approved):

ironfish-cli/src/commands/accounts/repair.ts
hughy: Updated 👍

ironfish-cli/src/commands/accounts/repair.ts
hughy: Updated

PR #2560, opened 2022-11-12
remove magic number for padding and formatted into a table

aditya-manit:
## Summary

Removes magic numbers added to pad the string, instead format the output into table which is much more readable code wise and display wise too

## Output before
Screenshot 2022-11-12 at 11 51 14 AM

## Output now
Screenshot 2022-11-12 at 11 49 51 AM

## Testing Plan

Tested on local

## Breaking Change

No

### DIFF
--- a/ironfish-cli/src/commands/workers/status.ts
+++ b/ironfish-cli/src/commands/workers/status.ts
@@ -4,6 +4,7 @@
 import { GetWorkersStatusResponse, PromiseUtils } from '@ironfish/sdk'
 import { Flags } from '@oclif/core'
 import blessed from 'blessed'
+import { table } from 'table'
 import { IronfishCommand } from '../../command'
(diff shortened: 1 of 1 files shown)


danield9tqh:
@aditya-manit I added this same comment to another one of your PRs, this also seems like a purely subjective UI improvement. We steer clear of accepting [new design decisions](https://github.com/iron-fish/ironfish/blob/master/CONTRIBUTING.md#making-new-product-design-decisions-that-have-no-precedent) as they are subjective.

Separately this PR has a bug where the `table` library is not a part of the production package. It is not in our package.json so this change will not work in production. If you're looking for more direction we do have a couple of issues in [Good First Issues Label](https://github.com/iron-fish/ironfish/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22)

PR #2562, opened 2022-11-12
improve output display

aditya-manit:
## Summary
Improves the output display of rpc:status
## Testing Plan
Tested on local

## Output before
Screenshot 2022-11-12 at 12 58 59 PM

## Output now
Screenshot 2022-11-12 at 1 10 56 PM

If there are multiple adapters it will output multiple such tables

## Breaking Change
No


danield9tqh:
@aditya-manit while we appreciate the effort in setting up the codebase, this seems like a purely subjective UI improvement. Additionally the `table` library is not a part of the production package. It is not in our package.json so this change will not work in production. If you're looking for more direction we do have a couple of issues in [Good First Issues Label](https://github.com/iron-fish/ironfish/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22)

PR #2563, opened 2022-11-12
print table for miners:pools:status

aditya-manit:
## Summary
Enhance printing output for yarn start miners:pools:status

## Output before
Screenshot 2022-11-12 at 1 34 32 PM

## Output now
Screenshot 2022-11-12 at 2 01 17 PM

## Testing Plan
Tested on local

## Breaking Change

No

## PS:
It took me a long time to setup the local environment and get everything running in local to test the changes, please accept atleast one of my PRs as a big one.

### DIFF
--- a/ironfish-cli/src/commands/miners/pools/status.ts
+++ b/ironfish-cli/src/commands/miners/pools/status.ts
@@ -108,7 +108,7 @@ export class PoolStatus extends IronfishCommand {
         alignment: 'center',
         content: `Status of mining pool: ${status.name}`,
       },
-      columns: [{ alignment: 'left', width: 100 }],
+      columns: [{ alignment: 'left', width: 110 }],
     }
 
     result.push(['Miners', `${status.clients}`])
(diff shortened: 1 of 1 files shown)


danield9tqh:
@aditya-manit I added this same comment to another one of your PRs, this also seems like a purely subjective UI improvement. It also has the same problem that the `table` library is not a part of the production package. It is not in our package.json so this change will not work in production. If you're looking for more direction we do have a couple of issues in [Good First Issues Label](https://github.com/iron-fish/ironfish/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22)

aditya-manit:
> I added this same comment to another one of your PRs, this also seems like a purely subjective UI improvement. We steer clear of accepting [new design decisions](https://github.com/iron-fish/ironfish/blob/master/CONTRIBUTING.md#making-new-product-design-decisions-that-have-no-precedent) as they are subjective.

Ah I see, but it looks much cleaner this way if you noticed. Any way this can be discussed internal and my PRs can be merged ?

> Separately this PR has a bug where the table library is not a part of the production package.

I can address and fix that

PR #2564, opened 2022-11-12
Defer syncing transaction until after broadcasting

dguenther:
## Summary

Syncing a transaction can be slow since it needs to call decryptNotes and update the wallet state, so we should defer it until after rebroadcasting the transaction to avoid delaying transaction propagation. Additionally, we should only need to sync the transaction once when it's initially accepted to the mempool.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/peerNetwork.ts
+++ b/ironfish/src/network/peerNetwork.ts
@@ -1352,16 +1352,19 @@ export class PeerNetwork {
         return
       }
 
-      // The accounts need to know about the transaction since it could be
-      // relevant to the accounts, despite coming from a different node.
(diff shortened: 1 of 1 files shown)


hughy (approved):
This also has the benefit of not syncing the transaction to the wallet if it wasn't accepted to the mempool!

dguenther:
I think I'm going to change this back to call `syncTransaction` every time, but still call it at the end of the function. What I'm probably really looking for here is the ability to turn off the wallet, basically, to save the `decryptNotes` CPU usage on nodes that are just validating/syncing.

PR #2565, opened 2022-11-12
fixes cause of negative balances

hughy:
## Summary

wallets may have negative balances after the following sequence of events:

- transaction *A* created, mined on block *x*
  - output notes from *A* now have `sequenceToNoteHash` mappings with sequence of *x*
- re-org removes transaction *A* from chain
  - notes from *A* added to `nonChainNoteHashes`
  - notes from *A* still have sequence *x* in `sequenceToNoteHash`
- balance calculated for account
  - notes from *A* subtracted from pending balance because they are marked as sequence *x* in `sequenceToNoteHash`
  - notes from *A* subtracted again from unconfirmed balance because they are in `nonChainNoteHashes`

the solution is to be sure to remove the entry in `sequenceToNoteHash` when *x* is removed from the chain. the notes do not have sequence data (yet), so we cannot remove this index when we update the note without keeping track of the former sequence of the transaction (*A*).

## Breaking Change

```
[ ] Yes
```


dguenther (approved):
Verified the test fails without the wallet changes, nice!

PR #2568, opened 2022-11-13
adds chain:rewind command

hughy:
## Summary

given a sequence, rewind the chain by disconnecting blocks until the head of the chain is at that sequence.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/chain/rewind.ts
+++ b/ironfish-cli/src/commands/chain/rewind.ts
@@ -1,0 +1,75 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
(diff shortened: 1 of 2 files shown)


hughy:
example output:
```
Chain head currently at 267000. Rewinding 200 blocks to 266800.
Rewinding chain: [████████████████████████████████████████] 100% | 200 / 200 blocks | 8.43/s | ETA: soon
✨  Done in 24.74s.
```

ironfish-cli/src/commands/chain/rewind.ts
```diff
@@ -0,0 +1,75 @@
...
+  async start(): Promise<void> {
+    const { args } = await this.parse(Rewind)
+
+    const node = await this.sdk.node()
+
+    await NodeUtils.waitForOpen(node)
+
+    const toDisconnect = node.chain.head.sequence - args.sequence
+
+    this.log(
```
danield9tqh, line 34: Should we check if sequence is less than head sequence first

NullSoldier (approved):

PR #2579, opened 2022-11-15
Replace Payment Address with Subgroup Point in Output and Spend

ygao76:
## Summary
Replace Payment Address with Subgroup Point in Output and Spend
## Testing Plan
unit tests
## Breaking Change

```
[ ] Yes
```
@mat-if @rohanjadvani

### DIFF
--- a/ironfish-zkp/src/circuits/output.rs
+++ b/ironfish-zkp/src/circuits/output.rs
@@ -3,8 +3,8 @@ use group::Curve;
 
 use bellman::{Circuit, ConstraintSystem, SynthesisError};
 
-use jubjub::{SubgroupPoint, ExtendedPoint};
-use zcash_primitives::sapling::{PaymentAddress, ValueCommitment};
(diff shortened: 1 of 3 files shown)


ironfish-zkp/src/circuits/output.rs
```diff
@@ -0,0 +1,265 @@
...
+        ecc::{self},
+        pedersen_hash,
+    },
+    constants::NOTE_COMMITMENT_RANDOMNESS_GENERATOR,
+};
+
+use super::util::expose_value_commitment;
+use bellman::gadgets::boolean;
+
+pub const PUBLIC_KEY_GENERATOR: SubgroupPoint = SubgroupPoint::from_raw_unchecked(
```
rohanjadvani, line 20: Can we move this to [`constants.rs`](https://github.com/iron-fish/ironfish/blob/network-reset/ironfish-zkp/src/constants.rs)?

mat-if (approved):

PR #2586, opened 2022-11-15
Add rewind to chain processor

danield9tqh:
## Summary
ChainProcessor was failing to update correctly if the head was behind the chain head

## Testing Plan
Unit tests and local testing

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/chainProcessor.ts
+++ b/ironfish/src/chainProcessor.ts
@@ -75,6 +75,9 @@ export class ChainProcessor {
 
     const { fork } = await this.chain.findFork(head, chainHead)
 
+    // All cases can be handled by rewinding to the fork point
+    // and then fast-forwarding to the destination. In cases where `head` and `chainHead`
+    // are on the same linear chain, either rewind or fast-forward will just be a no-op
     const iterBackwards = this.chain.iterateFrom(head, fork, undefined, false)
 
     for await (const remove of iterBackwards) {
--- a/ironfish/src/chainProcessor.ts
+++ b/ironfish/src/chainProcessor.ts
@@ -96,24 +96,6 @@ export class ChainProcessor {
       }
     }
 
-    if (chainHead.sequence < head.sequence) {
-      const iterBackwards = this.chain.iterateFrom(head, chainHead, undefined, false)
-
-      for await (const remove of iterBackwards) {
-        if (signal?.aborted) {
-          return { hashChanged: !oldHash || !this.hash.equals(oldHash) }
-        }
-
-        if (remove.hash.equals(chainHead.hash)) {
-          continue
-        }
-
-        await this.remove(remove)
-        this.hash = remove.hash
-        this.sequence = remove.sequence
-      }
-    }
-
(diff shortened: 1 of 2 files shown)


ironfish/src/chainProcessor.ts
```diff
@@ -96,20 +96,40 @@ export class ChainProcessor {
       }
     }
 
-    const iter = this.chain.iterateTo(fork, chainHead, undefined, false)
+    if (chainHead.sequence < head.sequence) {
```
NullSoldier: Why cant you just change the `isLinear` above to this condition?

hughy: `chainHead` could still be ahead of `head` and you'd need to do the block above. Suppose `chainHead` is chain A and `head` is chain B:
```
A1 -> A2 -> A3 -> A4
               -> B3
```
`head` needs to go backwards from B3 -> A2, and then forwards from A2 -> A3 -> A4

hughy: I think we might be able to change it to `if (fork.sequence < head.sequence)`. The block below seems to assume `fork === head`

NullSoldier: Sorry, yes. That's what I mean. fork.sequence < head.sequence

ironfish/src/chainProcessor.ts
```diff
@@ -73,32 +73,30 @@ export class ChainProcessor {
...
-        await this.remove(remove)
-        this.hash = remove.previousBlockHash
-        this.sequence = remove.sequence - 1
+      if (remove.hash.equals(fork.hash)) {
+        continue
       }
+
+      await this.remove(remove)
+      this.hash = remove.hash
+      this.sequence = remove.sequence
```
hughy, line 94: Should these be setting it to the block before remove?

ironfish/src/chainProcessor.ts
danield9tqh: Hugh was saying we can always just rewind to the fork then fast-forward to the destination and that should handle all cases. Seems sound. Added a comment here for explanation

ironfish/src/chainProcessor.ts
danield9tqh: Updated

hughy (approved):

ironfish/src/chainProcessor.ts
```diff
@@ -73,32 +73,30 @@ export class ChainProcessor {
       `Chain processor head not found in chain: ${this.hash.toString('hex')}`,
     )
 
-    const { fork, isLinear } = await this.chain.findFork(head, chainHead)
-    if (!fork) {
-      return { hashChanged: false }
-    }
+    const { fork } = await this.chain.findFork(head, chainHead)
```
NullSoldier, line 76: What is the fork here in the case that chainHead is ahead? This is ok?

danield9tqh: Fork will return the smaller of the two if they are on the same linear branch

PR #2588, opened 2022-11-15
Skip processing the chain if accounts is empty

dguenther:
## Summary

If we don't have any accounts, we should be able to stop the wallet from doing some chain processing.

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):

NullSoldier:
This was a regression from https://github.com/iron-fish/ironfish/pull/1324

PR #2592, opened 2022-11-16
Remove invalid transaction from mempool

hairtail:
## Summary
There are too many invalid transactions in mempool, which actually makes some invalid blocks for double spend and slows down the `blockTemplateStream`. The mempool information like nullifiers will be cleared once restart. This PR add some verify logic with blockchain nullifiers before add a transaction to mempool, which remove invalid transactions from mempool and makes mining more efficient. 
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/memPool/memPool.ts
+++ b/ironfish/src/memPool/memPool.ts
@@ -125,7 +125,7 @@ export class MemPool {
   /**
    * Accepts a transaction from the network
    */
-  acceptTransaction(transaction: Transaction): boolean {
+  async acceptTransaction(transaction: Transaction): Promise<boolean> {
     const hash = transaction.hash().toString('hex')
     const sequence = transaction.expirationSequence()
     if (this.exists(transaction.hash())) {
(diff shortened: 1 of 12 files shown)


ironfish/src/memPool/memPool.ts
```diff
@@ -142,6 +142,14 @@ export class MemPool {
       return false
     }
 
+    const { valid: isValid, reason } = await this.chain.verifier.verifyTransactionSpends(
+      transaction,
+    )
+    if (!isValid) {
+      this.logger.debug(`Invalid transaction '${hash}': ${reason ?? ''}`)
```
ygao76, line 149: can you use logger.info instead of debug, so the message will expose properly.

ygao76:
This is not the cause of double spend issue we recently experienced. What issue are you trying to fix it here?

NullSoldier:
We're not going to accept this PR right now

1. We're almost done with Phase 2
2. There's no evidence posted on what affect this will have in the network
3. Tests are failing
4. No new tests are added

I would prefer an actual reasoning posted along with the change, including its drawbacks. What negative affect will this have? I think if were going to change core logic, I think we should have some due dilligence on the part of the author. If we randomly change our behaviour then we're guarnateed not to get into a better state. What about TX that can be valid after a re-org? These won't let those in now.

PR #2595, opened 2022-11-16
avoids re-syncing expired transactions in wallet

hughy:
## Summary

if we receive a transaction message from a peer for a transaction that has already expired we will sync that transaction and add its notes to the wallet. since the transaction record in the database has not changed we do not add the transaction to pendingTransactionHashes and it never re-expires in the wallet. this results in the balance updating to include notes that do not exist.

updates syncTransaction to check whether an off-chain transaction is expired before syncing it.

if the transaction has a blockHash but would other be expired, such as when syncing during a re-org, we still sync the transaction.

adds tests of both cases.

## Testing Plan

updates unit tests

## Breaking Change

```
[ ] Yes
```


danield9tqh:
This is a little hacky but discussed with @hughy and there doesn't seem to be a much better way to do this without a larger refactor or database change. Worst case scenario I can think of is if we receive and expired transaction and don't sync it then the chain re-orgs so that the transaction is no longer expired and we've missed it. This cannot cause any database corruptions it's similar to just missing a gossiped transaction which is possible anyway.

This logic we'll want to revist after Phase 2 testnet when we do Wallet 2.1 refactor

danield9tqh (approved):

PR #2598, opened 2022-11-16
Encode hard fork into invalid blocks

NullSoldier:
## Summary

This should prevent some re-orging. And we should delete this in the network-reset branch.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -175,6 +175,12 @@ export class Blockchain {
     this.autoSeed = options.autoSeed ?? true
     this.consensus = new TestnetParameters()
 
+    // TODO: Remove this during network reset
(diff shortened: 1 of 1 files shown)


ironfish/src/blockchain/blockchain.ts
```diff
@@ -175,6 +175,12 @@ export class Blockchain {
     this.autoSeed = options.autoSeed ?? true
     this.consensus = new TestnetParameters()
 
+    // TODO: Remove this during network reset
+    this.invalid.set(
```
hughy, line 179: Since `invalid` is an LRU, do we need to worry about this getting evicted?

NullSoldier: it could, eventually, yeah.

hughy (approved):

PR #2602, opened 2022-11-17
remove custom eror message

aditya-manit:
## Summary

Remove custom error message to show config key

Yup automatically gives an appropriate error message with key name and error message, if a custom error message is defined - it will become our responsibility to add the key name in that custom error message.

Here a custom error message was defined without a key name, so in case on error in key value. We were just getting:

`Error: Invalid URL` -> with no idea which key it is for

Removing the custom message which automatically fix the issue
`Error: explorerBlocksUrl must be a valid URL
`
## Testing Plan

Tested on local

## Breaking Change

Is this a breaking change? If yes, add notes below on why this is breaking and
what additional work is required, if any.

No

PR #2606, opened 2022-11-18
fix config validator

aditya-manit:
## Summary

Remove custom error message to show config key

Yup automatically gives an appropriate error message with key name and error message, if a custom error message is defined - it will become our responsibility to add the key name in that custom error message.

Here a custom error message was defined without a key name, so in case on error in key value. We were just getting:

`Error: Invalid URL` -> with no idea which key it is for

Removing the custom message which automatically fix the issue
`Error: explorerBlocksUrl must be a valid URL
`
## Testing Plan

Tested on local

## Breaking Change

No

### DIFF
--- a/ironfish/src/fileStores/config.ts
+++ b/ironfish/src/fileStores/config.ts
@@ -259,7 +259,7 @@ export const noWhitespaceBegEnd = yup
   .string()
   .matches(reNoWhitespaceBegEnd, 'Path should not contain leading or trailing whitespace.')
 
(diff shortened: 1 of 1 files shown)


NullSoldier:
Hey, why did you close your old PR? https://github.com/iron-fish/ironfish/pull/2602

You can just use `git pull -r origin staging` to rebase it next time

NullSoldier (approved):

aditya-manit:
Yeah, but actually had to clone other branches in my fork - the only branch in my fork was master

PR #2610, opened 2022-11-18
Use the assigned schema validation results

NullSoldier:
## Summary

If you use any transformers in yup, we don't use the return value. So we can't use things like default(), or trim().

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/fileStores/keyStore.ts
+++ b/ironfish/src/fileStores/keyStore.ts
@@ -54,9 +54,7 @@ export class KeyStore<TSchema extends Record<string, unknown>> {
         throw new Error(error.message)
       }
 
-      if (result) {
-        Object.assign(data, result)
-      }
+      Object.assign(data, result)
     }
 
     this.keysLoaded.clear()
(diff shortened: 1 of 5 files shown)


ironfish/src/fileStores/keyStore.ts
```diff
@@ -48,10 +48,15 @@ export class KeyStore<TSchema extends Record<string, unknown>> {
...
     // Validate file store if we have a schema
     if (this.schema) {
-      const { error } = await YupUtils.tryValidate(this.schema, data)
+      const { error, result } = await YupUtils.tryValidate(this.schema, data)
+
       if (error) {
         throw new Error(error.message)
       }
+
+      if (result) {
```
danield9tqh, line 57: `result` and `error` are mutually exclusive on this return type right? Do we need to do this check here or can we just assign?

NullSoldier: Good find, I didn't see that it was typed like that.

ironfish/src/fileStores/keyStore.ts
```diff
@@ -48,10 +48,15 @@ export class KeyStore<TSchema extends Record<string, unknown>> {
...
+        Object.assign(data, result)
```
danield9tqh, line 58: Also I think we might need to check if `data === null` before this. If the file doesn't exist.

NullSoldier: added check

danield9tqh (approved):

Issue #2618, opened 2022-11-19
Feature: Validation and Rollback on config:edit


aditya-manit:
### Description

Current config:edit command has no validations, you can literally put anything into the config and save it.
It will just show

```
ironfish config:edit
Editing /Users/aditya.verma/.ironfish/config.json
```

When you do any other operations like config:show only then you know the config value that you had set wasn't right

```
 ironfish config:show
    ParseJsonError: Parsing config.json Failed
    Unexpected token "a" (0x61) in JSON at position 163 while parsing near "... \"databaseMigrate\": afalse\n}\n" in config.json 

      3 |     "editor": "/usr/bin/vim",
      4 |     "blockGraffiti": "9999Kingsuper",
    > 5 |     "databaseMigrate": afalse
        |                        ^
      6 | }
      7 |
```

To make the matter worse if you now try to edit config again it won't let you

```
ironfish config:edit
    ParseJsonError: Parsing config.json Failed
    Unexpected token "a" (0x61) in JSON at position 163 while parsing near "... \"databaseMigrate\": afalse\n}\n" in config.json 

      3 |     "editor": "/usr/bin/vim",
      4 |     "blockGraffiti": "9999Kingsuper",
    > 5 |     "databaseMigrate": afalse
        |                        ^
      6 | }
      7 |
```

Now you have to manually go into ~/.ironfish directory and edit config.json

## Summary
So to summarise:
1. config:edit gives no warning of invalid json or check any validation type errors
2. When you come to know about the error, it won't let you fix it

It's kinda of bug or miss, as other config operations have validation checks on them, I guess we are using an external editor to edit a file, things are a little trickier in this case and therefore it got missed.

aditya-manit:
Solving this in https://github.com/iron-fish/ironfish/pull/2619

PR #2619, opened 2022-11-19
add validation and rollback in config:edit command

aditya-manit:
## Summary
Please check https://github.com/iron-fish/ironfish/issues/2618 for description of bug / missing feature

After changes, it is not a valid json it will display the error message and rollback

```
yarn start config:edit
yarn run v1.22.19
$ yarn build && yarn start:js config:edit
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc --inspect=:0 --inspect-publish-uid=http --enable-source-maps bin/run config:edit
Editing /Users/aditya.verma/.ironfish/config.json
ParseJsonError: Parsing undefined Failed
Unexpected token "a" (0x61) in JSON at position 163 while parsing near "... \"databaseMigrate\": afalse\n}\n" 

  3 |     "editor": "/usr/bin/vim",
  4 |     "blockGraffiti": "9999Kingsuper",
> 5 |     "databaseMigrate": afalse
    |                        ^
  6 | }
  7 |
 Not a Valid JSON, Aborting changes !!
```

If its a valid json, next it run all the yup checks for type validation of config key and same goes here: if any validation check fails, it will rollback

```
yarn start config:edit
yarn run v1.22.19
$ yarn build && yarn start:js config:edit
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc --inspect=:0 --inspect-publish-uid=http --enable-source-maps bin/run config:edit
Editing /Users/aditya.verma/.ironfish/config.json
{
    "explorerBlocksUrl": "hsdbshttps://explorer.ironfish.network/blocks/",
    "editor": "/usr/bin/vim",
    "blockGraffiti": "9999Kingsuper",
    "databaseMigrate": false
}

ValidationError: explorerBlocksUrl must be a valid URL, Aborting changes !!
```
The problem of it not letting it fix it is also solved since the checks are now being run at the end

## Testing Plan

Tested on local

## Breaking Change

No

----
I am trying to get one of my PR counted as a big one, hopefully this will be it, took a lot of time to understand validations and directory structure and all that stuff

main-qimg-ee225d42d464aa7155add23140df0e1f-lq

### DIFF
--- a/ironfish-cli/src/commands/config/edit.ts
+++ b/ironfish-cli/src/commands/config/edit.ts
@@ -79,6 +79,7 @@ export class EditCommand extends IronfishCommand {
     const filePath = path.join(folderPath, DEFAULT_CONFIG_NAME)
 
     await writeFileAsync(filePath, output)
+    const existingContent = await readFileAsync(filePath, { encoding: 'utf8' })
     const code = await launchEditor(filePath, this.sdk.config)
 
     if (code !== 0) {
@@ -86,10 +87,26 @@ export class EditCommand extends IronfishCommand {
     }
 
     const content = await readFileAsync(filePath, { encoding: 'utf8' })
-    const config = JSONUtils.parse<Record<string, unknown>>(content)
 
-    await client.uploadConfig({ config })
-    this.log('Uploaded config successfully.')
-    this.exit(0)
+    try {
+      const config = JSONUtils.parse<Record<string, unknown>>(content)
+      this.log(jsonColorizer(content))
+      await ConfigOptionsSchema.validate(config, { strict: true })
+      await client.uploadConfig({ config })
+      this.log('Uploaded config successfully.')
+      this.exit(0)
+    } catch (err) {
+      if (err instanceof ParseJsonError) {
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+        this.log(`${err} Not a Valid JSON, Aborting changes !!`)
+        await writeFileAsync(filePath, existingContent)
+        this.exit(1)
+      } else {
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+        this.log(`${err}, Aborting changes !!`)
+        await writeFileAsync(filePath, existingContent)
+        this.exit(1)
+      }
+    }
   }
 }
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/config/edit.ts
```diff
@@ -48,17 +79,34 @@ export class EditCommand extends IronfishCommand {
...
+        this.log(`${err} Not a Valid JSON, Aborting changes !!`)
+        await writeFileAsync(filePath, existingContent)
+        this.exit(1)
+      } else {
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+        this.log(`${err}, Aborting changes !!`)
+        await writeFileAsync(filePath, existingContent)
+        this.exit(1)
+      }
+    }
```
NullSoldier, line 110: I don't think you read any of the code at all based on these changes.

1. There's already validation in the uploadConfig()
2. Why is this writing with existingContent to file path? It's a temporary file allocated in the os.tmpdir(). why would you need to revert it?
3. You don't understand how template strings work if you're needing to add `eslint-disable-next-line`. You're just doing this to ignore the error its telling you without fixing the actual error. You can't and should not pass objects into template expressions.
4. All of this catch logic is duplicating the logic in the base command
5. There's no need to catch the errors at all here as you're du;plicating

NullSoldier:
I'm closing this PR for being extremely low code quality. The issues in the PR have demonstrated the author clearly has not read any of the surrounding code while modifying it. Accepting code of this quality is a danger to the security, and hygiene of the code base.

NullSoldier:
I see you are trying to get your code submitted. Can you discuss product changes like this in our discord? This isn't just a fix, but changing the behaviour of this command. We're willing to discuss them first but you'll need to engage with us before hand.

hairtail:
It's hard to get a big PR, and so does Medium. I have about 30 PRs merged, all small.😅😅

hairtail:
This is a small PR from my friend, https://github.com/iron-fish/ironfish/pull/2368
Yeah, some really hard work should be done to get a medium and big.

ironfish-cli/src/commands/config/edit.ts
aditya-manit: 1. I have seen the check in upload config, but does that stops you from saving incorrect value in config ? No

If you are open to discuss and suggest improvements - I can fix the code and maybe we can merge this ?

aditya-manit:
The checks are in upload config, not while saving the config.json file -> which is the issue because the user can put incorrect values in config.json. This is exactly what I tried to mentioned here: https://github.com/iron-fish/ironfish/issues/2618

Also I think upload config is only called when --remote flag is passed, what about the other block of code ?

PR #2635, opened 2022-11-22
chore: database paths to one folder

jowparks:
## Summary
NETWORK RESET PR

Moving database paths to consistent subdirectory as delineated in task:
https://airtable.com/appIXmGgVqP9QdbCf/tblAWSTOQbTq6vh6X/viwHAfsDVPWLk63IW/recswRXAfNGoulBEt?blocks=hide
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/fileStores/config.ts
+++ b/ironfish/src/fileStores/config.ts
@@ -7,9 +7,10 @@ import { YupUtils } from '../utils'
 import { KeyStore } from './keyStore'
 
 export const DEFAULT_CONFIG_NAME = 'config.json'
-export const DEFAULT_DATABASE_NAME = 'default'
+export const DEFAULT_DATABASE_NAME = 'chain'
+export const DEFAULT_INDEX_DATABASE_NAME = 'mined'
 export const DEFAULT_DATA_DIR = '~/.ironfish'
-export const DEFAULT_WALLET_NAME = 'default'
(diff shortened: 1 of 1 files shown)


ironfish/src/fileStores/config.ts
```diff
@@ -337,15 +337,15 @@ export class Config extends KeyStore<ConfigOptions> {
...
   }
 
   get accountDatabasePath(): string {
-    return this.files.join(this.storage.dataDir, 'accounts', this.get('accountName'))
+    return this.files.join(this.storage.dataDir, 'databases', 'wallet', this.get('accountName'))
   }
 
   get indexDatabasePath(): string {
-    return this.files.join(this.storage.dataDir, 'indexes', this.get('databaseName'))
+    return this.files.join(this.storage.dataDir, 'databases', 'mined', this.get('databaseName'))
```
NullSoldier, line 348: We don't need the extra folder anymore now too.

We can remove the middle folder and just use the DB name since that becomes a folder

NullSoldier: also this should have it own DB name

jowparks: We can't do this with `accounts` unless we want to rename the default for the chain database to be not `default` or the default for accounts name to not be `default`, there is a collision. I left the folder path in there for accounts

dguenther (approved):

PR #2638, opened 2022-11-22
uses config tempDir for backup files

hughy:
## Summary

the backup command writes files to a temporary directory before uploading to S3. instead of generating a new directory on each command invocation we can use the tempDir specified by the config, which is located within the datadir.

the config tempDir makes backup consistent with other ironfish commands and makes the write location for backup consistent across invocations.

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):
This is great, it will leave the volumes clean on our docker images so they don't grow in size.

PR #2639, opened 2022-11-23
Add support for custom network definitions

mirayadav:
## Summary

- Changes consensus parameters from hardcoded constants to customizable parameters
- Standardizes an Iron Fish network definition format to include a unique network ID, genesis block, bootstrap nodes, and consensus parameters
- Allows a node to connect to a custom network from a specified network definition JSON file or to an official Iron Fish network with a specified network ID
- Adds a RPC endpoint that returns the consensus parameters of the network a node is connected to

## Testing Plan

Added and updated relevant tests.

## Breaking Change

This is not a breaking change as the default logic (when no network ID or custom network definition file is specified) uses an internal Testnet Phase 2 network definition. 

```
[ ] Yes
```


ironfish/src/node.ts
```diff
@@ -226,13 +234,35 @@ export class IronfishNode {
...
+    } else if (config.get('networkId') === 1) {
+      networkDefinitionJSON = TESTNET_PHASE_2
+    }
+
+    const networkDefinition = IJSON.parse(networkDefinitionJSON) as {
+      id: number
+      bootstrapNodes: string[]
+      genesis: SerializedBlock
+      consensus: ConsensusParameters
+    }
```
dguenther, line 254: It'd be nice to have a `yup` schema for this to validate the incoming type like we do for the config: https://github.com/iron-fish/ironfish/blob/22e9f8ce5cded0ce32ecdce8750869111638cb23/ironfish/src/fileStores/config.ts#L250

That way we get an error up front if the input is incorrectly formatted rather than somewhere else in the node. I realize this might be a bit inconvenient to do since there are several fields in the genesis block data + consensus params, but still think it'd be worthwhile.

ironfish/src/networkDefinitions.ts
```diff
@@ -0,0 +1,123 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+export const TESTING = `
```
dguenther, line 5: I think one of these networks should be `dev` or `testing`, and use the genesis block in `genesisBlockDev.ts`. It can have an empty array for bootstrap nodes. That genesis block is easily mineable, so it should make it easier for developers to start up a local network of nodes on their machine.

ironfish-cli/src/commands/chain/genesisblock.ts
```diff
@@ -79,7 +78,7 @@ export default class GenesisBlockCommand extends IronfishCommand {
       allocations: [
         {
           publicAddress: account.publicAddress,
-          amount: CurrencyUtils.decodeIron(GENESIS_SUPPLY_IN_IRON),
+          amount: CurrencyUtils.decodeIron(node.chain.consensus.parameters.genesisSupplyInIron),
```
dguenther, line 81: I think what this should probably do eventually is take `genesisSupplyInIron` as a flag to the command, then output a network definition file. Right now, it looks like there's sort of a chicken-egg problem where you need a network definition file (including a genesis block) to generate a genesis block.

I don't think this has to change in this PR though, it seems like it'll still work to assemble a new genesis block.

ironfish/src/testUtilities/nodeTest.ts
```diff
@@ -74,7 +74,7 @@ export class NodeTest {
 
     const sdk = await IronfishSdk.init({ dataDir, strategyClass })
 
-    sdk.config.setOverride('bootstrapNodes', [''])
```
dguenther, line 77: We might want to keep the bootstrap nodes empty here -- I haven't debugged into it to see if it actually happens, but the tests might try to connect to the bootstrap node otherwise

ironfish/src/networkDefinitions.ts
```diff
@@ -0,0 +1,123 @@
...
+        }
+        ]
+    },
+    "consensus": {
+        "genesisBlockPrevious": {
+            "type": "Buffer",
+            "data": "0000000000000000000000000000000000000000000000000000000000000000"
+        },
+        "genesisBlockSequence": 1,
```
dguenther, line 54: I realize this is after the fact, but we could probably just make this a constant in the source at some point -- A lot of the changes involve threading this value around, and I can't think of a reason it would be useful to change it to a value other than 1.

Not necessary for this PR though since you've already done the work!

ironfish/src/networkDefinitions.ts
```diff
@@ -0,0 +1,123 @@
...
+        }
+        ]
+    },
+    "consensus": {
+        "genesisBlockPrevious": {
```
dguenther, line 50: Same comment as on the genesis block sequence -- we could probably just make this a constant in the code, since I don't see a reason it would need to be configured, but not necessary for this PR.

ironfish/src/network/utils/serializers.ts
```diff
@@ -11,6 +11,8 @@ import { BigIntUtils } from '../../utils/bigint'
 
 export const MINERS_FEE_TRANSACTION_SIZE_BYTES = 562
 
+export const GRAFFITI_SIZE = 32
```
dguenther, line 14: This is probably used in a few places outside of the network code, so I'm curious if there's a better place to put this -- maybe exporting it from the block or block header file makes sense? Or creating a new block/blockchain constants file.

ironfish/src/node.ts
```diff
@@ -226,13 +234,35 @@ export class IronfishNode {
...
+
+    const bootstrapNodes =
+      config.get('bootstrapNodes').length === 0
```
dguenther, line 257: We'll probably need to add an `isSet` function to `config` for this. Right now, the only way to start a node up with 0 bootstrap nodes is by passing an empty list to the config flag, and that wouldn't work with this check.

ironfish/src/network/version.ts
```diff
@@ -4,3 +4,6 @@
 
 export const VERSION_PROTOCOL = 18
 export const VERSION_PROTOCOL_MIN = 18
+
+export const MAX_REQUESTED_BLOCKS = 50
+export const MAX_MESSAGE_SIZE = 256 * 1024 * 1024
```
dguenther, line 9: Not related to your PR at all, but we should really adjust this to something reasonable, hah

ironfish/src/network/utils/serializers.ts
mirayadav: Would it make sense to export it from `src/utils/graffiti.ts` or should I avoid doing that since it's a utils file?

ironfish/src/consensus/consensus.ts
```diff
@@ -92,14 +71,18 @@ export class ConsensusParameters {
...
 
-export class TestnetParameters extends ConsensusParameters {
-  constructor() {
-    super()
+export class TestnetConsensus extends Consensus {
+  constructor(parameters: ConsensusParameters) {
+    super(parameters)
     this.V1_DOUBLE_SPEND = 204000
     this.V2_MAX_BLOCK_SIZE = 255000
     this.V3_DISABLE_MINING_REWARD = 279900
```
dguenther, line 88: Also doesn't need to be done in this PR, but I wonder if these should be consensus params specified in the network definition file as well.

ironfish/src/fileStores/config.ts
```diff
@@ -245,6 +244,16 @@ export type ConfigOptions = {
...
   feeEstimatorPercentileMedium: number
   feeEstimatorPercentileHigh: number
+
+  /**
+   * Network ID of an official Iron Fish network
+   */
+  networkId: number
+
+  /**
+   * JSON file containing the network definition of a custom network
```
dguenther, line 254: Maybe "Path to a JSON file", not an important change though

ironfish-cli/src/commands/start.ts
```diff
@@ -100,6 +100,18 @@ export default class Start extends IronfishCommand {
       allowNo: true,
       description: 'Run migrations when an upgrade is required',
     }),
+    // TODO: Update description with a list of network IDs for mainet and official testnets.
```
dguenther, line 103:
```suggestion
    // TODO: Update description with a list of network IDs for mainnet and official testnets.
```

I'm actually okay just dropping this TODO as well

ironfish-cli/src/commands/start.ts
```diff
@@ -194,7 +220,7 @@ export default class Start extends IronfishCommand {
     this.log(`Peer Identity ${node.peerNetwork.localPeer.publicIdentity}`)
     this.log(`Peer Agent    ${node.peerNetwork.localPeer.agent}`)
     this.log(`Peer Port     ${peerPort}`)
-    this.log(`Bootstrap     ${bootstraps.join(',') || 'NONE'}`)
+    this.log(`Bootstrap     ${bootstraps.join(',') || 'FROM NETWORK DEFINITION'}`)
```
dguenther, line 223: Oh, are the bootstrap nodes not loaded at this point? Or could we pull them from the node somehow?

ironfish/src/consensus/consensus.ts
mirayadav: I'm going to figure out how we want to deal with consensus upgrades in a separate PR (part of the Consensus Upgrades project), so I've left these out of the consensus params for now.

ironfish-cli/src/commands/chain/genesisblock.ts
mirayadav: Ok I added a task for fixing this command

ironfish/src/networkDefinitions.ts
mirayadav: Added - however the genesis block in `genesisBlockDev.ts` has not been updated with the new `transactionCommitment` field in its header

ironfish/src/node.ts
mirayadav: Added `yup` validation

ironfish/src/testUtilities/nodeTest.ts
mirayadav: Added back

ironfish/src/networkDefinitions.ts
mirayadav: Made into a constant

ironfish/src/networkDefinitions.ts
mirayadav: Made into a constant

ironfish/src/network/utils/serializers.ts
mirayadav: Moved to block file along with the above two constants

ironfish/src/node.ts
mirayadav: Added `isBootstrapNodesSet` to `config`

ironfish/src/strategy.ts
```diff
@@ -34,7 +30,7 @@ export class Strategy {
    * Annual coin issuance from mining goes down every year. Year is defined here by the
    * number of blocks (IRON_FISH_YEAR_IN_BLOCKS)
```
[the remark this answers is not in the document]
mirayadav: Updated

ironfish/src/network/version.ts
mirayadav: Yup!

ironfish/src/fileStores/config.ts
mirayadav: Changed

ironfish-cli/src/commands/start.ts
mirayadav: Removed

ironfish-cli/src/commands/chain/genesisblock.ts
mirayadav: Added a `genesisSupplyInIron` flag to the command. However, as we discussed I'm going to keep the output as a genesis block and not a network definition file. Let me know if you'd still like me to change this to output a network definition file.

ironfish/src/fileStores/config.ts
```diff
@@ -338,9 +349,13 @@ export class Config extends KeyStore<ConfigOptions> {
     return this.files.join(this.storage.dataDir, 'temp')
   }
 
+  isBootstrapNodesSet(): boolean {
+    return this.keysLoaded.has('bootstrapNodes') || this.overrides.bootstrapNodes?.length !== 0
```
dguenther, line 353: This could use a test. The KeyStore does some prototypal inheritance with defaults+loaded+overrides, so you'll want to double-check that you're only checking the `overrides` object (and not its parents) for whether bootstrapNodes is set.

ironfish/src/networkDefinition.ts
```diff
@@ -0,0 +1,55 @@
...
+              .object({
+                commitment: yup.string().defined(),
+                size: yup.number().integer().defined(),
+              })
+              .defined(),
+            transactionCommitment: yup.mixed<Buffer>().defined(),
+            target: yup.string().defined(),
+            randomness: yup.string().defined(),
+            timestamp: yup.number().integer().defined(),
+            noteSize: yup.mixed<number | null>(),
```
dguenther, line 36: This should be something like `.number().integer().nullable()`, I don't think `mixed` does any actual validation on the values

mirayadav: Changed

ironfish/src/fileStores/config.ts
mirayadav: Added tests for this function

dguenther: Sorry, I didn't see this before, but this should check if `bootstrapNodes` is set in `overrides`, not just whether it has values of 0, so that we can allow people to pass an empty array. You can test this by changing the default config to `bootstrapNodes: ['']`, then making sure the `config.test.ts` tests still pass.

You can check if just the object (and not its prototype chain) has a property. It's just a bit inconvenient in JS: `Object.prototype.hasOwnProperty.call(this.overrides, 'bootstrapNodes')`

mirayadav: Updated to check this, I kept the default as `[]` but changed the test to test the case when people pass an empty array.

dguenther (approved):

PR #2640, opened 2022-11-23
Add back wallet unit tests to network-reset

ygao76:
the useAccountFixture used createAccount to create the account in a fixture, but importAccount when deserializing the fixture. one problem with this is that createAccount and importAccount set the head hash stored for an account differently. createAccount sets the head hash to the head of the chainProcessor -- we don't need to scan earlier blocks if we just created the account. importAccount sets the head hash to null -- we don't know when this account was created, so we need to scan the whole chain.

this leads to different behavior in tests when generating the fixtures vs. when reusing them.

these changes update useAccountFixture to set the head hash to the chainProcessor's hash after adding the account so that the head hash will match what it would have if we had created the account.

there is another problem with the fixture: we use UUIDs when creating and importing accounts. so if we import an account from its data, the account's id in the wallet will be different. this _should_ be ok, since tests should not depend on the UUID of the account.

- updates useAccountFixture to set head hash to chainProcessor hash during deserialization
- fixes tests that depended on head hash being set to null on creation

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/testUtilities/fixtures.ts
+++ b/ironfish/src/testUtilities/fixtures.ts
@@ -131,7 +131,11 @@ export async function useAccountFixture(
     },
 
     deserialize: async (accountData: AccountValue): Promise<Account> => {
-      return wallet.importAccount(accountData)
+      const account = await wallet.importAccount(accountData)
+
+      await wallet.updateHeadHash(account, wallet.chainProcessor.hash)
+
+      return account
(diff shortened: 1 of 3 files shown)


hughy:
IT looks like we might need to regenerate the fixtures for these since these commits went into staging and still have diversifier in the public address

PR #2641, opened 2022-11-23
feat: remove crypto wasm library chacha

jowparks:
## Summary
Removes dependency on `rust-crypto-wasm = "0.3.1" `
and adds `chacha20poly1305 = "0.9.0"`

Rest of PR is to work around those updates
## Breaking Change

```
[x] Yes
```


ironfish-rust/src/note.rs
```diff
@@ -209,10 +209,13 @@ impl<'a> Note {
...
         bytes_to_encrypt[(SCALAR_SIZE + AMOUNT_VALUE_SIZE)..].copy_from_slice(&self.memo.0[..]);
-        let mut encrypted_bytes = [0; ENCRYPTED_NOTE_SIZE + aead::MAC_SIZE];
-        aead::encrypt(shared_secret, &bytes_to_encrypt, &mut encrypted_bytes);
-
-        encrypted_bytes
+        let encrypted_bytes = aead::encrypt(shared_secret, &bytes_to_encrypt)?;
+        assert_eq!(encrypted_bytes.len(), ENCRYPTED_NOTE_SIZE + aead::MAC_SIZE);
+        
+        Ok(encrypted_bytes
+            .try_into()
```
jowparks, line 216: What type should this error be? Should I create a type? The type is going from `Vec<u8>` to `&[u8; 32]`

ironfish-rust/src/serializing.rs
```diff
@@ -90,40 +88,25 @@ pub(crate) mod aead {
...
-            &mut encrypted_output[..plaintext.len()],
-            &mut tag,
-        );
-        encrypted_output[plaintext.len()..].clone_from_slice(&tag);
+    /// This is just a facade around the ChaCha20Poly1305 struct. The nonce and
+    /// associated data are zeroed.
+    pub(crate) fn encrypt(key: &[u8; 32], plaintext: &[u8]) -> Result<Vec<u8>, IronfishError> {
+        let encryptor = ChaCha20Poly1305::new(Key::from_slice(key));
+        let result = encryptor
+            .encrypt(&Nonce::default(), plaintext)
```
jowparks, line 96: Want to verify that using `&Nonce::default()` is acceptable.

ironfish-rust/src/merkle_note.rs
```diff
@@ -89,8 +89,8 @@ impl MerkleNote {
             &note.commitment_point(),
             public_key,
         );
-        let mut note_encryption_keys = [0; NOTE_ENCRYPTION_KEY_SIZE];
-        aead::encrypt(&encryption_key, &key_bytes, &mut note_encryption_keys);
+        let note_encryption_keys: [u8; NOTE_ENCRYPTION_KEY_SIZE] =
```
jowparks, line 92: I didn't want to add a big refactor in here for returning results rather than unwrapping, I can do that in follow up PR if we want to. Also is there any reason why we would be returning plain objects rather than `Result`s?

ironfish-rust/src/merkle_note.rs
```diff
@@ -89,8 +89,11 @@ impl MerkleNote {
...
```
jowparks, line 92: Not a fan of this unwrapping, but I can fix in subsequent PR to not bloat this one (update results for return types)

dguenther: Sounds good, would definitely be nice to remove the unwraps. We're basically doing the same thing in `note.rs`'s `encrypt` function that we'd want to do here, it looks like

ironfish-rust/src/merkle_note.rs
```diff
@@ -138,7 +141,7 @@ impl MerkleNote {
             value_commitment: value_commitment.commitment().into(),
             note_commitment: note.commitment_point(),
             ephemeral_public_key: (*public_key),
-            encrypted_note,
+            encrypted_note: encrypted_note.unwrap(),
```
dguenther, line 144: We could probably avoid the unwrapping in this one as well, if you want to do that in a follow-up

ironfish-rust/src/serializing.rs
```diff
@@ -90,40 +88,25 @@ pub(crate) mod aead {
...
+            .map_err(|_| IronfishError::InvalidSigningKey)?;
+        Ok(result)
```
mat-if, line 98: We could probably avoid one layer of unwraps if we don't return `Vec<u8>` here. I think we can use [`encrypt/decrypt_in_place_detached`](https://docs.rs/chacha20poly1305/latest/chacha20poly1305/trait.AeadInPlace.html#tymethod.encrypt_in_place_detached) to work with arrays directly. This removes the need to call `try_into()` as a follow-up to these functions

```rust

    pub(crate) fn encrypt<const SIZE: usize>(
        key: &[u8; 32],
        plaintext: &[u8],
    ) -> Result<[u8; SIZE], IronfishError> {
        let mut encrypted_output = [0u8; SIZE];
        encrypted_output[..plaintext.len()].copy_from_slice(plaintext);

        let encryptor = ChaCha20Poly1305::new(Key::from_slice(key));

        let tag = encryptor
            .encrypt_in_place_detached(
                &Nonce::default(),
                &[],
                &mut encrypted_output[..plaintext.len()],
            )
            .map_err(|_| IronfishError::InvalidSigningKey)?;

        encrypted_output[plaintext.len()..].copy_from_slice(&tag);

        Ok(encrypted_output)
    }

    pub(crate) fn decrypt<const SIZE: usize>(
        key: &[u8; 32],
        ciphertext: &[u8],
    ) -> Result<[u8; SIZE], IronfishError> {
        let decryptor = ChaCha20Poly1305::new(Key::from_slice(key));

        let mut plaintext = [0u8; SIZE];
        plaintext.copy_from_slice(&ciphertext[..SIZE]);

        decryptor
            .decrypt_in_place_detached(
                &Nonce::default(),
                &[],
                &mut plaintext,
                ciphertext[SIZE..].into(),
            )
            .map_err(|_| IronfishError::InvalidDecryptionKey)?;

        Ok(plaintext)
    }
```

ironfish-rust/src/serializing.rs
dguenther: The key is a one-time-use key generated for this particular note/output, so this should be acceptable.

ironfish-rust/src/note.rs
dguenther: I think it's fine being generous with error types, maybe like `IronfishError::InvalidEncryptedNoteLength`. I don't think our error types are especially well thought-out so far, so I'm not too concerned with getting it right here

ironfish-rust/src/merkle_note.rs
dguenther: As far as I know, we usually only use `Result`s if we're propagating errors up, so we might not have been doing that before (it doesn't look like the old `crypto-wasm` returned a `Result`, so it probably wasn't necessary for us to return one either)

jowparks: From what I was reading, the result type is idiomatic (functional, strongly typed) rather than unwrap. Takes advantage of the functional nature of rust, which makes me 😄

ironfish-rust/src/note.rs
jowparks: Wow that is cool syntax @mat-if, thanks for the tip!

ironfish-rust/src/note.rs
```diff
@@ -247,8 +247,7 @@ impl<'a> Note {
         index += GENERATOR_SIZE;
 
         bytes_to_encrypt[index..].copy_from_slice(&self.asset_generator.to_bytes());
-        let mut encrypted_bytes = [0; ENCRYPTED_NOTE_SIZE + aead::MAC_SIZE];
-        aead::encrypt(shared_secret, &bytes_to_encrypt, &mut encrypted_bytes);
+        let  encrypted_bytes = aead::encrypt(shared_secret, &bytes_to_encrypt).unwrap();
```
jowparks, line 250: Remove this unwrap in subsequent PR.

ironfish-rust/src/merkle_note.rs
```diff
@@ -199,11 +199,9 @@ impl MerkleNote {
             &self.ephemeral_public_key,
         );
 
-        let mut note_encryption_keys = [0; ENCRYPTED_SHARED_KEY_SIZE];
-        aead::decrypt(
+        let note_encryption_keys = aead::decrypt::<ENCRYPTED_SHARED_KEY_SIZE>(
```
mat-if, line 202: Minor nit, up to you. Can also be written as `let note_encryption_keys: [u8; ENCRYPTED_SHARED_KEY_SIZE] = aead::decrypt(...)?`

ironfish-rust/src/serializing.rs
```diff
@@ -75,57 +73,76 @@ pub(crate) mod aead {
     ///
     /// This is just a facade around the ChaCha20Poly1305 struct. It ignores
     /// nonce and aad and automatically stores the mac tag.
```
mat-if, line 75:
```suggestion
```

ironfish-rust/src/serializing.rs
```diff
@@ -75,57 +73,76 @@ pub(crate) mod aead {
...
+            .map_err(|_| IronfishError::InvalidSigningKey)?;
+        encrypted_output[plaintext.len()..].copy_from_slice(&tag);
+
+        Ok(encrypted_output)
     }
 
     /// Decrypt the encrypted text using the given key and ciphertext, also checking
     /// that the mac tag is correct.
     ///
     /// Returns Ok(()) if the mac matches the decrypted text, Err(()) if not
```
mat-if, line 102:
```suggestion
```

Probably just dont need this line anymore, I think the type definition is clear enough

mat-if (approved):
🙌 looks good!

ironfish-rust/src/merkle_note.rs
jowparks: I like it

ironfish-rust/src/serializing.rs
```diff
@@ -63,69 +63,82 @@ pub fn hex_to_bytes(hex: &str) -> Result<Vec<u8>, IronfishError> {
...
-    pub(crate) fn encrypt(key: &[u8], plaintext: &[u8], encrypted_output: &mut [u8]) {
-        assert_eq!(encrypted_output.len(), plaintext.len() + MAC_SIZE);
-        let mut encryptor = ChaCha20Poly1305::new(key, &[0; 8], &[0; 8]);
-        let mut tag = [0; MAC_SIZE];
-        encryptor.encrypt(
-            plaintext,
-            &mut encrypted_output[..plaintext.len()],
-            &mut tag,
-        );
-        encrypted_output[plaintext.len()..].clone_from_slice(&tag);
```
jowparks, line 87: These two methods are the major changes

dguenther:
This looks good to me -- I'm fairly sure it'll just need fixtures regenerated for the failing tests.

jowparks:
To regenerate all fixtures, instructions by @dguenther :
remove all files
`find . -name "__fixtures__" | xargs rm -rf`

```
if you do that, you’ll want to run yarn test, yarn test:slow, and yarn test:perf in ironfish to regenerate all of them. sometimes they’ll fail due to taking too long, so you can add --testTimeout=600000 to give them a bit more leeway
```

ironfish/package.json
```diff
@@ -56,7 +56,9 @@
     "test:slow": "tsc -b && tsc -b tsconfig.test.json && cross-env TEST_INIT_RUST=true jest --testMatch \"**/*.test.slow.ts\" --testTimeout=60000 --testPathIgnorePatterns",
     "test:perf": "tsc -b && tsc -b tsconfig.test.json && cross-env TEST_INIT_RUST=true jest --testMatch \"**/*.test.perf.ts\" --testTimeout=60000 --testPathIgnorePatterns",
     "test:coverage:html": "tsc -b tsconfig.test.json && jest --coverage --coverage-reporters html --testPathIgnorePatterns",
-    "test:watch": "tsc -b tsconfig.test.json && jest --watch --coverage false"
+    "test:watch": "tsc -b tsconfig.test.json && jest --watch --coverage false",
+    "fixtures:regenerate": "find . -name \"__fixtures__\" | xargs rm -rf && yarn run test; yarn run test:slow; yarn run test:perf"
```
jowparks, line 60: Added this for... future use 🥷

jowparks:
@dguenther I am getting timeout test failures in the `Node.js CI` runner after regenerating the fixtures. Is this expected?

PR #2642, opened 2022-11-23
network-reset: Remove note tree size from block header

dguenther:
## Summary

Removes the note tree size from the network-serialized block and PartialBlock, leaving it on the database-stored block and fixtures. We can always do a migration to remove it from the database-stored block if it's no longer necessary, but we use it in some cases now.

## Breaking Change

```
[X] Yes
```


.github/workflows/rust_ci.yml
```diff
@@ -68,7 +68,7 @@ jobs:
         name: "Clippy check on ironfish-rust"
         with:
           command: clippy
-          args: --manifest-path ironfish-rust/Cargo.toml -- -D warnings
+          args: --manifest-path ironfish-rust/Cargo.toml --all-targets -- -D warnings
```
danield9tqh, line 71: This belong in this PR?

dguenther: Should be gone now!

ironfish/src/blockchain/database/headers.ts
```diff
@@ -18,8 +18,8 @@ export class HeaderEncoding implements IDatabaseEncoding<HeaderValue> {
 
     bw.writeU32(value.header.sequence)
     bw.writeHash(value.header.previousBlockHash)
-    bw.writeHash(value.header.noteCommitment.commitment)
-    bw.writeU32(value.header.noteCommitment.size)
+    bw.writeHash(value.header.noteCommitment)
+    bw.writeU32(value.header.noteSize ?? 0)
```
danield9tqh, line 22: Assert that it is not null here

ironfish/src/blockchain/blockchain.ts
```diff
@@ -1234,7 +1244,7 @@ export class Blockchain {
 
     // If the tree sizes don't match the previous block, we can't verify if the tree
     // sizes on this block are correct
-    const prevNotesSize = prev?.noteCommitment.size || 0
+    const prevNotesSize = prev?.noteSize || 0
```
danield9tqh, line 1247: Do we want to assert that `prev?.noteSize` is not null here?

ironfish/src/blockchain/database/headers.ts
```diff
@@ -40,7 +40,7 @@ export class HeaderEncoding implements IDatabaseEncoding<HeaderValue> {
     const sequence = reader.readU32()
     const previousBlockHash = reader.readHash()
     const noteCommitment = reader.readHash()
-    const noteCommitmentSize = reader.readU32()
+    const noteSize = reader.readU32() || null
```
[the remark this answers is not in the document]
dguenther: Done, added an Assert

ironfish/src/blockchain/database/headers.ts
dguenther: Done

danield9tqh (approved):

PR #2653, opened 2022-11-28
Add transaction commitment to block header

danield9tqh:
## Summary
Full description of changes outlined in this proposal https://coda.io/@danielfish/add-transaction-root-to-block-header

## Testing Plan
Unit tests

## Breaking Change

```
[ ] Yes
```


ironfish/src/blockchain/blockchain.ts
```diff
@@ -958,11 +964,15 @@ export class Blockchain {
 
       graffiti = graffiti ? graffiti : Buffer.alloc(32)
 
+      const transactionHashes = transactions.map((t) => t.hash())
+      const transactionCommitment = transactionMerkleRoot(transactionHashes)
```
dguenther, line 968: We'll want to check this commitment in a Verifier function at some point, so we can detect if the transactions don't match the block header commitment.

ironfish/src/primitives/blockheader.ts
```diff
@@ -68,6 +116,12 @@ export class BlockHeader {
    */
   public nullifierCommitment: { commitment: NullifierHash; size: number }
 
+  /**
+   * Commitment to the set of transactions in this block. Generated by a merkle
+   * tree of transasction hashes which include transaction data + witness/signature data.
```
dguenther, line 121:
```suggestion
   * tree of transaction hashes which include transaction data + witness/signature data.
```

dguenther:
Looks good! I think we decided we want to add in the verifier change as well -- ping me for a re-review once that's in and I'll take a peek.

ironfish/src/blockchain/blockchain.ts
danield9tqh: Added to `verifyBlock`

ironfish/src/consensus/verifier.ts
```diff
@@ -124,6 +124,11 @@ export class Verifier {
       return { valid: false, reason: VerificationResultReason.INVALID_MINERS_FEE }
     }
 
+    const expectedTransactionCommitment = transactionCommitment(block.transactions)
+    if (!expectedTransactionCommitment.equals(block.header.transactionCommitment)) {
+      return { valid: false, reason: VerificationResultReason.INVALID_TRANSACTION_COMMITMENT }
+    }
```
dguenther, line 130: We should do the transaction commitment check before the other transaction-related checks in this function, so that we're sure the transactions are expected for the block before we verify them.

I think in the addBlock process we should also say that if a block has an invalid transaction commitment, we don't mark the block's hash as invalid. We can't know if a peer sent the right transactions for the block or not. That could go in a different PR though

dguenther (changes requested):

ironfish/src/consensus/verifier.ts
danield9tqh: Yea you're right. I was thinking the other way around but we don't want to waste time verifying the transactions

dguenther (approved):

PR #2655, opened 2022-11-29
Improve note lengths in the codebase

ygao76:
## Summary
Create constants to store the hardcoded lengths
Export constants to ts package 

## Testing Plan
existing unit tests

## Breaking Change
No

```
[ ] Yes
```
@mat-if @rohanjadvani


ironfish-rust-nodejs/src/structs/note_encrypted.rs
```diff
@@ -9,20 +9,31 @@ use napi::bindgen_prelude::*;
...
 //  32 value commitment
 //+ 32 note commitment
 //+ 32 ephemeral public key
-//+ 100 encrypted note
-//+ 16 aead MAC
-//+ 64 note encryption keys
+//+ 104 encrypted note
 //+ 16 aead MAC
-//= 264 bytes
+//+ 80 note encryption keys
```
dguenther, line 35: `NOTE_ENCRYPTION_KEY_SIZE` is `ENCRYPTED_SHARED_KEY_SIZE` (64) + `aead::MAC_SIZE` (16). these should stay split out if we're going to split out the MAC size from the encrypted note as well

ironfish-rust-nodejs/src/structs/note_encrypted.rs
```diff
@@ -9,20 +9,31 @@ use napi::bindgen_prelude::*;
...
+//= 296 bytes
```
dguenther, line 36: Thanks for fixing the math 👍

ironfish/src/primitives/noteEncrypted.ts
```diff
@@ -4,13 +4,16 @@
...
+  MAC_LENGTH,
+  NOTE_ENCRYPTION_KEY_LENGTH,
   NoteEncrypted as NativeNoteEncrypted,
 } from '@ironfish/rust-nodejs'
 import bufio from 'bufio'
 import { Serde } from '../serde'
 import { Note } from './note'
 
-const PROOF_LENGTH = 192
+export const PROOF_LENGTH = 192
```
dguenther, line 16: Proofs aren't a property of notes, so I don't think it makes sense to export this constant from `noteEncrypted` -- maybe the Rust code could export it from somewhere general?

dguenther (changes requested):

ironfish-rust-nodejs/src/structs/note_encrypted.rs
```diff
@@ -9,20 +9,31 @@ use napi::bindgen_prelude::*;
...
 
 #[napi]
-pub const ENCRYPTED_NOTE_LENGTH: u32 = 296;
+pub const NOTE_ENCRYPTION_KEY_LENGTH: u32 = NOTE_ENCRYPTION_KEY_SIZE as u32;
+
+#[napi]
+pub const ENCRYPTED_NOTE_PLAINTEXT_LENGTH: u32 = ENCRYPTED_NOTE_SIZE as u32;
+
+#[napi]
+pub const MAC_LENGTH: u32 = 16;
```
mat-if, line 25: We do have this number defined here: https://github.com/iron-fish/ironfish/blob/master/ironfish-rust/src/serializing.rs#L86

Up to you whether you want to move it or anything, though

ygao76: Yeah I saw that, but serializing and aead are not public, so I created a new one. Should I make the serializing and aead public?

ironfish/src/primitives/noteEncrypted.ts
ygao76: ok

ironfish-rust-nodejs/src/structs/note_encrypted.rs
dguenther: It makes sense to me to make those public, I don't think we care too much about whether modules are public vs private in Rust. I'm okay with leaving it there, I don't have any particular ideas for where to move it to right now

ironfish/src/primitives/noteEncrypted.ts
```diff
@@ -39,16 +42,14 @@ export class NoteEncrypted {
...
-    reader.seek(72)
+    reader.seek(ENCRYPTED_NOTE_PLAINTEXT_LENGTH)
     // aead MAC
-    reader.seek(16)
+    reader.seek(MAC_LENGTH)
     // note encryption keys
-    reader.seek(64)
-    // aead MAC
-    reader.seek(16)
+    reader.seek(NOTE_ENCRYPTION_KEY_LENGTH)
```
dguenther, line 49: I think this should stay split out into ENCRYPTED_SHARED_KEY_SIZE (64) + MAC_LENGTH (16), since we also split out the ENCRYPTED_NOTE_PLAINTEXT_LENGTH from the MAC_LENGTH.

ironfish/src/primitives/noteEncrypted.ts
ygao76: Create proof length in output

ironfish-rust-nodejs/src/structs/note_encrypted.rs
ygao76: Reused the mac size defined in aead

ironfish-rust-nodejs/src/structs/note_encrypted.rs
dguenther: Decided to merge both together

dguenther (approved):

PR #2663, opened 2022-11-29
implements 'get' on database in storage layer

hughy:
## Summary

the current implementation of the storage layer uses stores to interact with the database.

implements 'get' on the database to retrieve serialized values

changes the store implementation of get to be a wrapper around db.get that encodes the key and deserializes the value

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/storage/database/database.ts
+++ b/ironfish/src/storage/database/database.ts
@@ -137,6 +137,15 @@ export interface IDatabase {
       SchemaValue<DatabaseSchema>
     >[],
   ): Promise<void>
+
+  /**
+   * Used to get a value from the database at a given key
+
+  * @param key - The key to fetch
+  *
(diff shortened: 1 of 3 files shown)


ironfish/src/storage/database/database.ts
```diff
@@ -137,6 +137,15 @@ export interface IDatabase {
...
   ): Promise<void>
+
+  /**
+   * Used to get a value from the database at a given key
+
+  * @param key - The key to fetch
+  *
+  * @returns resolves with the serialized value if found, or undefined if not found.
+  */
+  get(key: Readonly<SchemaKey<DatabaseSchema>>): Promise<Buffer | undefined>
```
NullSoldier, line 148: this should be Buffer

NullSoldier (approved):

PR #2668, opened 2022-11-30
Fix mining resulting from header changes

dguenther:
## Summary

A few fields weren't updated as a result of 
header changes, so mining isn't working on 
staging. This should fix them up.

## Breaking Change

```
[X] Yes
```

### DIFF
--- a/ironfish/src/mining/utils.ts
+++ b/ironfish/src/mining/utils.ts
@@ -5,8 +5,11 @@
 import bufio from 'bufio'
 import { SerializedBlockTemplate } from '../serde/BlockTemplateSerde'
 
+const MINEABLE_BLOCK_HEADER_SIZE = 224
+export const MINEABLE_BLOCK_HEADER_GRAFFITI_OFFSET = MINEABLE_BLOCK_HEADER_SIZE - 32
+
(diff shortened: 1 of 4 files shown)


ironfish/src/mining/utils.ts
```diff
@@ -5,8 +5,11 @@
 import bufio from 'bufio'
 import { SerializedBlockTemplate } from '../serde/BlockTemplateSerde'
 
+const MINEABLE_BLOCK_HEADER_SIZE = 224
+export const MINEABLE_BLOCK_HEADER_GRAFFITI_OFFSET = MINEABLE_BLOCK_HEADER_SIZE - 32
+
 export function mineableHeaderString(header: SerializedBlockTemplate['header']): Buffer {
-  const bw = bufio.write(240)
+  const bw = bufio.write(MINEABLE_BLOCK_HEADER_SIZE)
```
danield9tqh, line 12: Did we miss changing this in a previous PR and that's why it's changing?

danield9tqh (approved):

ironfish/src/mining/utils.ts
dguenther: Yep, I bet I missed changing it in at least one of mine, if not 2. Same with the graffiti offset, it was 176 but is now 192.

PR #2669, opened 2022-11-30
Remove double spend consensus parameter

danield9tqh:
## Summary
During the Phase 2 testnet we had a double spend bug. A fix was activated at a specified block height to no longer allow double spends. Now that we have the fix we want to remove that parameter for allowing double spends and disallow double spends entirely. In this PR I'm also refactoring block verification to no longer check for nullifiers (double spends) after a block has already been connected. We will only check before we connect a block. This should not remove any safety checks since the nullifier check after connecting a new head block was flawed (this is what caused the original bug) and the nullifier check before connecting was added as the fix.

## Testing Plan
Unit tests

## Breaking Change

```
[ ] Yes
```


ironfish/src/testUtilities/matchers/blockchain.ts
```diff
@@ -50,10 +50,27 @@ async function toAddBlock(self: Blockchain, other: Block): Promise<jest.CustomMa
...
+  // Mock nullifiers to allow creation of a double spend chain
+  const containsMock = jest.spyOn(self.nullifiers, 'contains').mockResolvedValue(false)
+  const result = await self.addBlock(other)
+  containsMock.mockRestore()
+
+  if (!result.isAdded) {
+    return makeResult(false, `Could not add block: ${String(result.reason)}`)
+  }
+
+  return makeResult(true, `Expected to not add block at ${String(other.header.sequence)}`)
```
NullSoldier, line 66: This is doing some very scary things that shouldn't be allowed by our consensus layer and corrupting the test chain DB. What is this used for? We should fix any test so this never gets added here.

ironfish/src/blockchain/blockchain.test.ts
```diff
@@ -759,74 +750,44 @@ describe('Blockchain', () => {
...
     })
   })
 
   it('rejects double spend during reorg', async () => {
-    // G -> A2 -> A3
-    //   -> B2 -> B3* -> A4
+    /**
+     * We do check double spends when connecting forks because we don't rebuild the nullifier
+     * set unless we're adding to the head. If we re-org to a fork that contains a double spend
+     * though, we should catch that
```
NullSoldier, line 770: This sentence is inverted I think.

On a side note, this is a bit confusing for me. There's only 1 condition for checking double spends, and that's when we're connecting blocks. Blocks are only connected to the main chain.

ironfish/src/blockchain/blockchain.test.ts
```diff
@@ -759,74 +750,44 @@ describe('Blockchain', () => {
...
       isAdded: false,
       reason: VerificationResultReason.DOUBLE_SPEND,
     })
   })
 
   it('rejects double spend during reorg', async () => {
-    // G -> A2 -> A3
-    //   -> B2 -> B3* -> A4
+    /**
+     * We do check double spends when connecting forks because we don't rebuild the nullifier
```
NullSoldier, line 768: This sentence is inverted I think.

```suggestion
     * We don't check double spends when connecting forks because we don't rebuild the nullifier
```

ironfish/src/consensus/verifier.ts
```diff
@@ -352,26 +365,30 @@ export class Verifier {
     tx?: IDatabaseTransaction,
```
NullSoldier, line 364: The comment above needs to be updated around nullifiers.

ironfish/src/blockchain/blockchain.test.ts
```diff
@@ -887,6 +851,10 @@ describe('Blockchain', () => {
         reason: VerificationResultReason.DOUBLE_SPEND,
       })
     }
+
+    // The chain should re-org back to the valid chain once it sees the next block
+    await expect(nodeA.chain).toAddBlock(blockA6)
+    expect(nodeA.chain.head.hash.equals(blockA6.header.hash)).toBe(true)
   })
 
   it('does not remove nullifiers from double spends during reorg', async () => {
```
NullSoldier, line 860: we should delete this entire test now since double spends are not allowed. It doesn't test anything anymore.

ironfish/src/blockchain/blockchain.test.ts
```diff
@@ -759,74 +750,44 @@ describe('Blockchain', () => {
...
-    // us add the transaction a third time
-    const block5 = await useMinerBlockFixture(node.chain, 5, undefined, undefined, [tx])
-    await expect(node.chain.addBlock(block5)).resolves.toMatchObject({
+    await expect(node.chain.addBlock(block4)).resolves.toMatchObject({
       isAdded: false,
       reason: VerificationResultReason.DOUBLE_SPEND,
     })
   })
 
   it('rejects double spend during reorg', async () => {
```
NullSoldier, line 766: After some thinking, I think we should delete this test too. This is testing the same thing as the above test, that we check for double spends inside the pre-connect when were connecting a block.

At most, if you wanted to add a test that demonstrates you can add a double spend to a fork that would be different but that's expected behaviour.

ironfish/src/consensus/verifier.ts
```diff
@@ -364,33 +355,14 @@ export class Verifier {
...
-        if (processedSpends.has(spend.nullifier)) {
-          return { valid: false, reason: VerificationResultReason.DOUBLE_SPEND }
-        }
-
-        if (spend.size > previousNotesSize) {
-          return {
-            valid: false,
-            reason: VerificationResultReason.NOTE_COMMITMENT_SIZE_TOO_LARGE,
-          }
+        const verificationError = await this.verifySpend(spend, previousNotesSize, tx)
```
NullSoldier, line 362: Do we need this at all?

1. We're removing the double spend check
2. Could spend size checking against previousNotesSize could be done before we add the block?
3. Could spend commitments be checked before add?

ironfish/src/blockchain/blockchain.test.ts
danield9tqh: Yea agreed, removed it

ironfish/src/blockchain/blockchain.test.ts
danield9tqh: Idk I think this is still useful. It is a test that demonstrates you can add a double spend to a fork, like you said, and also tests that `connect` is called correctly when doing a re-org. Or at least that the check for double spends is actually IN the connect function and not just in `addBlock`

ironfish/src/consensus/verifier.ts
danield9tqh: Yea that's a good point. The only case I can think of is if we would want to later allow spends to reference other spends in the same block. We're not doing that rn so yea I think it would be the same as checking before. Wdyt about that scenario though? Either way do you mind if I make that change separately / after this PR if we want to do it?

ironfish/src/blockchain/blockchain.test.ts
NullSoldier: That's fine to keep it if you want.

You could demonstrate that without adding a double spend, though. Add the TX to the main chain, and the fork, and it'll show the spends aren't validated on the fork because it would have caught it from the main chain block with that same TX.

NullSoldier (approved):

ironfish/src/consensus/verifier.ts
danield9tqh: This doesn't seem like a feature we'd want to have since we're a privacy coin and miners cannot order transactions as they can in non-privacy coins

PR #2670, opened 2022-11-30
Move IDatabaseStore.getAllIter() to IDatabase

NullSoldier:
## Summary

This moves core underlying DB iterator into the DB itself from the store.

## Testing Plan

Added tests.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/storage/database/types.ts
+++ b/ironfish/src/storage/database/types.ts
@@ -4,16 +4,11 @@
 
 import { IJsonSerializable } from '../../serde'
 
-export interface DatabaseIterOptions {
+export interface DatabaseKeyRange {
   gt?: Buffer
   gte?: Buffer
   lt?: Buffer
   lte?: Buffer
-}
-
-export interface DatabaseKeyRange {
-  gte: Buffer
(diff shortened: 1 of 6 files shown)


ironfish/src/storage/database/types.ts
```diff
@@ -4,6 +4,13 @@
 
 import { IJsonSerializable } from '../../serde'
 
+export interface DatabaseIterOptions {
```
hughy, line 7: nit - i think the naming clashes a little with the usage in `StorageUtils.isInRange`, which makes me think this is a range, or "range options"

it seems like this interface is the most general form of `DatabaseKeyRange`, and the interface below is a specific type of range

i don't think changing the name of `DatabaseKeyRange` is in the scope of this pr if we were to do it at all

hughy (approved):

ironfish/src/storage/database/types.ts
NullSoldier: Good call. I'll remove mine, use DatabaseKeyRange, and expand it to have the new fields.

PR #2673, opened 2022-11-30
feat: add sender address to notes and validate in circuit

jowparks:
:exclamation: NOTE - Only one line of actual change in typescript
https://github.com/iron-fish/ironfish/pull/2673/files#r1049084361
The rest are fixture regeneration

## Summary

To be listed on exchanges, our note receivers must know who the sender is received funds, but we want this sender address to be private to the spender and owner of those notes. This changes moves us toward that direction.

We accomplish 4 things in this PR:
1. Randomness for randomized public key is now calculated at transaction level and propagated to the spends and outputs. This allows us to validate that the sender address in the output notes matches the sender address in spend note. This means that the randomized address that was shown to own the originating note is the same address that is passed into the sender.

Previous:
```
	Transaction
		Spend1
			randomness1
		Spend2
			randomness2
		Output1
			randomness3
		Output2
			randomness4
```
Current:
```
	Transaction
		randomness1
		Spend1
			randomness1
		Spend2
			randomness1
		Spend3
			randomness1
		Spend4
			randomness1
```

2. Public key randomness is added as input to the Output ZKP circuit. This allows us to calculate the sender address in circuit and thereby show that the address listed in the circuit is the same as the one used for the spend

3. Adds sender address to the Spend circuit. This allows us to pass in the original sender address from previous transaction so that we can regenerate the note contents. This is all so that we can create the same note commitment in circuit that matches the expected commitment for the nullifier.

4. Circuits params are regenerated

## Testing Plan
Unittests pass
## Breaking Change

```
[x] Yes
```

This would require network reset, but since we are doing so for next phase of testnet, no additional changes/considerations are required


ironfish-zkp/src/circuits/output.rs
```diff
@@ -166,26 +260,32 @@ mod test {
...
 
             let proof_generation_key = ProofGenerationKey { ak, nsk };
 
             let viewing_key = proof_generation_key.to_viewing_key();
 
             let payment_address = PUBLIC_KEY_GENERATOR * viewing_key.ivk().0;
 
-            let commitment_randomness = jubjub::Fr::random(&mut rng);
-            let esk = jubjub::Fr::random(&mut rng);
+            // let sender_address = ak + (SPENDING_KEY_GENERATOR * ar);
```
jowparks, line 283: remove

ironfish-rust/src/transaction/outputs.rs
```diff
@@ -76,20 +76,22 @@ impl OutputBuilder {
     pub(crate) fn build(
         &self,
         spender_key: &SaplingKey,
+        public_key_randomness: &jubjub::Fr,
     ) -> Result<OutputDescription, IronfishError> {
         let diffie_hellman_keys = self.note.owner.generate_diffie_hellman_keys();
-
+
```
jowparks, line 82: remove

ironfish-rust/src/transaction/tests.rs
```diff
@@ -192,7 +192,7 @@ fn test_miners_fee() {
         NATIVE_ASSET_GENERATOR,
         spender_key.public_address(),
     );
```
jowparks, line 194: Is this worrying that this test was passing? I changed from receiver -> spender to get it to start passing.

ironfish-zkp/src/circuits/output.rs
```diff
@@ -166,26 +260,32 @@ mod test {
         // Seed a fixed rng for determinism in the test
         let mut rng = StdRng::seed_from_u64(0);
 
-        for _ in 0..50 {
```
jowparks, line 169: Saw 5 iterations in spend circuit, so I reduced this one

ironfish-rust/src/note.rs
```diff
@@ -273,12 +273,12 @@ impl<'a> Note {
 
     /// Computes the note commitment, returning the full point.
     fn commitment_full_point(&self) -> jubjub::SubgroupPoint {
-        // TODO: @yajun replace this with commitment_full_point_new() that takes sender address
-        commitment_full_point(
```
jowparks, line 277: @ygao76 we should clean this up after this PR, we will go back and get rid of `commitment_full_point` and refactor the `commitment_full_point_new` to have that name

ygao76: yep, thats the plan

ironfish-rust/src/transaction/mod.rs
```diff
@@ -92,9 +93,12 @@ pub struct ProposedTransaction {
...
     /// signed.
     spender_key: SaplingKey,
-    //
-    // NOTE: If adding fields here, you may need to add fields to
-    // signature hash method, and also to Transaction.
+
+    // randomness used for the transaction to calculate the randomized ak, which
+    // allows us to verify the sender address is valid and stored in the notes
+    public_key_randomness: jubjub::Fr, //
+                                       // NOTE: If adding fields here, you may need to add fields to
```
ygao76, line 100: nit: indent

ironfish-rust/src/transaction/outputs.rs
```diff
@@ -170,17 +197,21 @@ impl OutputDescription {
     /// Converts the values to appropriate inputs for verifying the bellman proof.
     /// Demonstrates knowledge of a note containing the value_commitment, public_key
     /// and note_commitment
```
ygao76, line 199: Can you update the comment to include randomized public key

ironfish/src/primitives/transaction.ts
```diff
@@ -79,7 +79,11 @@ export class Transaction {
       // proof
       reader.seek(PROOF_LENGTH)
 
-      return new NoteEncrypted(reader.readBytes(ENCRYPTED_NOTE_LENGTH, true))
+      const note = new NoteEncrypted(reader.readBytes(ENCRYPTED_NOTE_LENGTH, true))
+      // TODO(joe): remove once rpk removed from proof and put on transaction
```
ygao76, line 83: why the note has rpk? the sender address size should be updated in `ENCRYPTED_NOTE_LENGTH` right?

tools/reset-network-genesis.sh
```diff
@@ -22,5 +22,5 @@ cd "$(dirname "$0")"
     yarn start chain:genesisblock -a IronFishGenesisAccount
 
     echo ""
-    echo "Copy the above block into genesisBlock.ts"
+    echo "Copy the above block into `DEV_GENESIS` in ironfish/src/defaultNetworkDefinitions.ts"
```
ygao76, line 25: nit: both DEV_GENSIS and TESTING

ironfish/src/primitives/transaction.ts
jowparks: In this section we are actually deserializing the `OutputDescription` which are called "notes" in this context 😅

ygao76: https://vscode.dev/github/iron-fish/ironfish/blob/07f483eec85cf88bd017849e4a2c5fdb729c4668/ironfish-rust-nodejs/src/structs/note_encrypted.rs#L30
pub const ENCRYPTED_NOTE_LENGTH: u32 =
    NOTE_ENCRYPTION_KEY_LENGTH + ENCRYPTED_NOTE_PLAINTEXT_LENGTH + 96;
   
pub const ENCRYPTED_NOTE_PLAINTEXT_LENGTH: u32 = ENCRYPTED_NOTE_SIZE as u32 + MAC_LENGTH;

pub const ENCRYPTED_NOTE_SIZE: usize =
    SCALAR_SIZE + MEMO_SIZE + AMOUNT_VALUE_SIZE + GENERATOR_SIZE + PUBLIC_ADDRESS_SIZE;
https://github.com/iron-fish/ironfish/blob/07f483eec85cf88bd017849e4a2c5fdb729c4668/ironfish-rust/src/note.rs#L27-L28

The encrypted note size is already includes public address size which is the sender address

ygao76: ugh I see, can you rename it so its less confusing.

ironfish-rust/src/transaction/tests.rs
mat-if: Don't think so - it seems like this test is just testing the miners fee is has the right amount and uses the hardcoded, technically invalid encryption keys. It is technically valid to pay out to any address prior to your change, but now it has some verification involved on the sender address which it didn't have before (since the sender address didn't exist)

mat-if (approved):

tools/reset-network-genesis.sh
jowparks: This var is used in both places, so should be good.

ironfish/src/primitives/transaction.ts
jowparks: I am not sure we can. I think outputs are generally referred to as notes everywhere in typescript.

ironfish/src/primitives/transaction.ts
```diff
@@ -112,9 +116,8 @@ export class Transaction {
 
     // sender address
     // TODO(joe): read from bytes rather than hardcoded value
-    // this._senderAddress = reader.readBytes(PUBLIC_ADDRESS_LENGTH, true).toString('hex')
+    // this._sender = reader.readBytes(PUBLIC_ADDRESS_LENGTH, true).toString('hex')
```
ygao76, line 119: What is the plan to replace the hardcode value?  Anything blocks this change? @jowparks

PR #2675, opened 2022-11-30
Remove nullifier commitment from block header

danield9tqh:
## Summary
Nullifiers are kept track of by each node to check for double spends. Currently we track nullifiers in a merkle tree but we could easily track them in a set as well which we would like to do. The block header has a commitment to the nullifier merkle tree but this only serves as a sanity check that the nullifier set was calculated correctly by everyone in the network. We do not necessarily need this sanity check and it comes at the expense of having a larger block header. It also locks us into using a merkle tree as the data structure for tracking nullifiers which we don't want. The solution is to remove the nullifier commitment from the block header

## Testing Plan
Unit tests

## Breaking Change

```
[X] Yes
```


ironfish/src/primitives/blockheader.ts
```diff
@@ -180,27 +178,27 @@ export class BlockHeader {
...
     this.noteCommitment = noteCommitment
-    this.nullifierCommitment = nullifierCommitment
     this.transactionCommitment = transactionCommitment
     this.target = target
     this.randomness = randomness
     this.timestamp = timestamp || new Date()
     this.graffiti = graffiti
     this.noteSize = noteSize ?? null
     this.work = work
+    this.nullifierSize = nullifierSize ?? null
```
dguenther, line 201: This is a trivial change, but I think `nullifierSize` should be next to `noteSize` here and in the parameters as well, since they're next to each other as properties above the constructor.

danield9tqh: Yea I thought about this. I put them away from each other so someone doesn't accidentally mix them up when using the constructor. I think the real fix for this like we discussed is to have named params for the constructor

dguenther (approved):

PR #2677, opened 2022-12-01
feat(ironfish): Add assets store to blockchain

rohanjadvani:
## Summary

Persist newly created assets in the blockchain DB.

## Testing Plan

N/A

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/database/assets.ts
+++ b/ironfish/src/blockchain/database/assets.ts
@@ -19,8 +19,8 @@ export class AssetsValueEncoding implements IDatabaseEncoding<AssetsValue> {
   serialize(value: AssetsValue): Buffer {
     const bw = bufio.write(this.getSize(value))
     bw.writeHash(value.createdTransactionHash)
-    bw.writeString(value.metadata, 'utf8')
-    bw.writeString(value.name, 'utf8')
+    bw.writeVarString(value.metadata, 'utf8')
+    bw.writeVarString(value.name, 'utf8')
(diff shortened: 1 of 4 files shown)


ironfish/src/blockchain/database/assets.ts
```diff
@@ -0,0 +1,51 @@
...
+  owner: string
+  supply: bigint
+}
+
+export class AssetsValueEncoding implements IDatabaseEncoding<AssetsValue> {
+  serialize(value: AssetsValue): Buffer {
+    const bw = bufio.write(this.getSize(value))
+    bw.writeHash(value.createdTransactionHash)
+    bw.writeVarString(value.metadata, 'utf8')
+    bw.writeVarString(value.name, 'utf8')
```
NullSoldier, line 23: how is name a var string? Is this a var string in our rust side and not a fixed length buffer?

mat-if: It gets converted to a fixed length buffer in the Rust side. It's functionally identical to the way `memo` works on `Note`

mat-if (approved):

ironfish/src/blockchain/database/assets.ts
NullSoldier: Ok, what do you think about down the line making a pass and converting all these strings to buffers? They're fixed size  and don't need to contain human readable data?

NullSoldier (approved):

PR #2684, opened 2022-12-02
replaces unsigned transaction hashes in rpc responses

hughy:
## Summary

uses the signed transaction hash instead of the unsigned hash in every rpc response that returns a transaction hash.

we store transactions by signed hash in the wallet but return unsigned hashes in rpc responses. every user-facing transaction hash is currently an unsigned hash, so lookups iterate over all traansactions to find a match.

removes logic for finding transactions by unsigned hash.

NOTE: this will also have the effect of propagating signed hashes to the API instead of unsigned hashes

## Breaking Change

```
[ ] Yes
```


NullSoldier (commented):
Yeees, so glad to do this.

NullSoldier (approved):

PR #2686, opened 2022-12-02
feat: add sender to note ts

jowparks:
## Summary
Adds sender to note as well as a test to verify public address that is recorded
NOTE: dependent on first having changes in rust codebase
## Breaking Change

```
[ ] Yes
```


ironfish/src/primitives/note.ts
```diff
@@ -29,7 +29,7 @@ export class Note {
     // skip owner public address
     reader.seek(PUBLIC_ADDRESS_LENGTH)
 
-    this._assetIdentifier = reader.readBytes(PUBLIC_ADDRESS_LENGTH, true)
```
ygao76: the asset identifier length is define with another constant. Its defined here https://github.com/iron-fish/ironfish/blob/fe19ede6ca0631f42441ec4a52698966f31f163c/ironfish-zkp/src/constants.rs#L13
You will need to export in rust-nodejs package using napi, then you can refer it in ts.

ironfish/src/consensus/__fixtures__/verifier.test.ts.fixture
```diff
@@ -947,5 +947,34 @@
         }
       ]
     }
```
jowparks, line 949: Maybe someone forgot to commit this?

ygao76 (approved):

ironfish-rust/src/keys/mod.rs
```diff
@@ -132,17 +132,13 @@ impl SaplingKey {
...
-            Err(_) => Err(IronfishError::InvalidPaymentAddress),
-            Ok(bytes) => {
-                if bytes.len() != 32 {
-                    Err(IronfishError::InvalidPaymentAddress)
-                } else {
-                    let mut byte_arr = [0; 32];
-                    byte_arr.clone_from_slice(&bytes[0..32]);
-                    Self::new(byte_arr)
-                }
-            }
```
jowparks, line 145: This was swallowing errors from `hex_to_bytes`, I am ok with reverting this change, but I think the new version is more explicit

ironfish/src/workerPool/tasks/createTransaction.ts
```diff
@@ -274,7 +274,7 @@ export class CreateTransactionTask extends WorkerTask {
     }
 
     for (const { publicAddress, amount, memo } of receives) {
-      const note = new Note(publicAddress, amount, memo)
+      const note = new Note(publicAddress, amount, memo, transaction.sender())
```
jowparks, line 277: Note to self, this is an important line :)

ironfish/src/primitives/note.test.ts
```diff
@@ -0,0 +1,48 @@
...
+import { createNodeTest, useAccountFixture, useMinerBlockFixture } from '../testUtilities'
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+describe('Note', () => {
+  const nodeTest = createNodeTest()
+
+  it('check if a transaction is not a miners fee', async () => {
+    const nodeA = nodeTest.node
```
mat-if, line 13: The primitives tests should probably not involve an entire node and chain. I would expect them to test only the constructor and it's own data/functions. The test name seems counterintuitive to what the test is actually testing (that the miner's fee transaction is a specific sender key).  Tests related to the miner's fee would probably make more sense in the `strategy` or `createMinersFee` test files

ironfish/src/primitives/note.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    const accountB = await useAccountFixture(nodeA.wallet, () =>
+      nodeA.wallet.createAccount('testB'),
+    )
+
+    // Create a block with a miner's fee
+    const block1 = await useMinerBlockFixture(nodeA.chain, 2, accountA)
+    await nodeA.chain.addBlock(block1)
+    await nodeA.wallet.updateHead()
+
+    const transaction = await nodeA.wallet.createTransaction(
```
mat-if, line 28: If this test needs these and isn't going to be refactored away, use a transaction fixture here - transactions are the heaviest part of most code. `useTxFixture` probably? Been a bit since I've worked with them

ironfish/src/primitives/note.test.ts
```diff
@@ -0,0 +1,48 @@
...
+      0,
+    )
+    const encryptedNote = transaction.getNote(0)
+    const note = encryptedNote.decryptNoteForOwner(accountB.incomingViewKey)
+    // TODO(joe): unhardcode address in circuit updates and update this test
+    expect(note?.senderAddress()).toBe(
+      '8a4685307f159e95418a0dd3d38a3245f488c1baf64bc914f53486efd370c563',
+    )
+    // expect(note?.senderAddress()).toBe(accountA.publicAddress)
+  }, 500000)
```
mat-if, line 47:
```suggestion
  })
```

We generally group tests on the amount of time needed to run after fixtures are generated, so if a test needs more than the default, it probably should live in a `slow` test file. This one shouldn't need to with a tx fixture, though

ironfish/src/strategy.test.slow.ts
```diff
@@ -274,11 +279,18 @@ describe('Demonstrate the Sapling API', () => {
       transaction.spend(note, witness)
       receiverNote.returnReference()
 
-      const noteForSpender = new NativeNote(spenderKey.public_address, BigInt(10), '')
+      const receiverAddress = generateKeyFromPrivateKey(receiverKey.spending_key).public_address
```
mat-if, line 282:
```suggestion
      const receiverAddress = receiverKey.public_address
```

Oh I didn't see at first that this was existing code, well, that's weird. Must be some legacy thing we don't need to do anymore.

ironfish/src/consensus/__fixtures__/verifier.test.ts.fixture
mat-if: Yeah I was running into this one too 🤷

mat-if (approved):
Looks good, few small things with tests so I'm not gonna hold it up on a follow-up review

ironfish/src/primitives/note.test.ts
ygao76: nice catch

ironfish/src/strategy.test.slow.ts
```diff
@@ -102,6 +103,9 @@ describe('Demonstrate the Sapling API', () => {
       expect(minerTransaction.notesLength()).toEqual(1)
     })
```
jowparks, line 105: expected

ironfish-rust/src/transaction/mod.rs
```diff
@@ -111,6 +111,12 @@ impl ProposedTransaction {
         }
     }
 
+    // return the sender of the transaction
+    // TODO(joe): change to calculate via randomness with circuit update
```
jowparks, line 115: This will actually come from transaction.sender rather than calculated by randomess. Leaving comment since it will be removed in next PR.

PR #2690, opened 2022-12-02
Tweak network definitions

danield9tqh:
## Summary
Making some small changes to network definitions like removing definitions that are not currently used. I think for the near term it would be good to only have one genesis block so that when we regenerate it we don't have to update it in multiple places. Also would be good to provide a default value for the genesisSupplyIron so that regenerating the genesis block is easier.

## Testing Plan
Unit tests

## Breaking Change

```
[ ] Yes
```


ironfish/src/defaultNetworkDefinitions.ts
```diff
@@ -54,120 +54,4 @@ export const TESTING = `{
...
-        "genesisSupplyInIron": 42000000,
-        "targetBlockTimeInSeconds": 60,
-        "maxSyncedAgeBlocks": 60,
-        "targetBucketTimeInSeconds": 10,
-        "maxBlockSizeBytes": 2000000
-    }
-}
- `
-
-export const DEV = `
```
NullSoldier, line 125: Can we bring the dev network back?

ironfish/src/defaultNetworkDefinitions.ts
```diff
@@ -54,120 +54,4 @@ export const TESTING = `{
   }
 }`
 
-export const MAINNET = `
```
NullSoldier, line 57: Can we bring main net back so we can use it as a staging ground for the main net parameters?

ironfish/src/defaultNetworkDefinitions.ts
mirayadav: Yeah I think the only network that should be deleted is the one for testnet phase 2.

ironfish/src/defaultNetworkDefinitions.ts
danield9tqh: Ahh, ok yea I wasn't about a case where we'd use the consensus params but not the genesis block.

ironfish/src/defaultNetworkDefinitions.ts
danield9tqh: Yep I can add this back

ironfish/src/fileStores/config.ts
```diff
@@ -377,7 +377,7 @@ export class Config extends KeyStore<ConfigOptions> {
       feeEstimatorPercentileLow: 10,
       feeEstimatorPercentileMedium: 20,
       feeEstimatorPercentileHigh: 30,
-      networkId: 2,
+      networkId: 1,
```
mirayadav, line 380: Should we default to the dev network since we don't have mainnet running right now?

ironfish/src/defaultNetworkDefinitions.ts
```diff
@@ -70,58 +70,6 @@ export const MAINNET = `
...
-        "maxSyncedAgeBlocks": 60,
-        "targetBucketTimeInSeconds": 10,
-        "maxBlockSizeBytes": 2000000
-    }
-}
- `
-
 export const DEV = `
 {
     "id": 3,
```
mirayadav, line 75: Can we make this id 2?

ironfish/src/node.ts
```diff
@@ -242,8 +236,6 @@ export class IronfishNode {
       networkDefinitionJSON = TESTING
     } else if (config.get('networkId') === 1) {
       networkDefinitionJSON = MAINNET
-    } else if (config.get('networkId') === 2) {
-      networkDefinitionJSON = TESTNET_PHASE_2
     } else if (config.get('networkId') === 3) {
```
mirayadav, line 239: Will also have to change the id to 2 here.

ironfish/src/defaultNetworkDefinitions.ts
```diff
@@ -70,104 +70,50 @@ export const MAINNET = `
...
-          "data": "base64:KYJk06s6m1IblZiINE3U6cGN8YLwNpmawcr7N1XoqCE="
-        },
-        "target": "883423532389192164791648750371459257913741948437809479060803100646309888",
-        "randomness": "0",
-        "timestamp": 1669940355519,
-        "graffiti": "67656E6573697300000000000000000000000000000000000000000000000000",
-        "noteSize": 3,
-        "work": "0",
-        "nullifierSize": 1
+export const DEV = `{
```
danield9tqh, line 73: Updating the `DEV` to be the same as `TESTING` because I think there may have been some block changes that went in after this was last updated

mirayadav (approved):

PR #2693, opened 2022-12-05
Allow loading genesis block allocations from CSV

dguenther:
## Summary

The genesis block command doesn't currently allow for more than a single recipient for the genesis coins. This adds support for loading a batch of allocations from a CSV file.

## Testing Plan

* [x] Test some invalid inputs manually to make sure the command handles them.
* [x] Test allocating tokens to multiple addresses and make sure the tokens show up in `accounts:balance --all`

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/chain/genesisblock.ts
+++ b/ironfish-cli/src/commands/chain/genesisblock.ts
@@ -43,7 +43,7 @@ export default class GenesisBlockCommand extends IronfishCommand {
     allocations: Flags.string({
       required: false,
       description:
-        'A CSV file with the format address,amount,memo containing genesis block allocations',
+        'A CSV file with the format address,amountInIron,memo containing genesis block allocations',
       exclusive: ['account', 'memo'],
     }),
     genesisSupplyInIron: Flags.string({
@@ -146,7 +146,7 @@ const parseAllocationsFile = (
       continue
     }
 
-    const [address, amount, memo, ...rest] = line.split(',').map((v) => v.trim())
+    const [address, amountInIron, memo, ...rest] = line.split(',').map((v) => v.trim())
 
     if (rest.length > 0) {
       return {
@@ -164,7 +164,7 @@ const parseAllocationsFile = (
     }
 
     // Check amount is positive and decodes as $IRON
-    const amountInOre = CurrencyUtils.decodeIron(amount)
+    const amountInOre = CurrencyUtils.decodeIron(amountInIron)
     if (amountInOre < 0) {
       return {
         ok: false,
--- a/ironfish-cli/src/commands/chain/genesisblock.ts
+++ b/ironfish-cli/src/commands/chain/genesisblock.ts
@@ -46,10 +46,10 @@ export default class GenesisBlockCommand extends IronfishCommand {
         'A CSV file with the format address,amountInIron,memo containing genesis block allocations',
       exclusive: ['account', 'memo'],
     }),
-    genesisSupplyInIron: Flags.integer({
+    genesisSupplyInIron: Flags.string({
       char: 'g',
       required: true,
-      default: 42000000,
+      default: '42000000',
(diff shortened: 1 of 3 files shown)


ironfish-cli/src/commands/chain/genesisblock.ts
```diff
@@ -64,34 +74,119 @@ export default class GenesisBlockCommand extends IronfishCommand {
...
+
+      if (totalSupply !== expectedSupply) {
+        this.error(
+          `Allocations file contains ${CurrencyUtils.encodeIron(
+            totalSupply,
+          )} $IRON, but --genesisSupplyInIron expects ${flags.genesisSupplyInIron} $IRON.`,
+        )
+      }
+
+      allocations = result.allocations
```
ygao76, line 100: Do you think we need a way to dry run the command with the csv file? So in the dry run, it will print out the parsed csv file

danield9tqh: Yep, agreed. At least should probably print out addresses and confirm first

ironfish-cli/src/commands/chain/genesisblock.ts
```diff
@@ -36,6 +40,12 @@ export default class GenesisBlockCommand extends IronfishCommand {
       default: 'Genesis Block',
       description: 'The memo of the block',
     }),
+    allocations: Flags.string({
+      required: false,
+      description:
+        'A CSV file with the format address,amountInIron,memo containing genesis block allocations',
+      exclusive: ['account', 'memo'],
+    }),
     genesisSupplyInIron: Flags.integer({
```
danield9tqh, line 49: I think this needs to be changed back to a string. I changed it to `number` but didn't think about decimal places

danield9tqh:
Did a few tests
1. Tested genesis block with 6 different public addresses, mined a little bit after then checked the balance of all those accounts and they were correct. These all added up to 42000000
2. Tested again with amounts that did not add up to 42000000. The command makes sure you pass in the correct supply in order to continue and again all balances were correct after mining
3. Tested again with same list but added an additional item for one of the public addresses so that one public address had two outputs. All balances were correct after mining

ironfish-cli/src/commands/chain/genesisblock.ts
dguenther: Good idea. I updated this

ironfish-cli/src/commands/chain/genesisblock.ts
dguenther: Good idea, I added a `--dryRun` flag that logs the difficulty and list of allocations, then exits, and I also added a confirmation step to creating the genesis block that logs the difficulty and list of allocations and asks to confirm.

I also added duplicate detection -- It could be intentional that someone would include the same address multiple times, but worth informing the user in case it was a copy/paste error.

ironfish/src/genesis/makeGenesisBlock.ts
```diff
@@ -50,7 +52,7 @@ export async function makeGenesisBlock(
   // It should end up with 0 coins.
   const genesisKey = generateKey()
   // Create a genesis note granting the genesisKey allocationSum coins.
-  const genesisNote = new NativeNote(genesisKey.public_address, allocationSum, info.memo)
+  const genesisNote = new NativeNote(genesisKey.public_address, allocationSum, '')
```
danield9tqh, line 55: We're not using the memo here anymore?

dguenther: flags.memo is still in the single allocation when an `allocations` CSV isn't passed. It didn't make sense to put it on this Note since this note is only ever decryptable by the `genesisKey` above. The `genesisKey` is just used to temporarily hold the genesis supply before it's sent to the `allocations`.

danield9tqh (approved):

PR #2695, opened 2022-12-06
adds separate method for syncing pending transactions

hughy:
## Summary

pending transactions are currently added to the account via the same function, syncTransaction, that we use to process transaction when blocks are added to and removed from the chain. this can make transaction processing confusing.

creates a separate 'addPendingTransaction' method for transactions created by an account or gossiped to the node. it only adds a transaction to an account if the account has not processed the transaction before.

the addTransaction method on the account
- creates new decryptedNote records for each output note in the transaction received by the account
- adds each of these notes to nonChainNoteHashes
- marks each spent note as spent
- adds the transaction to pendingTransactionHashes
- creates new record in walletDb transactions store
- updates balance once with the delta from received notes and spent notes

these changes also update tests and fixtures to use addPendingTransaction instead of syncTransaction.

## Testing Plan

- updated unit tests
- added unit tests specific to new methods

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/wallet.ts
```diff
@@ -394,9 +394,32 @@ export class Wallet {
...
+
+    if (accounts.length === 0) {
+      return
+    }
+
+    const decryptedNotesByAccountId = await this.decryptNotes(transaction, null, accounts)
+
+    for (const [accountId, decryptedNotes] of decryptedNotesByAccountId) {
+      const account = this.accounts.get(accountId)
+      Assert.isNotUndefined(account, `addPendingTransaction: No account found for ${accountId}`)
```
NullSoldier, line 416: This is a code bomb. Decryption is slow, if you remove an account it's likely to crash if its in the middle of adding a pending transaction and that gossip is for the account being removed.

ironfish/src/wallet/account.ts
```diff
@@ -298,6 +379,10 @@ export class Account {
     return await this.walletDb.loadTransaction(this, hash, tx)
   }
 
+  async hasTransaction(hash: Buffer, tx?: IDatabaseTransaction): Promise<boolean> {
+    return (await this.getTransaction(hash, tx)) !== undefined
```
NullSoldier, line 383: we don't need to load the value + do deserialization, so we might as well make a version of hasTransaction that only loads the key.

hughy: Good point, this is lazy

ironfish/src/wallet/wallet.ts
hughy: Should we `continue` if the account is removed during decryption?

We have the same check in `syncTransaction`

ironfish/src/wallet/wallet.ts
```diff
@@ -425,7 +453,11 @@ export class Wallet {
 
     for (const [accountId, decryptedNotes] of decryptedNotesByAccountId) {
       const account = this.accounts.get(accountId)
-      Assert.isNotUndefined(account, `syncTransaction: No account found for ${accountId}`)
+
+      if (!account) {
+        this.logger.warn(`Account ${accountId} removed from wallet during note decryption`)
```
NullSoldier, line 458: warnings are for things going wrong that aren't fatal, but why is it wrong for someone to remove a wallet?

hughy: Fair point; it's may be 'wrong' from the perspective of the programmer who expects the account to exist, but not from the perspective of the user who removed the account

I'll remove the warning

ironfish/src/network/peerNetwork.ts
```diff
@@ -1368,7 +1368,7 @@ export class PeerNetwork {
 
       // Sync every transaction to the wallet, since senders and recipients may want to know
       // about pending transactions even if they're not accepted to the mempool.
-      await this.node.wallet.syncTransaction(transaction, {})
+      await this.node.wallet.addPendingTransaction(transaction, null)
```
NullSoldier, line 1371: Get rid of this second parameter, and always use chain.head.sequence no matter what.

ironfish/src/wallet/account.ts
```diff
@@ -195,6 +195,89 @@ export class Account {
...
+    tx?: IDatabaseTransaction,
+  ): Promise<void> {
+    await this.walletDb.db.withTransaction(tx, async (tx) => {
+      let balanceDelta = 0n
+      for (const decryptedNote of decryptedNotes) {
+        if (decryptedNote.forSpender) {
+          continue
+        }
+
+        await this.walletDb.nonChainNoteHashes.put([this.prefix, decryptedNote.hash], null, tx)
```
NullSoldier, line 211: We should use the account wrapper function here

ironfish/src/wallet/account.ts
```diff
@@ -195,6 +195,89 @@ export class Account {
...
+            blockHash: null,
+            sequence: null,
+          },
+          tx,
+        )
+
+        balanceDelta += note.value()
+      }
+
+      for (const spend of transaction.spends()) {
```
NullSoldier, line 233: If you have a chain of TX like `A -> B` and get B first, this won't mark any notes from A as spent. This code doesn't check notes and check if nullifiers exist in a pending state here. We should add a test for this, and fix this type of case in the future.

NullSoldier (approved):

NullSoldier:
I'm going to approve this but there is more work to be done here

 - We need to have a lock to avoid the issue of over-writing data when gossip comes in and blocks are being connected
 - We don't handle marking notes as spent if a later transaction has come in first. There is also no test to test this cause which doesn't workl
 - We can parallelize the hasTransaction() checks before decrypting

NullSoldier (approved):

PR #2696, opened 2022-12-06
Delete minedBlockIndexer

dguenther:
## Summary

Deletes the minedBlockIndexer, since we can add data to the wallet to track 
this rather than maintaining a separate database.

This PR is just for deleting the indexer -- Jason mentioned he'd add support 
for it in the next iteration of the wallet if it's useful for node 
operators.

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/miners/mined.ts
```diff
@@ -1,136 +0,0 @@
...
-} from '@ironfish/sdk'
-import { CliUx, Flags } from '@oclif/core'
-import readline from 'readline'
-import { parseNumber } from '../../args'
-import { IronfishCommand } from '../../command'
-import { RemoteFlags } from '../../flags'
-import { ProgressBar } from '../../types'
-import { linkText } from '../../utils/terminal'
-
-export class MinedCommand extends IronfishCommand {
```
dguenther, line 20: I could update this command to scan the chain instead for now, but I'm not sure whether it's useful for testnet participants now.

NullSoldier (approved):

ironfish-cli/src/commands/miners/mined.ts
dguenther: Discussed and decided that since we're not doing a release in a while, we have time to stop informing people to use this command, or rewrite it with the new wallet.

ironfish/src/migrations/data/index.ts
```diff
@@ -4,7 +4,6 @@
 
 import { Migration010 } from './010-blockchain'
 import { Migration011 } from './011-accounts'
-import { Migration012 } from './012-indexer'
 import { Migration013 } from './013-wallet-2'
 
-export const MIGRATIONS = [Migration010, Migration011, Migration012, Migration013]
+export const MIGRATIONS = [Migration010, Migration011, Migration013]
```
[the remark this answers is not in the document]
dguenther: Discussed and determined that this should work.

PR #2698, opened 2022-12-06
Enforce separate data directories per network

mirayadav:
## Summary

Enforces unique network per data directory by storing network ID in `internal.json` and confirming that network ID flag and network definition contain the same network ID. The genesis block hash in the network definition is also matched with the existing chain genesis block hash, if it exists. Copies custom network definitions into data directory so that the definition file doesn't have to be passed in every time.

## Testing Plan

Ran start command locally with different network ID flags and custom network files to check for correct behavior.

## Breaking Change

```
[ ] Yes
```


ironfish/src/networkDefinition.ts
```diff
@@ -47,3 +51,61 @@ export const networkDefinitionSchema: yup.ObjectSchema<NetworkDefinition> = yup
...
+      internal.isSet('networkId') &&
+      config.isSet('networkId') &&
+      networkId !== config.get('networkId')
+    ) {
+      throw Error('Network ID flag does not match network ID stored in datadir')
+    }
+
+    if (config.isSet('networkId')) {
+      networkId = config.get('networkId')
+    }
```
dguenther, line 78: This is kind of just a code style thing, but I'd suggest defining networkId after the `if` check, because I think it makes it easier to read if networkId can be declared as `const`. Like this:

```suggestion
    if (
      internal.isSet('networkId') &&
      config.isSet('networkId') &&
      internal.get('networkId') !== config.get('networkId')
    ) {
      throw Error('Network ID flag does not match network ID stored in datadir')
    }
    
    const networkId = config.isSet('networkId') ? config.get('networkId') : internal.get('networkId')
```

dguenther (approved):
Left a comment on code style but otherwise looks good to me 👍 feel free to merge once that's addressed

PR #2699, opened 2022-12-06
CI checks for yarn.lock changes manually

mat-if:
## Summary

Needed because --frozen-lockfile is broken in yarn 1.x. We can probably go back to using --frozen-lockfile if we upgrade to yarn 2.x

Fail with outsanding yarn.lock changes: https://github.com/iron-fish/ironfish/actions/runs/3633233416/jobs/6130008284
Pass with no outstanding yarn.lock changes: https://github.com/iron-fish/ironfish/actions/runs/3633244719/jobs/6130032328

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,16 @@ jobs:
           cache: 'yarn'
 
       - name: Install packages
-        run: yarn --non-interactive --frozen-lockfile --ignore-scripts
+        run: yarn --non-interactive --ignore-scripts
+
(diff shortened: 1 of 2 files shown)


hughy (approved):
What's stopping us from upgrading to yarn 2.x?

mat-if:
> What's stopping us from upgrading to yarn 2.x?

I think the only thing stopping us is taking the time to do it and make sure everything works etc. It should probably happen alongside the node version upgrade

PR #2702, opened 2022-12-06
Fix call to FeeEstimator.init() in openDB()

NullSoldier:
## Summary

The function is for opening DB connections, this function does not open a database.

The fee estimator should also not take a Wallet. It won't have access to it once we split it out, and the estimator should not be wallet aware. I will fix in a future PR.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/memPool/memPool.ts
+++ b/ironfish/src/memPool/memPool.ts
@@ -83,6 +83,10 @@ export class MemPool {
     })
   }
 
+  async start(): Promise<void> {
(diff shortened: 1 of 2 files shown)


ygao76:
The feeEstimator needs wallet now to create the transaction so it can estimate the transaction size. Once we have transaction builder or create transaction is separate from post, then we can have a direct call to get transaction size without wallet. The dependency of wallet can be removed at that point.

ygao76 (approved):

PR #2703, opened 2022-12-06
feat(ironfish,rust-nodejs): Update blockchain assets store when processing blocks

rohanjadvani:
## Summary

* Expose Asset-related fields and constants via Napi
* Fix asset serialization to use fixed-length buffers in the store
* Update description values from `number`s to `bigint`s
* Add a new fixture utility which accepts notes to spend and manually creates a transaction via the worker pool
* When a new block is processed:
  * Upsert assets from mint descriptions to the store. The creation will store the associated transaction hash
  * Decrement asset supply from burn descriptions
* When a block is removed:
  * Decrement supply if a mint is reverted
  * If a transaction hash matches the hash associated with an asset, remove that record if the supply has reached 0
  * Increment supply if a burn is reverted

## Testing Plan

Added unit tests

## Breaking Change

```
[x] Yes
```

This is a breaking change since we are updating the assets schema in the database and the wallet has not been yet updated to be asset-aware. We plan on merging this code change in since a release will not be pushed out until the wallet has been updated, and migrations are being skipped with a pending network reset.


ironfish/src/blockchain/blockchain.ts
```diff
@@ -1336,6 +1337,8 @@ export class Blockchain {
     const hashes = await this.sequenceToHashes.get(sequence, tx)
     await this.sequenceToHashes.put(sequence, { hashes: [...(hashes?.hashes || []), hash] }, tx)
 
+    await this.updateAssetsFromBlock(block, tx)
```
NullSoldier, line 1340: This is processing assets from forks.

ironfish/src/blockchain/blockchain.ts
```diff
@@ -1346,6 +1349,151 @@ export class Blockchain {
...
   }
 
+  private async updateAssetsFromBlock(block: Block, tx: IDatabaseTransaction): Promise<void> {
+    for (const transaction of block.transactions) {
+      await this.processMintsToAssetsStore(transaction, tx)
+      await this.processBurnsToAssetsStore(transaction, tx)
+    }
+  }
+
+  private async processMintsToAssetsStore(
```
NullSoldier, line 1359: This is a small nomenclature thing, but we use the term `connect` when were adding things to the main chain instead of `process`

ironfish/src/blockchain/blockchain.ts
```diff
@@ -1346,6 +1349,151 @@ export class Blockchain {
...
+          nonce: asset.nonce(),
+          owner: asset.owner(),
+          supply,
+        },
+        tx,
+      )
+    }
+  }
+
+  private async undoAssetChangesFromBlock(
```
NullSoldier, line 1417: nomenclature thing but we use the term disconnect to indicate a disconnection operation from the main chain

ironfish/src/blockchain/blockchain.ts
```diff
@@ -1154,6 +1154,7 @@ export class Blockchain {
         await this.sequenceToHashes.put(header.sequence, { hashes }, tx)
       }
 
+      await this.undoAssetChangesFromBlock(hash, tx)
```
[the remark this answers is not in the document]
rohanjadvani: Yep, updated

ironfish/src/blockchain/blockchain.ts
rohanjadvani: Fixed

ironfish/src/primitives/burnDescription.ts
```diff
@@ -3,12 +3,7 @@
...
-  readonly asset: Asset
-  readonly value: number
-
-  constructor(asset: Asset, value: number) {
-    this.asset = asset
-    this.value = value
-  }
+export interface BurnDescription {
+  asset: Asset
+  value: bigint
```
NullSoldier, line 8: Were using interfaces here, its ok for these to have other fields than this?

ironfish/src/primitives/transaction.ts
```diff
@@ -94,17 +94,17 @@ export class Transaction {
...
       reader.seek(64)
 
-      return new MintDescription(asset, value)
+      return { asset, value }
     })
 
     this._burns = Array.from({ length: _burnsLength }, () => {
       const asset = Asset.deserialize(reader.readBytes(ASSET_LENGTH))
-      const value = reader.readU8()
+      const value = BigInt(reader.readU64())
```
NullSoldier, line 102: This is a bug, readU64() cannot read greater than 2^53. You need to use the big versions of these functions.

NullSoldier (approved):
There's one small serialization bug I saw.

ironfish/src/primitives/transaction.ts
rohanjadvani: Yeah, I think that makes sense. Would [this](https://github.com/iron-fish/ironfish/blob/master/ironfish/src/primitives/note.ts#L25) be a bug as well then?

ironfish/src/blockchain/blockchain.ts
```diff
@@ -1292,6 +1290,11 @@ export class Blockchain {
       tx,
     )
 
+    for (const transaction of block.transactions) {
```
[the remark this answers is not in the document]
rohanjadvani: If you have a block with transaction A which mints some supply and transaction B which tries to spend that supply, that block will be invalid. Transaction B will have a spend which requires a valid witness, which won't hold true.

If you have a block with one transaction that mints some supply *and* has an output to transfer some tokens to another address, that will technically work (since the output circuit does not require a witness) and the value balancing will take care of the difference. It should work effectively the same if you replace the mint with a spend (assuming the merkle path verification passes).

We can add a test for this.

ironfish/src/primitives/transaction.ts
NullSoldier: Yep, there are bigint versions of all these functions in bufio. If we have a number, and were casting to BigInt or from BigInt we should almost always treat that as an error unless you can guarantee it won't overflow like the miners fee.

ironfish/src/blockchain/blockchain.ts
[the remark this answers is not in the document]
NullSoldier: Sounds good. Bitcoin has very clear same block guarantees, that are well defined in their consensus model. It would be nice if ours, or lack of ours were encoded in tests. Thanks

PR #2706, opened 2022-12-07
Remove element field from merkle tree leaves

dguenther:
## Summary

Removes the element field from merkle tree leaves, since we only use it for tests. This should save a minimum of 328MB per 1000 blocks, assuming that all 1000 blocks only have a miners transaction.

## Breaking Change

```
[X] Yes
```


NullSoldier (approved):
This looks great, a huge improvement to the code base.

PR #2707, opened 2022-12-07
refactor(rust,zkp): Move ValueCommitment and add constructor method

mat-if:
## Summary

- Moved ValueCommitment out of /circuits since it doesn't really belong there, into /primitives
- Added additional tests
- Added a `ValueCommitment::new()` constructor method to simplify code

## Breaking Change

```
[ ] Yes
```


mat-if:
Moving this to draft, I'll probably wait to merge this until some of the sender PRs are in to minimize merge conflicts. Also will probably refactor this a bit to make the struct fields private so we ensure we're using them with the proper constructor

ironfish-rust/src/merkle_note.rs
```diff
@@ -301,11 +299,7 @@ mod test {
         );
         let diffie_hellman_keys = EphemeralKeyPair::new();
 
-        let value_commitment = ValueCommitment {
-            value: note.value,
-            randomness: jubjub::Fr::random(thread_rng()),
-            asset_generator: note.asset_generator(),
-        };
+        let value_commitment = ValueCommitment::new(note.value, note.asset_generator());
```
danield9tqh, line 302: I'm not totally opposed to hiding this randomness but I think we should be careful putting it in the background. It could be helpful to explicitly show where we are getting randomness from or that a certain component depends on randomness.

jowparks (approved):

ironfish-rust/src/merkle_note.rs
mat-if: Hmm, I'd have to think about this more to form a better opinion, but my initial reaction is that randomness is a pretty minor implementation detail that doesn't need to be thought about unless you're working on the proofs or something.

Issue #2376, opened 2022-10-18
Disconnected due to TypeError: this._items.at is not a function


gjfrn:
### What happened?

Since updating from 0.1.47, peers keep getting disconnected due to the error in the log. This prevents my node from syncing.

### Version

0.1.49

### Debug output

```shell
Stopping sync from EUFyGNb (MSI2020) due to /Users/test/ironfish/ironfish/src/memPool/priorityQueue.ts:67
    return this._items.at(0)
                       ^

TypeError: this._items.at is not a function
    at PriorityQueue.peek (/Users/test/ironfish/ironfish/src/memPool/priorityQueue.ts:67:24)
    at MemPool.onConnectBlock (/Users/test/ironfish/ironfish/src/memPool/memPool.ts:169:44)
    at null.<anonymous> (/Users/test/ironfish/ironfish/src/memPool/memPool.ts:66:12)
    at Event.emitAsync (/Users/test/ironfish/ironfish/src/event.ts:65:31)
    at Blockchain.addHeadToChain (/Users/test/ironfish/ironfish/src/blockchain/blockchain.ts:726:31)
    at Blockchain.connect (/Users/test/ironfish/ironfish/src/blockchain/blockchain.ts:584:7)
    at null.<anonymous> (/Users/test/ironfish/ironfish/src/blockchain/blockchain.ts:371:31)
    at LevelupDatabase.withTransaction (/Users/test/ironfish/ironfish/src/storage/database/database.ts:211:22)
    at Blockchain.addBlock (/Users/test/ironfish/ironfish/src/blockchain/blockchain.ts:341:23)
    at Syncer.addBlock (/Users/test/ironfish/ironfish/src/syncer.ts:423:40)
Peer EUFyGNb (MSI2020) disconnected (DISCONNECTED) while syncing.
```

### Relevant log output

```shell
Stopping sync from EUFyGNb (MSI2020) due to /Users/test/ironfish/ironfish/src/memPool/priorityQueue.ts:67
    return this._items.at(0)
                       ^

TypeError: this._items.at is not a function
    at PriorityQueue.peek (/Users/test/ironfish/ironfish/src/memPool/priorityQueue.ts:67:24)
    at MemPool.onConnectBlock (/Users/test/ironfish/ironfish/src/memPool/memPool.ts:169:44)
    at null.<anonymous> (/Users/test/ironfish/ironfish/src/memPool/memPool.ts:66:12)
[...]

PR #2709, opened 2022-12-07
Require node 16

danield9tqh:
## Summary
The CLI is built to run on node version 16. We have run into one [issue](https://github.com/iron-fish/ironfish/issues/2376) that arose from incompatible node version. We should require users to be using the correct node version so we do not run into these versioning errors.

## Testing Plan
Local testing

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/bin/run
+++ b/ironfish-cli/bin/run
@@ -8,7 +8,7 @@ if (process.platform !== 'win32') {
 }
 
 if(process.versions.node.split('.')[0] !== '16') {
-  console.log('Node version ' +  process.versions.node + ' is not compatible. Must have node v16.x installed. You can install using nvm: https://github.com/nvm-sh/nvm')
+  console.log('NodeJS version ' +  process.versions.node + ' is not compatible. Must have node v16.x installed. You can install using nvm: https://github.com/nvm-sh/nvm')
   process.exit(1)
 }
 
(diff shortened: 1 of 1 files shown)


ironfish-cli/bin/run
```diff
@@ -7,6 +7,11 @@ if (process.platform !== 'win32') {
   require('segfault-handler').registerHandler('segfault.log')
 }
 
+if(process.versions.node.split('.')[0] !== '16') {
+  console.log('NodeJS version ' +  process.versions.node + ' is not compatible. Must have node v16.x installed. You can install using nvm: https://github.com/nvm-sh/nvm')
```
NullSoldier, line 11: Can we just link to node 16? NVM might be too hard for non technical people.

hughy (commented):
Could we enforce this on install?

e.g., https://stackoverflow.com/a/29349773

NullSoldier:
If you do the package strategy, does that work for execution too? I think we want it on both install / execution ideally. In the case that you're using NVM and forget to switch node versions.

ironfish-cli/bin/run
danield9tqh: Yea good call

NullSoldier:
@hughy: @danield9tqh  is going to look into preventing this on install as well since there's more questions about if that works properly. We'll merge this in and expect a PR open for fixing users installing it using the wrong version. We can stop them a lot sooner! Especially because it'll cause modules to get built under the correct node version.

NullSoldier (approved):
Tested manually, this works well.

PR #2710, opened 2022-12-08
adds separate methods for syncing transactions on a block

hughy:
## Summary

we currently use one method, syncTransaction, to process all transaction changes within the wallet. this makes it difficult to distinguish between different sets of logic for adding pending transactions vs. adding transactions mined on a block.

- implements addBlock method in wallet to process transactions when a block is added to the chain - for each account, add all transactions and update head hash within a single transaction - chainProcessor onAdd only calls addBlock for accounts that are up to date with the chain processor

- replaces call to syncTransaction during rescan with addBlock

- implements walletDb method addDecryptedNote to update note indices (nullifierToNoteHash, nonChainNoteHashes, sequenceToNoteHash) whenever a decrypted note is saved
- implements walletDb method addTransaction to update pendingTransactionHashes whenever a transaction is saved based on whether the transaction is on a block
- implements walletDb method spendNote to encapsulate note updates when a note is spent

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/wallet.ts
```diff
@@ -536,34 +553,9 @@ export class Wallet {
...
-        blockHash,
-        transaction,
-        initialNoteIndex,
-        sequence,
-      } of this.chain.iterateBlockTransactions(blockHeader)) {
-        if (scan.isAborted) {
-          scan.signalComplete()
-          this.scan = null
-          return
-        }
```
hughy, line 549: I need to go back and re-integrate this logic into `connectBlock`

ironfish/src/wallet/wallet.ts
```diff
@@ -397,6 +387,45 @@ export class Wallet {
     return decryptedNotes
   }
 
+  async connectBlock(
```
NullSoldier, line 390: later this should only update accounts where header.previousHash === account.head

NullSoldier (approved):

PR #2715, opened 2022-12-08
refactor(ci): Separate rust lint into a separate workflow

mat-if:
## Summary

The goal here is to allow testing and linting to happen simultaneously without stepping on each other due to failures. A quick 5 second lint check should not stop a 5 minute test run and vice-versa.

Confirming `fmt` still works: https://github.com/iron-fish/ironfish/actions/runs/3650307611/jobs/6166123169

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/.github/workflows/rust_ci.yml
+++ b/.github/workflows/rust_ci.yml
@@ -19,8 +19,8 @@ on:
 name: Rust CI
 
 jobs:
-  ironfish_rust:
-    name: ironfish-rust
+  rust_lint:
+    name: Lint Rust
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/rust_ci.yml
```diff
@@ -39,39 +39,29 @@ jobs:
...
       # clippy
       - uses: actions-rs/cargo@v1
         name: "Clippy check on ironfish-rust"
         with:
           command: clippy
-          args: --manifest-path ironfish-rust/Cargo.toml --all-targets -- -D warnings
+          args: --all-targets -- -D warnings
 
-  ironfish_rust_nodejs:
-    name: ironfish-rust-nodejs
```
mat-if, line 74: the nodejs job only ran lint, so it was removed entirely. since it should only be napi glue code, it will have it's typescript tests tested in the typescript CI

jowparks (approved):

PR #2717, opened 2022-12-08
Remove Wallet 2.0 migration

dguenther:
## Summary

I think at this point we can drop support for Wallet 1.0 databases, since we'll be resetting the chain anyway. If users who want to retrieve their keys from a wallet 1.0 database will need to export them on an older version. 

## Breaking Change

```
[X] Yes
```

Breaks migrating Wallet 1.0 wallets to Wallet 2.0


hughy (approved):
Pushed a commit to remove the reference to the migration in `index.ts`

PR #2718, opened 2022-12-08
Add network ID and genesis block hash to Identify Message

mirayadav:
## Summary

Add network ID and genesis block hash to the Identify message and confirm that connecting peers have the same network ID and genesis block hash as the current node. Ban peers that advertise an incorrect genesis block hash on their Identify message.

## Testing Plan

Added tests to `peerManager.test.ts` and `syncer.test.ts`.

## Breaking Change

```
[ ] Yes
```


ironfish/src/syncer.ts
```diff
@@ -456,6 +456,15 @@ export class Syncer {
...
     reason: VerificationResultReason | null
   }> {
+    if (
+      block.header.sequence === GENESIS_BLOCK_SEQUENCE &&
+      peer.genesisBlockHash &&
+      !block.header.hash.equals(peer.genesisBlockHash)
+    ) {
+      peer.punish(BAN_SCORE.MAX, VerificationResultReason.INVALID_GENESIS_BLOCK)
+      return { added: false, block, reason: VerificationResultReason.INVALID_GENESIS_BLOCK }
+    }
```
dguenther, line 466: I'm not certain whether this will practically be hit in any code paths right now. This `addBlock` should only be called during either gossip or syncing, and during syncing it's preceded by the ancestor search using `findAncestor`. I just tested out the case where a peer reports a correct genesisHash in their Identify message, but then returns all `0` hashes for block hashes requested in `findAncestor`, and it hits an Assert that throws an error.

So I think findAncestor might be a better place to check for this, but it'd be worth setting up a quick test case locally with a peer that reports a correct genesis block in their Identify message, but then returns incorrect responses later to GetBlockHashes.

mirayadav: Changed the check to be in `findAncestor` and also added a check to make sure peers aren't gossiping their genesis block. I'm going to run some additional tests locally tomorrow, but please let me know if the location of these checks seems right.

ironfish/src/network/peerNetwork.ts
```diff
@@ -1197,6 +1197,11 @@ export class PeerNetwork {
       return
     }
 
+    if (block.header.sequence === GENESIS_BLOCK_SEQUENCE) {
+      peer.punish(BAN_SCORE.MAX, 'Peer gossiped its genesis block')
+      return
```
dguenther, line 1202: We need to do this for compact blocks as well -- We won't ever hit this code without receiving a compact block first.

Also, this should be done next to the block header verification call since it's also verifying the block header, and should call `this.blockFetcher.removeBlock(block.header.hash)` (and probably `this.chain.addInvalid`)

ironfish/src/syncer.ts
dguenther: I think this is in the right place -- Just need to retest (manually or otherwise) to verify it's working

ironfish/src/syncer.ts
```diff
@@ -331,9 +331,19 @@ export class Syncer {
       lower = needle + 1
     }
 
-    Assert.isNotNull(ancestorHash)
     Assert.isNotNull(ancestorSequence)
 
+    if (ancestorSequence === GENESIS_BLOCK_SEQUENCE && ancestorHash === null) {
+      this.logger.warn(
+        `Peer ${peer.displayName} sent invalid genesis block hash in Identify message`,
```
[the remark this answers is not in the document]
mirayadav: Changed the logged message. Also, I tested this locally and realized that it had to be moved as the ancestorSequence would be null if the genesis block hashes didn't match.

ironfish/src/network/peerNetwork.ts
mirayadav: Added this check for compact blocks as well and moved it next to the block header verification call (with the additional `removeBlock and `addInvalid` calls). Also tested this locally.

ironfish/src/network/peerNetwork.ts
```diff
@@ -1204,6 +1211,13 @@ export class PeerNetwork {
     peer.knownBlockHashes.set(block.header.hash, KnownBlockHashesValue.Received)
 
     // verify the block header
+    if (block.header.sequence === GENESIS_BLOCK_SEQUENCE) {
+      peer.punish(BAN_SCORE.MAX, VerificationResultReason.GOSSIPED_GENESIS_BLOCK)
```
dguenther, line 1215: I think we can drop the ban here, since we don't do that for invalid blocks.

mirayadav: Dropped

dguenther (approved):

PR #2719, opened 2022-12-09
chore(ironfish,cli): Update jest to newer version

mat-if:
## Summary

Updating jest will allow us to use some of the newer features, like `--shard` which would enable us to improve our CI run times easily. This upgrade is not 100% complete, see the associated TODO in `jest.config.base.js` for the legacy options we are carrying forward for now.

## Breaking Change

```
[ ] Yes
```


rohanjadvani (approved):

ironfish/src/testUtilities/fixtures.ts
```diff
@@ -45,7 +46,7 @@ export async function useFixture<TFixture, TSerialized = unknown>(
   } = {},
 ): Promise<TFixture> {
   const testPath = getCurrentTestPath()
-  const testName = expect.getState().currentTestName
+  const testName = expect.getState().currentTestName || ''
```
dguenther, line 49: I'm not sure under what condition these would be null, so it should probably just throw an error here so something unexpected doesn't happen.

dguenther: Seems like this was returning undefined or an empty string before, and we call this outside of some tests via `makeDbName`. Leaving it for now

dguenther (approved):

PR #2722, opened 2022-12-10
Replace nullifier merkle tree with nullifier set

danield9tqh:
## Summary
The nullifier merkle tree was originally added for two purposes: to potentially [enable a faster sync](https://beanstalk.notion.site/Optimistic-Sync-91a880c2dedf491ba5b3779b07a1a216), and to potentially allow snapshot downloading of the trees. We didn’t end up getting the fast sync method to work out and we’ll probably consider an alternative implementation when we revisit fast syncing. For the snapshot downloading, I’m not sure whether it provides much value, since notes + nullifiers (including their proofs) are 95% of block size, and I think we could just use the chain data to generate snapshots anyway (maybe without the proofs).

Moving to a nullifier set is going to save us a little on code complexity since maintaining a set is easier than maintaining a merkle tree. It will also save space because we do not have to store internal nodes of the merkle tree.

## Testing Plan
Unit tests and running locally

## Breaking Change

```
[ ] Yes
```


ironfish/src/blockchain/nullifierSet/nullifierSet.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    await node.wallet.updateHead()
+
+    const { block: block3 } = await useBlockWithTx(node, accountA, accountA, false)
+    await expect(node.chain).toAddBlock(block3)
+
+    await node.wallet.updateHead()
+
+    const { block: block4 } = await useBlockWithTx(node, accountA, accountA, false)
+    await expect(node.chain).toAddBlock(block4)
+
```
dguenther, line 37: Maybe `await node.wallet.updateHead()`? I guess it doesn't do anything anyway

ironfish/src/blockchain/nullifierSet/nullifierSet.test.ts
```diff
@@ -0,0 +1,81 @@
...
+      expect(await set.contains(nullifier)).toBe(true)
+    }
+
+    for (const nullifier of [...block3Nullifiers]) {
+      expect(await set.contains(nullifier)).toBe(true)
+    }
+
+    for (const nullifier of [...block4Nullifiers]) {
+      expect(await set.contains(nullifier)).toBe(false)
+    }
```
dguenther, line 79: Maybe add a test for `contained` somewhere in here as well?

ironfish/src/blockchain/nullifierSet/nullifierSet.ts
```diff
@@ -0,0 +1,132 @@
...
+  value: NullifierInfo
+}
+
+export class NullifierSet {
+  readonly db: IDatabase
+
+  // Keep track of number of nullifiers in the set
+  private readonly counter: IDatabaseStore<CounterEntry<'Size'>>
+
+  // Nullifier -> BlockHash, TransactionHash
```
dguenther, line 35: Does this need to be updated to something like `Nullifier -> BlockHash, position`? Same with the below. I haven't looked to see what this is in other places

ironfish/src/blockchain/nullifierSet/nullifierSet.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  disconnectBlock(transactions: Transaction[], tx?: IDatabaseTransaction): Promise<void> {
+    return this.db.withTransaction(tx, async (tx) => {
+      let currentSize = await this.size(tx)
+
+      for (const transaction of transactions) {
+        for (const spend of [...transaction.spends()].reverse()) {
+          // Sanity check that we are removing the correct block
+          // Not necesarrally needed but
```
dguenther, line 117: I think this comment needs finishing
```suggestion
          // Not necessarily needed but
```

ironfish/src/blockchain/blockchain.ts
```diff
@@ -1313,10 +1298,9 @@ export class Blockchain {
     Assert.isNotNull(prev.noteSize)
     Assert.isNotNull(prev.nullifierSize)
 
-    await Promise.all([
-      this.notes.truncate(prev.noteSize, tx),
-      this.nullifiers.truncate(prev.nullifierSize, tx),
-    ])
+    await Promise.all([this.notes.truncate(prev.noteSize, tx)])
```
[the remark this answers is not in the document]
danield9tqh: Yea good catch will update this

ironfish/src/blockchain/nullifierSet/nullifierSet.test.ts
danield9tqh: added

ironfish/src/blockchain/nullifierSet/nullifierSet.ts
danield9tqh: Yea it should be `Nullifier -> TransactionHash, position`. Updated

ironfish/src/blockchain/nullifierSet/nullifierSet.ts
danield9tqh: thanks, fixed

ironfish/src/blockchain/blockchain.test.ts
```diff
@@ -409,11 +408,11 @@ describe('Blockchain', () => {
     expect(await node.chain.notes.size()).toBe(blockA3.header.noteSize)
   })
 
-  describe('MerkleTrees', () => {
-    it('should add notes and nullifiers to trees', async () => {
+  describe('MerkleTree + Nullifier Set', () => {
+    it('should add notes to trees and nullifiers to set', async () => {
```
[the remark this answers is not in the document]
danield9tqh: Nice, fixed

ironfish/src/blockchain/blockchain.ts
```diff
@@ -985,22 +976,19 @@ export class Blockchain {
...
-        }
       }
 
+      const originalNullifierSize = await this.nullifiers.size(tx)
       await this.notes.addBatch(blockNotes, tx)
-      await this.nullifiers.addBatch(blockNullifiers, tx)
 
       const noteCommitment = await this.notes.rootHash(tx)
       const noteSize = await this.notes.size(tx)
-      const nullifierSize = await this.nullifiers.size(tx)
```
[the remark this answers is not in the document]
dguenther: Closing this out, I'm more ambivalent since it does move us toward making `newBlock` more of a pure function.

ironfish-cli/src/commands/chain/repair.ts
```diff
@@ -153,14 +153,16 @@ export default class RepairChain extends IronfishCommand {
...
 
     CliUx.ux.action.start('Clearing notes MerkleTree')
     await node.chain.notes.truncate(noteSize)
     CliUx.ux.action.stop()
 
-    CliUx.ux.action.start('Clearing nullifier MerkleTree')
-    await node.chain.nullifiers.truncate(nullifierSize)
+    CliUx.ux.action.start('Clearing nullifier set')
+
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
```
NullSoldier, line 163: Don't add this.

ironfish/src/blockchain/blockchain.ts
```diff
@@ -255,16 +255,7 @@ export class Blockchain {
...
-      hasher: new NullifierHasher(),
-      leafIndexKeyEncoding: BUFFER_ENCODING,
-      leafEncoding: new LeafEncoding(),
-      nodeEncoding: new NodeEncoding(),
-      db: this.db,
-      name: 'u',
-      depth: 32,
-      defaultValue: Buffer.alloc(32),
-    })
+    this.nullifiers = new NullifierSet({ db: this.db, storeName: 'nu' })
```
NullSoldier, line 258: Why does this take a store name? We'll only ever have 1 of these per blockchain DB right?

ironfish/src/blockchain/blockchain.ts
```diff
@@ -985,22 +976,19 @@ export class Blockchain {
...
       }
 
+      const originalNullifierSize = await this.nullifiers.size(tx)
       await this.notes.addBatch(blockNotes, tx)
-      await this.nullifiers.addBatch(blockNullifiers, tx)
 
       const noteCommitment = await this.notes.rootHash(tx)
       const noteSize = await this.notes.size(tx)
-      const nullifierSize = await this.nullifiers.size(tx)
+      const addedNullifiers = transactions.reduce((count, t) => count + t.spendsLength(), 0)
```
NullSoldier, line 990: why add this separate reduce? Just calculate this in the loop above as the count of nullifiers in the block.

ironfish/src/blockchain/blockchain.ts
```diff
@@ -985,22 +976,19 @@ export class Blockchain {
...
+      const nullifierSize = originalNullifierSize + addedNullifiers
```
NullSoldier, line 991: Ok, so we're keeping this for an integrity check?

ironfish/src/blockchain/blockchain.ts
```diff
@@ -1313,10 +1291,9 @@ export class Blockchain {
     Assert.isNotNull(prev.noteSize)
     Assert.isNotNull(prev.nullifierSize)
 
-    await Promise.all([
-      this.notes.truncate(prev.noteSize, tx),
-      this.nullifiers.truncate(prev.nullifierSize, tx),
-    ])
+    await this.notes.truncate(prev.noteSize, tx)
+
+    await this.nullifiers.disconnectBlock(block.transactions, tx)
```
NullSoldier, line 1296: These can't be done in parallel anymore?

NullSoldier: This function is called disconnetBlock() but doesn't take a block. We should rename it right?

ironfish/src/blockchain/nullifierSet/nullifierSet.ts
```diff
@@ -0,0 +1,132 @@
...
+  DatabaseSchema,
+  IDatabase,
+  IDatabaseStore,
+  IDatabaseTransaction,
+  StringEncoding,
+  U32_ENCODING,
+} from '../../storage'
+import { NullifierInfo, NullifierLocationEncoding } from './encoding'
+
+interface CounterEntry<T extends string> extends DatabaseSchema {
```
NullSoldier, line 19: This doesn't matter too much but using interfaces for serialization is a little dangerous because it's permissive. It's better to use types if you ever loop over the keys.

ironfish/src/blockchain/nullifierSet/nullifierSet.ts
```diff
@@ -0,0 +1,132 @@
...
+    // Nullifier -> BlockHash, TransactionHash
+    this.nullifiers = this.db.addStore({
+      name: options.storeName,
+      keyEncoding: BUFFER_ENCODING,
+      valueEncoding: new NullifierLocationEncoding(),
+    })
+  }
+
+  size(tx?: IDatabaseTransaction): Promise<number> {
+    return this.db.withTransaction(tx, async (tx) => {
```
NullSoldier, line 56: why is this starting a transaction?

ironfish/src/blockchain/nullifierSet/nullifierSet.ts
```diff
@@ -0,0 +1,132 @@
...
+      const size = await this.counter.get('Size', tx)
+      return size === undefined ? 0 : size
+    })
+  }
+
+  contains(nullifier: Nullifier, tx?: IDatabaseTransaction): Promise<boolean> {
+    return this.db.withTransaction(tx, async (tx) => {
```
NullSoldier, line 63: why is this starting a transaction?

ironfish/src/blockchain/nullifierSet/nullifierSet.ts
```diff
@@ -0,0 +1,132 @@
...
+      if (nullifierInfo === undefined) {
+        return false
+      }
+
+      return nullifierInfo.position < size
+    })
+  }
+
+  get(nullifier: Nullifier, tx?: IDatabaseTransaction): Promise<NullifierInfo | undefined> {
+    return this.db.withTransaction(tx, async (tx) => {
```
NullSoldier, line 80: why does this start a transaction?

ironfish/src/blockchain/nullifierSet/nullifierSet.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  contains(nullifier: Nullifier, tx?: IDatabaseTransaction): Promise<boolean> {
+    return this.db.withTransaction(tx, async (tx) => {
+      return await this.nullifiers.has(nullifier, tx)
+    })
+  }
+
+  contained(nullifier: Nullifier, size: number, tx?: IDatabaseTransaction): Promise<boolean> {
+    return this.db.withTransaction(tx, async (tx) => {
```
NullSoldier, line 69: no transaction needed here

ironfish/src/blockchain/nullifierSet/nullifierSet.ts
```diff
@@ -0,0 +1,132 @@
...
+    })
+  }
+
+  contains(nullifier: Nullifier, tx?: IDatabaseTransaction): Promise<boolean> {
+    return this.db.withTransaction(tx, async (tx) => {
+      return await this.nullifiers.has(nullifier, tx)
+    })
+  }
+
+  contained(nullifier: Nullifier, size: number, tx?: IDatabaseTransaction): Promise<boolean> {
```
NullSoldier, line 68: Do we need this in a set? It seems like we should delete this function now right?

ironfish/src/blockchain/nullifierSet/encoding.ts
```diff
@@ -0,0 +1,44 @@
...
+import { TransactionHash } from '../../primitives/transaction'
+
+export type NullifierInfo = {
+  // transaction the nullfier is a part of on the main chain
+  transactionHash: TransactionHash
+  // zero indexed position of the nullfiier on the main chain
+  // Use 32 bit because we use 32 bit for notes tree. This is
+  // somewhat small but if we end up having to increase note tree size
+  // we can increase this nullifier set size at the same time
+  position: number
```
NullSoldier, line 16: It's a set now, there is no position in a set.

ironfish/src/blockchain/nullifierSet/nullifierSet.test.ts
danield9tqh: Thanks, added tests here

ironfish/src/blockchain/blockchain.ts
danield9tqh: Yea I passed the store name so that it is clear in the blockchain file which store names are currently taken. So someone else doesn't add another store with this prefix.

ironfish/src/blockchain/nullifierSet/nullifierSet.ts
danield9tqh: So we can call contains without passing in a transaction if we need to

ironfish/src/blockchain/nullifierSet/nullifierSet.ts
danield9tqh: So we can call contains without passing in a transaction if we need to

ironfish/src/blockchain/nullifierSet/nullifierSet.ts
danield9tqh: This will be useful for the case where we want to check whether a transaction is a double spend with a given confirmation range. In gossip code/mempool for example if we want to reject double spends that are older than 10 block confirmation range we can get the size at that block and call contained with it

ironfish-cli/src/commands/chain/repair.ts
danield9tqh: Fair, what do you think we should do? I don't really want to write a clear method on nullifier set because it's not used anywhere else and this command might be out of date eventually. We could delete this command also or not repair the nullifier set in it.

ironfish/src/blockchain/blockchain.ts
danield9tqh: Keeping annotated sizes on block headers for the use case I mentioned previously where we want to check if a nullifier exists in a certain confirmation range. We can get the size of the nullifier set on a specific block and check if the nullifier existed before that block

ironfish/src/blockchain/blockchain.ts
danield9tqh: Yea, alright changed that

ironfish/src/blockchain/blockchain.ts
```diff
@@ -96,7 +96,7 @@ export class Blockchain {
     SerializedNoteEncrypted,
     SerializedNoteEncryptedHash
   >
-  nullifiers: MerkleTree<Nullifier, NullifierHash, string, string>
+  nullifiers: NullifierSet
```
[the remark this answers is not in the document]
danield9tqh: Good catch, I removed them in this PR

ironfish/src/blockchain/nullifierSet/encoding.ts
```diff
@@ -0,0 +1,39 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import type { IDatabaseEncoding } from '../../storage/database/types'
+import bufio from 'bufio'
+import { TransactionHash } from '../../primitives/transaction'
+
+export type NullifierInfo = {
+  transactionHash: TransactionHash // transaction the nullfier is a part of on the main chain
+  position: number // zero indexed position of the nullfiier on the main chain
```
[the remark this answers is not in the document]
NullSoldier: fixed

ironfish/src/blockchain/blockchain.ts
NullSoldier: If you ever try to re-use the name it'll throw an error.

ironfish/src/blockchain/blockchain.ts
NullSoldier: It's hard to know how we'll want to accomplish that. Maybe we just build an in memory cache instead of using the DB here? It's unclear because it's not being used yet.

ironfish/src/blockchain/nullifierSet/nullifierSet.ts
```diff
@@ -0,0 +1,131 @@
...
+  U32_ENCODING,
+} from '../../storage'
+import { NullifierInfo, NullifierLocationEncoding } from './encoding'
+
+interface CounterEntry<T extends string> extends DatabaseSchema {
+  key: T
+  value: number
+}
+
+type NullifiersSchema extends DatabaseSchema = {
```
NullSoldier, line 24: nit: why is the above an interface and this is a type?

ironfish-cli/src/commands/chain/repair.ts
danield9tqh: Added a `clear` method here since we are keeping this repair command

ironfish/src/blockchain/blockchain.ts
danield9tqh: Yea I think having an in-mem cache does make sense for this use case. I'll remove the position from this set

ironfish/src/blockchain/nullifierSet/nullifierSet.test.ts
danield9tqh: Removed the `contained` method after all...

ironfish/src/blockchain/nullifierSet/nullifierSet.ts
danield9tqh: I don't think we can `extend` types that's why we have interface here. We do this for all other DBSchema fields in [schema.ts](https://github.com/iron-fish/ironfish/blob/97ab6751b0b698f8e05674b282e6b04ce3ea12d1/ironfish/src/blockchain/schema.ts) as well.

ironfish/src/blockchain/blockchain.ts
danield9tqh: Going to leave this here for consistency since even the merkle tree takes in a name field

ironfish/src/blockchain/blockchain.ts
danield9tqh: Updated to take a block

ironfish/src/blockchain/nullifierSet/nullifierSet.ts
danield9tqh: As you pointed out, `db.get` can also take an undefined transaction and will create its own so removing this.

ironfish/src/blockchain/nullifierSet/nullifierSet.ts
danield9tqh: removed

ironfish/src/blockchain/nullifierSet/nullifierSet.ts
danield9tqh: As you pointed out, `db.get` can also take an undefined transaction and will create its own so removing this.

ironfish/src/blockchain/nullifierSet/nullifierSet.ts
danield9tqh: removed

ironfish/src/blockchain/nullifierSet/nullifierSet.ts
danield9tqh: Discussed this and for this specific use case we could just keep a cache of nullifiers in previous 10 blocks for example the code would look like
```javascript
function isDoubleSpend(nullifier) {
  if(chain.nullifiers.contains(nullifier)) {
    return !last10BlocksCache.contains(nullifier)
  }
  return false
}
```

ironfish/src/blockchain/nullifierSet/nullifierSet.ts
danield9tqh: I don't think we can `extend` types that's why we have interface here. We do this for all other DBSchema fields in [schema.ts](https://github.com/iron-fish/ironfish/blob/97ab6751b0b698f8e05674b282e6b04ce3ea12d1/ironfish/src/blockchain/schema.ts) as well.

ironfish/src/blockchain/nullifierSet/nullifierSet.test.ts
```diff
@@ -0,0 +1,104 @@
...
+import { Assert } from '../../assert'
+import {
+  createNodeTest,
+  useAccountFixture,
+  useBlockWithTx,
+  useMinerBlockFixture,
+} from '../../testUtilities'
+import { NullifierSet } from './nullifierSet'
+
+describe('NullifierSet', () => {
```
NullSoldier, line 14: Should we just move these all into blockchain.test.ts, since they are blockchain tests?

ironfish/src/primitives/nullifier.ts
```diff
@@ -2,36 +2,5 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-import { Blake3Hasher } from '@napi-rs/blake-hash'
-import { MerkleHasher } from '../merkletree'
-import { BufferSerde, NullifierSerdeInstance } from '../serde'
-
 export type Nullifier = Buffer
 export type NullifierHash = Buffer
```
NullSoldier, line 6: This isnt used anymore is it?

NullSoldier (approved):

PR #2729, opened 2022-12-13
feat(ironfish): Update `balances` store in `walletdb` to include asset

rohanjadvani:
## Summary

* Update `balances` store to have a composite key of account prefix and asset identifier
* Default key usage to use native asset identifier (will be updated in subsequent PRs)

## Testing Plan

N/A

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/walletdb/walletdb.ts
+++ b/ironfish/src/wallet/walletdb/walletdb.ts
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+import { Asset } from '@ironfish/rust-nodejs'
 import { BufferMap } from 'buffer-map'
 import { Assert } from '../../assert'
 import { FileSystem } from '../../fileSystems'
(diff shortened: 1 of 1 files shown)


ironfish/src/wallet/walletdb/walletdb.ts
```diff
@@ -527,7 +531,12 @@ export class WalletDB {
...
 
   async getUnconfirmedBalance(account: Account, tx?: IDatabaseTransaction): Promise<bigint> {
-    const unconfirmedBalance = await this.balances.get(account.id, tx)
+    // TODO(mgeist,rohanjadvani): This will be updated in a subsequent PR to
+    // include an asset as an argument
+    const unconfirmedBalance = await this.balances.get(
+      [account.prefix, Asset.nativeIdentifier()],
+      tx,
+    )
     Assert.isNotUndefined(unconfirmedBalance)
```
hughy, line 540: Wondering if it makes sense to return 0 if there is no balance stored for the asset instead of asserting here

hughy (approved):

ironfish/src/wallet/walletdb/walletdb.ts
rohanjadvani: Yep, I am going to do that in a subsequent PR. This is stubbed out to just include the native identifier right now. When I add `assetIdentifier` added as an argument to this method, that logic will be updated to return `BigInt(0)` if no record is found.

mat-if (approved):

PR #2730, opened 2022-12-13
feat(ironfish): Add asset awareness when checking amounts needed for spends

rohanjadvani:
## Summary

Currently the wallet only uses the native asset and checks values across all notes when given a payload for transaction `receives`. Since this is now asset aware, we need to check amounts needed for each given asset type. This code change builds aggregate amounts keyed by asset identifier and fetches unspent notes with matching identifiers.

## Testing Plan

Added unit test

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -1,7 +1,8 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { generateKey } from '@ironfish/rust-nodejs'
+import { Asset, generateKey } from '@ironfish/rust-nodejs'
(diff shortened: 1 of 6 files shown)


ironfish/src/wallet/wallet.ts
```diff
@@ -747,15 +741,65 @@ export class Wallet {
...
+    for (const [assetIdentifier, amountNeeded] of amountsNeeded.entries()) {
+      const { amount, notes } = await this.createSpendsForAsset(
+        sender,
+        assetIdentifier,
+        amountNeeded,
+      )
+      if (amount < amountNeeded) {
+        throw new NotEnoughFundsError(
+          `Insufficient funds: Needed ${amountNeeded.toString()} but have ${amount.toString()}`,
+        )
```
hughy, line 785: Should we include the assetIdentifier in the message? Or even as an attribute of the error type?

hughy (approved):

mat-if (approved):

PR #2735, opened 2022-12-14
Remove nullifier size from block header

danield9tqh:
## Summary
We removed the nullifier size commitment from the block header but had kept the nullifierSize field on a database block in case we needed the field for some other use case. There were a couple use cases considered:
1. Having a nullifier size on each block would allow us to incrementally verify that the size of our nullifier set is consistent with what it should be
2.  If we wanted to test for inclusion of a nullifier before a certain block we could look at the nullifier set size at that block and then compare that with the nullifier's position in an ordered set.

For 1 we agreed it just didn't seem like it added much validation. It was essentially verifying the correctness of other code. Mainly that we were adding the nullifiers into the set correctly which seems simple enough that it doesn't need extra validation

For 2 there are other ways to tell if a nullifier was included before a certain block. If the block is not too far back (e.g. < 100 blocks) we could just keep a cache of nullifiers seen in those last ~100 blocks and check that the nullifier is not in that cache. We may want to do this for filtering out double spends from the mempool for example (but within a certain block confirmation range). 

The nullifierSize does not seem like it's adding much more so removing makes sense.

## Testing Plan
Unit tests

## Breaking Change

```
[ ] Yes
```


dguenther (approved):

dguenther:
Looks like there are some conflicts now, but otherwise looks good

PR #2738, opened 2022-12-14
feat(ironfish): Use `assetIdentifier` when saving unconfirmed balance

rohanjadvani:
## Summary

Pass in asset identifiers from notes when updating balances.

## Testing Plan

N/A

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/accounts/repair.ts
```diff
@@ -109,7 +110,8 @@ export default class Repair extends IronfishCommand {
     walletDb: WalletDB,
     chain: Blockchain,
   ): Promise<void> {
-    let unconfirmedBalance = 0n
+    const unconfirmedBalances = new BufferMap<bigint>()
+    const unconfirmedBalance = 0n
```
mat-if, line 114: I don't think we need this variable anymore - it's only used for the log, which probably needs to be updated too

ironfish-cli/src/commands/accounts/repair.ts
```diff
@@ -156,15 +158,23 @@ export default class Repair extends IronfishCommand {
...
+          const unconfirmedBalance = unconfirmedBalances.get(assetIdentifier) ?? BigInt(0)
+          unconfirmedBalances.set(
+            assetIdentifier,
+            unconfirmedBalance + decryptedNoteValue.note.value(),
+          )
         }
       }
     }
 
     this.log(
```
mat-if, line 171: We should move this into the loop below to output for each asset that it updates

hughy (approved):

mat-if (approved):
Should we add some test coverage for this? We get a lot for free for the native asset, but just to confirm it's working for a custom asset as well

ironfish-cli/src/commands/accounts/repair.ts
rohanjadvani: Done

ironfish-cli/src/commands/accounts/repair.ts
rohanjadvani: Ah yeah forgot to clean it up, thanks

PR #2744, opened 2022-12-14
updates balance only when blocks are connected or disconnected

hughy:
## Summary

wallet 2.0 stores the pending balance for each account in the wallet database. this balance includes all transactions involving the account that are either on the chain or are pending.

including pending transactions in the stored balance makes it difficult to store a correct balance that handles forks, double spends, expiring transactions, etc.

updating the balance only when blocks are connected or disconnected from the chain convertsthe balance storage to the unconfirmed balance that matches the state of the account on the chain at a particular block.

- isolates balance updates to three functions: - connectTransaction, called for each transaction in a block connected to the chain - bulkUpdateDecryptedNotes, only called for notes in a transaction in a block disconnected from the chain - processTransactionSpends, only called for transactions in a block disconnected from the chain

- changes getBalance to not calculate pending balance

- updates all uses of getBalance in codebase and tests

## Breaking Change

```
[ ] Yes
```


ironfish/src/genesis/genesis.test.slow.ts
```diff
@@ -140,8 +140,8 @@ describe('Create genesis block', () => {
     await expect(
       newNode.wallet.getBalance(accountNewNode, Asset.nativeIdentifier()),
     ).resolves.toMatchObject({
-      confirmed: amountBigint,
-      pending: amountBigint,
+      confirmed: amountBigint * 2n,
+      unconfirmed: amountBigint * 2n,
```
NullSoldier, line 144: What's the explanation here?

NullSoldier (approved):

NullSoldier (approved):

NullSoldier:
This is a huge improvement to fixing balance for users.

PR #2745, opened 2022-12-14
WIP adds separate wallet methods for disconnecting blocks

hughy:
## Summary

wallet disconnectBlock method processes each transaction on the block and updates the head hash for an account within a single database transaction.

account disconnectTransaction method
- deletes sequenceToNoteHash entry for each decrypted note
- adds an entry to nonChainNoteHashes for each decrypted note
- sets blockHash and sequence to null on each decrypted note
- sets nullifier and index to null on each decrypted note
- adds the transaction to pendingTransactionHashes
- sets blockHash and sequence to null on the transaction
- updates unconfirmed balance

defines disconnectNoteHashSequence in walletDb to encapsulate index updates when note is disconnectd

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):
Significant improvement to the clarity of this code path, thank you. Now we can finally remove syncTransaction.

PR #2746, opened 2022-12-14
Normalize transaction value accessors

NullSoldier:
## Summary

Some new descriptions didnt return iterators, and some had functions that returned length. None of that matters now since its all just arrays fmor the constructor.

## Breaking Change

```
[ ] Yes
```


hughy (approved):
LGTM

i'd like to go back and do the same with some of the other attributes (fee, expirationSequence, etc)

mat-if (approved):

PR #2747, opened 2022-12-14
Fix instances of reversing in place

NullSoldier:
## Summary

Array.reverse() will reverse in place, which we don't intend in all of these places.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -1392,7 +1392,7 @@ export class Blockchain {
     transaction: Transaction,
     tx: IDatabaseTransaction,
   ): Promise<void> {
-    for (const { assetIdentifier, value } of transaction.burns.reverse()) {
+    for (const { assetIdentifier, value } of transaction.burns.slice().reverse()) {
       const existingAsset = await this.assets.get(assetIdentifier, tx)
       Assert.isNotUndefined(existingAsset)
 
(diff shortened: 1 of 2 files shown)


mat-if (approved):

hughy (approved):

ironfish/src/migrations/migrator.ts
```diff
@@ -68,7 +68,7 @@ export class Migrator {
   async revert(options?: { dryRun?: boolean }): Promise<void> {
     const dryRun = options?.dryRun ?? false
 
-    const migrations = Array.from(this.migrations).reverse()
+    const migrations = this.migrations.slice().reverse()
```
dguenther, line 71: Is this a style-related change, or do these have functional differences?

ironfish/src/blockchain/blockchain.ts
```diff
@@ -1418,7 +1418,7 @@ export class Blockchain {
     transaction: Transaction,
     tx: IDatabaseTransaction,
   ): Promise<void> {
-    for (const { asset, value } of transaction.mints.reverse()) {
+    for (const { asset, value } of transaction.mints.slice().reverse()) {
```
dguenther, line 1421: Might be worth turning some of these into readonly arrays, like

`public readonly mints: readonly MintDescription[]`

ironfish/src/migrations/migrator.ts
NullSoldier: no functional difference, I just made it the same since I was fixing other bugs.

ironfish/src/blockchain/blockchain.ts
NullSoldier: Yeah, I guess the entire Transaction should be mutable.

PR #2751, opened 2022-12-15
Remove limiting miner fee transactions to single note

dguenther:
## Summary

Now that we have a block size limit, miners are lightly disincentivized from creating a miner's fee transaction with many notes. I don't see a reason to limit the number of notes 
in the transaction, since it allows miners additional flexibility when distributing block rewards.

## Breaking Change

```
[X] Yes
```

Multi-note miners fee transactions won't verify on older code.


NullSoldier:
It seems fine, I had to write out the pros and cons. How many blocks in BTC use multiple outputs per miners fee?

Pros
 - More flexible miners fees and payouts

Cons:
 - Feature adds complexity but won't be used because pools don't payout per block
 - Verification becomes less straight forward and we lose having a nice consistent standardized miner transaction
 - Larger miner transaction blocks

PR #2755, opened 2022-12-15
Move randomized public key to transaction

ygao76:
## Summary
Move rpk to transaction object 
Pass rpk from transaction to spend, mint and output.

This PR will reduce the output and spend size.

## Testing Plan
unit tests

## Breaking Change

```
[ x] Yes
```

@mat-if @jowparks


ironfish-rust/src/transaction/mints.rs
```diff
@@ -63,6 +63,7 @@ impl MintBuilder {
         &self,
         spender_key: &SaplingKey,
         public_key_randomness: &Fr,
```
jowparks, line 65: I wonder if this is better than regenerating from the randomness as it was doing before. You can still regenerate as it was doing in line 78, but keep all the other updates below you made to the MintDescription/verify proof. Up to you

ironfish/src/primitives/transaction.ts
```diff
@@ -76,15 +77,15 @@ export class Transaction {
...
       // proof
       reader.seek(PROOF_LENGTH)
 
-      const note = new NoteEncrypted(reader.readBytes(ENCRYPTED_NOTE_LENGTH, true))
-      // TODO(joe): remove once rpk removed from proof and put on transaction
-      // randomized public key
-      reader.seek(32)
-      return note
+      // output note
+      const outputNote = new NoteEncrypted(reader.readBytes(ENCRYPTED_NOTE_LENGTH, true))
```
jowparks, line 86: nit: can return directly rather than assigning here since there is no subsequent `seek`

jowparks (approved):

ironfish-rust/src/transaction/mints.rs
ygao76: yeah both would work. I figure since we store the rpk in transaction, I just pass in instead of regenerating individually in spend, output and mint since its repetitive computation.

ironfish/src/primitives/transaction.ts
ygao76: ok, I will update it in the next cleaning up PR.

PR #2756, opened 2022-12-15
refactor(rust): Encapsulate description verification and remove unused verify_proof functions

mat-if:
## Summary

Realized this code was too easy to get wrong because it was not clear what was happening.
- Removed `verify_proof` from the descriptions since it was only being used by the builders when creating the description. Since it wasn't used when verifying a transaction, it was too easy to add logic here that wouldn't get checked. The mint had a bug because of this - generator point was not being verified.
- Encapsulate all other verification logic into one function so as to not be a leaky abstraction

I think it's time to introduce a trait for descriptions, but I want to wait until the rust changes settle a bit so I'm not creating a ton of merge conflicts

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/transaction/mints.rs
+++ b/ironfish-rust/src/transaction/mints.rs
@@ -25,6 +25,8 @@ use crate::{
     SaplingKey,
 };
 
+use super::utils::verify_mint_proof;
+
 /// Parameters used to build a circuit that verifies an asset can be minted with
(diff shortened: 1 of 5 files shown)


rohanjadvani (approved):

dguenther:
Looks like some conflicts on this now 😬

PR #2758, opened 2022-12-16
Check isMinersFee when validating transactions

dguenther:
## Summary

isMinersFee has a few additional restrictions, but we didn't call isMinersFee as part of verification, so they weren't being enforced anywhere.

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):

ironfish/src/testUtilities/helpers/blockchain.ts
```diff
@@ -36,11 +43,25 @@ export async function makeBlockAfter(
...
+  const owner = key.public_address
+  const minerNote = new NativeNote(owner, BigInt(0), '', Asset.nativeIdentifier(), owner)
+  const transaction = new NativeTransaction(key.spending_key)
+  transaction.receive(minerNote)
+  const minerTransaction = new Transaction(transaction.post_miners_fee())
+
+  const noteCommitment = await chain.db.withTransaction(null, async (tx) => {
+    await chain.notes.addBatch(minerTransaction.notes, tx)
+    const commitment = await chain.notes.rootHash(tx)
+    await tx.abort()
```
danield9tqh, line 56: Not sure this will work unless we have all the previous miners fee transactions in note tree as well. This just adds the current one right? Definitely think we need to remove this `makeBlockAfter` function or find some way of making a long string of blocks without needing to update the DB every time (like in memory merkle tree for example). Also `makeBlockAfter` is used in a bunch of places so hard to get rid of

ironfish/src/consensus/verifier.test.ts
```diff
@@ -194,9 +260,11 @@ describe('Verifier', () => {
     const nodeTest = createNodeTest()
 
     it('says the block with no spends is valid', async () => {
-      const { chain, strategy } = nodeTest
-      strategy.disableMiningReward()
-      const block = await makeBlockAfter(chain, chain.head)
+      const { chain } = nodeTest
+      const block = await useMinerBlockFixture(chain, 2)
```
danield9tqh, line 264: can you remove this sequence?

danield9tqh (approved):

ironfish/src/consensus/verifier.test.ts
dguenther: Yep! Done

PR #2760, opened 2022-12-16
feat(ironfish): Pass `assetIdentifier` in `transaction/sendTransaction`

rohanjadvani:
## Summary

Add `assetIdentifier` as an argument to the RPC request and response

## Testing Plan

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/transactions/sendTransaction.ts
```diff
@@ -98,33 +102,41 @@ router.register<typeof SendTransactionRequestSchema, SendTransactionResponse>(
         publicAddress: receive.publicAddress,
         amount: CurrencyUtils.decode(receive.amount),
         memo: receive.memo,
-        assetIdentifier: Asset.nativeIdentifier(),
+        assetIdentifier: Buffer.from(receive.assetIdentifier, 'hex'),
       }
     })
 
     const fee = CurrencyUtils.decode(transaction.fee)
-    const sum = receives.reduce((m, c) => m + c.amount, fee)
```
mat-if, line 106: This line sets the initial value to the fee. I think we're missing the fee being added to the native asset amount in the new version

mat-if (approved):
Small thing about the fee but otherwise looks good

PR #2761, opened 2022-12-16
Clean up for sender address

ygao76:
## Breaking Change

```
[ ] Yes
```
@jowparks @mat-if


jowparks (approved):
LGTM

PR #2762, opened 2022-12-16
feat(cli,ironfish): Update balance CLI command to include `assetIdentifier`

rohanjadvani:
## Summary

Pass in asset identifier as a CLI flag for `accounts:balance`

## Testing Plan

Manually tested

```
Account: test
Balance: 76a45491b2e54ac9d1d6d0897daa7269d2b878b9cc44f8207d69f0516267e208 0.00000000
✨  Done in 2.31s.
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/utils/currency.ts
```diff
@@ -69,6 +69,20 @@ export class CurrencyUtils {
...
   }
+
+  static renderBalance(amount: bigint | string, assetIdentifier?: string): string {
+    if (typeof amount === 'string') {
+      amount = this.decode(amount)
+    }
+
+    const balance = FixedNumberUtils.render(amount, 8)
+
+    if (assetIdentifier) {
```
mat-if, line 80: What do you think about displaying $IRON if the native asset identifier? Might be able to fallback on renderIron() in this file or something

mat-if (approved):
Oops just saw I hadn't approved this. Whether we do the display changes in this PR or another is 🤷

ironfish/src/utils/currency.ts
```diff
@@ -69,6 +69,20 @@ export class CurrencyUtils {
 
     return ore
   }
+
+  static renderBalance(amount: bigint | string, assetIdentifier?: string): string {
```
ygao76, line 73: should the asset identifier be a enum type so its not free form text string

ironfish-cli/src/commands/accounts/balance.ts
```diff
@@ -41,32 +46,51 @@ export class BalanceCommand extends IronfishCommand {
   async start(): Promise<void> {
     const { flags, args } = await this.parse(BalanceCommand)
     const account = args.account as string | undefined
+    const assetIdentifier = flags.assetIdentifier ?? Asset.nativeIdentifier().toString('hex')
```
NullSoldier, line 49: This should be populated by the backend, not the front end. It should respect the balance from the node, node the CLI.

ironfish/src/utils/currency.ts
NullSoldier: no, it should be a string.

ironfish-cli/src/commands/accounts/balance.ts
```diff
@@ -78,24 +102,24 @@ export class BalanceCommand extends IronfishCommand {
...
     this.log(
-      `${response.pendingCount} notes worth ${CurrencyUtils.renderIron(
+      `${response.pendingCount} notes worth ${CurrencyUtils.renderBalance(
         pendingDelta,
-        true,
+        assetIdentifier,
       )} are waiting for miners to add them to the chain`,
     )
-    this.log(`Pending: ${CurrencyUtils.renderIron(pending, true)}`)
+    this.log(`Pending: ${CurrencyUtils.renderBalance(pending, assetIdentifier)}`)
```
NullSoldier, line 123: pending doesn't exist anymore

ironfish/src/utils/currency.ts
NullSoldier, line 73: The code above should be converted to use an optional assetIdentifier instead of adding this function. `Iron` and `Ore` refer to denominations rather than assets in the above context. It just defaults to the IRON ticket.

rohanjadvani: Yep, works for me

ironfish-cli/src/commands/accounts/balance.ts
rohanjadvani: Fixed

ironfish/src/utils/currency.ts
ygao76: okay but why

NullSoldier (approved):

PR #2767, opened 2022-12-19
feat(ci): Use 2 threads for rust tests to speed them up

mat-if:
## Summary

Draft because I want to run this a few times to see if it is a speed improvement and that it won't break anything

## Breaking Change

```
[ ] Yes
```


mat-if:
Looks like this is breaking, and wasn't any faster to begin with

PR #2769, opened 2022-12-19
refactor(rust): Use burn value directly for the binding signature of a transaction

mat-if:
## Summary

We were using the `ValueCommitment` which only makes sense when we are explicitly trying to not reveal the value. Since burns have a public value in their current form, we can use the value directly, much like the transaction fee

## Breaking Change

```
[ ] Yes
```


ironfish-rust/src/transaction/mod.rs
```diff
@@ -423,20 +431,19 @@ impl ProposedTransaction {
...
-            &binding_verification_key,
-            *self.value_balances.fee(),
-        )?;
+        let value_balance =
+            calculate_value_balance(&binding_verification_key, *self.value_balances.fee(), burns)?;
+
+        // Confirm that the public key derived from the binding signature key matches
+        // the final value balance point. The binding verification key is how verifiers
+        // check the consistency of the values in a transaction.
+        if value_balance != public_key.0 {
```
mat-if, line 444: Moved this check out of `check_value_consistency`, so that we could re-use the function in both the proposed transaction and the final transaction to de-dupe a bit of logic

ironfish-rust/src/transaction/mod.rs
```diff
@@ -707,25 +712,25 @@ fn value_balance_to_point(value: i64) -> Result<ExtendedPoint, IronfishError> {
...
+fn calculate_value_balance(
     binding_verification_key: &ExtendedPoint,
     value: i64,
-) -> Result<(), IronfishError> {
+    burns: &[BurnDescription],
+) -> Result<ExtendedPoint, IronfishError> {
     let value_balance_point = value_balance_to_point(value)?;
 
-    let calculated_public_key = binding_verification_key - value_balance_point;
+    let mut public_key = binding_verification_key - value_balance_point;
```
jowparks, line 726: Why is the balance called a `public_key`?

jowparks (approved):
LGTM

ironfish-rust/src/transaction/mod.rs
mat-if: Ah, that probably doesn't make sense anymore, or never really did I guess. Eventually it _is_ used as a public key (https://github.com/iron-fish/ironfish/pull/2769/files#diff-27fd1eb5ae888f12e53e1f310371941c240868ada1d0af6f33004d1f32acb74eR683), but naming it such here is not ideal. Will update it to something better

ironfish-rust/src/transaction/mod.rs
```diff
@@ -689,7 +694,7 @@ impl Transaction {
 
 /// Convert the integer value to a point on the Jubjub curve, accounting for
 /// negative values
-fn value_balance_to_point(value: i64) -> Result<ExtendedPoint, IronfishError> {
+fn fee_to_point(value: i64) -> Result<ExtendedPoint, IronfishError> {
```
mat-if, line 697: Renamed this function so it's more clear it should only be used by the transaction fee in it's current form

ironfish-rust/src/transaction/mod.rs
```diff
@@ -698,7 +703,7 @@ fn value_balance_to_point(value: i64) -> Result<ExtendedPoint, IronfishError> {
         None => return Err(IronfishError::IllegalValue),
     };
 
-    let mut value_balance = VALUE_COMMITMENT_VALUE_GENERATOR * jubjub::Fr::from(abs);
+    let mut value_balance = NATIVE_ASSET_GENERATOR * jubjub::Fr::from(abs);
```
mat-if, line 706: This is the same value - but we'll want to remove any usage of `VALUE_COMMITMENT_VALUE_GENERATOR` eventually and instead only rely on `NATIVE_ASSET_GENERATOR` (which is just a rename of `VALUE_COMMITMENT_VALUE_GENERATOR`) when working with the native asset - otherwise generator points should be using the asset's specific generator

rohanjadvani (approved):

PR #2775, opened 2022-12-20
Big int serialization

danield9tqh:
## Summary
Adding the latest `bufio` changes from [this PR](https://github.com/bcoin-org/bufio/pull/3) to be able to serialize 128-bit and 256-bit BigInts. Moving a few cases where we were already serializing fixed length BigInts to be serialized with `bufio`. 

## Testing Plan
Unit tests. Also tested equivalencies of bufio and current functions

## Breaking Change

Is this a breaking change? If yes, add notes below on why this is breaking and
what additional work is required, if any.

```
[ ] Yes
```

PR #2781, opened 2022-12-20
Replace all fixed sized BigInt serialization and deserialization with bufio

danield9tqh:
## Summary
Had a few more fixed sized BigInt serialization and deserialization to replace with bufio. Now that we've removed fixed size serialization with the `toBytes` function we can also remove the `size` parameter

## Testing Plan
Unit tests and testing equivalence of previous functions similar to [this PR](https://github.com/iron-fish/ironfish/pull/2775)

## Breaking Change

```
[ ] Yes
```


ironfish/src/workerPool/tasks/createMinersFee.ts
```diff
@@ -21,23 +21,23 @@ export class CreateMinersFeeRequest extends WorkerMessage {
 
   serialize(): Buffer {
     const bw = bufio.write(this.getSize())
-    bw.writeVarBytes(BigIntUtils.toBytes(this.amount))
+    bw.writeVarBytes(BigIntUtils.toBytesBE(this.amount))
```
NullSoldier, line 24: why are we using BE for this?

danield9tqh: This is just changing the name of `toBytes` to `toBytesBE` to be more explicit. Not changing any functionality. Although not really sure why we were using BE over LE here although seems somewhat arbitrary since it's just serializing a worker pool message

ironfish-rust-nodejs/index.js
```diff
@@ -11,8 +11,7 @@ function isMusl() {
   // For Node 10
   if (!process.report || typeof process.report.getReport !== 'function') {
     try {
-      const lddPath = require('child_process').execSync('which ldd').toString().trim();
-      return readFileSync(lddPath, 'utf8').includes('musl')
+      return readFileSync('/usr/bin/ldd', 'utf8').includes('musl')
```
dguenther, line 14: I think this may need to be reverted, I'm pretty sure it was added as part of an upgraded napi-rs version

ironfish/src/serde/PartialHeaderSerde.ts
```diff
@@ -14,8 +13,7 @@ export default class PartialBlockHeaderSerde {
     bw.writeHash(header.previousBlockHash)
     bw.writeHash(header.noteCommitment)
     bw.writeHash(header.transactionCommitment)
-    // TODO: Change to little-endian for consistency, since other non-bigint numbers are serialized as little-endian.
```
dguenther, line 17: Want to leave this comment until it's updated? Or do you think it should stay as BE

dguenther (changes requested):
Looks good! Just a couple comments

ironfish/src/serde/PartialHeaderSerde.ts
danield9tqh: I think it should stay as BE. It seemed to me like the target is not exactly a BigInt. It's more of a target hash value that the has of the block should be less than. For that reason I thought it made sense to serialize it in BE so that the bytes in the serialized format would be largest to smallest left to right to mirror a potential hash value. I don't think it makes too much of a difference but it did kind of make sense to me that it was BE

danield9tqh: Agreed that we will change the network and database serializations to be BE as well for consistency

dguenther (approved):

PR #2783, opened 2022-12-20
removes unused wallet code for syncing transactions

hughy:
## Summary

removes syncTransaction from wallet and account.

removes SyncTransactionParams.

removes bulkUpdateDecryptedNotes and processTransactionSpends which were only called from syncTransaction.

removes updateDecryptedNote and replaces remaining uses with calls to saveDecryptedNote.

removes updateTransaction and replaces remaining uses with calls to saveTransaction.

removes unnecessary join with transactions in getUnspentNotes; uses decryptedNote.sequence for block confirmations.

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/account.ts
```diff
@@ -548,7 +395,8 @@ export class Account {
             'nullifierToNote mappings must have a corresponding decryptedNote',
           )
 
-          await this.updateDecryptedNote(
+          await this.walletDb.saveDecryptedNote(
```
danield9tqh, line 398: So I understand this correctly. The only real logic change here is that we are no longer attempting to delete the nullifier for these notes. And that is because spends will always have nullifiers so no reason to ever delete them?

ironfish/src/wallet/wallet.ts
```diff
@@ -957,7 +903,7 @@ export class Wallet {
 
     if (noteMapValue) {
       this.logger.debug(`Unspent note has index ${String(noteMapValue.index)}`)
-      await sender.updateDecryptedNote(unspentNote.hash, {
+      await this.walletDb.saveDecryptedNote(sender, unspentNote.hash, {
```
danield9tqh, line 906: Same for this change.. We already do an early return if there are no nullifiers so no need to check deletion of nullifier here either?

ironfish/src/wallet/wallet.ts
```diff
@@ -676,20 +631,11 @@ export class Wallet {
...
-
-        Assert.isNotUndefined(
-          transaction,
-          `Transaction '${decryptedNote.transactionHash.toString(
-            'hex',
-          )}' missing for account '${account.id}'`,
-        )
-
-        if (!transaction.sequence) {
+        if (!decryptedNote.sequence) {
```
danield9tqh, line 634: I know this was done in another PR but curious why we moved blockHash/sequence to the note since it seems like a denormalization if the note also has the transactionHash

danield9tqh (approved):

ironfish/src/wallet/account.ts
hughy: Yeah exactly, the spends will still have nullifiers, and those would be removed only if the transaction is disconnected from the chain

ironfish/src/wallet/wallet.ts
hughy: We added blockHash and sequence to the note for performance -- avoiding looking up the transaction to get the sequence for a note

ironfish/src/wallet/wallet.ts
hughy: Yeah we know that the note has a nullifier already, so we wouldn't delete the index entry in this case

PR #2784, opened 2022-12-20
feat(ci): Fetch cargo-tarpaulin binary from github instead of using github action

mat-if:
## Summary

The tarpaulin github action that we were using seemed to have some kind of flake when fetching info from the github API. It also seems possibly unmaintained. So, let's just do the simple thing without layers of abstraction causing difficulty - fetch the built binary and "install" it.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/.github/workflows/rust_ci.yml
+++ b/.github/workflows/rust_ci.yml
@@ -31,6 +31,7 @@ jobs:
       - name: Cache Rust
         uses: Swatinem/rust-cache@v2
         with:
+          save-if: false
           shared-key: base
 
       - name: Check that cargo lockfile is up to date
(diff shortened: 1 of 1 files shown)


.github/workflows/rust_ci.yml
```diff
@@ -102,12 +104,13 @@ jobs:
...
           shared-key: zkp
 
-      # Build & Run test & Collect Code coverage
-      - name: Run cargo-tarpaulin on ironfish-zkp
-        uses: actions-rs/tarpaulin@v0.1
-        with:
-          version: "0.22.0"
-          args: --avoid-cfg-tarpaulin --manifest-path ironfish-zkp/Cargo.toml --release -- --test-threads 1
+      # Run tests to collect code coverage
+      - name: Run cargo-tarpaulin on ironfish-rust
```
mat-if, line 108:
```suggestion
      - name: Run cargo-tarpaulin on ironfish-zkp
```
💀 Well I wanted to trigger a rebuild, so I guess that gives me a good reason to.
I want to turn these into a matrix but it will require a bit more tedious config fiddling and I'm about tapped out on that for the day.

.github/workflows/rust_ci.yml
```diff
@@ -31,6 +31,7 @@ jobs:
       - name: Cache Rust
         uses: Swatinem/rust-cache@v2
         with:
+          save-if: false
```
mat-if, line 34: This was causing the "base" keyed cache to not get updated since this would finish before the other job, so the lint job just won't save.

dguenther (approved):

dguenther:
This is great, thanks for cleaning up this flaky action!

PR #2790, opened 2022-12-20
Transaction Ordering By Timestamp

mirayadav:
## Summary

Added support for iterating database entries by sorted key order. Added a timestamp field to `TransactionValue` and a new index `timestampToTransactionHash`. Used these changes to list transactions in `accounts:transaction` from most to least recent. 

## Testing Plan

Added a new test for checking if database sorted iteration is working correctly. Updated relevant tests.

## Breaking Change

```
[] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/getTransactions.ts
+++ b/ironfish/src/rpc/routes/wallet/getTransactions.ts
@@ -25,6 +25,7 @@ export type GetAccountTransactionsResponse = {
   mintsCount: number
   burnsCount: number
   expiration: number
+  timestamp: number
 }
 
 export const GetAccountTransactionsRequestSchema: yup.ObjectSchema<GetAccountTransactionsRequest> =
@@ -49,6 +50,7 @@ export const GetAccountTransactionsResponseSchema: yup.ObjectSchema<GetAccountTr
       mintsCount: yup.number().defined(),
       burnsCount: yup.number().defined(),
       expiration: yup.number().defined(),
+      timestamp: yup.number().defined(),
     })
     .defined()
 
@@ -145,7 +147,7 @@ const handleLimitedTransactions = async (
     }
   })
 
-  for await (const transaction of account.getTransactions()) {
+  for await (const transaction of account.getSortedTransactions()) {
     Assert.isNotNull(transaction)
     queue.add(transaction)
     // remove the earliest transaction when queue is full
@@ -170,7 +172,7 @@ const handleAllTransactions = async (
   account: Account,
   headSequence?: number | null,
 ): Promise<void> => {
-  for await (const transaction of account.getTransactions()) {
+  for await (const transaction of account.getSortedTransactions()) {
     if (request.closed) {
       break
     }
--- a/ironfish/src/rpc/routes/wallet/getTransactions.ts
+++ b/ironfish/src/rpc/routes/wallet/getTransactions.ts
@@ -147,7 +147,7 @@ const handleLimitedTransactions = async (
     }
   })
 
-  for await (const transaction of account.getOrderedTransactions()) {
+  for await (const transaction of account.getTransactionsByTime()) {
     Assert.isNotNull(transaction)
     queue.add(transaction)
     // remove the earliest transaction when queue is full
@@ -172,7 +172,7 @@ const handleAllTransactions = async (
   account: Account,
   headSequence?: number | null,
 ): Promise<void> => {
-  for await (const transaction of account.getOrderedTransactions()) {
+  for await (const transaction of account.getTransactionsByTime()) {
     if (request.closed) {
       break
     }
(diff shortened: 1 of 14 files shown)


ironfish/src/storage/levelup/store.ts
```diff
@@ -75,19 +75,32 @@ export class LevelupStore<Schema extends DatabaseSchema> extends DatabaseStore<S
     transaction?: IDatabaseTransaction,
     keyRange?: DatabaseKeyRange,
   ): AsyncGenerator<[SchemaKey<Schema>, SchemaValue<Schema>]> {
-    if (keyRange) {
-      keyRange = StorageUtils.addPrefixToRange(keyRange, this.prefixBuffer)
+    if (keyRange?.gt || keyRange?.gte || keyRange?.lt || keyRange?.lte) {
+      keyRange = {
+        ...keyRange,
+        ...StorageUtils.addPrefixToRange(keyRange, this.prefixBuffer),
+      }
```
hughy, line 82: I think `StorageUtils.addPrefixToRange` should change so that it propagates `ordered` and `reverse` if they're set.

ironfish/src/storage/levelup/store.ts
```diff
@@ -104,16 +117,40 @@ export class LevelupStore<Schema extends DatabaseSchema> extends DatabaseStore<S
...
+          ? key.compare(cacheElements[currElement].key) < 0
+          : key.compare(cacheElements[currElement].key) > 0)
+      ) {
+        yield [this.decodeKey(cacheElements[currElement].key), cacheElements[currElement].value]
+        currElement++
+      }
+
+      if (currElement < cacheElements.length && key.equals(cacheElements[currElement].key)) {
+        yield [this.decodeKey(cacheElements[currElement].key), cacheElements[currElement].value]
+        currElement++
```
hughy, line 143: Could we get rid of this block if we changed the compare functions above to use `<= 0` and `>= 0`?

hughy (commented):
I pulled this down to test with a local node, but I'm not seeing transactions in order: 
image

I think it could be an issue with LevelUp not iterating over the key range in order 🤔

ironfish/src/wallet/walletdb/walletdb.ts
```diff
@@ -295,6 +307,11 @@ export class WalletDB {
       }
 
       await this.transactions.put([account.prefix, transactionHash], transactionValue, tx)
+      await this.timestampToTransactionHash.put(
+        [account.prefix, transactionValue.timestamp.getTime()],
```
NullSoldier, line 311: nit: above you use `transactionValue.timestamp.valueOf()` and here you use `getTime()`

mirayadav:
> I pulled this down to test with a local node, but I'm not seeing transactions in order: image
> 
> I think it could be an issue with LevelUp not iterating over the key range in order 🤔

This should be fixed now (was caused by the fact that the key encoding was LE so they weren't actually sorted in order).

ironfish/src/rpc/routes/wallet/getTransactions.ts
```diff
@@ -141,7 +143,7 @@ const handleLimitedTransactions = async (
     }
   })
 
-  for await (const transaction of account.getTransactions()) {
+  for await (const transaction of account.getOrderedTransactions()) {
```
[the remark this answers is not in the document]
mirayadav: Makes sense, going to leave this for a future PR.

ironfish/src/storage/levelup/store.ts
```diff
@@ -75,19 +75,32 @@ export class LevelupStore<Schema extends DatabaseSchema> extends DatabaseStore<S
...
+      keyRange = {
+        ...keyRange,
+        ...StorageUtils.addPrefixToRange(keyRange, this.prefixBuffer),
+      }
     } else {
-      keyRange = this.allKeysRange
+      keyRange = keyRange ? { ...keyRange, ...this.allKeysRange } : this.allKeysRange
     }
 
+    const cacheElements = []
```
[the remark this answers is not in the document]
mirayadav: Pulled out this work into a different PR and made the change there - https://github.com/iron-fish/ironfish/pull/2799

ironfish/src/storage/levelup/store.ts
mirayadav: Made a change in another PR that removes the need for this change - https://github.com/iron-fish/ironfish/pull/2798

ironfish/src/storage/levelup/store.ts
mirayadav: Pulled out this work into a different PR and made the change there - https://github.com/iron-fish/ironfish/pull/2799

ironfish/src/wallet/walletdb/walletdb.ts
mirayadav: Changed to use `getTIme()` everywhere

NullSoldier (approved):

NullSoldier:
This is going to improve UX a lot for users. Thank you for your contributions.

PR #2791, opened 2022-12-20
Fix submittedSequence in connectTransaction

mirayadav:
## Summary

Get `submittedSequence` from the existing, stored `transactionValue` if it exists, otherwise set it to `null`. This change is needed because `submittedSequence` should be the sequence when the transaction was submitted and not the block's sequence.

## Breaking Change

```
[ ] Yes
```


hughy (approved):
Looks good! Thanks!

PR #2792, opened 2022-12-21
adds chain state fields to balances datastore

hughy:
## Summary

instead of storing only a bigint with the asset balance as the value in the balances store, stores an object containing the balance and the block hash and sequence of the most recent balance update.

now that we only update the balance when blocks are connected and disconnected from the chain the stored balance value corresponds to a specific chain state. storing this chain state alongside the balance will allow us to inform users of what chain state their balance reflects. a number of users were confused by balances while accounts were scanning during phase 2: their transactions would be on the chain, but the transaction wouldn't be reflected in their account balance because the account was behind the chain.

- defines BalanceValue type to include balance, blockHash, and sequence
- replaces value in balances datastore with BalanceValue
- propagates BalanceValue through all reads/writes on balances store

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/account.ts
+++ b/ironfish/src/wallet/account.ts
@@ -11,6 +11,7 @@ import { DatabaseKeyRange, IDatabaseTransaction } from '../storage'
 import { StorageUtils } from '../storage/database/utils'
 import { DecryptedNote } from '../workerPool/tasks/decryptNotes'
 import { AccountValue } from './walletdb/accountValue'
+import { BalanceValue } from './walletdb/balanceValue'
 import { DecryptedNoteValue } from './walletdb/decryptedNoteValue'
 import { TransactionValue } from './walletdb/transactionValue'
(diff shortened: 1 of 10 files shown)


ironfish/src/wallet/account.ts
```diff
@@ -447,7 +456,7 @@ export class Account {
   }> {
     let unconfirmedCount = 0
 
-    const unconfirmed = await this.getUnconfirmedBalance(assetIdentifier, tx)
+    const { balance: unconfirmed } = await this.getUnconfirmedBalance(assetIdentifier, tx)
```
NullSoldier, line 459: makes me think we should just have this as unconfirmed so its clear when looking at the model itself its unconfirmed

NullSoldier (approved):

ironfish/src/wallet/account.ts
hughy: Agreed, changed it to unconfirmed!

PR #2793, opened 2022-12-21
Display asset identifier for accounts:notes cli

ygao76:
## Testing Plan
Screen Shot 2022-12-21 at 4 52 05 PM

## Breaking Change

```
[ ] Yes
```
@rohanjadvani @mat-if


ironfish-cli/src/commands/accounts/notes.ts
```diff
@@ -37,9 +37,12 @@ export class NotesCommand extends IronfishCommand {
         [note],
         {
           amount: {
-            header: 'Amount ($IRON)',
+            header: 'Amount',
             get: (row) => CurrencyUtils.renderIron(row.amount),
           },
+          assetIdentifier: {
```
NullSoldier, line 43: This isn't that valuable we should show the name here and not the identifier. It's fine to keep this but keep it hidden by default like we do in Peers list.

ironfish/src/rpc/routes/chain/getAssetInfo.ts
```diff
@@ -0,0 +1,73 @@
...
+ 
+ router.register<typeof GetAssetInfoRequestSchema, GetAssetInfoResponse>(
+   `${ApiNamespace.chain}/getBlockInfo`,
+   GetAssetInfoRequestSchema,
+   async (request, node): Promise<void> => {
+     let header: BlockHeader | null = null
+     let error = ''
+ 
+     if (request.data.assetIdentifier.length != ASSET_IDENTIFIER_LENGTH) {
+        throw new ValidationError("Invalid asset identifier.")
```
NullSoldier, line 53: This isn't valid the asset length is byte length not encoded hex string length.

ironfish/src/rpc/routes/chain/getAssetInfo.ts
```diff
@@ -0,0 +1,73 @@
...
+     }
+
+     let assetIdentifier = Buffer.from(request.data.assetIdentifier, 'hex')
+ 
+     const asset = await node.chain.assets.get(assetIdentifier)  
+ 
+     Assert.isNotUndefined(asset, 'no asset is found')
```
NullSoldier, line 60: Asserts are not for expected paths. Render a view error here with a nice error.

ironfish/src/rpc/routes/chain/getAssetInfo.ts
```diff
@@ -0,0 +1,73 @@
...
+
+     request.end({
+        createdTransactionHash: BufferUtils.toHuman(asset.createdTransactionHash),
```
NullSoldier, line 63: Hashes should not be rendered into human format. It should always be hex.

ironfish/src/rpc/routes/chain/getAssetInfo.ts
```diff
@@ -0,0 +1,73 @@
...
+        identifier: BufferUtils.toHuman(asset.identifier),
+        metadata: BufferUtils.toHuman(asset.metadata),
+        name: BufferUtils.toHuman(asset.name),
+        nonce: Number(asset.nonce),
+        owner: BufferUtils.toHuman(asset.owner),
+        supply: Number(asset.supply),
```
NullSoldier, line 69: Don't cast currency to number as this is lossy. Use CurrencyUtils.encode

ironfish/src/rpc/routes/chain/getAssetInfo.ts
```diff
@@ -0,0 +1,73 @@
...
+ 
+     Assert.isNotUndefined(asset, 'no asset is found')
+
+     request.end({
+        createdTransactionHash: BufferUtils.toHuman(asset.createdTransactionHash),
+        identifier: BufferUtils.toHuman(asset.identifier),
+        metadata: BufferUtils.toHuman(asset.metadata),
+        name: BufferUtils.toHuman(asset.name),
+        nonce: Number(asset.nonce),
+        owner: BufferUtils.toHuman(asset.owner),
```
NullSoldier, line 68: This is a key not a human rebderable item

ironfish/src/rpc/routes/wallet/getNotes.ts
```diff
@@ -48,8 +50,13 @@ router.register<typeof GetAccountNotesStreamRequestSchema, GetAccountNotesStream
       const transaction = await account.getTransaction(transactionHash)
       Assert.isNotUndefined(transaction)
 
+      const asset = await node.chain.assets.get(note.assetIdentifier())  
+
+      Assert.isNotUndefined(asset)
+
       request.stream({
         amount: note.value().toString(),
+        assetIdentifier: asset.name.toString('hex') || "$IRON",
```
NullSoldier, line 59: This is a bug your passing the name as an identifier and hexing it.

ironfish/src/rpc/routes/wallet/getNotes.ts
```diff
@@ -48,8 +50,13 @@ router.register<typeof GetAccountNotesStreamRequestSchema, GetAccountNotesStream
       const transaction = await account.getTransaction(transactionHash)
       Assert.isNotUndefined(transaction)
 
+      const asset = await node.chain.assets.get(note.assetIdentifier())  
+
+      Assert.isNotUndefined(asset)
```
NullSoldier, line 55: You can't guarantee the asset is found because you could be using this during a reorg when an asset has been deleted.

ironfish/src/rpc/routes/chain/getAssetInfo.ts
```diff
@@ -0,0 +1,73 @@
...
+     const asset = await node.chain.assets.get(assetIdentifier)  
+ 
+     Assert.isNotUndefined(asset, 'no asset is found')
+
+     request.end({
+        createdTransactionHash: BufferUtils.toHuman(asset.createdTransactionHash),
+        identifier: BufferUtils.toHuman(asset.identifier),
+        metadata: BufferUtils.toHuman(asset.metadata),
+        name: BufferUtils.toHuman(asset.name),
+        nonce: Number(asset.nonce),
```
NullSoldier, line 67: This is already a number, why is it being casted?

ironfish-cli/src/commands/accounts/notes.ts
mat-if: The name isn't unique, therefore should not be used as a source of truth in situations like this. The identifier is the only way to truthfully show which asset is which - much like smart contracts have to rely on their addresses

ironfish/src/rpc/routes/chain/getAssetInfo.ts
```diff
@@ -0,0 +1,73 @@
...
+     let assetIdentifier = Buffer.from(request.data.assetIdentifier, 'hex')
+ 
+     const asset = await node.chain.assets.get(assetIdentifier)  
+ 
+     Assert.isNotUndefined(asset, 'no asset is found')
+
+     request.end({
+        createdTransactionHash: BufferUtils.toHuman(asset.createdTransactionHash),
+        identifier: BufferUtils.toHuman(asset.identifier),
+        metadata: BufferUtils.toHuman(asset.metadata),
```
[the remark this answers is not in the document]
mat-if: It's a field for users to put any extra info they want about an asset - there's no enforcement as to what they put in here besides a max length. I expect it to be primarily UTF8 strings

ironfish-cli/src/commands/chain/asset.ts
```diff
@@ -0,0 +1,34 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { IronfishCommand } from '../../command'
```
rohanjadvani, line 4: This seems like a separate change. We can remove this and keep it in a separate PR

ironfish-cli/src/commands/accounts/notes.ts
rohanjadvani: Agreed with @mat-if. I'd probably show both

ironfish/src/rpc/routes/wallet/getNotes.ts
rohanjadvani: +1

ironfish/src/rpc/routes/wallet/getNotes.ts
rohanjadvani: +1

rohanjadvani (changes requested):

ironfish-cli/src/commands/accounts/notes.ts
NullSoldier: My comment was to show both. The CLI here should show the name, but nothing accepts the name so they'll have to get the identifier and pass it in.

ironfish/src/rpc/routes/chain/getAssetInfo.ts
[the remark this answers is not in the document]
NullSoldier: Got it, that makes sense. We should hex encode this then and toHuman() this on the front end then or else devs cant get ths machine readable data here since toHuman() is lossy.

ironfish-cli/src/commands/chain/asset.ts
NullSoldier: Yes these 2 PRs shouldn't be conflated. Split this out into a different PR.

ironfish-cli/src/commands/accounts/notes.ts
ygao76: I will show both.

ironfish/src/utils/blockchain.ts
```diff
@@ -40,6 +42,23 @@ export function getBlockRange(
   return { start, stop }
 }
 
+export const NATIVE_ASSET_VALUE = {
+  createdTransactionHash: GENESIS_BLOCK_PREVIOUS,
+  identifier: Asset.nativeIdentifier(),
+  metadata: Buffer.from("Native asset of Iron Fish blockchain", 'utf8'),
+  name: Buffer.from("$IRON", 'utf8'),
+  nonce: 0,
```
ygao76, line 50: @rohanjadvani What value I should put for nonce?

ironfish/src/utils/blockchain.ts
```diff
@@ -40,6 +42,23 @@ export function getBlockRange(
...
+  owner: Buffer.from("Iron Fish", 'utf8'),
+  supply: BigInt(0),
```
ygao76, line 52: @rohanjadvani what is the supply number for IRON ?

rohanjadvani: This is defined in our [consensus parameters](https://github.com/iron-fish/ironfish/blob/master/ironfish/src/consensus/consensus.ts#L15), but I would omit this as mentioned above

ironfish/src/rpc/routes/wallet/getNotes.ts
```diff
@@ -48,8 +53,12 @@ router.register<typeof GetAccountNotesStreamRequestSchema, GetAccountNotesStream
       const transaction = await account.getTransaction(transactionHash)
       Assert.isNotUndefined(transaction)
 
+      const asset = await getAssetById(note.assetIdentifier(), node.chain)
```
rohanjadvani, line 56: It feels weird to have a utility function which takes the chain as an argument. What do you think about:
1. Moving `getAssetByIdentifier` to a method under the blockchain
2. Checking if the `note.assetIdentifier()` is the native identifier here in the RPC command
3. If (2) is false, call the method in (1)

rohanjadvani (changes requested):

ironfish/src/rpc/routes/wallet/getNotes.ts
NullSoldier: I agree with #1: We shouldn't have this hidden util function it should be `chain.getAssetById` or something like that.

I don't understand the suggestion in #2, what does that mean? We should always call getAssetById() and don't do carvouts.

ironfish/src/utils/blockchain.ts
NullSoldier: I don't know, I would prefer this type to match Asset or else it has to return Asset | typeof NATIVE_ASSET_VALUE

ironfish/src/utils/blockchain.ts
NullSoldier: As a note, the consensus parameters can't match this since the supply changes with every block mined.

ironfish/src/utils/blockchain.ts
```diff
@@ -40,6 +42,26 @@ export function getBlockRange(
...
+
+export async function getAssetById(
+  assetIdentifier: Buffer,
+  chain: Blockchain,
+): Promise<AssetsValue | undefined> {
+  if (Asset.nativeIdentifier().equals(assetIdentifier)) {
+    return NATIVE_ASSET_VALUE
+  }
+  return await chain.assets.get(assetIdentifier)
+}
```
NullSoldier, line 63: Agree with rohan, I expect this to be on block chain itself like all the other getter functions

ironfish/src/utils/blockchain.ts
```diff
@@ -40,6 +42,26 @@ export function getBlockRange(
   return { start, stop }
 }
 
+export const NATIVE_ASSET_VALUE = {
```
NullSoldier, line 45: This should move to primitives/asset and not a util file.

ironfish/src/rpc/routes/wallet/getNotes.ts
```diff
@@ -48,8 +53,12 @@ router.register<typeof GetAccountNotesStreamRequestSchema, GetAccountNotesStream
...
+
       request.stream({
         amount: note.value().toString(),
```
NullSoldier, line 59: This is mixing encodings in our code base, use `currencyUtils.encode()`

ironfish/src/rpc/routes/wallet/getNotes.ts
ygao76: I discussed with Mat about this, I dont like step 2. Currently this is the only place to do the condition check, but as we add more features in wallet for multi assets, there will be more read operation, where we need to do this check. I want to avoid to have the check for any read logic. As a util function, its a uniformed interface for `getAssetByIdentifier` so all the future read can use this as well.

This is the main reason Mat suggested to have the function in util instead under blockchain class.

ironfish-cli/src/commands/accounts/notes.ts
```diff
@@ -37,9 +37,15 @@ export class NotesCommand extends IronfishCommand {
         [note],
         {
           amount: {
-            header: 'Amount ($IRON)',
+            header: 'Amount',
```
NullSoldier, line 40: It doesn't the denomination now, but I guess that's ok. $IRON here is denomination rather than the currency but this is fine.

ironfish/src/rpc/routes/wallet/getNotes.ts
NullSoldier: I still don't understand step #2, but It should definitely go in blockchain like all other functions. I don't know what the wallet has to do with this, as they will both have a function like `wallet.getAssetById()` and `chain.getAssetById()` the code won't be shared because they won't even run on the same computer.

ygao76: I will move the function to blockchain.

ironfish/src/rpc/routes/wallet/getNotes.ts
NullSoldier, line 59: This is an existing bug I just noticed from the old code. This should be `CurrencyUtils.encode(note.value())`

NullSoldier: I'll fix it in another PR

ironfish/src/utils/blockchain.ts
ygao76: what supply value I should put for Iron?

ironfish/src/utils/blockchain.ts
ygao76: Return `Asset | typeof NATIVE_ASSET_VALUE` for the getAssetById

ygao76:
Moved the function to blockchain. Ready to review!

NullSoldier (approved):

PR #2794, opened 2022-12-21
Replace hex-array library with native (de)serialization

danield9tqh:
## Summary
Was doing some performance testing and realized that `hex-array` library is much slower (~7x) than native buffer serialization/deserialization from a hex string. The native code is also much more concise so removing this library in favor of that.

## Testing Plan
Unit tests and wrote tests to check equivalences of the native code and `hex-array` library

## Breaking Change

```
[ ] Yes
```


danield9tqh:
Tested out performance with the following code and native code came out to ~7x faster on average
```javascript
let start = Date.now()
for (const _ of [...new Array(100000)]) {
  const str = hexArray.toString(randomBytes(32))
  Buffer.from(hexArray.fromString(str))
}
console.log('hex-array', Date.now() - start)

start = Date.now()
for (const _ of [...new Array(100000)]) {
  const str = randomBytes(32).toString('hex')
  Buffer.from(str, 'hex')
}
console.log('native', Date.now() - start)
```

Also tested equivalence of the two functions with
```javascript
for (const _ of [...new Array(100000)]) {
  const bytes = randomBytes(32)

  const str1 = hexArray.toString(bytes)
  const str2 = bytes.toString('hex').toUpperCase()

  expect(str1).toEqual(str2)
}
```

dguenther (approved):

PR #2797, opened 2022-12-21
Add sources to transactions state in transaction fetcher

danield9tqh:
## Summary
A transaction's 'sources' are peers that we can request that full transactions from. We were storing this list of peers in a separate map of TransactionHash -> sources but it makes sense to store them in the map of TransactionHash -> TransactionState. This removes some unrepresentable states. For example when we've received a transaction and it is `PROCESSING` we no longer care about who we can request it from.

## Testing Plan
Unit tests

## Breaking Change

```
[ ] Yes
```


dguenther:
I'll take a look at this today.

dguenther (approved):

PR #2799, opened 2022-12-21
Add support for iterating database entries by key order

mirayadav:
## Summary

Add support for iterating database entries by key order (levelDB default iteration is by sorted key order but our transaction cache didn't support this)

## Testing Plan

Added tests in `database.test.ts`

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/storage/levelup/store.ts
+++ b/ironfish/src/storage/levelup/store.ts
@@ -3,6 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import type { LevelupDatabase } from './database'
 import { BufferSet } from 'buffer-map'
+import FastPriorityQueue from 'fastpriorityqueue'
 import MurmurHash3 from 'imurmurhash'
 import { Assert } from '../../assert'
 import { AsyncUtils } from '../../utils/async'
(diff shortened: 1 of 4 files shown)


hughy (approved):

ironfish/src/storage/levelup/store.ts
```diff
@@ -106,56 +111,85 @@ export class LevelupStore<Schema extends DatabaseSchema> extends DatabaseStore<S
           continue
         }
 
-        yield [this.decodeKey(key), value as SchemaValue<Schema>]
+        if (iteratorOptions?.ordered) {
```
NullSoldier, line 114: This should always be true right?

ironfish/src/storage/levelup/store.ts
```diff
@@ -84,6 +85,10 @@ export class LevelupStore<Schema extends DatabaseSchema> extends DatabaseStore<S
     }
 
     const seen = new BufferSet()
+    const cacheElements = new FastPriorityQueue<{ key: Buffer; value: SchemaValue<Schema> }>(
+      ({ key: a }, { key: b }) =>
+        iteratorOptions?.reverse ? b.compare(a) < 0 : a.compare(b) < 0,
```
NullSoldier, line 90: I think this differs from the interleaving comparater

PR #2801, opened 2022-12-21
Remove buildRawTransaction()

NullSoldier:
## Summary

Remove use of buildRawTransaction() method

## Breaking Change

```
[ ] Yes
```


ironfish/src/blockchain/blockchain.test.ts
```diff
@@ -1240,60 +1240,58 @@ describe('Blockchain', () => {
     })
 
     describe('when spending and burning the same note in a block', () => {
-      it('throws an exception', async () => {
+      it('fails validation as double spend', async () => {
```
NullSoldier, line 1243: no exception is thrown

rohanjadvani (approved):
LGTM, but it looks like there's an old fixture around because the Jest test description was changed. Can we remove that (or re-generate the whole fixture file)?

PR #2804, opened 2022-12-22
Remove useRawTxFixture() fixture

NullSoldier:
## Summary
The ultimate goal here is to remove
 - WorkerPool.createTransaction()
 - Fake witness creation in MASP tests
 - buildRawTransaction() in tx test helpers

## Testing Plan
This is only for tests, so just run the tests.

## Breaking Change

```
[ ] Yes
```


rohanjadvani (approved):
LGTM, but same comment on other PR about fixtures

PR #2806, opened 2022-12-22
Change RPC note.amount to be value

NullSoldier:
## Summary

This renamed a field on the note incorrectly

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/accounts/notes.ts
```diff
@@ -36,9 +36,9 @@ export class NotesCommand extends IronfishCommand {
       CliUx.ux.table(
         [note],
         {
-          amount: {
+          value: {
             header: 'Amount ($IRON)',
```
hughy, line 40: Should we change the header while we're at it?

hughy (approved):

PR #2809, opened 2023-01-03
Don't save latest until transaction commit

danield9tqh:
## Summary
A while back a change was made to do block database updates and block verification in parallel. This works because saving the block is done within an atomic transaction and if block verification later fails then the transaction can be rolled back. There are a few in-memory updates however that are not done within the database transaction and would not be rolled back such as updating the `chain.head` and `chain.latest` in memory objects. This PR fixes a potential bug where chain.latest can be updated outside of a database transaction. 

## Testing Plan
Unit tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -701,6 +701,11 @@ export class Blockchain {
 
     await tx.update()
     this.notes.pastRootTxCommitted(tx)
+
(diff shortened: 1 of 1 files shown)


NullSoldier:
Looks good. Luckily latest shouldn't be used for anything other than debugging but good to make this consitent.

NullSoldier (approved):

PR #2810, opened 2023-01-03
Add useMintBlockFixture for tests

NullSoldier:
## Summary
This new fixture is useful to create a block with a mint in it

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/testUtilities/fixtures.ts
+++ b/ironfish/src/testUtilities/fixtures.ts
@@ -15,7 +15,7 @@ import { SerializedTransaction, Transaction } from '../primitives/transaction'
 import { IJSON } from '../serde'
 import { Account, AccountValue, Wallet } from '../wallet'
 import { WorkerPool } from '../workerPool/pool'
-import { buildRawTransaction, createRawTransaction } from './helpers/transaction'
+import { createRawTransaction } from './helpers/transaction'
 import { getCurrentTestPath } from './utils'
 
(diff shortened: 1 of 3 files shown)


ironfish/src/testUtilities/fixtures.ts
```diff
@@ -301,15 +324,34 @@ export async function useBlockWithRawTxFixture(
   sequence: number,
 ): Promise<Block> {
   const generate = async () => {
-    const transaction = await buildRawTransaction(
```
NullSoldier, line 304: Inlined this function so no one else will use it while im trying to remove it.

ironfish/src/testUtilities/fixtures.ts
```diff
@@ -223,6 +223,29 @@ export async function useMinerBlockFixture(
   )
 }
 
+export async function useMintBlockFixture(options: {
```
hughy, line 226: With this fixture and `usePostTxFixture` we'll have added a couple that use an options object instead of a list of parameters. Should we make a point of updating the other fixtures to use `options` for consistency?

hughy (approved):

ironfish/src/testUtilities/fixtures.ts
NullSoldier: yes, im gonna do a full pass on that and clean up all the fixtures in one PR soon.

PR #2819, opened 2023-01-04
adds balance delta mapping to transactionValue

hughy:
## Summary

users currently cannot see how much they sent or received in each transaction listed in the CLI. the only way to present this information to the user with the current walletDb schema would be to load each note and spend for each transaction each time we look up the transaction.

stores the balance delta for each asset on the value in the wallet transactions store.

serializes buffer map as a list of key-value pairs prepended with the length of the list. keys are 32-byte asset identifiers and values are bigints.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/walletdb/transactionValue.ts
+++ b/ironfish/src/wallet/walletdb/transactionValue.ts
@@ -47,7 +47,7 @@ export class TransactionValueEncoding implements IDatabaseEncoding<TransactionVa
     const assetCount = value.assetBalanceDeltas.size
     bw.writeU32(assetCount)
 
-    for (const [assetIdentifier, balanceDelta] of value.assetBalanceDeltas.entries()) {
+    for (const [assetIdentifier, balanceDelta] of value.assetBalanceDeltas) {
       bw.writeHash(assetIdentifier)
       bw.writeBigU64(balanceDelta)
     }
--- a/ironfish/src/wallet/walletdb/transactionValue.ts
+++ b/ironfish/src/wallet/walletdb/transactionValue.ts
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import type { IDatabaseEncoding } from '../../storage/database/types'
+import { ASSET_IDENTIFIER_LENGTH } from '@ironfish/rust-nodejs'
+import { BufferMap } from 'buffer-map'
 import bufio from 'bufio'
 import { Transaction } from '../../primitives'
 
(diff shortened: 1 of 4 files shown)


ironfish/src/wallet/walletdb/transactionValue.ts
```diff
@@ -41,6 +44,14 @@ export class TransactionValueEncoding implements IDatabaseEncoding<TransactionVa
       bw.writeU32(sequence)
     }
 
+    const assetCount = value.assetBalanceDeltas.size
+    bw.writeU32(assetCount)
+
+    for (const [assetIdentifier, balanceDelta] of value.assetBalanceDeltas.entries()) {
```
NullSoldier, line 50: the default iterator for buffer set is entries, so you don't need the `.entries()` call here

ironfish/src/wallet/walletdb/transactionValue.ts
```diff
@@ -41,6 +44,14 @@ export class TransactionValueEncoding implements IDatabaseEncoding<TransactionVa
       bw.writeU32(sequence)
     }
 
+    const assetCount = value.assetBalanceDeltas.size
+    bw.writeU32(assetCount)
```
NullSoldier, line 48: I see why you did this but it's clear this encoding this BufferSet.

```suggestion
    bw.writeU32(value.assetBalanceDeltas.size)
```

NullSoldier, line 48: I see why you did this but it's clear this encoding this BufferSet.

```suggestion

    bw.writeU32(value.assetBalanceDeltas.size)
```

ironfish/src/wallet/walletdb/transactionValue.ts
```diff
@@ -41,6 +44,14 @@ export class TransactionValueEncoding implements IDatabaseEncoding<TransactionVa
...
+
+    for (const [assetIdentifier, balanceDelta] of value.assetBalanceDeltas.entries()) {
+      bw.writeHash(assetIdentifier)
```
NullSoldier, line 51: I don't think this will work, you need to use writeBytes with ASSET_IDENTIFIER. Hash assumes 32 bytes but ASSET_IDENTIFIER is 64 bytes.

ironfish/src/wallet/walletdb/transactionValue.ts
```diff
@@ -85,6 +112,8 @@ export class TransactionValueEncoding implements IDatabaseEncoding<TransactionVa
     if (value.sequence) {
       size += 4
     }
+    size += 4
+    size += value.assetBalanceDeltas.size * (ASSET_IDENTIFIER_LENGTH + 8)
```
NullSoldier, line 116: ASSET_IDENTIFIER_LENGTH is in bits but these are bytes right?

hughy: I wasn't sure, but it looks like ASSET_IDENTIFIER_LENGTH is 32 bytes: https://github.com/iron-fish/ironfish/blob/38d2528903d3def6be0eaa4eaf400deb541b404c/ironfish-zkp/src/constants.rs#L12

NullSoldier (approved):

PR #2824, opened 2023-01-04
Upgrading oclif/core to 1.23.1

NullSoldier:
## Summary

This adds a fix I commited to oclif/core to fix the prompt return type released in 1.20.1 here, https://github.com/oclif/core/releases/tag/1.20.1

```
> ironfish fees

Fee Rates ($ORE/kB)
low:    1
medium: 1
high:   1
```

## Breaking Change

```
[ ] Yes
```


jowparks (approved):
No tests for these, so I would just say add some output for commands that ran successfully :)

NullSoldier:
Since this only changes typescript, and not any code that gets included in the compiled bundle you can guarantee that the output has not changed as long as it builds.

PR #2795, opened 2022-12-21
feat: wire up multi asset syncer to API (2/2)

jowparks:
These changes are dependent on https://github.com/iron-fish/ironfish/pull/2835

## Summary
1. Wires up the asset work with the API and CLI multi asset syncer.
2. Adds tests for `getTransactionStream`
3. renames code from Masp -> multi asset

Related PRs in the API:
https://github.com/iron-fish/ironfish-api/pull/1156
https://github.com/iron-fish/ironfish-api/pull/1157

## Testing Plan
TODO: Write tests, also test in stage
## Breaking Change

Is this a breaking change? If yes, add notes below on why this is breaking and
what additional work is required, if any.

```
[ ] Yes
```

PR #2835, opened 2023-01-05
feat: update get transaction stream (1/2)

jowparks:
## Summary
Updates `getTransactionStream` to handle multi asset

Followup changes to multi-asset syncer will be done here:
https://github.com/iron-fish/ironfish/pull/2795
## Testing Plan
new unit tests
## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/chain/getTransactionStream.ts
```diff
@@ -13,30 +14,66 @@ import { ValidationError } from '../../adapters/errors'
...
   hash: string
   isMinersFee: boolean
   notes: Note[]
+  mints: Mint[]
+  burns: Burn[]
 }
 
 const NoteSchema = yup
   .object()
   .shape({
```
jowparks, line 43: I looked around code, this doesn't seem like a breaking change anywhere (other than masp syncer, which will be updated in subsequent PR). Just want to verify this rename from `amount`->`value` will be acceptable. This rename will make all our naming consistent in this rpc.

ironfish/src/rpc/routes/chain/getTransactionStream.ts
```diff
@@ -106,28 +143,55 @@ router.register<typeof GetTransactionStreamRequestSchema, GetTransactionStreamRe
...
+
+        for (const mint of tx.mints) {
+          mints.push({
+            value: CurrencyUtils.encode(mint.value),
+            assetId: mint.asset.identifier().toString('hex'),
+            assetName: mint.asset.name().toString(),
+          })
+        }
+
+        if (notes.length || burns.length || mints.length) {
```
jowparks, line 188: Verifying this conditional update makes sense, I think it does, but want to verify

ironfish/src/rpc/routes/chain/getTransactionStream.ts
```diff
@@ -106,28 +143,55 @@ router.register<typeof GetTransactionStreamRequestSchema, GetTransactionStreamRe
...
         for (const note of tx.notes) {
           const decryptedNote = note.decryptNoteForOwner(request.data.incomingViewKey)
 
           if (decryptedNote) {
+            const assetValue = await validatedAssetValue(decryptedNote.assetIdentifier())
             notes.push({
-              amount: CurrencyUtils.encode(decryptedNote.value()),
+              value: CurrencyUtils.encode(decryptedNote.value()),
               memo: decryptedNote.memo(),
+              assetId: decryptedNote.assetIdentifier.toString(),
```
NullSoldier, line 165: hex hex hex

ironfish/src/rpc/routes/chain/getTransactionStream.ts
```diff
@@ -106,28 +143,55 @@ router.register<typeof GetTransactionStreamRequestSchema, GetTransactionStreamRe
...
+              assetName: assetValue.name.toString(),
```
NullSoldier, line 166: hex hex hex

ironfish/src/rpc/routes/chain/getTransactionStream.ts
```diff
@@ -106,28 +143,55 @@ router.register<typeof GetTransactionStreamRequestSchema, GetTransactionStreamRe
...
           }
         }
 
-        if (notes.length) {
+        for (const burn of tx.burns) {
+          const assetValue = await validatedAssetValue(burn.assetIdentifier)
+          burns.push({
+            value: CurrencyUtils.encode(burn.value),
+            assetId: burn.assetIdentifier.toString('hex'),
+            assetName: assetValue.name.toString(),
```
NullSoldier, line 176: hex

ironfish/src/rpc/routes/chain/getTransactionStream.ts
```diff
@@ -106,28 +143,55 @@ router.register<typeof GetTransactionStreamRequestSchema, GetTransactionStreamRe
...
+          })
+        }
+
+        for (const mint of tx.mints) {
+          mints.push({
+            value: CurrencyUtils.encode(mint.value),
+            assetId: mint.asset.identifier().toString('hex'),
+            assetName: mint.asset.name().toString(),
```
NullSoldier, line 184: hex

ironfish/src/rpc/routes/chain/getTransactionStream.ts
NullSoldier: yes, if any of these have entries, add the TX to the list of returned ones.

ironfish/src/rpc/routes/chain/getTransactionStream.ts
```diff
@@ -147,20 +211,30 @@ router.register<typeof GetTransactionStreamRequestSchema, GetTransactionStreamRe
       })
     }
 
+    const validatedAssetValue = async (assetIdentifier: Buffer): Promise<AssetsValue> => {
+      const assetValue = await node.chain.assets.get(assetIdentifier)
+      if (!assetValue) {
```
NullSoldier, line 216: you should use node.chain.getAssetById and not this store because it handles the default asset which isnt in the DB yet.

ironfish/src/rpc/routes/chain/getTransactionStream.ts
```diff
@@ -147,20 +211,30 @@ router.register<typeof GetTransactionStreamRequestSchema, GetTransactionStreamRe
...
+        throw new ValidationError(
+          `Asset detected in chain is not in assetDB. Asset Identifier: ${assetIdentifier.toString()}`,
+        )
+      }
```
NullSoldier, line 220: This will crash incorrectly in the case that an asset has removed, we need to remove this error and handle this properly to avoid errors / crashes. Let's think about this case carefully

ironfish/src/rpc/routes/chain/getTransactionStream.ts
jowparks: Good call

NullSoldier (approved):

PR #2838, opened 2023-01-05
Add RPC for posting a raw transaction.

EvanJRichard:
## Summary
This PR proposes to add a new rpc endpoint `wallet/postTransaction`. It accepts the hex of a raw transaction, submits it to the node, and returns the hex of the posted transaction. This is similar to `wallet/sendTransaction`, but accepting a pre-constructed transaction, rather than building one from passed options.

## Testing Plan
Added `postTransaction.test.ts`, which passes.

## Breaking Change

```
[ ] Yes
[X] No
```


hairtail:
"transaction/sendTransaction" route is used for post raw transaction now.

ironfish/src/rpc/routes/wallet/postTransaction.test.ts
```diff
@@ -0,0 +1,47 @@
...
+
+describe('Route wallet/postTransaction', () => {
+  const routeTest = createRouteTest(true)
+
+  // TODO remove this when it's not needed for copying anymore
+  it('should return account status information', async () => {
+    const account = await routeTest.node.wallet.createAccount(uuid(), true)
+    const response = await routeTest.client
+      .request<any>('wallet/getAccountsStatus', {})
+      .waitForEnd()
```
NullSoldier, line 19:
```suggestion
    const response = await routeTest.client.getAccountsStatus({})

```

ironfish/src/rpc/routes/wallet/postTransaction.test.ts
```diff
@@ -0,0 +1,47 @@
...
+          headHash: 'NULL',
+          headInChain: false,
+          sequence: 'NULL',
+        },
+      ],
+    })
+  })
+
+  // TODO a valid raw transaction
+  it('should return the posted transaction', async () => {
```
NullSoldier, line 36: You can use this, https://github.com/iron-fish/ironfish/blob/staging/ironfish/src/testUtilities/helpers/transaction.ts#L28-L64

EvanJRichard:
@hairtail good point, updating description to clarify difference

ironfish/src/rpc/routes/wallet/postTransaction.ts
```diff
@@ -0,0 +1,37 @@
...
+  })
+  .defined()
+
+export const PostTransactionResponseSchema: yup.ObjectSchema<PostTransactionResponse> = yup
+  .object({
+    transaction: yup.string().defined(),
+  })
+  .defined()
+
+router.register<typeof PostTransactionRequestSchema, PostTransactionResponse>(
```
NullSoldier, line 26: This is missing on the client, we should add a function on the RpcClient as well like all the other rpcs.

ironfish/src/rpc/routes/wallet/postTransaction.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    const options = {
+      wallet: routeTest.node.wallet,
+      from: account,
+    }
+    const rawTransaction = await createRawTransaction(options)
+    const response = await routeTest.client
+      .request<any>('wallet/postTransaction', {
+        transaction: RawTransactionSerde.serialize(rawTransaction).toString('hex'),
+      })
+      .waitForEnd()
```
NullSoldier, line 28: Once we add this to the client we can do this...

```suggestion
    const response = await routeTest.client.postTransaction({
        transaction: RawTransactionSerde.serialize(rawTransaction).toString('hex'),
      })
```

hairtail:
Actually, i think this rpc is useless. To get a hex-encoded transaction, we need 
1. unspent notes to create the raw transaction
2. create transaction locally maybe
But, we cann't get unspent notes without an ironfish node now. So the only way to create a transaction is on the node side. If so, we never need this route since we are already on the node side.

hairtail:
Ok, seems that ironfish is about to support transaction creation locally. Sorry for that

> Actually, i think this rpc is useless. To get a hex-encoded transaction, we need
> 
> 1. unspent notes to create the raw transaction
> 2. create transaction locally maybe
>    But, we cann't get unspent notes without an ironfish node now. So the only way to create a transaction is on the node side. If so, we never need this route since we are already on the node side. I mean, if we already use the node for getting unspent notes, why not creating and sending the transaction with node directly?

NullSoldier:
This will be useful when we have createRawTransaction() RPC

ironfish/src/rpc/routes/wallet/postTransaction.test.ts
```diff
@@ -0,0 +1,45 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
NullSoldier, line 6: Do we still need these? They disable the compiler type checking.

ironfish/src/rpc/routes/wallet/postTransaction.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  })
+
+  it("should return an error if the transaction won't deserialize", async () => {
+    // const response = await routeTest.client.postTransaction({
+    //   transaction: '0xdeadbeef',
+    // })
+
+    // await expect(response).rejects.toMatchObject({
+    //   status: 400,
+    // })
```
hughy, line 35: Commented-out code left behind?

hughy (approved):
Looks great!

My only comment is about some commented-out code left in the tests

NullSoldier:
This PR is good, it just needs to be rebased and fixed because changes just merged in that break this from building.

NullSoldier (approved):

ironfish/src/rpc/routes/wallet/postTransaction.test.ts
EvanJRichard: Yes, removed :+1:

PR #2839, opened 2023-01-05
displays balance head hash and sequence in CLI

hughy:
## Summary

the account balance now includes chain state information (blockHash and sequence). includes this chain state in the balance CLI output so that it's clearer where on the chain the balance is reflective of.

## Testing Plan

image

image

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/balance.ts
+++ b/ironfish-cli/src/commands/wallet/balance.ts
@@ -3,7 +3,6 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { CurrencyUtils, GetBalanceResponse } from '@ironfish/sdk'
 import { Flags } from '@oclif/core'
-import { resolve } from 'path'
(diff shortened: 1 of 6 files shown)


ironfish-cli/src/commands/accounts/balance.ts
```diff
@@ -82,6 +84,8 @@ export class BalanceCommand extends IronfishCommand {
     const unconfirmedDelta = unconfirmed - confirmed
 
     this.log(`Account: ${response.account}`)
+    this.log(`Head Hash: ${response.blockHash || 'NULL'}`)
+    this.log(`Head Sequence: ${response.sequence || 'NULL'}`)
```
NullSoldier, line 88: Can this be in human format?

`Your balance is calculated on the chain at a block ${hash} at sequence {sequence}` or something like it?

NullSoldier (approved):

PR #2840, opened 2023-01-05
feat(rust): use `checked_add` for `ValueBalance`

mat-if:
## Summary

Right now, its possible to add enough value to a transaction via mints
or spends such that it would overflow the `i64` type in the
`ValueBalance` struct. By using `checked_add`, we get some cleaner error
handling around this logic. Note that this doesn't functionally change
much, it's just more explicit error handling.

As a future improvement, we could use an
arbitrarily large value type for `ValueBalance` like `i128`, and change
the logic on `Transaction::post` to loop over the values and create
multiple change notes, allowing value transfers of greater than
`u64::MAX`, if we ever needed to.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust-nodejs/src/structs/transaction.rs
+++ b/ironfish-rust-nodejs/src/structs/transaction.rs
@@ -169,19 +169,27 @@ impl NativeTransaction {
 
     /// Create a proof of a new note owned by the recipient in this transaction.
     #[napi]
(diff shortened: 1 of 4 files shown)


rohanjadvani (approved):

NullSoldier (approved):

NullSoldier:
https://doc.rust-lang.org/std/primitive.u64.html#method.checked_add

PR #2847, opened 2023-01-06
feat(cli,ironfish): Update mint to accept an optional identifier

rohanjadvani:
## Summary

Update the CLI / RPC to optionally accept an asset identifier or metadata + name.

## Testing Plan

Unit tests

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/wallet/mint.ts
```diff
@@ -88,23 +85,24 @@ export class Mint extends IronfishCommand {
...
+        name: flags.name,
+        value: flags.amount,
       })
 
       stopProgressBar()
 
       const response = result.content
       this.log(`
- Minted asset ${name} from ${account}
+ Minted asset ${flags.name} from ${flags.account}
```
NullSoldier, line 100: this should be in the result right?

mat-if (approved):

PR #2851, opened 2023-01-07
feat(rust,zkp): Remove nonce from asset info

mat-if:
## Summary

Builds on #2850 

This was necessary due to old blake hash setup for asset id. Since switching to pedersen hash, we don't need this since a pedersen hash always generates a valid point

## Breaking Change

```
[ ] Yes
```


ironfish-rust/src/assets/asset.rs
```diff
@@ -58,23 +54,19 @@ impl Asset {
...
     }
 
     fn new_with_raw_data(
         owner: PublicAddress,
         name: [u8; NAME_LENGTH],
         metadata: [u8; METADATA_LENGTH],
-        nonce: u8,
     ) -> Result<Asset, IronfishError> {
-        let capacity = METADATA_LENGTH + NAME_LENGTH + PUBLIC_ADDRESS_SIZE + 1;
+        let capacity = METADATA_LENGTH + NAME_LENGTH + PUBLIC_ADDRESS_SIZE;
```
rohanjadvani, line 65: Maybe we can do this to prevent duplicate lengths

```suggestion
        let capacity = ASSET_LENGTH;
```

rohanjadvani (approved):

PR #2852, opened 2023-01-07
removes miners fee transactions from forked blocks

hughy:
## Summary

minersFee transactions have no expiration so they will remain in a pending state forever unless removed. there is no need to persist the minersFee transactions form a fork in the wallet: if they are added to the main chain again in a re-org then we will re-sync them.

- defines removeTransaction to execute the same logic as expireTransaction, but deletes the transaction from the database
- cleans up some unnecessary code in expireTransaction and deleteDecryptedNote

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/account.ts
+++ b/ironfish/src/wallet/account.ts
@@ -342,12 +342,13 @@ export class Account {
     })
   }
 
-  async removeTransaction(transaction: Transaction, tx?: IDatabaseTransaction): Promise<void> {
+  async deleteTransaction(transaction: Transaction, tx?: IDatabaseTransaction): Promise<void> {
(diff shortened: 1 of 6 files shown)


ironfish/src/wallet/account.ts
```diff
@@ -342,20 +342,34 @@ export class Account {
     })
   }
 
+  async removeTransaction(transaction: Transaction, tx?: IDatabaseTransaction): Promise<void> {
```
NullSoldier, line 345: - account.removeTransaction
 - walletDb.deleteTransaction

NullSoldier (approved):

Issue #2811, opened 2023-01-03
Feature: Wait on an account being deleted


NullSoldier:
### Description

Since we support async account deletion, we should add a way to wait on an account to be fully deleted in the case of security and needing to ensure all account data is cleaned up.

NullSoldier:
These are being assigned interally right now, I'm deleting all of these feature requests so they don't get picked up.

PR #2853, opened 2023-01-07
Wait on an account being deleted

hairtail:
## Summary
Feature request https://github.com/iron-fish/ironfish/issues/2811
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/remove.ts
+++ b/ironfish-cli/src/commands/wallet/remove.ts
@@ -23,16 +23,22 @@ export class RemoveCommand extends IronfishCommand {
     confirm: Flags.boolean({
       description: 'Suppress the confirmation prompt',
     }),
+    wait: Flags.boolean({
+      default: false,
+      allowNo: true,
+      description: 'Wait for account remove finishes',
+    }),
   }
 
   async start(): Promise<void> {
     const { args, flags } = await this.parse(RemoveCommand)
     const confirm = flags.confirm
+    const wait = flags.wait
     const name = args.account as string
 
     const client = await this.sdk.connectRpc()
 
-    const response = await client.removeAccount({ name, confirm })
+    const response = await client.removeAccount({ name, confirm, wait })
 
     if (response.content.needsConfirm) {
       const value = await CliUx.ux.prompt(`Are you sure? Type ${name} to confirm`)
@@ -42,7 +48,11 @@ export class RemoveCommand extends IronfishCommand {
         this.exit(1)
       }
 
-      await client.removeAccount({ name, confirm: true })
+      if (wait) {
+        this.log(`Waiting for account delete finished ...`)
+      }
+
+      await client.removeAccount({ name, confirm: true, wait })
     }
 
     this.log(`Account '${name}' successfully removed.`)
--- a/ironfish-cli/src/commands/wallet/remove.ts
+++ b/ironfish-cli/src/commands/wallet/remove.ts
@@ -26,7 +26,7 @@ export class RemoveCommand extends IronfishCommand {
     wait: Flags.boolean({
       default: false,
       allowNo: true,
-      description: 'Wait for account remove finishes',
+      description: 'Wait for account removal to finish',
     }),
   }
 
@@ -49,7 +49,7 @@ export class RemoveCommand extends IronfishCommand {
       }
 
       if (wait) {
-        this.log(`Waiting for account delete finished ...`)
+        this.log(`Waiting for account deletion to finish . . .`)
       }
 
       await client.removeAccount({ name, confirm: true, wait })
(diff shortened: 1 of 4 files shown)


ironfish/src/wallet/walletdb/walletdb.ts
```diff
@@ -761,6 +761,39 @@ export class WalletDB {
     await this.pendingTransactionHashes.clear(tx, account.prefixRange)
   }
 
+  async cleanupDeletedAccount(accountId: string, signal?: AbortSignal): Promise<void> {
```
hughy, line 764: This duplicates most of the code in `cleanupDeletedAccounts`. Could we update one to call the other and avoid the duplication?

ironfish/src/rpc/routes/wallet/removeAccount.ts
```diff
@@ -49,7 +50,7 @@ router.register<typeof RemoveAccountRequestSchema, RemoveAccountResponse>(
       }
     }
 
-    await node.wallet.removeAccount(account.name)
+    await node.wallet.removeAccount(account.name, request.data.wait)
```
hughy, line 53: This might cause the CLI to seem unresponsive -- waiting on the account data to be removed might take a long time, and no response is returned until it's finished

That might be ok, and I think the only alternative would be to create a streaming endpoint that returns some progress/status messages

hairtail: I added a notify message in `removeAccount cli` for now.

ironfish/src/wallet/walletdb/walletdb.ts
```diff
@@ -778,6 +781,9 @@ export class WalletDB {
     ]
 
     for (const [accountId] of await this.accountIdsToCleanup.getAll()) {
+      if (accountIdToDelete && accountIdToDelete !== accountId) {
+        continue
+      }
```
hughy, line 786: This will work, but it looks a bit odd to have this conditional logic.

I was thinking that the logic in this loop could move to `cleanupDeletedAccount` so that the plural cleanup function calls the singular cleanup function for each account.

ironfish-cli/src/commands/wallet/remove.ts
```diff
@@ -42,7 +48,11 @@ export class RemoveCommand extends IronfishCommand {
         this.exit(1)
       }
 
-      await client.removeAccount({ name, confirm: true })
+      if (wait) {
+        this.log(`Waiting for account delete finished ...`)
```
hughy, line 52:
```suggestion
        this.log(`Waiting for account deletion to finish . . .`)
```

ironfish-cli/src/commands/wallet/remove.ts
```diff
@@ -23,16 +23,22 @@ export class RemoveCommand extends IronfishCommand {
     confirm: Flags.boolean({
       description: 'Suppress the confirmation prompt',
     }),
+    wait: Flags.boolean({
+      default: false,
+      allowNo: true,
+      description: 'Wait for account remove finishes',
```
hughy, line 29:
```suggestion
      description: 'Wait for account removal to finish',
```

ironfish-cli/src/commands/wallet/remove.ts
```diff
@@ -42,7 +48,11 @@ export class RemoveCommand extends IronfishCommand {
         this.exit(1)
       }
 
-      await client.removeAccount({ name, confirm: true })
+      if (wait) {
+        this.log(`Waiting for account deletion to finish . . .`)
+      }
```
NullSoldier, line 53: You should use oclif status spinner thing we use in other places to indicate a long action is being performed.  You'll need to just check what happens if it throws inbetween starting and ending the action though. It should be fine, but just throw a demo error and check the output is correct.

https://github.com/iron-fish/ironfish/blob/master/ironfish-cli/src/commands/chain/prune.ts#L80-L82

ironfish/src/wallet/wallet.ts
```diff
@@ -1201,7 +1204,11 @@ export class Wallet {
       return
     }
 
-    await this.walletDb.cleanupDeletedAccounts(this.eventLoopAbortController.signal)
+    if (accountId) {
+      await this.walletDb.cleanupDeletedAccount(accountId, this.eventLoopAbortController.signal)
```
NullSoldier, line 1208: This doesn't work as it does a mark and sweep amount of cleanup. Also these will now run in parallel with the event loop causing them to waste CPU and fight each other.

hairtail: Get it. Is there any other idea for this "wait for deletion finish" purpose?

NullSoldier: That's the entire feature request were asking to be built. If you cannot figure out how to implement it please don't pick up that feature request.

NullSoldier:
Closing this PR as it doesn't work and the author has not made more attempts to fix the issues.

PR #2864, opened 2023-01-09
feat(ironfish): Add simple validation for mints and burns

mat-if:
## Summary

Added validation `verifyBlock` and `verifyCreatedTransaction` to cover what I believe are the important code-paths for transaction verification. `verifyBlock` covers when new blocks are created and when new blocks are received from the peer network. `verifyCreatedTransaction` covers when the wallet creates new transactions and when new transactions are received from the peer network.

Mints validate that the given asset has a human-readable name set. Burns validate that the given asset is not the native asset.

## Breaking Change

```
[ ] Yes
```


ironfish/src/consensus/verifier.ts
```diff
@@ -446,26 +490,28 @@ export enum VerificationResultReason {
   DOUBLE_SPEND = 'Double spend',
   DUPLICATE = 'Duplicate',
   ERROR = 'Error',
-  GRAFFITI = 'Graffiti field is not 32 bytes in length',
   GOSSIPED_GENESIS_BLOCK = 'Peer gossiped its genesis block',
+  GRAFFITI = 'Graffiti field is not 32 bytes in length',
   HASH_NOT_MEET_TARGET = 'Hash does not meet target',
+  INVALID_ASSET_NAME = 'Asset name is blank',
+  INVALID_BURN = 'Attempting to burn the native asset',
```
rohanjadvani, line 497: This is a nit so feel free to ignore, but I wonder if we should change the enum to be a bit more explicit since this error is specific to the native asset. I'm unsure about what the name should be though ¯\_(ツ)_/¯

rohanjadvani (approved):

ironfish/src/consensus/verifier.ts
mat-if: Good call - renamed to `NATIVE_BURN`

PR #2866, opened 2023-01-09
Add instructions to upgrading Node version

danield9tqh:
## Summary
When users are told to upgrade to Node v18 they will also need to reinstall the `ironfish` package. This adds a message with those instructions. 

## Testing Plan
Tested locally

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/bin/run
+++ b/ironfish-cli/bin/run
@@ -8,7 +8,8 @@ if (process.platform !== 'win32') {
 }
 
 if(process.versions.node.split('.')[0] !== '18') {
-  console.log('NodeJS version ' +  process.versions.node + ' is not compatible. Must have node v18.x installed: https://nodejs.org/en/download/')
+  console.log('NodeJS version ' +  process.versions.node + ' is not compatible. Must have Node v18 installed: https://nodejs.org/en/download/')
+  console.log('After v18 is installed, MAKE SURE TO run `npm install -g ironfish` again to install ironfish with the correct Node version')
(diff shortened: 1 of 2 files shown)


NullSoldier (approved):

ironfish-cli/bin/run
```diff
@@ -8,7 +8,8 @@ if (process.platform !== 'win32') {
 }
 
 if(process.versions.node.split('.')[0] !== '18') {
-  console.log('NodeJS version ' +  process.versions.node + ' is not compatible. Must have node v18.x installed: https://nodejs.org/en/download/')
+  console.log('NodeJS version ' +  process.versions.node + ' is not compatible. Must have Node v18 installed: https://nodejs.org/en/download/')
+  console.log('After v18 is installed, MAKE SURE TO run `npm install -g ironfish` again to install ironfish with the correct Node version')
```
dguenther, line 12: Maybe a text change? Don't feel strongly about it

```suggestion
  console.log('After v18 is installed, run `npm install -g ironfish` again to install ironfish for the correct Node version')
```

dguenther (approved):

ironfish-cli/bin/run
danield9tqh: I put the MAKE SURE TO because they will have to remember to do it after going down the node18 install path. If they don’t they’ll end up with a confusing memory error saying something like node dependencies are out of date. So I think it’s worth emphasizing

PR #2867, opened 2023-01-09
feat(ironfish): Add getBalances method to wallet and account

rohanjadvani:
## Summary

Introduce a method to the account and wallet to fetch balances across all assets

## Testing Plan

Unit test

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/wallet.ts
```diff
@@ -511,6 +511,47 @@ export class Wallet {
...
+      unconfirmedCount,
+    } of account.getBalances(headSequence, minimumBlockConfirmations)) {
+      yield {
+        assetId,
+        blockHash,
+        confirmed,
+        sequence,
+        unconfirmed,
+        unconfirmedCount,
+      }
```
NullSoldier, line 551: we don't need to destructure it if were just gonna yield it again

NullSoldier (approved):

PR #2868, opened 2023-01-10
Fix test fails

hairtail:
## Summary
`getHeadHash` has been renamed as `getHead` in https://github.com/iron-fish/ironfish/commit/901b4ff2ccedb1ee2a3ef32a7eed1db9d5873932
## Breaking Change

```
[ ] Yes
```


mat-if (approved):
Good catch, thanks!

PR #2874, opened 2023-01-10
reads head sequence at account level in balance methods

hughy:
## Summary

reading the account head at the wallet level is unnecessary since its only use there is to be passed down to the account.

- updates getBalance and getBalances to read account head in the account-level methods
- removes unnecessary db transaction in getBalance

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/account.ts
+++ b/ironfish/src/wallet/account.ts
@@ -511,8 +511,8 @@ export class Account {
     const head = await this.getHead()
     if (!head) {
       return {
-        unconfirmed: BigInt(0),
-        confirmed: BigInt(0),
+        unconfirmed: 0n,
+        confirmed: 0n,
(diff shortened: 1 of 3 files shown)


ironfish/src/wallet/account.ts
```diff
@@ -505,10 +508,21 @@ export class Account {
     blockHash: Buffer | null
     sequence: number | null
   }> {
+    const head = await this.getHead()
+    if (!head) {
+      return {
+        unconfirmed: BigInt(0),
+        confirmed: BigInt(0),
```
NullSoldier, line 515: You can use literal syntax too if you want

```suggestion
        unconfirmed: 0n,
        confirmed: 0n,
```

NullSoldier (approved):

PR #2876, opened 2023-01-10
filters by sequence in getUnspentNotes at account level

hughy:
## Summary

the wallet getUnspentNotes method filters all unspent notes that the account level method returns by the account's head sequence.

moving this filtering down to the account level method puts all of the filtering logic for unspent notes in one place.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/account.ts
+++ b/ironfish/src/wallet/account.ts
@@ -94,7 +94,19 @@ export class Account {
 
   async *getUnspentNotes(
     assetId: Buffer,
+    options?: {
+      minimumBlockConfirmations?: number,
+    }
   ): AsyncGenerator<DecryptedNoteValue & { hash: Buffer }> {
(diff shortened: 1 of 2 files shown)


ironfish/src/wallet/account.ts
```diff
@@ -104,7 +116,7 @@ export class Account {
         continue
       }
 
-      if (!decryptedNote.index) {
+      if (!decryptedNote.sequence || decryptedNote.sequence > maxConfirmedSequence) {
```
NullSoldier, line 119: nit: Can we make this its own condition? It's kind of a style thing but I think it helps clearly understand each of the cases. Here we say if it's not on the chain, skip it, then next  we check that its in the confirmation range.

hughy: That sounds good to me! I'll make the change

PR #2878, opened 2023-01-10
Remove create transaction in worker pool

ygao76:
## Summary
- Replace fixture create transaction with raw transaction and post()
- Remove createTransaction
- Use size constants for hardcoded values
## Testing Plan
unit test
## Breaking Change

```
[ ] Yes
```


ironfish-rust-nodejs/index.d.ts
```diff
@@ -33,6 +33,10 @@ export interface NativeSpendDescription {
   nullifier: Buffer
 }
 export const PROOF_LENGTH: number
+export const TRANSACTION_SIGNATURE_LENGTH: number
```
NullSoldier, line 36: Can you split all of these changes out into another PR? They are not related to removing `workerPool.createTransaction()` right and they are fairly opinionated so mat or rohan should review them separately.

NullSoldier: That sounds good, but it's still unrelated to this PR. It's not needed to accomplish the PR task. It should be pretty easy to split this out into a different PR.

rohanjadvani: Yeah @ygao76 I think this is a sensible change; I'm all for removing magic numbers. But as @NullSoldier mentioned, we want to keep the code changes in this PR focused on just the worker pool. I'd recommend keeping this refactoring but just cherry picking those changes out in a new PR.

ygao76:
PTAL @NullSoldier

NullSoldier (approved):

NullSoldier:
Thanks, looks good!

PR #2881, opened 2023-01-10
Fix references to account DB

NullSoldier:
## Summary

Fixing places that still refer to to the wallet DB as account DB

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/backup.ts
+++ b/ironfish-cli/src/commands/backup.ts
@@ -28,12 +28,12 @@ export default class Backup extends IronfishCommand {
     chain: Flags.boolean({
       default: true,
       allowNo: true,
-      description: 'Export the chain DB',
+      description: 'Backup the chain DB',
     }),
-    accounts: Flags.boolean({
+    wallet: Flags.boolean({
       default: false,
       allowNo: true,
-      description: 'Export the accounts',
+      description: 'Backup the wallet',
     }),
     accessKeyId: Flags.string({
       char: 'a',
@@ -96,7 +96,7 @@ export default class Backup extends IronfishCommand {
       excludes.push(path.basename(path.dirname(this.sdk.config.chainDatabasePath)))
     }
 
-    if (!flags.accounts) {
+    if (!flags.wallet) {
       excludes.push(path.basename(path.dirname(this.sdk.config.walletDatabasePath)))
     }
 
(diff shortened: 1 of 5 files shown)


ironfish-cli/src/commands/reset.ts
```diff
@@ -54,7 +54,7 @@ export default class Reset extends IronfishCommand {
 
     const message =
       '\nYou are about to destroy your node databases. The following directories and files will be deleted:\n' +
-      `\nAccounts: ${accountDatabasePath}` +
```
ygao76, line 57: Can you change `Accounts` in the message?

ironfish/src/node.ts
```diff
@@ -246,7 +246,7 @@ export class IronfishNode {
     })
 
     const accountDB = new WalletDB({
```
ygao76, line 248: variable name is still accountDB

ironfish-cli/src/commands/backup.ts
```diff
@@ -97,7 +97,7 @@ export default class Backup extends IronfishCommand {
     }
 
     if (!flags.accounts) {
```
ygao76, line 99: flag variable name is still accounts

ironfish-cli/src/commands/reset.ts
```diff
@@ -45,7 +45,7 @@ export default class Reset extends IronfishCommand {
       this.exit(0)
```
ygao76, line 45: the warning message still has accounts.

ironfish-cli/src/commands/reset.ts
NullSoldier: Fixed

ironfish-cli/src/commands/reset.ts
NullSoldier: Fixed

ironfish-cli/src/commands/backup.ts
NullSoldier: Fixed

ironfish/src/node.ts
NullSoldier: Fixed

NullSoldier:
I changed the description on the PR now since we're catching more spots. Thanks for catching these other instances of account db.

rohanjadvani (approved):

PR #2883, opened 2023-01-11
moves isExpiredSequence off of Verifier

hughy:
## Summary

the logic of isExpiredSequence is very simple and is independent of any other state (i.e., it doesn't depend on the chain)

since this method is currently defined on the Verifier class it has a dependeny on the chain.

the wallet uses isExpiredSequence in several places. moving isExpiredSequence into a utils module removes a source of chain dependence in the wallet

## Breaking Change

```
[ ] Yes
```


jowparks (approved):
LGTM if tests pass, thanks for detangling

PR #2885, opened 2023-01-11
feat(cli,ironfish): Add `wallet:balances` command

rohanjadvani:
## Summary

* Update request `account` to be optional
* Return `account` name
* Add CLI command

## Testing Plan

```sh
λ ironfish wallet:balances

Account: default
 Asset Name Asset Id                                                         Confirmed Balance
 ────────── ──────────────────────────────────────────────────────────────── ─────────────────
 rohan      0d69819630f69b17b0c566bd258868e1796baee360f4839fdf1e90df35881aea 0.00000000
 $IRON      d7c86706f5817aa718cd1cfad03233bcd64a7789fd9422d3b17af6823a7e6ac6 60.00000000
✨  Done in 5.32s.
```

```sh
λ ironfish wallet:balances --all

Account: default
 Asset Name Asset Id                                                         Confirmed Balance Unconfirmed Balance Head Hash                                                        Head Sequence
 ────────── ──────────────────────────────────────────────────────────────── ───────────────── ─────────────────── ──────────────────────────────────────────────────────────────── ─────────────
 rohan      0d69819630f69b17b0c566bd258868e1796baee360f4839fdf1e90df35881aea 0.00000000        1225.00000000       000005b4a68cb0d18fb1202b4a52b10ffd48d29252ed570bef667693b25b9f95 6
 $IRON      d7c86706f5817aa718cd1cfad03233bcd64a7789fd9422d3b17af6823a7e6ac6 60.00000000       120.00000000        000004fc310941cba77f4e8ba99b18cb22d2aed20f362698159afe2c2f411393 7
✨  Done in 2.33s.
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/balances.ts
+++ b/ironfish-cli/src/commands/wallet/balances.ts
@@ -46,49 +46,47 @@ export class BalancesCommand extends IronfishCommand {
     })
     this.log(`Account: ${response.content.account}`)
 
-    let showHeader = true
-
-    for (const balance of response.content.balances) {
-      // TODO(rohanjadvani) We currently fetch the asset from the blockchain to
-      // populate the name when rendering balance. This can be refactored once
-      // the wallet persists assets.
-      const assetResponse = await client.getAsset({ id: balance.assetId })
-
-      let columns: CliUx.Table.table.Columns<Balance> = {
-        name: {
-          header: 'Asset Name',
-        },
-        assetId: {
-          header: 'Asset Id',
-        },
-        confirmed: {
-          header: 'Confirmed Balance',
-          get: (row) => CurrencyUtils.renderIron(row.confirmed),
-        },
-      }
-
-      if (flags.all) {
-        columns = {
-          ...columns,
-          unconfirmed: {
-            header: 'Unconfirmed Balance',
-            get: (row) => CurrencyUtils.renderIron(row.unconfirmed),
-          },
-          blockHash: {
-            header: 'Head Hash',
-            get: (row) => row.blockHash || 'NULL',
-          },
-          sequence: {
-            header: 'Head Sequence',
-            get: (row) => row.sequence || 'NULL',
-          },
-        }
-      }
-
-      const name = BufferUtils.toHuman(Buffer.from(assetResponse.content.name, 'hex'))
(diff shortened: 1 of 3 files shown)


ironfish-cli/src/commands/wallet/balances.ts
```diff
@@ -0,0 +1,94 @@
...
+    all: Flags.boolean({
+      default: false,
+      description: `Also show unconfirmed balance, head hash, and head sequence`,
+    }),
+    confirmations: Flags.integer({
+      required: false,
+      description: 'Minimum number of blocks confirmations for a note',
+    }),
+  }
+
```
jowparks, line 28: Do you think this should live in a separate command? Seems like this could also be a flag on the `wallet:balance` command

NullSoldier: I would say, consider the implication of that suggestion outside of this command and how to be consistent. Are you implying no pluralized command should exist? Should we merge `wallet:transactions` into `wallet:transaction`? `wallet:transactions` shows you a summary of each TX, `wallet:transaction` will dump the raw full data TX

I think it's good to have smaller very clearly purposed commands, which is the unix mantra.

rohanjadvani: Yeah. Balance for other blockchains (like Ethereum) generally includes a specific asset, so I chose to use the plural to indicate the command is across all assets. Open to changing it, though.

ironfish-cli/src/commands/wallet/balances.ts
```diff
@@ -0,0 +1,94 @@
...
+    const client = await this.sdk.connectRpc()
+
+    const account = args.account as string | undefined
+    const response = await client.getAccountBalances({
+      account,
+      minimumBlockConfirmations: flags.confirmations,
+    })
+    this.log(`Account: ${response.content.account}`)
+
+    let showHeader = true
```
NullSoldier, line 49: You don't need this because you're not streaming the responses back. You can just render the whole table.

rohanjadvani: Fixed

NullSoldier (approved):

PR #2888, opened 2023-01-11
feat(ironfish,cli): Change account import to only use spending key

mat-if:
## Summary

Since the public address size changed and we changed some of our key personalization strings, it was possibly to import an account that was in an invalid state by importing an incorrect (but technically valid) incoming and outgoing view key

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/account.ts
+++ b/ironfish/src/wallet/account.ts
@@ -19,6 +19,8 @@ import { WalletDB } from './walletdb/walletdb'
 
 export const ACCOUNT_KEY_LENGTH = 32
 
+export type AccountImport = { name: string; spendingKey: string }
+
 export class Account {
   private readonly walletDb: WalletDB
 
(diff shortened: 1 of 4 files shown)


ironfish/src/wallet/account.ts
```diff
@@ -19,6 +19,8 @@ import { WalletDB } from './walletdb/walletdb'
 
 export const ACCOUNT_KEY_LENGTH = 32
 
+export type AccountImport = { name: string; spendingKey: string }
```
mat-if, line 22: I'm not sure if we have a standardized pattern for re-usable types, so I wanted to specifically call this out - it feels kind of strange here but not sure if there's a better place for it.

jowparks (approved):
Tested in huddle, things are working as expected

ironfish/src/wallet/account.ts
NullSoldier: well, I don't really think we need this type exactly but it's fine either way. When you use it in importFile we just cast the result if JSON parsing to this type, but in reality it should be unknown and fed through a YUP schema.

NullSoldier (approved):

NullSoldier:
It works, I tested it by importing a file from phase 2

PR #2890, opened 2023-01-11
Replace hardcode values

ygao76:
## Summary
Use constant instead of hardcoded value

## Testing Plan
existing unit tests
## Breaking Change

```
[ ] Yes
```
@rohanjadvani @mat-if


ironfish-rust-nodejs/index.js
```diff
@@ -275,6 +275,10 @@ module.exports.DECRYPTED_NOTE_LENGTH = DECRYPTED_NOTE_LENGTH
 module.exports.Note = Note
 module.exports.TransactionPosted = TransactionPosted
 module.exports.PROOF_LENGTH = PROOF_LENGTH
+module.exports.TRANSACTION_SIGNATURE_LENGTH = TRANSACTION_SIGNATURE_LENGTH
+module.exports.TRANSACTION_PUBLIC_KEY_LENGTH = TRANSACTION_PUBLIC_KEY_LENGTH
+module.exports.TRANSACTION_EXPIRATION_LENGTH = TRANSACTION_EXPIRATION_LENGTH
+module.exports.TRANSACTION_FEE_LENGTH = TRANSACTION_FEE_LENGTH
```
NullSoldier, line 281: This isn't a transaction specific fee length, its a currency encoded size which we always treat as a U64, so I think we can just make 1 constant if you really wanted one to encode the length of a currency amount that is applicable to all currency amounts.

ironfish-rust-nodejs/index.js
```diff
@@ -275,6 +275,10 @@ module.exports.DECRYPTED_NOTE_LENGTH = DECRYPTED_NOTE_LENGTH
 module.exports.Note = Note
 module.exports.TransactionPosted = TransactionPosted
 module.exports.PROOF_LENGTH = PROOF_LENGTH
+module.exports.TRANSACTION_SIGNATURE_LENGTH = TRANSACTION_SIGNATURE_LENGTH
+module.exports.TRANSACTION_PUBLIC_KEY_LENGTH = TRANSACTION_PUBLIC_KEY_LENGTH
```
NullSoldier, line 279: Why is the transaction public key different than the account public key length?

ironfish-rust-nodejs/index.js
rohanjadvani: Agreed with Jason, this name can be more generic for any sort of amount.

ironfish-rust-nodejs/index.js
rohanjadvani: Agreed. The [comment](https://github.com/iron-fish/ironfish/pull/2890/files#diff-347e78ce8439f128b344eb339befd4f4899d395e0cc65c953d0fb1426aa237f0R57) also indicates there's an existing constant that can be used.

ironfish-rust-nodejs/src/structs/transaction.rs
```diff
@@ -33,6 +36,18 @@ pub struct NativeTransactionPosted {
 #[napi]
 pub const PROOF_LENGTH: u32 = PROOF_SIZE;
 
+#[napi]
+pub const TRANSACTION_SIGNATURE_LENGTH: u32 = TRANSACTION_SIGNATURE_SIZE;
```
rohanjadvani, line 40: Is this being used? I see `TRANSACTION_SIGNATURE_SIZE` but I don't think it's being referenced outside of exports

ironfish-rust-nodejs/index.js
```diff
@@ -275,6 +275,10 @@ module.exports.DECRYPTED_NOTE_LENGTH = DECRYPTED_NOTE_LENGTH
...
+module.exports.TRANSACTION_EXPIRATION_LENGTH = TRANSACTION_EXPIRATION_LENGTH
```
rohanjadvani, line 280: Should this just be something `SEQUENCE_LENGTH`?
```suggestion
module.exports.SEQUENCE_LENGTH = SEQUENCE_LENGTH
```

rohanjadvani (changes requested):

ironfish-rust-nodejs/index.js
ygao76: This is randomized public key of the sender for a transaction. For the account public key, are you referring to public address? if so that is different thing.

ironfish-rust-nodejs/index.js
ygao76: I want to point out Fee has type of I64,  but its the same length as U64. 
I agree with one constant of number length, it will be used for all the number types, eg notes size, spends size, tree size, authPath length. Since these x_sizes are fixed type.

But I don't want to mix transaction fee and transaction expiration with the above, due to
- have the flexibility of changing fee and expiration type in the future without touching other number type 
- improve code readability. If we need to add a new field in transaction, we can easily identify all the places to add.

ironfish-rust-nodejs/src/structs/transaction.rs
ygao76: I want to define in the transaction mod.rs so the constant is in the same file as the original definition. If I define in ts code, there is a gap. When people change the transaction class, it might not be aware of this constant in ts.

ironfish-rust-nodejs/index.js
rohanjadvani: Ah yeah, my bad forgot about that. What about using `RANDOMIZED_PUBLIC_KEY_LENGTH` then?

ironfish-rust-nodejs/index.js
rohanjadvani: I'm not sure if @NullSoldier was referring to conflating expiration length. I think we can just have a currency amount length (or some equivalent)

ironfish-rust-nodejs/index.js
ygao76: I want to keep the original name since its referring to the transaction expiration length. I know the type of expiration is a sequence.  If we want to change the type of expiration to other class X, then we can not just update this constant value, instead we need to change all the `SEQUENCE_LENGTH` constants to a new X_LENGTH.

ironfish-rust-nodejs/index.js
ygao76: I will rename to `TRANSACTION_PUBLIC_KEY__RANDOMNESS_LENGTH`

ironfish-rust-nodejs/index.js
ygao76: Added TYPE_NUMBER_SIZE constants to represent the size of type `number` in typescript. This will be the generic one for all the number type size. I keep all the constants for transaction attribute since transaction attribute has its own types defined in rs package.

ironfish/src/primitives/rawTransaction.ts
```diff
@@ -17,6 +22,8 @@ import { Transaction } from './transaction'
 // Needed for constructing a witness when creating transactions
 const noteHasher = new NoteHasher()
 
+export const TYPE_NUMBER_SIZE = 8
```
rohanjadvani, line 25: I think we should omit this from this PR since it focuses on hard-coded values for the transaction

rohanjadvani:
@ygao76 Are there any other updates to make to this PR? I'll close it soon otherwise; feel free to re-open after I do.

ygao76:
I am not clear what are you suggesting to do in your comment, lets have a huddle on this

ironfish-cli/src/commands/wallet/export.ts
```diff
@@ -1,7 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { wordsSpendingKey } from '@ironfish/rust-nodejs'
+import { spendingKeyToWords } from '@ironfish/rust-nodejs'
```
rohanjadvani, line 4: Does this need to be rebased with `staging`?

ironfish/src/primitives/rawTransaction.ts
rohanjadvani: Bump on this, I don't think we need this and I don't think it should be defined in the raw transaction file

ironfish-rust/src/transaction/tests.rs
```diff
@@ -103,6 +105,19 @@ fn test_transaction() {
         Transaction::read(&mut serialized_transaction[..].as_ref())
             .expect("should be able to deserialize valid transaction");
 
+    assert_eq!(read_back_transaction.version, TRANSACTION_VERSION);
+    assert_eq!(
+        read_back_transaction.expiration.to_le_bytes().len(),
+        TRANSACTION_EXPIRATION_SIZE
+    );
+    assert_eq!(
```
rohanjadvani, line 113: This is already tested below

ironfish-rust/src/transaction/tests.rs
```diff
@@ -103,6 +105,19 @@ fn test_transaction() {
         Transaction::read(&mut serialized_transaction[..].as_ref())
             .expect("should be able to deserialize valid transaction");
 
+    assert_eq!(read_back_transaction.version, TRANSACTION_VERSION);
```
rohanjadvani, line 108: This isn't related to the PR change, I don't know if we need this added

ygao76: I can remove this, but I noticed we miss this assertion in the test.

rohanjadvani (commented):
This PR is focused on changing hardcoded values but is separately also changing serialization behavior. We shouldn't combine these changes in one merge. Can we separate out that change for review in a different PR?

ygao76:
> This PR is focused on changing hardcoded values but is separately also changing serialization behavior. We shouldn't combine these changes in one merge. Can we separate out that change for review in a different PR?

the serialization is a bug that exposed during I replace hardcode constants. They can not commit separately.

When I use the constant `TRANSACTION_EXPIRATION_LENGTH` for raw transaction, it will break the serialization since the expiration size is 4 instead of 8. the reader and writer need to be updated with U32

rohanjadvani (approved):
Thanks for fixing this up. I'm going to approve it, but we'll need to update expiration serialization JS side across the board in a subsequent change.

ygao76:
The expiration serialization JS is correct for transaction, the bug is only introduced for ran transaction.

PR #2891, opened 2023-01-11
Fix missing schema minimumBlockConfirmations

NullSoldier:
## Summary

This was missing from the schema which means it would always be stripped out

## Breaking Change

```
[ ] Yes
```


rohanjadvani (approved):

hughy (approved):

NullSoldier:
This isn't needed anymore because of https://github.com/iron-fish/ironfish/pull/2892

PR #2894, opened 2023-01-11
feat(ironfish): Duplicate assets store into the walletdb

rohanjadvani:
## Summary

We want to eventually remove coupling between the node and the wallet, so the wallet should be aware of asset data. This code change duplicates the asset store available in the blockchain. The underlying schema is exactly the same, but the files have been copied in case we want to change what data is stored in each module and if we decide to further separate the wallet out in the future.

## Testing Plan

Duplicated unit tests

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/walletdb/assetsValue.ts
```diff
@@ -0,0 +1,57 @@
...
+  ASSET_ID_LENGTH,
+  ASSET_METADATA_LENGTH,
+  ASSET_NAME_LENGTH,
+  ASSET_OWNER_LENGTH,
+  PUBLIC_ADDRESS_LENGTH,
+} from '@ironfish/rust-nodejs'
+import bufio from 'bufio'
+import { BigIntUtils } from '../../utils'
+
+export interface AssetsValue {
```
NullSoldier, line 15: Singular since it represents 1 asset, like the other values.

rohanjadvani: Alright, there's some inconsistency in the blockchain as well. I'll fix that in a separate PR.

NullSoldier (approved):

PR #2899, opened 2023-01-12
chore: update Discord invite link

mat-if:
## Summary

Changing to a link we control so we don't have to worry about it breaking

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):
Great idea!

NullSoldier:
We should change the website too

PR #2903, opened 2023-01-12
wallet:send with fee rate

ygao76:
## Summary
- Create transaction API supporting fee and fee priority level
- Refactor wallet.createTransaction and postTransaction
- Update wallet:send with create transaction api and post transaction api so its a two step flow.

Create this draft to demo my solution and to get early feedback since there is refactoring of wallet code and send cmd.

## Testing Plan
WIP
## Breaking Change

```
[ ] Yes
```
@NullSoldier @hughy


ironfish/src/rpc/routes/wallet/createTransaction.ts
```diff
@@ -0,0 +1,187 @@
...
+        let priprityLevel: PriorityLevel
+        switch (transaction.feePriorityLevel) {
+            case 'low':
+              priprityLevel = 'low'
+              break
+            case 'high':
+              priprityLevel = 'high'
+              break
+            default:
+              priprityLevel = 'medium'
```
hughy, line 156: Is the `switch` necessary? It looks like its only effect is to default to `'medium'`

ironfish/src/wallet/wallet.ts
```diff
@@ -784,8 +766,19 @@ export class Wallet {
...
+    const transaction = await this.workerPool.postTransaction(raw)
+    const verify = this.chain.verifier.verifyCreatedTransaction(transaction)
+    if (!verify.valid) {
+      throw new Error(`Invalid transaction, reason: ${String(verify.reason)}`)
+    }
+
+    await this.addPendingTransaction(transaction)
+    memPool.acceptTransaction(transaction)
+    this.broadcastTransaction(transaction)
+    this.onTransactionCreated.emit(transaction)
```
hughy, line 779: Adding these steps here creates side effects in `postTransaction` that might break tests that use `useTxFixture`

It might be easier to split this change into a separate PR because a lot of test suites use `useTxFixture` and they'll all need to be updated for the chain to that fixture

ironfish/src/wallet/wallet.ts
```diff
@@ -698,38 +659,47 @@ export class Wallet {
...
+      assetId: Buffer
+    }[],
+    mints: MintDescription[],
+    burns: BurnDescription[],
+    priorityLevel: PriorityLevel,
+    memPool: MemPool,
+    transactionExpirationDelta: number,
+    expiration?: number | null,
+  ): Promise<RawTransaction>{
+    const estimatedFee = await memPool.feeEstimator.estimateFee(priorityLevel, sender, receives)
```
hughy, line 692: `estimateFee` will do most of the same work that `createTransaction` is doing (iterating over notes, creating witnesses).

Will we move the logic for getting a fee based on a fee rate and a raw transaction in a separate PR?

ironfish/src/wallet/wallet.ts
ygao76: that is the change I am holding off to do before get agreement on the approach.

ironfish-cli/src/commands/wallet/send.ts
```diff
@@ -130,65 +135,38 @@ export class Send extends IronfishCommand {
...
+
+    const rawTransactionBytes = Buffer.from(createResponse.content.transaction, 'hex')
+    const rawTransaction = RawTransactionSerde.deserialize(rawTransactionBytes)
+
     if (!flags.confirm) {
       this.log(`
 You are about to send:
 ${CurrencyUtils.renderIron(
-  amount,
+  rawTransaction.spends[0].note.value(),
```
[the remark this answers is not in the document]
ygao76: How should I get amount from Raw Transaction? sum up all the notes values?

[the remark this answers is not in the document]
hughy: We can use the amount that the user inputs

To get the amount from the raw transaction I think we'd need to iterate over the notes in `rawTransaction.receives` and exclude any that the sender is receiving as change. It might not be possible to distinguish those notes though: if an account sends a transaction to itself then all of the notes will belong to that account.

ironfish/src/rpc/routes/wallet/createTransaction.ts
ygao76: so `transaction.feePriorityLevel` is type of string. And priprityLevel has enum type of "low" | "medium" | "high". I can not directly convert because Type 'string' is not assignable to type '"low" | "medium" | "high"'

ironfish-cli/src/commands/wallet/send.ts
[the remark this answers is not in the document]
ygao76: I see, okay!

ironfish/src/testUtilities/fixtures/blocks.ts
```diff
@@ -251,7 +251,7 @@ export async function useBlockWithTx(
       options.expiration ?? 0,
     )
 
-    const transaction = await node.wallet.postTransaction(raw)
+    const transaction = await node.wallet.postTransaction(raw, node.memPool)
```
NullSoldier, line 254: The wallet will have no access to the mem pool so we can't do this.

ironfish/src/rpc/routes/wallet/createTransaction.ts
```diff
@@ -0,0 +1,181 @@
...
+     publicAddress: string
+     amount: string
+     memo: string
+     assetId?: string
+   }[]
+   fee?: string
+   feePriorityLevel?: string
+   expiration?: number | null
+   expirationDelta?: number | null
+ }
```
NullSoldier, line 26: This is missing mints and burns

ironfish/src/rpc/routes/wallet/createTransaction.ts
```diff
@@ -0,0 +1,181 @@
...
+ export type CreateTransactionRequest = {
+   fromAccountName: string
+   receives: {
+     publicAddress: string
+     amount: string
+     memo: string
+     assetId?: string
+   }[]
+   fee?: string
+   feePriorityLevel?: string
```
NullSoldier, line 23: this should take a fee rate not a fee priority level

ironfish/src/wallet/wallet.ts
hughy: Another way we could do this would be to modify the `fund` method to optionally take a fee rate.

`fund` takes a raw transaction and adds the spends needed to satisfy the transaction receives and fee. If it takes a fee rate it would need to do this in steps:
1. fund the receives
2. calculate the fee based on the size of the raw transaction
3. fund the fee

ironfish/src/rpc/routes/wallet/createTransaction.ts
ygao76: why you think it should be a fee rate? Other wallet like MetaMask, you specify the level. I can add fee rate as one of the request params but I dont want to limit the interface that user can not create transaction with fee level.

ironfish/src/rpc/routes/wallet/createTransaction.ts
ygao76: I will add

ironfish/src/rpc/routes/wallet/createTransaction.ts
```diff
@@ -0,0 +1,181 @@
...
+             publicAddress: yup.string().defined(),
+             amount: yup.string().defined(),
+             memo: yup.string().defined(),
+             assetId: yup.string().optional(),
+           })
+           .defined(),
+       )
+       .defined(),
+     fee: yup.string().optional(),
+     feePriorityLevel: yup.string().optional(),
```
NullSoldier, line 48: This will be fee rate now right?

PR #2907, opened 2023-01-13
adds IRON amount to wallet:transactions output

hughy:
## Summary

the transactionValue stored in the walletDb now includes a mapping from assetId to the balance delta for that account and asset in the transaction. this allows us to return transaction amounts in our RPC responses and display them in the CLI.

serializes assetBalanceDeltas as a JSON string because yup does not support map types.

only includes the balance delta for IRON in the CLI output table. we can show balance deltas for all assets in a transaction more easily in the output for a single transaction in wallet:transaction

## Testing Plan

image

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/types.ts
+++ b/ironfish/src/rpc/routes/wallet/types.ts
@@ -15,7 +15,7 @@ export type RpcAccountTransaction = {
   burnsCount: number
   expiration: number
   timestamp: number
-  assetBalanceDeltas: string
+  assetBalanceDeltas: Array<{ assetId: string; delta: string }>
 }
 
 export type RpcAccountDecryptedNote = {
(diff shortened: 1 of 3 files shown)


ironfish/src/rpc/routes/wallet/types.ts
```diff
@@ -15,6 +15,7 @@ export type RpcAccountTransaction = {
   burnsCount: number
   expiration: number
   timestamp: number
+  assetBalanceDeltas: string
```
NullSoldier, line 18: I think instead of sending this back as JSON we should do `Array<{ assetId: string, delta: number }>`

hughy: updated to `Array<{ assetId: string, delta: string }>`!

NullSoldier (approved):

PR #2912, opened 2023-01-13
feat(cli): Add asset selector for `wallet:send`

mat-if:
## Summary

Per the [oclif docs](https://oclif.io/docs/prompting#inquirer), they recommend the `inquirer` lib for this. We also had it installed already as a transitive dependency.

Since we're adding this new functionality: we should use this `inquirer` selector any time we have a known, finite list of things to select from. It's a huge improvement over making the user go look up the information out-of-band and then look up the flag to pass into the CLI.

Screen Recording 2023-01-13 at 10 30 31 AM

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/send.ts
+++ b/ironfish-cli/src/commands/wallet/send.ts
@@ -1,8 +1,10 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { CurrencyUtils, isValidPublicAddress } from '@ironfish/sdk'
+import { Asset } from '@ironfish/rust-nodejs'
+import { BufferUtils, CurrencyUtils, isValidPublicAddress, RpcClient } from '@ironfish/sdk'
 import { CliUx, Flags } from '@oclif/core'
+import inquirer from 'inquirer'
 import { IronfishCommand } from '../../command'
 import { RemoteFlags } from '../../flags'
 import { ProgressBar } from '../../types'
@@ -63,6 +65,7 @@ export class Send extends IronfishCommand {
     const { flags } = await this.parse(Send)
     let amount = null
     let fee = null
+    let assetId = null
     let to = flags.to?.trim()
     let from = flags.account?.trim()
     const expiration = flags.expiration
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -7040,30 +7054,10 @@ init-package-json@^3.0.2:
     validate-npm-package-license "^3.0.4"
     validate-npm-package-name "^4.0.0"
 
-inquirer@^8.0.0:
```
mat-if, line 7043: I got a little fiddly here to de-dupe these transitive dependencies, so we have 1 less thing downloaded/installed.

jowparks (approved):

ironfish-cli/src/commands/wallet/send.ts
```diff
@@ -265,4 +278,44 @@ Find the transaction on https://explorer.ironfish.network/transaction/${
...
+          name: `${assetId} (${displayName})`,
+        })
+      }
+    }
+
+    let assetId
+
+    if (balancesResponse.content.balances.length === 1) {
+      // If there's only one available asset, showing the choices is unnecessary
+      assetId = balancesResponse.content.balances[0].assetId
```
hughy, line 306: We could check this condition before the loop above to avoid an unnecessary `getAsset` call

hughy: I'm also thinking that in practice, if there's only one asset, doesn't it have to be $IRON in order to send a transaction? I don't think that we should hardcode that, but `send` might always fail if there's no iron for a tx fee

hughy (approved):
The selector looks great!

ironfish-cli/src/commands/wallet/send.ts
mat-if: That's a good point. I think we could improve the command's usability further with more guardrails. For example, the send command will happily let you start the command even if you have 0 balance, which can be confusing and strange. 

Also per the first comment, moved this up.

NullSoldier (approved):

ironfish-cli/src/commands/wallet/send.ts
```diff
@@ -265,4 +278,43 @@ Find the transaction on https://explorer.ironfish.network/transaction/${
...
+        })
+      }
+    }
+
+    const response: { assetId: string } = await inquirer.prompt<{ assetId: string }>([
+      {
+        name: 'assetId',
+        message: 'Select the asset you wish to send',
+        type: 'list',
+        choices: assetOptions,
```
NullSoldier, line 315: It would be nice to sort the native asset to the first option

PR #2916, opened 2023-01-13
fix(cli): Make `fee` denomination in mint/burn consistent with send

rohanjadvani:
## Summary

Use fee in $IRON for mint and burn

## Testing Plan

Manually tested

## Breaking Change

```
[ ] Yes
```


hughy (approved):
We may want to come back to add fee rate estimation to mint/burn once we've sorted out that flow for send

rohanjadvani:
Yeah, @NullSoldier mentioned that as well and to omit it for now. We can revisit.

PR #2917, opened 2023-01-13
fix(cli): Update amount denominations in mint/burn to be IRON

rohanjadvani:
## Summary

Make amounts consistent with send

## Testing Plan

Manual test

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/burnAsset.ts
+++ b/ironfish/src/rpc/routes/wallet/burnAsset.ts
@@ -75,7 +75,7 @@ router.register<typeof BurnAssetRequestSchema, BurnAssetResponse>(
     request.end({
       assetId: burn.assetId.toString('hex'),
       hash: transaction.hash().toString('hex'),
-      value: burn.value.toString(),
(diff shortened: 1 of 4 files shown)


mat-if:
I think we're double encoding a value or something on this display:

image

rohanjadvani:
Should be fixed now.

image

ironfish/src/rpc/routes/wallet/burnAsset.ts
```diff
@@ -73,6 +75,7 @@ router.register<typeof BurnAssetRequestSchema, BurnAssetResponse>(
     request.end({
       assetId: burn.assetId.toString('hex'),
       hash: transaction.hash().toString('hex'),
+      value: burn.value.toString(),
```
NullSoldier, line 78: CurrencyUtils.encode(burn.value)

NullSoldier, line 78:
```suggestion
      value: CurrencyUtils.encode(burn.value),
```

NullSoldier (approved):

PR #2923, opened 2023-01-13
adds asset name, id to notes in wallet:transaction output

hughy:
## Summary

updates the table of output notes to include the asset name and id for each note.

## Testing Plan

image

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/getTransaction.ts
+++ b/ironfish/src/rpc/routes/wallet/getTransaction.ts
@@ -53,6 +53,8 @@ export const GetAccountTransactionResponseSchema: yup.ObjectSchema<GetAccountTra
                 .object({
                   owner: yup.boolean().defined(),
                   value: yup.string().defined(),
+                  assetId: yup.string().defined(),
+                  assetName: yup.string().defined(),
                   sender: yup.string().defined(),
                   memo: yup.string().trim().defined(),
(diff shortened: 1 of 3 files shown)


ironfish/src/rpc/routes/wallet/getTransaction.ts
```diff
@@ -98,10 +100,14 @@ router.register<typeof GetAccountTransactionRequestSchema, GetAccountTransaction
...
         : new Note(decryptedNote.serializedNote)
 
+      const asset = await node.chain.getAssetById(note.assetId())
+
       serializedNotes.push({
         owner,
         memo: note.memo(),
         value: CurrencyUtils.encode(note.value()),
+        assetId: note.assetId().toString('hex'),
+        assetName: asset?.name.toString('utf8') || '',
```
NullSoldier, line 110: Buffers should remain buffers through our RPC layer, and rendered on the CLI side.

NullSoldier (approved):

ironfish/src/rpc/routes/wallet/getTransaction.ts
hughy: Discussed elsewhere; we'll merge these changes as they are for now and change strings to buffers in RPC in a separate PR.

PR #2924, opened 2023-01-13
chore: update genesis phase3

jowparks:
## Summary
New genesis block for phase 3 of testnet

## Breaking Change

```
[ ] Yes
```


ironfish/src/defaultNetworkDefinitions.ts
```diff
@@ -39,11 +39,45 @@ const DEV_GENESIS = `{
...
+    }
+  ]
+}`
 
 export const TESTNET = `{
   "id": 0,
   "bootstrapNodes": ["test.bn1.ironfish.network"],
-  "genesis": ${DEV_GENESIS},
+  "genesis": ${PHASE3_GENESIS},
```
NullSoldier, line 80: What difficulty did you use?

PR #2933, opened 2023-01-17
Display asset deltas when calling wallet:transaction

EvanJRichard:
## Summary
Builds off of #2907 to dump all asset deltas when inspecting a single transaction.
Looking for input as to whether this is correct & whether this is the desired UX. 

## Testing
Screen Shot 2023-01-17 at 4 04 16 PM
 Plan

## Breaking Change

```
[ ] Yes
[X] No
```


hughy (commented):
The changes look good!

I think rebasing onto staging and removing the `yarn.lock` changes will fix the build issues

hughy (approved):

PR #2934, opened 2023-01-17
adds timestamp to wallet:transaction

hughy:
## Summary

adds the stored transaction timestamp to the RPC response and CLI output of the wallet:transaction command

timestamp is the time that the transaction was first seen by the wallet

## Testing Plan

image

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/transaction.ts
+++ b/ironfish-cli/src/commands/wallet/transaction.ts
@@ -45,6 +45,7 @@ export class TransactionCommand extends IronfishCommand {
     this.log(`Transaction: ${hash}`)
     this.log(`Account: ${response.content.account}`)
     this.log(`Status: ${response.content.transaction.status}`)
+    this.log(`Timestamp: ${new Date(response.content.transaction.timestamp).toLocaleString()}`)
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/commands/wallet/transaction.ts
```diff
@@ -45,6 +45,7 @@ export class TransactionCommand extends IronfishCommand {
     this.log(`Transaction: ${hash}`)
     this.log(`Account: ${response.content.account}`)
     this.log(`Status: ${response.content.transaction.status}`)
+    this.log(`Timestamp: ${new Date(response.content.transaction.timestamp).toLocaleString()}`)
```
mat-if, line 48: How hard would it be to add a timezone? It's been a minute since I played with JS dates, but that could be nice to clarify "is this UTC or my timezone?"

hughy: Good point! I'll see what I can do about adding the timezone

NullSoldier: I like mat's suggestion, but let's do a sweep in another PR for this.

```
$ new Date().toLocaleString('en-us',{timeZoneName:'short'})
'1/17/2023, 1:12:47 PM EST'
```

PR #2945, opened 2023-01-17
Change NotEnoughFundsError to use CurrencyUtils.renderIron

EvanJRichard:
## Summary
`NotEnoughFundsError` just dumps the raw ore amounts and assetID. Use `CurrencyUtils.renderIron` instead.

## Testing Plan
`yarn test` locally detected no errors.

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/wallet/errors.ts
+++ b/ironfish/src/wallet/errors.ts
@@ -2,13 +2,13 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+import { CurrencyUtils } from "../utils"
+
 export class NotEnoughFundsError extends Error {
   name = this.constructor.name
 
   constructor(assetId: Buffer, amount: bigint, amountNeeded: bigint) {
     super()
-    this.message = `Insufficient funds: Needed ${amountNeeded.toString()} but have ${amount.toString()} for asset '${assetId.toString(
-      'hex',
-    )}'`
+    this.message = `Insufficient funds: Needed ${CurrencyUtils.renderIron(amountNeeded.toString())} but have '${CurrencyUtils.renderIron(amount, true, assetId.toString('hex',))}'`
   }
 }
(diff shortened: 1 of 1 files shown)


ironfish/src/wallet/errors.ts
```diff
@@ -2,13 +2,13 @@
...
+
 export class NotEnoughFundsError extends Error {
   name = this.constructor.name
 
   constructor(assetId: Buffer, amount: bigint, amountNeeded: bigint) {
     super()
-    this.message = `Insufficient funds: Needed ${amountNeeded.toString()} but have ${amount.toString()} for asset '${assetId.toString(
-      'hex',
-    )}'`
+    this.message = `Insufficient funds: Needed ${CurrencyUtils.renderIron(amountNeeded.toString())} but have '${CurrencyUtils.renderIron(amount, true, assetId.toString('hex',))}'`
```
NullSoldier, line 12: There is a sneaky extra comma in this toString()

NullSoldier: Also we don't need to do amountNeeded.toString() here since `renderIron` takes either ore as bigint or encoded ore as a string.

EvanJRichard: `assetId.toString('hex',)` becomes `assetId.toString('hex')

ironfish/src/wallet/errors.ts
```diff
@@ -2,13 +2,15 @@
...
 export class NotEnoughFundsError extends Error {
   name = this.constructor.name
 
   constructor(assetId: Buffer, amount: bigint, amountNeeded: bigint) {
     super()
-    this.message = `Insufficient funds: Needed ${amountNeeded.toString()} but have ${amount.toString()} for asset '${assetId.toString(
-      'hex',
-    )}'`
+    this.message = `Insufficient funds: Needed ${CurrencyUtils.renderIron(
+      amountNeeded.toString(),
```
NullSoldier, line 13: `CurrencyUtils.renderIron()` takes either an encoded currency value, or just a bigint currency value so we don't need to call toString() on this.

Also, should we display the ticker here too, or no?

EvanJRichard: removed `toString()`

EvanJRichard: I'm torn, `needed 1 but had 0 IRON` (or whatever other ticker) seems odd but `needed 1 IRON but had 0 IRON` seems redundant

wait I'll change it to `needed 1 IRON but had 0`

NullSoldier (approved):

PR #2950, opened 2023-01-18
Mint burn

lwisne:
## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/wallet/burn.ts
```diff
@@ -74,7 +74,7 @@ export class Burn extends IronfishCommand {
     if (flags.amount) {
       amount = CurrencyUtils.decodeIron(flags.amount)
     } else {
-      const input = await CliUx.ux.prompt('Enter the amount to burn in IRON', {
+      const input = await CliUx.ux.prompt('Enter the amount to burn in the custom asset', {
```
NullSoldier, line 77:
```suggestion
      const input = await CliUx.ux.prompt('Enter the amount to burn in the asset', {
```

NullSoldier (approved):

PR #2952, opened 2023-01-18
Create Transaction API

ygao76:
## Summary
create transaction API
support create raw transaction with fee or fee rate

## Testing Plan
unit test
## Breaking Change

```
[ ] Yes
```
@NullSoldier @hughy @danield9tqh


ironfish/src/wallet/wallet.ts
```diff
@@ -711,7 +719,25 @@ export class Wallet {
...
       raw.burns = burns
-      raw.fee = fee
+
+      if (fee) {
+        raw.fee = fee
+      }
+
+      if (feeRate) {
+        let size = 0
+        size += 8 // spends length
```
ygao76, line 729: this hardcode value will be updated to constant once I merge my other PR

ironfish/src/rpc/routes/wallet/createTransaction.ts
```diff
@@ -0,0 +1,233 @@
...
+    const burns: BurnDescription[] = []
+    if (transaction.burns) {
+      transaction.burns.map((burn) => {
+        let assetId = Asset.nativeId()
+        if (burn.assetId) {
+          assetId = Buffer.from(burn.assetId, 'hex')
+        }
+        burns.push({
+          assetId: assetId,
+          value: CurrencyUtils.decode(burn.value),
```
hughy, line 151: Do we need to account for the burn values in `totalByAssetIdentifier`?

ironfish/src/rpc/routes/wallet/createTransaction.ts
```diff
@@ -0,0 +1,233 @@
...
+
+      const sum = totalByAssetIdentifier.get(assetId) ?? BigInt(0)
+      totalByAssetIdentifier.set(assetId, sum + amount)
+    }
+
+    // Check that the node account is updated
+    for (const [assetId, sum] of totalByAssetIdentifier) {
+      const balance = await node.wallet.getBalance(account, assetId)
+
+      if (balance.confirmed < sum) {
```
hughy, line 179: I'm wondering if we should do this check in the RPC or in the wallet code: the wallet needs to redo the work of iterating over the receives and burns in order to determine the amount needed.

That change might be a task for a separate PR.

ironfish/src/rpc/routes/wallet/createTransaction.ts
```diff
@@ -0,0 +1,233 @@
...
+    if (transaction.fee) {
+      const fee = CurrencyUtils.decode(transaction.fee)
+      if (fee < 1n) {
+        throw new ValidationError(`Invalid transaction fee, ${transaction.fee}`)
+      }
+
+      totalByAssetIdentifier.set(Asset.nativeId(), fee)
+    }
+
+    for (const { assetId, amount } of receives) {
```
hughy, line 166: We iterate over the receives both here and on line 118. We could consolidate the two loops

ironfish/src/wallet/wallet.ts
```diff
@@ -682,15 +685,20 @@ export class Wallet {
...
     expirationDelta: number,
     expiration?: number | null,
+    feeRate?: bigint,
   ): Promise<RawTransaction> {
     const heaviestHead = this.chain.head
     if (heaviestHead === null) {
       throw new Error('You must have a genesis block to create a transaction')
     }
 
+    if (fee == null && feeRate == null) {
```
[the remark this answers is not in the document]
ygao76: ` == null` will check for both `undefined` or `null`

ironfish/src/rpc/routes/wallet/createTransaction.ts
ygao76: @rohanjadvani @mat-if Do you know the answer to this?

ironfish/src/wallet/wallet.ts
```diff
@@ -711,7 +719,25 @@ export class Wallet {
...
+        size += 8 // notes length
+        size += 8 // fee
+        size += 4 // expiration
+        size += 64 // signature
+
+        size += raw.spends.length * SPEND_SERIALIZED_SIZE_IN_BYTE
+
+        size += raw.receives.length * NOTE_ENCRYPTED_SERIALIZED_SIZE_IN_BYTE
+
+        raw.fee = getFee(feeRate, size)
```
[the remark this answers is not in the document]
ygao76: @rohanjadvani @mat-if what is the size in byte of each mint and burn here?

ironfish/src/rpc/routes/wallet/createTransaction.test.slow.ts
```diff
@@ -0,0 +1,227 @@
...
+import { useMinerBlockFixture } from '../../../testUtilities'
+import { createRouteTest } from '../../../testUtilities/routeTest'
+import { Account } from '../../../wallet'
+import { ERROR_CODES } from '../../adapters/errors'
+
+const REQUEST_PARAMS = {
+  fromAccountName: 'existingAccount',
+  receives: [
+    {
+      publicAddress: '0d804ea639b2547d1cd612682bf99f7cad7aad6d59fd5457f61272defcd4bf5b',
```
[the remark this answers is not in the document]
ygao76: I dont think so. This is the recipient address, its not in the fixtures. I tried to regenerate fixtures locally and it succeeded.

ironfish/src/wallet/wallet.ts
```diff
@@ -682,15 +685,20 @@ export class Wallet {
     }[],
     mints: MintDescription[],
     burns: BurnDescription[],
-    fee: bigint,
+    fee: bigint | null,
     expirationDelta: number,
     expiration?: number | null,
+    feeRate?: bigint,
```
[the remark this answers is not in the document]
ygao76: Use options in the function interface

ironfish/src/wallet/wallet.ts
```diff
@@ -711,7 +719,25 @@ export class Wallet {
...
+
+      if (feeRate) {
+        let size = 0
+        size += 8 // spends length
+        size += 8 // notes length
+        size += 8 // fee
+        size += 4 // expiration
+        size += 64 // signature
+
+        size += raw.spends.length * SPEND_SERIALIZED_SIZE_IN_BYTE
```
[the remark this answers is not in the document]
ygao76: `fund` takes fee and populate spends. In the case of fee rate,  I estimate fee without the size of spends and use it as input for the fund.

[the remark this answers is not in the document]
hughy: There is sort of a chicken/egg problem with fee and spends, but the spends will usually make up around half the size and possibly more than half the size of the transaction. If we don't include them at all then the size could be way off.

ironfish/src/rpc/routes/wallet/createTransaction.ts
ygao76: Currently we do the chekc in both PRC. and wallet code. It's redundant but I think we want to catch insufficient fund error early. If there are other people think we should remove the check I can make a separate PR.

hughy (approved):
RPC endpoint looks good!

We'll address including spends in the fee estimate and funding the fee in a separate PR

ironfish/src/rpc/routes/wallet/createTransaction.ts
```diff
@@ -0,0 +1,225 @@
...
+import { BurnDescription } from '../../../primitives/burnDescription'
+import { MintDescription } from '../../../primitives/mintDescription'
+import { RawTransactionSerde } from '../../../primitives/rawTransaction'
+import { CurrencyUtils } from '../../../utils'
+import { NotEnoughFundsError } from '../../../wallet/errors'
+import { ERROR_CODES, ValidationError } from '../../adapters/errors'
+import { ApiNamespace, router } from '../router'
+
+export type CreateTransactionRequest = {
+  fromAccountName: string
```
NullSoldier, line 16: This has 2 different names, `sender` in the `wallet.createTransaction()`, `fromAccountName` here, and `account` in other RPCs. Can we choose one?

ironfish/src/rpc/routes/wallet/createTransaction.ts
```diff
@@ -0,0 +1,225 @@
...
+  receives: {
+    publicAddress: string
+    amount: string
+    memo: string
+    assetId?: string
+  }[]
+  mints?: {
+    privateKey: string
```
NullSoldier, line 24: Let's not send private keys over RPC like this. Only use the keys already in the wallet.

ironfish/src/rpc/routes/wallet/createTransaction.ts
```diff
@@ -0,0 +1,225 @@
...
+    name: string
+    metadata: string
+    value: string
```
NullSoldier, line 27: How do you use this to mint supply for an existing asset? This has to take asset id too.

ironfish/src/rpc/routes/wallet/createTransaction.ts
```diff
@@ -0,0 +1,225 @@
...
+  .object({
+    transaction: yup.string().defined(),
+  })
+  .defined()
+
+router.register<typeof CreateTransactionRequestSchema, CreateTransactionResponse>(
+  `${ApiNamespace.wallet}/createTransaction`,
+  CreateTransactionRequestSchema,
+  async (request, node): Promise<void> => {
+    const transaction = request.data
```
NullSoldier, line 97: this isnt a transaction, lets delete this variable and just use request.data, or just call this data.

ironfish/src/rpc/routes/wallet/createTransaction.ts
```diff
@@ -0,0 +1,225 @@
...
+      if (receive.assetId) {
+        assetId = Buffer.from(receive.assetId, 'hex')
+      }
+
+      const amount = CurrencyUtils.decode(receive.amount)
+      if (amount < 0) {
+        throw new ValidationError(`Invalid transaction amount ${amount}.`)
+      }
+
+      const sum = totalByAssetIdentifier.get(assetId) ?? BigInt(0)
```
NullSoldier, line 139: nit:
```suggestion
      const sum = totalByAssetIdentifier.get(assetId) ?? 0n
```

ironfish/src/rpc/routes/wallet/createTransaction.ts
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    const receives = transaction.receives.map((receive) => {
+      let assetId = Asset.nativeId()
+      if (receive.assetId) {
+        assetId = Buffer.from(receive.assetId, 'hex')
+      }
+
+      const amount = CurrencyUtils.decode(receive.amount)
+      if (amount < 0) {
```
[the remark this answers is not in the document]
ygao76: I think we should expose this error to user since it might be a typo from user.

ironfish/src/rpc/routes/wallet/createTransaction.ts
ygao76: added asset id as input

ironfish/src/rpc/routes/wallet/createTransaction.ts
```diff
@@ -0,0 +1,250 @@
...
+
+          asset = new Asset(
+            account.spendingKey,
+            record.name.toString('utf8'),
+            record.metadata.toString('utf8'),
+          )
+          // Verify the stored asset produces the same identfier before building a transaction
+          if (!asset.id().equals(Buffer.from(mint.assetId, 'hex'))) {
+            throw new ValidationError(`Unauthorized to mint for asset '${mint.assetId}'`)
+          }
```
NullSoldier, line 170: We don't do this for mint, I don't think this is necessary if Rohan didn't do it. The TX won't validate.

https://github.com/iron-fish/ironfish/blob/master/ironfish/src/rpc/routes/wallet/mintAsset.ts#L74-L81

ironfish/src/rpc/routes/wallet/createTransaction.ts
```diff
@@ -0,0 +1,250 @@
...
+        )
+      }
+    }
+
+    let rawTransaction
+
+    try {
+      if (data.fee) {
+        rawTransaction = await node.wallet.createTransaction(account, receives, mints, burns, {
+          fee: BigInt(data.fee),
```
NullSoldier, line 217: CurrencyUtils.decode(data.fee)

ironfish/src/rpc/routes/wallet/createTransaction.ts
```diff
@@ -0,0 +1,250 @@
...
+          expirationDelta:
+            data.expirationDelta ?? node.config.get('transactionExpirationDelta'),
+          expiration: data.expiration,
+        })
+      } else {
+        const feeRate = BigInt(
```
NullSoldier, line 223: CurrencyUtils.decode(data.feeRate), fee rates are currency utils with a unit of `ore / 1 KB`

ironfish/src/rpc/routes/wallet/createTransaction.ts
ygao76: @rohanjadvani did this in wallet.mint(), I think we should either move that logic to wallet.creationTransaction() or I will have to keep it here.

rohanjadvani: Yeah, this existed in `wallet.mint`. It's a sanity check before it hits the circuits, so I think it's useful to have. I think we should include it in the wallet instead of RPC.

ygao76: I will move it to wallet.createTransaction from wallet.mint()

PR #2953, opened 2023-01-18
feat(ironfish): Save assets to wallet db

rohanjadvani:
## Summary

Update assets store in wallet db for incoming transactions if corresponding to stored accounts

## Testing Plan

Unit tests

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/account.ts
```diff
@@ -213,6 +216,156 @@ export class Account {
...
+      const supply = existingSupply - value
+      Assert.isTrue(supply >= BigInt(0))
+
+      // If we are reverting the transaction which matches the created at
+      // hash of the asset, delete the record from the store
+      if (
+        transaction.hash().equals(existingAsset.createdTransactionHash) &&
+        supply === BigInt(0)
+      ) {
+        await this.walletDb.deleteAsset(this, assetId, tx)
```
hughy, line 350: You might be planning to address this with assets in pending transactions -- the transaction may be pending after it's disconnected so we may not want to delete the asset.

ironfish/src/wallet/account.ts
```diff
@@ -213,6 +216,156 @@ export class Account {
...
+
+  private async saveConnectedBurnsToAssetsStore(
+    transaction: Transaction,
+    tx: IDatabaseTransaction,
+  ): Promise<void> {
+    for (const { assetId, value } of transaction.burns) {
+      const existingAsset = await this.walletDb.getAsset(this, assetId, tx)
+      if (!existingAsset) {
+        continue
+      }
```
hughy, line 263: This shouldn't happen, but what should happen if the wallet processes a burn transaction for an asset it owns, but the asset is missing from the db?

hughy (approved):
A couple of non-blocking comments.

LGTM! ✅

ironfish/src/wallet/account.ts
rohanjadvani: That seems like an invalid db state, so I'd expect an assertion to fail in that case. Similar to how we look up transactions via a note and throw assertion errors. We should do the same here. 

This block skips in a loop because this method will be called for all accounts in a wallet. I wouldn't want a crash if Account A is burning an asset it owns but Account B sees the burn

ironfish/src/wallet/account.ts
rohanjadvani: Yeah, that's a good point. I'll refactor this so the asset gets deleted when the associated created transaction is. Otherwise, we can just adjust the supply.

I'll also add a blockHash / sequence since that may be useful to infer pending status.

hughy (approved):

PR #2955, opened 2023-01-18
feat(cli): add asset selector to burn command

mat-if:
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/send.ts
+++ b/ironfish-cli/src/commands/wallet/send.ts
@@ -2,12 +2,12 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { Asset } from '@ironfish/rust-nodejs'
-import { BufferUtils, CurrencyUtils, isValidPublicAddress, RpcClient } from '@ironfish/sdk'
+import { CurrencyUtils, isValidPublicAddress } from '@ironfish/sdk'
 import { CliUx, Flags } from '@oclif/core'
-import inquirer from 'inquirer'
 import { IronfishCommand } from '../../command'
 import { RemoteFlags } from '../../flags'
 import { ProgressBar } from '../../types'
+import { selectAsset } from '../../utils/asset'
 
 export class Send extends IronfishCommand {
   static description = `Send coins to another account`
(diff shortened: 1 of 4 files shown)


ironfish-cli/src/commands/wallet/send.ts
```diff
@@ -278,43 +274,4 @@ Find the transaction on https://explorer.ironfish.network/transaction/${
       this.exit(2)
     }
   }
-
-  async selectAsset(
```
mat-if, line 282: Moved this into a util file

ironfish-cli/src/utils/asset.ts
```diff
@@ -0,0 +1,54 @@
...
+  const balancesResponse = await client.getAccountBalances({ account })
+  const assetOptions = []
+
+  let balances = balancesResponse.content.balances
+
+  if (!showNativeAsset) {
+    balances = balances.filter(
+      (balance) => balance.assetId !== Asset.nativeId().toString('hex'),
+    )
+  }
```
mat-if, line 23: This part is new, since we disallow burning the native asset, we need to be able to filter it

ironfish-cli/src/utils/asset.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  if (balances.length === 0) {
+    return undefined
+  } else if (balances.length === 1) {
+    // If there's only one available asset, showing the choices is unnecessary
+    return balancesResponse.content.balances[0].assetId
+  }
+
+  // Get the asset name from the chain DB to populate the display choices
+  for (const { assetId } of balancesResponse.content.balances) {
```
NullSoldier, line 33: loop over balances

ironfish-cli/src/utils/asset.ts
```diff
@@ -0,0 +1,54 @@
...
+    balances = balances.filter(
+      (balance) => balance.assetId !== Asset.nativeId().toString('hex'),
+    )
+  }
+
+  if (balances.length === 0) {
+    return undefined
+  } else if (balances.length === 1) {
+    // If there's only one available asset, showing the choices is unnecessary
+    return balancesResponse.content.balances[0].assetId
```
NullSoldier, line 29: and here

NullSoldier (approved):

PR #2957, opened 2023-01-18
feat(cli): Add asset selector to the mint command

mat-if:
## Summary

~~Builds on #2955~~

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/mint.ts
+++ b/ironfish-cli/src/commands/wallet/mint.ts
@@ -6,6 +6,7 @@ import { CliUx, Flags } from '@oclif/core'
 import { IronfishCommand } from '../../command'
 import { RemoteFlags } from '../../flags'
 import { ProgressBar } from '../../types'
+import { selectAsset } from '../../utils/asset'
 
 export class Mint extends IronfishCommand {
   static description = 'Mint tokens and increase supply for a given asset'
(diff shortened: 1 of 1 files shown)


NullSoldier (approved):

ygao76 (approved):

ironfish-cli/src/commands/wallet/mint.ts
```diff
@@ -96,9 +97,13 @@ export class Mint extends IronfishCommand {
...
       } else {
-        assetId = await CliUx.ux.prompt('Enter the Asset Identifier to mint supply for', {
-          required: true,
-        })
+        if (assetId == null) {
+          assetId = await selectAsset(client, account, false)
+        }
+
+        if (assetId == null) {
+          this.error(`You must have an existing asset, try creating a new one.`)
```
ygao76, line 105: nit: the error message is not very user friendly, it can be something like 'The given asset id is not found, please provide a valid asset id'

mat-if: It's not possible to get to this point if a user has a custom asset. If they have at least one, this error should never show. Telling them to provide an asset id at this point would be confusing, and the more that we can move away from explaining implementation details (asset id instead of asset), the more friendly the UX in my opinion

mat-if: I do want to do a pass on all these error messages tomorrow though. To make them friendlier, we should tell them run `wallet:mint`, for example.

PR #2960, opened 2023-01-18
Pass oclif table flags into tables for wallet:transaction

NullSoldier:
## Summary

When we use non streaming tables, we should accept the flags and pass them down

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/wallet/transaction.ts
```diff
@@ -61,41 +63,54 @@ export class TransactionCommand extends IronfishCommand {
...
+            header: 'Asset ID',
+          },
+          delta: {
+            header: 'Balance Change',
+          },
         },
-        delta: {
-          header: 'Balance Change',
+        {
+          ...flags,
```
hughy, line 111: Passing the flags to multiple tables will break for some flags, like filter:
image

NullSoldier:
I'll close this, I think we can only use this when there is one single output table.

PR #2962, opened 2023-01-18
Pass oclif table flags into peers table

NullSoldier:
## Summary

Need this for all oclif table features like --no-truncate

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/peers/index.ts
+++ b/ironfish-cli/src/commands/peers/index.ts
@@ -6,16 +6,20 @@ import { CliUx, Flags } from '@oclif/core'
 import blessed from 'blessed'
 import { IronfishCommand } from '../../command'
 import { RemoteFlags } from '../../flags'
+import { CommandFlags } from '../../types'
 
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/commands/peers/index.ts
```diff
@@ -95,13 +88,11 @@ export class ListCommand extends IronfishCommand {
...
   content: GetPeersResponse,
   flags: {
-    extended: boolean
     names: boolean
     all: boolean
-    sort: string
     agents: boolean
     sequence: boolean
-  },
+  } & ReturnType<typeof CliUx.ux.table.flags>,
```
hughy, line 95: It might make sense to change the `flags` that are currently passed in to `options` and have a separate `flags` argument for the table flags

I think the lint errors are coming from passing these booleans into `Cli.ux.table`

NullSoldier: Fixed this by adding a new CommandFlags type.

hughy (approved):

PR #2965, opened 2023-01-18
updates balances store for each block

hughy:
## Summary

defines AssetBalances class to encapsulate helpful increment and update functions on BufferMap<bigint>

returns AssetBalances from connectTransaction and disconnectTransaction with the asset balance deltas for the transaction

updates the balance store once for each account on each block connect and disconnect

changes updateUnconfirmedBalances to update balance for all assets with the new hash and sequence as well as any balance delta

## Testing Plan

adds tests of updating balances for all assets on each block connect and disconnect

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):
This looks really great, and very simple.

PR #2972, opened 2023-01-18
feat(cli): add a confirm flag/step to mint

mat-if:
## Summary

new asset without confirm flag
image

existing asset without confirm flag
image

with confirm flag
image

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/mint.ts
+++ b/ironfish-cli/src/commands/wallet/mint.ts
@@ -47,6 +47,10 @@ export class Mint extends IronfishCommand {
       description: 'Name for the asset',
       required: false,
     }),
+    confirm: Flags.boolean({
+      default: false,
+      description: 'Confirm without asking',
+    }),
   }
 
   async start(): Promise<void> {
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/wallet/mint.ts
```diff
@@ -138,6 +142,26 @@ export class Mint extends IronfishCommand {
...
+      const metadataString = metadata ? `Metadata: ${metadata}` : ''
+      const includeTicker = !!assetId
+      const amountString = CurrencyUtils.renderIron(amount, includeTicker, assetId)
+      const feeString = CurrencyUtils.renderIron(fee, true)
+      this.log(`
+You are about to mint ${nameString} ${metadataString}
+${amountString} plus a transaction fee of ${feeString} with the account ${account}
+
+* This action is NOT reversible *
+`)
```
mat-if, line 156: This string got pretty awkward to craft since everything is optional - it would be great to refactor this to prep a raw transaction before confirming, so we could parse out the asset id, name, etc always

NullSoldier:
Tested manually, good.

NullSoldier (approved):

PR #2982, opened 2023-01-18
fix: networkId defaults set upon run of reset command

jowparks:
## Summary
use of `reset` command will now change `networkId` to its default value
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/reset.ts
+++ b/ironfish-cli/src/commands/reset.ts
@@ -45,6 +45,9 @@ export default class Reset extends IronfishCommand {
       this.exit(0)
     }
 
+    this.sdk.internal.set('networkId', this.sdk.config.defaults.networkId)
+    this.sdk.internal.set('isFirstRun', true)
+    await this.sdk.internal.save()
     const walletDatabasePath = this.sdk.config.walletDatabasePath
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/reset.ts
```diff
@@ -45,6 +45,8 @@ export default class Reset extends IronfishCommand {
       this.exit(0)
     }
 
+    this.sdk.config.set('networkId', this.sdk.config.defaults.networkId)
+    await this.sdk.config.save()
```
hughy, line 49: I think these need to be `this.sdk.internal`

We may want to set `isFirstRun` to true, too: since we're deleting the wallet database the node will start up without creating an account unless it's the first run.

ironfish-cli/src/commands/reset.ts
```diff
@@ -45,6 +45,9 @@ export default class Reset extends IronfishCommand {
       this.exit(0)
     }
```
jowparks, line 47: We should get default from `this.sdk.config.defaults.networkId` right?

hughy: Yep, that looks right!

PR #2996, opened 2023-01-19
excludes fee paid from amount in transactions list

hughy:
## Summary

uses transaction type to determine whether to do fee correction on IRON amount in transaction.

renames 'Fee ($IRON)' column to 'Fee Paid ($IRON)' and only displays the fee if the account paid the transaction fee.

renames 'Amount ($IRON)' column to 'Net Amount ($IRON)' and only displays non-zero values.

defines type for TransactionRow and defines separate function for constructing the row from the transaction response.

## Testing Plan

Before:
image

After:
image

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/transactions.ts
+++ b/ironfish-cli/src/commands/wallet/transactions.ts
@@ -75,12 +75,12 @@ export class TransactionsCommand extends IronfishCommand {
           },
           amount: {
             header: 'Net Amount ($IRON)',
-            get: (row) => (row.amount > 0 ? CurrencyUtils.renderIron(row.amount) : ''),
+            get: (row) => (row.amount !== 0n ? CurrencyUtils.renderIron(row.amount) : ''),
             minWidth: 20,
           },
           feePaid: {
             header: 'Fee Paid ($IRON)',
-            get: (row) => (row.feePaid > 0 ? CurrencyUtils.renderIron(row.feePaid) : ''),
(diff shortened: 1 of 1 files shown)


mat-if:
Comparing master with this PR, it seems like the sends lose the net amount entirely and not just the fee
image

hughy:
> Comparing master with this PR, it seems like the sends lose the net amount entirely and not just the fee (to confirm: all but 1 of those sends were in $IRON)

Fixed! It wasn't displaying negative net $IRON amounts, but now it's changed to show any non-zero amount:
image

ironfish-cli/src/commands/wallet/transactions.ts
```diff
@@ -111,4 +114,42 @@ export class TransactionsCommand extends IronfishCommand {
...
+
+    let amount = BigInt(nativeAssetBalanceDelta?.delta ?? '0')
+
+    let feePaid = BigInt(transaction.fee)
+
+    if (transaction.type !== TransactionType.SEND) {
+      feePaid = 0n
+    } else {
+      amount += feePaid
+    }
```
NullSoldier, line 133: Can we also adjust the amount for miners transactions too?

hughy: This is what I get for miner's fees testing locally: 
image

Should it be adjusted differently?

NullSoldier (approved):

PR #3009, opened 2023-01-19
fix(cli): faucet service now checks public address validity

mat-if:
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/service/faucet.ts
+++ b/ironfish-cli/src/commands/service/faucet.ts
@@ -126,11 +126,11 @@ export default class Faucet extends IronfishCommand {
       return
     }
 
-    const baseFaucetTransactions = await api.getNextFaucetTransactions(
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/service/faucet.ts
```diff
@@ -126,14 +126,27 @@ export default class Faucet extends IronfishCommand {
       return
     }
 
-    let faucetTransactions = await api.getNextFaucetTransactions(MAX_RECIPIENTS_PER_TRANSACTION)
+    const baseFaucetTransactions = await api.getNextFaucetTransactions(
```
rohanjadvani, line 129: nit: `base` seems a little confusing to me. I think leaving this as `faucetTransactions` (or `unprocessedFaucetTransactions`) and using `validFaucetTransactions` and `invalidFaucetTransactions` below would be a little clearer.

rohanjadvani (approved):

PR #3044, opened 2023-01-20
wallet: send refactor

ygao76:
## Summary
Refactor wallet:send flow
Take in fee or fee rate
If user did not enter fee nor fee rate, prompt user with fee selection.
Fee selection is de-duped.
Call create tx and prompt user to confirm the raw transaction
Call post tx once user confirm it.

## Testing Plan
Screen Shot 2023-01-20 at 5 02 26 PM

## Breaking Change

```
[ ] Yes
```
@hughy @NullSoldier


ironfish-cli/src/commands/wallet/send.ts
```diff
@@ -143,56 +156,120 @@ export class Send extends IronfishCommand {
...
+            {
+              publicAddress: to,
+              amount: CurrencyUtils.encode(amount),
+              memo,
+              assetId,
+            },
+          ],
+          feeRate: feeRates.content.medium,
+          expiration: expiration,
+        })
```
hughy, line 200: Talked offline about reusing the repeated fields in the request object and using `Promise.all` on the multiple `createTransaction` calls

ygao76: updated

ironfish-cli/src/commands/wallet/send.ts
```diff
@@ -143,55 +157,109 @@ export class Send extends IronfishCommand {
...
+
+      const createResponses = await Promise.all(allPromises)
+
+      if (feeRates.content.low) {
+        const rawTransactionBytes = Buffer.from(createResponses[0].content.transaction, 'hex')
+        const rawTransaction = RawTransactionSerde.deserialize(rawTransactionBytes)
+
+        feeRateOptions.push({
+          value: 0,
+          name: `Low: ${rawTransaction.fee} IRON`,
```
hughy, line 206: `rawTransaction.fee` is in ORE, so it needs to be converted here

ironfish-cli/src/commands/wallet/send.ts
```diff
@@ -143,55 +157,109 @@ export class Send extends IronfishCommand {
...
+        }),
+        client.createTransaction({
+          ...createTransactionRequest,
+          feeRate: feeRates.content.high,
+        }),
+      ]
+
+      const createResponses = await Promise.all(allPromises)
+
+      if (feeRates.content.low) {
```
hughy, line 200: I know `low` is optional in the response content, and I think it is beyond the scope of this PR to change that RPC endpoint, but isn't it unexpected if the response doesn't include `low`?

What will happen here if the response doesn't include low?

hughy: I tested this case locally and I get `Invalid transaction fee rate, 0` from the `createTransaction` request above.

ironfish-cli/src/commands/wallet/send.ts
```diff
@@ -111,7 +117,15 @@ export class Send extends IronfishCommand {
...
-      fee = CurrencyUtils.decodeIron(flags.fee)
+      if (CurrencyUtils.decodeIron(flags.fee) < 1n) {
+        this.error(`The minimum fee is ${CurrencyUtils.renderOre(1n, true)}`)
+      }
+
+      fee = CurrencyUtils.encode(CurrencyUtils.decodeIron(flags.fee))
+    }
+
+    if (flags.feeRate) {
+      feeRate = CurrencyUtils.encode(CurrencyUtils.decodeIron(flags.feeRate))
```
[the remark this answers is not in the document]
ygao76: I dont think it simplify any codes. I will keep it here since this is validation check, all the validations are performed at beginning. We want to fail fast for the user.

ironfish-cli/src/commands/wallet/send.ts
```diff
@@ -143,55 +157,109 @@ export class Send extends IronfishCommand {
...
       }
 
-      const input = await CliUx.ux.prompt(
-        `Enter the fee amount in $IRON (min: ${CurrencyUtils.renderIron(
-          1n,
-        )} recommended: ${suggestedFee})`,
+      if (
+        feeRates.content.high !== feeRates.content.low &&
+        feeRates.content.high !== feeRates.content.medium
+      ) {
```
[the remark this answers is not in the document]
ygao76: Moved the check

ironfish-cli/src/commands/wallet/send.ts
ygao76: Yeah the `Invalid transaction fee rate, 0` is expected since fee rate returned is 0. I changed the fee rate rpc to return value large than 0. Since 0 fee rate is not a valid value to use.

ironfish-cli/src/commands/wallet/send.ts
```diff
@@ -36,6 +43,10 @@ export class Send extends IronfishCommand {
       char: 'o',
       description: 'The fee amount in IRON',
     }),
+    feeRate: Flags.string({
+      char: 'r',
+      description: 'The fee rate amount in IRON/Kilobyte',
```
NullSoldier, line 48: nit: `IRON/KB` is what you wrote in other places in our code.

ironfish-cli/src/commands/wallet/send.ts
```diff
@@ -143,55 +157,116 @@ export class Send extends IronfishCommand {
...
+            amount: CurrencyUtils.encode(amount),
+            memo,
+            assetId,
+          },
+        ],
+        expiration: expiration,
+      }
+
+      const allPromises = []
+      if (feeRates.content.low) {
```
NullSoldier, line 184: It's not acceptable to test truthiness against numbers in javascript. If you want to know if this is returned, you must check this against `!= null`, or `!== undefined` which is the expect value you expect. As an engineer, this immediately looks like a bug to me because I now have to dive into the backend to see if this is expected or not and that saves me 5-8 minutes if you are just specific. Many cases where numbers are checked for truthiness are just bugs because the developer didn't think about what the value could be, or worse the possible value changed later.

ironfish-cli/src/commands/wallet/send.ts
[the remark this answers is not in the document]
NullSoldier: It's unclear you're using a side effect of the decoder to use validation. That type of engineering is going to lead another developer to delete this code.

I think when writing this code, it should lead you to realize that Oclif has a strategy for this, and they expect you to write a custom flag which we haven't taken advantage of yet. https://oclif.io/docs/flags#custom-flags

If we wrote a FeeRate flag, it would ensure every time we take in a fee rate it would be consistent, it would also be a good separation of input validation.

ironfish-cli/src/commands/wallet/send.ts
ygao76: this is the external message to user, I want to be clear since some user might not aware KB is kilobyte in this context.

ironfish-cli/src/commands/wallet/send.ts
```diff
@@ -143,55 +157,116 @@ export class Send extends IronfishCommand {
...
+        )
+      }
+
+      const createResponses = await Promise.all(allPromises)
+      createResponses.forEach((createResponse, index) => {
+        const rawTransactionBytes = Buffer.from(createResponse.content.transaction, 'hex')
+        const rawTransaction = RawTransactionSerde.deserialize(rawTransactionBytes)
+
+        let name
+        switch (index) {
```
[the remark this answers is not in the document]
ygao76: I thought of that but its not, because Low, Medium, High are really just descriptions user see. So it makes sense to have the first estimate fee as Low.

hughy (approved):

PR #3050, opened 2023-01-21
fixes account reset to use mark and sweep

hughy:
## Summary

the purpose of resetting an account is to clear all of an account's data but preserve the account's keys and the record of the account.

this is functionally equivalent to deleting all account data except the data in the accounts store. however, this will crash for large accounts because we attempt to delete all data within a single database transaction.

instead of deleting the account outright we can use a mark and sweep strategy: 'resetAccount' marks an account as deleted so that its data is cleaned up asynchronously in the event loop.

to prevent deletion of the account record 'resetAccount' creates a new account with the same name and same key but a different account id. the id is not user facing and is only used for database storage.

- defines resetAccount
- splits the logic of removeAccount into removeAccount and removeAccountByName
- deletes the implementation of Account.reset

## Testing Plan

1. checked disk usage of wallet db, ran `wallet:balances`, `wallet:transactions`, `wallet:notes`
2. stopped node
3. ran `wallet:rescan --reset`
4. checked disk usage of wallet db, ran `wallet:balances`, `wallet:transactions`, `wallet:notes` again. verified command outputs were the saem and disk usage had ~doubled
5. started node again
6. checked disk usage of wallet db, saw that it had returned to approximately original value

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/removeAccount.ts
+++ b/ironfish/src/rpc/routes/wallet/removeAccount.ts
@@ -49,7 +49,7 @@ router.register<typeof RemoveAccountRequestSchema, RemoveAccountResponse>(
       }
     }
 
-    await node.wallet.removeAccount(account.name)
+    await node.wallet.removeAccountByName(account.name)
     request.end({})
   },
 )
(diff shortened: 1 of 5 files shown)


NullSoldier (approved):

NullSoldier:
Tested manually, looks good to me.

PR #3059, opened 2023-01-21
fixes calculation of confirmed balance

hughy:
## Summary

the confirmed balance for an account is equal to the unconfirmed balance of the account excluding any unconfirmed transactions.

we currently subtract the value of any notes received in unconfirmed transactions, but we do not add the value of any notes spent. this leads to a temporarily incorrect value computed for confirmed balance.

- fixes calculateUnconfirmedCountandConfirmedBalance to use transactions and transaction balance deltas
- changes CLI terms to refer to unconfirmed transactions instead of unconfirmed notes
- adds test confirming bug and fix

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/account.ts
+++ b/ironfish/src/wallet/account.ts
@@ -1,7 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { BufferMap } from 'buffer-map'
+import { BufferMap, BufferSet } from 'buffer-map'
 import MurmurHash3 from 'imurmurhash'
 import { Assert } from '../assert'
 import { BlockHeader, Transaction } from '../primitives'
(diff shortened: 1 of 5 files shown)


ironfish/src/wallet/account.ts
```diff
@@ -585,10 +587,20 @@ export class Account {
           continue
         }
 
-        if (!note.spent) {
-          unconfirmedCount++
-          confirmed -= note.note.value()
+        if (unconfirmedTransactionHashes.has(note.transactionHash)) {
```
dguenther, line 590: This makes the calculation of unconfirmed + confirmed balances better than it was, so I think we should merge it, but I think there's a further change needed.

It looks like this relies on the user having a note in the unconfirmed range, which in the case of this test, is happening because a change note is generated and returned to the user in the transaction spending another note. However, if the transaction spends an entire note without returning any change, then there won't be any notes in the confirmation range to trigger loading the transaction's balance deltas (only a note spend).

I don't think our fixtures allow us to change transaction amounts now, but I simulated this by editing fixtures to create a transaction that spends 2000000000n-value.

dguenther (approved):

ironfish/src/wallet/account.ts
hughy: Ah yeah we can't easily catch that case without iterating over all transactions right now 😬

I think adding an index 'sequenceToTransactionHash' will resolve this by providing a way for to us iterate over transactions in the sequence range.

PR #3105, opened 2023-01-24
Update CLI text descriptions 

wd021:
Updating some CLI text descriptions.

1 - Remove `IRON` from the description of send/burn/mint coin command, now that it's multi-asset
2 - wallet:burn/mint examples should be same short-form notation as wallet:send

## Breaking Change

```
[ ] Yes
```


hughy (approved):
Thanks!

Issue #3066, opened 2023-01-22
'Error: fractional component exceeds decimals' error appears when user tries to set fee 0.000000001


ipohosov:
### What happened?

STR:
1. Try to mint an asset using fee 0.000000001
Example: ironfish wallet:mint -a 100 -f mplife -m $iron_name -n $iron_name -o 0.00000001 --confirm

Actual Output:
`Error: fractional component exceeds decimals [ See: https://links.ethers.org/v5-errors-NUMERIC_FAULT ] (fault="underflow", operation="parseFixed", code=NUMERIC_FAULT, version=bignumber/5.7.0)
    at Logger.Logger.makeError (/usr/src/app/node_modules/@ethersproject/logger/src.ts/index.ts:269:28)
    at Logger.Logger.throwError (/usr/src/app/node_modules/@ethersproject/logger/src.ts/index.ts:281:20)
    at throwFault (/usr/src/app/node_modules/@ethersproject/bignumber/src.ts/fixednumber.ts:19:19)
    at Object.parseFixed (/usr/src/app/node_modules/@ethersproject/bignumber/src.ts/fixednumber.ts:104:9)
    at Function.decodeIron (/usr/src/app/node_modules/@ironfish/sdk/src/utils/currency.ts:23:12)
    at Mint.start (/usr/src/app/src/commands/wallet/mint.ts:132:27)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at Mint.run (/usr/src/app/src/command.ts:73:7)
    at async Mint._run (/usr/src/app/node_modules/@oclif/core/lib/command.js:80:22)
    at async Config.runCommand (/usr/src/app/node_modules/@oclif/core/lib/config/config.js:272:25)`

Expected Output: Some message about increasing fee because it's so small.

<img width="993" alt="Termius" src="https://user-images.githubusercontent.com/6196297/213920448-a5c21f5a-b857-4085-a250-19992db3fc3a.png">

### Version

0.1.62 @ 4ab43f3

### Debug output

```shell
Iron Fish version               0.1.62 @ 4ab43f3
Iron Fish library               0.0.39 @ 4ab43f3
Operating system                Linux x64
CPU model(s)                    AMD EPYC 7282 16-Core Processor
CPU threads                     8
RAM total                       29.39 GiB
Heap total                      4.00 GiB
Node version                    v18.12.1
ironfish in PATH                true
Garbage Collector Exposed       false
Telemetry enabled               true
[...]

Issue #3043, opened 2023-01-20
Token amount variable overflow


Qloun:
### What happened?

If you enter the number of tokens more than **92233720368.54775807**, then an error occurs:

![image](https://user-images.githubusercontent.com/35996623/213791845-3047f959-e362-45e6-80d7-8d9d48ee6dab.png)

If you enter the number of tokens **9961241799803.15787264** and more, then an overflow occurs:

![image](https://user-images.githubusercontent.com/35996623/213792562-2f9887a6-8f9c-4fc0-8733-04797528059b.png)

Thus, depending on the amount that was indicated during mint, we have the following result:

![Ironfish3](https://user-images.githubusercontent.com/35996623/213890169-228942a8-fc44-4d84-b74a-7cab5abd6693.jpg)

When overflowing during a transaction, the values of other tokens are also affected, but revert to their original value after the transaction ends. In this example, while adding the test6 token, the value of the test2 token also temporarily changed:

![image](https://user-images.githubusercontent.com/35996623/213792740-a65ac751-c747-419d-8ae9-0844f37feac2.png)

On the other wallet minting temporary affected on IRON balance:

(Mint t5)
![image](https://user-images.githubusercontent.com/35996623/213864562-9f6ab7f8-ef67-416c-a068-2534319d98bb.png)

### Version

0.1.62

### Debug output

```shell
Database in use, skipping output that requires database.
Stop the node and run the debug command again to show full output.

Iron Fish version               0.1.62 @ 4ab43f3
Iron Fish library               0.0.39 @ 4ab43f3
Operating system                Linux x64
CPU model(s)                    AMD Ryzen 7 5800X 8-Core Processor
CPU threads                     16
RAM total                       125.73 GiB
Heap total                      4.00 GiB
Node version                    v18.13.0
ironfish in PATH                true
Garbage Collector Exposed       false
Telemetry enabled               true
Node name                       megapapka
Block graffiti                  megapapka
```

### Graffiti

megapapka

Issue #2984, opened 2023-01-18
Attemtped to Mint 100000000000 tokens, but only 15908979783.59414784 created


ADADarkpool:
### What happened?

Attemtped to Mint 100000000000 tokens, but only 15908979783.59414784 created

### Version

v0.1.61

### Debug output

```shell
─$ docker run --rm --tty --interactive --network host --volume ~/.ironfishPhase3:/root/.ironfish ghcr.io/iron-fish/ironfish:latest wallet:mint -a 100000000000000000 -n ironSHIB
Enter metadata for the new asset: 
Enter the fee amount in $IRON (min: 0.00000001) [0.00000001]: 

You are about to mint Name: ironSHIB 
100000000000000000.00000000 plus a transaction fee of $IRON 0.00000001 with the account ironSHIB

* This action is NOT reversible *

Do you confirm (Y/N)?: Y
Creating the transaction: [████████████████████████████████████████] 100% | ETA: 0s

Minted asset ironSHIB from ironSHIB
Asset Identifier: 67c0ba876ff388f9eca93456914e14ed8548eacff796b89200f77260855a3971
Value: 15908979783.59414784

Transaction Hash: 85d23d55ae761a071116a26a919c7f568c0ea010b75b89899af71ee5438f239f

Find the transaction on https://explorer.ironfish.network/transaction/85d23d55ae761a071116a26a919c7f568c0ea010b75b89899af71ee5438f239f (it can take a few minutes before the transaction appears in the Explorer)
```

### Graffiti

DarkPool

PR #3106, opened 2023-01-24
Check validity of coin/fee input numbers

wd021:
## Summary

There have been a number of issues raised relating to overflow/underflow of coin amounts people are inputting. See https://github.com/iron-fish/ironfish/issues/2984, https://github.com/iron-fish/ironfish/issues/3043, https://github.com/iron-fish/ironfish/issues/3066. 

This task is aimed at improving input checks by 
1) catching underflow errors from `decodeIron` that result in https://github.com/iron-fish/ironfish/issues/3066
2) checking fee/coin inputs are within the valid range of **MINIMUM_ORE_AMOUNT** to **MAXIMUM_ORE_AMOUNT**

## Testing Plan
test `wallet:send/mint/burn` commands with fee and coin inputs that are outside the valid range

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/flags.ts
+++ b/ironfish-cli/src/flags.ts
@@ -8,6 +8,8 @@ import {
   DEFAULT_USE_RPC_IPC,
   DEFAULT_USE_RPC_TCP,
   DEFAULT_USE_RPC_TLS,
+  MAXIMUM_ORE_AMOUNT,
+  MINIMUM_ORE_AMOUNT,
 } from '@ironfish/sdk'
 import { Flags, Interfaces } from '@oclif/core'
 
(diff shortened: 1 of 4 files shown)


hughy (commented):
Checking the min/max ORE amount for every input is a bit repetitive. I think we could either incorporate that logic into `CurrencyUtils.decodeIron` or into a custom flag here: https://github.com/iron-fish/ironfish/blob/693ff0fb5292c4392c058ce98d3c2ce25075d897/ironfish-cli/src/flags.ts#L109

wd021:
Originally was thinking about adding the logic to `CurrencyUtils.decodeIron`, but think it's best to leave it in its simplified form. Makes sense to have a custom flag though, missed that completely. Will update this.

wd021:
1 - Updated `parseIron` to catch underflow errors from `CurrencyUtils.decodeIron` and check for min, max ore values.
2 - send/mint/burn now using `IronFlag` for both amount and fee inputs. 
3 - `parseIron` is also used now to catch input errors when using interactive `CliUx.ux.prompt`.

wd021:
@hughy is there a reason mint/burn isn't using feeRate?

ironfish-cli/src/flags.ts
```diff
@@ -111,14 +113,23 @@ export const IronFlag = Flags.custom<bigint, IronOpts>({
...
     const { largerThan, flagName } = opts ?? {}
-    const value = CurrencyUtils.decodeIron(input)
-    if (largerThan !== undefined && value <= largerThan) {
-      reject(new Error(`The minimum ${flagName} is ${CurrencyUtils.renderOre(1n, true)}`))
-    }
+    try {
+      const value = CurrencyUtils.decodeIron(input)
+
+      if (largerThan !== undefined && value <= largerThan) {
+        reject(new Error(`The minimum ${flagName} is ${CurrencyUtils.renderOre(1n, true)}`))
```
hughy, line 123: nit - I just noticed this in the existing error message -- it should use `largerThan` instead of `1n` for the minimum amount

hughy:
> @hughy is there a reason mint/burn isn't using feeRate?

we just haven't gotten around to adding the fee rate logic to those commands yet 😅

hughy (commented):
One small comment, but this looks great!

wd021:
updated

hughy (approved):
Thanks this looks great!

PR #3107, opened 2023-01-24
feat(ironfish): Add max mint value in verification

rohanjadvani:
## Summary

Check mint value against `Number.MAX_SAFE_INTEGER`. Should address #3043

## Testing Plan

Unit test

## Breaking Change

```
[ ] Yes
```


mat-if (approved):
We'll probably need to check this for burns and spends too, although spends will only be able to happen as UI validation I think

PR #3110, opened 2023-01-24
Trusted setup ceremony

danield9tqh:
## Summary
This branch includes changes for the trusted setup ceremony
- Code derived from [sapling-mpc](https://github.com/zcash-hackworks/sapling-mpc) and [ZCash's phase 2 repo](https://github.com/ebfull/phase2)
- Code for trusted setup coordination server
- Code for `ceremony` command for users to participate

## Testing Plan
Local testing + added unit tests

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/ceremony.ts
```diff
@@ -0,0 +1,170 @@
...
+
+export default class Ceremony extends IronfishCommand {
+  static description = 'Contribute randomness to the Iron Fish trusted setup'
+
+  static flags = {
+    [VerboseFlagKey]: VerboseFlag,
+    [DataDirFlagKey]: DataDirFlag,
+    host: Flags.string({
+      parse: (input: string) => Promise.resolve(input.trim()),
+      default: '127.0.0.1',
```
danield9tqh, line 23: Should add a default here. I've added `ceremony.ironfish.network` to DNS

ironfish-cli/src/commands/ceremony.ts
```diff
@@ -0,0 +1,170 @@
...
+    const tempDir = this.sdk.config.tempDir
+    await fsAsync.mkdir(tempDir, { recursive: true })
+
+    const inputPath = path.join(tempDir, 'params')
+    const outputPath = path.join(tempDir, 'newParams')
+
+    let localHash: string | null = null
+
+    // Prompt for randomness
+    let randomness: string | null = await CliUx.ux.prompt(
```
danield9tqh, line 46: Need to create this blog post and insert

ironfish-cli/src/commands/ceremony.ts
```diff
@@ -0,0 +1,170 @@
...
+
+      CliUx.ux.action.start('Contribution uploaded. Waiting for server to verify')
+    })
+
+    client.onContributionVerified.on(({ hash, downloadLink, contributionNumber }) => {
+      CliUx.ux.action.stop()
+
+      if (hash !== localHash) {
+        this.error(
+          'Hashes do not match. Please contact the Iron Fish team with this error message.',
```
danield9tqh, line 130: maybe we should print out the hashes here so we have more info to go on if there's an error

ironfish-cli/src/commands/ceremony.ts
```diff
@@ -0,0 +1,170 @@
...
+        )
+      }
+
+      this.log(
+        `Thank you for your contribution to the Iron Fish Ceremony. You have successfully contributed at position #${contributionNumber}. The public hash of your contribution is "${hash}". You can view your contributed file at ${downloadLink}.`,
```
danield9tqh, line 135: another ceremony I've seen divided the hash into multiple sections. I thought it was a good way to display the hash since it's very long
```
aaaaaa aaaaaa aaaaaa
aaaaaa aaaaaa aaaaaa
aaaaaa aaaaaa aaaaaa
```

ironfish-cli/src/trusted-setup/client.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  private onError = (error: unknown): void => {
+    this.logger.error(`Server error ${ErrorUtils.renderError(error)}`)
+  }
+
+  private onData(data: Buffer): void {
+    const message = data.toString('utf-8')
+    let parsedMessage
+    try {
+      parsedMessage = JSON.parse(message) as CeremonyServerMessage
```
danield9tqh, line 88: could cause a crash later on if it's not a CeremonyServerMessage but other valid JSON. Probably should use yup here instead

ironfish-mpc/COPYRIGHT
```diff
@@ -0,0 +1,14 @@
+Copyrights in the "sapling-mpc" library are retained by their contributors. No
```
danield9tqh, line 1: Maybe we should link original repo

ironfish-mpc/README.md
```diff
@@ -0,0 +1,19 @@
+# ironfish-mpc
```
danield9tqh, line 1: Nice

ironfish-cli/src/commands/ceremony.ts
dguenther: I'll switch this default over to that subdomain then.

ironfish-cli/src/commands/ceremony.ts
dguenther: I'm going to remove that for now and add an airtable task to re-add it once the blog post is up.

ironfish-cli/src/utils/s3.ts
```diff
@@ -181,4 +195,161 @@ async function uploadToBucket(
...
+    }
+
+    truncated = response.IsTruncated || false
+    commandParams = { Bucket: bucket, Marker: response.Contents?.slice(-1)[0]?.Key }
+  }
+
+  return keys
+}
+
+export async function copyBucketObject(
```
dguenther, line 280: Should we get rid of this? It's not used anywhere now that we're re-uploading

ironfish-cli/src/utils/s3.ts
```diff
@@ -181,4 +195,161 @@ async function uploadToBucket(
...
+    Bucket: bucket,
+    CopySource: `/${bucket}/${sourceFile}`,
+    Key: destFile,
+  }
+
+  const command = new CopyObjectCommand(commandParams)
+  return s3.send(command)
+}
+
+export async function deleteBucketObject(
```
dguenther, line 296: Could probably be named `deleteFromBucket` to align with the other function names a bit more

ironfish-cli/src/commands/ceremony.ts
dguenther: Changed this to print out the hashes.

ironfish-cli/src/commands/ceremony.ts
dguenther: Changed this to print out the hashes like so:
```
        95799e84 5202b5e2 e9d09f8c db981fcd
        ceec1d6e 784761e7 c4d88d5b 5ba5cb74
        dc9eda71 fa5e50c4 07957ae3 22b01fa2
        53f0c341 93b20d8b 7ee5e20f 464da9b1
```

ironfish-cli/src/trusted-setup/client.ts
dguenther: I'm not super worried about this since it should only be connecting to one of our servers, so we should be able to control the messages the client is receiving. On the server-side though, I think as long as we try/catch the JSON.parse, we should be able to check for the existence of `method`, and disconnect clients if they send any other message. Yup would be more thorough and what we use elsewhere though.

ironfish-cli/src/utils/s3.ts
danield9tqh: yea good idea

ironfish-cli/src/trusted-setup/client.ts
danield9tqh: yea I think I meant to comment on the server code.

ironfish-cli/src/utils/s3.ts
danield9tqh: good idea renamed

ironfish-cli/src/trusted-setup/client.ts
danield9tqh: Added a task to finish this

dguenther (approved):

PR #3111, opened 2023-01-24
Add char flag datadir

danield9tqh:
## Summary
Add a character shortcut for the --datadir flag since it's used often in development. Remove other -d flags because they are not used often

## Testing Plan
local testing. Also checked terraform to see that we are [not using -d flag](https://github.com/iron-fish/terraform/blob/1e667c6984b1827e8e7bc5f41bd94294acac7667/modules/pool/main.tf#L36)

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/chain/genesisblock.ts
+++ b/ironfish-cli/src/commands/chain/genesisblock.ts
@@ -168,7 +168,7 @@ export default class GenesisBlockCommand extends IronfishCommand {
     }
 
     // Exit if dry run, otherwise confirm
-    if (flags.dryRun) {
+    if (flags.dry) {
       this.exit(0)
     } else {
       const result = await CliUx.ux.confirm('\nCreate the genesis block? (y)es / (n)o')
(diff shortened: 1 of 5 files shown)


jowparks (approved):

ironfish-cli/src/commands/chain/genesisblock.ts
```diff
@@ -53,7 +53,6 @@ export default class GenesisBlockCommand extends IronfishCommand {
       description: 'The amount of coins in the genesis block',
     }),
     dryRun: Flags.boolean({
```
NullSoldier, line 55: We should rename this to --dry

NullSoldier (approved):

ironfish-cli/src/commands/chain/genesisblock.ts
danield9tqh: done

ironfish-cli/src/commands/chain/genesisblock.ts
```diff
@@ -52,8 +52,7 @@ export default class GenesisBlockCommand extends IronfishCommand {
       default: '42000000',
       description: 'The amount of coins in the genesis block',
     }),
-    dryRun: Flags.boolean({
-      char: 'd',
+    dry: Flags.boolean({
```
NullSoldier, line 55: this creates a bug

PR #3112, opened 2023-01-24
feat: export mnemonic (1/2)

jowparks:
## Summary
- Exposes Mnemonic phrase generation from the rust side into TS side. 
- Uses the `bip39` library we already have installed.
- Adds `LanguageCode` enum for usage typescript side. Sadly the support for enums is crap in napi, manual mapping is the best we can do.

### Implementation note
I considered adding the function to a new `.rs` file (ie `keys.ts`) but I saw there was a pattern for key helper functions already set up in `lib.rs` didn't want to diverge in this PR.

## Followup PR
[Next pull request will be added here. It will update the export command and call this NAPI function.](https://github.com/iron-fish/ironfish/pull/3207)

## Testing Plan
Tests
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust-nodejs/tests/demo.test.slow.ts
+++ b/ironfish-rust-nodejs/tests/demo.test.slow.ts
@@ -30,7 +30,6 @@ describe('Demonstrate the Sapling API', () => {
   it('Should return phrase words from key input', () => {
     const key = generateKey()
     const words = wordsSpendingKey(key.spending_key, LanguageCode.English);
-    console.log(words)
     expect(words.split(' ').length).toEqual(24)
   })
 
--- a/ironfish-rust-nodejs/tests/demo.test.slow.ts
+++ b/ironfish-rust-nodejs/tests/demo.test.slow.ts
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-import { Asset, DECRYPTED_NOTE_LENGTH } from '..'
+import { Asset, DECRYPTED_NOTE_LENGTH, LanguageCode, wordsSpendingKey } from '..'
 import {
   initializeSapling,
   generateKey,
@@ -25,6 +25,13 @@ describe('Demonstrate the Sapling API', () => {
     expect(typeof key.outgoing_view_key).toBe('string')
     expect(typeof key.public_address).toBe('string')
     expect(typeof key.spending_key).toBe('string')
+  })
+
+  it('Should return phrase words from key input', () => {
+    const key = generateKey()
+    const words = wordsSpendingKey(key.spending_key, LanguageCode.English);
+    console.log(words)
+    expect(words.split(' ').length).toEqual(24)
   })
 
   it('Should generate a new public address given a spending key', () => {
--- a/ironfish-rust/src/keys/mod.rs
+++ b/ironfish-rust/src/keys/mod.rs
@@ -197,8 +197,8 @@ impl SaplingKey {
     /// a seed. This isn't strictly necessary for private key, but view keys
     /// will need a direct mapping. The private key could still be generated
     /// using bip-32 and bip-39 if desired.
-    pub fn words_spending_key(&self, language: &Language) -> Result<String, IronfishError> {
-        let mnemonic = Mnemonic::from_entropy(&self.spending_key, *language)
+    pub fn words_spending_key(&self, language: Language) -> Result<String, IronfishError> {
+        let mnemonic = Mnemonic::from_entropy(&self.spending_key, language)
             .map_err(|_| IronfishError::InvalidEntropy)?;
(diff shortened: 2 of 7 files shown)


ironfish-rust/src/keys/mod.rs
```diff
@@ -194,11 +202,15 @@ impl SaplingKey {
     /// a seed. This isn't strictly necessary for private key, but view keys
     /// will need a direct mapping. The private key could still be generated
     /// using bip-32 and bip-39 if desired.
```
jowparks, line 204: Updated unused method

ironfish-rust-nodejs/src/structs/key.rs
```diff
@@ -0,0 +1,32 @@
...
+#[napi]
+pub const SPEND_KEY_LENGTH: u32 = SPEND_KEY_SIZE as u32;
+
+#[napi]
+pub const MNEMONIC_LENGTH: u32 = MNEMONIC_SIZE as u32;
+
+#[napi(js_name = "SaplingKey")]
+pub struct SaplingKey {}
+
+#[napi]
```
jowparks, line 20: Exposed method in napi

EvanJRichard (approved):
LGTM AFAICT! 🚀

ironfish-rust-nodejs/tests/demo.test.slow.ts
```diff
@@ -27,6 +27,13 @@ describe('Demonstrate the Sapling API', () => {
     expect(typeof key.spending_key).toBe('string')
   })
 
+  it('Should return phrase words from key input', () => {
+    const key = generateKey()
+    const words = wordsSpendingKey(key.spending_key, LanguageCode.English);
+    console.log(words)
```
mat-if, line 33:
```suggestion
```

ironfish-rust-nodejs/tests/demo.test.slow.ts
```diff
@@ -27,6 +27,13 @@ describe('Demonstrate the Sapling API', () => {
     expect(typeof key.spending_key).toBe('string')
   })
 
+  it('Should return phrase words from key input', () => {
```
mat-if, line 30: I love that you made a test here for this, it's something I've wanted us to do more. 🙌

ironfish-rust/src/keys/test.rs
```diff
@@ -112,3 +113,15 @@ fn test_hex_conversion() {
...
     assert!(PublicAddress::from_hex("invalid").is_err());
 }
+
+#[test]
+fn test_words_spending_key() {
+    let key = SaplingKey::generate_key();
+    let words = key
+        .words_spending_key(&bip39::Language::English)
+        .expect("Should return words");
+    let expected_words = MnemonicType::for_key_size(&key.spending_key[..].len() * 8)
```
mat-if, line 123:
```suggestion
    let expected_words = MnemonicType::for_key_size(key.spending_key.len() * 8)
```

This just takes a `usize`, so taking a reference of a slice of an array is not needed

ironfish-rust/src/keys/mod.rs
```diff
@@ -194,10 +197,9 @@ impl SaplingKey {
...
-    pub fn words_spending_key(&self, language_code: &str) -> Result<String, IronfishError> {
-        let language = Language::from_language_code(language_code)
-            .ok_or(IronfishError::InvalidLanguageEncoding)?;
-        let mnemonic = Mnemonic::from_entropy(&self.spending_key, language).unwrap();
+    pub fn words_spending_key(&self, language: &Language) -> Result<String, IronfishError> {
+        let mnemonic = Mnemonic::from_entropy(&self.spending_key, *language)
```
mat-if, line 201: Since `from_entropy` takes the `Language` struct directly, we should just take it in here as well instead of taking the reference. You'll have to remove a few `&` up the chain, but it works otherwise 

```suggestion
    pub fn words_spending_key(&self, language: Language) -> Result<String, IronfishError> {
        let mnemonic = Mnemonic::from_entropy(&self.spending_key, language)
```

mat-if (approved):
Few little things, but looks good 👍

ironfish-rust-nodejs/tests/demo.test.slow.ts
```diff
@@ -27,6 +27,12 @@ describe('Demonstrate the Sapling API', () => {
...
+    const key = generateKey()
+    const words = wordsSpendingKey(key.spending_key, LanguageCode.English);
+    expect(words.split(' ').length).toEqual(24)
```
NullSoldier, line 33: you can actually hard code the result here, because it should never change if you want

NullSoldier:
This is pretty good, just a couple good rust tips here from mat.

ironfish-rust-nodejs/tests/demo.test.slow.ts
jowparks: Tried it, `generateKey` actually generates a random key, so the value changes. Gonna leave as is

PR #3113, opened 2023-01-24
chore: const spending key size

jowparks:
## Summary
Instead of hardcoding `32` everywhere, moved this size to a constant. It will be useful for the import/export work that @EvanJRichard and I are working on.

## Testing Plan
unittests pass
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/keys/mod.rs
+++ b/ironfish-rust/src/keys/mod.rs
@@ -29,7 +29,7 @@ pub use view_keys::*;
 mod test;
 
 const EXPANDED_SPEND_BLAKE2_KEY: &[u8; 16] = b"Iron Fish Money ";
-
+pub const SPEND_KEY_SIZE: usize = 32;
 /// A single private key generates multiple other key parts that can
 /// be used to allow various forms of access to a commitment note:
 ///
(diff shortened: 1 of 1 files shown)


ironfish-rust/src/keys/mod.rs
```diff
@@ -29,7 +29,7 @@ pub use view_keys::*;
 mod test;
 
 const EXPANDED_SPEND_BLAKE2_KEY: &[u8; 16] = b"Iron Fish Money ";
-
+pub const KEY_SIZE: usize = 32;
```
mat-if, line 32: `SPEND_KEY_SIZE` or something might be a little more clear - we deal with a lot of different keys

ironfish-rust/src/keys/mod.rs
```diff
@@ -90,8 +90,9 @@ impl SaplingKey {
         let proof_authorizing_key =
             jubjub::Fr::from_bytes_wide(&Self::convert_key(spending_key, 1));
 
-        let mut outgoing_viewing_key = [0; 32];
-        outgoing_viewing_key[0..32].clone_from_slice(&Self::convert_key(spending_key, 2)[0..32]);
+        let mut outgoing_viewing_key = [0; KEY_SIZE];
+        outgoing_viewing_key[0..KEY_SIZE]
+            .clone_from_slice(&Self::convert_key(spending_key, 2)[0..32]);
```
mat-if, line 95:
```suggestion
            .clone_from_slice(&Self::convert_key(spending_key, 2)[0..KEY_SIZE]);
```

mat-if (approved):

ironfish-rust/src/keys/mod.rs
jowparks: Aren't all the keys the same size?

jowparks: Note: yes they will most likely stay the same size

PR #3114, opened 2023-01-24
CLI: optionally support import and export as base58 checksummed string, rather than hex string

EvanJRichard:
## Summary
This PR proposes to add an optional flag `--base58` to the `wallet:import` and `wallet:export` commands. If provided by a user, the CLI will use [bs58check-ts](https://www.npmjs.com/package/bs58check-ts) to convert any uses of the default hex encoding to checksummed base58 encoding.

Additionally, this adds new RPC endpoints that are as import and export, but with base58.

## Testing Plan
To-do.

## Breaking Change

```
[ ] Yes
[X] No
```


package.json
```diff
@@ -61,7 +61,9 @@
     "axios": "0.21.4",
     "ejs": "^3.1.7",
     "handlebars": "4.7.7",
-    "node-notifier": "8.0.1",
-    "eslint-plugin-import/**/json5": "^1.0.2"
+    "node-notifier": "8.0.1"
+  },
```
EvanJRichard, line 65: yarn keeps trying to remove `"eslint-plugin-import/**/json5": "^1.0.2"` so I obeyed it but we can remove this change

ironfish-cli/src/commands/wallet/export.ts
```diff
@@ -45,6 +51,12 @@ export class ExportCommand extends IronfishCommand {
     const client = await this.sdk.connectRpc(local)
     const response = await client.exportAccount({ account })
```
jowparks, line 53: I think instead of modifying the response inplace  it would be better to create a separate RPC (ie `exportAccountBase58`), or just extend the  RPC to take an input of `format` and pass in base58 (or hex as default). 

I know we talked about keeping things in the CLI layer, but it seems like the code will get more cruft if we do that here because of the way the response is used. Also we get the benefit of our RPC being exposed to other wallet developers

ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -35,11 +41,11 @@ export class ImportCommand extends IronfishCommand {
 
     let account: AccountImport | null = null
     if (importPath) {
```
jowparks, line 44: Instead of using/passing around flags, I think we could just try all the formats (currently hex, base58), and throw useful error if none work. This is better from a usability perspective

package.json
```diff
@@ -61,7 +61,6 @@
     "axios": "0.21.4",
     "ejs": "^3.1.7",
     "handlebars": "4.7.7",
-    "node-notifier": "8.0.1",
-    "eslint-plugin-import/**/json5": "^1.0.2"
```
EvanJRichard, line 65: yarn keeps trying to remove this package, should I obey it or remove the changeset from the PR?

ironfish-cli/src/commands/wallet/export.ts
EvanJRichard: Got it! 👍

ironfish-cli/src/commands/wallet/import.ts
EvanJRichard: I see, so for the import codepath, instead of using `--base58`, just try both (for example) `client.importAccount` and `importAccountBase58` and only throw if both fail?

ironfish/src/rpc/routes/wallet/exportAccountBase58.ts
```diff
@@ -0,0 +1,52 @@
...
+
+router.register<typeof ExportAccountRequestBase58Schema, ExportAccountBase58Response>(
+  `${ApiNamespace.wallet}/exportAccountBase58`,
+  ExportAccountRequestBase58Schema,
+  (request, node): void => {
+    const account = getAccount(node, request.data.account).serialize()
+    account.spendingKey = bs58safe.encode(Buffer.from(account.spendingKey, 'hex'))
+    account.incomingViewKey = bs58safe.encode(Buffer.from(account.incomingViewKey, 'hex'))
+    account.outgoingViewKey = bs58safe.encode(Buffer.from(account.outgoingViewKey, 'hex'))
+    request.end({ account: account })
```
NullSoldier, line 50: Should we encode the whole account as base58 instead of each individual key? Maybe I'm not sure what the encoding for this should be. Should it just be the key, or the whole account that you can just import as a base 58?

EvanJRichard:
Closing in favor of #3213.

PR #3116, opened 2023-01-24
Remove wallet in fee estimator

ygao76:
## Breaking Change

```
[ ] Yes
```
@NullSoldier @hughy


ironfish/src/memPool/feeEstimator.ts
```diff
@@ -60,6 +58,7 @@ export class FeeEstimator {
     }
 
     let currentBlockHash = chain.latest.hash
+    this.maxBlockSizeBytes = chain.consensus.parameters.maxBlockSizeBytes
```
NullSoldier, line 61: Lets use the consensus here

NullSoldier (approved):

PR #3127, opened 2023-01-25
fix: solo miner will wait for new work after submitting

mat-if:
## Summary

This will still submit multiple blocks in the case of having a high thread count and a very easy target, but it will be at most n submissions where n is the number of threads. This still allows the node to process and create a new block template, making it much harder for the node to get overloaded

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):
I tested this manually as well, it doesn't freeze anymore. Thank you!

PR #3141, opened 2023-01-26
Update rust-toolchain.toml

sinascorpion:
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "1.66.1"
+channel = "1.67.1"


NullSoldier:
We don't allow PR's from the community to upgrade any versions. Please see our contributing agreement, https://github.com/iron-fish/ironfish/blob/master/CONTRIBUTING.md#upgrading-package-versions

Even if you say it's harmless we will just refuse these types of PRs as a policy in the case that someone gets lazy and doesn't fully vet the upgrade.

On top of that, upgrading is not simple. Many upgrades have regressions and we have no evidence that you've fully tested for regressions and are taking on the responsibility of continuously monitoring for regressions and to revert the change. Such a regression happened when we upgraded NAPI recently.

If you want to upgrade one of our packages, please open a ticket asking for it as a feature request and why you want the upgrade.

PR #3145, opened 2023-01-26
automatically migrates Docker nodes

hughy:
## Summary

adds '--upgrade' to the start command in the Dockerfile

## Breaking Change

```
[ ] Yes
```


jowparks (approved):

hughy:
Closing since we don't want to automatically migrate nodes for all users who use Docker and there may be cases where we don't want to automatically migrate our own nodes

PR #3151, opened 2023-01-27
trusted-setup: Send correct queue location

dguenther:
## Summary

In the other place we send queueLocation, we treat "0" as the currently contributing client. This aligns the queueLocation to be the same.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/trusted-setup/server.ts
+++ b/ironfish-cli/src/trusted-setup/server.ts
@@ -170,7 +170,7 @@ export class CeremonyServer {
   private onConnection(socket: net.Socket): void {
     const client = new CeremonyServerClient({ socket, id: uuid(), logger: this.logger })
     this.queue.push(client)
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/trusted-setup/server.ts
```diff
@@ -170,7 +170,7 @@ export class CeremonyServer {
   private onConnection(socket: net.Socket): void {
     const client = new CeremonyServerClient({ socket, id: uuid(), logger: this.logger })
     this.queue.push(client)
-    client.send({ method: 'joined', queueLocation: this.queue.length })
+    client.send({ method: 'joined', queueLocation: this.queue.length - 1 })
```
danield9tqh, line 173: Is this right? if there are 4 people in the queue (including them) then their position would be 4 right?

danield9tqh:
Closing in favor of https://github.com/iron-fish/ironfish/pull/3224

PR #3152, opened 2023-01-27
Add chain.getDifficulty rpc route

hairtail:
## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/chain/getDiffifulty.ts
```diff
@@ -0,0 +1,57 @@
...
+    difficulty: yup.number().defined(),
+  })
+  .defined()
+
+router.register<typeof GetDifficultyRequestSchema, GetDifficultyResponse>(
+  `${ApiNamespace.chain}/getDifficulty`,
+  GetDifficultyRequestSchema,
+  async (request, node): Promise<void> => {
+    let sequence = null
+    let block = null
```
NullSoldier, line 39:
```suggestion
    let sequence = node.chain.head.sequence
    let block = node.chain.head
```

ironfish/src/rpc/routes/chain/getDiffifulty.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    if (request.data?.sequence) {
+      sequence = request.data.sequence
+      block = await node.chain.getHeaderAtSequence(sequence)
+    }
+    if (!block) {
+      block = node.chain.head
+      sequence = block.sequence
+    }
```
NullSoldier, line 48: This will use the head and give you the wrong difficulty in the case that the block is not found

ironfish/src/rpc/routes/chain/getDiffifulty.ts
```diff
@@ -0,0 +1,57 @@
...
+router.register<typeof GetDifficultyRequestSchema, GetDifficultyResponse>(
+  `${ApiNamespace.chain}/getDifficulty`,
+  GetDifficultyRequestSchema,
+  async (request, node): Promise<void> => {
+    let sequence = null
+    let block = null
+
+    if (request.data?.sequence) {
+      sequence = request.data.sequence
+      block = await node.chain.getHeaderAtSequence(sequence)
```
NullSoldier, line 43: throw ViewValidationError with a 404 here if the block is not found

ironfish/src/rpc/routes/chain/getDiffifulty.ts
```diff
@@ -0,0 +1,57 @@
...
+    if (!block) {
+      block = node.chain.head
+      sequence = block.sequence
+    }
+
+    Assert.isNotNull(sequence)
+
+    request.end({
+      sequence,
+      difficulty: Number(block.target.toDifficulty()),
```
NullSoldier, line 54: you cannot cast difficulty to a number, it will get truncated.

ironfish/src/rpc/routes/chain/getDiffifulty.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+    if (!block) {
+      block = node.chain.head
+      sequence = block.sequence
+    }
+
+    Assert.isNotNull(sequence)
+
+    request.end({
+      sequence,
```
NullSoldier, line 53: I think we should send the hash too

```suggestion
      sequence,
      hash: block.hash
```

ironfish/src/rpc/routes/chain/getDiffifulty.ts
```diff
@@ -0,0 +1,64 @@
...
+    let block = node.chain.head
+
+    if (request.data?.sequence) {
+      sequence = request.data.sequence
+      const sequenceBlock = await node.chain.getHeaderAtSequence(sequence)
+      if (!sequenceBlock) {
+        throw new ValidationError(`No block found at provided sequence`)
+      } else {
+        block = sequenceBlock
+      }
```
NullSoldier, line 51: I think this can be a little cleaner.

```suggestion
      const sequenceBlock = await node.chain.getHeaderAtSequence(request.data.sequence)
      if (!sequenceBlock) {
        throw new ValidationError(`No block found at provided sequence`)
      }

      sequence = sequenceBlock.sequence
      block = sequenceBlock
```

ironfish/src/rpc/routes/chain/getDiffifulty.ts
```diff
@@ -0,0 +1,64 @@
...
+  GetDifficultyRequestSchema,
+  async (request, node): Promise<void> => {
+    let sequence = node.chain.head.sequence
+    let block = node.chain.head
+
+    if (request.data?.sequence) {
+      sequence = request.data.sequence
+      const sequenceBlock = await node.chain.getHeaderAtSequence(sequence)
+      if (!sequenceBlock) {
+        throw new ValidationError(`No block found at provided sequence`)
```
NullSoldier, line 48: We should include the sequence in the error, in the case that the user didnt pass one they can't tell what the sequence was.

```suggestion
        throw new ValidationError(`No block found at sequence ${sequence}`)
```

ironfish/src/rpc/routes/chain/getDiffifulty.ts
```diff
@@ -0,0 +1,64 @@
...
+      } else {
+        block = sequenceBlock
+      }
+    }
+
+    if (!block) {
+      throw new ValidationError(`Chain header not found`)
+    }
```
NullSoldier, line 56: Is this check neccessary not that you added the check above? block is not nullable here right?

ironfish/src/rpc/routes/chain/getDiffifulty.ts
```diff
@@ -0,0 +1,59 @@
...
+      if (!sequenceBlock) {
+        throw new ValidationError(`No block found at sequence ${request.data.sequence}`)
+      }
+      sequence = sequenceBlock.sequence
+      block = sequenceBlock
+    }
+
+    request.end({
+      sequence,
+      hash: BlockHashSerdeInstance.serialize(block.hash),
```
NullSoldier, line 55:
```suggestion
      hash: block.hash.toString('hex'),
```

ironfish/src/rpc/routes/chain/index.ts
```diff
@@ -8,6 +8,7 @@ export * from './followChain'
 export * from './getBlock'
 export * from './getBlockInfo'
 export * from './getChainInfo'
+export * from './getDiffifulty'
```
NullSoldier, line 11: This file has a typo

NullSoldier (approved):

NullSoldier:
I added a test because the test was missing

PR #3155, opened 2023-01-27
Update MAX_BURN_VALUE in rawTransaction.ts 

Qloun:
Correcting MAX_MINT_VALUE according to #3043

## Summary
Due to bug #3043, currently the maximum number of tokens will be 92233720368.54775807.
In addition, setting the MAX_BURN_VALUE constant before determining the maximum possible number of tokens is incorrect, since MAX_TOKEN_AMOUNT can be greater than MAX_MINT_VALUE (and therefore it will not be possible to burn all the tokens at one transaction if the mint operation with the MAX_MIN_VALUE was previously executed multiple times).

## Breaking Change

```
[ ] Yes
```


rohanjadvani:
This is intended behavior. We want to limit the mint and burn value per transaction, so theoretically the total token supply can be much greater than a single mint amount from a description. Similarly, we don't intend on the supply to be burned all in one transaction; it's expected to take multiple transactions to burn if we're past the maximum

PR #3159, opened 2023-01-27
feat(ironfish): Manage pending state for assets

rohanjadvani:
## Summary

* Adds `blockHash` and `sequence` to asset records
  * There is no accompanying migration since this store is currently empty, so no records need to be migrated
* Adjusts backfill migration to populate `blockHash` and `sequence`
* Connects mints and burns when pending transactions are added to an account
* Increments supply only if connected mints and burns if the associated transaction is on a block

## Testing Plan

Added and adjusted unit tests

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/account.ts
```diff
@@ -259,19 +269,23 @@ export class Account {
         'Existing asset owner should match public address',
       )
 
-      const existingSupply = existingAsset.supply
-      const supply = existingSupply - value
+      let supply = existingAsset.supply
+      if (blockHash && sequence) {
```
hughy, line 273: If there's no blockHash and sequence we won't update anything on the asset. Should we return early?

ironfish/src/wallet/account.ts
```diff
@@ -398,19 +416,19 @@ export class Account {
...
+        transaction,
+        blockHash: null,
+        sequence: null,
+        submittedSequence,
+        timestamp: new Date(),
+        assetBalanceDeltas,
+      }
+
+      await this.saveConnectedMintsToAssetsStore(transactionValue, tx)
+      await this.saveConnectedBurnsToAssetsStore(transactionValue, tx)
```
hughy, line 429: We may not need to process pending burns at all if supply only changes on connected burns

ironfish/src/wallet/account.ts
```diff
@@ -398,19 +416,19 @@ export class Account {
...
+      const transactionValue = {
+        transaction,
+        blockHash: null,
+        sequence: null,
+        submittedSequence,
+        timestamp: new Date(),
+        assetBalanceDeltas,
+      }
+
+      await this.saveConnectedMintsToAssetsStore(transactionValue, tx)
```
hughy, line 428: nit - wondering if the `Connected` still makes sense in the name if we call this for pending transactions?

ironfish/src/wallet/account.ts
```diff
@@ -186,29 +186,26 @@ export class Account {
...
+
+      await this.walletDb.saveTransaction(this, transaction.hash(), transactionValue, tx)
     })
 
     return assetBalanceDeltas
   }
 
   async saveConnectedMintsToAssetsStore(
-    transaction: Transaction,
+    { blockHash, sequence, transaction }: TransactionValue,
```
hughy, line 208: I didn't know you could do this in TypeScript 🤯

rohanjadvani: :cowboy_hat_face:

ironfish/src/wallet/account.ts
rohanjadvani: Yep removed

hughy (approved):

PR #3160, opened 2023-01-27
Fix default host for trusted setup

danield9tqh:
## Summary
Trusted setup does not use http so we need to remove that. Also adding some logic to display connection errors

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/ceremony.ts
+++ b/ironfish-cli/src/commands/ceremony.ts
@@ -159,8 +159,9 @@ export default class Ceremony extends IronfishCommand {
     let connected = false
     while (!connected) {
       CliUx.ux.action.start('Connecting')
-      connected = await client.start()
-      CliUx.ux.action.stop(connected ? 'done' : 'error')
+      const error = await client.start()
+      connected = error === null
+      CliUx.ux.action.stop(error ? `Error connecting: ${error}` : 'done')
 
       if (!connected) {
         this.log('Unable to connect to contribution server. Retrying in 5 seconds.')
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/commands/ceremony.ts
```diff
@@ -159,8 +159,9 @@ export default class Ceremony extends IronfishCommand {
     let connected = false
     while (!connected) {
       CliUx.ux.action.start('Connecting')
-      connected = await client.start()
-      CliUx.ux.action.stop(connected ? 'done' : 'error')
+      const error = await client.start()
+      connected = error === null
+      CliUx.ux.action.stop(error ? `Error connecting: ${error}` : 'done')
```
dguenther, line 164: What do you think about doing the error logging as a separate `debug` log so it doesn't appear to users by default? These logs feel a bit too technical for average users:

```
Connecting... Error connecting: connect ECONNREFUSED 127.0.0.1:9040
Unable to connect to contribution server. Retrying in 5 seconds.
```

But I could also see the case where "Error" doesn't really help the user solve any connection issues.

dguenther (approved):

ironfish-cli/src/commands/ceremony.ts
danield9tqh: Yea I think it would be good to print it out to the user so they are able to report it to us. Ideally this should just not happen at all but if it does it would be good for them to report the specific error even if they don't understand it themselves

PR #3161, opened 2023-01-28
Add wallet:post command

NullSoldier:
## Summary

This will take an unsigned transaction and post the transaction.

```
ironfish wallet:post f9c799f94a0469c53ea1131c5d9823c44c809f5bd611468d949a0c61a17a3c66

Posting transaction... done
f116384fa4c18c1169962401a63cfa1041455032a48e87e350028c65295b2d1666ecfbc866078fb5ddfad428ddeb40ff
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/post.ts
+++ b/ironfish-cli/src/commands/wallet/post.ts
@@ -1,7 +1,9 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { CliUx } from '@oclif/core'
+import { CurrencyUtils, RawTransaction, RawTransactionSerde } from '@ironfish/sdk'
+import { CliUx, Flags } from '@oclif/core'
+import { flags } from '@oclif/core/lib/parser'
 import { IronfishCommand } from '../../command'
 import { RemoteFlags } from '../../flags'
 
@@ -18,19 +20,31 @@ export class PostCommand extends IronfishCommand {
 
   static flags = {
     ...RemoteFlags,
+    confirm: Flags.boolean({
+      default: false,
+      description: 'Confirm without asking',
+    }),
   }
 
   static args = [
     {
       name: 'transaction',
+      required: true,
       parse: (input: string): Promise<string> => Promise.resolve(input.trim()),
       description: 'The raw transaction in hex encoding',
     },
   ]
 
   async start(): Promise<void> {
-    const { args } = await this.parse(PostCommand)
+    const { flags, args } = await this.parse(PostCommand)
     const transaction = args.transaction as string
+
+    const serialized = Buffer.from(transaction)
+    const raw = RawTransactionSerde.deserialize(serialized)
+
+    if (!flags.confirm && !this.confirm(raw)) {
+      this.exit(0)
+    }
 
     CliUx.ux.action.start(`Posting transaction`)
     const client = await this.sdk.connectRpc()
@@ -38,5 +52,23 @@ export class PostCommand extends IronfishCommand {
     CliUx.ux.action.stop()
 
     this.log(response.content.transaction)
+  }
+
+  confirm(raw: RawTransaction): Promise<boolean> {
+    let spending = 0n
+    for (const recieve of raw.receives) {
+      spending += recieve.note.value()
+    }
+
+    this.log(
+      `You are about to post a transaction that sends ${spending}, with ${
+        raw.mints.length
+      } mints and ${raw.burns.length} burns with a fee ${CurrencyUtils.renderIron(
+        raw.fee,
+        true,
+      )}`,
+    )
+
+    return CliUx.ux.confirm('Do you want to post this (Y/N)?')
   }
 }
--- a/ironfish-cli/src/commands/wallet/post.ts
+++ b/ironfish-cli/src/commands/wallet/post.ts
@@ -1,7 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { CurrencyUtils, RawTransaction, RawTransactionSerde } from '@ironfish/sdk'
+import { CurrencyUtils, RawTransaction, RawTransactionSerde, Transaction } from '@ironfish/sdk'
 import { CliUx, Flags } from '@oclif/core'
 import { flags } from '@oclif/core/lib/parser'
 import { IronfishCommand } from '../../command'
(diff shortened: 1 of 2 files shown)


ygao76 (changes requested):
I don’t like the cli name

NullSoldier:
> I don’t like the cli name

Yeah, sounds good. Let's rename this, and the RPC all at once in it's own PRs.

ironfish-cli/src/commands/wallet/post.ts
```diff
@@ -0,0 +1,34 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { CliUx } from '@oclif/core'
+import { IronfishCommand } from '../../command'
+import { RemoteFlags } from '../../flags'
+
+export class PostCommand extends IronfishCommand {
+  static description = `Post a raw transaction`
```
ygao76, line 9: Can you be specific on what will happen when post a raw transaction. Also include the instruction on offline signing here, like what bitcoin does for the examples in https://developer.bitcoin.org/reference/rpc/sendrawtransaction.html#examples

ironfish-cli/src/commands/wallet/post.ts
```diff
@@ -0,0 +1,34 @@
...
+      parse: (input: string): Promise<string> => Promise.resolve(input.trim()),
+      description: 'The raw transaction in hex encoding',
+    },
+  ]
+
+  async start(): Promise<void> {
+    const { args } = await this.parse(PostCommand)
+    const transaction = args.transaction as string
+
+    CliUx.ux.action.start(`Posting transaction`)
```
ygao76, line 27: Can you deserialize the raw transaction, display the details and confirm with user they want to sign this transaction

ironfish-cli/src/commands/wallet/post.ts
```diff
@@ -0,0 +1,34 @@
...
+    const client = await this.sdk.connectRpc()
+    const response = await client.postTransaction({ transaction })
+    CliUx.ux.action.stop()
```
ygao76, line 30: Can you display the transaction returned in response? User needs this to broadcast transaction

NullSoldier: It's already happening below here, look a few lines down.

ironfish-cli/src/commands/wallet/post.ts
NullSoldier: I added docs. I think you asked me to add some docs here on how to get the raw transaction but this is opened before those commands exist, so that's not really possible. Let's add those docs in the PR for when we can get a raw tranaction.

ironfish-cli/src/commands/wallet/post.ts
NullSoldier: Yep, I'll add that

ironfish-cli/src/commands/wallet/post.ts
```diff
@@ -0,0 +1,42 @@
...
+import { CliUx } from '@oclif/core'
+import { IronfishCommand } from '../../command'
+import { RemoteFlags } from '../../flags'
+
+export class PostCommand extends IronfishCommand {
+  static summary = 'Post a raw transaction'
+
+  static description = `Use this command to post a raw transaction.
+
+  The output is a finalized posted transaction. The transaction is also added to the wallet, and sent out to the network.`
```
ygao76, line 13: Post will not add to wallet and send to network. That logic will move to broadcast.

ironfish-cli/src/commands/wallet/post.ts
ygao76: I mean display the details in transaction like what send cli does. 
```
Sending ${CurrencyUtils.renderIron(amount, true, assetId)} to ${to} from ${from}
Transaction Hash: ${transaction.hash().toString('hex')}
Transaction fee: ${CurrencyUtils.renderIron(transaction.fee(), true)}
```
also add description, user can use this result and call broadcast cli to post on network like what bitcoin dose.

ironfish-cli/src/commands/wallet/post.ts
NullSoldier: Added

ironfish-cli/src/commands/wallet/post.ts
NullSoldier: The fee is already in the confirmation, but I'll add the hash.

The CLI command shouldn't output tutorial steps. That's added to the docs / command CLI help like you asked for.

NullSoldier: I added the posted hash to the message.

ironfish-cli/src/commands/wallet/post.ts
NullSoldier: We don't have the ability to remove that from the wallet, it's an eventual consistency property of our wallet. No matter what, once the transaction is added to the wallet the wallet will always try to broadcast it out in the event loop.

ygao76: It will not in Post API and wallet.post(), we have discussed this last friday with all the engs. The approach is to split that logic from post to broadcast, broadcast still belongs wallet. But currently post logic will be new post() + broadcast()

ironfish-cli/src/commands/wallet/post.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  static examples = [
+    '$ ironfish wallet:post 618c098d8d008c9f78f6155947014901a019d9ec17160dc0f0d1bb1c764b29b4...',
```
ygao76, line 18: Can you add the previous step wallet:send and  next step wallet:broadcast in the examples? 

Follow what bitcoin does 
```
Create a transaction:

bitcoin-cli createrawtransaction "[{\"txid\" : \"mytxid\",\"vout\":0}]" "{\"myaddress\":0.01}"
Sign the transaction, and get back the hex:

bitcoin-cli signrawtransactionwithwallet "myhex"
Send the transaction (signed hex):

bitcoin-cli sendrawtransaction "signedhex"
```

ironfish-cli/src/commands/wallet/post.ts
NullSoldier: Ok, so we'll need to avoid adding it to the wallet as well then if we want to avoid broadcasting.

So create() -> post() -> add()

ironfish-cli/src/commands/wallet/post.ts
NullSoldier: That's not related to this overall command, that's "how to perform a task" that's outside the function of this specific command. We can put that into the description if you like, but creating raw transactions does not exist yet, right?

ygao76: Creating a raw transaction will be `wallet:send --rawTransaction`. It does not exist yet, I will update this once I add the flag.

ygao76 (approved):

ygao76:
Add this PR to https://github.com/iron-fish/ironfish/pull/3271

PR #3162, opened 2023-01-28
adds transactionAmounts datastore to walletDb

hughy:
## Summary

the transactions datastore includes a field 'assetBalanceDeltas' on each record that stores a mapping from asset ID to the net amount or balance delta for that asset in the transaction.

however, there are some cases, such as computing the amount of an asset that an account has available to spend, where it is useful to separate the amount that the account input into the transaction from the amount that the account received output from the transaction.

the transactionAmounts datastore stores the input and output amounts for each account, transaction hash, and asset id.

this datastore will allow us to efficiently compute the 'available' or 'spendable' balance by reading the amount spent in a transaction without iterating over a transaction's spends and looking up decrypted notes for those spends in the walletDb.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/__fixtures__/account.test.ts.fixture
+++ b/ironfish/src/wallet/__fixtures__/account.test.ts.fixture
@@ -1314,5 +1314,131 @@
         }
       ]
     }
+  ],
+  "Accounts addPendingTransaction should add transaction amounts to transactionAmounts": [
+    {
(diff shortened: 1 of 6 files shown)


mat-if (approved):

hughy:
Re-targeted to from `staging` to `available-balances` branch: I plan on adding all of the changes needed for calculating the available balance (migration for this new store, calculating the balance, etc.) there before merging into staging

PR #3164, opened 2023-01-28
Add wallet:transaction:add and wallet.addTransaction

NullSoldier:
## Summary

This new command will add a transaction to your wallet and optionally broadcast it, similar to the create transaction flow.

```
$ ironfish wallet:transaction:add 943ac7b05082160a9cdb0b946b13c328b6a2eb6a0f7334c8846ff7eaa801694d

Transaction added for account jason
```

## Testing Plan
1. Create a posted transaction
2. Serialize it to a hex string
3. Call `ironfish wallert:transactions:add <binary>`

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/wallet/addTransaction.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    node.memPool.acceptTransaction(transaction)
+
+    if (request.data.broadcast) {
+      node.wallet.broadcastTransaction(transaction)
+    }
+
+    request.end({
+      account: account.name,
+    })
```
hughy, line 68: A transaction could be related to multiple accounts in the wallet. Should we return them all in a response?

ygao76 (approved):

ironfish/src/rpc/routes/wallet/addTransaction.ts
NullSoldier: Good find, fixed.

ygao76 (approved):

ironfish/src/rpc/routes/wallet/addTransaction.ts
```diff
@@ -0,0 +1,70 @@
...
+      throw new ValidationError(`Invalid transaction, reason: ${String(verify.reason)}`, 400)
+    }
+
+    await node.wallet.addPendingTransaction(transaction)
+
+    const accounts = await AsyncUtils.filter(node.wallet.listAccounts(), (account) =>
+      account.hasTransaction(transaction.hash()),
+    )
+
+    if (accounts.length) {
```
hughy, line 54: Should this be `if (accounts.length === 0)`?

NullSoldier: Fixed

hughy (approved):

PR #3172, opened 2023-01-28
Use getBlockSize & getTransactionSize in getBlock rpc route

hairtail:
## Summary
I'm working on an ironfish explorer, which fetches most chain data with rpc call `getBlock` & `getBlockInfo`. Then I find that two `blockSize` of the specified block from, 1:  `chain.getBlock`, 2:  `chain.followChain`,  are different.
## Breaking Change

```
[ ] Yes
```


dguenther (approved):
Looks good to me. As far as I can tell, we don't use this RPC route in our own code anywhere, so it should be safe to change, and we want to use `getBlockSize` and `getTransactionSize` everywhere anyway.

PR #3178, opened 2023-01-28
Add the ability to watch a transaction status

NullSoldier:
## Summary

https://asciinema.org/a/aV8PobDK15767HElhB7yMKez3

You can now pass --watch to send or use the new command wallet:transaction:watch

## Testing Plan
Run `ironfish wallet:transaction:watch` or `ironfish wallet:send --watch`

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/utils/transaction.ts
+++ b/ironfish-cli/src/utils/transaction.ts
@@ -4,6 +4,7 @@
 
 import {
   Assert,
+  createRootLogger,
   Logger,
   PromiseUtils,
   RpcClient,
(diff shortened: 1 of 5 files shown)


ironfish-cli/src/utils/transaction.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    const now = Date.now()
+    const duration = now - lastTime
+    lastTime = now
+
+    CliUx.ux.action.stop(
+      `${last.content.transaction.status} -> ${
+        response.content.transaction.status
+      }: ${TimeUtils.renderSpan(duration)}`,
+    )
```
hughy, line 67: Do you think it would be helpful to display the block hash and/or sequence when the status changes?

ironfish-cli/src/utils/transaction.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    const response = await client.getAccountTransaction({ account, hash, confirmations })
+
+    if (response.content.transaction == null) {
+      CliUx.ux.action.stop(`Transaction ${hash} deleted while watching it.`)
+      return
+    }
+
+    if (response.content.transaction.status === last.content.transaction.status) {
+      await PromiseUtils.sleep(2000)
```
hughy, line 55: nit - since block time is ~1m we may not need to poll this frequently

hughy (approved):
this is awesome!

ironfish-cli/src/utils/transaction.ts
NullSoldier: I'll slow it down

ironfish-cli/src/utils/transaction.ts
NullSoldier: I think I can expand this to show some contextual information

Issue #3180, opened 2023-01-28
Feature: add balance to selection menu


anotherhabit:
### Description

quality of life improvement; display how many tokens available on the asset selection menu for burn, send and mint commands.

very helpful when have multiple assets, especially if more than one has same name for whatever reason.

PR #3190, opened 2023-01-29
Add balances to asset menu

wd021:
## Summary
Implementing https://github.com/iron-fish/ironfish/issues/3180. Adding confirmed balances to asset selection menu.

## Testing Plan
Run `ironfish wallet:send/burn/mint` and check balances when selecting the asset from menu

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/utils/asset.ts
+++ b/ironfish-cli/src/utils/asset.ts
@@ -41,7 +41,7 @@ export async function selectAsset(
       const displayName = BufferUtils.toHuman(Buffer.from(assetResponse.content.name, 'hex'))
       assetOptions.push({
         value: assetId,
-        name: `${assetId} (${displayName}) - ${CurrencyUtils.renderIron(confirmed)}`,
+        name: `${assetId} (${displayName}) (${CurrencyUtils.renderIron(confirmed)})`,
       })
     }
   }
--- a/ironfish-cli/src/utils/asset.ts
+++ b/ironfish-cli/src/utils/asset.ts
@@ -3,7 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 import { Asset } from '@ironfish/rust-nodejs'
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/utils/asset.ts
```diff
@@ -34,14 +34,14 @@ export async function selectAsset(
...
-  for (const { assetId } of balances) {
+  for (const { assetId, confirmed } of balances) {
     const assetResponse = await client.getAsset({ id: assetId })
 
     if (assetResponse.content.name) {
       const displayName = BufferUtils.toHuman(Buffer.from(assetResponse.content.name, 'hex'))
       assetOptions.push({
         value: assetId,
-        name: `${assetId} (${displayName})`,
+        name: `${assetId} (${displayName}) - ${CurrencyUtils.renderIron(confirmed)}`,
```
hughy, line 44: I worry that `$IRON - 1.00000000` might lead some users to think that we're displaying a negative balance here

hughy (commented):
Thanks for this!

Just one tiny comment on avoiding the appearance of negative balances

ironfish-cli/src/utils/asset.ts
wd021: yea, good point. it might just be me, but putting double brackets there feels like bad ux. how about formatting it with`${displayName}, id:${assetId}, balance:${CurrencyUtils.renderIron(confirmed)}`?

Screen Shot 2023-02-01 at 1 35 03 PM

Screen Shot 2023-02-01 at 1 36 16 PM

anotherhabit: in my opinion the second example without brackets feels more intuitive as it is closer to how the output of the balances command.

dguenther: I think I'd like to go with the first one, since it causes the asset IDs to align vertically. I can make that change here quick and merge it

dguenther (approved):

Issue #3191, opened 2023-01-29
Feature: Show asset name for burning


wd021:
### Description

Burning asset only shows assetId. Would be nice to also return the asset name.

PR #3192, opened 2023-01-29
Add asset name to wallet:burn confirmation

wd021:
## Summary
Implementing https://github.com/iron-fish/ironfish/issues/3191. Burning asset now shows asset name and id in transaction confirmation, in the same way that minting asset does.

## Testing Plan
Run `ironfish wallet:burn` and see that asset name shows in the transaction confirmation

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/burnAsset.test.ts
+++ b/ironfish/src/rpc/routes/wallet/burnAsset.test.ts
@@ -8,7 +8,7 @@ import {
   usePostTxFixture,
 } from '../../../testUtilities'
 import { createRouteTest } from '../../../testUtilities/routeTest'
-import { CurrencyUtils } from '../../../utils'
+import { BufferUtils, CurrencyUtils } from '../../../utils'
 
 describe('burnAsset', () => {
   const routeTest = createRouteTest(true)
@@ -87,6 +87,7 @@ describe('burnAsset', () => {
 
       expect(response.content).toEqual({
         assetId: asset.id().toString('hex'),
+        name: BufferUtils.toHuman(asset.name()),
         hash: burnTransaction.hash().toString('hex'),
         value: burnTransaction.burns[0].value.toString(),
       })
--- a/ironfish/src/rpc/routes/wallet/burnAsset.test.ts
+++ b/ironfish/src/rpc/routes/wallet/burnAsset.test.ts
@@ -8,7 +8,7 @@ import {
   usePostTxFixture,
 } from '../../../testUtilities'
 import { createRouteTest } from '../../../testUtilities/routeTest'
-import { BufferUtils, CurrencyUtils } from '../../../utils'
+import { CurrencyUtils } from '../../../utils'
 
 describe('burnAsset', () => {
   const routeTest = createRouteTest(true)
@@ -87,7 +87,7 @@ describe('burnAsset', () => {
 
       expect(response.content).toEqual({
         assetId: asset.id().toString('hex'),
-        name: BufferUtils.toHuman(asset.name()),
+        name: asset.name().toString('hex'),
         hash: burnTransaction.hash().toString('hex'),
         value: burnTransaction.burns[0].value.toString(),
       })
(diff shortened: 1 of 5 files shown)


ironfish/src/rpc/routes/wallet/burnAsset.ts
```diff
@@ -60,10 +62,15 @@ router.register<typeof BurnAssetRequestSchema, BurnAssetResponse>(
       throw new ValidationError('Invalid burn amount')
     }
 
+    const assetBuffer = Buffer.from(request.data.assetId, 'hex')
+
+    const assetResponse = await node.chain.getAssetById(assetBuffer)
```
hughy, line 67: @ygao76 do we plan to change the data included in `BurnDescription` as part of offline signing?

If requests and/or responses include the asset name then we may not need to look up the asset here.

ironfish/src/rpc/routes/wallet/burnAsset.test.ts
```diff
@@ -87,6 +87,7 @@ describe('burnAsset', () => {
 
       expect(response.content).toEqual({
         assetId: asset.id().toString('hex'),
+        name: BufferUtils.toHuman(asset.name()),
```
NullSoldier, line 90: we always send this back as hex16 encoded `assetName` then you can toHuman on the other side, or people cannot get the raw asset name buffer.

```suggestion
        name: asset.name().toString('hex'),
```

ironfish/src/rpc/routes/wallet/burnAsset.ts
```diff
@@ -19,6 +19,7 @@ export interface BurnAssetRequest {
 export interface BurnAssetResponse {
   assetId: string
   hash: string
+  name: string
```
NullSoldier, line 22: `assetName` here

ironfish/src/rpc/routes/wallet/burnAsset.ts
```diff
@@ -75,6 +82,7 @@ router.register<typeof BurnAssetRequestSchema, BurnAssetResponse>(
     request.end({
       assetId: burn.assetId.toString('hex'),
       hash: transaction.hash().toString('hex'),
+      name: BufferUtils.toHuman(assetResponse.name),
```
NullSoldier, line 85: We always send this back as hex16 encoded `assetName` field, like in other RPC. You can toHuman on the other side, or people cannot get the raw asset name buffer.

```suggestion
      assetName: assetResponse.name.toString('hex'),
```

ironfish/src/rpc/routes/wallet/burnAsset.ts
```diff
@@ -60,10 +62,15 @@ router.register<typeof BurnAssetRequestSchema, BurnAssetResponse>(
       throw new ValidationError('Invalid burn amount')
     }
 
+    const assetBuffer = Buffer.from(request.data.assetId, 'hex')
```
NullSoldier, line 65: this is not an asset buffer, it's the asset id in buffer form, but you can drop the buffer syntax since asset ids internally are buffers.

```suggestion
    const assetId = Buffer.from(request.data.assetId, 'hex')
```

ironfish/src/rpc/routes/wallet/burnAsset.ts
NullSoldier, line 67: you can just call this `asset` its not a response
```suggestion
    const asset = await node.chain.getAssetById(assetBuffer)
```

ironfish/src/rpc/routes/wallet/burnAsset.ts
wd021: also updated mintAssetResponse name to `hex`, instead of `utf8`

wd021:
@NullSoldier applied review changes. 

1 - updated the mintAssetResponse name to be `hex` and not `utf8`. 
2 - rebased and burn function no longer uses the rpc burnAsset method, instead just using postTransaction. I added a getAsset call after transaction to get assetName. The updates in this PR for burnAsset rpc are not really relevant, but i think its still fine to keep it.

NullSoldier (approved):

PR #3205, opened 2023-01-30
feat(ironfish): introduce an event loop for mining pool

mat-if:
## Summary

The goal here is to simplify the pool logic a little bit. With the coming refactors, it will be much easier if we have the main logic in one place, so it can be extended and refactored without introducing a bunch of random timeouts or intervals. This also will make it easier to read and maintain the code going forward, I hope.

One thing I want to point out: I intentionally left the existing intervals in the pool class, since they're a bit more time-sensitive. It would feel awkward if these were getting delayed because of the event loop, which I expect to be a long running function much of the time. It will eventually check for confirmed blocks, transactions, etc. on top of doing more logic around figuring out payout amounts, payout buckets, etc.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/mining/pool.ts
+++ b/ironfish/src/mining/pool.ts
@@ -212,8 +212,9 @@ export class MiningPool {
     const eventLoopStartTime = new Date().getTime()
 
     if (this.nextPayoutAttempt <= eventLoopStartTime) {
+      const nextPayoutAttempt = new Date().getTime() + this.attemptPayoutInterval * 1000
       await this.shares.createPayout()
-      this.nextPayoutAttempt = new Date().getTime() + this.attemptPayoutInterval * 1000
+      this.nextPayoutAttempt = nextPayoutAttempt
     }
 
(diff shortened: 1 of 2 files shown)


hughy (approved):
This looks good to me!


One thing I thought about while reading this is that the interval between pool payout attempts might be irregular and might not match up with `poolAttemptPayoutInterval` since it depends on the event loop interval and duration.

That seems fine to me, but I thought that we might change the name of the config. I can't think of anything that would make more sense than the current name though 🥴

mat-if:
@hughy That's a good point - I'll modify this so it more closely mimics existing behavior by using the timestamp taken before the function runs. This will allow it to run immediately after a previous payout attempt finishes if it takes more than the time interval, but that's actually an improvement over the current logic where they will execute simultaneously.

Eventually, this flag is going to be removed and it will be able to constantly create transactions as needed once more of the refactoring is done.

PR #3207, opened 2023-01-30
feat: export mnemonic cli (2/2)

jowparks:
## Summary
Adds support for `mnemonic` export of private key to the `ironfish` cli.

Depends on:
https://github.com/iron-fish/ironfish/pull/3112
## Testing Plan

https://user-images.githubusercontent.com/26990067/215891175-84310de2-a709-4b8d-a5b0-a977b7662e00.mov

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/utils/language.ts
+++ b/ironfish-cli/src/utils/language.ts
@@ -1,0 +1,35 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { LanguageCode } from '@ironfish/rust-nodejs'
+import inquirer from 'inquirer'
+
+export const LANGUAGES = {
+  English: LanguageCode.English,
+  ChineseSimplified: LanguageCode.ChineseSimplified,
+  ChineseTraditional: LanguageCode.ChineseTraditional,
+  French: LanguageCode.French,
+  Italian: LanguageCode.Italian,
+  Japanese: LanguageCode.Japanese,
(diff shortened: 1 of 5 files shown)


EvanJRichard (approved):
LGTM :rocket:

ironfish-cli/src/utils/language.ts
```diff
@@ -0,0 +1,35 @@
...
+  English = LanguageCode.English,
+  ChineseSimplified = LanguageCode.ChineseSimplified,
+  ChineseTraditional = LanguageCode.ChineseTraditional,
+  French = LanguageCode.French,
+  Italian = LanguageCode.Italian,
+  Japanese = LanguageCode.Japanese,
+  Korean = LanguageCode.Korean,
+  Spanish = LanguageCode.Spanish,
+}
+
```
jowparks, line 20: Any cleaner way to represent these types? Three separate values seems unfortunate to me

jowparks: Thank you @danield9tqh for the great suggestions :)

ironfish-cli/src/utils/language.ts
```diff
@@ -0,0 +1,35 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { LanguageCode } from '@ironfish/rust-nodejs'
+import inquirer from 'inquirer'
+
```
jowparks, line 7: Unfortunately we need to re-export the constant here because the export from napi is `const enum` which can only be referenced as values

PR #3208, opened 2023-01-30
adds a migration to backfill transactionAmounts

hughy:
## Summary

the migration iterates over the accounts in the walletDb to backfill the new transactionAmounts datastore

for each transaction in an account it
        - adds up the values of all the notes that the account spent in the
          transaction
        - adds up the values of all the notes that the account received in the
          transaction
        - adds records to the transactionAmounts store for each asset in the
          transaction that the account either spent or received

## Testing Plan

image

## Breaking Change

```
[ ] Yes
```


hughy:
The migration added here will change from `017-...` to `018-...` if `wallet-assets` merges to staging before `available-balance`

PR #3209, opened 2023-01-30
feat(ironfish): getBlockInfo RPC supports confirmation

mat-if:
## Summary

This adds a `confirmations` optional flag to the request, which will override the default config. Otherwise, it will use the default `confirmations` config to determine whether the block is confirmed or not. It will include this boolean in the response's `metadata` as `confirmed`

## Breaking Change

```
[ ] Yes
```


hughy (approved):
✅ 

I think this makes a lot of sense to have alongside `main`!

PR #3213, opened 2023-01-31
Support bech32-encoded account import/export in cli

EvanJRichard:
## Summary
This is the same as #3114, but using bech32 encoding instead of base58. It uses the workflows discussed. 
- Wallet export now defaults to `bech32` rather than `json`, prefixed with `ironfishaccount00000`. 
- Wallet import now tries bech32 before json automatically.
- Adds bech32 as a dependency.

Depends on #3251 

## Testing Plan
File workflow example: 

https://user-images.githubusercontent.com/5766842/216449967-9ec65994-68dc-458e-b5ef-49cbdad11730.mov

## Breaking Change

```
[ ] Yes
[x] No
```


jowparks:
Can you list out the main points of what the PR accomplishes in description?
ie
```
- makes Json non default param
- adds bech32 library as dependency for encoding
- ...
```

ironfish-cli/src/commands/wallet/export.ts
```diff
@@ -45,7 +50,18 @@ export class ExportCommand extends IronfishCommand {
     const client = await this.sdk.connectRpc(local)
     const response = await client.exportAccount({ account })
```
jowparks, line 52: You will want to rebase on current `staging`, there will be conflicts from my `mnemonic` pr

ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -65,7 +66,13 @@ export class ImportCommand extends IronfishCommand {
   async importFile(path: string): Promise<AccountImport> {
     const resolved = this.sdk.fileSystem.resolve(path)
     const data = await this.sdk.fileSystem.readFile(resolved)
-    return JSONUtils.parse<AccountImport>(data)
+    try {
+      const decodedWords = bech32m.bech32m.decode(data).words
+      const decodedBytes = bech32m.bech32m.fromWords(decodedWords)
+      return JSONUtils.parse<AccountImport>(decodedBytes.toString())
+    } catch (e) {
```
jowparks, line 73: Lets not swallow all errors, catch only the JSON parse error type (whatever it is)

ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -65,7 +66,13 @@ export class ImportCommand extends IronfishCommand {
...
+      return JSONUtils.parse<AccountImport>(data)
+    }
```
jowparks, line 75:
```suggestion
    try {
      const decodedWords = bech32m.bech32m.decode(data).words
      const decodedBytes = bech32m.bech32m.fromWords(decodedWords)
      data = decodedBytes.toString()
    } catch (e DecodingErrorWords DecodingErrorBytes) {
      // failure suggests already in json format
    }
    return JSONUtils.parse<AccountImport>(data)
```

jowparks: Not sure of exact error types, but this captures the essence

ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -84,10 +91,29 @@ export class ImportCommand extends IronfishCommand {
...
     process.stdin.off('data', onData)
 
-    return JSONUtils.parse<AccountImport>(data)
+    try {
+      const decodedWords = bech32m.bech32m.decode(data).words
+      const decodedBytes = bech32m.bech32m.fromWords(decodedWords)
+      return JSONUtils.parse<AccountImport>(decodedBytes.toString())
+    } catch (e) {
+      return JSONUtils.parse<AccountImport>(data)
+    }
```
jowparks, line 100: Same as above, also if this is a reusable pattern, should move it into its own method to avoid duplication.

ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -84,10 +91,29 @@ export class ImportCommand extends IronfishCommand {
...
 
   async importTTY(): Promise<AccountImport> {
+    const bech32 = await CliUx.ux.prompt(
+      "Enter the bech32 encoded account (string beginning with 'ironfishprivatekey'), or hit enter to begin legacy hex spending key workflow",
+      {
+        required: true,
+      },
+    )
+
+    if (bech32 !== '') {
```
jowparks, line 111: move to method

ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -84,10 +91,29 @@ export class ImportCommand extends IronfishCommand {
...
+      const decodedBytes = bech32m.bech32m.fromWords(decodedWords)
+      return JSONUtils.parse<AccountImport>(decodedBytes.toString())
+    } catch (e) {
+      return JSONUtils.parse<AccountImport>(data)
+    }
   }
 
   async importTTY(): Promise<AccountImport> {
+    const bech32 = await CliUx.ux.prompt(
+      "Enter the bech32 encoded account (string beginning with 'ironfishprivatekey'), or hit enter to begin legacy hex spending key workflow",
```
jowparks, line 105: The prefix `ironfishprivatekey` is a misnomer, this is actually an an entire account that is encoded in json.

ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -84,10 +91,29 @@ export class ImportCommand extends IronfishCommand {
...
+      const decodedWords = bech32m.bech32m.decode(data).words
+      const decodedBytes = bech32m.bech32m.fromWords(decodedWords)
+      return JSONUtils.parse<AccountImport>(decodedBytes.toString())
+    } catch (e) {
+      return JSONUtils.parse<AccountImport>(data)
+    }
   }
 
   async importTTY(): Promise<AccountImport> {
+    const bech32 = await CliUx.ux.prompt(
```
jowparks, line 104: Can use more interactive selector rather than requiring user to read entire message (with examples in text output)
https://github.com/iron-fish/ironfish/blob/027b96c406bb15e754e3829aef87c373313d382f/ironfish-cli/src/utils/asset.ts#L49-L56

ironfish/src/rpc/routes/wallet/exportAccount.ts
```diff
@@ -13,6 +13,7 @@ export type ExportAccountResponse = {
     incomingViewKey: string
     outgoingViewKey: string
     publicAddress: string
```
jowparks, line 15: We probably want to move this version stuff to a separate PR. It would be nice to write some tests to verify the version handling is working as expected (for future proofing when we go to v2)

ironfish-cli/src/commands/wallet/export.ts
EvanJRichard: Rebased

ironfish/src/rpc/routes/wallet/exportAccount.ts
EvanJRichard: Opened as #3222

ironfish-cli/src/commands/wallet/import.ts
EvanJRichard: `bech32` seems to only throw `Error` with no types, unless I'm missing something: https://github.com/bitcoinjs/bech32/blob/master/src/index.ts#L111

ironfish-cli/src/commands/wallet/import.ts
EvanJRichard: Nice, I like that much better.

package.json
```diff
@@ -63,5 +63,6 @@
     "handlebars": "4.7.7",
     "node-notifier": "8.0.1",
     "eslint-plugin-import/**/json5": "^1.0.2"
```
jowparks, line 65: There was a deletion of `"eslint-plugin-import/**/json5": "^1.0.2"` in this PR. I assumed it was an accident so I removed it.

ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -84,10 +98,34 @@ export class ImportCommand extends IronfishCommand {
...
+    try {
+      data = ImportCommand.bech32ToJSON(data)
+    } catch (e) {
+      CliUx.ux.info('Unable to decode bech32, assuming input is already JSON')
+    }
     return JSONUtils.parse<AccountImport>(data)
   }
 
   async importTTY(): Promise<AccountImport> {
+    const response: { decodingChoice: string } = await inquirer.prompt<{
```
jowparks, line 110: This UX isn't great, we can do better. Lets just let the user input text, try all formats, if fails then default to asking for account name and spending key.

ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -27,6 +29,13 @@ export class ImportCommand extends IronfishCommand {
     },
   ]
 
+  static bech32ToJSON(bech32: string): string {
```
jowparks, line 32: This is failable, we should wrap this in a try/catch and return null if it fails. That way we can try and check output for comment below.

ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -64,7 +73,12 @@ export class ImportCommand extends IronfishCommand {
 
   async importFile(path: string): Promise<AccountImport> {
     const resolved = this.sdk.fileSystem.resolve(path)
-    const data = await this.sdk.fileSystem.readFile(resolved)
+    let data = await this.sdk.fileSystem.readFile(resolved)
+    try {
+      data = ImportCommand.bech32ToJSON(data)
```
jowparks, line 78: Handle updated method for possible `null` return

ironfish-cli/src/commands/wallet/import.ts
jowparks: The reason I don't like this is most users will likely not know what either of these types are. They will just be carrying in the account info, copy pasta.

ironfish-cli/src/commands/wallet/import.ts
EvanJRichard: I see, so instead of 
```
try {
  bech32ToJSON
  } catch {
  ```
  everywhere, move the try/catch into bech32ToJSON, and check for null where bech32ToJSON gets used? Got it

File: ironfish-cli/src/commands/wallet/import.ts (hunk 10 above)
Comment (EvanJRichard) [reply ironfish-cli/src/commands/wallet/import.ts]: Good point.

File: ironfish-cli/src/commands/wallet/export.ts [hunk 13]
```diff
@@ -104,7 +116,7 @@ export class ExportCommand extends IronfishCommand {
       return
     }
 
-    if (color && !flags.language && !flags.mnemonic) {
+    if (color && flags.json) {
```
jowparks, line 119: I think this is broken, if you provide the `--json` flag it will try to output the `bech32m` output as json.

yarn.lock
```diff
@@ -4687,6 +4687,11 @@ base64-js@^1.0.2, base64-js@^1.3.1:
   resolved "https://registry.yarnpkg.com/base64-js/-/base64-js-1.5.1.tgz#1b1b440160a5bf7ad40b650f095963481903930a"
   integrity sha512-AKpaYlHn8t4SVbOHCy+b5+KKgvR4vrsD8vbvrbiQJps7fKDTkjkDry6ji0rUJjC0kzbNePLwzxq8iypo41qeWA==
```
jowparks, line 4689: woot 😃

ironfish/src/rpc/routes/wallet/exportAccount.ts
```diff
@@ -13,6 +13,7 @@ export type ExportAccountResponse = {
...
+    version: number
   }
```
jowparks, line 17: You can reset all the files for the `version` changes since they are all in different PR now

ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -84,10 +98,23 @@ export class ImportCommand extends IronfishCommand {
 
     process.stdin.off('data', onData)
 
-    return JSONUtils.parse<AccountImport>(data)
+    const convertedData = ImportCommand.bech32ToJSON(data)
+    return JSONUtils.parse<AccountImport>(convertedData !== null ? convertedData : data)
```
jowparks, line 102: Can be shortened to
`ImportCommand.bech32ToJSON(data) ?? data`

jowparks: Also it is totally possible that this JSON decode fails as well, probably want to throw a meaningful error message if that happens

ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -84,10 +96,23 @@ export class ImportCommand extends IronfishCommand {
...
+    const userInput = await CliUx.ux.prompt('Paste the output of wallet:export', {
+      required: true,
+    })
+    try {
+      const retData = JSONUtils.parse<AccountImport>(
+        ImportCommand.bech32ToJSON(userInput) ?? userInput,
+      )
+      return retData
+    } catch (e) {
+      CliUx.ux.info('Unable to decode the account, requesting account details individually')
```
jowparks, line 113: I think it would be better if the message said
'Failed to decode the account from the provided input, please continue with the manual input below'
also we should probably swap to `CliUx.ux.error`

EvanJRichard: Changing the message, agreed. I like `error` better, except that `error` exits, (or I don't see in the docs how to make `error` not exit).

EvanJRichard: And by docs I guess I mean reading the `oclif/core` code

EvanJRichard: Nevermind figured it out 😎

jowparks (approved):

PR #3217, opened 2023-01-31
trusted-setup: Add estimated time remaining to ceremony command

dguenther:
## Summary

The command previously showed how many people were in line in front of you, but it would
be helpful to have an ETA. I added an estimate of 10 minutes per contributor, but this could
go longer since the server verification time isn't factored into the estimate.

I could add a few minutes for the server estimate, but I don't think people will tend to take
the entire 10 minutes to contribute, so I'm hoping it balances out.

## Breaking Change

```
[ ] Yes
```


danield9tqh:
I'm thinking it might be easier to just pass back an expected time from the server side so we can include the variable timeouts. Would be slightly more accurate than hard coding 10 mins

dguenther:
Superceded by #3227

PR #3218, opened 2023-01-31
feat: import mnemonic from words (1/2)

jowparks:
## Summary
PR 1/2

Adds support rust/rust-nodejs for getting spending key string from words.
## Testing Plan
Unittests
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust-nodejs/tests/demo.test.slow.ts
+++ b/ironfish-rust-nodejs/tests/demo.test.slow.ts
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-import { Asset, DECRYPTED_NOTE_LENGTH, LanguageCode, wordsSpendingKey } from '..'
+import { Asset, DECRYPTED_NOTE_LENGTH, LanguageCode, spendingKeyToWords, wordsToSpendingKey } from '..'
 import {
   initializeSapling,
   generateKey,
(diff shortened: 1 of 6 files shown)


EvanJRichard (approved):
LGTM :ship:

ironfish-rust/src/keys/mod.rs
```diff
@@ -197,12 +185,22 @@ impl SaplingKey {
...
     /// will need a direct mapping. The private key could still be generated
     /// using bip-32 and bip-39 if desired.
-    pub fn words_spending_key(&self, language: Language) -> Result<String, IronfishError> {
+    pub fn to_words(&self, language: Language) -> Result<String, IronfishError> {
         let mnemonic = Mnemonic::from_entropy(&self.spending_key, language)
             .map_err(|_| IronfishError::InvalidEntropy)?;
         Ok(mnemonic.phrase().to_string())
     }
 
+    /// takes words and turns them into spending key
```
mat-if, line 194:
```suggestion
    /// Takes words and turns them into spending key
```

As a side note, it's not immediately clear what 'words' are, as someone new to this code. The doc-string here (and any other related functions) could be more descriptive about that, even something simple as "Takes a bip-39 phrase as a string..." would help

ironfish-rust/src/keys/test.rs
```diff
@@ -115,13 +115,26 @@ fn test_hex_conversion() {
 }
 
 #[test]
-fn test_words_spending_key() {
+fn test_to_words() {
     let key = SaplingKey::generate_key();
```
mat-if, line 119: If you wanted to make these tests more deterministic, you could do something like this

```rust
let byte_array = [213, 107, ... ]; // copied from below
SaplingKey::new(byte_array)
assert_eq(words, key.spending_key)
```

ironfish-rust-nodejs/tests/demo.test.slow.ts
```diff
@@ -29,10 +29,16 @@ describe('Demonstrate the Sapling API', () => {
 
   it('Should return phrase words from key input', () => {
     const key = generateKey()
```
mat-if, line 31: Same thing with this test, you could make it deterministic if you wanted with `generateKeyFromPrivateKey` which takes a hex string. It might be worth writing tests here and in the rust to do the "full flow" of starting at 1 point, say, a key, going to words, then back to the key and asserting its the same as well. I guess if you have both tests using deterministic inputs its essentially the same thing.

ironfish-rust/src/keys/mod.rs
```diff
@@ -197,12 +185,22 @@ impl SaplingKey {
...
+    pub fn from_words(words: String, language: Language) -> Result<Self, IronfishError> {
+        let mnemonic = Mnemonic::from_phrase(&words, language)
+            .map_err(|_| IronfishError::InvalidPaymentAddress)?;
```
mat-if, line 197: This would be easier to use if we implemented the From error type wrapper on IronfishError the way we do for other error types, and just wrap the error. Given this already existed, just pointing it out, not asking for it to be changed. Then you're able to skip the `map_err`, since `?` implicitly calls `.into()` if i remember correctly

mat-if (approved):
Looks good. Few minor comments, dealers choice

ironfish-rust/src/keys/mod.rs
jowparks: Added it as a follow up (along with returning Mnemonic rather than string in rust layer)

PR #3220, opened 2023-01-31
trusted-setup: Add target beacon randomness value

dguenther:
## Summary

Picks a future randomness source that will be used to seal the parameters after public contributions are finished.

https://drand.love/ was chosen as a source of randomness, since it has the advantage of being both a distributed source of entropy and verifiable.

## Breaking Change

```
[ ] Yes
```


ironfish-mpc/README.md
```diff
@@ -2,6 +2,10 @@
 
 Much of the code in this folder was originally forked from https://github.com/zcash-hackworks/sapling-mpc. The original licenses and copyright are retained in this folder.
 
+## Beacon
+
+Our final contribution will be seeded using the randomness generated from [The League of Entropy's drand network](https://drand.love/) in round #2759370.
```
danield9tqh, line 7: Can we move it back a couple days. I calculated 2768010 is about March 10

danield9tqh (approved):

PR #3222, opened 2023-01-31
Exported accounts have a version

EvanJRichard:
## Summary
This PR proposes to add a `version` member to accounts, in case encoding/decoding changes in the future. Currently, that version is always `1`.

## Testing Plan
Unit tests.

## Breaking Change

```
[ ] Yes
[X] No
```


jowparks:
Changes look reasonable, CI is failing though. You might also want to run this by #engineering to make sure we are ok with encoding the version in the json like this.

NullSoldier (approved):
Will there be an encoding version too?

EvanJRichard:
@NullSoldier I should add import version, too, good point. Will do.
@jowparks CI fixing adventures in progress in #engineering.

PR #3228, opened 2023-01-31
calculates available balance

hughy:
## Summary

the available or 'spendable' balance for an account is equal to the unconfirmed balance minus any outputs received in unconfirmed transactions and any inputs spent in pending transactions.

the available balance is useful for determining whether an account is able to create a new transaction that spends an asset.

calculates the available balance for an asset by iterating over unconfirmed transactions and subtracting output amounts from the unconfirmed balance then iterating over pending transactions and subtracting input amounts.

combines the logic for calculating confirmed, pending, and available balances into a single function, 'calculateBalances', to avoid iterating over transactions multiple times

adds available balance to RPCs

updates getBalance and getBalances to include available balance and updates RPC test

## Breaking Change

```
[ ] Yes
```


dguenther (approved):
Let me see if I have this right:

### Terms

`n`: Head of the chain
`c`: Confirmation sequence, <= `n`

### Calculations

* Unconfirmed: outputs at `n` - inputs at `n`
* Confirmed: Unconfirmed - outputs from `c`..`n` - inputs from `c`..`n`
* Pending: Unconfirmed + outputs from mempool - inputs from mempool
* Available: Unconfirmed - outputs from `c`..`n` - inputs from mempool

### Definitions

* Unconfirmed: Amount you have on-chain at `n`
* Confirmed: Amount you have on-chain at `c`
* Pending: Amount you have on-chain at `n`, as well as your transactions that are not yet on-chain
* Available: Amount you have on-chain at `c`, minus the value of notes that are spent between `c` and `n` and notes that are spent in transactions that are not yet on-chain

### Summary

With the availability check, are there edge cases around if you have a separate account with a different confirmation window? For example, if I have note1 with value 500 and note2 with value 500, and a confirmation range of 5. If, on a separate node with confirmation range 0, I create 2 transactions sending 1 each, spending note1 and receiving note1a with 499, then spending note1a and receiving note1b with 498, would Available be 500 + 500 - 500 - 499?

I think the Available balance is overall more clear for users who want to spend notes vs the other two, so this looks good to me!

hughy:
> For example, if I have note1 with value 500 and note2 with value 500, and a confirmation range of 5. If, on a separate node with confirmation range 0, I create 2 transactions sending 1 each, spending note1 and receiving note1a with 499, then spending note1a and receiving note1b with 498, would Available be 500 + 500 - 500 - 499?

Hmm this is a good example. The _correct_ available balance with `c = 5` should be 500: note2 is confirmed and unspent.

This could be a problem just on one node using different confirmation ranges for different RPC calls. I'm thinking that it's possible that the only way to calculate the available balance independently of confirmation range might be to sum the values of all of the notes that are confirmed and unspent.

dguenther:
Yeah, I'm leaning in that direction as well. I'm not sure if it's possible to calculate the Available balances without edge cases unless you're aware of the notes that are being spent, or at least a dependency graph of, for example, which notes are being spent in a transaction, or which block the notes originate from.

I think this is a step in the right direction in terms of user experience, but I wonder if we might need to change the data models up a bit to be able to dynamically calculate these balances correctly. @NullSoldier mentioned wanting to review this as well

hughy:
Closing this PR since we'll need to change to approach and data modeling

PR #3229, opened 2023-02-01
chore: remove codecov cli

jowparks:
## Summary
We are not writing tests currently for `ironfish-cli` so this is conflating our actual code coverage reports. Removing this, we can revert this change if we start writing tests again for cli.
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -32,10 +32,6 @@ flags:
     paths:
       - ironfish-rust
     carryforward: true
-  ironfish-cli:
-    paths:
-      - ironfish-cli
-    carryforward: true
   ironfish:
     paths:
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -1,5 +1,6 @@
 ignore:
   - config/*
+  - ironfish-cli/*
```
NullSoldier, line 3: don't you need to remove it below too? Or does this work enough?

NullSoldier (approved):

codecov.yml
jowparks: Yea lets try that first, not sure if there will be a caching error

Issue #3219, opened 2023-01-31
Transactions timestamps after rescan are wrong


Ryushka:
### What happened?

After the wallet:rescan --reset procedure, timestamps of transactions in wallet:balances --all  are just the "timestamps" when rescan function rescaned this specific transactions...

![image](https://user-images.githubusercontent.com/113109172/215847832-8b29274f-af8e-4cb9-8596-7868959ba2db.png)

### Version

0.1.63

### Debug output

```shell
Iron Fish version               0.1.63 @ src
Iron Fish library               0.0.40 @ src
Operating system                Linux x64
CPU model(s)                    Intel(R) Core(TM) i7-10700K CPU @ 3.80GHz
CPU threads                     16
RAM total                       15.52 GiB
Heap total                      3.99 GiB
Node version                    v18.13.0
ironfish in PATH                true
Garbage Collector Exposed       true
Telemetry enabled               true
Node name                       PaRUSaI
Block graffiti                  PaRUSaI
```

### Graffiti

PaRUSaI

PR #3230, opened 2023-02-01
sets connected transaction timestamp to block header timestamp

hughy:
## Summary

the wallet currently sets the timestamp field of a record in the transactions datastore to the time that the wallet first synced that transaction.

however, if the wallet rescans transactions, then these timestamps will all be set to a time during the rescan, and the relative time of transaction is lost.

instead we can set the timestamp of the transaction equal to the timestamp in the block header when that transaction is connected to the chain.

reported in #3219

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/account.ts
```diff
@@ -193,7 +191,7 @@ export class Account {
           blockHash,
           sequence,
           submittedSequence,
-          timestamp,
+          timestamp: blockHeader.timestamp,
```
NullSoldier, line 194: This means that transactions will move in your `wallet:transactions` list, is that ok? Do we intend that behavior?

NullSoldier (approved):

PR #3236, opened 2023-02-01
Update post with spending key

ygao76:
## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/wallet/postTransaction.test.ts
```diff
@@ -2,23 +2,29 @@
...
 import { createRawTransaction } from '../../../testUtilities/helpers/transaction'
 import { createRouteTest } from '../../../testUtilities/routeTest'
+import { Account } from '../../../wallet/account'
 
 describe('Route wallet/postTransaction', () => {
   const routeTest = createRouteTest(true)
+  let account: Account
+
+  beforeAll(async () => {
+    account = await useAccountFixture(routeTest.node.wallet, 'existingAccount')
```
NullSoldier, line 16: It's preferable not to use fixtures in beforeAll, these fixtures get generated for every test weather they use them or not when more tests are added in the future. You can just use it in each test.

ironfish/src/rpc/routes/wallet/postTransaction.ts
```diff
@@ -27,9 +32,21 @@ router.register<typeof PostTransactionRequestSchema, PostTransactionResponse>(
   `${ApiNamespace.wallet}/postTransaction`,
   PostTransactionRequestSchema,
   async (request, node): Promise<void> => {
+    const data = request.data
+
+    const account = node.wallet.getAccountByName(data.sender)
+
+    if (!account) {
+      throw new ValidationError(`No account found with name ${data.sender}`)
+    }
```
NullSoldier, line 41: We already have `getAccount()` RPC util that does this that most RPC uses

ironfish/src/rpc/routes/wallet/postTransaction.ts
```diff
@@ -3,16 +3,21 @@
...
+  transaction: string
+}
 
 export type PostTransactionResponse = {
   transaction: string
 }
 
 export const PostTransactionRequestSchema: yup.ObjectSchema<PostTransactionRequest> = yup
   .object({
+    sender: yup.string().defined(),
```
NullSoldier, line 20: This should be optional to match the default account pattern of other RPCs

NullSoldier (approved):
Looks good, there's just a few small things.

rohanjadvani (approved):

PR #3246, opened 2023-02-01
Swap wordsToSpendingKey for spendingKeyToWords

dguenther:
## Summary

This was a misnamed function causing lints on staging to fail after merging #3218 and #3207 separately. Also, since this is exporting, I think it should be taking a spending key and outputting words, rather than taking words and outputting a spending key.

## Breaking Change

```
[ ] Yes
```


holahula (approved):

EvanJRichard (approved):
LGTM :shipit:

Issue #3234, opened 2023-02-01
"ironfish  chain:asset" get error


weekendfootballhappy:
### What happened?

"ironfish  chain:asset" get error
![image](https://user-images.githubusercontent.com/124109740/215924863-1c52ef68-dcb9-4a62-ba8f-e98c82165fa6.png)

### Version

0.1.63 @ 693ff0f

### Graffiti
weekend.football

NullSoldier:
Fixed by https://github.com/iron-fish/ironfish/pull/3247

PR #3247, opened 2023-02-01
Fix chain:asset args name

wd021:
## Summary
Fixing bug https://github.com/iron-fish/ironfish/issues/3234. Args name should be `id`.

## Testing Plan
Run `chain:asset` and check that bug is gone

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):
good find, thanks

PR #3248, opened 2023-02-01
feat(ironfish): Enrich assets from chain DB

rohanjadvani:
## Summary

* Check if the Wallet DB has matching records for the given Asset ID when decrypting notes
* If not, fetch them from the chain (this will be migrated in the future to RPC)
* Null out block hash and sequence during disconnect

## Testing Plan

Unit tests

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/account.ts
```diff
@@ -204,6 +204,40 @@ export class Account {
...
+      metadata,
+      name,
+      owner,
+    }: {
+      createdTransactionHash: Buffer
+      metadata: Buffer
+      id: Buffer
+      name: Buffer
+      owner: Buffer
+    },
```
hughy, line 220: This looks a little odd; can we pass each field in as a typed argument? Or use the type returned by `chain.getAssetById`?

ironfish/src/wallet/wallet.ts
```diff
@@ -411,6 +411,8 @@ export class Wallet {
 
           assetBalanceDeltas.update(transactionDeltas)
 
+          await this.upsertAssetsFromDecryptedNotes(account, decryptedNotes, tx)
```
hughy, line 414: Should we do this for pending transactions too?

rohanjadvani: Ah yeah, forgot to paste that in. Updated

rohanjadvani: Actually - I think this is ok. We only need to do this for assets available in the chain. Pending mints/burns should have that available before it hits this point. If something isn't confirmed, then we shouldn't need it from the chain.

hughy: I'm thinking about a pending transaction where one account, _A_, is sending some of a custom asset to another account, _B_, for the first time.

Would _B_ already have that asset in its wallet DB?

rohanjadvani: Fixed! Thanks for pointing this out

hughy (approved):

PR #3255, opened 2023-02-02
add name to trusted setup submissions

danield9tqh:
## Summary
For later usage we want to give the user the option to add a name to their contribution. This is in case we want to publicly display the contributions later on.

## Testing Plan
Local testing

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/ceremony.ts
+++ b/ironfish-cli/src/commands/ceremony.ts
@@ -11,7 +11,6 @@ import { pipeline } from 'stream/promises'
 import { IronfishCommand } from '../command'
 import { DataDirFlag, DataDirFlagKey, VerboseFlag, VerboseFlagKey } from '../flags'
 import { CeremonyClient } from '../trusted-setup/client'
-import { nameIsValid } from '../trusted-setup/schema'
 
 export default class Ceremony extends IronfishCommand {
   static description = 'Contribute randomness to the Iron Fish trusted setup'
@@ -52,15 +51,10 @@ export default class Ceremony extends IronfishCommand {
     )
     randomness = randomness.length ? randomness : null
 
-    let name = 'anonymous'
-    do {
-      // Prompt for graffiti / name
-      const nameInput = await CliUx.ux.prompt(
-        `\nIf you'd like to associate a name with this contribution, type it here, then press Enter. Name must only include letters, numbers and spaces and be less than 30 characters. To accept the default, just press Enter (anonymous)`,
-        { required: false },
-      )
-      name = nameInput.length ? nameInput.trim() : name
-    } while (!nameIsValid(name))
+    const name = await CliUx.ux.prompt(
+      `If you'd like to associate a name with this contribution, type it here, then press Enter. Otherwise, to contribute anonymously, just press Enter`,
+      { required: false },
+    )
 
     // Create the client and bind events
     const client = new CeremonyClient({
--- a/ironfish-cli/src/commands/ceremony.ts
+++ b/ironfish-cli/src/commands/ceremony.ts
@@ -11,6 +11,7 @@ import { pipeline } from 'stream/promises'
 import { IronfishCommand } from '../command'
 import { DataDirFlag, DataDirFlagKey, VerboseFlag, VerboseFlagKey } from '../flags'
 import { CeremonyClient } from '../trusted-setup/client'
+import { nameIsValid } from '../trusted-setup/schema'
 
 export default class Ceremony extends IronfishCommand {
   static description = 'Contribute randomness to the Iron Fish trusted setup'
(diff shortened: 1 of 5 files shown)


ironfish-cli/src/commands/ceremony.ts
```diff
@@ -46,11 +47,21 @@ export default class Ceremony extends IronfishCommand {
...
+      `If you'd like to contribute your own randomness to the ceremony, type it here, then press Enter. For more information on where this should come from and its importance, please read https://setup.ironfish.network. If you'd like the command to generate some randomness for you, just press Enter`,
       { required: false },
     )
     randomness = randomness.length ? randomness : null
 
+    let name = 'anonymous'
+    do {
+      // Prompt for graffiti / name
+      const nameInput = await CliUx.ux.prompt(
+        `\nIf you'd like to associate a name with this contribution, type it here, then press Enter. Name must only include letters, numbers and spaces and be less than 30 characters. To accept the default, just press Enter (anonymous)`,
```
ygao76, line 59: nit:  it might be better to phrase like `To accept the default as anonymous, just press Enter`

ironfish-cli/src/trusted-setup/server.ts
```diff
@@ -101,7 +101,7 @@ export class CeremonyServer {
     const paramFileNames = await S3Utils.getBucketObjects(this.s3Client, this.s3Bucket)
     const validParams = paramFileNames
       .slice(0)
-      .filter((fileName) => /^params_\d{4}$/.exec(fileName)?.length === 1)
+      .filter((fileName) => /^params_\d{5}.*$/.test(fileName))
```
ygao76, line 104: is the exec -> test intentional ?

ygao76 (approved):
Looks good, left a few small comments

ironfish-cli/src/trusted-setup/schema.ts
```diff
@@ -27,11 +27,26 @@ export type CeremonyServerMessage =
...
+const nameRegexes = [/^[ a-zA-Z0-9]+$/, /^[^ ].*/, /.*[^ ]$/]
+
 export const CeremonyClientMessageSchema: yup.ObjectSchema<CeremonyClientMessage> = yup
   .object({
-    method: yup.string().oneOf(['contribution-complete', 'upload-complete']).required(),
+    method: yup.string().oneOf(['contribution-complete', 'upload-complete', 'join']).required(),
+    name: yup
+      .string()
+      .max(30)
+      .matches(nameRegexes[0]) // all characters must be alphanumeric or space
```
lwisne, line 43: dashes?

ironfish-cli/src/trusted-setup/server.ts
```diff
@@ -329,7 +345,8 @@ export class CeremonyServer {
     const hash = await verifyTransform(oldParamsDownloadPath, newParamsDownloadPath)
 
     client.logger.info(`Uploading verified contribution`)
-    const destFile = 'params_' + nextParamNumber.toString().padStart(4, '0')
+    const nameSuffix = client.name === undefined ? '' : `_${client.name.split(' ').join('_')}`
+    const destFile = 'params_' + nextParamNumber.toString().padStart(5, '0') + nameSuffix
```
lwisne, line 349: Paired with the comment above, if you remove restrictions from inputted text, you probably want to not use that text for file system naming. I would suggest just always using the anon path. Is there a value add in having the user name in the file name?

ironfish-cli/src/commands/ceremony.ts
dguenther: Maybe "To contribute anonymously, just press Enter" or "Otherwise, to contribute anonymously, just press Enter". I do agree it's a little bit confusing what the default is

ironfish-cli/src/trusted-setup/server.ts
dguenther: Yeah, [Regexp.test](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test) returns true/false, so is more convenient for filtering than [Regexp.exec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec). They're both searching the string for a match.

ironfish-cli/src/trusted-setup/server.ts
dguenther: We don't have any database beyond S3 for the trusted setup process (or rather, we're treating the S3 key-value store as our database). So the most immediately obvious place to store the contributor's name is in the filename, which is limited to support for these characters: https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html

Other options I can think of:

- Add the contributor to the file's metadata. It's a bit more inconvenient to retrieve, but S3 apparently allows arbitrary metadata values on files: https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html
- Create a separate file with metadata for the contribution. Might be a bit more straightforward than using metadata since we already have support for uploading files, but does raise concurrency issues (if upload of params succeeds but metadata fails, or vice versa, how do we handle that error state?)

Generally though, I think this is pretty low priority. If one of the above are easy to implement, feel free, but I don't think this should block. E.g. [Aleo](https://setup.aleo.org/) and [ZCash](https://github.com/zcash-hackworks/sapling-mpc/wiki) didn't support complex unicode characters

danield9tqh: Yea there is no value in having the name in the file name. This was more a matter of just finding a way to store a name in general like Derek mentioned. This wasn't in our original spec and isn't necessary per se. But I was thinking that if we did want to have some display of contributors later on, it wouldn't be possible to retroactively get peoples' names. Really just a simple name would do the job of identification. I understand the concern that we may be making the overall experience slightly worse by asking them to enter a name using a very restrictive character set. I'm going to look into S3 metadata quickly to see if that would be easy to implement

ironfish-cli/src/commands/ceremony.ts
danield9tqh: Yea good suggestions

dguenther (approved):

svarichevskiy:
why error ? 
version 0.1.67

root@ubuntu-s-4vcpu-8gb-blr1-01:~# ironfish ceremony
    ParseJsonError: Parsing internal.json Failed
    Unexpected end of JSON input while parsing empty string in internal.json

PR #3256, opened 2023-02-02
Raw transaction flag for wallet:send and wallet:mint

ygao76:
## Testing Plan
locally run `wallet:send --rawTransaction`
and `wallet:mint --rawTransaction`
## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/wallet.ts
```diff
@@ -690,6 +690,10 @@ export class Wallet {
       },
     )
 
+    if (options.rawTransaction === true) {
+      return raw
+    }
```
hughy, line 695: I was expecting that `mint` and `burn` would follow the same logical flow as `send` with create, post, and broadcast as separate steps and separate RPC calls.

Should we do mints differently?

PR #3257, opened 2023-02-02
fixes little-endian encodings in walletDb

hughy:
## Summary

we store sequences in several walletDb datastores: sequenceToNoteHash, sequenceToTransactionHash, and pendingTransactionHashes

we use little-endian encodings in the keys for each of these stores. the problem is that we rely on iterating over the keys of these stores *in order* so that we can load keys using sequence ranges. we cannot use little-endian encodings for this kind of ordering; we need to use big-endian encoding instead.

- adds UInt32EncodingBE for big-endian encoding of Uint32 values
- replaces datastores using little-endian encodings with new ones using big-endian encodings
- adds data migration to replace the old stores with the new stores
- adds tests of WalletDB methods that rely on ordered iteration

## Testing Plan

- Added unit tests, ran on old datastores to confirm the issue
- Ran migration on local node, tested balance computations with different confirmation values

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/migrations/data/017-sequence-encoding.ts
+++ b/ironfish/src/migrations/data/017-sequence-encoding.ts
@@ -1,0 +1,158 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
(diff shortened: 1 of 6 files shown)


NullSoldier (approved):
Thanks for fixing this. @mirayadav pointed out this issue but I forgot to create a task for it.

PR #3265, opened 2023-02-02
test(ironfish): introduce some test scaffolding for the mining pool

mat-if:
## Summary

This introduces some basic scaffolding for the poolDatabase and MiningPoolShares classes. While this change doesn't introduce any meaningful unit tests, it will enable testing with the refactor PRs, so that we can be more confident in the changes being made.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/mining/poolDatabase/database.test.ts
+++ b/ironfish/src/mining/poolDatabase/database.test.ts
@@ -1,0 +1,43 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { Config } from '../../fileStores'
+import { NodeFileProvider } from '../../fileSystems'
+import { createRootLogger } from '../../logger'
+import { getUniqueTestDataDir } from '../../testUtilities/utils'
+import { PoolDatabase } from './database'
+
+describe('poolDatabase', () => {
+  let db: PoolDatabase
+
+  beforeEach(async () => {
+    const logger = createRootLogger().withTag('test')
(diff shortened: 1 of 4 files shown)


ironfish/src/mining/poolDatabase/database.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    const dataDir = getUniqueTestDataDir()
+    const fileSystem = new NodeFileProvider()
+    await fileSystem.init()
+    // TODO(mat): It would be convenient if we didn't need a filesystem for Config for tests
+    const config = new Config(fileSystem, dataDir)
+
+    db = await PoolDatabase.init({
+      config,
+      logger,
+      dbPath: ':memory:',
```
hughy, line 25: 🤯

ironfish/src/mining/poolDatabase/database.test.ts
```diff
@@ -0,0 +1,43 @@
...
+import { createRootLogger } from '../../logger'
+import { getUniqueTestDataDir } from '../../testUtilities/utils'
+import { PoolDatabase } from './database'
+
+describe('poolDatabase', () => {
+  let db: PoolDatabase
+
+  beforeEach(async () => {
+    const logger = createRootLogger().withTag('test')
+    const dataDir = getUniqueTestDataDir()
```
hughy, line 16: Do we need to do any cleanup on these?

hughy (approved):

ironfish/src/mining/poolDatabase/database.test.ts
mat-if: Based on a bit of poking around, I don't think we do any cleanup on test data on a test-run. Instead, it cleans it up on your _next_ test run. So you always have 1 run of data laying around. It'd be worth cleaning that up in another PR at some point

PR #3266, opened 2023-02-02
feat(ironfish): Add RPC function to get network hash power

holahula:
## Summary
Added functionality for getting the network hashes per second processing power. 

## Testing Plan
- Added unit tests for the new route

- Sanity check against the testnet with `[lookup=120, height=0]` returns `77253067370.02484` ~= `77 GH`

## Breaking Change

```
[ ] Yes
[X] No
```


ironfish/src/rpc/routes/chain/getNetworkHashPower.ts
```diff
@@ -0,0 +1,96 @@
...
+  `${ApiNamespace.chain}/getNetworkHashPower`,
+  GetNetworkHashPowerRequestSchema,
+  async (request, node): Promise<void> => {
+    // default values for lookup and height
+    let lookup = 120
+    let height = -1
+
+    if (request.data?.lookup) {
+      lookup = request.data.lookup
+    }
```
hughy, line 42: nit - In a lot of places in our codebase you'll see null coalescence for things like this:

```
let lookup = request.data?.lookup ?? 120
```

You don't need to make this change, just something you might see around the code!

ironfish/src/rpc/routes/chain/getNetworkHashPower.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    // Don't divide by 0
+    if (startTime === endTime) {
+      request.end({
+        hashesPerSecond: 0,
+      })
+      return
+    }
+
+    const workDifference = Number(endBlock.work - startBlock.work)
```
hughy, line 88: This might be ok depending on how big this difference can get, but I'm wary of coercing BigInt to Number because of loss of precision.

If we want to return a number with floating point precision we have `BigIntUtils.divide` for dividing two BigInts

holahula: Made the change to use `BigIntUtils.divide`

ironfish/src/rpc/routes/chain/getNetworkHashPower.ts
holahula: Made the change

hughy (approved):

PR #3267, opened 2023-02-02
trusted-setup: Handle client disconnects while server is verifying contributions

dguenther:
- Add a VERIFYING state to the trusted setup server
- WIP

## Summary

If a client disconnects while the server is verifying parameters, the server will continue with verifying + uploading the parameters, but will start the next client in the queue at the same time. We should effectively be locking around the process of verifying+uploading verified params to S3, and once a client has uploading their parameters, it's not too important whether they remain connected or not.

This PR adds an additional client state, VERIFYING, that is transitioned to once the client uploads their parameters. If the current contributor disconnects while in this state, the verification will continue.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/trusted-setup/server.ts
+++ b/ironfish-cli/src/trusted-setup/server.ts
@@ -11,11 +11,21 @@ import { v4 as uuid } from 'uuid'
 import { S3Utils } from '../utils'
 import { CeremonyClientMessageSchema, CeremonyServerMessage } from './schema'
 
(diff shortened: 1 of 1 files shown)


danield9tqh:
only thing I could think of for that is if a client sends a message out of order (or sends an additional message) then it will kick them and discard their contribution. I think that is ok though since something would probably be wrong with the client in that scenario

danield9tqh (approved):

PR #3270, opened 2023-02-02
updates wallet:notes table formatting

hughy:
## Summary

uses the same column formatting for 'Asset', 'Asset Name', and 'Asset ID' columns as wallet:transactions

moves logic for constructing asset column(s) to utils

replaces 'x' for unspent notes with '' in Spent column

reorders columns so that amount appears right after asset name

## Testing Plan

image

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/notes.ts
+++ b/ironfish-cli/src/commands/wallet/notes.ts
@@ -36,7 +36,7 @@ export class NotesCommand extends IronfishCommand {
 
     for await (const note of response.contentStream()) {
       CliUx.ux.table(
-        [{ ...note, assetName: BufferUtils.toHuman(Buffer.from(note.assetName, 'hex')) }],
+        [note],
         {
           memo: {
             header: 'Memo',
@@ -49,12 +49,6 @@ export class NotesCommand extends IronfishCommand {
           transactionHash: {
             header: 'From Transaction',
           },
-          ...TableCols.asset({ extended: flags.extended }),
-          value: {
-            header: 'Amount',
-            get: (row) => CurrencyUtils.renderIron(row.value),
-            minWidth: 16,
-          },
           isSpent: {
             header: 'Spent',
             get: (row) => {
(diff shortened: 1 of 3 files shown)


ironfish-cli/src/commands/wallet/notes.ts
```diff
@@ -34,19 +36,8 @@ export class NotesCommand extends IronfishCommand {
 
     for await (const note of response.contentStream()) {
       CliUx.ux.table(
-        [note],
+        [{ ...note, assetName: BufferUtils.toHuman(Buffer.from(note.assetName, 'hex')) }],
```
NullSoldier, line 39: Won't this already be done in tablecols Asset?

ironfish-cli/src/utils/table.ts
```diff
@@ -55,12 +59,52 @@ const timestamp = <T extends Record<string, unknown>>(options?: {
...
+      },
+    }
+  } else {
+    return {
+      asset: {
+        header: 'Asset',
+        get: (row) => {
+          Assert.isString(row.assetName)
+          Assert.isString(row.assetId)
+          const assetName = truncateCol(row.assetName, MIN_ASSET_NAME_COLUMN_WIDTH)
```
NullSoldier, line 91: we need to tohuman this don't we since it always comes in as hex16?

ironfish-cli/src/commands/wallet/notes.ts
```diff
@@ -58,18 +49,24 @@ export class NotesCommand extends IronfishCommand {
...
+            minWidth: 16,
+          },
           isSpent: {
             header: 'Spent',
             get: (row) => {
               if (row.spent === undefined) {
                 return '-'
               } else {
-                return row.spent ? `✔` : `x`
+                return row.spent ? `✔` : ``
```
NullSoldier, line 64: This gets pushed out because value is variable width. Should we just re-order this so it doesn't get shifted around?

ironfish-cli/src/commands/wallet/notes.ts
hughy: Updated!

ironfish-cli/src/commands/wallet/notes.ts
hughy: Reordered so that this appears before `Asset` and `Amount`

NullSoldier (approved):
Looks good, thanks!

PR #3275, opened 2023-02-02
feat(ironfish): mining pool tracks status of its submitted blocks

mat-if:
## Summary

**Builds on #3268** 

This introduces a new table to the mining pool database, block. The pool will insert a row every time it successfully submits a block. The event loop will check any unconfirmed blocks, and update their status accordingly. For now, that status includes whether it's on the main chain, and whether it's confirmed. This gives us the ability to track how much reward was earned within a specific timeframe.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/mining/poolDatabase/migrations/006-add-block-table.ts
+++ b/ironfish/src/mining/poolDatabase/migrations/006-add-block-table.ts
@@ -1,0 +1,30 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { Database } from 'sqlite'
+import { Migration } from '../migration'
+
+export default class Migration006 extends Migration {
+  name = '006-add-block-table'
+
+  async forward(db: Database): Promise<void> {
+    await db.run(`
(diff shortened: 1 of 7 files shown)


ironfish/src/mining/poolDatabase/migrations/006-add-block-table.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  async forward(db: Database): Promise<void> {
+    await db.run(`
+      CREATE TABLE block (
+        id INTEGER PRIMARY KEY,
+        createdAt INTEGER NOT NULL DEFAULT CURRENT_TIMESTAMP,
+        blockSequence INTEGER NOT NULL,
+        blockHash TEXT NOT NULL,
+        minerReward TEXT NOT NULL,
+        confirmed BOOLEAN DEFAULT FALSE,
```
hughy, line 19: This depends on the config value of `confirmations`. Would we run into problems if we materialize `confirmed` and later want to change the config value on the pool?

mat-if: I don't think it would adversely affect any of the logic, other than the fact that it would potentially not check some since they _were_ "confirmed" and thus wouldn't get checked anymore, even if you made the confirmation larger.

NullSoldier (approved):

PR #3279, opened 2023-02-03
Remove tree size

leanthebean:
## Summary
Removing tree size from the Spend Description as it appears to no longer be necessary for any real validation. 

## Testing Plan
Relying on existing tests as this was not really used anywhere. 

## Breaking Change
Looks like it's not as it wasn't even used in db serialization


ygao76 (commented):
Looks like the serialization need to updated on TS side.

leanthebean:
we still use tree size in verification (I thought we moved away from that), so closing

PR #3283, opened 2023-02-03
feat(ironfish): Store assets in the wallet database

rohanjadvani:
## Summary

This is a series of previously approved pull requests. This code change:
* Updates the assets store to make supply nullable and add block hash + sequence
  * No migration is needed to transform data since no writes previously occurred
  * Block hash and sequence are written for assets an account owns when the transaction is added on the main chain
  * Supply is set for assets the account owns
* Updates supply for main chain transactions
* Adds a migration to iterate through a wallet's transactions to backfill assets
* Upserts assets when decrypting notes in processing pending and connected transactions
  * Falls back on chain assets db

## Testing Plan

Unit tests

## Breaking Change

```
[ ] Yes
```


hughy (approved):

ironfish/src/wallet/walletdb/walletdb.ts
```diff
@@ -385,6 +386,19 @@ export class WalletDB {
     await this.nonChainNoteHashes.clear(tx, account.prefixRange)
   }
 
+  async *loadSequenceToTransactionHashes(
```
NullSoldier, line 389: We already have this added by hugh

ironfish/src/wallet/wallet.ts
```diff
@@ -428,6 +429,30 @@ export class Wallet {
...
+    decryptedNotes: DecryptedNote[],
+    tx?: IDatabaseTransaction,
+  ): Promise<void> {
+    for (const { serializedNote } of decryptedNotes) {
+      const note = new Note(serializedNote)
+      const asset = await this.walletDb.getAsset(account, note.assetId(), tx)
+
+      if (!asset) {
+        const chainAsset = await this.chain.getAssetById(note.assetId())
+        Assert.isNotNull(chainAsset, 'Asset must be non-null in the chain')
```
NullSoldier, line 443: I'm gonna write a future note here because we had some out of github conversation on this.

If this doesn't exist in the chain, it can crash. It's unclear how often this can happen so we're gonna ignore this for now. If it happens more, we'll need to figure out a way to save a pending asset, and then add the details later.

ironfish/src/wallet/account.ts
```diff
@@ -422,6 +621,16 @@ export class Account {
     return this.walletDb.loadTransactionsByTime(this, tx)
   }
 
+  async *getTransactionsOrderedBySequence(
```
NullSoldier, line 624: I think hugh added this if you rebase

ironfish/src/wallet/walletdb/walletdb.ts
rohanjadvani: Hugh's method takes a sequence range. I'll leave this as is

ironfish/src/wallet/account.ts
rohanjadvani: Renamed

NullSoldier (approved):

PR #3285, opened 2023-02-03
feat(ironfish): mining pool tracks status of transactions its created

mat-if:
## Summary

**Builds on #3275**

This introduces a new table which allows the pool to insert a row whenever it makes a new transaction. The event loop will update the status of any unconfirmed and unexpired transactions. This allows the pool to know what transactions it has made, as well as giving it the ability to re-pay out transactions in the event of one expiring.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/mining/pool.ts
+++ b/ironfish/src/mining/pool.ts
@@ -14,6 +14,7 @@ import { BigIntUtils } from '../utils/bigint'
 import { ErrorUtils } from '../utils/error'
 import { FileUtils } from '../utils/file'
 import { SetIntervalToken, SetTimeoutToken } from '../utils/types'
+import { TransactionStatus } from '../wallet'
 import { MiningPoolShares } from './poolShares'
 import { MiningStatusMessage } from './stratum/messages'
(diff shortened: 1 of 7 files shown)


ironfish/src/mining/pool.ts
```diff
@@ -541,4 +543,25 @@ export class MiningPool {
...
     }
   }
+
+  async updateUnconfirmedPayoutTransactions(): Promise<void> {
+    const unconfirmedTransactions = await this.shares.unconfirmedPayoutTransactions()
+
+    for (const transaction of unconfirmedTransactions) {
+      const transactionInfoResp = await this.rpc.getAccountTransaction({
+        hash: transaction.transactionHash,
+        confirmations: this.config.get('confirmations'),
```
NullSoldier, line 553: We rely on the confirmations of the node for blocks, but use the confirmation of the pool here. This isn't a big deal just pointing out should we just delete this and use the nodes confirmations config too?

NullSoldier (approved):

PR #3288, opened 2023-02-03
feat: detect language mnemonic export

jowparks:
## Summary
Detects language based on canonical env vars in shell for language codes
## Testing Plan

https://user-images.githubusercontent.com/26990067/216724561-8c00df9b-fe94-4a6c-aa1b-ecb77fa4bc57.mov

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/utils/language.ts
+++ b/ironfish-cli/src/utils/language.ts
@@ -16,10 +16,22 @@ export const LANGUAGES = {
   Spanish: LanguageCode.Spanish,
 } as const
 
+type LanguageCodeKey = keyof typeof LANGUAGE_CODES
 type LanguageKey = keyof typeof LANGUAGES
 
 export const LANGUAGE_KEYS = Object.keys(LANGUAGES) as Array<LanguageKey>
 export const LANGUAGE_VALUES = Object.values(LANGUAGES) as Array<LanguageCode>
+
+const LANGUAGE_CODES = {
+  en: LanguageCode.English,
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/utils/language.ts
```diff
@@ -34,3 +46,30 @@ export async function selectLanguage(): Promise<LanguageCode> {
...
+  }
+  if (languageId && languageId.slice(0, 5) === 'zh_TW') {
+    return LANGUAGE_CODES['zh_TW']
+  }
+  const languageCode = languageId?.split('_')[0]
+  if (languageCode && languageCode in LANGUAGE_CODES) {
+    return LANGUAGE_CODES[languageCode as LanguageCodeKey]
+  }
+  return null
+}
```
NullSoldier, line 67: I think we can change this to use the built in locale detection that dates use.

```
Intl.DateTimeFormat().resolvedOptions().locale
```

NullSoldier (approved):

xx-134:
Updated to 1.65.
There they changed the import export of the wallet,
before the wallet was restored when entering "spendingKey":
And now he demands a long phrase that I don't have.
How to recover having what is in the second picture?

EvanJRichard:
@xx-134 The workflow may have changed, but importing keys the way you describe is still supported. The easiest way to get help might be on the discord. 

(While trying to help you, I found a bug that might be impacting you, going to investigate and fix now.)

PR #3289, opened 2023-02-03
feat: move bech32 to SDK

jowparks:
## Summary
This moves the bech32 encode/decode scheme to the SDK for reuse
## Testing Plan
tests
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/export.ts
+++ b/ironfish-cli/src/commands/wallet/export.ts
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { spendingKeyToWords } from '@ironfish/rust-nodejs'
-import { ErrorUtils } from '@ironfish/sdk'
+import { Bech32m, ErrorUtils } from '@ironfish/sdk'
 import { CliUx, Flags } from '@oclif/core'
 import { bech32m } from 'bech32'
 import fs from 'fs'
@@ -66,7 +66,6 @@ export class ExportCommand extends IronfishCommand {
 
     const client = await this.sdk.connectRpc(local)
     const response = await client.exportAccount({ account })
-    const responseJSONString = JSON.stringify(response.content.account)
 
     let output
     if (flags.language) {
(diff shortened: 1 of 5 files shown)


NullSoldier:
What is this used for?

jowparks:
> Reference

I thought we would expose it so that exteneral devs would have access to this encode/decode strategy outside of TS. Where do you want it to live?

EvanJRichard:
My mind goes to some class or set of functions in `ironfish/src/[...some package, not sure which]`. `utils`, `serde`, and `common` come to mind, but I don't know as much about the sdk structure as the cli structure. 
Then the CLI (and in the future, devs working on top of ironfish) could import the class or functions from the sdk and use them.

ironfish/src/utils/bech32m.ts
```diff
@@ -0,0 +1,22 @@
...
+
+// 1023 is maximum character count where the checksum will still catch errors
+export function bech32mDecode(bech32: string): string | null {
+  try {
+    const decodedOutput = bech32m.decode(bech32, 1023)
+    const decodedWords = decodedOutput.words
+    const decodedBytes = bech32m.fromWords(decodedWords)
+    return Buffer.from(decodedBytes).toString()
+  } catch (e) {
+    return null
```
NullSoldier, line 14: You don't wan't to catch errors like this in JS, because errors are untyped. It won't surface decoding errors properly.

ironfish/src/utils/bech32m.ts
```diff
@@ -0,0 +1,22 @@
...
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { bech32m } from 'bech32'
+
+// 1023 is maximum character count where the checksum will still catch errors
+export function bech32mDecode(bech32: string): string | null {
+  try {
+    const decodedOutput = bech32m.decode(bech32, 1023)
+    const decodedWords = decodedOutput.words
+    const decodedBytes = bech32m.fromWords(decodedWords)
+    return Buffer.from(decodedBytes).toString()
```
NullSoldier, line 12: hmm, this encoding is not symetrical encoding

You can't encode and decode and get the same output.

ironfish-cli/src/commands/wallet/export.ts
```diff
@@ -88,13 +88,11 @@ export class ExportCommand extends IronfishCommand {
...
         : responseJSONString
     } else {
-      const responseBytes = Buffer.from(responseJSONString)
-      const lengthLimit = 1023
-      output = bech32m.encode(
-        'ironfishaccount00000',
-        bech32m.toWords(responseBytes),
-        lengthLimit,
-      )
+      const encoded = Bech32m.encode(responseJSONString, 'ironfishaccount00000')
```
NullSoldier, line 91: You can unpack this to be more clear

NullSoldier (approved):

PR #3291, opened 2023-02-04
Rename transaction output

ygao76:
## Testing Plan
existing unit test
## Breaking Change

```
[ ] Yes
```


jowparks (dismissed):
Looks good, seems like the TS is choking, probably need to update some calls to the napi bindings

PR #3307, opened 2023-02-05
split change into 2 notes in certain cases

lwisne:
## Summary
We have found in Phase 2 that users end up with one big note that they keep spending. This creates the impression that transactions are blocked while another tx is processing. That's annoying, and also a self-imposed problem.

An easy fix is to split change into multiple notes in certain cases. There are a lot of ways to do this, but one that is relatively simple is just to look at a tx and split the change if it's a relatively large proportion of the original tx. I chose >50% here.

Example: if you have a 10 $IRON note, and you're spending 1 $IRON and getting 9 $IRON back in change, that's a sign that it would nice if you had some smaller notes. So we give you two change bills - a 6 $IRON note and a 3 $IRON note.

This would repeat again if you did the same transaction: you now spend 1 $IRON via a 3 $IRON note and get 2 $IRON in change, in increments of 1.33333333 $IRON and 0.66666667 $IRON. 

Now, if you spend that 1 $IRON a third time you have a note that's close in size, and we only deliver one note change.

This process should lead to accounts having a nice diversity of different sized notes that, long term, will help keep spendable balances somewhere near their real value while transactions are pending.

## Breaking Change

```
[ ] Yes
```


dguenther:
I definitely see the value in helping users avoid single-note transaction gating. I think we could probably spend a little more time investing in UX improvements to determine whether this is an issue that could be solved with UX -- For example, giving users a view of which notes they're spending in a transaction, or allowing users to customize how they get change notes back.

I could see a case for trying this out, since rolling back would only impact users who learn to rely on this behavior. I think my main potential concern would be whether it raises more questions without UX to explain what's happening -- for example, if users are having transactions gated now, we can pretty concretely say "We don't provide an easy way to split up your funds." But with this change, it would take a bit more debugging to determine whether they were hitting the 1-change-note arm of the match or not.

So I think it would be helpful to have more visibility into notes getting spent/created as part of a transaction before making changes to the notes we create, to avoid making the change process harder to intuit.

danield9tqh:
@dguenther that was my first thought as well. Is this going to make it more complicated for us / users to debug transaction errors. But the more I thought about it I think I've come around to the idea that it will not make that much of a difference in debugging transactions. 
1. We already don't have good visibility into what notes a user is spending so we aren't currently using that information. I think we're already making the assumption that transactions can send or receives notes in any quantity or size
2. If we are making that assumption anywhere it seems like a good idea to stop doing that. Right now transactions are (presumably) only created by our wallet but after mainnet all bets are off. Anyone could be writing custom transaction creation and probably will. So I think it would be good to get used to that.

On that note though I do agree that we could have a config option for this. (not sure if a config just makes things even more complicated as now there are two possibilities).

Yea my intuition is that this could be a project in itself and the precise nature of how to split up change notes should be further investigated and pros/cons evaluated more formally. But this method of doing it actually seems pretty reasonable on first pass. I'd vote yes on it since it solved an immediate user need.

ironfish-rust/src/transaction/mod.rs
```diff
@@ -220,6 +223,28 @@ impl ProposedTransaction {
         self._partial_post()
     }
 
+    fn _change_note_amounts(&self, asset_id: &[u8; 32], change_amount: &i64) -> Vec<i64> {
```
danield9tqh, line 226: I do think it would make sense if we are going to do this to make it more parameterized either by passing in the threshold for splitting notes or passing in a list of change amounts that we want. That way the wallet can make the decision of how to create the transaction instead of the rust code doing it automatically

NullSoldier:
After thinking about this pretty carefully, I really do not think this should be done without user intervention. I have a feeling that in general a financial system should not do anything unexpected. I think that if we want to do this, it should either be

 - User initiated
 - User configured
 - Automatically done higher up by things like the Wallet GUI app once the user has been asked

Mainly because this produces some behaviour that may end up making transactions more expensive, or have unpredictable results. I could imagine this being done in a few ways

- Have various change calculating modes that are selectable when creating a transaction such as "one note", "or multiple notes"
- Add commands to manage subdivision of notes
- Have certain user simplification apps like the wallet app prompt the user to configure one time which type of change they want, or just pick the one that optimizes usability

lwisne:
> After thinking about this pretty carefully, I really do not think this should be done without user intervention. I have a feeling that in general a financial system should not do anything unexpected. I think that if we want to do this, it should either be
> 
> * User initiated
> * User configured
> * Automatically done higher up by things like the Wallet GUI app once the user has been asked
> 
> Mainly because this produces some behaviour that may end up making transactions more expensive, or have unpredictable results. I could imagine this being done in a few ways
> 
> * Have various change calculating modes that are selectable when creating a transaction such as "one note", "or multiple notes"
> * Add commands to manage subdivision of notes
> * Have certain user simplification apps like the wallet app prompt the user to configure one time which type of change they want, or just pick the one that optimizes usability

I agree with the sentiment that this doesn't belong here. Was more food for thought.

I also agree with a wallet implementing. Specifically, having some sort of setup wizard with these options and sane defaults. I would actually default it to some sort of intelligent note splitting, but that's a difference conversation.

My only ask is that we do _something_ on this front at some point in the near future. We have seen multiple users complain about this and/or be confused by it, and I personally have found it to be a very odd behavior in the system. I like to make attempts to address user feedback like this.

PR #3308, opened 2023-02-06
Add data buffer to trusted setup client

danield9tqh:
## Summary
Sometime messages from the server get broken up into multiple calls to `onData`. In these cases we want to buffer the message until we receive the delimiter character `\n` and only process once we have the full message

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/trusted-setup/client.ts
```diff
@@ -86,8 +86,26 @@ export class CeremonyClient {
...
+      return
+    }
+
+    for (const message of messages.slice(0, -1)) {
+      this.onMessage(message)
+    }
+
+    this.dataBuffer = messages[message.length - 1] || ''
+  }
+
```
NullSoldier, line 107: You can use hugh's patented MessageBuffer here

danield9tqh: Nice, added

NullSoldier (approved):

PR #3316, opened 2023-02-06
feat(ironfish): add new share concept to mining pool

mat-if:
## Summary

**Builds on #3285** 

This is functionally very similar to the existing shares. The only differences are that it is associated to a payout period and a payout transaction. This way, we know which payout period it belongs to, and which transaction it belongs to. This also allows us to make sure that a share is paid out again in the event of a payout transaction expiring.

The old logic/table will be removed in a follow-up commit that removes all of the old stuff in one change, to minimize disruption to the existing pool and make it easier to transition to the new logic.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/mining/poolDatabase/database.ts
+++ b/ironfish/src/mining/poolDatabase/database.ts
@@ -8,6 +8,8 @@ import { NodeFileProvider } from '../../fileSystems/nodeFileSystem'
 import { Logger } from '../../logger'
 import { Migrator } from './migrator'
 
+const MAX_ADDRESSES_PER_PAYOUT = 250
+
 export class PoolDatabase {
   private readonly db: Database
   private readonly config: Config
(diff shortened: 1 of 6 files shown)


ironfish/src/mining/poolDatabase/database.ts
```diff
@@ -217,8 +230,74 @@ export class PoolDatabase {
...
+        payoutPeriodId = ?
+        AND publicAddress IN (
+          SELECT publicAddress
+          FROM payoutShare
+          WHERE
+            payoutPeriodId = ?
+            AND payoutTransactionId IS NULL
+          GROUP BY publicAddress
+          LIMIT ?
+        )
```
NullSoldier, line 271: Are we gonna do this query then fetch the paid shares after this query runs later?

NullSoldier: Ok, we're gonna try to convert this to a WHERE IN query

NullSoldier (approved):

PR #3317, opened 2023-02-06
feat(ironfish): Add RPC route for `/wallet/getAssets`

rohanjadvani:
## Summary

Adds RPC route to stream assets the wallet has

## Testing Plan

Unit test

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/wallet/getAssets.ts
```diff
@@ -0,0 +1,66 @@
...
+      if (request.closed) {
+        break
+      }
+
+      request.stream({
+        createdTransactionHash: asset.createdTransactionHash.toString('hex'),
+        id: asset.id.toString('hex'),
+        metadata: asset.metadata.toString('hex'),
+        name: asset.name.toString('hex'),
+        owner: asset.owner.toString('hex') === account.publicAddress,
```
hughy, line 58: Is there a case when we'd want to return the owner's address?

The chain RPC route give the owner address, so there is still a way to get it, but it's indirect

hughy (approved):

ironfish/src/rpc/routes/wallet/getAssets.ts
```diff
@@ -0,0 +1,66 @@
...
+        pending: !asset.blockHash,
```
hughy, line 59: What do you think of using `status` with pending/confirmed/unconfirmed?

That might require accepting a 'confirmations' option in the request too 🤔

ironfish/src/rpc/routes/wallet/getAssets.ts
rohanjadvani: Good point. I'll keep that behavior for consistency

ironfish/src/rpc/routes/wallet/getAssets.ts
rohanjadvani: Yep! Added

hughy (approved):

PR #3318, opened 2023-02-06
feat(cli): Add `wallet:assets`

rohanjadvani:
## Summary

Add CLI command to fetch assets stream and render table

## Testing Plan

```
git/ironfish/ironfish-cli  feat/rohan/cli-wallet-assets ✗                                                                                                             23m ◒
▶ ironfish wallet:assets
Now using node v18.12.1 (npm v8.19.2)
yarn run v1.22.19
 Name             ID                                                               Metadata Status      Supply          Owner
 ──────────────── ──────────────────────────────────────────────────────────────── ──────── ─────────── ─────────────── ────────────────────────────────────────────────────────────────
 rohanjadvani     421e78f2fc229fa11f06303f85dfc0051aebb2f17bc6d859c5afeebdf4755130 asset    confirmed   241200000000    a68aff2b847615c20b2a1f5a81b15d2f41aa7f84bccba1b254569f298cb00d02
 backup           6d6a154113dfad9102a4139fdde1dfd67c6acc4d4203f104a2961a7c83d1f842          pending     NULL            d292bc32f58873ed327f6161a972f34a112242cbf7f8a20d5bce09f956cdaccf
✨  Done in 6.51s.

```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/wallet/assets.ts
```diff
@@ -0,0 +1,89 @@
...
+    let showHeader = true
+
+    for await (const asset of response.contentStream()) {
+      CliUx.ux.table(
+        [asset],
+        {
+          name: {
+            header: 'Name',
+            minWidth: assetNameWidth,
+            get: (row) => BufferUtils.toHuman(Buffer.from(row.name, 'hex')),
```
hughy, line 55: If the width is set to `MIN_ASSET_NAME_COLUMN_WIDTH` but we have an asset with a longer name then the column widths might change between rows.

We have `truncateCol` in `utils/table.ts`, but I'm thinking of defining a fixed-width column type to handle this

rohanjadvani: Yeah, makes sense. Updated to use that function

ironfish-cli/src/commands/wallet/assets.ts
```diff
@@ -0,0 +1,89 @@
...
+          },
+          id: {
+            header: 'ID',
+            minWidth: ASSET_ID_LENGTH + 1,
+          },
+          metadata: {
+            header: 'Metadata',
+            get: (row) => BufferUtils.toHuman(Buffer.from(row.metadata, 'hex')),
```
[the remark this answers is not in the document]
rohanjadvani: Yep updated

hughy (approved):

PR #3320, opened 2023-02-06
feat(ironfish): Return `mints` and `burns` in follow chain response

rohanjadvani:
## Summary

We use this RPC command when uploading blocks to the API. Upload mints and burns as well

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/chain/followChain.ts
+++ b/ironfish/src/rpc/routes/chain/followChain.ts
@@ -37,6 +37,17 @@ export type FollowChainStreamResponse = {
       fee: number
       notes: Array<{ commitment: string }>
       spends: Array<{ nullifier: string }>
+      mints: Array<{
+        id: string
+        metadata: string
+        name: string
+        owner: string
+        value: string
+      }>
+      burns: Array<{
+        id: string
+        value: string
+      }>
     }>
   }
 }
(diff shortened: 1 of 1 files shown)


mat-if (approved):

ironfish/src/rpc/routes/chain/followChain.ts
```diff
@@ -125,6 +159,17 @@ router.register<typeof FollowChainStreamRequestSchema, FollowChainStreamResponse
             spends: transaction.spends.map((spend) => ({
               nullifier: spend.nullifier.toString('hex'),
             })),
+            mints: transaction.mints.map((mint) => ({
+              id: mint.asset.id().toString('hex'),
+              metadata: BufferUtils.toHuman(mint.asset.metadata()),
+              name: BufferUtils.toHuman(mint.asset.name()),
```
NullSoldier, line 165: hex these and do this on the other side, we just ran into issues with this today.

NullSoldier (approved):

ironfish/src/rpc/routes/chain/followChain.ts
rohanjadvani: Leaving as is for consistency with block graffiti

PR #3321, opened 2023-02-06
feat: miners fee public address

jowparks:
## Summary
WIP, pending discussion here:
https://iron-fish.slack.com/archives/C02PF56HESK/p1675715166660979
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/workerPool/tasks/createMinersFee.ts
+++ b/ironfish/src/workerPool/tasks/createMinersFee.ts
@@ -1,7 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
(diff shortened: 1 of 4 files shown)


ironfish/src/workerPool/tasks/createMinersFee.ts
```diff
@@ -75,16 +75,14 @@ export class CreateMinersFeeTask extends WorkerTask {
...
-    const minerNote = new Note(
-      minerPublicAddress,
-      amount,
-      memo,
-      Asset.nativeId(),
-      minerPublicAddress,
-    )
+    const minerNote = new Note(publicAddress, amount, memo, Asset.nativeId(), publicAddress)
 
     const transaction = new Transaction(spendKey)
```
jowparks, line 87: Need a valid spend key to pass in here, but it doesn't need to be attached to any real account.

jowparks:
Looks like we actually cannot do this, it the miners key is used to sign the mining transaction, which will not create a valid proof (even though it isn't used to generate the MerkleNote)

PR #3322, opened 2023-02-06
feat(ironfish): new payout logic for mining pool

mat-if:
## Summary

**Builds on #3316** 

This introduces the core logic needed to create payouts with the new additions to the codebase. Specifically, it now knows which shares have been paid out, how much reward was earned in a payout period, whether its blocks and transactions are confirmed and whether transactions need to be undone in the database if they expire.

It lacks some of the polish of the existing payout logic, which will be moved over and tweaked as needed as we begin to remove the old logic.

## Breaking Change

```
[x] Yes
```

### DIFF
--- a/ironfish/src/mining/poolDatabase/database.ts
+++ b/ironfish/src/mining/poolDatabase/database.ts
@@ -9,6 +9,7 @@ import { NodeFileProvider } from '../../fileSystems/nodeFileSystem'
 import { Logger } from '../../logger'
 import { Migrator } from './migrator'
 
+const PREVIOUS_PAYOUT_PERIODS = 3
 const MAX_ADDRESSES_PER_PAYOUT = 250
 
 export class PoolDatabase {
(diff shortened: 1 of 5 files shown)


ironfish/src/mining/poolDatabase/database.ts
```diff
@@ -290,6 +291,86 @@ export class PoolDatabase {
...
+      [payoutPeriodId - 2]: BigInt(15), // 15% of payout period x-2
+      [payoutPeriodId - 3]: BigInt(10), // 10% of payout period x-3
+    }
+
+    // Safety check in case the associated const is changed
+    Assert.isEqual(
+      PREVIOUS_PAYOUT_PERIODS + 1,
+      Object.keys(percentAmount).length,
+      'Payout period percent amount needs to have a value for each period',
+    )
```
NullSoldier, line 340: love this

ironfish/src/mining/poolDatabase/database.ts
```diff
@@ -290,6 +291,86 @@ export class PoolDatabase {
...
+
+    let totalReward = BigInt(0)
+    for (const result of results) {
+      const reward = BigInt(result.reward || '0')
+      const amount = (reward * percentAmount[result.id]) / BigInt(100)
```
NullSoldier, line 345: Juste a note, but this will leave some ore dust left over after each transaction, but that's fine

ironfish/src/mining/poolShares.ts
```diff
@@ -282,15 +283,110 @@ export class MiningPoolShares {
...
+      return
+    }
+
+    const assetId = Asset.nativeId().toString('hex')
+    const outputs = []
+    for (const payout of payoutAddresses) {
+      const amount = amountPerShare * BigInt(payout.shareCount)
+      outputs.push({
+        publicAddress: payout.publicAddress,
+        amount: amount.toString(),
```
NullSoldier, line 348:
```suggestion
        amount: CurrencyUtils.encode(amount),
```

ironfish/src/mining/poolShares.ts
```diff
@@ -282,15 +283,110 @@ export class MiningPoolShares {
...
+        memo: `${this.poolName} payout`,
```
NullSoldier, line 349: Do we want to include the payout ID here for debugging?

NullSoldier (approved):

PR #3325, opened 2023-02-07
Change wallet import/export to mirror each other, and to expect blobs

EvanJRichard:
## Summary
This PR proposes to change the CLI `wallet:import` workflow to default to expecting a blob, such as: `ironfish wallet:import pasted_blob`, rather than what it currently expects, a path: `ironfish wallet:import /path/to/a/blob`.  Path can still be supplied with `--importPath`. If the user somehow supplies both a blob and a path, just the blob is used.

Mirroring this behavior, `wallet:export` now needs an explicit argument `--path` if you want to export to a path; it defaults to assuming you would like a blob output to console.

## Testing Plan

https://user-images.githubusercontent.com/5766842/217118129-15011ebe-d730-46d1-8fa2-c5d5fa4a22a4.mov

## Breaking Change
Not a breaking change, though users may have to update scripts if they use any, since the usage default changed.
```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/import.ts
+++ b/ironfish-cli/src/commands/wallet/import.ts
@@ -23,10 +23,10 @@ export class ImportCommand extends IronfishCommand {
 
   static args = [
     {
-      name: 'path',
+      name: 'blob',
       parse: (input: string): Promise<string> => Promise.resolve(input.trim()),
       required: false,
-      description: 'The path to import the account from',
+      description: 'The copy-pasted output of wallet:export',
     },
   ]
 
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -19,14 +19,19 @@ export class ImportCommand extends IronfishCommand {
       default: true,
       description: 'Rescan the blockchain once the account is imported',
     }),
```
jowparks, line 21: This no longer matches the signature of the `export` command. I think we should make the `path` a labeled argument on the other command as well (instead of a positional argument). 

Also I would name this `path` not `importPath`. I am guessing there is a `--importPath` flag currently.

Last do you think `p` is a good char choice? Honest question, I was checking around saw one or two other commands use that flag as path, but also saw some using it as port

EvanJRichard: Good ideas, implementing. As far as `p` goes, I personally prefer just using the long argument, so I am inclined to remove the `char` entirely to make it unambiguous.

jowparks (approved):

PR #3331, opened 2023-02-07
chore: package updates

jowparks:
## Summary
Updates dependencies for `eslint`, `jest`, and `lerna` to fix some vulnerabilities
## Testing Plan
unittests
## Breaking Change

```
[ ] Yes
```


jowparks:
dependencies of build system in `lerna` still uses outdated packages, this will not fix unfortunately.

PR #3335, opened 2023-02-07
feat: account spending key optional

jowparks:
## Summary
- This PR makes spending key optional. The reason for this change is to allow usage of `View Only` wallets, instead of only having `Spend` wallets.
- Creates a type for `SpendingAccount` that requires a spending key
- updates tests/fixtures relative to these types

Eventually (in subsequent PR) this will lead to two different types of exports:
SPEND WALLET
```
{
    "name": "throwaway",
    "spendingKey": "8b610cdb12f5b4d84034325b6bfbde84ca8e45b5ee826e7d2e734fe62e52cd5b",
    "incomingViewKey": "9007719c56e84730db86f1674f4e1a65041d3e6ca08cba1cf7db5326cb0ca807",
    "outgoingViewKey": "b1e6de94e6fbb13fe3a905e94b3c5df6f8f5f0ea7010b5fbcfbcd535e6da7788",
    "publicAddress": "445721a8e2ea387a2f9cd351e32d0b9b3eaec91e9895efe9efe92bc4d2a56928"
}
```
VIEW ONLY WALLET
```
{
    "name": "throwaway",
    "spendingKey": null,
    "incomingViewKey": "9007719c56e84730db86f1674f4e1a65041d3e6ca08cba1cf7db5326cb0ca807",
    "outgoingViewKey": "b1e6de94e6fbb13fe3a905e94b3c5df6f8f5f0ea7010b5fbcfbcd535e6da7788",
    "publicAddress": "445721a8e2ea387a2f9cd351e32d0b9b3eaec91e9895efe9efe92bc4d2a56928"
}
```
### Migration testing
Tested `v1.61.0`->`this branch`
```
fish start  --upgrade --datadir=~/.irfTestOptionalSpending
Running 021-add-version-to-accounts...
  Migrating account default
Running 022-add-view-key-account...
  Migrating account default
Running 023-wallet-optional-spending-key...
  Migrating account default
Successfully applied 8 migrations
```
Tested `staging`->`this branch`
```
fish start  --upgrade --datadir=~/.irfTestOptionalSpending
Applying 1 migrations:
  023-wallet-optional-spending-key

Running 023-wallet-optional-spending-key...
  Migrating account default
```

Tested mint/burn/send on branch:

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/walletdb/accountValue.ts
```diff
@@ -10,18 +10,19 @@ const KEY_LENGTH = 32
...
   id: string
   name: string
-  spendingKey: string
+  spendingKey?: string
   incomingViewKey: string
   outgoingViewKey: string
   publicAddress: string
 }
 
 export class AccountValueEncoding implements IDatabaseEncoding<AccountValue> {
```
jowparks, line 19: This creates a null placeholder value for the spending key in a serialized account. It was either this or have a variable size serialized account. I tried out both options and this looked cleaner/easier to reason about IMO.

ironfish/src/wallet/account.ts
```diff
@@ -29,7 +29,7 @@ export class Account {
   readonly id: string
   readonly displayName: string
   name: string
```
jowparks, line 31: This is the main change in the PR that caused cascading of all the other changes.

ironfish/src/testUtilities/types.ts
```diff
@@ -0,0 +1,4 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
```
jowparks, line 3: This type is used for test fixtures so that we can guarantee the spendingKey is present on the given account.

ironfish/src/rpc/routes/wallet/exportAccount.ts
```diff
@@ -9,7 +9,7 @@ export type ExportAccountRequest = { account?: string }
 export type ExportAccountResponse = {
   account: {
     name: string
```
jowparks, line 11: This is no longer a required value. We will need to handle this with the `export` command updates, included in spec.

ironfish/src/testUtilities/fixtures/account.ts
```diff
@@ -1,24 +1,35 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Assert } from '../../assert'
 import { Account, AccountValue, Wallet } from '../../wallet'
+import { WithRequired } from '../types'
 import { FixtureGenerate, useFixture } from './fixture'
```
jowparks, line 8: The account fixtures should always have a `spendingKey`, so we are representing that here so that we can make the type system accept that.

ironfish/src/workerPool/tasks/decryptNotes.ts
```diff
@@ -26,6 +26,7 @@ export interface DecryptedNote {
 }
 
 export class DecryptNotesRequest extends WorkerMessage {
```
jowparks, line 28: Another null value place holder for serialization/deserialization. Open to comments

ironfish/src/wallet/account.ts
```diff
@@ -583,6 +578,10 @@ export class Account {
 
     return assetBalanceDeltas
   }
+  spendingAccount(): SpendingAccount {
+    Assert.isTruthy(this.spendingKey, 'The provided account is not a spending account')
```
hughy, line 582: I don't think it makes a practical difference, but I was a bit surprised to see `isTruthy` instead of `isNotUndefined`

ironfish/src/workerPool/tasks/decryptNotes.ts
```diff
@@ -237,7 +238,7 @@ export class DecryptNotesTask extends WorkerTask {
           forSpender: false,
           hash: note.hash(),
           nullifier:
-            currentNoteIndex !== null
+            currentNoteIndex !== null && spendingKey
```
hughy, line 241: Is it possible to add a spendingKey to an account?

The account would need to rescan in order to get the `decryptedNotes` and `nullifierToNoteHash` stores filled correctly

ironfish/src/workerPool/tasks/decryptNotes.ts
```diff
@@ -26,6 +26,7 @@ export interface DecryptedNote {
 }
 
 export class DecryptNotesRequest extends WorkerMessage {
+  static nullValue = '0'.repeat(KEY_LENGTH)
```
hughy, line 29: nit - I'm wondering if we should define a `NULL_KEY` somewhere since this gets used in a couple of places?

ironfish/src/wallet/walletdb/accountValue.ts
hughy: This seems OK to me. We use bitflags in a lot of serialized values to handle optional fields. We could do the same here for consistency, but that would require data migration ⚖️

ironfish/src/wallet/account.ts
jowparks: Seems reasonable to me and is more idiomatic

ironfish/src/workerPool/tasks/decryptNotes.ts
jowparks: Currently it is not supported. I don't think it would be worth it to support, instead the user could just delete the account and reimport using spending key. That way all the existing machinery just takes care of the process.

ironfish/src/wallet/walletdb/accountValue.ts
jowparks: Yea I think I should follow the standard pattern of bit flags since that seems like the current pattern. I propose we ship it this way, then I can make migration/update. Otherwise this PR is gonna get even bigger. Wdyt?

ironfish/src/rpc/routes/mining/blockTemplateStream.test.slow.ts
```diff
@@ -70,7 +76,12 @@ describe('Block template stream', () => {
     )
 
     // Generate another block
```
jowparks, line 78: This was a bug

ironfish/src/wallet/walletdb/accountValue.ts
hughy: That sounds good to me 👍

hughy (approved):

ironfish/src/workerPool/tasks/decryptNotes.ts
jowparks: Will delete this in subsequent migration PR

ironfish/src/wallet/wallet.ts
```diff
@@ -672,7 +672,7 @@ export class Wallet {
 
   async send(
     memPool: MemPool,
-    sender: Account,
+    sender: SpendingAccount,
```
NullSoldier, line 675: I thought based on the original spec, we were just gonna take a normal account here and do the check in here? This kind of puts a code burden on the external caller to do this check everywhere which balloons out the places needed to do this check. It's ok but it's kind of an irritating SDK design when in reality we have good infrastructure to automatically catch those errors and the user knows what to do with them.

jowparks: Adding this to the typing system is more explicit IMO. Lets consider this `send` example

If someone writes new code and wants to call `send`, they could pass an account (without spending key) and compiler checks would say they are good. It wouldn't be until they wrote a test (which we can only hope would test both types of accounts) or ran the code that they would see there is a failure. We are more explicitly defining our signature and therefore the interface without the user needing to understand the intracicies of what happens via the `send` command. 

The send method does not need an `Account` passed, it _needs_ an account with a spending key.

ironfish/src/testUtilities/fixtures/account.ts
jowparks: Note that this is almost a must since all old tests assumed account had a spending key. 50-100 tests would require addition of assertions if I didn't make this change. I can do that, or we can use `SpendingAccount`s for our fixtures

ironfish/src/wallet/wallet.ts
```diff
@@ -690,8 +690,12 @@ export class Wallet {
       expiration: expiration ?? undefined,
       confirmations: confirmations ?? undefined,
     })
-
-    return this.post(raw, memPool, sender.spendingKey)
+    const { spendingKey } = sender
```
NullSoldier, line 693: Let's avoid destructuring to remove things from the models they belong to.

ironfish/src/wallet/walletdb/accountValue.ts
```diff
@@ -10,18 +10,19 @@ const KEY_LENGTH = 32
...
   name: string
-  spendingKey: string
+  spendingKey: string | null
   incomingViewKey: string
   outgoingViewKey: string
   publicAddress: string
 }
 
 export class AccountValueEncoding implements IDatabaseEncoding<AccountValue> {
+  nullValue = '0'.repeat(KEY_LENGTH)
```
NullSoldier, line 20: This should just write a boolean and not write the bits if its nul.

ironfish/src/wallet/walletdb/accountValue.ts
```diff
@@ -10,18 +10,19 @@ const KEY_LENGTH = 32
...
   serialize(value: AccountValue): Buffer {
     const bw = bufio.write(this.getSize(value))
     bw.writeVarString(value.id, 'utf8')
     bw.writeVarString(value.name, 'utf8')
-    bw.writeBytes(Buffer.from(value.spendingKey, 'hex'))
+    bw.writeBytes(Buffer.from(value.spendingKey || this.nullValue, 'hex'))
```
NullSoldier, line 25: This should just write a boolean and not write the bits if its nul.

ironfish/src/workerPool/tasks/decryptNotes.ts
```diff
@@ -12,7 +12,7 @@ export interface DecryptNoteOptions {
   serializedNote: Buffer
   incomingViewKey: string
   outgoingViewKey: string
-  spendingKey: string
+  spendingKey: string | null
```
NullSoldier, line 15: This should just be deleted from the task right? It doesn't look like it's used

ironfish/src/workerPool/tasks/decryptNotes.ts
```diff
@@ -237,7 +239,7 @@ export class DecryptNotesTask extends WorkerTask {
           forSpender: false,
           hash: note.hash(),
           nullifier:
-            currentNoteIndex !== null
+            currentNoteIndex !== null && spendingKey
```
NullSoldier, line 242: Passing null here means the note being decrypted is not in the tree. Now if this is will all notes decrypted will be interpreted as not being on the tree. Do we need to fix this first in another PR? This seems like it could hint that we may be dead in the water with this. What's the implication to the wallet to not being able to create nullifiers to notes?

ironfish/src/wallet/wallet.ts
```diff
@@ -1291,15 +1303,17 @@ export class Wallet {
...
 
-    if (this.listAccounts().find((a) => toImport.spendingKey === a.spendingKey)) {
+    if (this.listAccounts().find((a) => spendingKey === a.spendingKey)) {
       throw new Error(`Account already exists with provided spending key`)
     }
-
-    const key = generateKeyFromPrivateKey(toImport.spendingKey)
+    // TODO(evan): upon adding multiple account import types, handle this error
+    Assert.isNotNull(spendingKey, 'Spending key is required for importing account')
+    const key = generateKeyFromPrivateKey(spendingKey)
```
NullSoldier, line 1316: I think this can all be reverted if we just use `toImport`

ironfish/src/workerPool/tasks/decryptNotes.ts
jowparks: The downstream of this is here:
https://github.com/iron-fish/ironfish/blob/ece1454b6b355146bd137edcf6dc0b55edd3af14/ironfish/src/wallet/walletdb/walletdb.ts#L574-L580
Particularly for the `nullifierToNoteHash`. 
I will look at where that table is utilized

jowparks: It seems like the main issue is this line:
https://github.com/iron-fish/ironfish/blob/1a051809dfd6f59d7178e174a5a79d98bf5fc756/ironfish/src/wallet/walletdb/walletdb.ts#L505-L510

We are getting note hashes by their nullifiers. I think you are right, the nullifierToNoteHash is used in a variety of very important methods for handling `spends`. The assumption is that the note is only a spend if the table has a corresponding `nullifierToNoteHash`. I will convene with @hughy and we can discuss plan of action.

ironfish/src/wallet/wallet.test.slow.ts
```diff
@@ -617,6 +619,96 @@ describe('Accounts', () => {
...
+      ...accountValue,
+      walletDb: node.wallet.walletDb,
+    })
+
+    await node.wallet.walletDb.db.transaction(async (tx) => {
+      await node.wallet.walletDb.setAccount(viewOnlyAccount, tx)
+      await node.wallet.skipRescan(viewOnlyAccount, tx)
+    })
+    // TODO(joe): remove this reference once `importAccount` supports view only wallet
+    // and change accounts back to protected
```
jowparks, line 649: Currently `wallet.importAccount` doesn't support view only wallets, so I hacked this test until we merge that PR.

ironfish/src/migrations/data/023-wallet-optional-spending-key.ts
```diff
@@ -0,0 +1,44 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
```
jowparks, line 3: review migration

ironfish/src/migrations/data/021-add-version-to-accounts/schemaOld.ts
```diff
@@ -98,7 +98,7 @@ export async function GetOldAccounts(
     accounts.push(
       new Account({
         ...account,
```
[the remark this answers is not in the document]
hughy: Yeah this is perfect since we backfilled with 1

ironfish/src/migrations/data/023-wallet-optional-spending-key.ts
hughy: It will be a race between this and #3515 to see which gets Migration023!

ironfish/src/genesis/genesis.test.slow.ts
```diff
@@ -63,7 +63,7 @@ describe('Create genesis block', () => {
     const amountBigint = BigInt(amountNumber)
 
     // Construct parameters for the genesis block
-    const account = await useAccountFixture(node.wallet, 'test')
```
NullSoldier, line 66: revert

ironfish/src/migrations/data/021-add-version-to-accounts/schemaOld.ts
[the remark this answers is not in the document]
NullSoldier: What is the type error? This arg looks like it should accept undefined.

ironfish/src/migrations/data/023-wallet-optional-spending-key.ts
NullSoldier: Looks good. We should test run it.,

ironfish/src/mining/manager.ts
```diff
@@ -133,6 +133,7 @@ export class MiningManager {
 
     const account = this.node.wallet.getDefaultAccount()
     Assert.isNotNull(account, 'Cannot mine without an account')
+    Assert.isNotNull(account.spendingKey, 'Account must have spending key in order to mine')
```
NullSoldier, line 136: You shouldn't need to, is this something we can fix after this PR?

ironfish/src/migrations/data/021-add-version-to-accounts/schemaOld.ts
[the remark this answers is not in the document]
jowparks: - backfill to add version is 1, should be one anyway
  - this function for `GetOldAccounts` doesn't even use version

ironfish/src/rpc/routes/wallet/postTransaction.ts
```diff
@@ -35,6 +36,10 @@ router.register<typeof PostTransactionRequestSchema, PostTransactionResponse>(
   PostTransactionRequestSchema,
   async (request, node): Promise<void> => {
     const account = getAccount(node, request.data.account)
+    Assert.isNotNull(
+      account.spendingKey,
+      'Account must have spending key to post a transaction',
+    )
```
NullSoldier, line 42: Can we just move this into node.wallet.post so it doesn't need to be done everywhere?

NullSoldier (approved):
Approved with some final comments

hughy (approved):
LGTM 🥇

ironfish/src/mining/manager.ts
jowparks: Are you sure? I thought a couple weeks ago I followed this down the call stack and saw in order to sign the transaction and generate the proofs you need a spend key.

jowparks:
Did some final testing, checked out v1.61.0, created account, went to staging, migrated account, went to this branch, migrated account. Everything works as expected

PR #3339, opened 2023-02-07
finishes renaming 'receives' -> 'outputs'

hughy:
## Summary

finds remaining uses of 'receives' to describe the outputs of a transaction and renames to 'outputs'

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/mining/poolShares.ts
+++ b/ironfish/src/mining/poolShares.ts
@@ -162,7 +162,7 @@ export class MiningPoolShares {
 
       const transaction = await this.rpc.sendTransaction({
         fromAccountName: this.accountName,
-        receives: transactionReceives,
+        outputs: transactionReceives,
         fee: transactionReceives.length.toString(),
       })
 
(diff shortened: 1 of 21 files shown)


ironfish/src/mining/poolShares.ts
```diff
@@ -173,7 +170,7 @@ export class MiningPoolShares {
         w.poolPayoutSuccess(
           payoutId,
           transaction.content.hash,
-          transactionReceives,
+          transactionOutputs,
           shareCounts.totalShares,
         ),
       )
```
NullSoldier, line 176: Mat is re-writing this right now, can you link him to this to make sure this won't impact him? He might be able to just do this on his branch more easily.

hughy: Chatted with Mat and reverted changes to `poolShares`

NullSoldier:
Still a few references in ironfish/src/testUtilities/helpers/transaction.ts

PR #3340, opened 2023-02-07
fix: autoimport spend key + clarifications

jowparks:
## Summary
- Adds attempt to import raw spending key
- Clarifies the account import when a failure occurs
- swaps the order of the prompt for failed import to disambiguate the requested "name"
## Testing Plan

https://user-images.githubusercontent.com/26990067/217382192-ca672514-865d-4541-8404-8c81bce7fecf.mov

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/import.ts
+++ b/ironfish-cli/src/commands/wallet/import.ts
@@ -1,7 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { wordsToSpendingKey } from '@ironfish/rust-nodejs'
+import { generateKeyFromPrivateKey, wordsToSpendingKey } from '@ironfish/rust-nodejs'
 import { AccountImport, Bech32m, JSONUtils, PromiseUtils } from '@ironfish/sdk'
 import { CliUx, Flags } from '@oclif/core'
 import { IronfishCommand } from '../../command'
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -71,8 +51,10 @@ export class ImportCommand extends IronfishCommand {
     }
 
     if (account === null) {
-      this.log('No account to import provided')
-      return this.exit(1)
+      CliUx.ux.warn(
+        'Failed to decode the account from the provided input, please continue with the manual input below',
+      )
+      account = await this.manualImport()
```
NullSoldier, line 57: I think this flow got a bit worse in the failure case. If I enter an invalid key, I get asked twice, even though the validation already happens in importTTY. I don't think there's any reasonable way to get to this code path, right?

```
ironfish wallet:import
Paste the output of wallet:export: b4174e1c50453ac1638915136ab9eaff0e7e028c427754ef2163858eb5491d2
 ›   Warning: Failed to decode the account from the provided input, please continue with the manual input below
Enter the account spending key: b4174e1c50453ac1638915136ab9eaff0e7e028c427754ef2163858eb5491d2
Enter a new account name: jason2
InvalidPaymentAddress
```

jowparks: Yea your are right, we should just delete this path and exit if this occurs

jowparks: Screen Shot 2023-02-07 at 4 22 26 PM

jowparks: Screen Shot 2023-02-07 at 4 49 21 PM

NullSoldier (approved):

PR #3341, opened 2023-02-07
defines reusable fixedWidth column

hughy:
## Summary

encapsulates logic for creating an oclif table column with fixed width. creates a column with minWidth set to the width and truncates the value returned by 'get' to the specified width

uses fixedWidth column in wallet:assets command

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/assets.ts
+++ b/ironfish-cli/src/commands/wallet/assets.ts
@@ -7,7 +7,7 @@ import {
   ASSET_NAME_LENGTH,
   PUBLIC_ADDRESS_LENGTH,
 } from '@ironfish/rust-nodejs'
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/commands/wallet/assets.ts
```diff
@@ -57,25 +57,20 @@ export class AssetsCommand extends IronfishCommand {
       CliUx.ux.table(
         [asset],
         {
-          name: {
+          name: TableCols.fixedWidth<GetAssetsResponse>({
```
NullSoldier, line 60: you don't need to specify this generic, it'll infer it properly.

NullSoldier (approved):

PR #3352, opened 2023-02-08
wallet.importAccount supports view-only versions

EvanJRichard:
## Summary
This changes `wallet.importAccount` to allow importing an account from provided view keys, rather than a spending key.

## Testing Plan
Added a unit test.

## Breaking Change

```
[ ] Yes
[X] No
```


ironfish/src/wallet/validator.ts
```diff
@@ -60,11 +60,7 @@ export function validateAccount(toImport: Partial<AccountValue>): void {
     throw new Error(`Provided incoming view key ${toImport.incomingViewKey} is invalid`)
   }
 
-  if (!toImport.spendingKey) {
-    throw new Error(`Imported account has no spending key`)
-  }
-
-  if (!isValidSpendingKey(toImport.spendingKey)) {
+  if (toImport.spendingKey && !isValidSpendingKey(toImport.spendingKey)) {
```
EvanJRichard, line 63: This change seems fine but I'm sweating that I'm weakening assertions in a wallet validator.

ironfish/src/wallet/wallet.ts
```diff
@@ -1292,18 +1292,29 @@ export class Wallet {
...
+        incomingViewKey: key.incoming_view_key,
+        outgoingViewKey: key.outgoing_view_key,
+        publicAddress: key.public_address,
+      }
+    } else { 
+      // if spending key is not provided, use the provided view keys
+      accountValue = {
+        ...toImport,
+        id: uuid(),
+        // TODO BREADCRUMB: how to get public address without a spending key? should accountValues be changed to not need it? should it be required to import?
```
EvanJRichard, line 1316: @jowparks Would love your opinion on this - `AccountValue`s need a `publicAddress` but it seems like this must be derived from the `spendingKey`. Do I have that right? If so - what do you think a good path forward is? Maybe exported accounts provide public address too so that it can then be imported later?

ironfish/src/wallet/account.ts
```diff
@@ -23,7 +23,12 @@ import { WalletDB } from './walletdb/walletdb'
 
 export const ACCOUNT_KEY_LENGTH = 32
```
jowparks, line 25: I think a better update to this would be something like
```
AccountImport = {name: string, spendingKey:string} | {name: string, incomingViewKey: string, outgoingViewKey: string} 
```

ironfish/src/wallet/wallet.ts
```diff
@@ -1292,18 +1293,34 @@ export class Wallet {
...
+        publicAddress: key.public_address,
+      }
+    } else if (toImport.incomingViewKey && toImport.outgoingViewKey) {
+      // if spending key is not provided, use the provided view keys
+      const publicAddress = generatePublicAddressFromIncomingViewKey(toImport.incomingViewKey)
+      accountValue = {
+        ...toImport,
+        id: uuid(),
+        incomingViewKey: toImport.incomingViewKey,
+        outgoingViewKey: toImport.outgoingViewKey,
```
jowparks, line 1321: these two lines are unnecessary

ironfish-rust-nodejs/index.d.ts
```diff
@@ -61,6 +61,7 @@ export function generateKey(): Key
 export function spendingKeyToWords(privateKey: string, languageCode: LanguageCode): string
 export function wordsToSpendingKey(words: string, languageCode: LanguageCode): string
 export function generateKeyFromPrivateKey(privateKey: string): Key
+export function generatePublicAddressFromIncomingViewKey(incomingViewKey: string): string
```
jowparks, line 64: This name is crazy long, lets call it `incomingViewKeyToPublicAddress`

jowparks (commented):
Generally looks good, just some small comments. Also lets write some tests for both `AccountImport` types

ironfish/src/wallet/wallet.ts
EvanJRichard: thought so too, but an error occurs without them:

```
Type '{ id: string; publicAddress: string; name: string; spendingKey?: string | undefined; incomingViewKey?: string | undefined; outgoingViewKey?: string | undefined; }' is not assignable to type 'AccountValue'.
  Types of property 'incomingViewKey' are incompatible.
    Type 'string | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'.ts(2322)
```

EvanJRichard: Actually, with the type change suggested elsewhere, no error B)

jowparks: Yea you can't do assertions of object properties and have the type system pick that up unfortunately, you have to unpack and do assertions on them. Either way this should be a moot point if we do the OR type rather than optional values

ironfish/src/wallet/wallet.test.slow.ts
```diff
@@ -4,6 +4,7 @@
 import { Asset, generateKey } from '@ironfish/rust-nodejs'
 import { Assert } from '../assert'
 import { Target } from '../primitives/target'
+import { Account } from './account'
```
EvanJRichard, line 7: In line 654 of this file, I get that
`    node.wallet.accounts.set(viewOnlyAccount.id, viewOnlyAccount)`
is trying to access inaccessible member `accounts`. Did I do a bad merge @jowparks?

jowparks: Hmm not sure, lets pair and we can post results back in PR here

EvanJRichard: This ended up being a temporary issue while the main PR gets tested. Going to ignore.

PR #3354, opened 2023-02-08
calculates available balance from unspent notes

hughy:
## Summary

uses the unspentNoteHashes store to calculate the available balance

the available balance is the amount of an asset that an account currently has available to spend in a new transaction. the available balance is equal to the sum of the values of all notes for the asset that are confirmed on the chain and have not been spent in any confirmed, unconfirmed, or pending transaction

adds 'loadUnspentNoteValues' to walletDb to iterate over the values of unspent notes for a given asset

calculates the available balance by using 'loadUnspentNoteValues' to sum the values of all unspent notes with confirmed sequences

Depends on #3349 

## Breaking Change

```
[ ] Yes
```


rohanjadvani (approved):
Your tests are great :ok_hand:

PR #3355, opened 2023-02-08
adds backfill for unspentNoteHashes

hughy:
## Summary

iterates over all notes for each account

if a note is on the chain and has not been spent the migration adds an entry to the unspentNoteHashes datastore

Depends on #3349 

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/storage/database/encoding.ts
+++ b/ironfish/src/storage/database/encoding.ts
@@ -172,7 +172,7 @@ export class BigIntLEEncoding implements IDatabaseEncoding<BigInt> {
   }
 }
 
-export class BigIntBEEncoding implements IDatabaseEncoding<BigInt> {
+export class BigU64BEEncoding implements IDatabaseEncoding<BigInt> {
   serialize(value: bigint): Buffer {
     const buffer = bufio.write(8)
     buffer.writeBigU64BE(value)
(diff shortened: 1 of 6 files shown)


ironfish/src/migrations/data/020-unspent-notes.ts
```diff
@@ -0,0 +1,57 @@
...
+      )
+    }
+
+    logger.info(`Indexing unspent notes for ${accounts.length} accounts`)
+
+    for (const account of accounts) {
+      let unspentNotes = 0
+
+      logger.info(` Indexing unspent notes for account ${account.name}`)
+      for await (const note of account.getNotes()) {
```
NullSoldier, line 41: were getting into a tricky place where joe is changing the account schema so that will make migrations like this fail that don't materialize the account encoding

ironfish/src/storage/database/encoding.ts
```diff
@@ -172,6 +172,19 @@ export class BigIntLEEncoding implements IDatabaseEncoding<BigInt> {
   }
 }
 
+export class BigIntBEEncoding implements IDatabaseEncoding<BigInt> {
```
NullSoldier, line 175: this specifically is U64, we should put that in the encoding

hughy: I'll rename this to `BigU64BEEncoding`

ironfish/src/migrations/data/020-unspent-notes.ts
hughy: Do you mean that this will fail specifically for migrations that change the account encoding, or for all migrations?

hughy: I now understand what you mean (e.g., #3487 )

rohanjadvani (approved):

PR #3356, opened 2023-02-08
Make the default pool host ipv6 compatible

NullSoldier:
## Summary

By default if you listen on 0.0.0.0 then the node net server won't listen on the ipv6 adapter. IF you listen on :: which is the ipv6 equivilent it will set the net server in multi adapter mode and listen on both ipv6 and ipv4. If you want ipv6 only there is an extra parameter.

Read more, https://github.com/nodejs/node/issues/9390

## Breaking Change

```
[ ] Yes
```


jowparks:
Just want to verify the expected behavior is auto-listen on ipv4
https://github.com/nodejs/node/issues/9390#issuecomment-280394892

jowparks (approved):

mat-if (approved):

PR #3359, opened 2023-02-09
adds available balance to cli output

hughy:
## Summary

passes available balance through RPC endpoints

includes the 'available' balance in all balance outputs from the 'wallet:balance' and 'wallet:balances' commands

displays 'Available Balance' instead of the confirmed balance when running 'wallet:balance'

## Testing Plan

output of `wallet:balance --all` and `wallet:balances --all` with a pending transaction sending `hughy-if`:

image

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/wallet/getBalances.ts
```diff
@@ -82,6 +84,7 @@ router.register<typeof GetBalancesRequestSchema, GetBalancesResponse>(
         unconfirmedCount: balance.unconfirmedCount,
         pending: CurrencyUtils.encode(balance.pending),
         pendingCount: balance.pendingCount,
+        available: balance.available.toString(),
```
rohanjadvani, line 87: Does this need to be encoded?

hughy: Yes, good catch!

rohanjadvani (approved):

PR #3367, opened 2023-02-10
feat: optional spending key migration

jowparks:
## Summary
Creates migration for making account serialization compatible with optional spending key.
Related to:
https://github.com/iron-fish/ironfish/pull/3335

## Testing Plan
upgrade
```
Applying 1 migrations:
  020-wallet-optional-spending-key

Running 020-wallet-optional-spending-key...
Gathering accounts for migration
Making wallet compatible with view only accounts

  Migrating account default
  Completed migration for account default
WebSocket server started at :::9033
```

REPL
```
> f = wallet.walletDb.loadAccounts()
Object [AsyncGenerator] {}
> for await (a of f) { console.log(a) }
{
  id: 'a8e8907d-bcc3-44a9-9eee-b624e42b0715',
  name: 'default',
  publicAddress: '511d9aff078202fe1a74036022b82e36f4f5f30726b218258015d780e67bdc20',
  incomingViewKey: 'ceb44a9cd2d53aa523d3d2302a81e70ac91923cade243c718e5bb4be3dfe6700',
  outgoingViewKey: '5c7fbb7d4fcede23d32c5eb08868938885c7824f925898d4d46b49a5280fe4ee',
  spendingKey: '61b30c86603903badb91eab58f4af338f9bd01552b3983b69fbc52d07719a38a'
}
```
## Breaking Change

```
[ ] Yes
```


ironfish/src/migrations/data/020-wallet-optional-spending-key.ts
```diff
@@ -0,0 +1,120 @@
...
+        await walletDb.setAccount(account, tx)
+        logger.info(`  Completed migration for account ${account.name}`)
+      }
+    })
+
+    await oldAccountStore.clear(tx)
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  async backward(): Promise<void> {}
```
jowparks, line 55: A reverse migration doesn't make sense here right? If we did a reverse the walletdb would then be incompatible with the current serialization code.

hughy (approved):
This looks good to me!

The backwards migration isn't strictly necessary: it mostly helps developers switching between branches, but might also be useful if users had to downgrade to a previous version

ironfish/src/migrations/data/020-wallet-optional-spending-key.ts
NullSoldier: usually it's for the purpose that the current code is busted, but yeah we don't need to implement one.

PR #3374, opened 2023-02-10
Remove extra query from selectAsset

NullSoldier:
## Summary

This also makes it so it returns the whole "asset" and not just the ID

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/utils/asset.ts
+++ b/ironfish-cli/src/utils/asset.ts
@@ -15,48 +15,63 @@ export async function selectAsset(
     showSingleAssetChoice: boolean
     confirmations?: number
   },
-): Promise<string | undefined> {
+): Promise<
+  | {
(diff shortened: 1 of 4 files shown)


hughy (approved):

ironfish-cli/src/utils/asset.ts
```diff
@@ -15,48 +15,63 @@ export async function selectAsset(
...
+      name: string
+    }
+  }>([
     {
-      name: 'assetId',
+      name: 'asset',
       message: `Select the asset you wish to ${options.action}`,
       type: 'list',
-      choices: assetOptions,
+      choices: choices,
```
rohanjadvani, line 72:
```suggestion
      choices,
```

rohanjadvani (approved):

PR #3378, opened 2023-02-10
fix(ironfish): Skip assertion when connecting burns

rohanjadvani:
## Summary

There is a bug in which the wallet crashing when processing burns for an asset it does not own. Since the wallet stores all assets it owns, we should not assert on the owner field when connecting burns.

```
/Users/rohan/git/ironfish/ironfish/src/assert.ts:38
      throw new Error(message || `Expected values to be equal: ${String(a)} vs ${String(b)}`)
            ^


Error: Existing asset owner should match public address
    at Function.isEqual (/Users/rohan/git/ironfish/ironfish/src/assert.ts:38:13)
    at Account.saveConnectedBurnsToAssetsStore (/Users/rohan/git/ironfish/ironfish/src/wallet/account.ts:324:14)
    at <anonymous> (/Users/rohan/git/ironfish/ironfish/src/wallet/account.ts:200:7)
    at LevelupDatabase.withTransaction (/Users/rohan/git/ironfish/ironfish/src/storage/database/database.ts:246:22)
    at Account.connectTransaction (/Users/rohan/git/ironfish/ironfish/src/wallet/account.ts:144:5)
    at <anonymous> (/Users/rohan/git/ironfish/ironfish/src/wallet/wallet.ts:415:37)
    at LevelupDatabase.withTransaction (/Users/rohan/git/ironfish/ironfish/src/storage/database/database.ts:246:22)
    at Wallet.connectBlock (/Users/rohan/git/ironfish/ironfish/src/wallet/wallet.ts:392:7)
    at <anonymous> (/Users/rohan/git/ironfish/ironfish/src/wallet/wallet.ts:137:7)
    at async Promise.all (index 0)
```

## Testing Plan

Unit tests

## Breaking Change

```
[ ] Yes
```


hughy (approved):

ironfish/src/wallet/account.test.ts
```diff
@@ -680,6 +681,42 @@ describe('Accounts', () => {
         supply: mintValue - burnValue,
       })
       expect(await accountB['walletDb'].getAsset(accountB, asset.id())).toBeUndefined()
+
+      // Send some of Account A coins to Account B
```
NullSoldier, line 685: Does this crash before the fix?

rohanjadvani: Yeah

PR #3411, opened 2023-02-13
feat: wallet using notes instead of nullifiers

jowparks:
## Summary
WIP
## Breaking Change

```
[ ] Yes
```


jowparks:
It is note possible to use only notes instead of nullifiers via testing

PR #3412, opened 2023-02-13
feat(ironfish): Introduce new `GetBlockHeaders` network message

mat-if:
## Summary

This message will replace `GetBlockHashes` as the message used for syncing in a follow-up change. It also includes a `skip` and `reverse` flag, for more flexible future usage.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/messages/getBlockHeaders.ts
+++ b/ironfish/src/network/messages/getBlockHeaders.ts
@@ -1,0 +1,93 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import bufio from 'bufio'
+import { BlockHeader } from '../../primitives'
+import { NetworkMessageType } from '../types'
+import { getBlockHeaderSize, readBlockHeader, writeBlockHeader } from '../utils/serializers'
+import { Direction, RpcNetworkMessage } from './rpcNetworkMessage'
+
+export class GetBlockHeadersRequest extends RpcNetworkMessage {
+  readonly start: number
+  readonly limit: number
+  readonly skip: number
+  readonly reverse: boolean
+
+  constructor(start: number, limit: number, skip: number, reverse: boolean, rpcId?: number) {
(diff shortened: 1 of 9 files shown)


ironfish/src/network/peerNetwork.ts
```diff
@@ -993,6 +1021,69 @@ export class PeerNetwork {
...
+
+    if (request.message.limit === 0) {
+      peer.punish(
+        BAN_SCORE.LOW,
+        `Peer sent GetBlockHeaders with limit of ${request.message.limit}`,
+      )
+      return new GetBlockHeadersResponse([], rpcId)
+    }
+
+    if (request.message.limit > MAX_REQUESTED_BLOCKS) {
```
dguenther, line 1038: This could be done in a separate PR, but I think we could consider creating a separate MAX_REQUESTED_HEADERS value and increasing it, since block headers are much smaller than blocks. I'd probably take a look around bitcoin and/or eth and see what they consider reasonable values.

ironfish/src/network/messages/getBlockHeaders.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export class GetBlockHeadersRequest extends RpcNetworkMessage {
+  readonly start: number
+  readonly limit: number
+  readonly skip: number
+  readonly reverse: boolean
+
+  constructor(start: number, limit: number, skip: number, reverse: boolean, rpcId?: number) {
+    super(NetworkMessageType.GetBlockHeadersRequest, Direction.Request, rpcId)
+    this.start = start
```
dguenther, line 19: This is my fault for not noticing earlier, but the GetBlockHeaders method in eth takes either a sequence or a blockHash for the start value: https://github.com/ethereum/devp2p/blob/master/caps/eth.md#getblockheaders-0x03

I'm not sure that we use both now, but could we add that? The main benefit I think would be if the chain forks and rolls back while you're fetching block headers, then the sequences may continue to match up but not connect with the previous fetched headers.

dguenther (approved):
That's all I have for suggestions. The tests are really great 👍 Since this isn't a breaking change, I don't mind if you'd prefer to address the comments in separate PRs, so I'll hit approve on it

PR #3417, opened 2023-02-14
add eligibility check to mint

lwisne:
## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/wallet/mint.ts
```diff
@@ -339,4 +342,26 @@ Find the transaction on https://explorer.ironfish.network/transaction/${transact
...
+    const user = await api.findUser({ graffiti: graffiti })
+    if (!user) {
+      this.log(`WARNING: Could not find a user with graffiti ${graffiti}`)
+    } else {
+      if (!user.is_verified) {
+        this.log(
+          `WARNING: No verified email on account for graffiti ${graffiti}. You need this email to claim testnet rewards. Visit https://ironfish.network and click "Log In" to verify.`,
+        )
+      }
+      if (user.node_uptime_score < 14) {
```
NullSoldier, line 358: Can we just encode this here so we can change this in the API if we need to?

https://github.com/iron-fish/ironfish-api/pull/1227/files

NullSoldier: What if we send back 

total_hours and total_hours_needed, or what ever you want to name it just under that principal. Once this goes out we can't really change it easily since people aggressively use older versions.

lwisne: done.

ironfish-cli/src/commands/wallet/mint.ts
```diff
@@ -339,4 +342,29 @@ Find the transaction on https://explorer.ironfish.network/transaction/${transact
       this.exit(2)
     }
   }
+
+  async doElegibilityCheck(): Promise<void> {
+    const api = new WebApi()
+    const graffiti = this.sdk.config.get('blockGraffiti')
```
NullSoldier, line 348: There is one issue with this, this will use the graffiti on the CLI and not neccessarily the node. Lots of users connect to a remote node in docker, or on another machine. It will tell them they aren't eligibile. We should pass in the client from right outside this call, and use the client to get the graffiti from 

```
const graffiti = (await node.getConfig({ name: 'blockGraffiti' })).content.blockGraffiti
```

just like like our testnet command here https://github.com/iron-fish/ironfish/blob/master/ironfish-cli/src/commands/testnet.ts#L121

ironfish-cli/src/commands/wallet/mint.ts
```diff
@@ -339,4 +342,29 @@ Find the transaction on https://explorer.ironfish.network/transaction/${transact
...
+    const user = await api.findUser({ graffiti: graffiti })
+    if (!user) {
+      this.log(`WARNING: Could not find a user with graffiti ${graffiti}`)
+    } else {
+      if (!user.verified) {
+        this.log(
+          `WARNING: No verified email on account for graffiti ${graffiti}. You need this email to claim testnet rewards. Visit https://ironfish.network and click "Log In" to verify.`,
```
NullSoldier, line 355: You can just link to the login directly, https://testnet.ironfish.network/login

ironfish-cli/src/commands/wallet/mint.ts
```diff
@@ -339,4 +342,29 @@ Find the transaction on https://explorer.ironfish.network/transaction/${transact
       this.exit(2)
     }
   }
+
+  async doElegibilityCheck(): Promise<void> {
+    const api = new WebApi()
+    const graffiti = this.sdk.config.get('blockGraffiti')
+    const user = await api.findUser({ graffiti: graffiti })
+    if (!user) {
+      this.log(`WARNING: Could not find a user with graffiti ${graffiti}`)
```
NullSoldier, line 351: If they have not set a graffiti yet it just prints

```bash
$ ironfish wallet:mint

WARNING: Could not find a user with graffiti
Do you want to create a new asset (Y/N)?:
```
It's a bit weird, should we avoid doing this or print a different message if they have no block graffiti?

ironfish-cli/src/commands/wallet/mint.ts
lwisne: Good to know, will change. I will note that graffiti is fetched this way in a number of different places, so there may be other existing bugs in this category.

lwisne:
Screen Shot 2023-02-16 at 7 57 42 PM

NullSoldier (approved):

PR #3418, opened 2023-02-14
feat(ironfish): GetBlockHeaders can take sequence or hash for start

mat-if:
## Summary

This allows us to use either a sequence or hash for the start parameter of a GetBlockHeaders request

## Testing Plan

Unit tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/messages/getBlockHeaders.ts
+++ b/ironfish/src/network/messages/getBlockHeaders.ts
@@ -31,13 +31,12 @@ export class GetBlockHeadersRequest extends RpcNetworkMessage {
   serialize(): Buffer {
     const bw = bufio.write(this.getSize())
 
-    const isBuffer = Buffer.isBuffer(this.start)
-    bw.writeU8(Number(isBuffer))
-
-    if (isBuffer) {
-      bw.writeHash(this.start as Buffer)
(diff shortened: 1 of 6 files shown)


ironfish/src/network/messages/getBlockHeaders.ts
```diff
@@ -24,7 +30,16 @@ export class GetBlockHeadersRequest extends RpcNetworkMessage {
...
-    bw.writeU32(this.start)
+
+    const isBuffer = Buffer.isBuffer(this.start)
+    bw.writeU8(Number(isBuffer))
+
+    if (isBuffer) {
+      bw.writeHash(this.start as Buffer)
+    } else {
+      bw.writeU32(this.start as number)
+    }
```
dguenther, line 41: Nitpicky, but would it work to do this? Main reason I'd ask is I prefer to avoid `as` casting where possible since it overrides the type system without any checks.

```suggestion
    if (Buffer.isBuffer(this.start)) {
      bw.writeU8(1)
      bw.writeHash(this.start)
    } else {
      bw.writeU8(0)
      bw.writeU32(this.start)
    }
```

dguenther (approved):
Left one comment, otherwise looks good!

PR #3423, opened 2023-02-14
feat: view key type wrapper (1/2)

jowparks:
## Summary
Creates a type wrapper in Rust. Working towards storing the viewKey (hex from this PR) in the wallet/account instead of the `incomingViewKey`
## Testing Plan
Tests pass
## Breaking Change

```
[ ] Yes
```


ironfish-rust/src/keys/mod.rs
```diff
@@ -210,23 +210,22 @@ impl SaplingKey {
...
-    pub fn view_keys(&self) -> ViewKeys {
-        ViewKeys {
-            incoming: self.incoming_view_key().clone(),
-            outgoing: self.outgoing_view_key().clone(),
-        }
-    }
-
-    /// Adapter to convert this key to a viewing key for use in sapling
+    /// -    /// Adapter to convert this key to a viewing key for use in sapling
     /// functions.
```
jowparks, line 216: We will delete this method in a subsequent PR and replace with `ViewKey`

ironfish-rust/src/keys/mod.rs
```diff
@@ -210,23 +210,22 @@ impl SaplingKey {
...
     pub(crate) fn sapling_viewing_key(&self) -> ViewingKey {
         ViewingKey {
             ak: self.authorizing_key,
             nk: self.nullifier_deriving_key,
         }
     }
```
jowparks, line 222: We will delete this method in a subsequent PR and replace with ViewKey

ironfish-rust/src/keys/view_keys.rs
```diff
@@ -92,6 +92,51 @@ impl IncomingViewKey {
...
+            Err(_) => Err(IronfishError::InvalidViewKey),
+            Ok(bytes) => {
+                if bytes.len() != 64 {
+                    Err(IronfishError::InvalidViewKey)
+                } else {
+                    let mut authorizing_key_bytes = [0; 32];
+                    let mut nullifier_deriving_key_bytes = [0; 32];
+                    authorizing_key_bytes.clone_from_slice(&bytes[..32]);
+                    nullifier_deriving_key_bytes.clone_from_slice(&bytes[32..]);
+                    let authorizing_key =
```
jowparks, line 114: Not stoked about these `unwrap`s, but the code will get way uglier if we try to deal with the underlying types of `CtOption`. I was thinking of extending the CtOption in our codebase with something like
```
impl From<CtOption<T>> for Result<T, IronfishError> {
...
```

But we don't include the `subtle` library dependency, so not sure if we can do that without explicitly adding the dependency.

ironfish-rust/src/keys/view_keys.rs
```diff
@@ -143,14 +195,6 @@ impl OutgoingViewKey {
     }
 }
```
jowparks, line 197: Unused struct, deleteing

ironfish-rust/src/keys/view_keys.rs
```diff
@@ -92,6 +92,58 @@ impl IncomingViewKey {
...
+#[derive(Clone)]
+pub struct ViewKey {
+    pub authorizing_key: jubjub::SubgroupPoint,
+    pub nullifier_deriving_key: jubjub::SubgroupPoint,
+}
+
+impl ViewKey {
+    /// Load a key from a string of hexadecimal digits
+    pub fn from_hex(value: &str) -> Result<Self, IronfishError> {
+        match hex_to_bytes(value) {
```
jowparks, line 104: I will put in subsequent PR to rename this error to `InvalidViewKey`, it is present in a few places and has incorrect tense.

dguenther (approved):
This generally looks good to me. I wonder if we could have simplified some of the renaming, but I don't have any objections here since `ViewKeys` isn't the best name, and we've tended to create our own structs vs extending the zcash ones (`ViewingKey`).

I'd like to wait for @mat-if 's review too before merging

ironfish-rust/src/keys/view_keys.rs
```diff
@@ -92,6 +92,58 @@ impl IncomingViewKey {
     }
 }
 
+#[derive(Clone)]
+pub struct ViewKey {
+    pub authorizing_key: jubjub::SubgroupPoint,
+    pub nullifier_deriving_key: jubjub::SubgroupPoint,
```
jowparks, line 98: It is possible for us to tack on `outgoing_view_key` here. It would also mean we could use a name like `FullViewKey` wonder if the wallet cleanliness is worth the added change? We could also make this change later, which probably is the safest option considering we are close to launch

ironfish-rust/src/keys/mod.rs
mat-if, line 216: Can you clean up this comment? Think it got clobbered when you deleted the other struct

mat-if (approved):
Looks good - just a few minor nits. 👍

ironfish-rust/src/keys/view_keys.rs
jowparks: Nice, ty for suggestion

ironfish-rust/src/keys/test.rs
```diff
@@ -134,3 +137,20 @@ fn test_from_and_to_words() {
         SaplingKey::from_words(words, bip39::Language::English).expect("key should be created");
     assert_eq!(key.spending_key, key_bytes);
 }
+
+#[test]
```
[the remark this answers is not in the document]
jowparks: I was wondering why the `keys` had a separate test file, I just shrugged and added the test there. SGTM, moving

ironfish-rust/src/keys/view_keys.rs
```diff
@@ -92,6 +92,58 @@ impl IncomingViewKey {
...
+                } else {
+                    let mut authorizing_key_bytes = [0; 32];
+                    let mut nullifier_deriving_key_bytes = [0; 32];
+                    authorizing_key_bytes.clone_from_slice(&bytes[..32]);
+                    nullifier_deriving_key_bytes.clone_from_slice(&bytes[32..]);
+
+                    let authorizing_key_option: Option<SubgroupPoint> =
+                        SubgroupPoint::from_bytes(&authorizing_key_bytes).into();
+                    let authorizing_key =
+                        authorizing_key_option.ok_or(IronfishError::InvalidAuthorizingKey)?;
```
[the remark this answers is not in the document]
jowparks: Oh nice, I like it. Thanks for the sugar, tastes goooood

ironfish-rust/src/keys/view_keys.rs
jowparks: Closing comment, as well will do this as a followup later

PR #3425, opened 2023-02-14
feat(ironfish): use GetBlockHeaders if peer has the appropriate version

mat-if:
## Summary

**Builds on #3418** 

This allows us to use the new GetBlockHeaders message without requiring
a minimum version bump, as syncing from a peer on an older version will
default to the previously expected behavior of using GetBlockHashes.

## Testing Plan

Local testing

## Breaking Change

```
[ ] Yes
```


dguenther (approved):
Tested with one node running 0.1.67 and one node running this branch to confirm GetBlockHashes was sent. Also tested with two nodes on this branch to confirm GetBlockHeaders was sent.

PR #3429, opened 2023-02-15
feat: SaplingKey.view_key update (PR 2)

jowparks:
## Summary
Updates the `SaplingKey` struct to contain the `ViewKey` struct instead of the `authorizing_key` and `nullifier_deriving_key` components.
Lots of reference changes from `SaplingKey.authorizing_key` to `SaplingKey.view_key.authorizing_key`.

Part 1 here https://github.com/iron-fish/ironfish/pull/3423
Part 3 here https://github.com/iron-fish/ironfish/pull/3433
Part 4 here https://github.com/iron-fish/ironfish/pull/3436
## Testing Plan
tests pass
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/keys/mod.rs
+++ b/ironfish-rust/src/keys/mod.rs
@@ -13,7 +13,7 @@ use group::GroupEncoding;
 use ironfish_zkp::constants::{
     CRH_IVK_PERSONALIZATION, PROOF_GENERATION_KEY_GENERATOR, SPENDING_KEY_GENERATOR,
 };
-use ironfish_zkp::{ProofGenerationKey, ViewingKey};
+use ironfish_zkp::ProofGenerationKey;
 use jubjub::SubgroupPoint;
 use rand::prelude::*;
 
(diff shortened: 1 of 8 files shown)


ironfish-rust/src/keys/mod.rs
```diff
@@ -210,28 +207,11 @@ impl SaplingKey {
...
-    /// Retrieve the sapling_representation of the ViewKey. These would normally used for third-party audits
-    /// or for light clients.
-    /// Adapter to convert this key to a viewing key for use in sapling
-    /// functions.
-    pub(crate) fn sapling_viewing_key(&self) -> ViewingKey {
-        ViewingKey {
-            ak: self.authorizing_key,
-            nk: self.nullifier_deriving_key,
-        }
-    }
```
jowparks, line 222: Is there some reason we want to use getter pattern here, or can I just delete and access the struct directly? I noticed that the `outgoing` and `incoming` view keys have the pattern of getter method, but wondering what is idiomatic in rust

ironfish-rust/src/note.rs
```diff
@@ -298,7 +298,7 @@ impl<'a> Note {
                 .hash_length(32)
                 .personal(PRF_NF_PERSONALIZATION)
                 .to_state()
-                .update(&private_key.sapling_viewing_key().nk.to_bytes())
+                .update(&private_key.view_key.nullifier_deriving_key.to_bytes())
```
jowparks, line 301: One step closer to this using `view_key`

ironfish-rust/src/keys/mod.rs
jowparks: I am just gonna leave this method for now since I don't want to diverge from the cannon of how the fields on the `SaplingKey` are approached, but I think this is a useful discussion.

mat-if: This sums up my feelings about getters: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b78b2a1de22a603e84625616b49e3a33

If there's not a scenario where we would expect re-assignment, why even allow it? Accidents can happen. Yeah, we get some safety with mutable/immutable vars, but it's a lot easier to toss `mut` on something than to have to go change a whole function signature. In that vein, I want to go through and remove most of our `pub` keywords one of these days

mat-if: In this case, I'd say we're running into two slightly different patterns. Most of these previously were just functions to return a subset of the SaplingKey data via a more-specific struct. I'm not sure off the top of my head why IVK/OVK diverge from that pattern, and whether that makes sense or not. I'll defer to you on that for now

mat-if (approved):

PR #3431, opened 2023-02-15
feat(ironfish): improve the limits of GetBlockHeaders

mat-if:
## Summary

This introduces a limit of 1024 headers that can be requested per GetBlockHeaders request. A block header is currently ~180 bytes, which would make the max message size around 185kb. This also introduces a limit to the number of total block headers that can be looked up from disk, given that GetBlockHeaders can use a `skip` amount to potentially use an excessive amount of disk usage.

## Testing Plan

Unit tests

## Breaking Change

```
[ ] Yes
```


ironfish/src/network/peerNetwork.ts
```diff
@@ -1058,6 +1060,9 @@ export class PeerNetwork {
     const headers = []
     let skipCounter = skip
 
+    // Limit the total number of lookups to avoid excessive disk usage
+    let remainingLookups = MAX_REQUESTED_HEADERS * 2
```
mat-if, line 1064: Not sure if it made sense to make this a const in version.ts since its pretty specific to this logic right now, open to suggestions on this

ironfish/src/network/peerNetwork.ts
```diff
@@ -1066,6 +1071,11 @@ export class PeerNetwork {
       : (from: BlockHeader) => this.chain.iterateTo(from)
 
     for await (const header of iterationFunction(from)) {
+      remainingLookups -= 1
+      if (remainingLookups === 0) {
```
dguenther, line 1075: I think there's an off-by-one here -- e.g. if remainingLookups starts at 1, it'll immediately decrement to 0 and break. Might be worth a test, if it's easy to write one.

mat-if: Good call - I had been avoiding trying to figure out how to test the max given it was a const in another file, but decided to just figure it out (tldr: not hard, just mockModule)

ironfish/src/network/peerNetwork.test.ts
```diff
@@ -534,6 +542,109 @@ describe('PeerNetwork', () => {
...
+        const peerIdentity = peer.getIdentityOrThrow()
+
+        const sendSpy = jest.spyOn(peer, 'send')
+
+        const rpcId = 432
+        const message = new GetBlockHeadersRequest(block2.header.sequence, 4, 7, false, rpcId)
+        // The start header, block2, is the first lookup. Then we skip 7,
+        // resulting in 8 lookups. Thus, the header from block10 is not included
+        // since it would be the 9th lookup.
+        const response = new GetBlockHeadersResponse([block2.header], rpcId)
```
mat-if, line 637: I'm not the biggest fan of these really fiddly tests that require comments to clearly understand what it is testing, but not sure of a better alternative

ironfish/src/network/peerNetwork.ts
dguenther: I think I'd put in in `version.ts` as well, it might make it a bit more clear that adjusting `MAX_REQUESTED_HEADERS` will affect the max lookups if the two numbers are adjacent in `version.ts`

ironfish/src/network/peerNetwork.ts
```diff
@@ -1066,6 +1071,11 @@ export class PeerNetwork {
       : (from: BlockHeader) => this.chain.iterateTo(from)
 
     for await (const header of iterationFunction(from)) {
+      if (remainingLookups === 0) {
```
dguenther, line 1074: I think this technically does an extra lookup as well -- like if remainingLookups is 1, it'll fetch a header, decrement to 0, fetch another header, see that remainingLookups is 0, and break. I don't really mind in this case though because it accomplishes the goal of limiting the returned blocks correctly, so I'd be okay looking at this in a followup

ironfish/src/network/peerNetwork.test.ts
dguenther: The only change I'd want to make is maybe move the mocking of MAX_REQUESTED_HEADERS closer to the actual test. I guess also it's tricky because it mocks MAX_REQUESTED_HEADERS, but it's actually testing that maxLookups is maxheaders * 2, whereas maybe a better test would set a value for maxLookups then test that you don't go over it.

dguenther (approved):

PR #3439, opened 2023-02-16
Add route and CLI command for showing transactions from mempool

danield9tqh:
## Summary
Useful to be able to see what kind of transaction fees and expiration sequences are in the mempool. We can use this both to debug large user mempools and to help debug future mempool changes in upcoming projects.

Usage
```bash
ironfish mempool:transactions                       # returns a transactions in mempool

ironfish mempool:transactions --fee 50:100          # returns all transactions with fees between 50 and 100
ironfish mempool:transactions --fee 50:             # returns all transactions with fees greater than 50
ironfish mempool:transactions --fee :100            # returns all transactions with fees less than 100

ironfish mempool:transactions --expiresIn :5        # returns all transactions that expire in the next 5 blocks
ironfish mempool:transactions --position   :100     # returns all the top 100 transactions in the mempool

ironfish mempool:transactions --position   :100     # returns all the top 100 transactions in the mempool

ironfish mempool:transactions --csv |> ./file.csv   # export all mempool transactions to a csv file

ironfish mempool:transactions --show 100            # display 100 transactions in the output
```

## Testing Plan
Local testing

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/mempool/transactions.ts
```diff
@@ -0,0 +1,204 @@
...
+      for await (const value of response.contentStream()) {
+        text.clearBaseLine(0)
+        text.setContent(renderTable(value, flags))
+        screen.render()
+      }
+    }
+  }
+}
+
+type TransactionRow = {
```
NullSoldier, line 127: you don't need this since you aren't doing any transforming

ironfish/src/rpc/routes/mempool/getTransactions.ts
```diff
@@ -0,0 +1,137 @@
...
+        )
+        .defined(),
+    })
+    .defined()
+
+router.register<typeof MempoolTransactionsRequestSchema, GetMempoolTransactionsResponse>(
+  `${ApiNamespace.mempool}/getTransactions`,
+  MempoolTransactionsRequestSchema,
+  (request, node): void => {
+    if (!request.data.stream) {
```
NullSoldier, line 71: I don't think we should implement this API, it should always be streaming.

ironfish/src/rpc/routes/mempool/getTransactions.ts
```diff
@@ -0,0 +1,137 @@
...
+  const serializedTransactions: GetMempoolTransactionResponse[] = []
+
+  let position = 0
+  for (const transaction of memPool.orderedTransactions()) {
+    if (request.limit !== undefined && serializedTransactions.length >= request.limit) {
+      break
+    }
+
+    if (includeTransaction(transaction, position, request)) {
+      serializedTransactions.push({
```
NullSoldier, line 106: this can blow node memory with how large the mempool can get you should stream these back

ironfish/src/rpc/routes/mempool/getTransactions.ts
danield9tqh: Yea I had the same concerns about having a large mempool. The main reason I kept it was that miners already run this function about once every minute to create new blocks. Running it once every 5 seconds is adding load but I thought it was reasonably similar to mining load to be possible.

Streaming is nice to have but not necessary for debugging use cases. Maybe we could implement a max mempool size to stream? Streaming 500tx mempool is still pretty fast but maybe we just disallow streaming with greater mempool sizes. Or I can remove the streaming all-together

ironfish/src/rpc/routes/mempool/getTransactions.ts
danield9tqh: Ok so sounds like your suggesting is to just get rid of mempool streaming altogether (we only allow one time queries of the mempool). And those one time queries we would stream the results. I think that makes sense to me. Streaming the transaction list with small mempool was pretty cool to see transactions being added in real time but it's by no means necessary so I'm ok with getting rid of it

dguenther (approved):

dguenther:
I think we can merge this, it's not very consequential if it doesn't work for some reason since it's mostly for debugging output.

PR #3450, opened 2023-02-16
feat: wallet supports importing viewonly accounts

EvanJRichard:
## Summary
Relies on merging https://github.com/iron-fish/ironfish/pull/3469

This changes wallet.importAccount to allow importing an account from provided view keys, rather than a spending key.
## Testing Plan
New test passes
## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/wallet.ts
```diff
@@ -1301,21 +1302,34 @@ export class Wallet {
...
+        outgoingViewKey: key.outgoing_view_key,
+        publicAddress: key.public_address,
+      }
+    } else {
+      // if spending key is not provided, use the provided view keys
+      const publicAddress: string = incomingViewKeyToPublicAddress(toImport.incomingViewKey)
+      accountValue = {
+        ...toImport,
+        id: uuid(),
+        spendingKey: null,
```
jowparks, line 1330: This is missing `viewKey`.

ironfish/src/wallet/account.ts
```diff
@@ -24,7 +24,15 @@ export const ACCOUNT_KEY_LENGTH = 32
 
 export const ACCOUNT_SCHEMA_VERSION = 1
 
-export type AccountImport = { name: string; spendingKey: string | null; version: number }
+export type AccountImport =
+  | { name: string; spendingKey: string; version: number }
```
jowparks, line 28: Lets split these into two non-exported types, `SpendingAccountImport`, `ViewAccountImport`

ironfish/src/wallet/wallet.test.ts
```diff
@@ -542,6 +542,26 @@ describe('Accounts', () => {
         'Account already exists with provided spending key',
       )
     })
+
+    it('should be able to import an account from solely its view keys', async () => {
+      const { node } = nodeTest
+      const account = await useAccountFixture(node.wallet, 'accountA')
+      expect(node.wallet.accountExists(account.name)).toEqual(true)
```
jowparks, line 549:
```suggestion
const key = generateKey()
```

ironfish/src/wallet/wallet.ts
```diff
@@ -1301,21 +1302,34 @@ export class Wallet {
...
+        publicAddress: publicAddress,
```
jowparks, line 1331:
```suggestion
        ...toImport,
        id: uuid(),
```

ironfish/src/wallet/wallet.ts
```diff
@@ -1301,21 +1302,34 @@ export class Wallet {
...
+      // if spending key is provided, derive everything from that, even if view keys were also provided.
+      const key = generateKeyFromPrivateKey(toImport.spendingKey)
+      accountValue = {
+        ...toImport,
+        version: ACCOUNT_SCHEMA_VERSION,
+        id: uuid(),
+        viewKey: key.view_key,
+        incomingViewKey: key.incoming_view_key,
+        outgoingViewKey: key.outgoing_view_key,
+        publicAddress: key.public_address,
```
jowparks, line 1322:
```suggestion
        ...key,
        version: ACCOUNT_SCHEMA_VERSION,
        id: uuid(),
```

ironfish/src/wallet/wallet.ts
```diff
@@ -1301,21 +1302,34 @@ export class Wallet {
...
+      }
+    } else {
+      // if spending key is not provided, use the provided view keys
+      const publicAddress: string = incomingViewKeyToPublicAddress(toImport.incomingViewKey)
```
jowparks, line 1326: Will not need this method when you update `ViewAccountImport`

ironfish/src/wallet/wallet.ts
```diff
@@ -1308,21 +1308,45 @@ export class Wallet {
       throw new Error(`Account already exists with the name ${toImport.name}`)
     }
```
jowparks, line 1310: I wonder if it is worth it to have two separate import methods for the two types?

EvanJRichard: something like:
```
  async importAccount(toImport: AccountImport): Promise<Account> {
    if (toImport.name && this.getAccountByName(toImport.name)) {
      throw new Error(`Account already exists with the name ${toImport.name}`)
    }
    if (spendingKey in toImport) {
       return importSpendingAccount(toImport)
    }  else {
       return importViewOnlyAccount(toImport)
    }
}
```
? We could do that. When the two paths were only slightly different, keeping it as one method made more sense, but now it's an awful lot of code all in one block. Implementing.

ironfish/src/wallet/wallet.ts
```diff
@@ -1303,28 +1303,60 @@ export class Wallet {
...
   }
 
-  async importAccount(toImport: AccountImport): Promise<Account> {
-    if (toImport.name && this.getAccountByName(toImport.name)) {
-      throw new Error(`Account already exists with the name ${toImport.name}`)
+  async importSpendingAccount(toImport: AccountImport): Promise<Account> {
```
jowparks, line 1306: Now that we have two methods we should use the two types, that way we wouldn't need this assertion

jowparks (approved):

PR #3451, opened 2023-02-16
CLI supports exporting (and importing) viewonly accounts

EvanJRichard:
## Summary
This PR proposes to add a flag `--viewonly` to the export command, ensuring an exported account will not have its spending key exported if supplied. (Import command already supports importing viewonly accounts.)

## Testing Plan
Unit tests added and pass. E2E test below: 

First, export showing a spendkey-less json:
```
yarn start wallet:export importtest --datadir=~/.ironfish-old --json --viewonly
```
```
{
    "version": 1,
    "name": "importtest",
    "spendingKey": null,
    "viewKey": "0fca6234c0cb83ec348a3771d747e24990613801e91cd2cf104dbb367f245c068caaded32f1d0ac860c9ea6773fd9fdfba84ecb692ac542258f45ac26445933f",
    "incomingViewKey": "090a2a6fa0e4f67923e1802aaa59dbe30229916307d81c5cd9c70f7853e65c06",
    "outgoingViewKey": "02dea4335761ed93e88c69251c323b3f1e22632efadfc1ccf852bc6a0d6d3bb5",
    "publicAddress": "e0555fa13f968e598ae9afd05d41cf17ae91c91423f3f07091871ed1a08d7ca6"
}
```
Next, get that as a bech blob, then delete it. Deletion console commands removed for brevity:
```
yarn start wallet:export importtest --datadir=~/.ironfish-old --viewonly
```
```
ironfishaccount0000010v38vetjwd5k7m3z8gcjcgnwv9kk2g36yf5k6ur0wf68getnws3zcgnnwpjkuerfden5keteygaxuatvdskzyanfv4m5keteygazyvrxvdsnvv3nx33nqcmz8qek2cenxsuxzvehxuckgde5xajnydpe8ycrvvfn8qcrzefex93kgvnrvccnqdryvf3rxd3hvcergdtrxqmrscmpv9jx2epnxfnrzepsv93nsd3svvuk2cfkxumnxeny89nxgenzvyurgetrvgmrjvnpvv6ngv3jx5uxvdp4v93nyd35xs6njvenvc3zcgnfde3k7mtfden4v6t9wa9k27fz8g3rqwfsvyexzdnxvycx2drxxcmnjv3nv5cnsvpjv9skzdfev33x2vesxgerjwf3xcenqdmy8qckxdtrvsukxdesvcmnsdfnv5mr2cesxc3zcgn0w46xwmmfden4v6t9wa9k27fz8g3rqvnyv4sngvenx5mnvvt9vsunxefc8p3nvwfjx5ckxvejxd3rxe33v5eryd3nxfjkvctyve3nzcmrvcur2vnzvvmxzvryxejrxcnzx53zcgnsw43xc6trg9jxgun9wdejyw3zv5cr2df4vesnzvmx8ymrsef48yuxzefev9nxgvp4vs6rzcmxxymkzefex93njvf5xgekvvmxxqmnqwf38qmnzetyx9snqwryxa3kzd3z05ln9wq6
```
Now import the deleted account:
```
yarn start wallet:import --datadir=~/.ironfish.old
```
```
Paste the output of wallet:export: ironfishaccount0000010v38vetjwd5k7m3z8gcjcgnwv9kk2g36yf5k6ur0wf68getnws3zcgnnwpjkuerfden5keteygaxuatvdskzyanfv4m5keteygazyvrxvdsnvv3nx33nqcmz8qek2cenxsuxzvehxuckgde5xajnydpe8ycrvvfn8qcrzefex93kgvnrvccnqdryvf3rxd3hvcergdtrxqmrscmpv9jx2epnxfnrzepsv93nsd3svvuk2cfkxumnxeny89nxgenzvyurgetrvgmrjvnpvv6ngv3jx5uxvdp4v93nyd35xs6njvenvc3zcgnfde3k7mtfden4v6t9wa9k27fz8g3rqwfsvyexzdnxvycx2drxxcmnjv3nv5cnsvpjv9skzdfev33x2vesxgerjwf3xcenqdmy8qckxdtrvsukxdesvcmnsdfnv5mr2cesxc3zcgn0w46xwmmfden4v6t9wa9k27fz8g3rqvnyv4sngvenx5mnvvt9vsunxefc8p3nvwfjx5ckxvejxd3rxe33v5eryd3nxfjkvctyve3nzcmrvcur2vnzvvmxzvryxejrxcnzx53zcgnsw43xc6trg9jxgun9wdejyw3zv5cr2df4vesnzvmx8ymrsef48yuxzefev9nxgvp4vs6rzcmxxymkzefex93njvf5xgekvvmxxqmnqwf38qmnzetyx9snqwryxa3kzd3z05ln9wq6
Account importtest imported.
The default account is now: importtest
```

## Breaking Change

the `importAccount` rpc endpoint is now `importSpendAccount`, so any consumers of this rpc endpoint will need to update the path used.

```
[X] Yes
```


ironfish/src/rpc/routes/wallet/exportAccount.ts
```diff
@@ -29,6 +31,7 @@ export const ExportAccountResponseSchema: yup.ObjectSchema<ExportAccountResponse
       .object({
         name: yup.string().defined(),
         spendingKey: yup.string().defined(),
```
jowparks, line 33: Should `spendingKey` not be not required for defined? If it can be null on return

ironfish-cli/src/commands/wallet/export.ts
```diff
@@ -60,14 +64,14 @@ export class ExportCommand extends IronfishCommand {
...
+    const viewOnly = flags.viewonly
 
     if (flags.language) {
       flags.mnemonic = true
     }
 
     const client = await this.sdk.connectRpc(local)
-    const response = await client.exportAccount({ account })
-
+    const response = await client.exportAccount({ account: account, viewOnly: viewOnly })
```
jowparks, line 74:
```suggestion
    const response = await client.exportAccount({ account, viewOnly })
```

jowparks (commented):
Looks pretty good, only one comment about historical RPC setting

jowparks:
Couple test failures

jowparks:
This looks good, we should add some tests for these RPCs though

ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -99,6 +127,7 @@ export class ImportCommand extends IronfishCommand {
 
   static verifySpendingKey(spendingKey: string): string | null {
     try {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-return
```
jowparks, line 130: Why do we need this comment line? We should be able to use:
```
return generateKeyFromPrivateKey(spendingKey)?.spendingKey ?? null
```

ironfish/src/rpc/routes/wallet/exportAccount.ts
jowparks, line 33: Isn't this optional now? It might be null

ironfish/src/rpc/routes/wallet/exportAccount.ts
```diff
@@ -5,11 +5,12 @@ import * as yup from 'yup'
 import { ApiNamespace, router } from '../router'
 import { getAccount } from './utils'
```
jowparks, line 7: I know you didn't introduce this change, but why is `account?` possibly undefined? Shouldn't we always requested account always have a value? I suppose this might be to handle export of default acccount, but it seems like that might be buried in the RPC somewhere in `getAccount`. Might be worth it to make it more explicit.

ironfish/src/rpc/routes/wallet/importViewAccount.ts
```diff
@@ -0,0 +1,78 @@
...
+    let isDefaultAccount = false
+    if (!node.wallet.hasDefaultAccount) {
+      await node.wallet.setDefaultAccount(account.name)
+      isDefaultAccount = true
+    }
+
+    request.end({
+      name: account.name,
+      isDefaultAccount,
+    })
```
jowparks, line 76: This is duplicated code from other RPC, we should consolidate. Not sure what the pattern is for handling these sorts of cases, but you might be able to find an example in other RPCs where code is reused across them.

ironfish/src/rpc/routes/wallet/exportAccount.ts
EvanJRichard: It'll be defined, but with value `null`, the way it is currently written.

ironfish/src/rpc/routes/wallet/exportAccount.ts
EvanJRichard: Yeah, if no `account` is defined, the default is account is exported, exactly.

ironfish/src/rpc/routes/wallet/exportAccount.ts
jowparks: https://github.com/iron-fish/ironfish/blob/38d2528903d3def6be0eaa4eaf400deb541b404c/ironfish/src/rpc/routes/peers/getPeers.ts#L63

ironfish/src/rpc/routes/wallet/importViewAccount.ts
EvanJRichard: The closest thing I can find is the mostly-empty `utils`, I'll push a changeset that uses that and we'll see how it looks

ironfish/src/rpc/routes/wallet/importSpendAccount.ts
```diff
@@ -5,20 +5,21 @@ import { generateKeyFromPrivateKey } from '@ironfish/rust-nodejs'
 import { v4 as uuid } from 'uuid'
 import * as yup from 'yup'
 import { ApiNamespace, router } from '../router'
```
jowparks, line 7: I don't think this is an issue since we are still in testnet, but this is a breaking change to the SDK. In the future, we would keep the route for `importAccount` and just redirect to `importSpendAccount`

ironfish/src/rpc/routes/wallet/importViewAccount.test.ts
```diff
@@ -0,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
```
jowparks, line 4: Why do you need these eslint disables? We should fix the type error, not disable the linter

ironfish/src/rpc/routes/wallet/importViewAccount.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export type ImportViewAccountRequest = {
+  account: ViewAccountImport
+  rescan?: boolean
+}
+
+export type ImportViewAccountResponse = {
+  name: string
+  isDefaultAccount: boolean
+}
```
jowparks, line 26: This is a duplicate type, we could consolidate in utils or make them equivalent.

ironfish/src/rpc/routes/wallet/importViewAccount.test.ts
```diff
@@ -0,0 +1,40 @@
...
+import { createRouteTest } from '../../../testUtilities/routeTest'
+
+describe('Route wallet/importViewAccount', () => {
+  const routeTest = createRouteTest(true)
+
+  it('should import a view only account that has no spending key', async () => {
+    const account = await routeTest.node.wallet.createAccount(uuid(), true)
+
+    // delete the account or else the import will fail
+    await routeTest.node.wallet.removeAccount(account)
```
jowparks, line 18:
```suggestion
const key = generateKey()
```

ironfish/src/rpc/routes/wallet/importViewAccount.test.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    const response = await routeTest.client
+      .request<any>('wallet/importViewAccount', {
+        account: {
+          name: account.name,
+          viewKey: account.viewKey,
+          publicAddress: account.publicAddress,
+          incomingViewKey: account.incomingViewKey,
+          outgoingViewKey: account.outgoingViewKey,
+          version: account.version,
```
jowparks, line 28:
```suggestion
          name: 'foo',
          viewKey: key.viewKey,
          publicAddress: key.publicAddress,
          incomingViewKey: key.incomingViewKey,
          outgoingViewKey: key.outgoingViewKey,
          version: 1,
```

ironfish/src/rpc/routes/wallet/importSpendAccount.test.ts
```diff
@@ -0,0 +1,37 @@
...
+import { createRouteTest } from '../../../testUtilities/routeTest'
+
+describe('Route wallet/importSpendAccount', () => {
+  const routeTest = createRouteTest(true)
+
+  it('should import a spending account', async () => {
+    const account = await routeTest.node.wallet.createAccount(uuid(), true)
+
+    // delete the account or else the import will fail
+    await routeTest.node.wallet.removeAccount(account)
```
jowparks, line 18: Same as below, just use `generateKey`

ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -4,16 +4,26 @@
...
 
+type AccountImport = {
+  name: string
+  spendingKey?: string
+  viewKey?: string
+  publicAddress?: string
+  incomingViewKey?: string
+  outgoingViewKey?: string
+  version: number
+}
```
jowparks, line 25: Split into two types, this will make code below much cleaner.

ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -62,11 +72,29 @@ export class ImportCommand extends IronfishCommand {
...
+          name: account.name,
+          viewKey: account.viewKey ?? '',
+          publicAddress: account.publicAddress ?? '',
+          incomingViewKey: account.incomingViewKey ?? '',
+          outgoingViewKey: account.outgoingViewKey ?? '',
+          version: account.version,
+        },
+        rescan: flags.rescan,
+      })
+    }
```
jowparks, line 97: If you split `AccountImport` into two types, this should reduce to something like follows:
```suggestion
      const rescan = flags.rescan
      result = 'spendingKey' in account  ? 
      await client.importSpendAccount({account, rescan}) : 
      await client.importViewAccount({account, rescan})
```
syntax probably isn't right for awaits, but you get the point

jowparks (changes requested):
Looks good generally, just some cleanup that needs to happen

ironfish/src/rpc/routes/wallet/importViewAccount.test.ts
EvanJRichard: That snuck in from the test that was used as a template. Removing/fixing.

ironfish-cli/src/commands/wallet/import.ts
EvanJRichard: I ran into a roadblock doing that, but didn't document what it was, hm. Will retry/re-do and see where we get.

EvanJRichard: Hm, worked fine! Wonder what I was doing wrong last time.

ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -62,11 +65,11 @@ export class ImportCommand extends IronfishCommand {
...
     }
 
-    const result = await client.importAccount({
-      account,
-      rescan: flags.rescan,
-    })
-
+    const rescan = flags.rescan
+    const result =
+      'spendingKey' in account && account.spendingKey
```
jowparks, line 70: Nit: Type system should be able to pick up on the ternary and the type is implied to be SpendingAccountImport on this line, and ViwAccountImport on next. Also you can change "account: account", to just "account"

jowparks (approved):
One nit, looks good

ironfish-cli/src/commands/wallet/import.ts
EvanJRichard: Changed `account: account` to just `account`, but still need the `as ViewAccountImport` cast on the next line

ironfish/src/rpc/routes/wallet/exportAccount.test.ts
```diff
@@ -0,0 +1,59 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
EvanJRichard, line 6: I missed that this is still in here, worth it to remove you think @jowparks ?

jowparks: Good call, I fixed it up, along with some of the import stuff to make the types live in the right spot

PR #3455, opened 2023-02-17
feat: use view key for nullifier

jowparks:
## Summary
We only need the `nullifier_deriving_key` for verifying the nullifier, not the `spend_key`.  These changes are a result of the identification of view only wallet incompatibility from this PR review comment:
https://github.com/iron-fish/ironfish/pull/3335#discussion_r1103230014

TODO:
add links to previous PR in series and rename
## Testing Plan
Tests pass
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust-nodejs/src/structs/note.rs
+++ b/ironfish-rust-nodejs/src/structs/note.rs
@@ -5,11 +5,12 @@
 use ironfish_rust::{
     assets::asset::{asset_generator_from_id, ID_LENGTH as ASSET_ID_LENGTH},
     note::{AMOUNT_VALUE_SIZE, GENERATOR_SIZE, MEMO_SIZE, SCALAR_SIZE},
+    ViewKey,
 };
 use napi::{bindgen_prelude::*, JsBuffer};
 use napi_derive::napi;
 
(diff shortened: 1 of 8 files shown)


NullSoldier (approved):

ironfish-rust-nodejs/src/structs/note.rs
```diff
@@ -145,12 +146,12 @@ impl NativeNote {
     /// only at the time the note is spent. This key is collected in a massive
     /// 'nullifier set', preventing double-spend.
     #[napi]
```
jowparks, line 148: I need a seal.

mat-if (changes requested):
https://github.com/iron-fish/ironfish/blob/feat-view-key-for-nullifier/ironfish/src/primitives/note.ts#L92-L93 If I'm not mistaken, this needs to be updated. I'm a little worried that there was no test coverage to catch this.

mat-if (approved):
Nvm, false alarm. Spoke with Joe, it's just a naming thing

jowparks:
> Nvm, false alarm. Spoke with Joe, it's just a naming thing

Still a good catch, no more cruft!

PR #3456, opened 2023-02-17
Add mempool to wallet constructor

NullSoldier:
## Summary

No reason to have to pass it in, make it consistent with our other objects. If you need a high level system you should take it into the constructor.

## Breaking Change

```
[ ] Yes
```


jowparks (approved):
LG as long as tests pass.

rohanjadvani (approved):

PR #3458, opened 2023-02-17
Remove use of routeTest.client.request

NullSoldier:
## Summary

We don't need to use this unless you want a raw untyped request. Most of these tests only want the typed request, and in fact typing these found some bugs in the tests.

## Testing Plan

these are changes to the tests

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/chain/getTransactionStream.test.ts
+++ b/ironfish/src/rpc/routes/chain/getTransactionStream.test.ts
@@ -11,16 +11,9 @@ import {
   useMintBlockFixture,
 } from '../../../testUtilities/fixtures'
 import { createRouteTest } from '../../../testUtilities/routeTest'
-import { GetTransactionStreamResponse } from './getTransactionStream'
 
 describe('Route chain.getTransactionStream', () => {
   const routeTest = createRouteTest()
-
-  it('should fail if no incoming view key is specified', async () => {
-    await expect(
-      routeTest.client.request('chain/getTransactionStream', {}).waitForEnd(),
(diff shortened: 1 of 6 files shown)


ironfish/src/rpc/routes/chain/getTransactionStream.test.ts
```diff
@@ -11,17 +11,10 @@ import {
...
-import { GetTransactionStreamResponse } from './getTransactionStream'
 
 describe('Route chain.getTransactionStream', () => {
   const routeTest = createRouteTest()
 
-  it('should fail if no incoming view key is specified', async () => {
-    await expect(
-      routeTest.client.request('chain/getTransactionStream', {}).waitForEnd(),
-    ).rejects.toThrow('Request failed (400) validation: incomingViewKey is a required field')
-  })
```
NullSoldier, line 23: Removed because this duplicates YUP testing

ironfish/src/rpc/routes/wallet/create.test.slow.ts
```diff
@@ -47,28 +47,14 @@ describe('Route wallet/create', () => {
...
-      await routeTest.client.request('wallet/create').waitForEnd()
-    } catch (e: unknown) {
-      if (!(e instanceof RpcRequestError)) {
-        throw e
-      }
-      expect(e.status).toBe(400)
-      expect(e.code).toBe(ERROR_CODES.VALIDATION)
-      expect(e.message).toContain('name')
-    }
-  })
```
NullSoldier, line 62: Removed because this duplicates YUP testing

mat-if (approved):

PR #3467, opened 2023-02-17
chore: remove snake case key

jowparks:
## Summary
Refactor NAPI nodejs representation of `Key` to use default variable name export (instead of camel case).
## Testing Plan
tests pass
## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/wallet.ts
```diff
@@ -1269,11 +1269,11 @@ export class Wallet {
       version: ACCOUNT_SCHEMA_VERSION,
       id: uuid(),
       name,
```
jowparks, line 1271: These will be cleaned up in subsequent PR

mat-if (approved):

PR #3468, opened 2023-02-18
Change eligibiliy check

NullSoldier:
## Summary
 - Added to burn / send as well as mint
 - Colorized output to be yellow 
 - Skip in case API is unreachable
 - Allow user to disable eligibility check for offline signing to work

Screen Shot 2023-02-17 at 7 14 34 PM

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/burn.ts
+++ b/ironfish-cli/src/commands/wallet/burn.ts
@@ -16,6 +16,7 @@ import { IronfishCommand } from '../../command'
 import { IronFlag, parseIron, RemoteFlags } from '../../flags'
 import { ProgressBar } from '../../types'
 import { selectAsset } from '../../utils/asset'
+import { doEligibilityCheck } from '../../utils/testnet'
 
 export class Burn extends IronfishCommand {
   static description = 'Burn tokens and decrease supply for a given asset'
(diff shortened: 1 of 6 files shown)


jowparks (approved):
LGTM, real changes are just adding chalk and making the eligibility check reusable

ironfish-cli/src/commands/wallet/burn.ts
```diff
@@ -67,12 +68,21 @@ export class Burn extends IronfishCommand {
...
+      allowNo: true,
+      description: 'check testnet eligibility',
+    }),
   }
 
   async start(): Promise<void> {
     const { flags } = await this.parse(Burn)
     const client = await this.sdk.connectRpc(false, true)
 
+    if (flags.eligibility) {
```
lwisne, line 82: I had only added it to the mint on the assumption that people would be generally running both commands and it would reduce API load.

Not strongly opinionated her just giving context

lwisne (approved):

ironfish-cli/src/commands/wallet/burn.ts
NullSoldier: oh yeah, that's a good reason. I think our API is performing OK now so it should be fine.

Issue #3477, opened 2023-02-19
wallet:mint shouldn't show non-owner assets


wd021:
### What happened?

You should only be allowed to increase mint for assets that you're the owner of, but `wallet:mint` currently shows non-owner assets as options as well. When you select one of these, it'll just mint a new asset, but we should remove it from the list.. 

As can be seen in the output below, i selected an asset that was sent to me (flamengo) and when i select it to mint, it mints a new assetId

### Version

0.1.67

### Relevant log output

```shell
Dannys-MacBook-Pro-2:ironfish-cli dannywitters$ yarn start wallet:mint
yarn run v1.22.10
$ yarn build && yarn start:js wallet:mint
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc --inspect=:0 --inspect-publish-uid=http --enable-source-maps bin/run wallet:mint
Do you want to create a new asset (Y/N)?: n
? Select the asset you wish to mint de195b7f36883a41cfee50594e212c8ca054829c7f78998124e578c2211335ad (flamengo) (10.00000000)
Enter the amount to mint in the custom asset: 10
? Select the fee you wish to use for this transaction slow: 0.00000001 IRON

You are about to mint  
de195b7f36883a41cfee50594e212c8ca054829c7f78998124e578c2211335ad 10.00000000 plus a transaction fee of $IRON 0.00000001 with the account abcd

* This action is NOT reversible *

Do you confirm (Y/N)?: y
Creating the transaction: [████████████████████████████████████████] 100% | ETA: 0s

Minted asset flamengo from abcd
Asset Identifier: f0143ad69433e3a060018dcee0f622a9f06c3e77c67494d4f684a3562b701165
Value: f0143ad69433e3a060018dcee0f622a9f06c3e77c67494d4f684a3562b701165 10.00000000

Transaction Hash: 16b0dcf8f33a33eac345cf472d676251f7cc20b04f423da6fd6cf3ec40656c1f

Find the transaction on https://explorer.ironfish.network/transaction/16b0dcf8f33a33eac345cf472d676251f7cc20b04f423da6fd6cf3ec40656c1f (it can take a few minutes before the transaction appears in the Explorer)
✨  Done in 18.17s.
```

### Graffiti

wd021

PR #3478, opened 2023-02-19
Filter out non-owner assets from mint screen

wd021:
## Summary
Fix https://github.com/iron-fish/ironfish/issues/3477. `wallet:mint` should filter out non-owner assets. Adding `ownerOnly` flag to GetBalancesRequest rpc.

## Testing Plan
Run `ironfish wallet:mint` and see that non-owner assets no longer appear rom select screen. Run `ironfish wallet:send/burn` and confirm it still shows all assets.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/getBalances.ts
+++ b/ironfish/src/rpc/routes/wallet/getBalances.ts
@@ -9,6 +9,7 @@ import { getAccount } from './utils'
 export interface GetBalancesRequest {
   account?: string
   confirmations?: number
+  ownerOnly?: boolean
 }
 
 export interface GetBalancesResponse {
@@ -31,6 +32,7 @@ export const GetBalancesRequestSchema: yup.ObjectSchema<GetBalancesRequest> = yu
   .object({
     account: yup.string().optional(),
     confirmations: yup.number().min(0).optional(),
+    ownerOnly: yup.boolean().optional(),
   })
   .defined()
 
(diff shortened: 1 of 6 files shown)


wd021:
rebased

ironfish/src/rpc/routes/wallet/getBalances.ts
```diff
@@ -9,6 +9,7 @@ import { getAccount } from './utils'
 export interface GetBalancesRequest {
   account?: string
   confirmations?: number
+  ownerOnly?: boolean
```
dguenther, line 12: I think this is a good improvement, but If you don't mind changing it up, I'd prefer doing the filtering in the CLI rather than adding filtering to the request, since we haven't established a pattern for doing that in RPC calls yet.

Basically this would mean:
* `getBalances` should return an `assetOwner` with each balance
* in `utils\asset.ts`, there should be a check at the top of the function similar to `options.showNativeAsset` that fetches the public address for an account, then filters `balances` by `assetOwner` using that public address.

wd021: ok, updated to return `assetOwner` with getBalances rpc. renamed utils/asset.ts flag to `option.showNonOwnerAsset`.

dguenther:
Cool, this looks good! I want to manually test this out, then this looks good to merge

dguenther (approved):

PR #3487, opened 2023-02-20
fixes walletdb migrations that rely on old accounts

hughy:
## Summary

we've recently made changes to the way that accounts are stored in the walletdb. it's possible that a user may have a database that needs to apply migrations that rely on the old AccountValue encoding before applying the migration to the new encoding.

this results in an encoding error when an older migration (pre-021) attempts to read from the accounts store, which has not yet been migrated.

adds a function, GetOldAccounts, to simplify reading accounts from the old accounts store in older migrations. uses this function for loading accounts in all migrations before Migration021.

## Testing Plan

Running migrations before changes:
image

Running migrations after changes:
image

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):

NullSoldier:
Ok, let's establish a standard for writing future-proof migrations. We need to establish a more formal template.

PR #3499, opened 2023-02-21
feat(ironfish): add a recently evicted cache to the mempool

holahula:
## Summary
Adds a new `RecentlyEvictedCache` to the mempool. 

A later PR will introduce a maximum capacity to the mempool. When the mempool is full, it is possible for transactions to be evicted to make room for more appealing ones. These evicted transactions will be added to this new cache to prevent immediate re-submission into the mempool. As transaction validation is expensive, it does not make sense to validate a transaction, _t_,  if it is known that _t_ will not make it into the mempool (as it was recently evicted). 

The `RecentlyEvictedCache` has a maximum capacity determined by the upstream maximum capacity of the mempool. Transactions can only be removed from this cache in the two following ways:

1. The cache is full and a new transaction is added. The transaction with the highest `feeRate` (fee/byte) will be removed as this has the highest chance to be re-introduced into the mempool
2. A new block is added. Transactions that have spent longer than their `maxAge` in the cache will be removed. This is to prevent transactions from never being re-added to the mempool because they were stuck indefinitely in the cache 

## Testing Plan
See included unit tests

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/memPool/recentlyEvictedCache.ts
+++ b/ironfish/src/memPool/recentlyEvictedCache.ts
@@ -1,0 +1,225 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { BufferMap } from 'buffer-map'
+import { Assert } from '../assert'
+import { createRootLogger, Logger } from '../logger'
+import { MetricsMonitor } from '../metrics'
+import { getTransactionSize } from '../network/utils/serializers'
+import { Transaction, TransactionHash } from '../primitives/transaction'
+import { PriorityQueue } from './priorityQueue'
+
+interface EvictionEntry {
+  hash: TransactionHash
+  feeRate: bigint
+}
+
+interface ExpirationEntry {
+  hash: TransactionHash
+  expiration: number
+}
+
+export class RecentlyEvictedCache {
+  private readonly metrics: MetricsMonitor
+  private readonly logger?: Logger
+
+  private readonly capacity: number = 1000 // TODO: fix this default initialization
+
+  private readonly transactions = new BufferMap<Transaction>()
+  /**
+   * Keep track of number of bytes stored in the transaction map
+   * */
+  private transactionsBytes = 0
+
+  private readonly maxJailTime: number = 10 // TODO: fix -> this is max duration txns exist in cache
+
+  private currentSequence = 0
+
+  /**
+   * A priority queue of transactions to evict from this cache when full.
+   *
+   * Transactions are evicted in order of `feeRate` descending because
+   * they are most likely to be re-included into the mempool
+   */
+  private readonly evictionQueue: PriorityQueue<EvictionEntry>
+
+  private readonly expirationQueue: PriorityQueue<ExpirationEntry>
+
+  /**
+   * Creates a new cache to track transactions that have recently been evicted.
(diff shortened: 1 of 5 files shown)


ironfish/src/memPool/recentlyEvictedCache.ts
```diff
@@ -0,0 +1,225 @@
...
+  private readonly metrics: MetricsMonitor
+  private readonly logger?: Logger
+
+  private readonly capacity: number = 1000 // TODO: fix this default initialization
+
+  private readonly transactions = new BufferMap<Transaction>()
+  /**
+   * Keep track of number of bytes stored in the transaction map
+   * */
+  private transactionsBytes = 0
```
danield9tqh, line 32: I think we might be able to get rid of this. We keep track of this in mempool because transaction entries can be variable size but here they are all the same size right? So this could be deduced from the total size

ironfish/src/memPool/recentlyEvictedCache.ts
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  private remove(hash: TransactionHash): Transaction | void {
+    const transactionToRemove = this.get(hash)
+
+    if (!transactionToRemove) {
+      return
+    }
+
+    this.transactions.delete(hash)
```
danield9tqh, line 79: I think I would feel more comfortable keeping all these strongly in sync and removing from all queues at the same time. I'm trying to think of scenarios where a transaction gets removed by one process then re-added and it gets added to a queue twice somehow

danield9tqh: Also if we keep them in sync then we don't even need the `this.transactions` map we can just use one of the priority queues which has an internal map

ironfish/src/memPool/recentlyEvictedCache.ts
```diff
@@ -0,0 +1,225 @@
...
+   *
+   */
+  add(transaction: Transaction): void {
+    const hash = transaction.hash()
+
+    if (this.has(hash)) {
+      return
+    }
+
+    if (this.isFull()) {
```
danield9tqh, line 143: Do we want to add first then evict? So the current transaction `feeRate` is taken into account

ironfish/src/memPool/recentlyEvictedCache.ts
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    return this.remove(hashToRemove)
+  }
+
+  count(): number {
+    return this.transactions.size
+  }
+
+  isFull(): boolean {
```
danield9tqh, line 123: Probably > capacity would be safer here

ironfish/src/memPool/recentlyEvictedCache.ts
holahula: removed

ironfish/src/memPool/recentlyEvictedCache.ts
```diff
@@ -0,0 +1,198 @@
...
+   * @param transactionHash the hash of the transaction to add
+   * @param feeRate the fee/byte rate of the transaction
+   * @param currentBlockSequence the current block sequence when the transaction was added
+   *
+   * @returns true if the transaction was successfully added to the cache, false if it was already present
+   */
+  add(
+    transactionHash: TransactionHash,
+    feeRate: bigint,
+    currentBlockSequence: number,
```
danield9tqh, line 143: Maybe we could just pass this in as `expirationSequence`

danield9tqh (approved):

PR #3501, opened 2023-02-21
fix: fixes 'expired transaciton' test failure on fixture regenerate

jowparks:
## Summary
When trying to regenerate fixtures, this test was failing because there was likely new validation added to wallet. @hughy paired with me to update the test so validation passes.

## Testing Plan
tests pass
## Breaking Change

```
[ ] Yes
```


hughy (approved):
🙌

PR #3502, opened 2023-02-21
Remove potential race condition when disconnecting blocks from mempool

danield9tqh:
## Summary
When `onDisconnectBlock` is called in the mempool it doesn't currently check for existing transactions with overlapping nullifiers. This is an optimization because if transactions are being added from disconnected blocks synchronously then it would be impossible for there to be a double spend in the mempool because these transactions were confirmed on chain. We check the chain before adding double spends into the mempool so in theory there should be no overlapping nullifiers from a disconnecting block. However because the `onDisconnectBlock` function [is called asynchronously](https://github.com/iron-fish/ironfish/blob/9f96511beec7250f996bb1f1fa72aca9d99512ff/ironfish/src/blockchain/blockchain.ts#L651) it is possible (although highly unlikely) that a transaction with overlapping nullifiers could be added to the mempool. 

Additionally this logic relies on double spends never being accepted into the mempool which is the case currently but that could change in the future if for example we implement a confirmation period (~5 block) for the mempool. In this case we want `addTransaction` to never put the internal mempool data structures into an inconsistent state

## Breaking Change

```
[ ] Yes
```


EvanJRichard (approved):
Seems cleaner too, I like the additional tests.

PR #3504, opened 2023-02-21
adds template migration

hughy:
## Summary

creates a template migration, 000-template, to define best practices for developing future-proof migrations.

we've written several migrations that depend on schemas that are defined outside of the migrations directory. if those schemas later change (such as in a later migration), then that migration will break.

the template migration copies from Migration022 as a model for future-proof migrations.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/migrations/data/000-template.ts
+++ b/ironfish/src/migrations/data/000-template.ts
@@ -67,7 +67,5 @@ export class Migration000 extends Migration {
       await stores.old.accounts.put(account.id, account, tx)
     }
 
-    // and clears data from new stores
-    await stores.new.accounts.clear(tx)
   }
 }
--- a/ironfish/src/migrations/data/000-template.ts
+++ b/ironfish/src/migrations/data/000-template.ts
@@ -47,8 +47,6 @@ export class Migration000 extends Migration {
       await stores.new.accounts.put(account.id, migrated, tx)
     }
 
-    // and clears data from old stores
-    await stores.old.accounts.clear(tx)
   }
 
   async backward(
--- a/ironfish/src/migrations/data/000-template.ts
+++ b/ironfish/src/migrations/data/000-template.ts
@@ -14,7 +14,7 @@ export class Migration000 extends Migration {
   path = __filename
 
   prepare(node: IronfishNode): IDatabase {
-    /* replace with node.chain.location if applying migration to the blockchain
+    /* replace line below with node.chain.location if applying migration to the blockchain
      * database
      */
     return createDB({ location: node.config.walletDatabasePath })
--- a/ironfish/src/migrations/data/000-template.ts
+++ b/ironfish/src/migrations/data/000-template.ts
@@ -1,0 +1,75 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { generateKeyFromPrivateKey } from '@ironfish/rust-nodejs'
+import { Logger } from '../../logger'
+import { IronfishNode } from '../../node'
+import { IDatabase, IDatabaseTransaction } from '../../storage'
+import { createDB } from '../../storage/utils'
+import { Migration } from '../migration'
+import { GetNewStores } from './000-template/schemaNew'
+import { GetOldStores } from './000-template/schemaOld'
+
+export class Migration000 extends Migration {
+  path = __filename
+
+  prepare(node: IronfishNode): IDatabase {
+    /* replace with node.chain.location if applying migration to the blockchain
+     * database
+     */
+    return createDB({ location: node.config.walletDatabasePath })
+  }
+
+  async forward(
+    node: IronfishNode,
+    db: IDatabase,
+    tx: IDatabaseTransaction | undefined,
+    logger: Logger,
+  ): Promise<void> {
+    // use GetOldStores and GetNewStores to attach datastores with the old and new schemas to the database
+    const stores = {
+      old: GetOldStores(db),
+      new: GetNewStores(db),
+    }
+
+    // forward migration inserts from old stores into new stores
+    for await (const account of stores.old.accounts.getAllValuesIter(tx)) {
+      logger.info(`  Migrating account ${account.name}`)
+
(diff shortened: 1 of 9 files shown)


ironfish/src/migrations/data/000-template/schemaOld.ts
```diff
@@ -0,0 +1,92 @@
...
+    return size
+  }
+}
+
+export function GetOldStores(db: IDatabase): {
+  accounts: IDatabaseStore<{ key: string; value: AccountValue }>
+} {
+  const accounts: IDatabaseStore<{ key: string; value: AccountValue }> = db.addStore(
+    {
+      name: 'a21',
```
jowparks, line 84:
```suggestion
      name: 'a',
```

This was done this way because we wrote the migration incorrectly at first. The `false` option below allows us to overwrite table so we can use the same table name for this particular migration

ironfish/src/migrations/data/000-template/schemaOld.ts
```diff
@@ -0,0 +1,92 @@
...
+import { IDatabase, IDatabaseEncoding, IDatabaseStore, StringEncoding } from '../../../storage'
+
+/* The schemaOld.ts file must define the value schema and database encoding for
+ * ALL datastores that the migration reads from. Even if the migration does not
+ * modify a datastore _A_, if the migration needs to read data from _A_ in order
+ * to write to another datastore _B_, then the schema and encoding for _A_ must
+ * be defined in schemaOld.ts.
+ *
+ * The example below is taken from Migration022, which added the viewKey field
+ * to the AccountValue schema. */
```
jowparks, line 15: We can probably simplify:

If you read or write from a store, you will need to have a copy of the schema (example `AccountValue`) and IDatabaseEncoding (example `AccountValueEncoding`) here. Generally for `schemaOld.ts`, you can copy the schema and encoding from the originating file on `staging` branch.

ironfish/src/migrations/data/000-template/schemaNew.ts
```diff
@@ -0,0 +1,97 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { PUBLIC_ADDRESS_LENGTH } from '@ironfish/rust-nodejs'
+import bufio from 'bufio'
+import { IDatabase, IDatabaseEncoding, IDatabaseStore, StringEncoding } from '../../../storage'
+
```
jowparks, line 7: Add line:
This schema/encoding should match the schema of your PR branch. This way, once your branch is merged it will be compatible with the migrated database

ironfish/src/migrations/data/000-template.ts
```diff
@@ -0,0 +1,75 @@
...
+      const migrated = {
+        ...account,
+        viewKey: key.viewKey,
+      }
+
+      await stores.new.accounts.put(account.id, migrated, tx)
+    }
+
+    // and clears data from old stores
+    await stores.old.accounts.clear(tx)
```
jowparks, line 51:
```suggestion
```

jowparks: Since the name for the table in the `new`/`old` files are the same, we can delete the clear. The accounts will be overwritten in transaction as the `account.id` is unchanged.

ironfish/src/migrations/data/000-template.ts
```diff
@@ -0,0 +1,75 @@
...
+import { createDB } from '../../storage/utils'
+import { Migration } from '../migration'
+import { GetNewStores } from './000-template/schemaNew'
+import { GetOldStores } from './000-template/schemaOld'
+
+export class Migration000 extends Migration {
+  path = __filename
+
+  prepare(node: IronfishNode): IDatabase {
+    /* replace with node.chain.location if applying migration to the blockchain
```
jowparks, line 17:
```suggestion
    /* replace line below with node.chain.location if applying migration to the blockchain
```

ironfish/src/migrations/data/000-template.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    // backward migration re-inserts data from new stores into old stores
+    for await (const account of stores.new.accounts.getAllValuesIter(tx)) {
+      logger.info(`  Migrating account ${account.name}`)
+      await stores.old.accounts.put(account.id, account, tx)
+    }
+
+    // and clears data from new stores
+    await stores.new.accounts.clear(tx)
```
jowparks, line 73:
```suggestion
```

ironfish/src/migrations/data/000-template/schemaNew.ts
```diff
@@ -0,0 +1,97 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { PUBLIC_ADDRESS_LENGTH } from '@ironfish/rust-nodejs'
+import bufio from 'bufio'
+import { IDatabase, IDatabaseEncoding, IDatabaseStore, StringEncoding } from '../../../storage'
```
NullSoldier, line 6: Should we make the template assuming more than 1 table will be modified? That's more of a pain but more of a scalable example.

hughy: I think this is a good idea! I'll change the example into something contrived with more than 1 table read and more than 1 table write.

ironfish/src/migrations/data/000-template/schemaOld.ts
hughy: That sounds good. The `_A_` and `_B_` isn't very clear

NullSoldier (approved):

ironfish/src/migrations/data/000-template/schemaNew.ts
NullSoldier: Done

ironfish/src/migrations/data/000-template/schemaOld.ts
NullSoldier: Took jow's example, but trimmed the part about staging

PR #3505, opened 2023-02-21
Replace progress bar with action in wallet:post

NullSoldier:
## Summary
The progress bar is faked, and we should use an action spinner instead. Also added UX guidelines for this.

## Breaking Change

```
[ ] Yes
```


mat-if (approved):

rohanjadvani (approved):
This change looks good but it seems like there are 3 different things happening:
* Updating style guide
* Using `broadcast` instead of `offline`
* Changing the progress bar to the action

Generally we've split out these changes per PR in the past. I'll leave it up to you if you want to do that.

NullSoldier:
Good call, the style guide and action are the same change but the offline -> broadcast is unrelated. I'll separate these out better in the future.

PR #3506, opened 2023-02-22
Handles insufficient funds for one or more fee rates

hughy:
## Summary

we use Promise.all to parallelize createTransaction requests for all three fee rates (slow, average, fast) when neither a fee nor fee rate are specified in the send command.

however, if the sender account does not have enough funds to pay the fast fee rate, then the whole transaction fails. the user cannot select the slow fee rate even if their account has enough funds to pay that fee.

handles insufficient balance errors when creating transactions for different fee rates. inserts an 'Insufficient funds' option into the inquirer prompt and throws the error if that option is selected.

removes parallelization of the three fee rate requests to simplify error handling. Promise.allSettled is an alternative that would preserve parallelization, but makes it more difficult to distinguish the type of error when a request fails.

https://asciinema.org/a/VA4SGLM1Xklb3oI1nqwMhrMrj

## Testing Plan

Randomly throw errors in the backend to test what happens

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/utils/currency.ts
+++ b/ironfish-cli/src/utils/currency.ts
@@ -46,11 +46,11 @@ export async function promptCurrency(options: {
     return null
   }
 
-  const [fee, error] = CurrencyUtils.decodeTry(input)
+  const [amount, error] = CurrencyUtils.decodeTry(input)
 
   if (error) {
     throw error
   }
 
-  return fee
+  return amount
 }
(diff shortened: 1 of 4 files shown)


ironfish-cli/src/utils/currency.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  const input = await CliUx.ux.prompt(text, {
+    required: options.required,
+  })
+
+  if (!input) {
+    return null
+  }
+
+  const [fee, error] = CurrencyUtils.decodeTry(input)
```
hughy, line 49: nit - should this be something like `value` or `amount` instead of `fee`?

We could use this for prompting fees or amounts to send/mint/burn

NullSoldier (approved):

ironfish-cli/src/utils/currency.ts
NullSoldier: Fixed

ygao76:
Do we want to fix the issue in the same way for wallet:mint and burn?

PR #3508, opened 2023-02-22
changes remaining low/medium/high references

hughy:
## Summary

finishes renaming fee rate priority levels from low, medium, and high to slow, average, and fast.

## Breaking Change

```
[ ] Yes
```


EvanJRichard (approved):
Lints are good, tests are good, all LGTM. Old configs will have the wrong variable names but that's fine right?

hughy:
> Old configs will have the wrong variable names but that's fine right?

There may be some users who have set overrides for these, but I'd be surprised if there are many (or any) 😅 

Even for those users there shouldn't be any impact for the rest of the testnet: since our blocks are mostly empty the estimated fee rate will probably remain at 1 ORE

PR #3509, opened 2023-02-22
Fix mempool tests

danield9tqh:
## Summary
Mempool tests were not passing when the fixtures were regenerated. This was because the default for transaction broadcasting changed without regenerating the tests.

## Breaking Change

```
[ ] Yes
```


ironfish/src/testUtilities/fixtures/blocks.ts
```diff
@@ -215,6 +215,7 @@ export async function useBlockWithTx(
     expiration?: number
     fee?: number
   } = { expiration: 0 },
+  broadcast = true,
```
hughy, line 218: Should this be a part of `options`?

hughy (approved):

ironfish/src/testUtilities/fixtures/blocks.ts
danield9tqh: I tried a few different APIs for this function that being one of them. Ultimately decided on this one bc it was the made the fewest changes in other test files. This function definition definitely need to be improved though

hughy: I wonder if a lot of the fixture functions could be easier to use if they expected a single object with all of the arguments as named fields 🤔 

I've found it tricky with some of the fixtures to have to specify a long list of arguments that I don't care about just to set the one that I need, like a transaction expiration or fee.

PR #3514, opened 2023-02-22
feat(ironfish): add max capacity to mempool

holahula:
## Summary
The primary objective of this PR is to add a configurable max capacity to the mempool. 

The following functionality is included:
- Add a max size for the mempool
- Add eviction logic that evicts the transactions with the lowest fee rate (fee/byte) when newer, more appealing transactions are introduced into a full mempool
- Add a recently evicted cache to the mempool to track transactions that have recently been evicted. This serves as an optimization mechanism to prevent nodes from processing transactions that will not make it into the mempool
- Makes transaction gossip changes to account for not all transactions being in the mempool 
- Adds telemetry support to various metrics surrounding mempool size, saturation and eviction statistics. 

This PR encompasses a few smaller PRs that each implement a required functionality. For deeper descriptions for a specific section, see below. 
- a configurable max capacity for the mempool 
    - #3524
- a recently evicted cache for transactions removed from the mempool when full 
    - #3499
- required transaction gossip changes following mempool changes
    - #3575
- telemetry to track the impact of introducing a max size to the mempool
    - #3533

## Testing Plan
See included unit / integration tests. We will also build a test network infrastructure to test sending a large number of transactions

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/node/getStatus.ts
```diff
@@ -300,6 +316,13 @@ function getStatus(node: IronfishNode): GetNodeStatusResponse {
     memPool: {
       size: node.metrics.memPoolSize.value,
       sizeBytes: node.memPool.sizeBytes(),
+      maxSizeBytes: node.memPool.maxSizeBytes,
+      saturation: node.memPool.saturation(),
```
danield9tqh, line 320: Could probably remove this saturation since we're not using it

ironfish/src/rpc/routes/node/getStatus.ts
```diff
@@ -143,6 +150,15 @@ export const GetStatusResponseSchema: yup.ObjectSchema<GetNodeStatusResponse> =
       .object({
         size: yup.number().defined(),
         sizeBytes: yup.number().defined(),
+        maxSizeBytes: yup.number().defined(),
+        saturation: yup.number().defined(),
```
danield9tqh, line 154: Could probably remove this saturation since we're not using it

dguenther:
I'm going to review this today -- generally for a branch of this size, I'd prefer a little commentary in the PR description around main motivations for the change or any considerations when working on the branch, I understand that this started off as a holding branch though.

ironfish/src/fileStores/config.ts
```diff
@@ -315,6 +329,11 @@ export const ConfigOptionsSchema: yup.ObjectSchema<Partial<ConfigOptions>> = yup
     networkId: yup.number().integer().min(0),
     customNetwork: yup.string().trim(),
     maxSyncedAgeBlocks: yup.number().integer().min(0),
+    mempoolMaxSizeBytes: yup
+      .number()
+      .integer()
+      .min(2 * MEGABYTES), // one blocks worth of transactions
```
dguenther, line 335: This should be using `consensus.maxBlockSizeBytes`, so we don't have to change multiple places if we support variable block sizes.

It's not very easy to access that in the config though, so I wonder if it might be better to have the config min at 0, then where `mempoolMaxSizeBytes` is used, use the max of `maxBlockSizeBytes, mempoolMaxSizeBytes`. Alternatively, is it that much of an issue if someone sets a very small mempool size? They'll probably have to spend a bit more on bandwidth fetching missing transactions

ironfish/src/fileStores/config.ts
```diff
@@ -399,6 +418,8 @@ export class Config extends KeyStore<ConfigOptions> {
       networkId: DEFAULT_NETWORK_ID,
       customNetwork: '',
       maxSyncedAgeBlocks: 60,
+      memPoolMaxSizeBytes: 60 * MEGABYTES,
```
dguenther, line 421: Do you have any reasoning for this number? It seems large enough that it shouldn't be concerning unless we have heavy transaction volume

ironfish/src/network/peerNetwork.ts
```diff
@@ -899,6 +898,14 @@ export class PeerNetwork {
         continue
       }
 
+      // Recently evicted means the transaction is relatively low feeRate
+      // mempool + wallet have already processed it. It could stil be download
+      // and forward to other peers but since it is a low feeRate and
+      // other peer mempools are likely at capacity, just drop it
```
dguenther, line 904:
```suggestion
      // Recently evicted means the transaction is relatively low feeRate and the
      // mempool + wallet have already processed it. It could still be downloaded
      // and forwarded to other peers, but since it is a low feeRate and
      // other peer mempools are likely at capacity, just drop it
```

ironfish/src/network/peerNetwork.ts
```diff
@@ -899,6 +898,14 @@ export class PeerNetwork {
...
+      if (this.node.memPool.recentlyEvicted(hash)) {
```
dguenther, line 905: I hadn't thought of this before, but I think this kind of assumes that most peer mempools will be the same size -- if you assume that every peer will have drastically different mempool sizes (like your node has 3MB mempool and peers have 100MB), is this still useful to check?

ironfish/src/telemetry/telemetry.ts
```diff
@@ -184,6 +184,41 @@ export class Telemetry {
         type: 'integer',
         value: this.metrics.memPoolSize.value,
       },
+      {
```
dguenther, line 187: I think this holding branch wasn't originally intended for review, so my comments here might not really be relevant. But I think in the future I'd prefer spending a little extra effort to put the core logic changes in a standalone PR into staging (behind a config if necessary). Then these telemetry, getStatus, and metrics changes could be separated out. I'd have to think more about the best way to split these out, but it does get tougher to provide quality review on a 1000s line PR, plus some of these issues may have already been commented on in the originating PRs.

ironfish/src/memPool/recentlyEvictedCache.ts
```diff
@@ -0,0 +1,218 @@
...
+  /**
+   * A priority queue of transaction hashes ordered by `feeRate` descending.
+   * This is the eviction order for the cache when full.
+   */
+  private readonly evictionQueue: PriorityQueue<EvictionQueueEntry>
+
+  /**
+   * A priority queue of transaction hashes ordered by their expiration sequence.
+   */
+  // TODO(holahula): this can just be a normal queue because if y is inserted after x, y cannot expire before x
```
dguenther, line 49: Is this true? If it's referencing the expiration sequence on the transaction, that's a user-set value, and there's no guarantee transactions will have the same interval (in other words, some users might set an expiration sequence 30 blocks out, and some users might set an expiration sequence 10 blocks out)

ironfish/src/memPool/recentlyEvictedCache.ts
```diff
@@ -0,0 +1,218 @@
...
+/**
+ * A cache to track transactions that have recently been evicted from the mempool.
+ *
+ * This is primarily a mechanism to prevent transactions from being re-added to the mempool
+ * within a reasonable duration if for some reason they are evicted (i.e. the mempool is full
+ * and the transaction is underpriced).
+ *
+ * When full, transactions are evicted in order of `feeRate` descending to make room for the
+ * new transaction. This order is chosen because these transactions most likely to be successfully
+ * re-introduced into the mempool
```
dguenther, line 28: This is a little unclear -- I'd say something like "When the RecentlyEvictedCache is full, the transaction with the lowest feeRate is removed to make room for the new transaction. This transaction is chosen because it's least likely to be re-introduced into the mempool"

ironfish/src/memPool/recentlyEvictedCache.ts
```diff
@@ -0,0 +1,218 @@
...
+   */
+  // TODO(holahula): this can just be a normal queue because if y is inserted after x, y cannot expire before x
+  private readonly expireAtQueue: PriorityQueue<ExpireAtQueueEntry>
+
+  /**
+   * Creates a new RecentlyEvictedCache.
+   * Transactions that are evicted from the mempool should be added here.
+   *
+   * @constructor
+   * @param options.capacity the maximum number of hashes to store in the cache
```
dguenther, line 57:
```suggestion
   * @param options.maxSize the maximum number of hashes to store in the cache
```

ironfish/src/memPool/recentlyEvictedCache.ts
holahula: It references the current block + maximum time a txn can spend in the cache. Usage of `expiration` in the comment is a bit ambigous as to whether it's expiration from mempool / evicted cache so I'll change that

ironfish/src/memPool/recentlyEvictedCache.ts
dguenther: Okay, I think my understanding of this was wrong. It should be "When the RecentlyEvictedCache is full, the transaction with the highest feeRate is removed to make room for the new transaction. This transaction is most likely to make it into the mempool, so by removing it, we allow it to be re-fetched."

ironfish/src/memPool/recentlyEvictedCache.ts
dguenther: Resolving, I think I misunderstood that this isn't the expiration sequence set on the transaction, it's actually a sequence generated by the RecentlyEvictedCache.

ironfish/src/rpc/routes/node/getStatus.ts
dguenther: Is this done? It's not clear to me why this is in the yup schema but not in the TS GetNodeStatusResponse type.

ironfish/src/memPool/recentlyEvictedCache.ts
```diff
@@ -0,0 +1,218 @@
...
+   */
+  constructor(options: { logger: Logger; metrics: MetricsMonitor; maxSize: number }) {
+    this.maxSize = options.maxSize
+    this.metrics = options.metrics
+    this.logger = options.logger.withTag('RecentlyEvictedCache')
+
+    this.metrics.memPool_RecentlyEvictedCache_MaxSize.value = this.maxSize
+
+    this.evictionQueue = new PriorityQueue<EvictionQueueEntry>(
+      (t1, t2) => t1.feeRate > t2.feeRate,
```
dguenther, line 67: should this use `mempoolEntryComparator`?

ironfish/src/memPool/recentlyEvictedCache.ts
```diff
@@ -0,0 +1,218 @@
...
+  has(hash: string): boolean {
+    return this.expireAtQueue.has(hash)
+  }
+
+  /**
+   * Flushes the cache of any transactions that will expire after adding the given block sequence.
+   *
+   * @param maxSequence All transactions with an expiration sequence <= `maxSequence` will be flushed.
+   */
+  flush(maxSequence: number): void {
```
dguenther, line 190: maybe just call this `sequence` to avoid terminology confusion around max vs min sequence

ironfish/src/memPool/memPool.ts
```diff
@@ -180,47 +231,47 @@ export class MemPool {
...
-        Assert.isNotUndefined(existingTransactionHash)
-
-        const existingTransaction = this.transactions.get(existingTransactionHash)
-        if (!existingTransaction) {
-          continue
-        }
+      const existingHash = this.nullifiers.get(spend.nullifier)
+      const existingTransaction = existingHash && this.transactions.get(existingHash)
 
+      if (existingTransaction) {
```
dguenther, line 274: I think I did prefer flattening this with a `continue` rather than changing it to an `if`. Also definitely helps to keep refactors out of these PRs

ironfish/src/rpc/routes/node/getStatus.ts
holahula: yes, removed

ironfish/src/fileStores/config.ts
danield9tqh: Yea copying some reasoning from the doc

* 60 MB of RAM is a reasonably small amount for a personal machine to allocate for a node
* Default transaction expiration in our wallet is 15 blocks. With 30 blocks worth of transactions any default transaction that gets evicted from the mempool would not have made it into a block anyway
* 60Mb is about 60,000 single input/output transactions. When testing mempool changes in Phase 2, a node running on a M1 MacBook Pro was able to comfortably hold more than 60k transactions in the mempool (even >100) and remain synced
* Setting the limit too high is preferable to setting it too low. If it’s set too low then failure would look like transactions getting lost in the network which is a hard problem to pin down. If it is set too high then failure would be nodes running out of memory which is much easier to pin down whether it’s related to their mempool size or not

ironfish/src/memPool/recentlyEvictedCache.ts
holahula: changed

ironfish/src/memPool/recentlyEvictedCache.ts
holahula: fixed

ironfish/src/memPool/recentlyEvictedCache.ts
holahula: reworded

ironfish/src/memPool/recentlyEvictedCache.ts
```diff
@@ -0,0 +1,218 @@
...
+  public readonly maxSize: number
+
+  /**
+   * A priority queue of transaction hashes ordered by `feeRate` descending.
+   * This is the eviction order for the cache when full.
+   */
+  private readonly evictionQueue: PriorityQueue<EvictionQueueEntry>
+
+  /**
+   * A priority queue of transaction hashes ordered by their expiration sequence.
```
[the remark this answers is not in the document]
holahula: renamed from `expireAtQueue` to `removeAtQueue` to remove ambiguity from a transaction's expiration from the mempool sequence

ironfish/src/memPool/recentlyEvictedCache.ts
danield9tqh: yea good call

ironfish/src/memPool/memPool.ts
danield9tqh: yea will change it back

ironfish/src/telemetry/telemetry.ts
danield9tqh: Yea agreed telemetry and getStatus changes could have gone in after this

ironfish/src/fileStores/config.ts
danield9tqh: I remember we had this discussion and decided to put this config value in terms of straight bytes instead of blocks so it's makes more sense to users (they can see a real byte value). I will remove this comment connecting it to block size

ironfish/src/network/peerNetwork.ts
danield9tqh: Yea agreed this could mean that nodes with smaller mempools forward fewer transactions. I increased the min maxMempoolSizeBytes to 20Mb which should alleviate some of this. Also at this point we have already broadcasted it once to all the peers we had when it was originally added to the mempool so this would just be downloading to potentially broadcast it to newly added peers. Given that it is also evicted value of downloading doesn't seem worth it. I'll update this comment to also include the fact that we've already broadcasted it once

dguenther (approved):
Thanks for thinking through this with me, I don't have any other comments 👍

PR #3523, opened 2023-02-23
feat(ironfish): Fix double spend issue in transaction expiration

rohanjadvani:
## Summary

This is a series of previously approved pull requests:
* #3518 - Create db store
* #3519 - Update db store on pending transactions, connected transactions, expired transactions
* #3520 - Check transaction hash in expiration before marking note as unspent
* #3522 - Backfill nullifier to transaction hash mapping

## Testing Plan

Manually run and unit tests

## Breaking Change

```
[ ] Yes
```


hughy (approved):

ironfish/src/wallet/account.ts
```diff
@@ -714,15 +736,25 @@ export class Account {
...
+              noteHash,
+              {
+                ...decryptedNote,
+                spent: false,
+              },
+              tx,
+            )
+          }
+
           await this.walletDb.addUnspentNoteHash(this, noteHash, decryptedNote, tx)
```
hughy, line 758:
```suggestion

            await this.walletDb.addUnspentNoteHash(this, noteHash, decryptedNote, tx)
          }
```

The `addUnspentNoteHash` call was added in `available-balances`. We should only add the note back into the unspent notes index if we're expiring the only transaction that spent it.

rohanjadvani: Fixed! Good catch

NullSoldier:
I think this PR looks good, but I want to document something I found that I also talked to Rohan about. I still think there is more work to do there and that this is a good but not complete solution.

Because we're only storing the pending transaction if it's not here, this only protects against 1 level of double spend.

1. Create pending transaction C that spends nullifier N
2. nullifier N is marked as spent and added to the index (N -> A)
3. Bump fee and resend transaction as transaction B
2. nullifier N is marked as spent but the index is not updated (N -> A)
3. Transaction A expires
4. nullifier N is unmarked as spent because (N -> A === A)
5. nullifier N is not incorrectly showing as unspent even though its in transaction B

I think the solution would be to have this store for every transaction, and only unmark it as spent in the future when all transactions that mark it as spent are removed.

NullSoldier (approved):

NullSoldier (approved):

PR #3524, opened 2023-02-23
Add max size to mempool

danield9tqh:
## Summary
Currently, there is no max size in the mempool. This means the mempool can grow indefinitely and is open to flooding attacks. A capacity will be implemented so that when reached, lowest-priority transactions will be evicted. The priority of a transaction is determined by its feeRate. We will also store a cache of recently evicted transactions as optimization for not re-processing transactions that have recently been evicted. For a recently-evicted transaction to be re-inserted into the mempool, it must increase its feeRate beyond the current mempool’s minimum transaction feeRate. 

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/fileStores/config.ts
+++ b/ironfish/src/fileStores/config.ts
@@ -17,6 +17,8 @@ export const DEFAULT_POOL_HOST = '::'
 export const DEFAULT_POOL_PORT = 9034
 export const DEFAULT_NETWORK_ID = 0
 
+const MEGABYTES = 1000 * 1000
+
 export type ConfigOptions = {
   bootstrapNodes: string[]
   databaseMigrate: boolean
@@ -235,8 +237,8 @@ export type ConfigOptions = {
   memPoolMaxSizeBytes: number
 
   /**
-   * Limit how many entries the mempool's recently evicted caches stores. This cache is used
-   * to keep track of transaction hashes recently evicted from the mempool because mempool exceeds mempoolMaxSizeBytes
+   * Limit how many entries the mempool's recently evicted caches stores. This cache is used to keep
+   * track of transaction hashes recently evicted from the mempool after it exceeds mempoolMaxSizeBytes
    */
   memPoolRecentlyEvictedCacheSize: number
 
@@ -302,7 +304,10 @@ export const ConfigOptionsSchema: yup.ObjectSchema<Partial<ConfigOptions>> = yup
     networkId: yup.number().integer().min(0),
     customNetwork: yup.string().trim(),
     maxSyncedAgeBlocks: yup.number().integer().min(0),
-    mempoolMaxSizeBytes: yup.number().integer(),
+    mempoolMaxSizeBytes: yup
+      .number()
+      .integer()
+      .min(2 * MEGABYTES), // one blocks worth of transactions
     memPoolRecentlyEvictedCacheSize: yup.number().integer(),
     networkDefinitionPath: yup.string().trim(),
   })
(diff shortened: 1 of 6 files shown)


ironfish/src/memPool/memPool.ts
```diff
@@ -46,22 +62,25 @@ export class MemPool {
...
+    this.maxSizeBytes = options.maxSizeBytes
+    this.consensus = options.consensus
     this.head = null
 
-    this.queue = new PriorityQueue<MempoolEntry>(
-      (firstTransaction, secondTransaction) => {
-        if (firstTransaction.feeRate !== secondTransaction.feeRate) {
-          return firstTransaction.feeRate > secondTransaction.feeRate
-        }
+    this.queue = new PriorityQueue<MempoolEntry>(compareMempoolEntries, (t) =>
```
holahula, line 78: as there are multiple queues now, can `this.queue` be named into something more indicative of what it does?i.e. `this.toBlockQueue` or something

ironfish/src/memPool/memPool.ts
```diff
@@ -25,17 +25,33 @@ interface ExpirationMempoolEntry {
   hash: TransactionHash
 }
 
+export function compareMempoolEntries(
```
holahula, line 28: nit: 
```suggestion
export function mempoolEntryComparator(
```

ironfish/src/memPool/memPool.ts
```diff
@@ -239,18 +270,79 @@ export class MemPool {
...
+   */
+  private evictTransactions(): Transaction[] {
+    const evictedTransactions: Transaction[] = []
+
+    while (this.full()) {
+      const next = this.evictionQueue.peek()
+
+      const transaction = next && this.transactions.get(next.hash)
+      if (!transaction) {
+        break
```
holahula, line 320:
```suggestion
        // need to remove the txn from the eviction queue
        continue
```
1. imo continue > break if the transaction cannot be found for some reason we should continue to evict
2. need to remove this faulty txn from the eviction queue as otherwise on the next call of `evictTransactions` nothing will be evicted as `queue.peek()` will return the same obj at the front of the pq

ironfish/src/memPool/memPool.test.ts
```diff
@@ -546,4 +568,109 @@ describe('MemPool', () => {
...
+  return [...transactions].sort((t1, t2) => {
+    const greater = compareMempoolEntries(
+      { hash: t1.hash(), feeRate: getFeeRate(t1) },
+      { hash: t2.hash(), feeRate: getFeeRate(t2) },
+    )
+    return greater ? -1 : 1
+  })
+}
+
+// return the first transactions that fit within the target byte size
```
holahula, line 656: update comment to reflect return arrays changing

ironfish/src/fileStores/config.ts
```diff
@@ -374,6 +388,8 @@ export class Config extends KeyStore<ConfigOptions> {
       networkId: DEFAULT_NETWORK_ID,
       customNetwork: '',
       maxSyncedAgeBlocks: 60,
+      memPoolMaxSizeBytes: 60000000,
+      memPoolRecentlyEvictedCacheSize: 60000,
```
holahula, line 392: as mempool size is determined in bytes, should the cache be as well? then can add a downstream `sizeInTxns` that is determined from the size in bytes

danield9tqh: Yea could be a good idea. I don't wanna do it right now though since it complicates things and don't anticipate people really changing this config setting

ironfish/src/memPool/memPool.ts
danield9tqh: Yea I like that. Changing to `feeRateQueue`

ironfish/src/memPool/memPool.ts
danield9tqh: Discussed, we'll change this to `Assert` that transaction is always defined. This should never be undefined because if the mempool is full there should always be another transaction present

holahula (approved):

holahula:
lgtm

PR #3525, opened 2023-02-23
Change send to use new CLI prompts

NullSoldier:
## Summary

This now changes cli send to use the new fee and currency prompts. It also simplifies the logic a lot.

## Testing Plan

Create some transactions using it.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/send.ts
+++ b/ironfish-cli/src/commands/wallet/send.ts
@@ -240,7 +240,7 @@ export class Send extends IronfishCommand {
         .toString('hex')}`,
     )
 
-    if (flags.watch && params) {
+    if (flags.watch) {
       this.log('')
 
       await watchTransaction({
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/commands/wallet/send.ts
```diff
@@ -90,240 +93,154 @@ export class Send extends IronfishCommand {
...
-      if (error instanceof Error) {
-        this.error(error.message)
-      }
-      this.exit(2)
-    }
+        .toString('hex')}`,
+    )
 
-    if (flags.watch && transaction) {
+    if (flags.watch && params) {
```
hughy, line 243: nit - do we need `&& params`? Won't it always be truthy?

hughy (approved):

ironfish-cli/src/commands/wallet/send.ts
NullSoldier: fixed

PR #3530, opened 2023-02-23
changes mint to use new utils

hughy:
## Summary

updates the mint command to use new utils. uses 'selectFee' and 'promptCurrency'.

simplifies some of the logical flow and messages to be more consistent with the send command where applicable (e.g., confirming before post)

adds 'offline' and 'watch' flags to match send

matches changes to the send command in #3525

## Testing Plan

ran 'mint' with a few different options, e.g.,
image

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/mint.ts
+++ b/ironfish-cli/src/commands/wallet/mint.ts
@@ -97,7 +97,7 @@ export class Mint extends IronfishCommand {
 
   async start(): Promise<void> {
     const { flags } = await this.parse(Mint)
-    const client = await this.sdk.connectRpc(false, true)
+    const client = await this.sdk.connectRpc()
 
     if (flags.eligibility) {
       await doEligibilityCheck(client, this.logger)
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/wallet/mint.ts
```diff
@@ -96,44 +104,41 @@ export class Mint extends IronfishCommand {
     }
 
     const status = await client.getNodeStatus()
```
NullSoldier, line 106: this can move into the offline if statement, so we don't need to do this request.

ironfish-cli/src/commands/wallet/mint.ts
```diff
@@ -96,44 +104,41 @@ export class Mint extends IronfishCommand {
...
+    if (!flags.offline) {
+      if (!status.content.blockchain.synced) {
+        this.log(
+          `Your node must be synced with the Iron Fish network to send a transaction. Please try again later`,
+        )
+        this.exit(1)
+      }
     }
 
     let account = flags.account?.trim()
```
NullSoldier, line 116: After some testing, we don't need to trim flags. It looks like heroku does that automatically

NullSoldier (approved):

PR #3531, opened 2023-02-23
Add subprotocols to identify message

dguenther:
## Summary

Adds supported sub-protocols to the identify message. This will allow nodes to indicate which sub-protocols they support in the identify message, as well as allowing independent development of sub-protocol versions. Sub-protocols contain an ID and a version.

## Testing Plan

Tested 0.1.68 connecting to this branch to ensure it doesn't break backward compatibility in the Identify message.

## Breaking Change

```
[ ] Yes
```


mat-if (approved):
Ran multiple nodes on this version. Also ran it on the testnet - had peers as far back as 0.1.63 👍

dguenther:
Closing for #3551

PR #3532, opened 2023-02-23
Replace message type with sub-protocol and message type

dguenther:
## Summary

Given we'll be restructuring messages to be part of various sub-protocols, it would be helpful to include the sub-protocol ID along with the message ID in the network message header. At the cost of an additional byte per message, this will allow us to have independent message IDs per-protocol -- when using a single global message ID, the messages from different sub-protocols can be mixed (for example, messages from sync might be 1 and 2, then messages from peer discovery might be 3 and 4, followed by messages from sync at 5 and 6).

It'll be nice to have this backward-compatible to test on the testnet prior to mainnet launch. So in this branch, if we detect a node on a newer network version, we'll send both the sub-protocol and message ID, otherwise for older nodes, we'll just send the old message ID. We'll plan to make a breaking change to remove this code and add sub-protocol + message ID to the Identify message prior to mainnet.

## Testing Plan

Tested this with two nodes on this branch connecting to each other to check that sub-protocols and message types are sent, and tested connecting to 0.1.68 to check that only message types (and not sub-protocols) are sent.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/messageRegistry.test.ts
+++ b/ironfish/src/network/messageRegistry.test.ts
@@ -8,7 +8,7 @@ describe('messageRegistry', () => {
   describe('parseNetworkMessage', () => {
     describe('with a malformed header', () => {
       it('throws an error', () => {
-        expect(() => parseNetworkMessage(Buffer.from(''))).toThrow()
+        expect(() => parseNetworkMessage(Buffer.from(''), true)).toThrow()
       })
     })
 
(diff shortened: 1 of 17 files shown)


mat-if (approved):

danield9tqh:
I think there are def some cons to consider adding the protocol to the message itself. For example changing protocols becomes more difficult. If we want to re-arrange messages or protocols we now have to change the messages themselves. It looks like we are already defining (in code) that a protocol is a collection of specific messages we can still do that without adding the protocol itself to the message. 

The only advantage listed is that we can have different message numbering per protocol so all the message numbers are grouped together but that seems a little arbitrary since nobody is actually looking at the raw message numbers. It's really only a serialization thing

dguenther:
Closing for #3551

PR #3533, opened 2023-02-23
[wip] feat: telemetry for mempool + recently evicted cache

holahula:
## Breaking Change

```
[ ] Yes
```


ironfish/src/memPool/memPool.test.ts
```diff
@@ -22,7 +21,7 @@ import { mempoolEntryComparator } from './memPool'
 // but not the mempool
 async function createTransactions(
   node: IronfishNode,
-  from: SpendingAccount,
+  from: Account,
```
danield9tqh, line 24: can we rebase and get rid of this somehow?

ironfish/src/memPool/memPool.ts
```diff
@@ -115,17 +116,53 @@ export class MemPool {
...
   }
 
+  /**
+   *
+   * @returns The maximum number of transactions that can be stored in the mempool.
+   *
+   * This assumes that all transactions are the maximum size, and so the true
+   * maximum in practice can be higher.
+   */
+  maxCount(): number {
```
danield9tqh, line 134: Are we using this anywhere?

ironfish/src/memPool/memPool.ts
```diff
@@ -115,17 +116,53 @@ export class MemPool {
...
    */
   sizeBytes(): number {
     return this.transactionsBytes
   }
 
+  /**
+   *
+   * @returns The maximum number of bytes that can be stored in the mempool
+   */
+  maxSizeBytes(): number {
```
danield9tqh, line 154: maybe we can just make the `maxBytes` field public. This seems kinda verbose

ironfish/src/memPool/memPool.test.ts
holahula: yes, i'll fix this issue in the dst branch

ironfish/src/memPool/memPool.ts
holahula: this is used for for the max capacity and saturation recently evicted cache metrics

ironfish/src/memPool/memPool.ts
holahula: is there a standard between private field + public `get()` function vs public field?

danield9tqh: don't think we have a standard but less boilerplate is always nice

ironfish/src/memPool/recentlyEvictedCache.ts
```diff
@@ -117,6 +117,22 @@ export class RecentlyEvictedCache {
...
+   */
+  saturation(): number {
+    return this.size() / this._capacity
+  }
+
+  /**
+   *
+   * @returns the maximum capacity of the cache, in number of transactions
+   */
+  maxSize(): number {
```
danield9tqh, line 132: Could just make this public also

danield9tqh (changes requested):

ironfish/src/memPool/memPool.ts
```diff
@@ -280,12 +293,13 @@ export class MemPool {
...
-    this.metrics.memPoolSize.value = this.count()
+    this._sizeBytes += getTransactionSize(transaction)
+
+    this.updateMetrics()
 
-    // TODO: Add telemetry for evictions
     if (this.full()) {
-      this.evictTransactions()
+      const evicted = this.evictTransactions()
+      this.metrics.memPoolEvictions.add(evicted.length)
```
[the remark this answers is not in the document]
holahula: added a check at the start of `addTransaction` so that txns in the recently evicted cache are rejected at the start of the function

danield9tqh (approved):

PR #3534, opened 2023-02-23
Update CLI import prompt

EvanJRichard:
## Summary
`wallet:import` doesn't just accept the output of `wallet:export`, it'll also accept a raw spending key. Update the prompt to reflect this.

## Testing Plan
Cosmetic prompt change, just leaning on unit tests.

## Breaking Change

```
[ ] Yes
[X] No
```


rohanjadvani (approved):
lgtm, can we update the description of the `blob` arg too?

PR #3538, opened 2023-02-24
Fix mempool transaction expiry bug

danield9tqh:
## Summary
We keep a PriorityQueue of transactions to expire when a new block is connected. They are ordered by lowest expiration sequence. However we also use an expiration sequence of `0` to indicate a transaction that never expires. These 0 expiration transactions were going to the top of the queue and when checking to see if the next transaction should be expired it would look at a 0 expiration and say that the transaction does not need to be expired. This was blocking other transactions that actually did need to be expired.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/memPool/memPool.ts
+++ b/ironfish/src/memPool/memPool.ts
@@ -63,7 +63,7 @@ export class MemPool {
     )
 
     this.expirationQueue = new PriorityQueue<ExpirationMempoolEntry>(
-      (t1, t2) => t1.expiration < t2.expiration,
+      (t1, t2) => (t1.expiration === 0 ? false : t1.expiration < t2.expiration),
       (t) => t.hash.toString('hex'),
     )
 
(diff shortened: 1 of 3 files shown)


ironfish/src/memPool/memPool.ts
```diff
@@ -63,7 +63,7 @@ export class MemPool {
     )
 
     this.expirationQueue = new PriorityQueue<ExpirationMempoolEntry>(
-      (t1, t2) => t1.expiration < t2.expiration,
+      (t1, t2) => (t1.expiration === 0 ? false : t1.expiration < t2.expiration),
```
hughy, line 66: Should we add transactions with an expiration of 0 to this queue at all?

hughy (approved):
good find!

ironfish/src/memPool/memPool.ts
danield9tqh: Yea that's a good point. I think you're right we can safely just not add them to the queue

PR #3544, opened 2023-02-24
chore: upgrade ws websocket library

mat-if:
## Summary

- Upgrade `ws` to the latest version
- Remove the dependency from CLI since it isn't being used
- Only big change is that "WebSocket was closed before the connection was established" error no longer raises an exception, but fires an onError event in the _next tick_ (https://github.com/websockets/ws/commit/d086f4bcbbe235f12f6fa2ddba5a8ce1342dac58) which means we can't easily remove the onerror handler. If we still want to remove the handler, we will likely have to do it in a `process.nextTick` call.

If you wish to confirm that the removal of the try-catch works, adding a nice obvious log to the onerror handler when the state is disconnected. Within about 2-5 minutes, you'll see the "WebSocket was closed before the connection was established" error being logged and no exception getting raised.

## Testing Plan

Manually tested 2 nodes on this version, as well as running live on the testnet for several hours.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/package.json
+++ b/ironfish-cli/package.json
@@ -28,7 +28,6 @@
     "@types/inquirer": "8.2.5",
     "@types/node": "18.11.16",
     "@types/tar": "6.1.1",
-    "@types/ws": "7.4.5",
     "chai": "4.2.0",
     "cross-env": "7.0.3",
     "eslint-config-ironfish": "*",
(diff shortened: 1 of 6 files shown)


dguenther:
I think we might want to call terminate on all of the clients when the server closes now:

https://github.com/websockets/ws/releases/tag/8.0.0

Otherwise this looks good. The only other idea I have in mind for the event unbinding would be maybe the server can unbind the error event in a handler for the close event, but I don't think it's too important 🤔

PR #3545, opened 2023-02-24
feat: replace segfault-handler

mat-if:
## Summary

segfault-handler is a useful library, but requires a lot of extra
dependencies for our users since it compiles binaries on installation.
We can replace this with some napi rust functions that give us most of
what we need. This can later be extended to cover all the cases we want
it to, or split out into its own package.

### TODOs
- [x] Ensure this catches SEGFAULTs triggered by other code.
  - [x] Catches NAPI segfaults
  - [x] Catches an unrelated C/C++ NAPI module segfault? See https://github.com/iron-fish/ironfish/tree/mat/segfault-testing for more

    - https://user-images.githubusercontent.com/97762857/222219291-74da43c8-8f09-4130-bdd4-3615ae863e34.png

- [x] Ensure catching SIGTRAPs won't cause unintended issues (have ran a node on testnet for a couple hours)
- [x] Clean up the code
- [x] Actually remove segfault-handler dependency
- [x] Get better backtraces by using libc's backtrace instead of Rust.

## Testing Plan

Manually tested:
- Windows: only needed to install nodejs, yarn, rust. Built and started a node
- NixOS: Ran in a pure shell with only nodejs, yarn, rust. Built and started a node
- M1 Mac: Started and ran a node on testnet for several hours

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/bin/run
+++ b/ironfish-cli/bin/run
@@ -1,18 +1,21 @@
 #!/usr/bin/env node
 
-// segfault-handler causes crashes in node-webrtc on Windows because
-// it catches all exceptions rather than only exceptions that would
-// crash the process
+// the signal handler does not support windows yet
 if (process.platform !== 'win32') {
-  require('segfault-handler').registerHandler('segfault.log')
+  require('@ironfish/rust-nodejs').initSignalHandler()
 }
 
-if(process.versions.node.split('.')[0] !== '18') {
-  console.log('NodeJS version ' +  process.versions.node + ' is not compatible. Must have Node v18 installed: https://nodejs.org/en/download/')
-  console.log('After v18 is installed, MAKE SURE TO run `npm install -g ironfish` again to install ironfish with the correct Node version')
+if (process.versions.node.split('.')[0] !== '18') {
+  console.log(
+    `NodeJS version ${process.versions.node} is not compatible. Must have Node v18 installed: https://nodejs.org/en/download/`,
+  )
+  console.log(
+    'After v18 is installed, MAKE SURE TO run `npm install -g ironfish` again to install ironfish with the correct Node version',
+  )
   process.exit(1)
 }
 
-require('@oclif/core').run()
-.then(require('@oclif/core/flush'))
-.catch(require('@oclif/core/handle'))
+require('@oclif/core')
+  .run()
+  .then(require('@oclif/core/flush'))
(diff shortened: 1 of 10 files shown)


ironfish/src/sdk.ts
```diff
@@ -183,6 +184,8 @@ export class IronfishSdk {
     autoSeed?: boolean
     privateIdentity?: PrivateIdentity
   } = {}): Promise<IronfishNode> {
+    // triggerSeg()
```
jowparks, line 187: left over from testing?

dguenther:
> Ensure this catches SEGFAULTs trigger by other code

This is probably the main bit it'd be nice to check. It'd be much cleaner if this works decently well. It'd also be nice to double-check that this works on Windows, since I suspect a decent number of users will be on there, and the current version of segfault-handler seems to work: https://github.com/ddopson/node-segfault-handler

mat-if:
> > Ensure this catches SEGFAULTs trigger by other code
> 
> This is probably the main bit it'd be nice to check. It'd be much cleaner if this works decently well. It'd also be nice to double-check that this works on Windows, since I suspect a decent number of users will be on there, and the current version of segfault-handler seems to work: https://github.com/ddopson/node-segfault-handler

I'm confident that this currently wouldn't work for Windows, but we already weren't using segfault-handler in Windows: https://github.com/iron-fish/ironfish/blob/staging/ironfish-cli/bin/run#L3-L6

Once I've reached ~feature parity with segfault-handler, I would like to investigate Windows usage

ironfish-cli/bin/run
```diff
@@ -1,18 +1,21 @@
...
+// the signal handler does not support windows yet
 if (process.platform !== 'win32') {
-  require('segfault-handler').registerHandler('segfault.log')
+  require('@ironfish/rust-nodejs').initSignalHandler()
 }
 
-if(process.versions.node.split('.')[0] !== '18') {
-  console.log('NodeJS version ' +  process.versions.node + ' is not compatible. Must have Node v18 installed: https://nodejs.org/en/download/')
-  console.log('After v18 is installed, MAKE SURE TO run `npm install -g ironfish` again to install ironfish with the correct Node version')
+if (process.versions.node.split('.')[0] !== '18') {
```
mat-if, line 8: auto formatter 🤷

dguenther (approved):

Issue #3549, opened 2023-02-24
Feature: Display Error Message when burning, minting, sending transaction


irontankfish:
### Description

Currently when sending, minting or burning a transaction, no Error message is displayed to let the user know, that the amount that they entered is 0, or less than 0. An output message, would be useful:

Burning:
<img width="888" alt="before-burn" src="https://user-images.githubusercontent.com/114656810/221302362-70e490cb-6fb0-4d51-a19a-ff29a9663c1f.png">

Minting:
<img width="894" alt="before-mint" src="https://user-images.githubusercontent.com/114656810/221302365-32e992c5-41a0-4f24-857e-b3c0fc1c34c4.png">

Sending:
<img width="886" alt="before-send" src="https://user-images.githubusercontent.com/114656810/221302369-57eb8679-036a-427d-841a-d9875b7f70e2.png">

hughy:
Resolved by #3567

PR #3550, opened 2023-02-24
Output error message when transaction amount is not greater than 0

irontankfish:
## Summary
Output error message: "Please enter an amount greater than 0", when transaction amount is not greater than 0
Fixes: #3549 

## Testin
After burn:
after-burn

After mint:
after-mint

After send:
after-send
g Plan

## Breaking Change

```
[ X] No
```

### DIFF
--- a/ironfish-cli/src/utils/currency.ts
+++ b/ironfish-cli/src/utils/currency.ts
@@ -5,6 +5,8 @@
 import { Asset } from '@ironfish/rust-nodejs'
 import { Assert, CurrencyUtils, RpcClient } from '@ironfish/sdk'
 import { CliUx } from '@oclif/core'
+import {createRootLogger, Logger} from '@ironfish/sdk'
+
 
 export async function promptCurrency(options: {
   client: RpcClient
@@ -16,6 +18,7 @@ export async function promptCurrency(options: {
     assetId?: string
     confirmations?: number
   }
+  logger?: Logger
 }): Promise<bigint>
 
 export async function promptCurrency(options: {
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/utils/currency.ts
```diff
@@ -59,6 +59,7 @@ export async function promptCurrency(options: {
     Assert.isNotNull(amount)
 
     if (options.minimum != null && amount < options.minimum) {
+      console.log("Please enter an amount greater than 0")
```
hughy, line 62: This message won't always be correct: what if `options.minimum` isn't equal to 1?

We also prefer not to use `console.log`.

I think that it might be better to update `text` to include the minimum (if any), similar to what we do for the balance on line 40.

NullSoldier: Yes we shouldn't use console.log. You need to pass a logger in, and use that.

I'm ok if we don't update the text, I did that in my original Pr and it felt like too much information for an error case that most users won't hit.

irontankfish: Perfect! I used a logger, tested it, and pushed a new commit

NullSoldier (approved):

NullSoldier:
I took this code and did it for minimum, and other types of errors. We'll still give you credit but I'm going to close this PR. 

https://github.com/iron-fish/ironfish/pull/3567

NullSoldier:
@irontankfish What graffiti do you want this credited to?

irontankfish:
@NullSoldier thanks for reaching out, and offering me credit. That is very nice of you guys. It should be under the same name: irontankfish

PR #3551, opened 2023-02-24
Add Features field to Identify message

dguenther:
## Summary

Adds a features field to the Identify message. For now, it allows for broadcasting to other peers whether your node supports syncing.

## Testing Plan

* [x] Can sync from 0.1.69 to this branch
* [x] Can sync from this branch to this branch
* [x] Does not gossip from this branch to this branch (with enableSyncing false)

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/peers/localPeer.ts
+++ b/ironfish/src/network/peers/localPeer.ts
@@ -26,6 +26,8 @@ export class LocalPeer {
   readonly webSocket: IsomorphicWebSocketConstructor
   // the unique ID number of the network
   readonly networkId: number
+  // true if the peer supports syncing and gossip messages
+  readonly enableSyncing: boolean
 
   // optional port the local peer is listening on
   port: number | null
@@ -41,6 +43,7 @@ export class LocalPeer {
     chain: Blockchain,
     webSocket: IsomorphicWebSocketConstructor,
     networkId: number,
+    enableSyncing: boolean,
   ) {
     this.privateIdentity = identity
     this.publicIdentity = privateIdentityToIdentity(identity)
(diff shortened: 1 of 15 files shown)


ironfish/src/network/peers/localPeer.ts
```diff
@@ -71,6 +75,9 @@ export class LocalPeer {
       work: this.chain.head.work,
       networkId: this.networkId,
       genesisBlockHash: this.chain.genesis.hash,
+      features: {
+        enableSyncing: this.enableSyncing,
```
NullSoldier, line 79: would be nice to name this syncing

ironfish/src/network/peers/peerFeatures.ts
```diff
@@ -0,0 +1,13 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+export const FEATURES_MIN_VERSION = 21
+
+export interface Features {
+  enableSyncing: boolean
```
NullSoldier, line 8: would be nice to name this syncing

NullSoldier (dismissed):

NullSoldier:
This looks good. I pulled it down and tested it and it doesn't run atm. I originally approved it because the overall approach is good. Once we can test this works, I'm good to merge.

ironfish/src/network/peers/peerFeatures.ts
dguenther: renamed this

NullSoldier (approved):

NullSoldier:
Looks good, im gonna run this overnight to test.

dguenther:
Tested that when this branch is connected to another node on this branch with enableSyncing set to false, that node does not receive any gossip messages. Also tested that when enableSyncing is true, it's possible to sync between two nodes on this branch.

PR #3558, opened 2023-02-25
Remove IncomingPeerMessage

dguenther:
## Summary

This abstraction isn't really useful anymore since we pass the peer around with the message, so we can remove it and just emit the message directly.

## Testing Plan

Tests should pass.

## Breaking Change

```
[ ] Yes
```


mat-if (approved):
🎉

PR #3559, opened 2023-02-25
Display features in the peers list

dguenther:
## Summary

#3551 added features to the Identify message. This PR adds a `--features` flag to the `peers` command that displays what features a node has advertised in its Identify message.

## Testing Plan

Ran the command on a node with the syncing feature enabled and disabled to check that it appears (and doesn't) in the FEATURES column.

## Breaking Change

```
[ ] Yes
```


dguenther:
image

mat-if (approved):

PR #3560, opened 2023-02-25
Trim file contents when importing an account

NullSoldier:
## Summary

Many editors put new lines at the end of file, and any file with a newline that contains bech32 won't be importable.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/import.ts
+++ b/ironfish-cli/src/commands/wallet/import.ts
@@ -127,7 +127,7 @@ export class ImportCommand extends IronfishCommand {
   async importFile(path: string): Promise<AccountImport> {
     const resolved = this.sdk.fileSystem.resolve(path)
     const data = await this.sdk.fileSystem.readFile(resolved)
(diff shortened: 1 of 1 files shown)


ygao76 (approved):

rohanjadvani (approved):

ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -127,7 +127,7 @@ export class ImportCommand extends IronfishCommand {
   async importFile(path: string): Promise<AccountImport> {
     const resolved = this.sdk.fileSystem.resolve(path)
     const data = await this.sdk.fileSystem.readFile(resolved)
-    return this.stringToAccountImport(data)
```
lwisne, line 130: suggest adding a test for this case

NullSoldier: Seems like a good idea, once we unblock our users we can look into how to not only write tests, but make this code handle old schemas in a real way.

PR #3561, opened 2023-02-25
Fix bech32 import for accounts without view key

NullSoldier:
## Summary

This is breaking import of bech32 accounts without a view key. Those keys will fail YUP validation during import.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/import.ts
+++ b/ironfish-cli/src/commands/wallet/import.ts
@@ -84,7 +84,7 @@ export class ImportCommand extends IronfishCommand {
       }
     }
     CliUx.ux.error(
-      `Detected mnemonic input, but the import failed. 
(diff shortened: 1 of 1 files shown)


rohanjadvani (approved):

ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -101,7 +101,16 @@ export class ImportCommand extends IronfishCommand {
     // bech32 encoded json
     const [decoded, _] = Bech32m.decode(data)
     if (decoded) {
-      return JSONUtils.parse<AccountImport>(decoded)
```
lwisne, line 104: add test case

jowparks: We have 0 tests in the cli

PR #3565, opened 2023-02-25
Change promptCurrency to accept iron not ore

NullSoldier:
## Summary

Users were typing iron in here, but this prompt accepted ore. Changes this to accept and decode iron now.

## Breaking Change

```
[ ] Yes
```


rohanjadvani (approved):
Lgtm after lint passes

PR #3567, opened 2023-02-26
Show nicer parsing errors in promptCurrency

NullSoldier:
## Summary

This will help parse the untyped errors from bignumber and add a new function to try and display them more cleanly.

It also now logs the reason for prompt errors. 

```bash
$ fishy wallet:send

Enter the amount (balance 0.00000000): 0.000000000005
Error: fractional component exceeds 8 decimals
Enter the amount (balance 0.00000000): fisdfjgdsdsfds
Error: invalid decimal value
Enter the amount (balance 0.00000000): -4
Error: Minimum is 0.00000001
Enter the amount (balance 0.00000000): 4
Enter the public address of the recipient:
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/burn.ts
+++ b/ironfish-cli/src/commands/wallet/burn.ts
@@ -139,6 +139,7 @@ export class Burn extends IronfishCommand {
         required: true,
         text: 'Enter the amount of the custom asset to burn',
         minimum: 1n,
+        logger: this.logger,
         balance: {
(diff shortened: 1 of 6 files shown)


hughy (approved):
Should we also check and render an error message if the value passed in is greater than the available balance?

NullSoldier:
Hmmm let's get this in and see if users want that next

PR #3572, opened 2023-02-27
Update balance.ts

leestore:
In the "if (flags.all)" condition block, you need to add a check for the presence of a block hash and sequence, because they can be undefined.

## Breaking Change

```
[ ] Yes
```


hughy:
This does not make any functional changes to the code; only formatting changes

PR #3573, opened 2023-02-27
Update balance.ts

leestore:
In the explainBalance() function, the parameter assetId is not used. This can be removed because the value of assetId can be obtained from response.content.assetId.

## Breaking Change

```
[ ] Yes
```


hughy:
This will break the balance command when `--explain` is used because it removes the assetId argument but does not update `explainBalance`.

PR #3577, opened 2023-02-27
feat: adds support to configure the max rolling avg samples for meters

holahula:
## Summary
Adds support to configure the max # of samples tracked in a `meter.rollling_avg`

Using `rate5m` is causing issues with the telemetry graphs for `create_new_block_template` as the inclusion of 0 values when new block templates are not being created are skewing the results. Switching to `avg` so sample are only added when new block templates are being created. Max # of values tracking in the rolling avg is set to `100`

## Testing Plan
existing tests

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/metrics/meter.ts
+++ b/ironfish/src/metrics/meter.ts
@@ -5,6 +5,13 @@ import { SetIntervalToken } from '../utils'
 import { EwmAverage } from './ewmAverage'
 import { RollingAverage } from './rollingAverage'
 
+export type newMeterOptions = {
+  /**
+   * The maximum number of samples to keep in the rolling average
+   */
+  maxRollingAverageSamples?: number
+}
+
 /**
  * A metric type useful for recording metered things like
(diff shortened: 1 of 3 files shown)


ironfish/src/metrics/meter.ts
```diff
@@ -31,15 +38,17 @@ export class Meter {
   private _intervalMs: number
   private _intervalLastMs: number | null = null
 
-  constructor() {
+  constructor(options?: newMeterOptions) {
```
NullSoldier, line 41: Can we in-line this type? It makes people have to look at a separate interview just for parameters of a function

NullSoldier (approved):

ironfish/src/metrics/meter.ts
holahula: the options type is also used upstream, in `MetricsMonitor.addMeter()` should it be in-lined in both places?

PR #3580, opened 2023-02-27
Remove unnecessary checks for transactions in transaction fetcher

danield9tqh:
## Summary
We are doing some checks that are redundant in the transaction fetcher. In the middle of the fetching cycle we are making sure the transactions didn't get added to a block or added to the mempool. But if they did the transaction fetcher would have already been notified and we would've stopped downloading the transaction. So doing these checks doesn't add any benefit

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/peerNetwork.test.ts
+++ b/ironfish/src/network/peerNetwork.test.ts
@@ -40,6 +40,7 @@ import {
   PooledTransactionsResponse,
 } from './messages/pooledTransactions'
 import { PeerNetwork } from './peerNetwork'
+import { Peer } from './peers/peer'
 import {
   expectGetBlockHeadersResponseToMatch,
   expectGetBlockTransactionsResponseToMatch,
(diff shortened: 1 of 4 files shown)


dguenther (approved):

ironfish/src/network/peerNetwork.test.ts
```diff
@@ -739,7 +740,14 @@ describe('PeerNetwork', () => {
 
         const accountA = await useAccountFixture(wallet, 'accountA')
         const accountB = await useAccountFixture(wallet, 'accountB')
-        const { transaction } = await useBlockWithTx(node, accountA, accountB)
+        const { transaction } = await useBlockWithTx(
```
danield9tqh, line 743: Updating fixtures here. This test should not sync to mempool when creating this transaction

ironfish/src/network/peerNetwork.test.ts
```diff
@@ -767,7 +775,14 @@ describe('PeerNetwork', () => {
         const { wallet, memPool, chain } = node
         const accountA = await useAccountFixture(wallet, 'accountA')
         const accountB = await useAccountFixture(wallet, 'accountB')
-        const { transaction } = await useBlockWithTx(node, accountA, accountB)
+        const { transaction } = await useBlockWithTx(
```
danield9tqh, line 778: Updating fixtures here. This test should not sync to mempool when creating this transaction

ironfish/src/network/peerNetwork.test.ts
```diff
@@ -867,7 +882,14 @@ describe('PeerNetwork', () => {
 
         const accountA = await useAccountFixture(wallet, 'accountA')
         const accountB = await useAccountFixture(wallet, 'accountB')
-        const { block, transaction } = await useBlockWithTx(node, accountA, accountB)
+        const { block, transaction } = await useBlockWithTx(
```
danield9tqh, line 885: Updating fixtures here. This test should not sync to mempool when creating this transaction

PR #3583, opened 2023-02-27
feat: smaller max message size and soft message size limit

mat-if:
## Summary

This implements a much smaller MAX_MESSAGE_SIZE of 16mb. It also introduces a SOFT_MAX_MESSAGE_SIZE of 12mb. This allows us to fill messages to a reasonable size without running the risk of fetching data that we end up not including in messages.

This also introduces changes to allow GetBlocksResponse to work up to the SOFT_MAX_MESSAGE_SIZE limit while also allowing the syncer to work as expected.

## Breaking Change

```
[x] Yes
```


dguenther (approved):
Nice, I think this looks good. I'm not generally too concerned with backward compatibility here since the current message should already fit in the soft cap.

PR #3585, opened 2023-02-28
Refactor checking the mempool in PeerNetwork.onNewTransaction

danield9tqh:
## Summary
Moving some logic around to make incoming changes related to the mempool more seamless. Essentially this does an early return if we have already seen the transaction in the mempool (while still broadcasting it to un-knowing peers). 

## Breaking Change

```
[ ] Yes
```


ironfish/src/network/peerNetwork.ts
```diff
@@ -1361,15 +1358,11 @@ export class PeerNetwork {
...
-    }
-
-    // Check 'exists' rather than 'accepted' to allow for rebroadcasting to nodes that
-    // may not have seen the transaction yet
-    if (this.node.memPool.exists(transaction.hash())) {
-      this.broadcastTransaction(transaction)
-    }
+    // This transaction is valid and has not yet been seen so sync it to the
+    // mempool, gossip to all peers and emit telemetry event for new transaction seen
+    this.node.memPool.acceptTransaction(transaction)
```
danield9tqh, line 1363: Doing an early return now so at this point the transaction has not already been seen in the mempool. We've also confirmed that it is valid. If it hasn't been seen in the mempool we always want to

1. gossip
2. accept into the mempool
3. sync to the wallet
4. submit telemetry event that a new transaction was seen

ironfish/src/network/peerNetwork.ts
```diff
@@ -1361,15 +1358,11 @@ export class PeerNetwork {
       return
     }
 
-    if (this.node.memPool.acceptTransaction(transaction)) {
-      this.onTransactionAccepted.emit(transaction, received)
-    }
-
-    // Check 'exists' rather than 'accepted' to allow for rebroadcasting to nodes that
-    // may not have seen the transaction yet
-    if (this.node.memPool.exists(transaction.hash())) {
```
danield9tqh, line 1370: Not checking the result of this function anymore has the interesting result of gossiping transactions with overlapping nullifiers in the mempool (and lower fee rate). This could be a good think since not gossiping these transactions seemed opinionated about which transactions other nodes would prefer

ironfish/src/network/peerNetwork.ts
```diff
@@ -1361,13 +1358,14 @@ export class PeerNetwork {
       return
     }
 
-    if (this.node.memPool.acceptTransaction(transaction)) {
-      this.onTransactionAccepted.emit(transaction, received)
-    }
+    this.onTransactionAccepted.emit(transaction, received)
```
danield9tqh, line 1361: Not checking the result of `memPool.acceptTransaction` anymore because that was incorrect. Even if it is not accepted for some reason like it's expired or it has overlapping nullifiers with existing transactions it still should be sent to telemetry as a new transaction seen.

dguenther: What do you think about renaming to `onTransactionSeen`? The event is only used for the telemetry event anyway, and now it's emitted when a new transaction is seen rather than accepted, I suppose

danield9tqh: Yea I originally had that exact rename in this PR but removed it so it was easier to review with less changes 😆

danield9tqh: I can add it back in after you finish reviewing

ironfish/src/network/peerNetwork.ts
danield9tqh: Deciding to continue to not gossip transactions with overlapping nullifiers and lower fees. Looked into btc-core and go-eth and both seem to not send out lower fee conflicting transactions

dguenther (approved):

PR #3590, opened 2023-02-28
feat: Don't send outbound messages that exceed the max size

mat-if:
## Summary

This adds a check to all outbound messages to ensure they fit within the MAX_MESSAGE_SIZE. This ensures a node is not sending invalid messages, intentionally or not.

## Testing Plan

Unit test

## Breaking Change

```
[ ] Yes
```


ironfish/src/network/peers/connections/connection.ts
```diff
@@ -114,6 +115,47 @@ export abstract class Connection {
...
+          object.type,
+        )} message in state ${this.state.type} ${ErrorUtils.renderError(e)}`,
+      )
+      this.close(e)
+      return false
+    }
+
+    if (sendResult) {
+      this.metrics?.p2p_OutboundTraffic.add(byteCount)
+      this.metrics?.p2p_OutboundTraffic_WebRTC.add(byteCount)
```
mat-if, line 152: 🤦

ironfish/src/network/peers/connections/webRtcConnection.ts
```diff
@@ -197,42 +197,18 @@ export class WebRtcConnection extends Connection {
...
-    }
-
     if (!this.datachannel.isOpen()) {
       this.logger.debug('Datachannel no longer open, closing connection')
       this.close()
       return false
     }
 
-    const data = message.serializeWithMetadata()
-    try {
```
dguenther, line 220: I can't say for sure that this try/catch is no longer needed -- might be worth re-adding it unless you checked it out

mat-if: I hoisted it up into the send function: https://github.com/iron-fish/ironfish/pull/3590/files#diff-ed1400fde596eef381f9b8abb8faaa1965e28b481b5f87c829c95c16df06f918R139

Happy to move it back if we prefer - I don't feel strongly one way or the other, I think I moved it just to minimize the amount of differences between the `_send` functions

dguenther: Resolving this, I noticed you boosted it up to the connection level

dguenther (approved):

PR #3591, opened 2023-02-28
Remove testnet command and eligibility flag

dguenther:
## Summary

Now that the incentivized testnet is ending, we can remove the `testnet` command and the `eligibility` flags from the CLI.

## Breaking Change

```
[ ] Yes
```


mat-if (approved):
image
😍

PR #3592, opened 2023-02-28
uses unspentNoteHashes in getUnspentNotes

hughy:
## Summary

we use getUnspentNotes when funding transactions to find notes to spend.

the current implementation of getUnspentNotes iterates over all notes in the wallet in order to find notes for the right asset that are confirmed and unspent.

the unspentNoteHashes index that we added to support computing available balance can also be used to efficiently find confirmed, unspent notes for a given asset.

implements walletDb methods for iterating over unspent notes by assetId and sequence range.

rewrites getUnspentNotes to use the unspentNoteHashes index via the new walletDb methods.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/account.test.ts
+++ b/ironfish/src/wallet/account.test.ts
@@ -420,7 +420,7 @@ describe('Accounts', () => {
       await node.wallet.updateHead()
 
       let unspentA = await AsyncUtils.materialize(
-        accountA['walletDb'].loadUnspentNoteHashes(accountA),
+        accountA['walletDb'].loadUnspentNoteHashes(accountA, Asset.nativeId()),
       )
 
       expect(unspentA).toHaveLength(1)
@@ -429,7 +429,7 @@ describe('Accounts', () => {
       await useTxFixture(node.wallet, accountA, accountB)
 
       unspentA = await AsyncUtils.materialize(
-        accountA['walletDb'].loadUnspentNoteHashes(accountA),
+        accountA['walletDb'].loadUnspentNoteHashes(accountA, Asset.nativeId()),
       )
       expect(unspentA).toHaveLength(0)
     })
@@ -448,7 +448,7 @@ describe('Accounts', () => {
       await useTxFixture(node.wallet, accountA, accountB)
 
       const unspentB = await AsyncUtils.materialize(
-        accountB['walletDb'].loadUnspentNoteHashes(accountB),
+        accountB['walletDb'].loadUnspentNoteHashes(accountB, Asset.nativeId()),
       )
       expect(unspentB).toHaveLength(0)
     })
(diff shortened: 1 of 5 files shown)


EvanJRichard (approved):
Seems much cleaner. `Test Plan` is empty, are the unit tests sufficient?

ironfish/src/wallet/account.test.ts
```diff
@@ -1834,25 +1834,68 @@ describe('Accounts', () => {
...
+      )
+
+      expect(unspentNotes).toHaveLength(1)
+
+      unspentNotes = await AsyncUtils.materialize(
+        accountA.getUnspentNotes(Asset.nativeId(), { confirmations: 1 }),
+      )
+
+      expect(unspentNotes).toHaveLength(0)
+    }, 50000)
```
dguenther, line 1899: Was this long timeout just for the initial fixture generation?

ironfish/src/wallet/account.test.ts
```diff
@@ -1834,25 +1834,68 @@ describe('Accounts', () => {
...
+      await node.wallet.updateHead()
+
+      const unspentNotes = await AsyncUtils.materialize(
+        accountA.getUnspentNotes(Asset.nativeId()),
+      )
+
+      expect(unspentNotes).toHaveLength(2)
+    })
+
+    it('loads filters unspent notes by confirmation range', async () => {
```
dguenther, line 1879: maybe change 'loads filters' to 'filters'

ironfish/src/wallet/account.test.ts
```diff
@@ -1834,25 +1834,68 @@ describe('Accounts', () => {
...
       unspentHashes = await AsyncUtils.materialize(
-        accountA['walletDb'].loadUnspentNoteHashes(accountA),
+        accountA['walletDb'].loadUnspentNoteHashes(accountA, Asset.nativeId()),
       )
       expect(unspentHashes).toHaveLength(1)
       expect(unspentHash).toEqualBuffer(unspentHashes[0])
     })
   })
+
+  describe('getUnspentNotes', () => {
```
dguenther, line 1859: I'm not sure how inconvenient it is to test, but maybe a test that tests filtering by asset ID could be useful?

dguenther (approved):

ironfish/src/wallet/account.test.ts
hughy: Yes, removed!

ironfish/src/wallet/account.test.ts
hughy: fixed in test and fixture

ironfish/src/wallet/account.test.ts
hughy: Not too inconvenient at all! I added a test of filtering with multiple assets.

PR #3593, opened 2023-02-28
Feature: accounts:delete --wait

EvanJRichard:
## Summary
As it stands currently, `accounts:delete` only marks an account for deletion. This account is later deleted in the event loop. 

This feature proposes to add an option `--wait`. This will immediately force the cleanup code to be called. 

## Testing Plan
```
yarn start wallet:use killme
The default account is now: killme
```
```
yarn start wallet:delete --wait killme
Deleting account 'killme'... done
Account 'killme' successfully deleted.
```
```
yarn start wallet:accounts
default
```
## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -1404,10 +1404,10 @@ export class Wallet {
       }
       // no this.walletDb.removeAccount(account, tx) - we don't want to add the account to cleanup list
       await this.walletDb.removeHead(account, tx)
-      this.accounts.delete(account.id)
       await this.walletDb.deleteAccount(account, tx)
-      this.onAccountRemoved.emit(account)
     })
+    this.accounts.delete(account.id)
+    this.onAccountRemoved.emit(account)
   }
 
   async cleanupDeletedAccounts(): Promise<void> {
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -1400,7 +1400,11 @@ export class Wallet {
       return
     }
 
-    await this.walletDb.cleanupDeletedAccounts(1000, this.eventLoopAbortController.signal)
+    const recordsToCleanup = 1000
+    await this.walletDb.cleanupDeletedAccounts(
+      recordsToCleanup,
+      this.eventLoopAbortController.signal,
+    )
   }
 
   get hasDefaultAccount(): boolean {
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -1397,8 +1397,17 @@ export class Wallet {
   }
 
   async removeAccountSynchronous(account: Account, tx?: IDatabaseTransaction): Promise<void> {
-    await this.removeAccount(account, tx)
-    await this.walletDb.deleteAccount(account)
+    await this.walletDb.db.withTransaction(tx, async (tx) => {
+      if (account.id === this.defaultAccount) {
+        await this.walletDb.setDefaultAccount(null, tx)
+        this.defaultAccount = null
+      }
+      // no this.walletDb.removeAccount(account, tx) - we don't want to add the account to cleanup list
+      await this.walletDb.removeHead(account, tx)
+      this.accounts.delete(account.id)
+      await this.walletDb.deleteAccount(account, tx)
+      this.onAccountRemoved.emit(account)
+    })
   }
 
   async cleanupDeletedAccounts(): Promise<void> {
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -1388,7 +1388,7 @@ export class Wallet {
   }
 
   async forceDeletedAccountCleanup(): Promise<void> {
-    await this.walletDb.cleanupAllAccountsNow(this.eventLoopAbortController.signal)
+    await this.walletDb.forceDeletedAccountCleanup(this.eventLoopAbortController.signal)
   }
 
   async cleanupDeletedAccounts(): Promise<void> {
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -1372,15 +1372,6 @@ export class Wallet {
     await this.removeAccount(account)
   }
 
-  async removeAccountByNameSynchronous(name: string): Promise<void> {
-    const account = this.getAccountByName(name)
-    if (!account) {
-      return
-    }
-
-    await this.removeAccountSynchronous(account)
-  }
-
   async removeAccount(account: Account, tx?: IDatabaseTransaction): Promise<void> {
     await this.walletDb.db.withTransaction(tx, async (tx) => {
       if (account.id === this.defaultAccount) {
@@ -1396,8 +1387,7 @@ export class Wallet {
     this.onAccountRemoved.emit(account)
   }
 
-  async removeAccountSynchronous(account: Account, tx?: IDatabaseTransaction): Promise<void> {
-    await this.removeAccount(account, tx)
+  async forceDeletedAccountCleanup(): Promise<void> {
     await this.walletDb.cleanupAllAccountsNow(this.eventLoopAbortController.signal)
   }
 
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -1387,8 +1387,8 @@ export class Wallet {
     this.onAccountRemoved.emit(account)
   }
 
-  async forceDeletedAccountCleanup(): Promise<void> {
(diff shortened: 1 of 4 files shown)


ironfish/src/wallet/walletdb/walletdb.ts
```diff
@@ -1023,6 +1023,29 @@ export class WalletDB {
...
+    ]
+    const prefix = calculateAccountPrefix(account.id)
+    const range = StorageUtils.getPrefixKeyRange(prefix)
+
+    for (const store of stores) {
+      for await (const key of store.getAllKeysIter(undefined, range)) {
+        await store.del(key)
+      }
+    }
+  }
```
EvanJRichard, line 1047: This is basically copied from `cleanupDeletedAccounts`. I wonder if the list `stores` could be built and shared between the two functions.

ironfish/src/wallet/wallet.ts
```diff
@@ -1387,6 +1396,10 @@ export class Wallet {
     this.onAccountRemoved.emit(account)
   }
 
+  async removeAccountSynchronous(account: Account): Promise<void> {
+    await this.walletDb.deleteAccount(account)
+  }
```
EvanJRichard, line 1401: Does this also call for an `emit`?

ironfish/src/wallet/wallet.ts
```diff
@@ -1387,6 +1396,10 @@ export class Wallet {
     this.onAccountRemoved.emit(account)
   }
 
+  async removeAccountSynchronous(account: Account): Promise<void> {
+    await this.walletDb.deleteAccount(account)
```
dguenther, line 1400: I'm pretty sure this also needs to make some of the same calls in `wallet.removeAccount`. I tried running `wallet:delete --wait default`, and when I ran `wallet:accounts`, the same `default` account still showed up.

EvanJRichard: Thanks for trying it out. I don't understand the event loop too well yet, pushed a version that does the same stuff in `wallet.removeAccount` too. I can't tell if that's safe, though? @dguenther

hughy (dismissed):
looks good to me! 👍

ironfish/src/wallet/wallet.ts
```diff
@@ -1387,6 +1396,20 @@ export class Wallet {
...
+    await this.walletDb.db.withTransaction(tx, async (tx) => {
+      if (account.id === this.defaultAccount) {
+        await this.walletDb.setDefaultAccount(null, tx)
+        this.defaultAccount = null
+      }
+      // no this.walletDb.removeAccount(account, tx) - we don't want to add the account to cleanup list
+      await this.walletDb.removeHead(account, tx)
+      this.accounts.delete(account.id)
+      await this.walletDb.deleteAccount(account, tx)
+      this.onAccountRemoved.emit(account)
```
hughy, line 1409: It might be good to move `this.accounts.delete(account.id)` and `this.onAccountRemoved.emit(account)` outside of the database transaction. That way these will only run if the transaction is committed.

It's probably OK in this case because `accounts` is an in-memory cache populated when the node starts.

dguenther (changes requested):
This leaves the wallet in an unrecoverable state, I think because it's missing a `this.accounts.del`. With no node running, I ran:

```
yarn start wallet:create dag
yarn start wallet:delete --wait dag
yarn start wallet:accounts
```

and got:

```
Error: Expected value not to be undefined
    at Function.isNotUndefined (/Users/derek/repos/ironfish/ironfish/src/assert.ts:20:13)
    at WalletDB.getHead (/Users/derek/repos/ironfish/ironfish/src/wallet/walletdb/walletdb.ts:344:12)
    at Wallet.getLatestHeadHash (/Users/derek/repos/ironfish/ironfish/src/wallet/wallet.ts:1499:20)
    at Wallet.load (/Users/derek/repos/ironfish/ironfish/src/wallet/wallet.ts:206:32)
    at Wallet.open (/Users/derek/repos/ironfish/ironfish/src/wallet/wallet.ts:189:5)
    at IronfishNode.openDB (/Users/derek/repos/ironfish/ironfish/src/node.ts:302:7)
    at Object.waitForOpen (/Users/derek/repos/ironfish/ironfish/src/utils/node.ts:17:7)
    at IronfishSdk.connectRpc (/Users/derek/repos/ironfish/ironfish/src/sdk.ts:256:5)
    at AccountsCommand.start (/Users/derek/repos/ironfish/ironfish-cli/src/commands/wallet/accounts.ts:22:20)
    at AccountsCommand.run (/Users/derek/repos/ironfish/ironfish-cli/src/command.ts:73:7)
```

It'd definitely be great if the wallet could recover from these issues better, but aside from that, I think it'd be helpful to unify the code paths of the synchronous and async deletes more. Two things I noticed:

* It's not the worst if the account ends up in the cleanup list, since it'll quickly discover nothing to do and remove it next time cleanup runs, but if we really want to remove all traces, maybe a flag that determines whether the account is added to the cleanup list? 
* `deleteAccount` and `cleanupDeletedAccounts` are quite similar... maybe they could share the same code, but the sync version would set `recordsToCleanup` to `Infinity`?

hughy:
> It's not the worst if the account ends up in the cleanup list, since it'll quickly discover nothing to do and remove it next time cleanup runs, but if we really want to remove all traces, maybe a flag that determines whether the account is added to the cleanup list?

I suggested avoiding inserting the account into the cleanup list to avoid parallel deletes for the same account 😅 

Thinking more about this, I don't think that that's possible since the synchronous deletion will be blocking.

ironfish/src/wallet/wallet.ts
dguenther: The event loop is just a factor here in order to periodically call `cleanupDeletedAccounts` -- it lets us limit the amount of cleanup done at one time so that it doesn't use too much of the main thread.

You might be asking about updating the database though -- the database is single-process, and database writes done inside of a transaction don't get committed until the end of the transaction, so there's no place where concurrent reads can fetch half-deleted state from the DB.

EvanJRichard:
@hughy @dguenther Thanks for the review, guys. It's simpler now, and seems to work - at least I haven't been able to break it yet, and the bug Derek pointed out above doesn't reproduce. Going to keep testing it.

EvanJRichard:
Marking for auto-merge on approval, so don't approve if you don't think it should merge.

ironfish/src/wallet/wallet.ts
```diff
@@ -1372,6 +1372,15 @@ export class Wallet {
     await this.removeAccount(account)
   }
 
+  async removeAccountByNameSynchronous(name: string): Promise<void> {
```
lwisne, line 1375: Quick drive-by: it's confusing to me as a random reader that there is a function with the form `async ...Synchronous`. Maybe this is a pattern we use elsewhere, but I'm not sure what to do with this mentally.

ironfish/src/wallet/walletdb/walletdb.ts
```diff
@@ -1023,9 +1023,11 @@ export class WalletDB {
     await this.pendingTransactionHashes.clear(tx, account.prefixRange)
   }
 
-  async cleanupDeletedAccounts(signal?: AbortSignal): Promise<void> {
-    let recordsToCleanup = 1000
+  async cleanupAllAccountsNow(signal?: AbortSignal): Promise<void> {
+    return this.cleanupDeletedAccounts(Number.MAX_SAFE_INTEGER, signal)
```
EvanJRichard, line 1027: this could be `Number.POSITIVE_INFINITY` if we want

ironfish/src/wallet/wallet.ts
EvanJRichard: Good point. `remove...Blocking` or `remove...WithCleanup` would be more accurate. Changing.

EvanJRichard: Actually, recent changes make this function entirely superfluous. Removing.

ironfish/src/wallet/wallet.ts
```diff
@@ -1387,6 +1387,10 @@ export class Wallet {
     this.onAccountRemoved.emit(account)
   }
 
+  async forceDeletedAccountCleanup(): Promise<void> {
```
dguenther, line 1390: I'd suggest renaming to `forceCleanupDeletedAccounts` to make the name align a bit more with `cleanupDeletedAccounts`

ironfish/src/wallet/walletdb/walletdb.ts
dguenther: Yeah, I think `Number.POSITIVE_INFINITY` would be good. It aligns more with the dev expectation that it should wait for the deletion of every record

dguenther:
Added a couple more comments, but otherwise looks cleaner to me, I like it! Going to test it out as well

dguenther (approved):

PR #3594, opened 2023-02-28
adds createdAt field to Account

hughy:
## Summary

the createdAt field or 'account birthday' will be used during account scanning to skip blocks that were created before the account was created.

adds a nullable createdAt date field. the field may be null for imported accounts. if an account is imported we cannot determine when the account was created without scanning the chain for the earliest transactions that involved that account.

theses changes do not yet support importing an account with a creation date. followup changes should add createdAt to account exports and make the field optional for imports.

## Breaking Change

```
[ ] Yes
```


rohanjadvani (approved):

dguenther (approved):
Tested that `wallet:address` still works to load an older account, and an account created with this PR checked out. I'm not sure if there's anything else that can be easily tested beyond that so far

PR #3597, opened 2023-03-02
sets the account birthday on block connect/disconnect

hughy:
## Summary

sets the account birthday to the earliest transaction timestamp when a block is connected or disconnected if the following are all true:
- the account was involved in a transaction on the block
- the account birthday is null or the block timestamp is earlier than the account birthday

defines 'getEarliestTransactionTimestamp' methods for reading the earliest timestamp from the wallet database using the timestampToTransactionHash index.

defines wallet methods for updating the account createdAt field. updateAccount updates the account record in the wallet database and updates the entry in the wallet's cache.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/__fixtures__/wallet.test.ts.fixture
+++ b/ironfish/src/wallet/__fixtures__/wallet.test.ts.fixture
@@ -4805,5 +4805,217 @@
         }
       ]
     }
+  ],
+  "Accounts connectBlock should not set account createdAt if the connected block is later than the current value": [
+    {
+      "version": 1,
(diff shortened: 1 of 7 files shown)


hughy:
I chatted with @NullSoldier about account birthdays.

Setting or updating the account `createdAt` field for an existing account is difficult to do correctly. The goal of adding account birthdays is to optimize scanning. We can support this for new accounts and imported accounts that include `createdAt` in the imported data without updating `createdAt` during a scan. Users can still manually set `createdAt` in a JSON import.

PR #3599, opened 2023-03-02
Remove promise all

ygao76:
## Testing Plan
```
yajun@Yajuns-MBP ironfish-cli % yarn start wallet:send -t e4f6cab7e8774f74310fbfb4a345fac8fec289531914
yarn run v1.22.19
$ yarn build && yarn start:js wallet:send -t e4f6cab7e8774f74310fbfb4a345fac8fec28953191bcc908265b74e4
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc --inspect=:0 --inspect-publish-uid=http4
? Select the asset you wish to send d7c86706f5817aa718cd1cfad03233bcd64a7789fd9422d3b17af6823a7e6ac6
($IRON) (0.00000012)
Enter the amount (balance 0.00000012): 0.00000012
? Select the fee you wish to use for this transaction (Use arrow keys)
  - Slow  (Not enough $IRON)
  - Average  (Not enough $IRON)
  - Fast  (Not enough $IRON)
❯ Enter a custom fee
```
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/utils/fees.ts
+++ b/ironfish-cli/src/utils/fees.ts
@@ -26,23 +26,23 @@ export async function selectFee(options: {
 }): Promise<RawTransaction> {
   const feeRates = await options.client.estimateFeeRates()
 
-  const [slow, average, fast] = await Promise.all([
-    getTxWithFee(
+  const [slow, average, fast] = [
+    await getTxWithFee(
       options.client,
       options.transaction,
       CurrencyUtils.decode(feeRates.content.slow),
     ),
-    getTxWithFee(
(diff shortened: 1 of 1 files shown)


EvanJRichard (approved):

hughy (approved):
This looks good to me!

I think we should get to the bottom of why this doesn't work with `Promise.all` -- I suspect there may be something going wrong with our RPC calls. This does solve the problem, so I think we should get it in and investigate `Promise.all` separately!

PR #3600, opened 2023-03-02
fixes account import with account birthdays

hughy:
## Summary

import accounts fails following the addition of account birthdays with 'value.createdAt.getTime is not a function' because we do not parse any 'createdAt' into a date.

adds 'createdAt' to the required RPC fields in importAccount and ensures that the field is included in requests to importAccount.

## Testing Plan

manually tested various export/import conditions with and without the createdAt field:
- bech32
- json
- spendingKey
- mnemonic
- viewonly

## Breaking Change

```
[ ] Yes
```


EvanJRichard (approved):
:ship:

dguenther (approved):
Tested that exporting a view-only account and importing it into another wallet works. Thanks!

PR #3602, opened 2023-03-02
fixes feeEstimator tests, regenerates fixtures

hughy:
## Summary

many of the feeEstimator tests were implemented expecting that creating transactions would not be accepted to the mempool as part of fixture generation. however we now add transactions to the mempool during fixture generation by calling 'wallet.post'.

these tests fail when fixtures are regenerated.

fixes the tests by explicitly passing 'broadcast = false' during fixture generation to avoid adding transaction fixtures to the mempool.

## Breaking Change

```
[ ] Yes
```


EvanJRichard (approved):
Do we want to add a comment explaining this anywhere or nah since the test will break if it's undone anyways?

dguenther (approved):
I'm down to merge this in if it fixes the fixture generation, but I think we should remove the `broadcast` flag from `useBlockWithTx` -- The fixture generation needs to be stateless (or at least everything it does needs to be encapsulated in the output) in order to prevent these issues.

We might be able to use one of the other functions besides `generate`, but I'd have to look a bit closer at how the fixture generation works.

hughy:
> I'm down to merge this in if it fixes the fixture generation, but I think we should remove the `broadcast` flag from `useBlockWithTx` -- The fixture generation needs to be stateless (or at least everything it does needs to be encapsulated in the output) in order to prevent these issues.
> 
> We might be able to use one of the other functions besides `generate`, but I'd have to look a bit closer at how the fixture generation works.

I think making `useBlockWithTx` stateless is a much better solution than merging this in. There are a few other tests that need the same fix (peerNetwork, transactionFetcher). I'll give fixing the fixture generation a try!

hughy:
Closing in favor of https://github.com/iron-fish/ironfish/pull/3603, which fixes fixture generation

PR #3603, opened 2023-03-03
removes broadcast option from block fixtures

hughy:
## Summary

removes the 'broadcast' option from block fixture generation so that creating transactions will not affect the state of the wallet or the mempool.

changes block fixture generation to use worker pool 'postTransaction' instead of 'wallet.post', which has side effects when used with 'broadcast'. 'wallet.post' will sync transactions to the wallet and add them to the mempool if called with 'broadcast' set to true.

regenerates fixtures.

## Breaking Change

```
[ ] Yes
```


dguenther (approved):
Nice, thanks for this cleanup!

PR #3604, opened 2023-03-03
feat(ironfish): add support for a minimum required transaction fee

holahula:
## Breaking Change

```
[ ] Yes
```


ironfish/src/consensus/consensus.ts
```diff
@@ -28,6 +28,11 @@ export type ConsensusParameters = {
    * Max block size
    */
   maxBlockSizeBytes: number
+
+  /**
+   * The minimum fee that a transaction must have to be accepted
+   */
+  minFee: number
```
NullSoldier, line 35: This is an amount in ore so it should not be a number

ironfish/src/defaultNetworkDefinitions.ts
```diff
@@ -84,7 +84,8 @@ export const TESTNET = `{
       "genesisSupplyInIron": 42000000,
       "targetBlockTimeInSeconds": 60,
       "targetBucketTimeInSeconds": 10,
-      "maxBlockSizeBytes": 2000000
+      "maxBlockSizeBytes": 2000000,
+      "minFee": 1
```
danield9tqh, line 88: Let's set this to 0 so we don't hurt phase3 merging this is

ironfish/src/consensus/consensus.ts
holahula: leaving as is due to issues with yup supporting `BigInt`

ironfish/src/defaultNetworkDefinitions.ts
holahula: changed to 0

ironfish/src/consensus/verifier.ts
```diff
@@ -504,6 +522,7 @@ export enum VerificationResultReason {
   MAX_BLOCK_SIZE_EXCEEDED = 'Block size exceeds maximum',
   MAX_TRANSACTION_SIZE_EXCEEDED = 'Transaction size exceeds maximum',
   MINERS_FEE_EXPECTED = 'Miners fee expected',
+  MINIMUM_TRANSACTION_FEE_NOT_MET = 'Transaction fee is below the minimum required fee',
```
danield9tqh, line 525: Probably could use `MINIMUM_FEE_NOT_MET` here for succinctness too

ironfish/src/consensus/verifier.test.ts
```diff
@@ -254,6 +274,19 @@ describe('Verifier', () => {
       })
     })
 
+    it('rejects a block with a transaction with fee less than minimum', async () => {
```
danield9tqh, line 277: could we add one more test that tests `chain.addBlock()`

danield9tqh (approved):

PR #3606, opened 2023-03-04
Add mempool status command

danield9tqh:
## Summary
Adding a status command for the mempool to monitor the recently evicted cache. Also fixes a telemetry bug by sending number instead of a float

## Breaking Change

```
[ ] Yes
```


holahula (approved):

plast73:
yes

PR #3608, opened 2023-03-06
adds bech32 to dependencies for @ironfish/sdk

hughy:
## Summary

`yarn add bech32`

after installing @ironfish/sdk, bech32 module cannot be found (e.g., https://discord.com/channels/771503434028941353/831945060777852998/1082211129796874311)

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/package.json
+++ b/ironfish/package.json
@@ -21,6 +21,7 @@
     "@ironfish/rust-nodejs": "0.1.27",
     "@napi-rs/blake-hash": "1.3.3",
     "axios": "0.21.4",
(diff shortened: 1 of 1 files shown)


hughy:
We define util functions that use bech32 in the SDK, but it looks like we only use them in the CLI right now. An alternative to this dependency is to move those util definitions to the CLI.

dguenther (approved):
This seems like an adequate quick fix -- I'm not expecting the yarn.lock to need updating since it's already in CLI. Could you test that it fixes the issue with the sdk before merging?

PR #3609, opened 2023-03-06
fixes balance calculation when confirmations >= chain length

hughy:
## Summary

if the length of the chain is shorter than the confirmation range we calculate a negative number for the maximum confirmed sequence. this results in an error when attempting to construct a key range with a negative value since the db encoding requires a non-negative integer value.

however, the genesis block is always confirmed, so the maximum confirmed sequence can never be less than 1.

fixes calculation of available balance by setting maxConfirmedSequence to be at least 1.

adds unit test for this case

## Testing Plan

Steps to reproduce:
```
rm -rf ~/.ironfishdev
yarn start start --datadir ~/.ironfishdev --networkId 2
yarn start wallet:balances --datadir ~/.ironfishdev
> The value of "value" is out of range. It must be >= 0 and <= 4294967295. Received -1
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/__fixtures__/account.test.ts.fixture
+++ b/ironfish/src/wallet/__fixtures__/account.test.ts.fixture
@@ -3295,5 +3295,44 @@
         }
       ]
     }
+  ],
+  "Accounts getBalance should calculate balances on a chain with no confirmed blocks": [
+    {
+      "version": 1,
+      "id": "9bca25a4-cb27-4217-aa32-cacff1b03dac",
+      "name": "accountA",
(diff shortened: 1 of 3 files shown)


dguenther (approved):
Ran the new test without the change to verify that it fails, then ran the CLI steps to verify the error doesn't reproduce:

```
rm -rf ~/.ironfishdev
yarn start start --datadir ~/.ironfishdev --networkId 2
yarn start wallet:balances --datadir ~/.ironfishdev
```

PR #3610, opened 2023-03-06
Close trusted setup command

danield9tqh:
## Summary
Giving some more information for users who try to run the `ironfish ceremony` command by telling them the ceremony is complete. We may still need this command in case of future setups / network upgrades. So for now, just exit the command with some information

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/ceremony.ts
+++ b/ironfish-cli/src/commands/ceremony.ts
@@ -47,6 +47,12 @@ export default class Ceremony extends IronfishCommand {
     let refreshEtaInterval: NodeJS.Timeout | null = null
     let etaDate: Date | null = null
 
+    // Trusted setup has ended so just
+    this.log(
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/ceremony.ts
```diff
@@ -47,6 +47,13 @@ export default class Ceremony extends IronfishCommand {
     let refreshEtaInterval: NodeJS.Timeout | null = null
     let etaDate: Date | null = null
 
+    // Trusted setup has ended but this command may still be needed in case of future
+    // setups / network upgrades. So for now, just exit the command with some information
+    this.log(
+      `The trusted setup ceremony was completed on Mar 3, 2023. For more information on the trusted setup process and its completion please read https://setup.ironfish.network.`,
```
danield9tqh, line 53: We should add an amendment to this blog post with link to the final parameters

dguenther (approved):

PR #3614, opened 2023-03-07
Add chron job for testing regenerate fixtures

danield9tqh:
## Summary
Tests are failing sometimes when regenerating fixtures because regenerating fixtures is not done on every code change. It's an expensive process. There should be some check on this though so adding a Github action to regenerate and run tests once a week seems like a good short term solution

## Breaking Change

```
[ ] Yes
```


.github/workflows/ci-regenerate-fixtures.yml
```diff
@@ -0,0 +1,39 @@
+name: Test With Regenerate Fixtures
+
+on:
+  schedule:
+    - cron: '0 15 * * 1' # Every Monday 7AM PST
```
hughy, line 5: Would we ever have to respond to the result of this job, kill it, or prevent it from running? 7AM PST might be earlier than we're online some weeks, especially if that Monday is a holiday!

.github/workflows/ci-regenerate-fixtures.yml
```diff
@@ -0,0 +1,39 @@
...
+          shared-key: nodejs
+
+      - name: Install packages
+        run: yarn --non-interactive --frozen-lockfile
+
+      - name: Delete fixture files
+        run: node ./tools/delete-fixtures.js
+
+      - name: Run tests
+        run: JEST_TIMEOUT=1000000000 yarn test:coverage --maxWorkers=2 --workerIdleMemoryLimit=2000MB
```
hughy, line 36: What is the `workerIdleMemoryLimit` based on?

tools/delete-fixtures.js
```diff
@@ -0,0 +1,32 @@
...
+    .filter(dirent => dirent.isDirectory())
+    .map(dirent => resolve(dir, dirent.name))
+}
+
+// Recursively get all sub-directories including the root directory
+async function getDirectoriesRecursivePlusRoot(dir) {
+  const subDirs = await getDirectories(dir)
+
+  const list = await Promise.all(subDirs.map((subDir) => getDirectoriesRecursivePlusRoot(subDir)))
+  return [dir, ...list.reduce((a, f) => a.concat(f), [])]
```
hughy, line 16: nit - would `list.flat()` work instead of the `reduce`?

hughy (commented):
I think we should definitely be doing this regularly!

Do you think we should also do this as part of the release workflow? For example, we might put out a release on a Friday and then find out on Monday that tests were broken.

.github/workflows/ci-regenerate-fixtures.yml
dguenther: This is carried over from our other CI test workflows, but I think it's roughly correlated to the amount of memory on GH Actions VMs, with a bit of headroom. https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

Unfortunately Jest + NodeJS tests seem to leak memory over time and crash, so they added this flag to restart workers when they approach the memory limit.

.github/workflows/ci-regenerate-fixtures.yml
danield9tqh: Yea I put 7AM PST because it takes about an hour to run so it should be complete by 8AM. Also since we have some people on east coast that's 11AM which seems pretty safe. Maybe we can discuss this amongst the team

tools/delete-fixtures.js
danield9tqh: yep changed

.github/workflows/ci-regenerate-fixtures.yml
danield9tqh: Decided to change to Tuesday morning in case of monday holiday but still before most releases. And adding a badge to the Github repo so that failures aren't missed

hughy (approved):

PR #3618, opened 2023-03-07
Make faucet only for testnet

ygao76:
## Summary
check network Id for faucet command.
## Testing Plan
local yarn start faucet

```
yajun@Yajuns-MBP ironfish-cli % yarn start faucet
yarn run v1.22.19
$ yarn build && yarn start:js faucet
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc --inspect=:0 --inspect-publish-uid=httpt

::::::::::          :::::::::::::::::
::::::::::::       :::::::::::::::::::
:::::::::::::     :::::::::::::::::::::
::::::::::::::   ::::::::::::::::::::::::
 ::::::::::::: ::::::::::        :::::::::
   :::::::::::::::::::::          ::::::::::
   :::::::::::::::::::::          :::::::::
 ::::::::::::: ::::::::::        :::::::::
::::::::::::::   ::::::::::::::::::::::::
:::::::::::::     :::::::::::::::::::::
::::::::::::       :::::::::::::::::::
::::::::::           ::::::::::::::::

Enter your email to stay updated with Iron Fish:
```
## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/faucet.ts
```diff
@@ -33,10 +33,18 @@ export class FaucetCommand extends IronfishCommand {
       this.exit(1)
     }
 
-    this.log(ONE_FISH_IMAGE)
-
     const client = await this.sdk.connectRpc()
 
+    const configResponse = await client.getConfig({
+      name: 'networkId',
+    })
```
hughy, line 40: I don't think we can use `getConfig` for this -- the node's config might not have the value of `networkId` that's used for the network definition

https://iron-fish.slack.com/archives/C02PF56HESK/p1678236428921539

dguenther: @ygao76 added an RPC for this in #3622

ygao76: updated the PR using the new RPC getNetworkInfo

dguenther (approved):
Could you also add the `chain/getNetworkInfo` to the RPC docs in https://github.com/iron-fish/website ? We just added those yesterday lol

PR #3619, opened 2023-03-07
feat(cli,ironfish): Offer name change on duplicate name during import

rohanjadvani:
## Summary

If you import your default account (which has the name default) into a new wallet / data directory (and that database already has a default account), you will encounter an error that shows a duplicate name error. This is not a great UX, and it'd be nice to offer the user a chance to rename the account during the import.

* In the RPC request, throw a validation error if the account name already exists
* In the CLI check for a validation error and error code. If the code is account exists, prompt for a different name.
* If the same name is provided, error out

## Testing Plan

Existing behavior

https://user-images.githubusercontent.com/5459049/223557261-07949a5d-1ce6-4d21-8a38-8125cb188390.mp4

New behavior for same name

https://user-images.githubusercontent.com/5459049/223557529-aa1167d9-7966-46ae-b712-5d5f19c7d082.mp4

New behavior for new name

https://user-images.githubusercontent.com/5459049/223557809-ed0210c3-02e0-4d94-9fc9-6b75d461a79e.mp4

New behavior with an existing error

https://user-images.githubusercontent.com/5459049/223557981-e4ff50db-3ab4-4d79-b92c-4352aa564eb8.mp4

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/import.ts
+++ b/ironfish-cli/src/commands/wallet/import.ts
@@ -2,20 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { generateKeyFromPrivateKey, wordsToSpendingKey } from '@ironfish/rust-nodejs'
-import {
-  ACCOUNT_SCHEMA_VERSION,
-  Bech32m,
-  ERROR_CODES,
-  ImportResponse,
-  JSONUtils,
-  PromiseUtils,
-  RpcResponseEnded,
-} from '@ironfish/sdk'
+import { ACCOUNT_SCHEMA_VERSION, Bech32m, JSONUtils, PromiseUtils } from '@ironfish/sdk'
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -62,7 +71,30 @@ export class ImportCommand extends IronfishCommand {
     }
 
     const rescan = flags.rescan
-    const result = await client.importAccount({ account, rescan })
+
+    let result: RpcResponseEnded<ImportResponse>
+    try {
+      result = await client.importAccount({ account, rescan })
```
jowparks, line 77: Instead of trying and failing, can we use the `AccountImport.name` field and compare to existing accounts in the wallet? Then prompt for rename if there is a duplicate?

rohanjadvani: Yeah that sounds good.

danield9tqh (approved):

NullSoldier (approved):

PR #3620, opened 2023-03-07
Update testnet genesis block to use new parameters

danield9tqh:
## Summary
Updating the testnet genesis block to be compatible with the new parameters generated for mainnet

## Breaking Change

```
[ ] Yes
```


dguenther (approved):

dguenther:
Thanks! 👍 I pushed the commit onto the network-reset branch for now, it was quicker than rebasing this one on that branch

PR #3622, opened 2023-03-08
Get network info rpc

ygao76:
## Summary
Get Network Info RPC, it will return network id and other config in internal config.

## Testing Plan
unit test
## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/chain/getNetworkInfo.ts
```diff
@@ -0,0 +1,48 @@
...
+       throw new ValidationError(`No config option ${String(request.data.name)}`)
+     }
+ 
+     let pickKeys: string[] | undefined = undefined
+     
+     if (request.data?.name) {
+       pickKeys = [request.data.name]
+     }
+ 
+     const data = JSON.parse(JSON.stringify(node.internal.config, pickKeys)) as GetNetworkInfoResponse
```
dguenther, line 43: This won't work, because `JSON.stringify` just does first-level keys in the object, but the `config` object uses prototype inheritance to override properties.

For now, let's just return the `networkId` property, since that's the only one we want. Some of the other `internal` keys aren't network-related, and we also don't want to expose all of them over RPC (like the RPC auth token)

ironfish/src/rpc/routes/chain/getNetworkInfo.ts
```diff
@@ -0,0 +1,30 @@
...
+  networkId?: number
+}
+
+export const GetNetworkInfoRequestSchema: yup.MixedSchema<GetNetworkInfoRequest> = yup
+  .mixed()
+  .oneOf([undefined] as const)
+
+export const GetNetworkInfoResponseSchema: yup.ObjectSchema<GetNetworkInfoResponse> = yup
+  .object({
+    networkId: yup.number().optional(),
```
dguenther, line 18: Nice, this looks good to me. One more thing, could you see if it's possible to remove this `.optional()`? The network ID might be always initialized to the default value, but I'm not sure.

dguenther (approved):

PR #3624, opened 2023-03-08
Replace bigint casting with bigint literal

NullSoldier:
## Summary

There's no reason to wrap bigints in the bigint type. You can just use literals. This only replaces some places.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/primitives/target.ts
+++ b/ironfish/src/primitives/target.ts
@@ -7,28 +7,25 @@ import { BigIntUtils } from '../utils/bigint'
 /**
  *  Minimum difficulty, which is equivalent to maximum target
  */
-const MIN_DIFFICULTY = BigInt(131072)
+const MIN_DIFFICULTY = 131072n
 
 /**
  * Maximum target, which is equivalent of minimum difficulty of 131072
  * target == 2**256 / difficulty
  */
-const MAX_TARGET = BigInt(
-  '883423532389192164791648750371459257913741948437809479060803100646309888',
-)
+const MAX_TARGET = 883423532389192164791648750371459257913741948437809479060803100646309888n
 
 /**
  *  Maximum number to represent a 256 bit number, which is 2**256 - 1
  */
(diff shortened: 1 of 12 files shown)


ironfish/src/strategy.test.slow.ts
```diff
@@ -213,7 +213,7 @@ describe('Demonstrate the Sapling API', () => {
         workerPool: new WorkerPool(),
         consensus: new TestnetConsensus(consensusParameters),
       })
-      const minersFee = await strategy.createMinersFee(BigInt(0), 0, key.spendingKey)
+      const minersFee = await strategy.createMinersFee(0, 0, key.spendingKey)
```
hughy, line 216:
```suggestion
      const minersFee = await strategy.createMinersFee(0n, 0, key.spendingKey)
```

Missing `n` to make this a bigint literal

ironfish/src/primitives/target.ts
```diff
@@ -7,28 +7,25 @@ import { BigIntUtils } from '../utils/bigint'
...
-)
+const MAX_TARGET = 883423532389192164791648750371459257913741948437809479060803100646309888n
 
 /**
  *  Maximum number to represent a 256 bit number, which is 2**256 - 1
  */
-const MAX_256_BIT_NUM = BigInt(
-  '115792089237316195423570985008687907853269984665640564039457584007913129639935',
-)
+const MAX_256_BIT_NUM =
```
danield9tqh, line 21: Here do you want to do this for readability?
```javascript
const MAX_256_BIT_NUM = 2n ** 256n - 1n
```

danield9tqh: could do in another change too

dguenther (approved):

ironfish/src/primitives/target.ts
dguenther: I think that's clean, but I'd recommend doing it in another PR just to avoid the back-and-forth

PR #3625, opened 2023-03-08
changes timestampToTransactionHash to one-to-many

hughy:
## Summary

it's possible for multiple transactions to have the same timestamp in a wallet.

for example, if an account receives two transactions that are on the same block, but did not sync those transactions until they were on the chain, then both transactions will have the block header timestamp.

this example might occur in practice if a node was offline when the transactions were first created and/or did not receive the transactions in gossip messages before they were added to the chain.

this might also happen for an account that is imported to a wallet after the transactions were added to the chain. when rescanning the chain for the imported account all transactions will use the timestamp of the block header of the block they are found on.

we use the 'timestampToTransactionHash' index to load transactions in order for the 'getAccountTransactions' RPC. the impact of the one-to-one datastore is that the RPC would miss transactions that the account has.

## Testing Plan

- adds unit test for loading multiple transactions from the same block
- manual testing of migration
- manual testing of `wallet:transactions` after migration

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/account.test.ts
```diff
@@ -1982,4 +1982,46 @@ describe('Accounts', () => {
...
+      await nodeA.chain.addBlock(block4)
+      await nodeB.chain.addBlock(block4)
+      await nodeA.wallet.updateHead()
+      await nodeB.wallet.updateHead()
+
+      const accountATx = await AsyncUtils.materialize(accountA.getTransactionsByTime())
+      const accountBTx = await AsyncUtils.materialize(accountB.getTransactionsByTime())
+
+      // 3 block rewards plus 2 outgoing transactions
+      expect(accountATx).toHaveLength(5)
```
danield9tqh, line 2021: Can we just add a little more expect here maybe expect each hash in these lists? Could expect ordering too

hughy: Good idea!

I added tests for the ordering of the transactions returned

danield9tqh (approved):

PR #3627, opened 2023-03-08
Add network id check for get fund rpc

ygao76:
## Testing Plan
unit test
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/faucet/getFunds.test.ts
+++ b/ironfish/src/rpc/routes/faucet/getFunds.test.ts
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import axios, { AxiosError } from 'axios'
+import { DEFAULT_NETWORK_ID } from '../../../fileStores'
 import { createRouteTest } from '../../../testUtilities/routeTest'
 import { RpcRequestError } from '../../clients'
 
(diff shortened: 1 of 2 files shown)


ironfish/src/rpc/routes/faucet/getFunds.ts
```diff
@@ -29,6 +30,14 @@ router.register<typeof GetFundsRequestSchema, GetFundsResponse>(
   `${ApiNamespace.faucet}/getFunds`,
   GetFundsRequestSchema,
   async (request, node): Promise<void> => {
+    // check node network id
+    const networkId = node.internal.get('networkId')
+
+    if (networkId !== DEFAULT_NETWORK_ID) {
```
dguenther, line 36: Let's hardcode the testnet network ID here, since we're planning to change `DEFAULT_NETWORK_ID` to mainnet when we launch mainnet.

I think it's okay to hardcode it since right now I think we hardcode the network IDs everywhere. (except for `DEFAULT_NETWORK_ID`). If you want, it could be useful in a separate PR to make an enum of Network IDs, so we can reference `NetworkID.Testnet` rather than just hardcoding `0`.

dguenther (changes requested):
One comment, otherwise looks good!

ironfish/src/rpc/routes/faucet/getFunds.ts
ygao76: I will hardcode the network id in this PR. I will make the enum of network Id as a follow up when I have time.

ironfish/src/rpc/routes/faucet/getFunds.test.ts
```diff
@@ -74,4 +76,15 @@ describe('Route faucet.getFunds', () => {
       )
     })
   })
+
+  describe('should fail when non testnet node', () => {
+    it('throws an error', async () => {
+      routeTest.node.internal.set('networkId', 2)
+      //const apiResponse = new Error('API failure') as AxiosError
+      //axios.post = jest.fn().mockRejectedValueOnce(apiResponse)
```
dguenther, line 84:
```suggestion

```

dguenther (approved):
Could you delete the commented code? Otherwise looks good!

PR #3630, opened 2023-03-09
Feature: RPC HTTP Adapter

EvanJRichard:
## Summary
This PR proposes to add `HttpAdapter` to `ironfish/src/rpc/adapters`. It is gated by `enableRpcHttp`, which defaults to `false`. 

To simplify things, everything will be a `POST` request.

## Testing Plan
WIP, unit tests + manual testing

## Breaking Change

```
[ ] Yes
```


ironfish/src/fileStores/config.ts
```diff
@@ -388,6 +397,8 @@ export class Config extends KeyStore<ConfigOptions> {
       rpcTcpPort: 8020,
       tlsKeyPath: files.resolve(files.join(dataDir, 'certs', 'node-key.pem')),
       tlsCertPath: files.resolve(files.join(dataDir, 'certs', 'node-cert.pem')),
+      rpcHttpHost: 'localhost',
```
NullSoldier, line 400: This doesn't work for ipv6

ironfish/src/fileStores/config.ts
```diff
@@ -388,6 +397,8 @@ export class Config extends KeyStore<ConfigOptions> {
...
+      rpcHttpPort: 8080,
```
NullSoldier, line 401: why not use port 80?

ironfish/src/rpc/adapters/httpAdapter.ts
```diff
@@ -0,0 +1,145 @@
...
+  server: http.Server | null = null
+  router: Router | null = null
+
+  readonly host: string
+  readonly port: number
+  readonly logger: Logger
+  readonly namespaces: ApiNamespace[]
+  readonly maxRequestSize: number
+
+  // TODO(daniel): implement https + basic authentication with rpcToken
```
NullSoldier, line 23: we can delete this, but after it merges we could enable HTTPS using the certs that exist

ironfish/src/rpc/adapters/httpAdapter.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  constructor(host: string, port: number, logger: Logger, namespaces: ApiNamespace[]) {
```
NullSoldier, line 25: make logger optional like all the other adapters

ironfish/src/rpc/adapters/httpAdapter.ts
```diff
@@ -0,0 +1,145 @@
...
+    this.host = host
+    this.port = port
+    this.logger = logger
+    this.namespaces = namespaces
+    // TODO(daniel): do we need a max message size and what should it be?
```
NullSoldier, line 30: make this configurable in the constructor by default here

ironfish/src/rpc/adapters/httpAdapter.ts
```diff
@@ -0,0 +1,184 @@
...
+          req.on('close', () => {
+            this.cleanUpRequest(reqId)
+          })
+
+          void this.handleRequest(req, res, reqId).catch((e) => {
+            // TODO: handle (e instanceof ResponseError) here?
+            const error = ErrorUtils.renderError(e)
+            this.logger.debug(`Error in HTTP adapter: ${error}`)
+            res.writeHead(500)
+            res.end(JSON.stringify({ error }))
```
NullSoldier, line 76: This is a non RPC uniform error. We should return a uniform RPC error so they don't have to handle this specially.

ironfish/src/rpc/adapters/httpAdapter.ts
```diff
@@ -0,0 +1,184 @@
...
+        resolve()
+      }
+
+      server.on('error', onError)
+      server.on('listening', onListening)
+      server.listen(this.port, this.host)
+    })
+  }
+
+  async stop(): Promise<void> {
```
NullSoldier, line 90: this isnt idempotent, look at socketAdapter

ironfish/src/rpc/adapters/httpAdapter.ts
```diff
@@ -0,0 +1,184 @@
...
+    for (const { req, rpcRequest } of this.requests.values()) {
+      req.destroy()
```
NullSoldier, line 92: https://nodejs.org/api/http.html#requestdestroyerror

ironfish/src/rpc/adapters/httpAdapter.ts
```diff
@@ -0,0 +1,184 @@
...
+    this.logger.debug(
+      `Call HTTP RPC: ${request.method || 'undefined'} ${request.url || 'undefined'}`,
+    )
+
+    // TODO(daniel): better way to parse method from request here
+    const url = new URL(request.url, `http://${request.headers.host || 'localhost'}`)
+    const route = url.pathname.substring(1)
+
+    if (request.method !== 'POST') {
+      response.writeHead(404, `Route does not exist, Did you mean to use POST?`)
```
NullSoldier, line 137: non standard RPC error

ironfish/src/fileStores/config.ts
EvanJRichard: won't this resolve just fine to `127.0.0.1` on an ipv4 system and `::1` on an ipv6 system? I don't quite follow @NullSoldier

ironfish/src/rpc/adapters/httpAdapter.ts
danield9tqh: Yea good suggestion. modified to handle more of our error types

ironfish/src/rpc/adapters/httpAdapter.ts
danield9tqh: Ok solved this in a little bit differently than the socketAdapter. This was slightly easier but lmk if you want to keep consistency with socketAdapter I can change it to that

NullSoldier (approved):

ironfish/src/fileStores/config.ts
NullSoldier: No we ran into an issue where if you listen on localhost, ipv6 connections won't work because of host node sockets work. We had to switch the pool host over. See this, https://github.com/iron-fish/ironfish/pull/3356/files

ironfish/src/rpc/adapters/httpAdapter.ts
NullSoldier: I'm not sure I see the solution but we can ignore this for now.

PR #3634, opened 2023-03-10
fixes genesis block transaction status

hughy:
## Summary

all transactions on the genesis block are confirmed.

'wallet:balance --all' shows a 'confirmed' balance of 0 for the genesis account when the chain contains only the genesis block.

fixes calculation of confirmed balance by ensuring that the range of unconfirmed sequences starts at at least 2

'wallet:transactions' shows a single 'unconfirmed' transaction for the genesis account when the chain contains only the genesis block.

fixes display of transaction status by checking if transaction has sequence equal to genesis block sequence

## Testing Plan

Before:
image

After:
image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/account.ts
+++ b/ironfish/src/wallet/account.ts
@@ -959,10 +959,8 @@ export class Account {
     if (confirmations > 0) {
       const unconfirmedSequenceEnd = headSequence
 
-      const unconfirmedSequenceStart = Math.max(
-        unconfirmedSequenceEnd - confirmations + 1,
-        GENESIS_BLOCK_SEQUENCE,
-      )
+      const unconfirmedSequenceStart =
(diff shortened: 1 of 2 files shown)


ygao76 (approved):
Looks good, can you test it with regular account so it does not break any common use cases.

hughy:
> Looks good, can you test it with regular account so it does not break any common use cases.

Tested with a wallet on the Phase 3 network.

Steps:
- Send transaction from account 'test' to default account, 'hughy-if', using '--watch' to watch transaction progress
- Wait until transaction is unconfirmed
- Stop node
- View balances and transactions list to confirm that transaction status is correctly tracked as unconfirmed
image

- Restart node, continue watching transaction with 'wallet:transaction:watch' until confirmed
- View balances and transactions list to see that the transaction is labeled as confirmed
image

PR #3637, opened 2023-03-10
Use confirmations when checking balance in sendTransaction

dguenther:
## Summary

I'm reasonably sure the custom confirmation range should be passed in here, since we pass it in to node.wallet.send.

I could use a review from someone familiar with the wallet code to double-check though.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):
yes this looks correct, good find

PR #3640, opened 2023-03-10
avoids deleting wallet in reset command

hughy:
## Summary

we likely do NOT want to delete users' accounts and keys after mainnet launch.

removes deletion of wallet database from 'reset' command. when the user starts a node on a new network after running reset, the wallet will automatically reset because the account heads will not be on the same chain as the network.

## Testing Plan

- manual testing

1. verify that node databases are on the wrong network
image

2. run `reset`
image

3. verify that accounts are still there
image

4. verify that wallet is reset once node starts successfully
image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/reset.ts
+++ b/ironfish-cli/src/commands/reset.ts
@@ -32,23 +32,9 @@ export default class Reset extends IronfishCommand {
   async start(): Promise<void> {
     const { flags } = await this.parse(Reset)
 
-    let confirmed = flags.confirm
-
-    const warningMessage =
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/reset.ts
```diff
@@ -57,12 +43,12 @@ export default class Reset extends IronfishCommand {
 
     const message =
       '\nYou are about to destroy your node databases. The following directories and files will be deleted:\n' +
```
dguenther, line 45: This is a rather minor nitpick, but there's only one database being deleted now, so maybe a slight reword:
```suggestion
      '\nYou are about to destroy your local copy of the blockchain. The following directories and files will be deleted:\n' +
```

dguenther (approved):

PR #3641, opened 2023-03-11
improves output of stop command

hughy:
## Summary

the 'stopNode' RPC shuts down the node before returning a response, so the cli client receives an error from the lost connection. IronfishCommand handles RPC connection errors and outputs 'Cannot connect to your node, start your node first.'. this is confusing both because we were able to connect to the node and because we successfully stopped it.

the stop command also silences RPC connection errors, so nothing is output if the node is already stopped when stop is run.

- returns a response from stopNode before shutting down the node
- removes error handling from 'client.connect' in stop command since the IronfishCommand abstract class handles these errors
- adds output to 'stop' command indicating that it asks the node to shut down

## Testing Plan

manual testing:
- start a node
- use 'ironfish stop' to stop the node
- use 'ironfish stop' to try to stop the node again once it has shut down

Before:
image

After:
image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/stop.ts
+++ b/ironfish-cli/src/commands/stop.ts
@@ -1,7 +1,8 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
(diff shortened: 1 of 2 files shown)


dguenther (approved):
Tested running it against a started node and a stopped node, looks good!

PR #3642, opened 2023-03-11
Add 'mainnet' command for migrating from testnet DB to mainnet DB

dguenther:
## Summary

Most testnet users will have created accounts on a node running a testnet DB, and we should expect they'll want to switch to mainnet in order to verify their airdrop.

In order to switch to mainnet, they'll need to reset their network ID in internal.json and delete their chain database. Alternatively, they can export their account, upgrade to 1.0, delete their data directory, then import their account.

We have an existing reset command that helps with this, but on 0.1.72 it also deletes the wallet database, which isn't ideal. We changed the command in #3640, but I'm hesitant to recommend it to users, since if they accidentally run the command before upgrading and don't read the logs, it'll delete their wallet. So, I added a `mainnet` command.

It's almost the same as the `reset` command, but with a few changes:
* It does nothing if the network ID is already set to 1 (mainnet).
* It tells the user to try to stop any running nodes if it encounters an error while deleting the database (they're probably in use)

## Testing Plan

* [x] Tested creating a data directory and two accounts with 0.1.65, then running `yarn start mainnet` on this branch
* [x] Tested running `yarn start mainnet` on this branch while another node is already running in the same data directory
* [x] Tested running `yarn start mainnet` on this branch with the network ID in internal.json set to 1 (mainnet)

## Documentation

```
[x] Yes
```

I'm working on a testnet-to-mainnet upgrade doc that will use this command.

## Breaking Change

```
[ ] Yes
```


dguenther:
With another node running on the same data directory:
image

With another node running on the same data directory and --verbose:
image

dguenther:
Tests need to be fixed in the base branch, so they won't pass here.

hughy (approved):
Looks good 👍 

Tested with a copy of my Phase 3 wallet from v0.1.70

charlie-wasp:
@dguenther sorry for bothering in obsolete PR, but I wonder, is this `mainnet` command still will be used after postponed launch? I see that `mainnet` branch is not updating anymore

PR #3645, opened 2023-03-13
Add startup flags for HTTP Adapter

danield9tqh:
## Summary
Adding flags `--rpc.tcp.port` and `--rpc.http.host` to ironfish commands to configure the HTTP adapter

## Testing Plan
Tested setting both flags locally

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/command.ts
```diff
@@ -17,6 +17,8 @@ import {
   ConfigFlagKey,
   DataDirFlagKey,
   RpcAuthFlagKey,
+  RpcHttpHostFlagKey,
+  RpcHttpPortFlagKey,
```
NullSoldier, line 21: I think we're still missing the flag to enable http

PR #3648, opened 2023-03-14
Move ceremony command to R2

danield9tqh:
## Summary
For cost saving we are moving future trusted setup ceremonies to use Cloudflare R2 instead of S3

## Testing Plan
Locally tested a few contributions

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/service/ceremony.ts
+++ b/ironfish-cli/src/commands/service/ceremony.ts
@@ -34,7 +34,8 @@ export default class Ceremony extends IronfishCommand {
       parse: (input: string) => Promise.resolve(input.trim()),
       required: false,
       description: 'Prefix for contribution download URLs',
-      default: 'https://a93bebf26da4c2fe205f71c896afcf89.r2.cloudflarestorage.com',
+      // TODO: update this to non-dev endpoint to avoid rate limiting
+      default: 'https://pub-6a239e04e140459087cf392ffc3245b1.r2.dev',
     }),
     contributionTimeoutMs: Flags.integer({
       required: false,
--- a/ironfish-cli/src/commands/service/ceremony.ts
+++ b/ironfish-cli/src/commands/service/ceremony.ts
@@ -11,7 +11,7 @@ import { S3Utils } from '../../utils'
 const CONTRIBUTE_TIMEOUT_MS = 5 * 60 * 1000
 const UPLOAD_TIMEOUT_MS = 5 * 60 * 1000
 const PRESIGNED_EXPIRATION_SEC = 5 * 60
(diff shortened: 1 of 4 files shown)


ironfish-cli/src/commands/service/ceremony.ts
```diff
@@ -29,6 +29,14 @@ export default class Ceremony extends IronfishCommand {
       description: 'S3 bucket to download and upload params to',
       default: 'ironfish-contributions',
     }),
+    downloadPrefix: Flags.string({
+      char: 'b',
+      parse: (input: string) => Promise.resolve(input.trim()),
+      required: false,
+      description: 'Prefix for contribution download URLs',
+      // TODO: update this to non-dev endpoint to avoid rate limiting
+      default: 'https://pub-6a239e04e140459087cf392ffc3245b1.r2.dev',
```
dguenther, line 38: This isn't the bucket, right? I assume it's like an account name

dguenther (approved):

ironfish-cli/src/utils/s3.ts
```diff
@@ -309,6 +318,31 @@ export function getS3Client(
...
+}): S3Client {
+  return new S3Client({
+    region: 'auto',
+    endpoint: R2_ENDPOINT,
+    credentials: {
+      accessKeyId: credentials.r2AccessKeyId,
+      secretAccessKey: credentials.r2SecretAccessKey,
+    },
+  })
+}
```
NullSoldier, line 333: These are binary compatible with S3, so I wouldnt expect this to exist. I would just expect these to be parameters that you pass into getS3Client but we can refactor this later.

NullSoldier (approved):

ironfish-cli/src/commands/service/ceremony.ts
ygao76: its the dev testing url. We will replace it with ironfish.network domain name when close to launch.

ironfish-cli/src/utils/s3.ts
```diff
@@ -309,6 +318,31 @@ export function getS3Client(
...
+
+export async function getR2Credentials(): Promise<R2Secret | undefined> {
+  const client = new SecretsManagerClient({ region: 'us-east-1' })
```
ygao76, line 336: I am not sure about this region config. If the ceremony server has terraform config, it prob wont need it. 
Maybe try it without region on the github workflow

ironfish-cli/src/commands/service/ceremony.ts
danield9tqh: Yea that is basically a link to the bucket

PR #3654, opened 2023-03-15
fixes rescanning wallet without --reset

hughy:
## Summary

the logic of 'connectBlock' filters any account that does not have its 'head' equal to the previous block. this means that if we don't reset account heads during rescan nothing will happen.

this is also true of rescanning using the '--from' flag. this flag lets users start the rescan at a particular sequence. however, if we don't set account heads to that sequence then scanning won't do anything.

- removes the 'reset' option and always resets accounts
- throws an error if the 'from' sequence is not in the chain or the header at that sequence is not on the main chain
- sets account heads equal to the block header before the from sequence if it is not the genesis block

## Testing Plan

- updated unit tests
- manual testing

before:
image
Finishes scanning almost immediately and then hits an error

after:
image

## Documentation

```
[X] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/rescan.ts
+++ b/ironfish-cli/src/commands/wallet/rescan.ts
@@ -10,7 +10,7 @@ import { ProgressBar } from '../../types'
 import { hasUserResponseError } from '../../utils'
 
 export class RescanCommand extends IronfishCommand {
-  static description = `Rescan the blockchain for transaction`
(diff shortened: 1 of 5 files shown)


EvanJRichard (approved):
Removing redundant rpc options? Increasing unit testing? I dig it. :shipit:

PR #3655, opened 2023-03-15
converts account createdAt to store a HeadValue

hughy:
## Summary

we added the createdAt field to AccountValue with a plan in mind to use it for timestamp-based account birthdays. the createdAt field could be used to speed up scanning by skipping blocks with a timestamp earlier than the account's createdAt.

however, we can't rely on block timestamps in this way: we cannot guarantee that the timestamp for a given block is not far before the wall clock time that that block was mined at.

to support block-based account birthdays we will store the chain head (block hash and sequence) at the time the account was created.

- changes the type of createdAt from Date to HeadValue
- updates account import and export for changes to createdAt field

## Testing Plan

- updated unit test for account serde
- manual testing
  - created account, exported it, deleted it, and re-imported it
  - imported account from v0.1.65

## Documentation

```
[X] Yes
```

## Breaking Change

```
[ ] Yes
```


EvanJRichard (approved):
I was a little sketched out by the changes to import, export, and serde, but the tests convinced me. Is it necessary that we regenerate the fixtures? If so, LGTM

hughy:
>  Is it necessary that we regenerate the fixtures?

It's a little overkill to regenerate all of them 😅 

We really only need to regenerate the existing account fixtures since they were all generated with a date in the `createdAt` field. It's easier just to regenerate everything though!

PR #3656, opened 2023-03-15
migrates account createdAt from Date to HeadValue

hughy:
## Summary

to support block-based account birthdays we will store the chain head (block hash and sequence) at the time the account was created.

adds migration to set createdAt to null for all existing accounts

## Testing Plan

- ran data migration and verified that createdAt was set to null with no other account data affected

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/migrations/data/027-account-created-at-block.ts
+++ b/ironfish/src/migrations/data/027-account-created-at-block.ts
@@ -5,7 +5,6 @@ import { Logger } from '../../logger'
 import { IronfishNode } from '../../node'
 import { IDatabase, IDatabaseTransaction } from '../../storage'
 import { createDB } from '../../storage/utils'
-import { Account } from '../../wallet'
 import { Migration } from '../migration'
 import { GetStores } from './027-account-created-at-block/stores'
 
@@ -27,18 +26,26 @@ export class Migration027 extends Migration {
     logger.info(`Migrating account data to store block-based creation time`)
 
     for await (const accountValue of stores.old.accounts.getAllValuesIter(tx)) {
-      const account = new Account({
-        ...accountValue,
-        createdAt: null,
-        walletDb: node.wallet.walletDb,
-      })
+      logger.info(` Migrating account ${accountValue.name}`)
 
-      logger.info(` Migrating account ${account.name}`)
-
-      await stores.new.accounts.put(account.id, account, tx)
+      await stores.new.accounts.put(accountValue.id, { ...accountValue, createdAt: null }, tx)
     }
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-empty-function
-  async backward(): Promise<void> {}
+  async backward(
+    node: IronfishNode,
+    db: IDatabase,
+    tx: IDatabaseTransaction | undefined,
+    logger: Logger,
+  ): Promise<void> {
+    const stores = GetStores(db)
+
+    logger.info(`Reverting migration of account data to store block-based creation time`)
+
+    for await (const accountValue of stores.new.accounts.getAllValuesIter(tx)) {
+      logger.info(` Reverting migration for account ${accountValue.name}`)
+
+      await stores.old.accounts.put(accountValue.id, { ...accountValue, createdAt: null }, tx)
+    }
+  }
 }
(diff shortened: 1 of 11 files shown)


ironfish/src/migrations/data/027-account-created-at-block.ts
```diff
@@ -0,0 +1,51 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
EvanJRichard, line 1: I recommend getting a second set of eyes on this migration. LGTM, but the migration I wrote didn't go well, so it's worth getting more scrutiny.

ironfish/src/migrations/data/027-account-created-at-block.ts
```diff
@@ -0,0 +1,51 @@
...
+    logger: Logger,
+  ): Promise<void> {
+    const stores = GetStores(db)
+
+    logger.info(`Reverting migration of account data to store block-based creation time`)
+
+    for await (const accountValue of stores.new.accounts.getAllValuesIter(tx)) {
+      logger.info(` Reverting migration for account ${accountValue.name}`)
+
+      await stores.old.accounts.put(accountValue.id, { ...accountValue, createdAt: null }, tx)
```
EvanJRichard, line 48: should this omit `createdAt` or no? seems interesting that both `forward` and `backward` do the same `put` operation

ironfish/src/migrations/data/027-account-created-at-block/old/index.ts
```diff
@@ -0,0 +1,20 @@
...
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { IDatabase, IDatabaseStore, StringEncoding } from '../../../../storage'
+import { AccountValue, AccountValueEncoding } from './accountValue'
+
+export function GetOldStores(db: IDatabase): {
+  accounts: IDatabaseStore<{ key: string; value: AccountValue }>
+} {
+  const accounts: IDatabaseStore<{ key: string; value: AccountValue }> = db.addStore(
+    {
+      name: 'a',
```
EvanJRichard, line 12: Thinking out loud so it's in someone else's head: it's totally outside the scope of this PR but in the future we really should change this `'a'` to an enum or a constant or something

ironfish/src/migrations/data/027-account-created-at-block/new/headValue.ts
```diff
@@ -0,0 +1,41 @@
...
+import bufio from 'bufio'
+import { IDatabaseEncoding } from '../../../../storage'
+
+export type HeadValue = {
+  hash: Buffer
+  sequence: number
+}
+
+export class NullableHeadValueEncoding implements IDatabaseEncoding<HeadValue | null> {
+  readonly nonNullSize = 32 + 4
```
EvanJRichard, line 13: can and should `nonNullSize` be derived from other constants or anything? I think the answer is "no, literal `32 + 4` is what makes the most sense here" but worth a double-check

ironfish/src/migrations/data/027-account-created-at-block/new/accountValue.ts
```diff
@@ -0,0 +1,108 @@
...
+      outgoingViewKey,
+      spendingKey,
+      publicAddress,
+      createdAt,
+    }
+  }
+
+  getSize(value: AccountValue): number {
+    let size = 0
+    size += 1
```
EvanJRichard, line 90: why this magic `+= 1`?

EvanJRichard (approved):
Migrations make me sweat, might be worth getting even more review ----- but it does LGTM! A couple questions and nits, no blockers.

ironfish/src/migrations/data/027-account-created-at-block/new/accountValue.ts
hughy: The `1` is for the byte that stores flags. I think we could improve the readability of the size calculations in some of these encodings with comments on this sort of thing

ironfish/src/migrations/data/027-account-created-at-block/new/headValue.ts
hughy: I looked around for any constants we have for the length of hashes (32 bytes), but didn't find any 🤔 I suspect that if we add one we'll want to replace a lot of `32` literals

For the `4` I think the literal `4` might make the most sense since it's just referring to the byte length of the 32-bit integer

NullSoldier (approved):

NullSoldier:
This looks good to me.

PR #3659, opened 2023-03-16
starts scanTransactions after account creation

hughy:
## Summary

if a node is online and syncing at the time of account creation it's possible to for an account head to be set to a block that is behind the wallet's chainProcessor.

if the account's head is behind the chain processor then the wallet will not scan any transactions for the account until the node is restarted.

runs scanTransactions after creating an account in the 'create' RPC so that a newly created account will be synced with the chainProcessor if it is behind. this matches the logic in the 'importAccount' RPC.

## Testing Plan

manual testing:
- start a new node so that the node is syncing
- create a new account
- use 'wallet:status' to see the the new account's head is moving forward with the default account

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/create.test.slow.ts
+++ b/ironfish/src/rpc/routes/wallet/create.test.slow.ts
@@ -14,9 +14,11 @@ describe('Route wallet/create', () => {
   jest.setTimeout(15000)
   const routeTest = createRouteTest()
 
-  it('should create an account', async () => {
+  beforeEach(() => {
     jest.spyOn(routeTest.node.wallet, 'scanTransactions').mockReturnValue(Promise.resolve())
+  })
 
+  it('should create an account', async () => {
     await routeTest.node.wallet.createAccount('existingAccount', true)
 
     const name = uuid()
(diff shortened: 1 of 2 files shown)


ironfish/src/rpc/routes/wallet/create.test.slow.ts
```diff
@@ -13,7 +13,10 @@ import { RpcRequestError } from '../../clients/errors'
 describe('Route wallet/create', () => {
   jest.setTimeout(15000)
   const routeTest = createRouteTest()
+
   it('should create an account', async () => {
+    jest.spyOn(routeTest.node.wallet, 'scanTransactions').mockReturnValue(Promise.resolve())
```
holahula, line 18: can these all be converted into a single
```
jest.beforeEach(() => {
    jest.spyOn(routeTest.node.wallet, 'scanTransactions').mockReturnValue(Promise.resolve())
})
```

hughy: updated to use `beforeEach`!

PR #3660, opened 2023-03-17
sets chainProcessor sequence on wallet load

hughy:
## Summary

the chainProcessor tracks the head of the chain using both the hash and sequence of a block. when the wallet loads we currently only set the has of the chainProcessor equal to the latest headd hash from among the accounts in the wallet.

one consequence of not setting the sequence at load time is that if a user creates an account while the node is offline, the chainProcessor is not fully initialized and the newly created account has its head set to null instead of set equal to the latest head in the wallet. the new account then needs to scan the whole chain  to sync, even though it was just created and won't have any transactions earlier in the chain.

- defines getLatestHead to load the latest head (hash and sequence) from among all accounts
- sets both the hash and sequence of the chainProcessor from the latest head

## Testing Plan

- adds unit test for wallet load
- manual testing:
  - use a node that has synced at least some blocks in a chain
  - with the node offline, run 'wallet:create' to create a new account
  - run 'wallet:status' to see that the head of the new account is equal to the latest head in the wallet

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -599,7 +599,7 @@ export class Wallet {
 
     if (this.chainProcessor.hash === null) {
       const latestHead = await this.getLatestHead()
-      Assert.isNotNull(latestHead, `scanTransactions: No latest head hash found`)
+      Assert.isNotNull(latestHead, `scanTransactions: No latest head found`)
 
       this.chainProcessor.hash = latestHead.hash
       this.chainProcessor.sequence = latestHead.sequence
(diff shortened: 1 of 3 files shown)


ironfish/src/wallet/wallet.ts
```diff
@@ -592,10 +598,11 @@ export class Wallet {
     }
 
     if (this.chainProcessor.hash === null) {
-      const latestHeadHash = await this.getLatestHeadHash()
-      Assert.isNotNull(latestHeadHash, `scanTransactions: No latest head hash found`)
+      const latestHead = await this.getLatestHead()
+      Assert.isNotNull(latestHead, `scanTransactions: No latest head hash found`)
```
EvanJRichard, line 602: Is it worth it to change this assert error to be more precise (` no head found`, because the code finds the head and then accesses the hash) or nah?

EvanJRichard (approved):
Seems straightforward to me! :ship:

ironfish/src/wallet/wallet.ts
hughy: Good catch, I think it's worth it!

PR #3665, opened 2023-03-17
Add noteSize and noteCommitment to getBlock RPC

danield9tqh:
## Summary
Adding note size and commitment to the block header RPC

## Testing Plan
Local testing

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/chain/getBlock.ts
+++ b/ironfish/src/rpc/routes/chain/getBlock.ts
@@ -23,6 +23,8 @@ export type GetBlockResponse = {
     previousBlockHash: string
     sequence: number
     timestamp: number
+    noteSize: number
+    noteCommitment: string
     transactions: Array<{
       fee: string
       hash: string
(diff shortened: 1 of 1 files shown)


ironfish/src/rpc/routes/chain/getBlock.ts
```diff
@@ -124,6 +128,10 @@ router.register<typeof GetBlockRequestSchema, GetBlockResponse>(
       throw new ValidationError(error)
     }
 
+    if (header.noteSize === null) {
```
danield9tqh, line 131: Adding this here because we [never save to database without the noteSize](https://github.com/iron-fish/ironfish/blob/8b765694d9350e19b82ff6886457f9ff992b7fef/ironfish/src/blockchain/database/headers.ts#L18). Ideal solution would be to have separate types for a network block header and a database block header but we don't have that yet. This should never be null so failing loudly here

PR #3666, opened 2023-03-18
Add RPC route to get note witness

danield9tqh:
## Summary
Adding RPC route to get the witness for a specific note index.

## Testing Plan
Unit test and local testing

## Documentation

Documentation: https://github.com/iron-fish/website/pull/342

```
[X] Yes
```

## Breaking Change

```
[ ] Yes
```


hairtail:
This is really helpful. With this pr, we can create transaction and sign a transaction locally i guess. There is one more route `chain/postTransaction` needed i think. I would like to provide service with all `chain/xxx` routes but i will disable rpc call for to `wallet/xxx` for my node.

NullSoldier:
nice can you also add the documentation for this too?

ironfish/src/rpc/routes/chain/getNoteWitness.ts
```diff
@@ -0,0 +1,78 @@
...
+import { ApiNamespace, router } from '../router'
+
+export type GetNoteWitnessRequest = {
+  index: number
+}
+
+export type GetNoteWitnessResponse = {
+  noteSize: number
+  rootHash: string
+  authPathLength: number
```
NullSoldier, line 15: we don't need this since its the length of authPath right?

ironfish/src/rpc/routes/chain/getNoteWitness.ts
```diff
@@ -0,0 +1,78 @@
...
+  authPath: {
+    side: 'Left' | 'Right'
+    hashOfSibling: string
+  }[]
+}
+
+export const GetNoteWitnessRequestSchema: yup.ObjectSchema<GetNoteWitnessRequest> = yup
+  .object()
+  .shape({
```
NullSoldier, line 24: I think you just want to pass this to object() and not use shape()

ironfish/src/rpc/routes/chain/getNoteWitness.ts
```diff
@@ -0,0 +1,78 @@
...
+import * as yup from 'yup'
+import { ValidationError } from '../../adapters'
+import { ApiNamespace, router } from '../router'
+
+export type GetNoteWitnessRequest = {
+  index: number
+}
+
+export type GetNoteWitnessResponse = {
+  noteSize: number
```
NullSoldier, line 13: this is usually called treeSize in our code so I think this will be confusing since it's suddenly a new concept, and also it's not the size of a note.

ironfish/src/rpc/routes/chain/getNoteWitness.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    const authPath = witness.authenticationPath.map((step) => {
+      return {
+        side: step.side,
+        hashOfSibling: step.hashOfSibling.toString('hex'),
+      }
+    })
+
+    request.end({
+      noteSize,
```
NullSoldier, line 72: This is the wrong value here, should be this
```suggestion
      noteSize: witness.treeSize(),
```

ironfish/src/rpc/routes/chain/getNoteWitness.ts
```diff
@@ -0,0 +1,78 @@
...
+      rootHash: rootHash.toString('hex'),
```
NullSoldier, line 73: this is the wrong value here, it should be witness.rootHash.toString('hex')

```suggestion
      rootHash: witness.rootHash.toString('hex'),
```

ironfish/src/rpc/routes/chain/getNoteWitness.ts
danield9tqh: yea probably, not needed I'll remove

ironfish/src/rpc/routes/chain/getNoteWitness.ts
danield9tqh: yea thanks for catching this

ironfish/src/rpc/routes/chain/getNoteWitness.ts
danield9tqh: thanks for catching

danield9tqh:
@hairtail the `wallet/postTransaction` also marks the notes in that transaction as spent. Where would you keep track of unspent notes if not in the wallet? Are you doing that outside the node currently?

rohanjadvani (approved):

hairtail:
> @hairtail the `wallet/postTransaction` also marks the notes in that transaction as spent. Where would you keep track of unspent notes if not in the wallet? Are you doing that outside the node currently?

I want to send transaction with a known note which called `causal send`. That to say, 
1. I receive some token in transaction A.
2. I want to send the received token in transaction A to address X in transaction B locally with these rpc calls without wallet import.
3. The consensus is responsible for the validation of transaction B.

danield9tqh:
@hairtail We do have some of that functionality but it is a little confusing currently. The one thing we don't have is an endpoint to create + sign a raw transaction from a list of notes + outputs which I think is what you're asking for. I think @hughy is going to pick this up in the coming weeks. 

We also have a endpoint that you can broadcast a signed transaction. That's in `wallet/addTransaction`

PR #3670, opened 2023-03-20
Fix request closing early

danield9tqh:
## Summary
Fixing a bug where the request close callback happens immediately. `req` object represents data coming from the client so once the client has streamed it's request 'close' event will be emitted. We want to wait until the client disconnects or the request is completed. This would be on the `res.on('close')` event. The response stream is separate from the request stream.

## Testing Plan
Tested locally

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


EvanJRichard (approved):
Good fix

PR #3675, opened 2023-03-20
updates account createdAt if block disconnected

hughy:
## Summary

when using block-based account birthdays we need to ensure that the account birthday always refers to a block in the main chain.

on a block disconnect checks whether the account's createdAt field refers to the disconnected block. if it does, updates the account's createdAt field to the previous block. this ensures that createdAt refers to a block in the main chain at the end of a re-org.

- defines updateAccount to update the account stored in the walletDb and in the wallet's in-memory cache
- adds conditional update at the end of disconnectBlock

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/__fixtures__/wallet.test.ts.fixture
+++ b/ironfish/src/wallet/__fixtures__/wallet.test.ts.fixture
@@ -4969,5 +4969,87 @@
         "sequence": 3
       }
     }
+  ],
+  "Accounts disconnectBlock should update an account createdAt field if that block is disconnected": [
+    {
+      "version": 1,
+      "id": "1f36ba0d-fc27-4e21-93ec-f74d90aee106",
+      "name": "accountA",
+      "spendingKey": "9cebffd7d3a2ab3132264fae01cdaebb8fc5f627b4cc110929452ac4cc2b9964",
+      "viewKey": "3de1429fef73ab4317333dd7836a92b4a2464b5288ea8122d8005e9180f3b8658ff8085f655156a0ff76b5d14492d844f3b40ebce32b5d5430c97438acf21bb7",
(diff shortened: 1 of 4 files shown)


EvanJRichard (approved):
Non-blocking question: would we get value out of adding a test for a multi-block re-org? This looks like it would handle a chain like
```
A -> B -> C_(created_at) -> D
  -> E -> F -> G
```
What I mean is, it seems like it would handle a reorg from the A-B-C-D chain to A-E-F-G just fine, but should we add an explicit test for it? ("Nah, that's redundant with the existing test" is a fine answer if you think so, you've been staring at this stuff for way longer than me)

hughy:
> Non-blocking question: would we get value out of adding a test for a multi-block re-org? This looks like it would handle a chain like
> 
> ```
> A -> B -> C_(created_at) -> D
>   -> E -> F -> G
> ```
> 
> What I mean is, it seems like it would handle a reorg from the A-B-C-D chain to A-E-F-G just fine, but should we add an explicit test for it? ("Nah, that's redundant with the existing test" is a fine answer if you think so, you've been staring at this stuff for way longer than me)

I think it's a good idea to have a test with a longer re-org!

We've had a couple of instances where testing only with short chains and/or small values has missed bugs

hughy:
> Non-blocking question: would we get value out of adding a test for a multi-block re-org? This looks like it would handle a chain like
> 
> ```
> A -> B -> C_(created_at) -> D
>   -> E -> F -> G
> ```
> 
> What I mean is, it seems like it would handle a reorg from the A-B-C-D chain to A-E-F-G just fine, but should we add an explicit test for it? ("Nah, that's redundant with the existing test" is a fine answer if you think so, you've been staring at this stuff for way longer than me)

Added a test with this fork case:

```
A1 -> A2 -> A3 (createdAt)
     -> B2 -> B3 -> B4
```
re-org to
```
A1 (createdAt)
     -> B2 -> B3 -> B4
```

PR #3676, opened 2023-03-21
Call forceCleanupDeletedAccounts at end of wallet:prune

EvanJRichard:
## Testing Plan
Ran `wallet:prune` without incident.

## Documentation

```
[ ] Yes
[X] No
```

## Breaking Change

```
[ ] Yes
[X] No
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/prune.ts
+++ b/ironfish-cli/src/commands/wallet/prune.ts
@@ -71,6 +71,10 @@ export default class PruneCommand extends IronfishCommand {
       CliUx.ux.action.stop()
     }
 
+    CliUx.ux.action.start(`Cleaning up deleted accounts`)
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/wallet/prune.ts
```diff
@@ -71,6 +71,10 @@ export default class PruneCommand extends IronfishCommand {
       CliUx.ux.action.stop()
     }
 
+    CliUx.ux.action.start(`Cleaning up deleted accounts`)
+    await node.wallet.forceCleanupDeletedAccounts()
```
hughy, line 75: We should probably do this before the call to `compact`, above, to get the most we can out of compaction

hughy (approved):
LGTM! Tested locally - created an account, deleted it, pruned it

PR #3677, opened 2023-03-21
sets createdAt on first on-chain transaction

hughy:
## Summary

even if we have an account without a birthday we can use the createdAt field to save time in future wallet rescans by setting createdAt the first time the wallet connects a block that contains a transaction for that account.

updates connectBlock to set account.createdAt if the account had a transaction on the block and createdAt is null. if assetBalanceDeltas is not empty then the account had a transaction on the block.

## Testing Plan

- adds unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/__fixtures__/wallet.test.ts.fixture
+++ b/ironfish/src/wallet/__fixtures__/wallet.test.ts.fixture
@@ -5237,5 +5237,159 @@
         }
       ]
     }
+  ],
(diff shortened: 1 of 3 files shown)


EvanJRichard (approved):
These small digestible PRs make me so happy :') 54 lines of tests in response to an 8 line logic change we love to see it :')

PR #3679, opened 2023-03-22
IFL-27: resets account birthdays if not in chain

hughy:
## Summary

we need to ensure that the createdAt field of each account always refers to a block on the main chain (or is null). if we start the node and an account's createdAt block is not on the node's chain then we can't be sure that it ever will be (it could be on a fork at any time before or after the node's current head). in this case we need to reset the account's createdAt to null.

checks that createdAt is on chain for each account on wallet start and resets account data (including createdAt) if not

Closes IFL-27

## Testing Plan

- added unit tests
- manual testing
  - created account
  - used `ironfish repl` to set createdAt to fake block
  - started node again
  - used `ironfish repl` to see that account's createdAt reset to null

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -230,7 +230,11 @@ export class Wallet {
     this.isStarted = true
 
     for (const account of this.listAccounts()) {
-      if (account.createdAt === null) {
+      if (account.createdAt === null || this.chainProcessor.sequence === null) {
+        continue
+      }
+
+      if (account.createdAt.sequence > this.chainProcessor.sequence) {
         continue
       }
 
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -229,6 +229,16 @@ export class Wallet {
     }
     this.isStarted = true
 
+    for (const account of this.listAccounts()) {
+      if (account.createdAt === null) {
+        continue
+      }
+
+      if (!(await this.chain.hasBlock(account.createdAt.hash))) {
+        await this.resetAccount(account, { resetCreatedAt: true })
+      }
+    }
+
     if (this.chainProcessor.hash) {
       const hasHeadBlock = await this.chain.hasBlock(this.chainProcessor.hash)
 
@@ -296,7 +306,7 @@ export class Wallet {
 
   private async resetAccounts(tx?: IDatabaseTransaction): Promise<void> {
     for (const account of this.listAccounts()) {
-      await this.resetAccount(account, tx)
+      await this.resetAccount(account, { tx })
     }
   }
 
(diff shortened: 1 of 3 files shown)


EvanJRichard (approved):
TIL how to use `ironfish repl`. Anyways, a straightforward change.

I spent a long time staring at the move to using `options?` from `tx?` in `resetAccount`....something about that stands out to me, but I think it's that I just personally don't like the typical typescript pattern of using an `options` object. Standard practice, so not recommending you change it.

ironfish/src/wallet/wallet.ts
```diff
@@ -1374,15 +1384,21 @@ export class Wallet {
     return this.getAccountByName(name) !== null
   }
 
-  async resetAccount(account: Account, tx?: IDatabaseTransaction): Promise<void> {
+  async resetAccount(
+    account: Account,
+    options?: {
+      resetCreatedAt?: boolean
+      tx?: IDatabaseTransaction
```
NullSoldier, line 1391: I think the tx should be outside like `getTransactionStatus`

ironfish/src/wallet/wallet.ts
```diff
@@ -229,6 +229,16 @@ export class Wallet {
     }
     this.isStarted = true
 
+    for (const account of this.listAccounts()) {
+      if (account.createdAt === null) {
+        continue
+      }
+
+      if (!(await this.chain.hasBlock(account.createdAt.hash))) {
```
NullSoldier, line 237: What if we import this into an wallet that hasn't synced the chain, it'll clear the createdAt even if the block will exist later?

hughy: It will clear createdAt in this case since we can't know for sure if the block will exist later. 

We could check `createdAt.sequence` to check whether createdAt might be ahead of the chain an only clear createdAt if its sequence is behind the current head and the block isn't on the chain.

If we do that I think we'll also need to update the scan logic to check whether createdAt is on the main chain when we get to that sequence and reset if it isn't.

NullSoldier: That sounds good, but is there any danger to just letting it stay as a block that's not on the same chain?

Is it that you might skip blocks between a fork point and the current head when you should check those for transactions?

hughy: > Is it that you might skip blocks between a fork point and the current head when you should check those for transactions?

Yeah exactly, if the account was created on a fork or on another chain (e.g., testnet), then scanning might skip blocks that we should check.

hughy:
> What if we import this into an wallet that hasn't synced the chain, it'll clear the createdAt even if the block will exist later?

Updated to only reset accounts if `createdAt` has a sequence less than or equal to the wallet's chain processor. I'll include the logic for resetting the account if the chain reaches its birthday and the birthday is not on the chain in the next PR (#3685)

PR #3681, opened 2023-03-22
removes account hasSpend method

hughy:
## Summary

when adding a transaction to the wallet we use 'hasSpend' to iterate over all spends in the transaction to see whether the account spent any notes in the transaction before using 'addPendingTransaction' or 'connectTransaction'.

however, both of those methods _also_ iterate over all spends in the transaction. this can result in unnecessary, duplicative iteration over spends.

there is another drawback to the way we currently use 'hasSpend': the account 'addPendingTransaction' and 'connectTransaction' methods depend on the caller not to call them with any account and transaction where the account was not involved in the transaction. this is a bug waiting to happen.

- removes 'hasSpend' method
- does not check 'hasSpend' at wallet level before adding transaction to account
- updates account 'addPendingTransaction' and 'connectTransaction' methods to return early if the account did not receive or spend any notes in the transaction

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/wallet.test.ts
+++ b/ironfish/src/wallet/wallet.test.ts
@@ -1343,28 +1343,6 @@ describe('Accounts', () => {
   })
 
   describe('addPendingTransaction', () => {
-    it('should add transactions to accounts involved in the transaction', async () => {
-      const { node } = await nodeTest.createSetup()
-
-      const accountA = await useAccountFixture(node.wallet, 'a')
-      const accountB = await useAccountFixture(node.wallet, 'b')
-
-      const blockA1 = await useMinerBlockFixture(node.chain, undefined, accountA, node.wallet)
-      await expect(node.chain).toAddBlock(blockA1)
-      await node.wallet.updateHead()
-
-      const addSpyA = jest.spyOn(accountA, 'addPendingTransaction')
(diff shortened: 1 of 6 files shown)


ironfish/src/wallet/wallet.test.ts
```diff
@@ -1343,28 +1343,6 @@ describe('Accounts', () => {
...
-
-      await useTxFixture(node.wallet, accountA, accountA)
-
-      // tx added to accountA
-      expect(addSpyA).toHaveBeenCalledTimes(1)
-
-      // tx not added to accountB
-      expect(addSpyB).toHaveBeenCalledTimes(0)
-    })
-
```
EvanJRichard, line 1367: I was sketched out we were removing a test, but I see this is mirrored/replaced over in `account.test.ts`

EvanJRichard (approved):
Looking good 👍

hughy:
> @hughy Is the test plan just the unit tests? Pulled it down and nothing's obviously wrong, just wondering

In this case I believe that unit tests are sufficient for testing 😁 

These changes shouldn't change any of the behavior of the wallet: they only move a bit of logic from the wallet-level down to the account-level

PR #3689, opened 2023-03-23
Tag docker deploys with Github info

danield9tqh:
## Summary
When we deploy any docker images either from master or from another branch we only tag with `latest`. This can run into potential issues with production deploys if we don't always deploy from the master branch. Tagging docker images with their Github hash+branch / tag name will make our production deployments more stable

## Testing Plan
Testing by running Github action

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/scripts/deploy-docker.sh
+++ b/ironfish-cli/scripts/deploy-docker.sh
@@ -12,7 +12,18 @@ if [ -z "${PACKAGE_NAME-}" ]; then
     exit 1
 fi
 
+if [ -z "${GITHUB_SHA-}" ]; then
+    echo "Set GITHUB_SHA before running deploy-docker.sh"
+    exit 1
+fi
+
+if [ -z "${GITHUB_REF-}" ]; then
+    echo "Set GITHUB_REF before running deploy-docker.sh"
+    exit 1
+fi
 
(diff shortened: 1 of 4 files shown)


ironfish-cli/scripts/deploy-docker.sh
```diff
@@ -12,7 +12,18 @@ if [ -z "${PACKAGE_NAME-}" ]; then
...
+    echo "Set GITHUB_SHA before running deploy-docker.sh"
+    exit 1
+fi
+
+if [ -z "${GITHUB_REF-}" ]; then
+    echo "Set GITHUB_REF before running deploy-docker.sh"
+    exit 1
+fi
 
 docker tag ironfish:latest ${REGISTRY_URL}/${PACKAGE_NAME}:latest
```
hughy, line 25: Should we only use the `latest` tag if building from master? It seems like this will help distinguish images built on other branches, but it won't solve the problem of `latest` coming from a different branch.

danield9tqh: Still tagging it with latest atm so nothing unexpected breaks. After this merges I will go through terraform and migrate to using the master tag. Then we can remove latest. Good suggestion tho

hughy: Sounds good to me!

hughy (approved):

PR #3690, opened 2023-03-23
feat(cli): reset CLI command takes networkId flag and only modifies if given

mat-if:
## Summary

Currently, there is no way to modify your network ID unless you manually edit the internal.json file. Combined with the usage of reset changing slightly to work more like a soft reset to allow changing between networks or just reset chain data, it makes sense to only optionally re-write the network ID if passed via flag. Otherwise, leave it unmodified

## Testing Plan

image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/reset.ts
+++ b/ironfish-cli/src/commands/reset.ts
@@ -43,10 +43,13 @@ export default class Reset extends IronfishCommand {
       HOST_FILE_NAME,
     )
 
+    const existingId = this.sdk.internal.get('networkId')
+
     let networkIdMessage = ''
-    if (flags.networkId != null) {
-      const existingId = this.sdk.internal.get('networkId')
+    if (flags.networkId != null && flags.networkId !== existingId) {
       networkIdMessage = `\n\nThe network ID will be changed from ${existingId} to the new value of ${flags.networkId}`
+    } else {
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/reset.ts
```diff
@@ -38,11 +43,18 @@ export default class Reset extends IronfishCommand {
       HOST_FILE_NAME,
     )
 
+    let networkIdMessage = ''
+    if (flags.networkId != null) {
+      const existingId = this.sdk.internal.get('networkId')
+      networkIdMessage = `\n\nThe network ID will be changed from ${existingId} to the new value of ${flags.networkId}`
```
hughy, line 49: nit - do we still want to print this message if `existingId === flags.networkId`?

hughy (approved):

ironfish-cli/src/commands/reset.ts
dguenther: I don't have evidence to support one option being better -- I think skipping the message would be a bit unclear as to whether a change will happen or not, but I'm not sure if it's more helpful to have a message that says `Your network ID will stay unchanged as ${existingId}`. I'm okay with it either way and changing it if we get user feedback.

mat-if: Good call - that makes it clear in all scenarios. Change incoming

PR #3691, opened 2023-03-24
Fix GitHub branch ref

danield9tqh:
## Summary
`GITHUB_REF` is not exactly the branch name so extracting it here

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


hughy (commented):
We might be able to use `GITHUB_REF_NAME` instead. That might support branch names that contain a `/`

hughy (approved):

PR #3695, opened 2023-03-24
Add runtime flag for json logging

danield9tqh:
## Summary
When programmatically starting up a node for the first time like we are doing in the simulator, a config value may not be set. This allows logging easily parable logs (with tags included) on startup without having to modify a config file first.

## Testing Plan
Tested locally

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/command.ts
+++ b/ironfish-cli/src/command.ts
@@ -16,6 +16,8 @@ import { CLIError, ExitError } from '@oclif/core/lib/errors'
 import {
   ConfigFlagKey,
   DataDirFlagKey,
+  JsonLogsFlag,
+  JsonLogsFlagKey,
   RpcAuthFlagKey,
   RpcHttpHostFlagKey,
   RpcHttpPortFlagKey,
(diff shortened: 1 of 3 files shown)


ironfish-cli/src/command.ts
```diff
@@ -167,6 +170,11 @@ export abstract class IronfishCommand extends Command {
       configOverrides.logLevel = '*:verbose'
     }
 
+    const jsonLogsFlag = getFlag(flags, JsonLogsFlagKey)
+    if (typeof jsonLogsFlag === 'boolean' && jsonLogsFlag !== JsonLogsFlag.default) {
+      configOverrides.jsonLogs = jsonLogsFlag
+    }
+
```
NullSoldier, line 177: Only a computer would want this right? Why would a computer not set the config first and use this field?

danield9tqh: In the use case we want to get all the node logs from startup as JSON. When starting up the nodes datadirs aren't initialized yet and I don't think we have an option to pass in a config file on startup. We could startup then change the config but then we'd have to restart the node or miss some logs. 

I think it could also be useful to encode for terraform scripts if we want to log to docker or something and don't want to have to make sure the config is set correctly in the node's data dir

danield9tqh:
Closing in favor of setting the config value using the SDK

PR #3696, opened 2023-03-24
fixes count of subscribed pool miners

hughy:
## Summary

the StratumServer has a property, 'subscribed', that it uses to count the number of subscribed miners.

we don't decrement that counter when a subscribed miner is disconnected because of an error. this causes the miner count in status messages to hover at an incorrect level after miner errors.

we also use the number of clients to indicate the number of miners when sending notifications for submitted blocks. this is not the right number since there may be connected clients that are only following the pool for status messages and aren't mining.

## Testing Plan

manual testing with discord hook to my own discord server. modified miner to send malformed messages.
before changes:
image
"Miners" keeps increasing after the miner idsconnects due to error

after changes:
image
"Miners" never increases from 0

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/mining/pool.ts
+++ b/ironfish/src/mining/pool.ts
@@ -312,7 +312,7 @@ export class MiningPool {
           )}/s`,
         )
         this.webhooks.map((w) =>
-          w.poolSubmittedBlock(hashedHeaderHex, hashRate, this.stratum.clients.size),
+          w.poolSubmittedBlock(hashedHeaderHex, hashRate, this.stratum.subscribed),
         )
       } else {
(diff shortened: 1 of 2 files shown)


danield9tqh:
Is subscribed number going to be any different here than `this.clients.count?`

danield9tqh (approved):

hughy:
> Is subscribed number going to be any different here than `this.clients.count?`

We chatted elsewhere, but I'll answer here too for completeness.

`clients` could include non-miner clients that are listening to status updates.

PR #3697, opened 2023-03-25
uses default node account in pool by default

hughy:
## Summary

by default the name of the account that a mining pool uses for payout transactions is hardcoded to 'default'. this poses two potential problems:

1. if the default account on the pool node is not 'default', then 'default' may not have any funds for payouts
2. 'default' may not exist on the pool node

instead of hardcoding the default value of the pool account name we can leave it unset by default. at the time of a payout the pool can then use the default account of its node.

## Testing Plan

- adds unit tests
- existing unit tests cover fetching default account from pool node via rpc

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/fileStores/config.ts
+++ b/ironfish/src/fileStores/config.ts
@@ -148,7 +148,7 @@ export type ConfigOptions = {
   /**
    * The name of the account that the pool will use to payout from.
    */
-  poolAccountName: string
+  poolAccountName?: string
 
   /**
    * Should pool clients be banned for perceived bad behavior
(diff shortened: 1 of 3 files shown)


mat-if (approved):
Looks good to me. Love the tests. What do you think about also checking the account on start? It's not perfect, say if they change their accounts after it is started, but that way it's a quick feedback loop if its incorrect and they aren't changing things. Could easily be a separate PR.

hughy:
> What do you think about also checking the account on start?

That sounds like a good idea to me!

I'll add that check in a separate PR (https://linear.app/if-labs/issue/IFL-315/check-whether-poolaccountname-exists-on-start)

PR #3707, opened 2023-03-28
Autotest account imports

EvanJRichard:
## Summary
This PR proposes to add a bash script that tries to import many different accounts generated across many different versions through three different supported import codepaths. This will make account import regressions easier to catch.

At time of writing, the test inputs are only spending accounts, no viewonly accounts are included.

- [x] get `expect` (or something) working with the interactive path
- [x] handle the case where `import` requests a name for the account
- [x] examine output or return codes if applicable and signal success or failure
- [x] add the script to the tests automatically run by github

## Testing Plan
Added the test to `slow-tests`.

## Documentation

```
[ ] Yes
[X] No
```

## Breaking Change

```
[ ] Yes
[X] No
```


ironfish-cli/scripts/import-export-test-vector/0p1p65_blob.txt
```diff
@@ -0,0 +1 @@
+ironfishaccount0000010v38vetjwd5k7m3z8gcjcgnwv9kk2g36ygc8qvtsxc647cnvda3zytpzwdcx2mnyd9hxwjm90y3r5g34xa3nswpkvf3xzwrzvejnvd33vcekzen9x9nx2v3exq6xyvmyvgckzvn9xvmrqen9xqmn2cnrv4nxzef5xgenydfkve3n2vp5vgcr2g3vyfmxjethfdjhjg36ygmnzwfevv6kxwryxejrvcenxqurjce3xymnqvenvf3xzcmzxuerzetzxsurvc3sxgur2vfhx9snyvp3xguxvvtrvdjngvpkxc6rwve4xuckvwfhxu6rwctzxycnqvnxx56xydfnxp3xgefj89jk2wp4xv6kvwpkxdjrwdf4vdskvdnxx4nxvdp5v5erydmrvymrwenr89jnqc3z9s3xjmnrdakkjmn82e5k2a6tv4ujyw3zxp3nxve5v5uxzenrvgmkzvpcv93nyvf4xcmnwd3cvdjxyetr8y6kye3cxyergvf5vc6ngdmrvyunvvnyxycn2efc8qekgwtxvgenqvpz9s3x7at5vahkjmn82e5k2a6tv4ujyw3zvfsn2wt9x93nxc3hvvenywtzxq6n2vp3xp3n2enrxgux2de4x4jx2v3hxyunjd33vvuryv3nvscrwen9vgervvpev3nrwenx8qckxv3z9s38qatzd35kxstyv3ex2umnygazydpexejngdm9vserxctxvejkvd35x43kxe3jv33xxef4vg6rsdeevfjnwep3xgcxyenxxycxzc34x9snswp5vfnxgven8qcnywf589jzylg7sfvp6
```
NullSoldier, line 1: We usually call these fixtures in our code base, it would be nice to remove the suffix or call them fixtures. It'd be fine to leave the suffix off too.

package.json
```diff
@@ -29,8 +29,8 @@
     "test:changed": "lerna run --since origin/master --include-dependents test",
     "test:update": "lerna run test -- -u",
     "test:update:changed": "lerna run --since origin/master --include-dependents test -- -u",
-    "test:slow": "lerna run test:slow",
-    "test:slow:coverage": "lerna run test:slow --stream -- --collect-coverage --testPathIgnorePatterns",
+    "test:slow": "lerna run test:slow && bash ./ironfish-cli/scripts/import-export-test.sh",
+    "test:slow:coverage": "lerna run test:slow --stream -- --collect-coverage --testPathIgnorePatterns && bash ./ironfish-cli/scripts/import-export-test.sh",
```
NullSoldier, line 33: I think we should separate out these tests into a different test command. test:importexport?

ironfish-cli/scripts/import-export-test.sh
```diff
@@ -0,0 +1,150 @@
...
+    # verify return code of delete
+    if [ $? -ne 0 ]; then
+        echo "Deletion failed for $ACCOUNT_NAME"
+        exit 1
+    fi
+    check_delete_success "$DELETE_OUTPUT" "$ACCOUNT_NAME"
+}
+
+#this script is always run from the root of ironfish
+TEST_VECTOR_LOCATION='./ironfish-cli/scripts/import-export-test-vector/'
```
NullSoldier, line 133: You don't need to depend on this. You can setup bash scripts so it doesn't matter where they are run from. Generally in a bash script the first few lines should be

1. Set error mode to catch errors in the scripts
2. CD to the current directory of the script

By doing that, you never have to care what your CWD is when you execute this script. I ran into this issue when I was just using it.

ironfish-cli/scripts/import-export-test.sh
```diff
@@ -0,0 +1,150 @@
...
+FORMAT_ARRAY=( blob json mnemonic)
+for VERSION in {65..72}
```
NullSoldier, line 135: This needs to be updated every release. What if we change this to just loop over every file in the fixtures directory and run an import test on it?

ironfish-cli/scripts/import-export-test.sh
```diff
@@ -0,0 +1,150 @@
+#!/bin/bash
```
NullSoldier, line 1: If you need something installed like `expect`, our other commands make that obvious with a nice error by checking if the command is installed.

https://github.com/iron-fish/ironfish/blob/master/ironfish-cli/scripts/build.sh#L11

EvanJRichard:
Review comments addressed. It's worth calling out in case we care, this new CI step takes ~15 minutes to run (16m 38s on latest version)

ironfish-cli/scripts/import-export-test.sh
```diff
@@ -0,0 +1,152 @@
...
+            }
+            eof {
+                set output \$expect_out(buffer)
+            }
+        }
+        puts \$output
+    "
+    check_error "Import failed for $account_name"
+    check_import_success "$account_name"
+    yarn --cwd .. start wallet:delete $account_name --wait
```
NullSoldier, line 133: we don't need to use --wait here since were just using the accounts existence in the test and not any of its data

EvanJRichard:
Import/export regression test now takes ~<10 minutes.

hughy (approved):
Excellent! I just used this to test the account changes in #3719 

I think if we want to reduce the runtime of the tests we might be able to eliminate some of the test cases: some of the versions between 0.1.65 and 0.1.72 may not have included changes to the accountdata.

PR #3709, opened 2023-03-29
Add index & noteIndex in chain/getTransaction

hairtail:
## Documentation
https://github.com/iron-fish/website/pull/362

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/chain/getTransaction.ts
+++ b/ironfish/src/rpc/routes/chain/getTransaction.ts
@@ -12,14 +12,11 @@ export type GetTransactionRequest = { blockHash: string; transactionHash: string
 export type GetTransactionResponse = {
   fee: string
   expiration: number
-  initialNoteIndex: number
+  noteSize: number
   notesCount: number
   spendsCount: number
   signature: string
-  notesEncrypted: {
-    noteIndex: number
-    noteData: string
-  }[]
+  notesEncrypted: string[]
   mints: {
     assetId: string
     value: string
@@ -40,20 +37,11 @@ export const GetTransactionResponseSchema: yup.ObjectSchema<GetTransactionRespon
   .object({
     fee: yup.string().defined(),
     expiration: yup.number().defined(),
-    initialNoteIndex: yup.number().defined(),
+    noteSize: yup.number().defined(),
     notesCount: yup.number().defined(),
     spendsCount: yup.number().defined(),
     signature: yup.string().defined(),
-    notesEncrypted: yup
-      .array(
-        yup
-          .object({
-            noteIndex: yup.number().defined(),
-            noteData: yup.string().defined(),
-          })
-          .defined(),
-      )
-      .defined(),
+    notesEncrypted: yup.array(yup.string().defined()).defined(),
     mints: yup
       .array(
         yup
@@ -95,7 +83,7 @@ router.register<typeof GetTransactionRequestSchema, GetTransactionResponse>(
     const rawTransaction: GetTransactionResponse = {
       fee: '0',
       expiration: 0,
-      initialNoteIndex: 0,
+      noteSize: 0,
       notesCount: 0,
       spendsCount: 0,
       signature: '',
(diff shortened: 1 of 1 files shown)


hairtail:
I will keep the RPC api doc up to date once this PR is approved.

ironfish/src/rpc/routes/chain/getTransaction.ts
```diff
@@ -72,36 +86,44 @@ router.register<typeof GetTransactionRequestSchema, GetTransactionResponse>(
...
       spendsCount: 0,
       signature: '',
       notesEncrypted: [],
       mints: [],
       burns: [],
     }
+    const transactions = await node.chain.getBlockTransactions(blockHeader)
 
-    block.transactions.map((transaction) => {
+    transactions.map(({ transaction, initialNoteIndex }) => {
```
NullSoldier, line 108: Can you revert all the above? All the data you need is already on the block header as `noteSize`

hairtail: Reason for this change is that i want to use an indeed same logic to get transaction initialNoteIndex as `wallet.connectBlock` does. I have get some errors caused by wrong calculation of transaction initialNoteIndex.

ironfish/src/rpc/routes/chain/getTransaction.ts
```diff
@@ -72,36 +86,44 @@ router.register<typeof GetTransactionRequestSchema, GetTransactionResponse>(
...
+          notesEncrypted.push({
+            noteIndex: noteIndex,
+            noteData: note.serialize().toString('hex'),
+          })
+          noteIndex++
         }
 
         rawTransaction.fee = fee
         rawTransaction.expiration = expiration
+        rawTransaction.index = initialNoteIndex
```
NullSoldier, line 126: This is not the transaction index. A transaction index is an index in the block. That's common word used by bitcoin and other cryptocurrencies.

ironfish/src/rpc/routes/chain/getTransaction.ts
```diff
@@ -72,36 +86,44 @@ router.register<typeof GetTransactionRequestSchema, GetTransactionResponse>(
...
         const fee = transaction.fee().toString()
         const expiration = transaction.expiration()
         const signature = transaction.transactionSignature()
         const notesEncrypted = []
 
+        let noteIndex = initialNoteIndex
         for (const note of transaction.notes) {
-          notesEncrypted.push(note.serialize().toString('hex'))
+          notesEncrypted.push({
+            noteIndex: noteIndex,
```
NullSoldier, line 118: We can just store this on the transaction and not duplicate it, since it can be calculated by the consumer of this.

hairtail: I have tried this way. Store the transaction initialNoteIndex and then calculation the noteIndex with the initialNoteIndex and transaction notes. You know, the explorer backend database can do this, but i did get some bad errors. But idk whether that would happen with node.

ironfish/src/rpc/routes/chain/getTransaction.ts
hairtail: Rename as `initialNoteIndex` in the latest commit.

ironfish/src/rpc/routes/chain/getTransaction.ts
danield9tqh: What type of errors were you getting? I would be concerned if transactions come out of order or something

ironfish/src/rpc/routes/chain/getTransaction.ts
danield9tqh: What type of errors were you getting here @hairtail ?

ironfish/src/rpc/routes/chain/getTransaction.ts
hairtail: As you say, transactions come out of order, then this transaction order and block header `treeSize` used in calculation of transaction `initialNoteIndex`, `note index`. Then wrong `noteWitness` to this `note` is used in transaction creation and this transaction creation failed (transaction proof verification failed).

ironfish/src/rpc/routes/chain/getTransaction.ts
```diff
@@ -12,10 +12,14 @@ export type GetTransactionRequest = { blockHash: string; transactionHash: string
 export type GetTransactionResponse = {
   fee: string
   expiration: number
+  initialNoteIndex: number
```
NullSoldier, line 15: This is named noteSize in other PRs

ironfish/src/rpc/routes/chain/getTransaction.ts
```diff
@@ -12,10 +12,14 @@ export type GetTransactionRequest = { blockHash: string; transactionHash: string
...
   notesCount: number
   spendsCount: number
   signature: string
-  notesEncrypted: string[]
+  notesEncrypted: {
+    noteIndex: number
+    noteData: string
+  }[]
```
NullSoldier, line 22: This shouldnt be specified, normally you re-calculate this if your fetching the root noteSize

danield9tqh (approved):

PR #3714, opened 2023-03-29
Remove async init from node filesystem

danield9tqh:
## Summary
This seems like just legacy code of awaiting imports so removing it and using default node libraries.

## Testing Plan
Running unit tests and running node. Also tested out outputting logs to a file

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/fileStores/fileStore.test.ts
+++ b/ironfish/src/fileStores/fileStore.test.ts
@@ -8,7 +8,7 @@ import { FileStore } from './fileStore'
 describe('FileStore', () => {
   it('should load file store', async () => {
     const dir = getUniqueTestDataDir()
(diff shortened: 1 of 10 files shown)


ironfish/src/fileStores/fileStore.test.ts
```diff
@@ -8,7 +8,7 @@ import { FileStore } from './fileStore'
 describe('FileStore', () => {
   it('should load file store', async () => {
     const dir = getUniqueTestDataDir()
-    const files = await new NodeFileProvider().init()
+    const files = new NodeFileProvider()
```
NullSoldier, line 11: You need to do this if you ever want it to be compatible in the web.

PR #3717, opened 2023-03-30
addresses zero division errors in Target calculation

hughy:
## Summary

- fromDifficulty: division by zero when difficulty is 0
- toDifficulty: division by zero when target is 0

any difficulty value > 2**256 will give a target value of 0

note: ethereum uses the same calculation (2**256 / difficulty) for target, but some ethereum implementations (e.g., cpp-ethereum) used a 256-bit integer to store difficulty in the block header, which made a target of 0 impossible.

## Testing Plan

- adds unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/primitives/target.ts
+++ b/ironfish/src/primitives/target.ts
@@ -142,8 +142,8 @@ export class Target {
    * Converts difficulty to Target
    */
   static fromDifficulty(difficulty: bigint): Target {
-    if (difficulty <= 1n) {
-      return new Target(MAX_256_BIT_NUM)
+    if (difficulty <= Target.minDifficulty()) {
+      return Target.maxTarget()
     }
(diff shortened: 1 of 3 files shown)


ironfish/src/primitives/target.ts
```diff
@@ -152,6 +152,9 @@ export class Target {
    * Return the difficulty representation as a big integer
    */
   toDifficulty(): bigint {
+    if (this.targetValue === 0n) {
+      return 2n ** 256n + 1n
```
hughy, line 156: We could return `MAX_256_BIT_NUM` for targets of 0 or 1 to avoid overflow.

dguenther (approved):

PR #3720, opened 2023-04-03
feat: airdrop split command

jowparks:
## Summary
Adds split command so that we can have pre-split notes for airdrop. Each airdrop transaction will contain all the ORE required for `AIRDROP_NOTES_IN_BLOCK` * FEE (10 ORE) plus the payouts for that transaction.
## Testing Plan
test in simulator

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/airdrop/constants.ts
+++ b/ironfish-cli/src/commands/airdrop/constants.ts
@@ -1,0 +1,5 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+export const AIRDROP_NOTES_IN_BLOCK = 600
+export const FEE_ORE_PER_AIRDROP = 10n
--- a/ironfish-cli/src/commands/airdrop/split.ts
+++ b/ironfish-cli/src/commands/airdrop/split.ts
@@ -1,0 +1,68 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Flags } from '@oclif/core'
+import fs from 'fs/promises'
+import { IronfishCommand } from '../../command'
+import { LocalFlags } from '../../flags'
+import { parseAllocationsFile } from '../../utils/allocations'
+import { AIRDROP_NOTES_IN_BLOCK, FEE_ORE_PER_AIRDROP } from './constants'
+
+export default class AirdropSplit extends IronfishCommand {
+  static aliases = ['airdrop:split']
+  static hidden = true
+
+  static flags = {
+    ...LocalFlags,
+    account: Flags.string({
+      required: true,
+      description: 'The name of the account to use for sending airdrop',
+    }),
+    allocations: Flags.string({
+      required: true,
+      description:
+        'A CSV file with the format address,amountInOre,memo containing airdrop allocations',
+    }),
+    output: Flags.string({
+      required: false,
+      default: 'split_transaction.txt',
+      description: 'A serialized raw transaction for splitting originating note',
+    }),
+  }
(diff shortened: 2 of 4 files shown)


ironfish-cli/src/utils/allocations.ts
```diff
@@ -0,0 +1,61 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
```
jowparks, line 3: No source change, just moved for reusage.

ironfish-cli/src/commands/airdrop/split.ts
```diff
@@ -0,0 +1,68 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Flags } from '@oclif/core'
+import fs from 'fs/promises'
+import { IronfishCommand } from '../../command'
+import { LocalFlags } from '../../flags'
+import { parseAllocationsFile } from '../../utils/allocations'
+import { AIRDROP_NOTES_IN_BLOCK, FEE_ORE_PER_AIRDROP } from './constants'
+
+export default class AirdropSplit extends IronfishCommand {
```
NullSoldier, line 11: Can you move this into wallet/split?

ironfish-cli/src/utils/allocations.ts
dguenther: This is also used in the `genesisadd` command if you wanted to delete from there as well, i'm not too concerned about it though

ironfish-cli/src/commands/airdrop/constants.ts
```diff
@@ -0,0 +1,5 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+export const AIRDROP_NOTES_IN_BLOCK = 600
+export const FEE_ORE_PER_AIRDROP = 10n
```
dguenther, line 5: Might be slightly clearer to call this `FEE_ORE_PER_NOTE` but I'm not too concerned with it

ironfish-cli/src/commands/airdrop/constants.ts
```diff
@@ -0,0 +1,5 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+export const AIRDROP_NOTES_IN_BLOCK = 600
```
dguenther, line 4: `AIRDROP_NOTES_PER_TRANSACTION` might be clearer, but again, don't mind it too much

ironfish-cli/src/commands/airdrop/split.ts
```diff
@@ -0,0 +1,68 @@
...
+        publicAddress: publicKey,
+        amount: ore.toString(),
+        memo: '',
+      })
+    }
+    const transaction = await client.createTransaction({
+      account,
+      outputs,
+      // uses dust from flooring airdrop
+      fee: '10',
```
dguenther, line 64: For the airdrop transactions we reserved notes_in_transaction * 10, is it okay that this is a flat 10, or should it be outputs.length * 10?

ironfish-cli/src/commands/airdrop/constants.ts
jowparks: We can update this in later PR, I will keep a list of notes for what we want to do after getting these merged. I just don't want to go through all the rebases again.

ironfish-cli/src/commands/airdrop/split.ts
jowparks: The split is a single transaction of about 50 notes, so I just hardcoded to 10.

PR #3721, opened 2023-04-03
feat: airdrop allocations to raw transactions

jowparks:
## Summary
Takes input allocations generated [from export](https://coda.io/d/_dMnayiE39lL/Airdrop_su_t8#_luHf9), bundles them based on max output notes per transaction, and writes them to file.
## Testing Plan
test in simulator
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/airdrop/raw-transactions.ts
+++ b/ironfish-cli/src/commands/airdrop/raw-transactions.ts
@@ -1,0 +1,81 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { isValidPublicAddress } from '@ironfish/rust-nodejs'
+import { Flags } from '@oclif/core'
+import fs from 'fs/promises'
+import { IronfishCommand } from '../../command'
+import { LocalFlags } from '../../flags'
+import { parseAllocationsFile } from '../../utils/allocations'
+import { AIRDROP_NOTES_IN_BLOCK, FEE_ORE_PER_AIRDROP } from './constants'
+
+export default class AirdropRawTransactions extends IronfishCommand {
+  static aliases = ['airdrop:raw']
+  static hidden = true
+
+  static flags = {
+    ...LocalFlags,
+    account: Flags.string({
+      required: true,
+      description: 'The name of the account used to create raw transactions',
+    }),
+    allocations: Flags.string({
+      required: true,
+      description:
+        'A CSV file with the format address,amountInIron,memo containing genesis block allocations',
+    }),
+    raw: Flags.string({
+      required: false,
+      default: 'raw_transactions.txt',
+      description: 'where to output the raw transactions',
+    }),
+  }
+  async start(): Promise<void> {
+    const { flags } = await this.parse(AirdropRawTransactions)
+    const account = flags.account
+    const client = await this.sdk.connectRpc()
+
+    const csv = await fs.readFile(flags.allocations, 'utf-8')
+    const result = parseAllocationsFile(csv)
+
+    if (!result.ok) {
+      this.error(result.error)
+    }
+    const allocations = result.allocations
+
+    await fs.rm(flags.raw)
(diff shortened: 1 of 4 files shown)


ironfish-cli/src/commands/airdrop/raw-transactions.ts
```diff
@@ -0,0 +1,74 @@
...
+          amount: output.amountInOre.toString(),
+          memo: output.memo,
+        })
+      }
+
+      const result = await client.createTransaction({
+        account,
+        outputs,
+        fee: String(BigInt(AIRDROP_NOTES_IN_BLOCK) * FEE_ORE_PER_AIRDROP),
+      })
```
dguenther, line 69: I don't think `createTransaction` marks any notes as spent, so will this end up generating multiple transactions that spend the same note, if the note is large enough?

jowparks: Oh interesting, how should we mark the note spent? Or should we use a different method call here?

hughy: The wallet will only mark notes as spent when a posted transaction is added. 

That won't work for us here because we can't alternate create/post 🤔

dguenther: Adding the ability to spend specific notes would be nice, but might be outside the scope of this PR.

Another option could be to boot up a node in-process (rather than connecting to one over RPC), then you can reach into the wallet directly and call a function on it. You could just call this and I think it would be good enough to prevent the note from appearing in getUnspentNotes:

```ts
await this.walletDb.deleteUnspentNoteHash(this, spentNoteHash, spentNote, tx)
```

Or make a custom `createTransaction` that takes a note to spend. The tricky bit will be generating the correct witness for the note.

jowparks: https://iron-fish.slack.com/archives/C02PF56HESK/p1681337839023619

https://github.com/iron-fish/ironfish/pull/3779

jowparks: Note that I decided to make this a separate RPC because I want to run this code in the simulator. I don't think the simulator is set up well to handle new node spawns in CLI layer, so I just forked cli command and added these lines as the only changes (thanks @hughy for help converting note types ):
```
      for (const spend of transaction.spends) {
        const decryptedNote = await account.getDecryptedNote(spend.note.hash())
        Assert.isNotUndefined(decryptedNote, 'Decrypted note that is being deleted is null.')
        await node.wallet.walletDb.deleteUnspentNoteHash(
          account,
          spend.note.hash(),
          decryptedNote,
        )
      }
```

ironfish/src/rpc/routes/wallet/createTransactionAirdrop.ts
```diff
@@ -0,0 +1,197 @@
...
+      params.feeRate = CurrencyUtils.decode(request.data.feeRate)
+    } else {
+      params.feeRate = node.memPool.feeEstimator.estimateFeeRate('average')
+    }
+
+    try {
+      const transaction = await node.wallet.createTransaction(params)
+      for (const spend of transaction.spends) {
+        const decryptedNote = await account.getDecryptedNote(spend.note.hash())
+        Assert.isNotUndefined(decryptedNote, 'Decrypted note that is being deleted is null.')
```
hughy, line 178: very minor nit - we aren't deleting the note here, just marking it as spent after we used it in the transaction

ironfish/src/rpc/routes/wallet/createTransactionAirdrop.ts
```diff
@@ -0,0 +1,197 @@
...
+        await node.wallet.walletDb.deleteUnspentNoteHash(
+          account,
+          spend.note.hash(),
+          decryptedNote,
+        )
```
hughy, line 183: This will get the result that we want as far as not re-using notes.

We could also change the `spent` field in the saved note to `true`. It won't make a difference for re-using notes, but I'm wondering if it might be useful for debugging or looking at the status of notes 🤔

hughy (approved):
A couple of minor comments, but looks good to me! 👍

PR #3723, opened 2023-04-03
feat: accept transactions response update

jowparks:
## Summary
Currently we have no info returned from RPC about the transaction that was added, this adds both if the transaction was accepted and what the hash is.
## Testing Plan
tests pass
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/addTransaction.ts
+++ b/ironfish/src/rpc/routes/wallet/addTransaction.ts
@@ -14,6 +14,8 @@ export type AddTransactionRequest = {
 
 export type AddTransactionResponse = {
   accounts: string[]
+  hash: string
+  accepted: boolean
 }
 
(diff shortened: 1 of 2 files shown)


ironfish/src/rpc/routes/wallet/addTransaction.ts
```diff
@@ -14,6 +14,8 @@ export type AddTransactionRequest = {
 
 export type AddTransactionResponse = {
   accounts: string[]
+  hash: string
```
hughy, line 17: We have a ticket for adding the hash to the response of `wallet/postTransaction`, too. If that endpoint returns the hash then the client will already have the hash before calling `wallet/addTransaction`.

I think it's fine to have the hash here too, though!

hughy (approved):

PR #3726, opened 2023-04-03
refactor(ironfish): do not quit transaction:watch if the transaction is not found

holahula:
## Summary
`ironfish wallet:transactions:watch` no longer quits if the transaction is not found. The `NOT_FOUND` state has been added as a `TransactionStatus` and is now integrated into the watch status loop 

Closes IFL-413

## Testing Plan
Successful watch:
```
➜  ironfish git:(holahula/fix/transaction-watch-404) ✗ fish wallet:send -d ~/.ironfish-simulator/node-e050 --watch
yarn run v1.22.19
$ yarn build && yarn start:js wallet:send -d /Users/austino/.ironfish-simulator/node-e050 --watch
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc --inspect=:0 --inspect-publish-uid=http --enable-source-maps bin/run wallet:send -d /Users/austino/.ironfish-simulator/node-e050 --watch
Enter the amount (balance 48405.00000000): 1
Enter the public address of the recipient: beb858a08898867fa7917c45f4ff6c5c67cb590dc654204330b0b69cc647ef1f
? Select the fee you wish to use for this transaction Enter a custom fee
Enter the fee amount in $IRON (balance 48405.00000000): 1
You are about to send a transaction: $IRON 1.00000000 plus a transaction fee of $IRON 1.00000000 to beb858a08898867fa7917c45f4ff6c5c67cb590dc654204330b0b69cc647ef1f from the account default
Do you confirm (Y/N)?: Y
Sending the transaction... done
Sent $IRON 1.00000000 to beb858a08898867fa7917c45f4ff6c5c67cb590dc654204330b0b69cc647ef1f from default
Hash: 7e9ae0a153e0b6bc8375c67ee5be8f62f6a715144461656ab76b676b098a2fc4
Fee: $IRON 1.00000000

If the transaction is mined, it will appear here https://explorer.ironfish.network/transaction/7e9ae0a153e0b6bc8375c67ee5be8f62f6a715144461656ab76b676b098a2fc4

{"level":2,"tag":"wallet:send","date":"2023-04-05T17:07:23.987Z","message":"Watching transaction 7e9ae0a153e0b6bc8375c67ee5be8f62f6a715144461656ab76b676b098a2fc4"}
Current Status... pending -> unconfirmed: 20s
Current Status... unconfirmed -> confirmed: 10s
Current Status... done after 30s
✨  Done in 53.00s.
```

If txn not found, then `watchTransaction()` loop forever
```
➜  ironfish git:(holahula/fix/transaction-watch-404) ✗ fish wallet:transaction:watch abc -d ~/.ironfish-simulator/node-75d5
yarn run v1.22.19
$ yarn build && yarn start:js wallet:transaction:watch abc -d /Users/austino/.ironfish-simulator/node-75d5
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc --inspect=:0 --inspect-publish-uid=http --enable-source-maps bin/run wallet:transaction:watch abc -d /Users/austino/.ironfish-simulator/node-75d5
{"level":2,"tag":"wallet:transaction:watch","date":"2023-04-05T17:11:56.931Z","message":"Watching transaction abc"}
Current Status... ⢿ not found 20s
```

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/utils/transaction.ts
+++ b/ironfish-cli/src/utils/transaction.ts
@@ -3,7 +3,6 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 import {
-  Assert,
   createRootLogger,
   Logger,
   PromiseUtils,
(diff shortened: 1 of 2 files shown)


hughy (approved):

ironfish-cli/src/commands/wallet/transaction/watch.ts
```diff
@@ -39,12 +40,22 @@ export class WatchTxCommand extends IronfishCommand {
...
     const client = await this.sdk.connectRpc()
 
-    await watchTransaction({
-      client,
-      logger: this.logger,
-      account,
-      hash,
-      confirmations: flags.confirmations,
-    })
+    const TIMEOUT = 5000
```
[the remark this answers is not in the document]
holahula: changed to regular lowercase local var

ironfish-cli/src/commands/wallet/transaction/watch.ts
```diff
@@ -39,12 +40,22 @@ export class WatchTxCommand extends IronfishCommand {
...
-      logger: this.logger,
-      account,
-      hash,
-      confirmations: flags.confirmations,
-    })
+    const timeout = 5000
+    let watched = false
+
+    while (!watched) {
+      watched = await watchTransaction({
```
dguenther, line 47: `watchTransaction` is a CLI utility function -- if we're just looping until the transaction is found, could we do the loop inside of `watchTransaction`?

holahula: moved the loop to `watchTransaction`

dguenther (approved):
This looks good. I think the one other UX change I'd consider making is integrating the "not found" status into the other loop, so the CLI output of both loops matched up but displays e.g. "not found" instead of "pending". This is a net improvement though so I'm good with merging it!

holahula:
Made the UX changes, now the spinner shows when the transaction is in the `not found` state, and logs a state change when it transitions from `not found` -> `pending` / `unconfirmed` / etc.

ironfish-cli/src/utils/transaction.ts
```diff
@@ -36,11 +36,43 @@ export async function watchTransaction(options: {
 
   const startTime = lastTime
 
-  if (last.content.transaction == null) {
-    logger.log(`Tried to watch transaction ${options.hash} but it's missing.`)
-    return
+  let state = ''
```
NullSoldier, line 39: There are now 2 states, when there used to be 1 state. Do we need 2 states?

holahula: I'll refactor the old code to use the new state system. I needed to introduce new state to know if the while loop was entered

ironfish-cli/src/utils/transaction.ts
```diff
@@ -36,11 +36,43 @@ export async function watchTransaction(options: {
...
+    logger.log(`Retrying in ${pollFrequencyMs / 1000} s...`)
+
+    await PromiseUtils.sleep(pollFrequencyMs)
+
+    last = await options.client.getAccountTransaction({
+      account: options.account,
+      hash: options.hash,
+      confirmations: options.confirmations,
+    })
+  }
```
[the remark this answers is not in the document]
holahula: The logic below has an invariant that `last.content.transaction != null`. The code above could be moved within the `while (true)` loop but it would probably have to change the entire structure of the loop. 

Also, most of the code changes is UX improvements, without it the core change is ~5 lines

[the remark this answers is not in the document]
dguenther: We should be able to use a `prevStatus` variable that's declared outside of the loop to add this logic to the existing while loop. The loop is really just handling status transitions for the transaction, and we want to add support for a new status of "not found". 

I think the while loop needs some refactoring, but I'd expect something like this:

```
prevStatus = null
while true {
  resp = await getTransaction()
  const currentStatus = resp?.transaction ?? 'not found'

  if prevStatus != null && prevStatus != 'not found' && currentStatus == 'not found' {
    log "transaction was deleted"
    return
  }

  if currentStatus == waitUntil {
    log "reached status waitUntil"
    return
  }

  if prevStatus != null && currentStatus != prevStatus {
    log "prevStatus -> currentStatus"
  }
  
  log currentStatus xx seconds
  prevStatus = currentStatus
  sleep
}
```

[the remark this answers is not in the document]
holahula: fixed the status updating

ironfish-cli/src/utils/transaction.ts
```diff
@@ -36,64 +35,68 @@ export async function watchTransaction(options: {
 
   const startTime = lastTime
 
-  if (last.content.transaction == null) {
-    logger.log(`Tried to watch transaction ${options.hash} but it's missing.`)
-    return
-  }
+  let prevStatus = last?.content.transaction?.status ?? TransactionStatus.NOT_FOUND
```
dguenther, line 38: I don't think we should add NOT_FOUND to TransactionStatus, since TransactionStatus is used for the status field of an existing transaction. In this case, we're just using it for display purposes.

Another way I think about it is that the `ironfish` package internally doesn't use the NOT_FOUND status, so I don't think we should change the enum just for the convenience of writing external (`ironfish-cli`) code.

I think you could type this as `type TransactionWatchStatus = TransactionStatus | 'not found'` and everything should just work.

holahula: removed changes to `TransactionStatus`

PR #3722, opened 2023-04-03
feat: airdrop post transactions

jowparks:
## Summary
This will be run on air-gapped node to post the transactions with the spend key for airdrops.
## Testing Plan
simulation test
## Documentation

Does this change require any updates to the Iron Fish Docs (ex. [the RPC API
Reference](https://ironfish.network/docs/onboarding/rpc/chain))? If yes, link a
related documentation pull request for the website.

```
[ ] Yes
```

## Breaking Change

Is this a breaking change? If yes, add notes below on why this is breaking and
what additional work is required, if any.

```
[ ] Yes
```

PR #3727, opened 2023-04-03
feat: airdrop airdrop

jowparks:
## Summary
This command takes the raw->posted transactions file from: 
https://github.com/iron-fish/ironfish/pull/3722
And adds them to the network.
## Testing Plan
simulator test
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/airdrop/airdrop.ts
```diff
@@ -0,0 +1,46 @@
...
+          hash: response.content.hash,
+        })
+      } else {
+        this.logger.warn(
+          `Skipping ${response.content.hash} transaction (#${lineNum}), transaction was not accepted`,
+        )
+      }
+    }
+  }
+}
```
NullSoldier, line 46: can we move this into the service commands?

jowparks: I don't think it matters where we put this stuff, we are never going to merge it into staging/master (unless you think we should).

I have a root branch `airdrop` that I was planning on using for the actual airdrop.

ironfish-cli/src/commands/airdrop/airdrop.ts
```diff
@@ -0,0 +1,48 @@
...
+    const lines = fileContent.split(/[\r\n]+/)
+    const client = await this.sdk.connectRpc()
+    for (const line of lines) {
+      lineNum++
+      CliUx.ux.action.start(`Adding transaction #${lineNum}`)
+      const response = await client.addTransaction({ transaction: line.trim() })
+      CliUx.ux.action.stop()
+      if (response.content.accepted) {
+        this.logger.info(`Added ${response.content.hash} transaction (#${lineNum})`)
+        await watchTransaction({
```
dguenther, line 35: Is it important to wait all the way until a transaction is confirmed before sending the next? They should be independent, but I could see not wanting to flood the network + mempool with a bunch of airdrop transactions.

If you do want to release them one at a time, it could be a bit quicker to wait for UNCONFIRMED. That way they'll be sent out in back-to-back blocks rather than separated by CONFIRMATION_RANGE # of blocks.

Also, do you have a plan to watch for transactions that expire without getting added to a block?

jowparks: Yea I would be fine sending them all at once, for some reason I thought Jason wanted to do individual in a row. As for the expiration I actually had a note about this, I was wondering if we should just make these have really long expiration times?

jowparks: Current state is that I have confirmations set to 0 and we are waiting, open to remove the wait if that is desirable.

jowparks:
> Looks good to me
> 
> Do you think it would be useful to collect a file with transaction hashes? Or to write to a file any transactions that were skipped?

Maybe? Let me add to the list of things we need to complete before shipping kyc

PR #3731, opened 2023-04-03
adds decrypted notes to getAccountTransactions

hughy:
## Summary

optionally decrypts transaction output notes and includes them in the response from getAccountTransactions with each transaction.

adding the decrypted notes will support adding both sender address and recipient address to the output of `wallet:transactions`: each note includes both the sender address and the 'owner' address, which is the recipient address.

## Testing Plan

- adds unit test

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/getTransactions.ts
+++ b/ironfish/src/rpc/routes/wallet/getTransactions.ts
@@ -9,7 +9,12 @@ import { Account } from '../../../wallet/account'
 import { TransactionValue } from '../../../wallet/walletdb/transactionValue'
 import { RpcRequest } from '../../request'
 import { ApiNamespace, router } from '../router'
-import { getAssetBalanceDeltas, serializeRpcAccountTransaction } from './types'
+import {
+  getAccountDecryptedNotes,
+  getAssetBalanceDeltas,
+  RpcAccountDecryptedNote,
+  serializeRpcAccountTransaction,
+} from './types'
 import { getAccount } from './utils'
 
 export type GetAccountTransactionsRequest = {
@@ -19,6 +24,7 @@ export type GetAccountTransactionsRequest = {
   limit?: number
   offset?: number
   confirmations?: number
+  notes?: boolean
 }
 
 export type GetAccountTransactionsResponse = {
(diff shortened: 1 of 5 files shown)


ironfish/src/rpc/routes/wallet/types.ts
```diff
@@ -72,3 +75,41 @@ export async function getAssetBalanceDeltas(
 
   return assetBalanceDeltas
 }
+
```
jowparks, line 78: This probably belongs in a different file, it isn't a type.

ironfish/src/rpc/routes/wallet/types.ts
```diff
@@ -72,3 +75,41 @@ export async function getAssetBalanceDeltas(
...
+    const asset = await node.chain.getAssetById(note.assetId())
+
+    serializedNotes.push({
+      isOwner,
+      owner: note.owner(),
+      memo: note.memo(),
+      value: CurrencyUtils.encode(note.value()),
+      assetId: note.assetId().toString('hex'),
+      assetName: asset?.name.toString('hex') || '',
```
jowparks, line 108: In what cases won't the asset be defined? Wouldn't it be a corrupt database if the asset isn't found but the note is decrypted?

ironfish/src/rpc/routes/wallet/getTransactions.ts
```diff
@@ -33,6 +39,7 @@ export type GetAccountTransactionsResponse = {
   expiration: number
   timestamp: number
   assetBalanceDeltas: Array<{ assetId: string; assetName: string; delta: string }>
+  notes?: RpcAccountDecryptedNote[]
```
jowparks, line 42: I am partial to `notes: RpcAccountDecryptedNote[] | undefined` since it is more explicit, not sure if this is the pattern we use though

jowparks (approved):
LGTM, few small comments/questions

ironfish/src/rpc/routes/wallet/types.ts
hughy: this logic came before assets were stored in the wallet db and looks the asset up from the chain database. if the note is from a pending mint transaction then the asset might not have been added to the chain yet.

we may be able to update this to use the wallet database instead of the chain now that it's available

ironfish/src/rpc/routes/wallet/types.ts
hughy: yeah, i'll add a separate `utils` module and then move the other non-type declarations from this file as well

ironfish/src/rpc/routes/wallet/getTransactions.ts
hughy: I suspect that we're not very consistent about using one vs the other. I'll add a task to the RPC uniformity project to address this and make it consistent!

PR #3732, opened 2023-04-03
feat: airdrop export sql updates api

jowparks:
## Summary
We want to update the testnet website with transaction hashes for airdrops once they are complete. To do this we will look at notes of the airdrop account and run updates for the memos. In airdrop case, the memo's should be the `graffiti`, if it is an unrelated note, the update will just not update any rows.
## Testing Plan
simulator test, test local db copy.
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/airdrop/export.ts
```diff
@@ -0,0 +1,53 @@
...
+    const response = client.getAccountTransactionsStream({
+      account,
+    })
+    await fs.rm(flags.exported)
+    const fileHandle = await fs.open(flags.exported, 'a')
+    for await (const transaction of response.contentStream()) {
+      const fullTransaction = await client.getAccountTransaction({
+        account,
+        hash: transaction.hash,
+      })
```
dguenther, line 39: I think you can pass the `notes` flag to `getAccountTransactionsStream` and avoid the need to call `getAccountTransaction` on every transaction.

ironfish-cli/src/commands/airdrop/export.ts
```diff
@@ -0,0 +1,53 @@
...
+      const notes = fullTransaction.content.transaction?.notes
+      if (!notes) {
+        continue
+      }
+      for (const note of notes) {
+        await fs.appendFile(
+          fileHandle,
+          `update redemptions set transaction_hash='${transaction.hash}', sent_ore=${note.value}  where user_id = (SELECT id from users where graffiti = '${note.memo}');\n`,
```
dguenther, line 47: Any reason to consider potential SQL injections on the graffiti here?

dguenther (approved):

ironfish-cli/src/commands/airdrop/export.ts
jowparks: No it should be fine since we are running this command ourself and used typed values for note and transaction, but it is a good callout.

PR #3738, opened 2023-04-04
Change importexport tests to use test datadir

NullSoldier:
## Summary

I noticed this was failing to run because my datadir database was a later version. I see it's using my actual datadir which I have my real data in. Changing this so it does not use a users real database to perform tests, and instead uses a test datadir.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/scripts/import-export-test.sh
+++ b/ironfish-cli/scripts/import-export-test.sh
@@ -5,6 +5,8 @@ set -e # exit immediately if anything returns with non-zero exit code
 # Change working directory to the script's directory
 cd "$(dirname "$0")"
 
+DATA_DIR="../testdbs/importexport"
+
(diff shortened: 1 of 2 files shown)


ironfish-cli/scripts/import-export-test.sh
```diff
@@ -5,6 +5,8 @@ set -e # exit immediately if anything returns with non-zero exit code
 # Change working directory to the script's directory
 cd "$(dirname "$0")"
 
+DATA_DIR="../testdbs/importexport"
```
hughy, line 8: nit - I think if you use `IRONFISH_DATA_DIR` here you won't need to set it in each command

hughy (approved):

PR #3740, opened 2023-04-04
bumps account schema version for block birthday changes

hughy:
## Summary

changing the createdAt field to store a HeadValue instead of a Date qualifies as a change to the account schema.

- bumps ACCOUNT_SCHEMA_VERSION to 2
- uses account version to set createdAt to null for account imports on version 1

## Testing Plan

- adds import test cases for 0.1.73

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -134,7 +134,7 @@ export class ImportCommand extends IronfishCommand {
         }
       }
 
-      if (typeof data.createdAt === 'string') {
+      if (data.version === 1) {
         data.createdAt = null
```
NullSoldier, line 138: This is optional but it will allow any migrations later to work fully. We've fully upgraded it here so we can bump the version.

```suggestion
        data.createdAt = null
        data.version = 2
```

ironfish-cli/scripts/import-export-test/0p1p73_mnemonic.txt
```diff
@@ -0,0 +1 @@
+bicycle game eight minimum rival shoulder airport birth section flash walnut seat omit front carpet cheap wall arch brick wink kangaroo safe gas hub
```
NullSoldier, line 1: This is ok, but since its just a spend key we don't need to include it since the spend key format didn't change only the ones with metadata.

NullSoldier (approved):

NullSoldier:
I wrote some comments but approving since the code functions.

ironfish-cli/scripts/import-export-test/0p1p73_mnemonic.txt
hughy: I'll remove this one. I think we could probably remove the mnemonic test cases from some of the earlier versions to reduce the test duration

PR #3747, opened 2023-04-05
feat(IFL-446):airdrop simulator

jowparks:
## Summary
WIP
## Testing Plan
Checked output with script below:
```typescript
import * as fs from 'fs';

// Define types for the data in the first CSV file
type FirstCsvData = {
  public_address: string;
  iron: number;
  graffiti: string;
};

// Define types for the data in the second CSV file
type SecondCsvData = {
  transaction_hash: string;
  sent_ore: number;
  graffiti: string;
};

// Read the first CSV file
const firstCsvData: FirstCsvData[] = fs
  .readFileSync('allocations.txt', 'utf-8')
  .split('\n')
  .map(line => line.trim())
  .filter(line => line)
  .map(line => {
    const [public_address, iron, graffiti] = line.split(',');
    return { public_address, iron: Number(iron), graffiti };
  });

// Read the second CSV file
const secondCsvData: SecondCsvData[] = fs
  .readFileSync('exported_airdrop.txt', 'utf-8')
  .split('\n')
  .map(line => line.trim())
  .filter(line => line)
  .map(line => {
    const [, transaction_hash, ore] = line.match(/set transaction_hash='(.+)', sent_ore=(.+) where user_id =/)!;
    const [, graffiti] = line.match(/SELECT id from users where graffiti = '(.*)'/)!;
    return { transaction_hash, sent_ore: parseInt(ore), graffiti };
  });

// Map the first CSV data into a map where the key is the graffiti
const firstCsvMap = new Map<string, FirstCsvData>();
firstCsvData.forEach(data => firstCsvMap.set(data.graffiti, data));
// Loop through the second CSV data and compare the sent_ore values
secondCsvData.forEach(data => {
  const firstCsvData = firstCsvMap.get(data.graffiti);
//   console.log(data, firstCsvData?.sent_ore)

  // sending iron to a public address that should not have an allocation
  if(!firstCsvData && data.graffiti) {
    console.log('ERROR', data.graffiti, data.sent_ore)
  }
  // ore sent does not match
  if (firstCsvData && firstCsvData.iron*100000000 !== data.sent_ore) {
    console.log('ERROR', data.graffiti, data.sent_ore, firstCsvData.iron)
  }
});
console.log('DONE')
```
Output:
```
ts-node compare_airdrop.ts                                                                                                                                                                                              
ERROR Genesis Block 4200000000000000
DONE
```
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


simulator/src/simulator/simulation-node.ts
```diff
@@ -33,7 +33,7 @@ import {
 import { sleep } from './misc'
 import { getLatestBlockHash } from './utils/chain'
```
jowparks, line 35: We should parameterize this or make it somehow global definition of this root repo (so build is used and not installed binary)

ironfish/src/wallet/wallet.ts
```diff
@@ -1062,9 +1062,12 @@ export class Wallet {
     if (!head) {
       return { amount, notes }
     }
```
jowparks, line 1064: Needed to keep the notes in correct order for payment.

ironfish-cli/src/commands/airdrop/raw-transactions.ts
```diff
@@ -55,19 +55,30 @@ export default class AirdropRawTransactions extends IronfishCommand {
           )
           continue
         }
```
jowparks, line 57: Bug due to simulated ORE values being 0. This is a simple workaround.

PR #3753, opened 2023-04-06
Allow opening the repl without opening the DB

NullSoldier:
## Summary

You can now use --no-db to not load the databases. This gives you access to the SDK without needing to load the data directory.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


NullSoldier:
I think it's fine, because those examples are still valid.

1. The examples show me how to do things
2. Only advanced user will use --no-opendb
3. They'll know to call node.open() if they want to use the DB
4. They'll get an nice error saying the DB is not open

PR #3757, opened 2023-04-06
feat(simulator): Support accessing the genesis block in the simulator

holahula:
## Summary
Add support for accessing the genesis block in the Simulator. 
- Adds util functions for setting default account and importing accounts
- Add config to import the genesis account onto the node at spawn
- The dev genesis block account and name are exported in `@ironfish/sdk` 

The account is taken from #3644 

## Testing Plan
Tests pass

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


jowparks (approved):

simulator/src/simulator/simulation-node.ts
```diff
@@ -31,21 +32,35 @@ import {
...
 } from './events'
 import { sleep } from './misc'
-import { getLatestBlockHash } from './utils/chain'
+import { getLatestBlockHash, importAccount } from './utils'
 
 export const rootCmd = 'ironfish'
 
+/**
+ * Additional configuration options for the node. These are not part of the `ConfigOptions` interface
+ */
```
jowparks, line 41: Why have a separate type for these? why not just make them optional and set sane defaults when constructing?

ironfish/src/defaultNetworkDefinitions.ts
```diff
@@ -6,6 +6,18 @@ export function isDefaultNetworkId(networkId: number): boolean {
   return networkId <= 100
 }
 
+/**
+ * The name of the account that contains the funds in the genesis block.
+ */
+export const DEV_GENESIS_IRONFISH_ACCOUNT_NAME = 'IronFishGenesisAccount'
```
jowparks, line 12: 🥳

simulator/src/simulator/utils/accounts.ts
```diff
@@ -57,3 +57,33 @@ export async function getDefaultAccount(node: SimulationNode): Promise<string> {
...
+ * @param node The node to import the account on
+ * @param account The account to import, in the form of a string blob
+ * @param rescan Whether to explicitly rescan the blockchain for transactions involving the account
+ */
+export async function importAccount(
+  node: SimulationNode,
+  account: string,
+  rescan?: boolean,
+): Promise<void> {
+  await node.executeCliCommandAsync('wallet:import', [account])
```
danield9tqh, line 85: Can we just use the RPC command for this? If we store the account as spendKey, publicKey etc then we can just use the RPC command to import

holahula: The genesis account is stored as a string blob right now, is there an easy way to go from string -> {spendkey, publickey} without re-writing a lot of the logic in `wallet:import`?

simulator/src/simulator/simulation-node.ts
```diff
@@ -41,26 +42,42 @@ export const rootCmd = 'ironfish'
...
+    | 'rpcTcpHost'
+    | 'rpcTcpPort'
+    | 'bootstrapNodes'
+  >
 >
 
+/**
+ * Additional configuration options for the node. These are not part of the `ConfigOptions` interface
+ */
+export type OptionalSimluationNodeConfig = {
```
dguenther, line 68:
```suggestion
export type OptionalSimulationNodeConfig = {
```

simulator/src/simulator/utils/accounts.ts
dguenther: Since this is supposed to basically take any account that comes out of wallet:export, I'm okay leaving this as-is. I think there's a general discussion to be had around moving our account parsing logic into the SDK though

simulator/src/simulator/utils/accounts.ts
```diff
@@ -57,3 +57,33 @@ export async function getDefaultAccount(node: SimulationNode): Promise<string> {
...
+/**
+ * Sets the default account on a node.
+ *
+ * @param node The node to set the default account on
+ * @param account The account to set as the default account
+ * @throws Error if the account does not exist
+ */
+export async function setDefaultAccount(node: SimulationNode, account: string): Promise<void> {
+  await node.client.useAccount({ account })
+}
```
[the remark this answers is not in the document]
dguenther: Removed this for now

PR #3762, opened 2023-04-06
makes blockHash optional in chain/getTransaction requests

hughy:
## Summary

after adding an index from transaction hash to block hash in #3761 we no longer need to include a block hash in requests to chain/getTransaction. instead we can use the index to look up the hash of the block on the main chain that includes that transaction, if any.

- defines 'getHashByTransactionHash' to look up a block hash in 'transactionHashToBlockHash'
- changes 'blockHash' to an optional request parameter

## Testing Plan

- adds route tests for chain/getTransaction

## Documentation

```
[X] Yes
```
https://github.com/iron-fish/website/pull/359

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -1049,7 +1049,7 @@ export class Blockchain {
     return hash || null
   }
 
-  async getHashByTransactionHash(
+  async getBlockHashByTransactionHash(
     transactionHash: TransactionHash,
     tx?: IDatabaseTransaction,
   ): Promise<BlockHash | null> {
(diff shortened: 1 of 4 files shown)


ironfish/src/blockchain/blockchain.ts
```diff
@@ -1049,6 +1049,14 @@ export class Blockchain {
     return hash || null
   }
 
+  async getHashByTransactionHash(
```
EvanJRichard, line 1052: is `getBlockHashByTransactionHash` too wordy, or unnecessary given where the function lives and its return type? I ask because when I initially read this, until I saw the return type, I was confused, "wait, if we have the hash why do we need a hash, is this a no-op?"

hughy: I think that `getBlockHashByTransactionHash` would be clearer

I left out the `Block` because there are a lot of other methods like `getHashAtSequence` that have names that leave the 'block' part implicit. Since there are two different hashes at play here I think it would be better to specify

EvanJRichard (approved):

rohanjadvani (approved):

PR #3764, opened 2023-04-07
adds hash to response of wallet/postTransaction

hughy:
## Summary

sendTransaction includes the hash in its responses, but postTransaction does not.

a client _could_ hash the output themselves, but this adds an extra step and perhaps an extra dependency for the client.

## Testing Plan

- updates unit tests

## Documentation

```
[X] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/postTransaction.test.ts
+++ b/ironfish/src/rpc/routes/wallet/postTransaction.test.ts
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+import { Transaction } from '../../../primitives'
 import { RawTransactionSerde } from '../../../primitives/rawTransaction'
 import { useAccountFixture } from '../../../testUtilities'
 import { createRawTransaction } from '../../../testUtilities/helpers/transaction'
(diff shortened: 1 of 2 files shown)


ironfish/src/rpc/routes/wallet/postTransaction.test.ts
```diff
@@ -27,6 +27,7 @@ describe('Route wallet/postTransaction', () => {
 
     expect(addSpy).toHaveBeenCalledTimes(0)
     expect(response.status).toBe(200)
+    expect(response.content.hash).toBeDefined()
```
EvanJRichard, line 30: Is it worth it to calculate what the `hash` should be and check equality, do you think?

EvanJRichard (approved):
Simple, convenient, and unifies RPCs a little more. One question about tests, but non-blockin g. :shipit:

ironfish/src/rpc/routes/wallet/postTransaction.test.ts
hughy: It's a little tricky to do this: there's randomness involved in posting transactions, so posting the same raw transaction twice will result in two different hashes 😓

EvanJRichard: Ah, true. Forget it, then, imo.

danield9tqh (approved):

NullSoldier (approved):

ironfish/src/rpc/routes/wallet/postTransaction.test.ts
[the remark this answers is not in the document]
rohanjadvani: I don't think this works in Jest with `toBe`. It seems like CI is failing with this.

PR #3773, opened 2023-04-11
fix(simulator): gracefully shutdown simulator on unexpected exit

holahula:
## Summary
Attempting to hook into `process.on('uncaughtException')` in the simulator to catch exceptions before process shutdwn does not work because oclif has built-in uncaught exception handlin so the exception never propagates to this level. Thus, when unexpected errors occur in user-written simulations, the simulator is unaware of them and is unable to gracefully shutdown. This can leave running `ironfish` processes hanging in the background which affects future simulation runs. 

By wrapping the simulation execution in a try-catch block, the uncaught exceptions / unhandled rejections can be caught before they reach oclif and the simulator can gracefully shutdown before exit.

The simulator needs to be created in the `start` cmd so it can catch unexpected simulation exits. 

Closes IFL-566

## Testing Plan
Sample simulation that returns a rejected promise
```
export async function run(simulator: Simulator, logger: Logger): Promise<void> {
  const nodes = []
  for (let i = 0; i < 2; i++) {
    nodes.push(await simulator.startNode())
  }

  await rejectPromise()

  // wait for all nodes to be stopped
  await simulator.waitForShutdown()

  logger.log('nodes stopped, shutting down...')
}
```
gracefully shuts down when unexpected promise rejection reached:
```
➜  simulator git:(holahula/feat/simulator-exception-handling) sim start test
yarn run v1.22.19
$ yarn build && yarn start:js start test
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc --inspect=:0 --inspect-publish-uid=http --enable-source-maps bin/run start test
(node:4092) Error Plugin: simulator: command process not found
module: @oclif/core@1.23.1
task: toCached
plugin: simulator
root: /Users/austino/i/ironfish/simulator
See more details with DEBUG=*
Starting node {
  cfg: `{"nodeName":"node-6c93","blockGraffiti":"node-6c93","peerPort":7001,"networkId":2,"rpcTcpHost":"localhost","rpcTcpPort":9001,"bootstrapNodes":["''"],"dataDir":"~/.ironfish-simulator/node-6c93","verbose":false,"importGenesisAccount":false}`
}
started node: node-6c93
Starting node {
  cfg: '{"nodeName":"node-4f98","blockGraffiti":"node-4f98","peerPort":7002,"networkId":2,"rpcTcpHost":"localhost","rpcTcpPort":9002,"bootstrapNodes":["localhost:7001"],"dataDir":"~/.ironfish-simulator/node-4f98","verbose":false,"importGenesisAccount":false}'
}
started node: node-4f98
simulation encountered rejected!, shutting down...
cleaning up node-6c93...
cleaning up node-4f98...
cleaning up data dirs
removing data dir: /Users/austino/.ironfish-simulator/node-6c93
removing data dir: /Users/austino/.ironfish-simulator/node-4f98
exiting...
running simulation test... done
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```
## Documentation

```
[ ] Yes
[X] No
```

## Breaking Change

```
[X] Yes
```

Changes the `Simulation` interface from 
```
  run(logger: Logger, options?: { persist?: boolean; duration?: number }): Promise<void>
```
to
```
  run(simulator: Simulator, logger: Logger): Promise<void>
```
Anyone actively writing simulations needs to change their simulation function signature to follow the new interface definition.

This change does not break the core `Ironfish` sdk

### DIFF
--- a/simulator/src/commands/start.ts
+++ b/simulator/src/commands/start.ts
@@ -10,8 +10,6 @@ import { Simulator } from '../simulator'
 export abstract class Start extends Command {
   static description = 'Start a simulation'
 
-  static simulator: Simulator
-
   static args = [
     {
       name: 'simulation',
@@ -65,13 +63,13 @@ export abstract class Start extends Command {
     // The simulator is created here because oclif catches errors so we can't throw them
     // and handle `uncaughtException` in the simulator. Having this try-catch block is a workaround
     // to ensure the simulator gracefully exits when an error occurs.
-    Start.simulator = new Simulator(logger, { persist, duration })
+    const simulator = new Simulator(logger, { persist, duration })
 
     try {
-      await simulation.run(Start.simulator, logger)
+      await simulation.run(simulator, logger)
     } catch (e) {
       logger.error(`simulation encountered ${String(e)}, shutting down...`)
-      Start.simulator.exit(1)
+      simulator.exit(1)
     }
 
     CliUx.ux.action.stop(`stop simulation ${simName}`)
--- a/simulator/src/commands/start.ts
+++ b/simulator/src/commands/start.ts
@@ -5,9 +5,12 @@ import { createRootLogger } from '@ironfish/sdk'
 import { CliUx, Command, Config } from '@oclif/core'
 import { Flags } from '@oclif/core'
 import { SIMULATIONS } from '../simulations'
+import { Simulator } from '../simulator'
 
 export abstract class Start extends Command {
   static description = 'Start a simulation'
+
+  static simulator: Simulator
 
   static args = [
     {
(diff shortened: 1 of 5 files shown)


simulator/src/commands/start.ts
```diff
@@ -58,8 +61,20 @@ export abstract class Start extends Command {
...
     // need to remove the spinner
     CliUx.ux.action.start(`running simulation ${simName}`)
-    await simulation.run(logger, { persist, duration })
-    CliUx.ux.action.start(`stop simulation ${simName}`)
-    this.exit()
+
+    // The simulator is created here because oclif catches errors so we can't throw them
+    // and handle `uncaughtException` in the simulator. Having this try-catch block is a workaround
+    // to ensure the simulator gracefully exits when an error occurs.
+    Start.simulator = new Simulator(logger, { persist, duration })
```
danield9tqh, line 68: Do we need to set this to `Start.simulator`? Could this just be a normal const?

holahula: no, made a normal const

jowparks:
Seems fine, looks like you call `logger.error` so I am guessing the colorization will be taken care of to surface the error.

jowparks (approved):

PR #3779, opened 2023-04-12
adds markSpent parameter to createTransaction RPC

hughy:
## Summary

optionally marks notes used in creating a transaction as spent. this is useful for creating multiple raw transactions in succession without spending the same note(s) in more than one transaction.

we have an example of this use case in the airdrop: we need to create many raw transactions and then post them separately before broadcasting to the network.

## Testing Plan

- added unit tests
- manual testing
  - observed notes using 'wallet:notes'
  - created raw transaction with 'markSpent' set to true by modifying send command
  - confirmed that notes were marked as spent

## Documentation

```
[X] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -857,6 +857,7 @@ export class Wallet {
     expiration?: number
     expirationDelta?: number
     confirmations?: number
+    markSpent?: boolean
   }): Promise<RawTransaction> {
     const heaviestHead = this.chain.head
     if (heaviestHead === null) {
@@ -936,6 +937,26 @@ export class Wallet {
           fee: raw.fee,
           account: options.account,
           confirmations: confirmations,
+        })
+      }
+
+      if (options.markSpent) {
+        await this.walletDb.db.transaction(async (tx) => {
+          for (const spend of raw.spends) {
+            const decryptedNote = await options.account.getDecryptedNote(spend.note.hash())
+
+            Assert.isNotUndefined(decryptedNote)
+
+            await this.walletDb.saveDecryptedNote(
+              options.account,
+              spend.note.hash(),
+              {
+                ...decryptedNote,
+                spent: true,
+              },
+              tx,
+            )
+          }
         })
       }
 
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -947,6 +947,13 @@ export class Wallet {
 
             Assert.isNotUndefined(decryptedNote)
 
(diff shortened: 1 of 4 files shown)


ironfish/src/wallet/wallet.ts
```diff
@@ -939,6 +940,26 @@ export class Wallet {
...
       }
 
+      if (options.markSpent) {
+        await this.walletDb.db.transaction(async (tx) => {
+          for (const spend of raw.spends) {
+            const decryptedNote = await options.account.getDecryptedNote(spend.note.hash())
+
+            Assert.isNotUndefined(decryptedNote)
+
+            await this.walletDb.saveDecryptedNote(
```
dguenther, line 950: Should this also be calling this?

```ts
await this.walletDb.deleteUnspentNoteHash(this, spentNoteHash, spentNote, tx)
```

otherwise next time `createSpendsForAsset` calls `getUnspentNotes`, I think it'll still return the note

dguenther: I'd probably expect something like this code block, but I haven't looked enough to see if all of these are necessary: https://github.com/iron-fish/ironfish/blob/5f68460f860d5447f9152d218effa466b58d798b/ironfish/src/wallet/account.ts#L495-L512

hughy: Yes! We need to update the index, too.

I was also looking at your comment here: https://github.com/iron-fish/ironfish/pull/3721/files#r1164678053, and I think that that might be a better option than adding a permanent parameter 🤔

hughy: > I'd probably expect something like this code block, but I haven't looked enough to see if all of these are necessary:
> 
> https://github.com/iron-fish/ironfish/blob/5f68460f860d5447f9152d218effa466b58d798b/ironfish/src/wallet/account.ts#L495-L512

We can't update the `nullifierToTransactionHash` index here because we don't have the transaction hash until the transaction is posted

dguenther: I'm not certain what people will want to do long-term -- They almost definitely will want to specify which notes will go into a transaction though, but that's probably a bit trickier to get ready.

I could see a use for this, but it's a bit dangerous in that there isn't a way to un-spend a note without adding a corresponding transaction and waiting for it to expire.

Adding a temporary hack would probably be the least commitment, and I'm okay with that since it'd just live in the `airdrop` branch

NullSoldier:
This is a bit dangerous because if the TX gets dropped and never posted then the note will be marked spent permanently.

NullSoldier:
There is no way to unmark these notes as spent.

PR #3789, opened 2023-04-14
Add transaction version, double spend verification before broadcast

hairtail:
## Summary
`verifyCreatedTransaction` is a subset of `VerifyNewTransaction`. `VerifyNewTransaction` has more verifications like `transaction version`, `double spend`, `transaction proof`, which are really necessary before the transaction is broadcasted to the p2p network in my opinion.
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/chain/broadcastTransaction.ts
+++ b/ironfish/src/rpc/routes/chain/broadcastTransaction.ts
@@ -32,11 +32,11 @@ export const BroadcastTransactionResponseSchema: yup.ObjectSchema<BroadcastTrans
 router.register<typeof BroadcastTransactionRequestSchema, BroadcastTransactionResponse>(
   `${ApiNamespace.chain}/broadcastTransaction`,
   BroadcastTransactionRequestSchema,
-  (request, node): void => {
+  async (request, node): Promise<void> => {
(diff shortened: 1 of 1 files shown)


danield9tqh:
I think this is a little subjective. `verifyNewTransaction` is used when receiving new transactions from peers that the node has no previous knowledge of. So verifying every aspect of the transaction makes sense. It also verifies everything before gossiping because we don't want to programmatically bring down the network by re-gossiping invalid transactions.

This RPC endpoint is used ad-hoc though. It's meant to be used to broadcast a transaction that was created by a trusted source so not everything needs to be verified. For example what if a user wanted to broadcast a double spend transaction because they had some knowledge that their previous transaction was on a fork. This change would not allow more nuanced cases like this. I think this endpoint should do as little verification as possible because the transaction has already been verified elsewhere.

Can you explain a little more about the use case of why you want to catch double spends in this endpoint? Maybe there's a better solution for it that doesn't involve this change

dguenther:
Going to close this for now, since I think the current intent of the endpoint is to broadcast transactions created by a trusted source, and we'd need more information about intended usage to justify this change.

PR #3791, opened 2023-04-14
avoids deserializing values in getAllKeysIter

hughy:
## Summary

we delete old records from wallet datastores in batches in 'cleanupDeletedAccounts' to avoid blocking the node during lengthy delete operations.

however, we use 'getAllKeysIter' to iterate over keys and delete them one by one. 'getAllKeysIter' calls 'getAllIter' which iterates over keys and values. during this iteration 'getAllIter' decodes the key and deserializes the value. this deserialization is problematic if the value encoding has changed.

for example, the latest network reset changed asset generation. the database
        fails to deserialize old transaction data because the asset identifiers
        in the old transactions are now invalid:

```
Error: InvalidAssetIdentifier
    at <anonymous> (/Users/hughy/Code/ironfish/ironfish/src/primitives/transaction.ts:95:27)
    at Function.from (<anonymous>)
    at Transaction (/Users/hughy/Code/ironfish/ironfish/src/primitives/transaction.ts:91:24)
    at TransactionValueEncoding.deserialize (/Users/hughy/Code/ironfish/ironfish/src/wallet/walletdb/transactionValue.ts:58:25
```

we can solve this problem by preventing 'getAllKeysIter' from deserializing values unnecessarily.

- implements '_getAllIter' to iterate over database entries without decoding keys and deserializing values
- use '_getAllIter' in 'getAllKeysIter' and decode keys before yielding

'_getAllIter' currently has to handle both serialized and deserialized values because a transaction cache contains values that have already been deserialized. in the future we might revisit the transaction cache and change it to only store encoded keys and serialized values to match the database.

## Testing Plan

manual testing:

    start a node on v0.1.75
    mint an asset
    upgrade ironfish to v0.1.76
    run 'ironfish reset' to reset the chain
    start the node
    a. without changes applied, you'll run into an error during account data cleanup
    b. with changes applied the node will start and clean up old data in the background

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/storage/levelup/store.ts
+++ b/ironfish/src/storage/levelup/store.ts
@@ -78,6 +78,20 @@ export class LevelupStore<Schema extends DatabaseSchema> extends DatabaseStore<S
     keyRange?: DatabaseKeyRange,
     iteratorOptions?: DatabaseIteratorOptions,
   ): AsyncGenerator<[SchemaKey<Schema>, SchemaValue<Schema>]> {
+    for await (const [key, value] of this._getAllIter(transaction, keyRange, iteratorOptions)) {
+      if (value instanceof Buffer) {
+        yield [this.decodeKey(key), this.valueEncoding.deserialize(value)]
+      } else {
+        yield [this.decodeKey(key), value]
+      }
+    }
+  }
+
+  async *_getAllIter(
+    transaction?: IDatabaseTransaction,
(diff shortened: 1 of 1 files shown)


hughy:
Closing this because #3797 resolves the issue, albeit in a different way.

We can revisit this change later if we want to avoid deserializing values during key iteration.

PR #3797, opened 2023-04-18
clears walletdb datastores during reset

hughy:
## Summary

following a network reset the walletdb may have data that is incompatible with database schemas. for instance, if asset identifiers in transactions or notes change, then any datastores that encode that data must be migrated or cleared.

we avoid deleting the walletdb during 'reset' so that we do not delete user keys. however, the node will fail to start if it needs to apply migrations on incompatible data or even if it needs to deserialize incompatible data in the course of deleting it.

to resolve this we can use 'clear' on each walletdb datastore during the 'reset' command. clear will delete all keys within a range without deserializing the values.

note that clear soft-deletes data that is cleanup up later during database compactions rather than deleting immediately.

- clears each walletdb datastore, except accounts and meta, during 'reset'
- adds missing datastores to cleanupDeletedAccounts

## Testing Plan

manual testing:
1. start with wallet on earlier version before network reset (e.g., v0.1.70)
2. upgrade to v0.1.76
3. run `ironfish reset`
4. start the node

tested with pool account from phase 3 (~10k transactions and ~10k notes) and resetting wallet data using these changes took less than 10s.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/reset.ts
+++ b/ironfish-cli/src/commands/reset.ts
@@ -1,7 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { HOST_FILE_NAME, IronfishNode } from '@ironfish/sdk'
+import { HOST_FILE_NAME, IDatabaseStore, IronfishNode } from '@ironfish/sdk'
 import { CliUx, Flags } from '@oclif/core'
 import fsAsync from 'fs/promises'
 import { IronfishCommand } from '../command'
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/commands/reset.ts
```diff
@@ -80,6 +80,32 @@ export default class Reset extends IronfishCommand {
...
+      walletDb.nonChainNoteHashes,
+      walletDb.transactions,
+      walletDb.sequenceToTransactionHash,
+      walletDb.pendingTransactionHashes,
+      walletDb.timestampToTransactionHash,
+      walletDb.assets,
+      walletDb.nullifierToTransactionHash,
+      walletDb.unspentNoteHashes,
+    ]
+
```
EvanJRichard, line 102: this list isn't reused anywhere right? was wondering if it's worth setting it up as a constant (or something like that) somewhere, but if this is the only place it's used, no point

EvanJRichard (approved):
Straightforward improvement. :shipit:

PR #3815, opened 2023-04-21
feat(cli): add memo to interactive mode and display it in confirmation

mat-if:
## Summary

image

image

Closes IFL-690

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/send.ts
+++ b/ironfish-cli/src/commands/wallet/send.ts
@@ -99,6 +99,7 @@ export class Send extends IronfishCommand {
     let assetId = flags.assetId
     let to = flags.to?.trim()
     let from = flags.account?.trim()
+    let memo = flags.memo?.trim()
 
     const client = await this.sdk.connectRpc()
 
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/wallet/send.ts
```diff
@@ -260,7 +267,7 @@ export class Send extends IronfishCommand {
       )} plus a transaction fee of ${CurrencyUtils.renderIron(
         fee,
         true,
-      )} to ${to} from the account ${from}`,
+      )} to ${to} from the account "${from}" with the memo "${memo}"`,
```
EvanJRichard, line 270: Is it okay that this might sometimes display `with the memo ""`?

EvanJRichard (approved):
One UX question, but looks right to me.

ironfish-cli/src/commands/wallet/send.ts
mat-if: Well, it's true.

I don't feel one way or the other, but I do feel like this should be changed to be more list-like instead of a sentence, if we want to improve it.

Something like:
"You are about to send the following transaction:
Amount: 1234
Fee: 0.00001
Memo: ""
To: asdf12345"

EvanJRichard: Makes sense, let's leave it as is :+1:

rohanjadvani (approved):

NullSoldier (approved):

PR #3816, opened 2023-04-21
Separate telemetry by endpoint

danield9tqh:
## Summary
Send telemetry to the correct API endpoint depending on the network id

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/node.ts
+++ b/ironfish/src/node.ts
@@ -113,14 +113,12 @@ export class IronfishNode {
       metrics,
       workerPool,
       localPeerIdentity: privateIdentityToIdentity(identity),
-      defaultTags: [
-        { name: 'version', value: pkg.version },
-        { name: 'networkId', value: networkId.toString() },
-      ],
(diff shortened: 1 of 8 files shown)


ironfish/src/node.ts
```diff
@@ -106,6 +106,11 @@ export class IronfishNode {
 
     const identity = privateIdentity || new BoxKeyPair()
 
+    let telemetryApiHost = 'https://api.ironfish.network'
+    if (networkId === 0) {
+      telemetryApiHost = 'https://testnet.api.ironfish.network'
+    }
```
NullSoldier, line 112: I think we should do this but just do this in the config, not hard coded.

NullSoldier (approved):

NullSoldier (approved):

NullSoldier:
I tested this manually, and it loaded the right endpoint in mainnet and testnet. I did not test a custom network definition.

PR #3828, opened 2023-04-24
adds support for loading unspent notes in different orderings

hughy:
## Summary

by default unspent notes for a given account and asset are always loaded in order of increasing block sequence.

it may be advantageous sometimes to load notes by value so that transactions can be funded with larger notes first.

renames 'loadUnspentNoteHashes' and 'loadUnspentNotes' to reflect that they are loaded by sequence.

adds walletDb methods to support loading unspent notes in order of value instead of sequence (i.e., 'loadUnspentNoteHashesByValue').

adds 'reverse' parameter to each of these methods to support loading in reverse order.

## Testing Plan

- adds unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/account.test.ts
+++ b/ironfish/src/wallet/account.test.ts
@@ -420,7 +420,7 @@ describe('Accounts', () => {
       await node.wallet.updateHead()
 
       let unspentA = await AsyncUtils.materialize(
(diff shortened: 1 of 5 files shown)


hughy:
Closing this for now. I think that adding ordering to notes endpoints may be out of scope for now.

PR #3842, opened 2023-04-25
Add RPC for validating address

danield9tqh:
## Summary
Adding endpoint for validating a public address

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/chain/isValidPublicAddress.ts
```diff
@@ -0,0 +1,38 @@
...
+      valid: yup.boolean().defined(),
+    })
+    .defined()
+
+router.register<typeof IsValidPublicAddressRequestSchema, IsValidPublicAddressResponse>(
+  `${ApiNamespace.chain}/isValidPublicAddress`,
+  IsValidPublicAddressRequestSchema,
+  (request): void => {
+    request.end({
+      valid: isValidPublicAddress(request.data.address),
```
ygao76, line 35: Should we use this `isValidPublicAddress` https://github.com/iron-fish/ironfish/blob/bb99849b03d744454f1098029f836704af94ed7a/ironfish/src/wallet/validator.ts#L12

dguenther: Yeah, it won't practically make a difference, but I think this is a good idea. We should have a cleanup task for standardizing all uses of the function to go through the SDK, and this function is exported through the SDK + used in the wallet

danield9tqh: Yea sounds good I updated it

holahula (approved):

hughy (approved):

PR #3854, opened 2023-04-26
ifl-744 add test/updates for calculating raw transaction size

jowparks:
## Summary
Updates size to match https://github.com/iron-fish/ironfish/blob/master/ironfish/src/primitives/transaction.ts#L44

We have to account for change notes as well. We can determine if there will be change for each asset by calculating the total spends/mints/burns/outputs. If they equal 0, there will be no change note
## Testing Plan
tests pass
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust-nodejs/src/structs/note_encrypted.rs
+++ b/ironfish-rust-nodejs/src/structs/note_encrypted.rs
@@ -28,12 +28,6 @@ pub const ENCRYPTED_NOTE_PLAINTEXT_LENGTH: u32 = ENCRYPTED_NOTE_SIZE as u32 + MA
 #[napi]
 pub const ENCRYPTED_NOTE_LENGTH: u32 =
     NOTE_ENCRYPTION_KEY_LENGTH + ENCRYPTED_NOTE_PLAINTEXT_LENGTH + 96;
-//  32 value commitment
-//+ 32 note commitment
-//+ 32 ephemeral public key
-//+ 120 encrypted note
-//+ 80 note encryption keys
-//= 296 bytes
 
 #[napi(js_name = "NoteEncrypted")]
 pub struct NativeNoteEncrypted {
--- a/ironfish/src/primitives/rawTransaction.test.ts
+++ b/ironfish/src/primitives/rawTransaction.test.ts
@@ -84,6 +84,10 @@ describe('RawTransaction', () => {
     const mintedValue = valuesByAsset.get(asset.id())
     Assert.isNotUndefined(mintedValue)
     expect(mintedValue).toEqual(1n)
+
+    // should have same size for posted transaction and estimated size from raw transaction
+    const serializedPost = posted.serialize()
+    expect(raw.postedSize(account.publicAddress)).toEqual(serializedPost.byteLength)
   })
 
   it('should throw an error if the max mint value is exceeded', async () => {
--- a/ironfish/src/primitives/rawTransaction.ts
+++ b/ironfish/src/primitives/rawTransaction.ts
@@ -6,9 +6,12 @@ import {
   AMOUNT_VALUE_LENGTH,
   ASSET_LENGTH,
   generateKeyFromPrivateKey,
+  PROOF_LENGTH,
   Transaction as NativeTransaction,
   TRANSACTION_EXPIRATION_LENGTH,
   TRANSACTION_FEE_LENGTH,
+  TRANSACTION_PUBLIC_KEY_RANDOMNESS_LENGTH,
+  TRANSACTION_SIGNATURE_LENGTH,
 } from '@ironfish/rust-nodejs'
 import { Asset, ASSET_ID_LENGTH } from '@ironfish/rust-nodejs'
 import bufio from 'bufio'
@@ -16,6 +19,7 @@ import { Witness } from '../merkletree'
 import { NoteHasher } from '../merkletree/hasher'
 import { Side } from '../merkletree/merkletree'
 import { CurrencyUtils } from '../utils/currency'
+import { AssetBalances } from '../wallet/assetBalances'
 import { BurnDescription } from './burnDescription'
 import { Note } from './note'
 import {
@@ -54,17 +58,44 @@ export class RawTransaction {
     >
   }[] = []
 
-  size(): number {
+  postedSize(publicAddress: string): number {
     let size = 0
+    size += 1 // version
     size += 8 // spends length
     size += 8 // notes length
-    size += 8 // fee
-    size += 4 // expiration
-    size += 64 // signature
-    size += this.outputs.length * NOTE_ENCRYPTED_SERIALIZED_SIZE_IN_BYTE
-    size += this.mints.length * (ASSET_LENGTH + 8)
-    size += this.burns.length * (ASSET_ID_LENGTH + 8)
+    size += 8 // mints length
+    size += 8 // burns length
+    size += TRANSACTION_FEE_LENGTH // fee
+    size += TRANSACTION_EXPIRATION_LENGTH // expiration
+    size += TRANSACTION_PUBLIC_KEY_RANDOMNESS_LENGTH // public key randomness
     size += this.spends.length * SPEND_SERIALIZED_SIZE_IN_BYTE
+    size += this.outputs.length * NOTE_ENCRYPTED_SERIALIZED_SIZE_IN_BYTE
+    size +=
+      this.mints.length *
+      (PROOF_LENGTH + ASSET_LENGTH + AMOUNT_VALUE_LENGTH + TRANSACTION_SIGNATURE_LENGTH)
+    size += this.burns.length * (ASSET_ID_LENGTH + 8)
+    size += TRANSACTION_SIGNATURE_LENGTH // signature
+
+    // Each asset might have a change note, which would need to be accounted for
+    const assetTotals = new AssetBalances()
+    for (const mint of this.mints) {
+      const asset = new Asset(publicAddress, mint.name, mint.metadata)
+      assetTotals.increment(asset.id(), mint.value)
+    }
+    for (const burn of this.burns) {
+      assetTotals.increment(burn.assetId, -burn.value)
+    }
+    for (const spend of this.spends) {
+      assetTotals.increment(spend.note.assetId(), -spend.note.value())
+    }
+    for (const output of this.outputs) {
+      assetTotals.increment(output.note.assetId(), output.note.value())
+    }
+    for (const [, value] of assetTotals) {
+      if (value !== 0n) {
+        size += NOTE_ENCRYPTED_SERIALIZED_SIZE_IN_BYTE
+      }
+    }
     return size
   }
 
--- a/ironfish/src/primitives/spend.ts
+++ b/ironfish/src/primitives/spend.ts
@@ -4,7 +4,7 @@
 
 import { Nullifier } from './nullifier'
 
-export const SPEND_SERIALIZED_SIZE_IN_BYTE = 388
+export const SPEND_SERIALIZED_SIZE_IN_BYTE = 356
 
 export interface Spend {
   nullifier: Nullifier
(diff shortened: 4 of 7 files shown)


ironfish/src/primitives/rawTransaction.ts
```diff
@@ -53,17 +57,54 @@ export class RawTransaction {
...
-    size += this.mints.length * (ASSET_LENGTH + 8)
+    size +=
+      this.mints.length *
+      (PROOF_LENGTH + ASSET_LENGTH + AMOUNT_VALUE_LENGTH + TRANSACTION_SIGNATURE_LENGTH)
     size += this.burns.length * (ASSET_ID_LENGTH + 8)
-    size += this.spends.length * SPEND_SERIALIZED_SIZE_IN_BYTE
+    size += TRANSACTION_SIGNATURE_LENGTH // signature
+
+    // Each asset might have a change note, which would need to be accounted for
+    const assetTotals = new BufferMap<bigint>()
```
jowparks, line 79: If there is only a mint, will there be an output to the sender as a result of the mint? Test is passing but I wanted to double check.

ironfish/src/primitives/spend.ts
```diff
@@ -4,7 +4,7 @@
 
 import { Nullifier } from './nullifier'
 
-export const SPEND_SERIALIZED_SIZE_IN_BYTE = 388
+export const SPEND_SERIALIZED_SIZE_IN_BYTE = 356
```
jowparks, line 7: This number was wrong.

ironfish/src/primitives/rawTransaction.ts
mat-if: Yes - it should function the same as if you create a transaction with spends and no outputs, it would collect that and create a change note of the full value

ironfish/src/primitives/rawTransaction.ts
```diff
@@ -53,17 +57,54 @@ export class RawTransaction {
     >
   }[] = []
 
-  size(): number {
+  size(senderPublicAddress: string): number {
```
mat-if, line 60: What do you think about renaming this? I feel like `.size()` sets a specific expectation. Maybe `postedSize()`, I don't know. I don't feel super strongly here

ironfish/src/primitives/rawTransaction.ts
```diff
@@ -53,17 +57,54 @@ export class RawTransaction {
...
+        assetValue ? assetValue - spend.note.value() : -spend.note.value(),
+      )
+    }
+    for (const output of this.outputs) {
+      const assetValue = assetTotals.get(output.note.assetId())
+      assetTotals.set(
+        output.note.assetId(),
+        assetValue ? assetValue + output.note.value() : output.note.value(),
+      )
+    }
```
mat-if, line 102: In the rust side, we have a `ValueBalances` struct, I wonder if it's worth exposing this struct to the TS side and using it.  Filing a ticket since that's non-trivial due to the bigint stuff, doesn't need to be dealt with in this PR. Overall, it's way more work but leads to less duplicated logic

https://linear.app/if-labs/issue/IFL-763/expose-valuebalances-to-typescript

ironfish/src/primitives/spend.ts
mat-if: Yeah, that's not ideal. I filed a ticket to look at these later https://linear.app/if-labs/issue/IFL-764/const-cleanup

ironfish/src/primitives/rawTransaction.ts
```diff
@@ -53,17 +57,54 @@ export class RawTransaction {
...
+      this.mints.length *
+      (PROOF_LENGTH + ASSET_LENGTH + AMOUNT_VALUE_LENGTH + TRANSACTION_SIGNATURE_LENGTH)
     size += this.burns.length * (ASSET_ID_LENGTH + 8)
-    size += this.spends.length * SPEND_SERIALIZED_SIZE_IN_BYTE
+    size += TRANSACTION_SIGNATURE_LENGTH // signature
+
+    // Each asset might have a change note, which would need to be accounted for
+    const assetTotals = new BufferMap<bigint>()
+    for (const mint of this.mints) {
+      const asset = new Asset(senderPublicAddress, mint.name, mint.metadata)
```
hughy, line 81: Will creating an asset here have any performance implications?

It looks like we only need the ID here as a way to group mints of the same asset, so I'm wondering if there's another way around this.

ironfish/src/primitives/rawTransaction.ts
hughy: I think that the usual result of a mint is to give the sender an output note for the minted asset

ironfish/src/primitives/rawTransaction.test.ts
```diff
@@ -84,6 +84,10 @@ describe('RawTransaction', () => {
     const mintedValue = valuesByAsset.get(asset.id())
     Assert.isNotUndefined(mintedValue)
     expect(mintedValue).toEqual(1n)
+
+    // should have same size for posted transaction and estimated size from raw transaction
+    const serializedPost = posted.serialize()
+    expect(raw.size(account.publicAddress)).toEqual(Buffer.byteLength(serializedPost))
```
mat-if, line 90: 💯

ironfish/src/primitives/rawTransaction.ts
```diff
@@ -53,17 +57,54 @@ export class RawTransaction {
...
+    size += TRANSACTION_FEE_LENGTH // fee
+    size += TRANSACTION_EXPIRATION_LENGTH // expiration
+    size += TRANSACTION_PUBLIC_KEY_RANDOMNESS_LENGTH // public key randomness
+    size += this.spends.length * SPEND_SERIALIZED_SIZE_IN_BYTE
     size += this.outputs.length * NOTE_ENCRYPTED_SERIALIZED_SIZE_IN_BYTE
-    size += this.mints.length * (ASSET_LENGTH + 8)
+    size +=
+      this.mints.length *
+      (PROOF_LENGTH + ASSET_LENGTH + AMOUNT_VALUE_LENGTH + TRANSACTION_SIGNATURE_LENGTH)
     size += this.burns.length * (ASSET_ID_LENGTH + 8)
```
hughy, line 75: Should this be `(ASSET_ID_LENGTH + AMOUNT_VALUE_LENGTH)` for consistency?

ironfish/src/primitives/rawTransaction.ts
hughy, line 60: Do you think we should rename this to something like `postedSize`?

mat-if (approved):

ironfish/src/primitives/rawTransaction.ts
hughy: There's also a class on the TS side, `AssetBalances`, that does something similar

ironfish/src/primitives/rawTransaction.ts
mat-if: It will run a few blake2s hashes and check for validity, but it's a fairly cheap check on its own. It's not _ideal_, but I think it's worth the hit to get an exact size. I think the alternative is a best-guess where we might be wrong since we can't calculate if change notes will be created or not.

Ran a quick test on my laptop:
```
Elapsed: 291.958µs
Elapsed: 591.791µs
Elapsed: 235.958µs
Elapsed: 118.75µs
Elapsed: 236.208µs
```

Code:
``` rust
for i in 0..5 {
    let start = Instant::now();
    let a = Asset::new(public_address, &format!("hello {}", i), "metadata").unwrap();
    let elapsed = start.elapsed();
    println!("Elapsed: {:?}", elapsed);
}
```

ironfish/src/primitives/rawTransaction.ts
```diff
@@ -53,17 +57,54 @@ export class RawTransaction {
...
-    size += this.spends.length * SPEND_SERIALIZED_SIZE_IN_BYTE
+    size += TRANSACTION_SIGNATURE_LENGTH // signature
+
+    // Each asset might have a change note, which would need to be accounted for
+    const assetTotals = new AssetBalances()
+    for (const mint of this.mints) {
+      const asset = new Asset(senderPublicAddress, mint.name, mint.metadata)
```
dguenther, line 81: I think the `Asset` constructor takes a spending key, not a public address, even though it just converts to a public address internally. I'd have expected this to error, but maybe the two are compatible with each other now.

dguenther (changes requested):

ironfish/src/primitives/noteEncrypted.ts
```diff
@@ -13,6 +13,7 @@ import bufio from 'bufio'
 import { Serde } from '../serde'
 import { Note } from './note'
 
+//192 + 328 = 520
```
dguenther, line 16: I think we can remove this for now. I'm getting a little wary of adding comments that may end up out-of-date, hah

ironfish/src/primitives/rawTransaction.test.ts
```diff
@@ -84,6 +84,10 @@ describe('RawTransaction', () => {
     const mintedValue = valuesByAsset.get(asset.id())
     Assert.isNotUndefined(mintedValue)
     expect(mintedValue).toEqual(1n)
+
+    // should have same size for posted transaction and estimated size from raw transaction
+    const serializedPost = posted.serialize()
+    expect(raw.postedSize(account.publicAddress)).toEqual(Buffer.byteLength(serializedPost))
```
dguenther, line 90: Since `serializedPost` is already a Buffer, let's prefer calling byteLength directly:

```suggestion
    expect(raw.postedSize(account.publicAddress)).toEqual(serializedPost.byteLength)
```

ironfish/src/primitives/rawTransaction.test.ts
```diff
@@ -84,6 +84,10 @@ describe('RawTransaction', () => {
     const mintedValue = valuesByAsset.get(asset.id())
     Assert.isNotUndefined(mintedValue)
     expect(mintedValue).toEqual(1n)
```
jowparks, line 86: This is a nice future proofing test

ironfish-rust-nodejs/src/structs/note_encrypted.rs
```diff
@@ -28,12 +28,6 @@ pub const ENCRYPTED_NOTE_PLAINTEXT_LENGTH: u32 = ENCRYPTED_NOTE_SIZE as u32 + MA
 #[napi]
 pub const ENCRYPTED_NOTE_LENGTH: u32 =
     NOTE_ENCRYPTION_KEY_LENGTH + ENCRYPTED_NOTE_PLAINTEXT_LENGTH + 96;
-//  32 value commitment
```
ygao76, line 31: is this comment calculation wrong?

ironfish/src/primitives/spend.ts
ygao76: Can you list out the calculation in the comment like what we have in the existing code so that we can keep this value up to date?

ygao76: // 32 nullifier
// 4 size
// 160?? commitment

ironfish-rust-nodejs/src/structs/note_encrypted.rs
jowparks: Yea it was, it also will go stale with any changes to the code above, you can hover the constant and get the value

ironfish/src/primitives/rawTransaction.ts
```diff
@@ -54,17 +58,54 @@ export class RawTransaction {
...
     size += this.burns.length * (ASSET_ID_LENGTH + 8)
-    size += this.spends.length * SPEND_SERIALIZED_SIZE_IN_BYTE
+    size += TRANSACTION_SIGNATURE_LENGTH // signature
+
+    // Each asset might have a change note, which would need to be accounted for
+    const assetTotals = new AssetBalances()
+    for (const mint of this.mints) {
+      const asset = new Asset(publicAddress, mint.name, mint.metadata)
+      const assetValue = assetTotals.get(asset.id())
+      assetTotals.set(asset.id(), assetValue ? assetValue + mint.value : mint.value)
```
danield9tqh, line 84: Should we use `assetTotals.increment()` here?

jowparks: Oh yea that is real nice! Cleans up this code a bunch. TY

ygao76 (approved):

danield9tqh (approved):

dguenther (approved):

PR #3856, opened 2023-04-27
feat(tools): Version bump script can bump all version numbers

mat-if:
## Summary

Makes the default version bump the "minor" version instead of the patch as per the documentation outlined here: https://github.com/iron-fish/website/pull/422

## Testing Plan

Ran the following and inspected the git diff:
`node ./tools/version-bump.js`
`node ./tools/version-bump.js --rust`
`node ./tools/version-bump.js --patch`
`node ./tools/version-bump.js --patch --rust`
`node ./tools/version-bump.js --major`
`node ./tools/version-bump.js --major --rust`

## Documentation

```
[x] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/tools/version-bump.js
+++ b/tools/version-bump.js
@@ -33,12 +33,12 @@ const bumpVersion = (version) => {
 
 const bumpMajor = (version) => {
   const { major, minor, patch } = parseVersion(version)
-  return `${parseInt(major) + 1}.${minor}.${patch}`
+  return `${parseInt(major) + 1}.0.0`
 }
 
 const bumpDefault = (version) => {
   const { major, minor, patch } = parseVersion(version)
-  return `${major}.${parseInt(minor) + 1}.${patch}`
+  return `${major}.${parseInt(minor) + 1}.0`
 }
 
 const bumpPatch = (version) => {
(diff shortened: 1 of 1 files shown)


jowparks:
I am guessing you tested this? Can you add test out of file. Looks good

jowparks (approved):

tools/version-bump.js
```diff
@@ -18,9 +18,31 @@ const RUST_PACKAGE = path.join(__dirname, '../ironfish-rust-nodejs/package.json'
...
+    return bumpPatch(version)
+  } else {
+    return bumpDefault(version)
+  }
+}
+
+const bumpMajor = (version) => {
+  const { major, minor, patch } = parseVersion(version)
+  return `${parseInt(major) + 1}.${minor}.${patch}`
+}
```
NullSoldier, line 37: this doesn't set the others back to 0, should it?

NullSoldier:
I pulled it down and tested, it but the versions of the others don't reset back to 0 upon a larger point update. Is that expected?

tools/version-bump.js
mat-if: Oh right, good catch! Fixed

PR #3859, opened 2023-04-27
adds a simulation for fetching transactions from getAccountTransaction

hughy:
## Summary

we have heard reports that the 'wallet/getAccountTransaction' endpoint does not always return every transaction.

this simulation creates three nodes and has two 'sender' nodes repeatedly send transactions to the 'receiver' node. for each transaction the simulation checks that the transaction can be found on the sender node and the receiver node using the 'wallet/getAccountTransaction' endpoint over HTTP.

running the simulation for ~1000 blocks did not find any transactions missing from either the sender or receiver.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ygao76 (approved):

holahula:
lgtm, i'll add default support for HTTP like it is for TCP

PR #3861, opened 2023-04-28
implements wallet/getNotes endpoint with pagination

hughy:
## Summary

adds a wallet RPC endpoint to retrieve notes for an account. uses pagination instead of streaming.

derives page tokens from the database key for the last note in a page.

next steps:
- add filtering to request parameters

## Testing Plan

- adds unit tests

## Documentation

```
[X] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/wallet/getNotes.ts
```diff
@@ -0,0 +1,116 @@
...
+    nextPageToken: yup.string(),
+  })
+  .defined()
+
+router.register<typeof GetNotesRequestSchema, GetNotesResponse>(
+  `${ApiNamespace.wallet}/getNotes`,
+  GetNotesRequestSchema,
+  async (request, node): Promise<void> => {
+    const account = getAccount(node, request.data.account)
+    const pageSize = request.data.pageSize ?? 100
```
holahula, line 70: thoughts on implementing a max page size?

hughy: That's a good idea. 100 was an arbitrary default choice 😅 

Some max/default size based on the size of the response would be better

hughy: Thinking about this a bit more and I worry about setting a max page size that's too small (or too big) for integration developers.

I think that's something that I would leave up to the node runner to decide, but we don't have a pattern yet for per-endpoint configuration.

holahula: sgtm

hughy:
Moving this back into draft until dependent PRs are in

ironfish/src/wallet/account.ts
```diff
@@ -84,8 +84,10 @@ export class Account {
     }
   }
 
-  async *getNotes(): AsyncGenerator<DecryptedNoteValue & { hash: Buffer }> {
-    for await (const decryptedNote of this.walletDb.loadDecryptedNotes(this)) {
+  async *getNotes(
+    keyRange?: DatabaseKeyRange,
+  ): AsyncGenerator<DecryptedNoteValue & { hash: Buffer }> {
+    for await (const decryptedNote of this.walletDb.loadDecryptedNotes(this, keyRange)) {
```
ygao76, line 90: Im a little confused here, walletDb.loadDecryptedNotes() takes IDatabaseTransaction, how does DatabaseKeyRange converts to IDatabaseTransaction?

ironfish/src/rpc/routes/wallet/getNotes.ts
```diff
@@ -0,0 +1,72 @@
...
+export type GetNotesResponse = {
+  notes: Array<RpcWalletNote>
+  nextPageToken: string | null
+}
+
+export const GetNotesRequestSchema: yup.ObjectSchema<GetNotesRequest> = yup
+  .object({
+    account: yup.string().trim(),
+    pageSize: yup.number().min(1),
+    pageToken: yup.string(),
```
ygao76, line 27: The name pageToken seems not very clear, should we call it a cursor or something like database object key/id

ironfish/src/wallet/account.ts
hughy: After #3873 it takes an optional key range as well as an optional transaction

ironfish/src/rpc/routes/wallet/getNotes.ts
hughy: Since this endpoint will be paginated each response will be a 'page'. To me it made sense to use page terms for pageSize, pageToken, and nextPageToken

dguenther: I'd probably lean toward calling these `pageCursor` and `nextPageCursor` instead -- it doesn't make a huge deal, but a few APIs I glanced at tend to use the term `cursor`:
https://developer.squareup.com/docs/build-basics/common-api-patterns/pagination
https://dev.twitch.tv/docs/api/guide/#pagination
https://shopify.dev/docs/api/usage/pagination-rest

ironfish/src/rpc/routes/wallet/getNotes.test.ts
```diff
@@ -0,0 +1,97 @@
...
+      expect(responseNotes.length).toBe(1)
+
+      const expectedNote = accountNotesByHash.get(Buffer.from(responseNotes[0].noteHash, 'hex'))
+      expect(responseNotes[0]).toEqual(expectedNote)
+
+      pageToken = nextPageToken ?? undefined
+      responseNoteHashes.add(responseNotes[0].noteHash)
+    }
+
+    // ensure that all notes were returned with no duplicates
```
dguenther, line 94: Maybe add an `expect` here that `nextPageToken` is null?

hughy: A good test, because it isn't null here 😟

ygao76 (approved):

PR #3840, opened 2023-04-24
feat(simulator): misc changes from working on airdrop

holahula:
## Summary
These are a set of miscellaneous changes from working on the `airdrop` project. Since that project will never be merged into staging, these changes were pulled out.

## Testing Plan
Tests coming soon

## Documentation

Does this change require any updates to the Iron Fish Docs (ex. [the RPC API
Reference](https://ironfish.network/docs/onboarding/rpc/chain))? If yes, link a
related documentation pull request for the website.

```
[ ] Yes
```

## Breaking Change

Is this a breaking change? If yes, add notes below on why this is breaking and
what additional work is required, if any.

```
[ ] Yes
```

PR #3862, opened 2023-04-28
fix: use exec instead of spawn for complex cli commands

holahula:
## Summary
`child_process.spawn()` cannot handle stringified JSON arguments. Importing the genesis block converts the genesis account into a string which needs `child_process.exec()` to run. 

This was broken by #3840, so `executeCliCommandAsync()` was re-added and renamed to `executeCliCommandWithExec()` with a disclaimer as to when to use this one. 

## Testing Plan
Soon

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/simulator/src/simulator/simulation-node.ts
+++ b/simulator/src/simulator/simulation-node.ts
@@ -14,7 +14,8 @@ import {
   RpcTcpClient,
   YupUtils,
 } from '@ironfish/sdk'
-import { ChildProcessWithoutNullStreams, spawn } from 'child_process'
+import { ChildProcessWithoutNullStreams, exec, spawn } from 'child_process'
+import { promisify } from 'util'
 import {
   defaultOnError,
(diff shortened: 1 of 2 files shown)


jowparks (approved):

simulator/src/simulator/simulation-node.ts
```diff
@@ -583,6 +584,49 @@ export class SimulationNode {
       })
     })
   }
+
+  /**
```
jowparks, line 588: Might be useful to include a short description of how exec is different from spawn in terms of execution runtime.

hughy (approved):

simulator/src/simulator/simulation-node.ts
holahula: updated the comment to include `exec()` buffering all output and the potential for a buffer overflow

PR #3867, opened 2023-04-28
adds filtering to wallet/getNotes endpoint

hughy:
## Summary

supports filtering notes by value, assetId, assetName, memo, sender, noteHash, transactionHash, index, and spent.

implements naive filtering: iterates over all notes and includes notes that meet the filter criteria. this doesn't make use of walletDb indexes that could be helpful depending on the filter. for example, if 'spent = false', we could use 'unspentNoteHashes' to avoid iterating over all of the spent notes.

## Testing Plan

- adds unit tests

## Documentation

```
[X] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/wallet/getNotes.ts
```diff
@@ -25,6 +44,23 @@ export const GetNotesRequestSchema: yup.ObjectSchema<GetNotesRequest> = yup
     account: yup.string().trim(),
     pageSize: yup.number().min(1),
     pageToken: yup.string(),
+    filter: yup
```
ygao76, line 47: is this the same as GetNotesRequestFilter ?

ironfish/src/rpc/routes/wallet/getNotes.ts
```diff
@@ -68,3 +110,19 @@ router.register<typeof GetNotesRequestSchema, GetNotesResponse>(
     })
   },
 )
+
+function includeNote(note: RpcWalletNote, filter: GetNotesRequestFilter): boolean {
+  return (
+    (filter.value?.min === undefined || BigInt(note.value) >= BigInt(filter.value.min)) &&
+    (filter.value?.max === undefined || BigInt(note.value) <= BigInt(filter.value.max)) &&
+    (filter.assetId === undefined || note.assetId === filter.assetId) &&
+    (filter.assetName === undefined || note.assetName === filter.assetName) &&
```
dguenther, line 119: Do you think it'd be okay to leave out `assetName` as a filter unless there's a specific use-case for it? I'm concerned about people treating "$IRON" as our native asset rather than our native asset ID

hughy: Agreed, removed filtering by `assetName`

dguenther (approved):

PR #3868, opened 2023-05-01
feat(ifl 776): size method note primitives

jowparks:
## Summary
- Adding size methods for Note/NoteEncrypted primitives
- Add test to make sure these stay up to date

### Reason for addition
Our constants were becoming stale rather quickly (e.g. `NOTE_ENCRYPTED_SERIALIZED_SIZE_IN_BYTE`), so this pattern attempts to do two things. 
1. Tightly couple the size via a static method to our objects (rather than an unrelated constant)
2. Add a test that verifies this always stays up to date. It is currently possible to update the serialized size of these objects (rust side) without updating these methods. Now if these values change, the tests will break.

## Testing Plan
tests pass
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/primitives/note.ts
+++ b/ironfish/src/primitives/note.ts
@@ -43,16 +43,13 @@ export class Note {
     this._sender = reader.readBytes(PUBLIC_ADDRESS_LENGTH, true).toString('hex')
   }
 
-  static size(): number {
-    return (
-      PUBLIC_ADDRESS_LENGTH +
-      ASSET_ID_LENGTH +
-      8 + // VALUE
-      RANDOMNESS_LENGTH +
-      MEMO_LENGTH +
-      PUBLIC_ADDRESS_LENGTH
-    )
-  }
+  static size =
+    PUBLIC_ADDRESS_LENGTH +
+    ASSET_ID_LENGTH +
+    8 + // VALUE
+    RANDOMNESS_LENGTH +
+    MEMO_LENGTH +
+    PUBLIC_ADDRESS_LENGTH
 
   serialize(): Buffer {
     return this.noteSerialized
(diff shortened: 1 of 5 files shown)


NullSoldier:
This change looks good, but can we get a comment on the systemic implications here? Is this a new pattern we should be going for? When ever we start moving the code base in a new direction with a new pattern it'd be nice to comment on that in the decsription.

ironfish/src/primitives/note.ts
```diff
@@ -43,6 +43,17 @@ export class Note {
     this._sender = reader.readBytes(PUBLIC_ADDRESS_LENGTH, true).toString('hex')
   }
 
+  static size(): number {
+    return (
+      PUBLIC_ADDRESS_LENGTH +
+      ASSET_ID_LENGTH +
+      8 +
```
NullSoldier, line 50:
```suggestion
      8 + // VALUE
```

ironfish/src/primitives/noteEncrypted.ts
```diff
@@ -45,9 +41,10 @@ export class NoteEncrypted {
 
     // note encryption keys
     reader.seek(NOTE_ENCRYPTION_KEY_LENGTH)
+  }
 
-    // total serialized size: 192 (proof from transaction)
-    // + 32 + 32 + 32 + 104 + 16 + 64 + 16 = 488 bytes
+  static size(): number {
+    return 32 + 32 + 32 + ENCRYPTED_NOTE_PLAINTEXT_LENGTH + NOTE_ENCRYPTION_KEY_LENGTH
```
NullSoldier, line 47:
```suggestion
    return
        32 + // value commitment
        32 + // note commitment
        32 + // ephemeral public key
        ENCRYPTED_NOTE_PLAINTEXT_LENGTH +
        NOTE_ENCRYPTION_KEY_LENGTH
```

NullSoldier (approved):

PR #3870, opened 2023-05-01
feat(ifl-679): validate string input fields in RPC

jowparks:
## Summary
Validate the following fields:
- asset name
- asset metadata
- note memo
## Testing Plan
tests pass
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/wallet/createTransaction.test.ts
```diff
@@ -240,8 +240,8 @@ describe('Route wallet/createTransaction', () => {
       ],
       mints: [
         {
```
jowparks, line 242: The current RPC usage via CLI is using plaintext input for these params, so I think these tests aren't representative.

ironfish/src/rpc/routes/wallet/createTransaction.ts
```diff
@@ -141,10 +143,18 @@ router.register<typeof CreateTransactionRequestSchema, CreateTransactionResponse
 
           name = asset.name.toString('utf8')
           metadata = asset.metadata.toString('utf8')
```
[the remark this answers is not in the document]
jowparks: I think this was just a test setup issue. RPC usage in CLI actually just passes plaintext rather than hex encoded bytes.

ironfish/src/rpc/routes/wallet/sendTransaction.ts
```diff
@@ -64,6 +65,7 @@ router.register<typeof SendTransactionRequestSchema, SendTransactionResponse>(
   `${ApiNamespace.wallet}/sendTransaction`,
   SendTransactionRequestSchema,
   async (request, node): Promise<void> => {
+    request.data.outputs.map((o) => ValidationUtils.assertValidMemo(o.memo))
```
mat-if, line 68: Instead of introducing these custom validation functions, it would probably be easier to change line 43 to `memo: yup.string().defined().max(32)`

You can apply this to all the changes in this PR I think.

mat-if:
This should be pointing to staging not master

ironfish/src/rpc/routes/wallet/sendTransaction.ts
jowparks: Oh cool, I didn't know yup could do that. Yea this is much better!

PR #3871, opened 2023-05-01
Check network version on chaindownload

ygao76:
## Summary
Check network Id in chain:download cli. Download testnet snapshot if the network Id is 0.

## Testing Plan
local 
```
yajun@Yajuns-MBP ironfish-cli % yarn start chain:download --datadir=~/testnet/.ironfish
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc --inspect=:0 --inspect-publish-uid=http --enable-source-maps bin/run chain:download '--datadir=~/testnet/.ironfish'
root: /Users/yajun/ironfish/ironfish/ironfish-cli
See more details with DEBUG=*
Download 16.98 MB snapshot to update from block 4348 to 9700? 
At least 33.97 MB of free disk space is required to download and unzip the snapshot file.
Are you sure? (Y)es / (N)o: Y
Downloading snapshot from https://testnet.snapshots.ironfish.network/ironfish_snapshot_1682953324920.tar.gz to /Users/yajun/testnet/.ironfish/temp/ironfish_snapshot_1682953324920.tar.gz
Downloading snapshot: [████████████████████████████████████████] 100% | 16.98 MB / 16.98 MB | 0 B/s | ETA: N/A
Unzipping snapshot: [████████████████████████████████████████] 100% | 16 / 16 entries | 0.00/s | ETA: N/A
Removing existing chain data at /Users/yajun/testnet/.ironfish/databases/chain before importing snapshot... done
Moving snapshot files from /Users/yajun/testnet/.ironfish/temp/snapshot to /Users/yajun/testnet/.ironfish/databases/chain... done
Cleaning up snapshot file at /Users/yajun/testnet/.ironfish/temp/ironfish_snapshot_1682953324920.tar.gz... done
```

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```
@iron-fish/engineering

### DIFF
--- a/ironfish-cli/src/commands/chain/download.ts
+++ b/ironfish-cli/src/commands/chain/download.ts
@@ -77,8 +77,10 @@ export default class Download extends IronfishCommand {
       if (networkId === 0) {
         // testnet
         manifestUrl = Download.defaultTestnetManifestUrl
-      } else {
+      } else if (networkId === 1) {
         manifestUrl = Download.defaultMainnetManifestUrl
+      } else {
+        this.log(`Manifest url for the snapshots are not available for network ID ${networkId}`)
       }
     }
 
--- a/ironfish-cli/src/commands/chain/download.ts
+++ b/ironfish-cli/src/commands/chain/download.ts
@@ -70,7 +70,7 @@ export default class Download extends IronfishCommand {
 
     const networkId = node.internal.get('networkId')
 
-    let manifestUrl
+    let manifestUrl = ''
     if (flags.manifestUrl) {
       manifestUrl = flags.manifestUrl
     } else {
@@ -81,6 +81,7 @@ export default class Download extends IronfishCommand {
         manifestUrl = Download.defaultMainnetManifestUrl
       } else {
         this.log(`Manifest url for the snapshots are not available for network ID ${networkId}`)
+        this.exit(1)
       }
     }
 
@@ -89,7 +90,7 @@ export default class Download extends IronfishCommand {
     if (flags.path) {
       snapshotPath = this.sdk.fileSystem.resolve(flags.path)
     } else {
-      if (!manifestUrl) {
+      if (!manifestUrl || manifestUrl === '') {
         this.log(`Cannot download snapshot without manifest URL`)
         this.exit(1)
       }
--- a/ironfish-cli/src/commands/chain/download.ts
+++ b/ironfish-cli/src/commands/chain/download.ts
@@ -28,13 +28,15 @@ export default class Download extends IronfishCommand {
 
   static description = `Download and import a chain snapshot`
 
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/chain/download.ts
```diff
@@ -66,17 +66,25 @@ export default class Download extends IronfishCommand {
     const node = await this.sdk.node()
     await NodeUtils.waitForOpen(node)
 
+    const networkId = node.internal.get('networkId')
+
+    let manifestUrl = flags.manifestUrl
+    if (networkId === 0) {
+      // testnet
+      manifestUrl = `https://testnet.snapshots.ironfish.network/manifest.json`
+    }
```
hughy, line 75: this will make this URL the only one you can use if `networkId === 0`

right now this is the only I know of that you'd want to download, but this change will prevent this command from being used to download snapshots that anyone else publishes

ygao76: Yeah how about create a new flag `testnetManifestUrl`? it will be used if the networkId is 0

hughy: Could we use the value that the user passes in if they use the flag, but default to either the mainnet or testnet URLs based on networkId if they didn't use the flag?

I think it would be confusing to have two different flags

hughy:
It might also be worthwhile (in a separate PR) to add `networkId` to the snapshot manifest so that `chain:download` can warn users if they're trying to download a snapshot for a different network

ironfish-cli/src/commands/chain/download.ts
ygao76: Updated! I created a default manifest url values instead of using tthe default value from flag.

hughy (approved):

ironfish-cli/src/commands/chain/download.ts
```diff
@@ -66,17 +68,31 @@ export default class Download extends IronfishCommand {
...
+    let manifestUrl
+    if (flags.manifestUrl) {
+      manifestUrl = flags.manifestUrl
+    } else {
+      if (networkId === 0) {
+        // testnet
+        manifestUrl = Download.defaultTestnetManifestUrl
+      } else {
+        manifestUrl = Download.defaultMainnetManifestUrl
+      }
```
dguenther, line 82: Could we exit with an error if `networkId` is not 0 or 1 (Testnet or Mainnet), and `flags.manifestUrl` is not set? Something like:

```suggestion
      } else if (networkId === 1) {
        manifestUrl = Download.defaultMainnetManifestUrl
      } else {
        this.log(`Snapshots are not available for network ID ${networkId}`)
      }
```

dguenther (approved):

PR #3876, opened 2023-05-02
Add RPCs for tracking full lifecycle of peer connection

danield9tqh:
## Summary
If we want to track the status of a adding a peer manually we'll need

1. Track the peer as a PeerCandidate - this will show us whether the peer is accepting incoming connections or if the connection failed because they requested a disconnect. We can track this by adding an RPC for getting peer candidates

2. Add ability to filter peer/getPeers RPC by host and port. This will allow us to follow a single connection that we added manually. Currently the only way to get this data is to query all peers and filter that response

## Testing Plan
Unit tests + testing on local node through HTTP

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


danield9tqh:
Implementing in https://github.com/iron-fish/ironfish/pull/3885 instead

PR #3877, opened 2023-05-02
test: introduce some basic benchmarks

mat-if:
## Summary

Due to limitations of how Cargo and Rust work, we would have been limited in how robust we could make the benchmark code without doing things like including `criterion` as a non-dev dependency. While we could still gate it as an optional dependency, it just feels kind of bad to do it that way. 

By creating a separate package, we can create all the benchmark code we want without worrying about it getting included in the binaries we deliver with the node.

Eventually, we may even consider adding Typescript benchmark code here as well. Haven't really thought through that, but we have plenty of flexibility now.

Closes IFL-813
Closes IFL-815
Closes IFL-819
Closes IFL-820
Closes IFL-824

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/benchmarks/src/lib.rs
+++ b/benchmarks/src/lib.rs
@@ -1,0 +1,12 @@
+use std::time::Duration;
+
+use criterion::Criterion;
+
+pub const LONG_BENCH_DURATION: Duration = Duration::from_secs(60);
+pub const LONG_BENCH_SAMPLE_SIZE: usize = 50;
+
+pub fn slow_config() -> Criterion {
+    Criterion::default()
+        .measurement_time(LONG_BENCH_DURATION)
+        .sample_size(LONG_BENCH_SAMPLE_SIZE)
+}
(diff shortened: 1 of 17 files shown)


ironfish-rust/src/test_util.rs
```diff
@@ -7,13 +7,13 @@ use super::{
     witness::{Witness, WitnessNode},
     MerkleNoteHash,
 };
+use bls12_381::Scalar;
 use ironfish_zkp::constants::TREE_DEPTH;
 use rand::{thread_rng, Rng};
 
 /// Given a note, construct a Witness with a valid root_hash and authentication
 /// path placing that note at a random location in a Merkle tree.
-#[cfg(test)]
```
mat-if, line 15: Gating these here is kind of unnecessary if we gate the module definition in lib.rs (see above)

benchmarks/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+use std::time::Duration;
+
+use criterion::Criterion;
+
+pub const LONG_BENCH_DURATION: Duration = Duration::from_secs(60);
```
mat-if, line 5: I expect this will all change quite a bit as we flesh it out

dguenther (approved):

PR #3879, opened 2023-05-02
feat(ifl-406): combine service commands

jowparks:
## Summary
Combines fee-rates and forks service command into one. Once we are converted over I will delete the other two commands.

## Testing Plan
Tested against local host, was hitting `/telemetry` endpoint and subsequently `influx` for both:

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/service/stats.ts
+++ b/ironfish-cli/src/commands/service/stats.ts
@@ -1,0 +1,150 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { CurrencyUtils, PromiseUtils } from '@ironfish/sdk'
+import { WebApi } from '@ironfish/sdk'
+import { Flags } from '@oclif/core'
+import { IronfishCommand } from '../../command'
+import { RemoteFlags } from '../../flags'
+import { IronfishCliPKG } from '../../package'
+import { GossipForkCounter } from '../../utils/gossipForkCounter'
+
+export default class Stats extends IronfishCommand {
+  static hidden = true
+
+  static description = `Submits stats to telemetry API`
+
+  static flags = {
+    ...RemoteFlags,
+    endpoint: Flags.string({
+      char: 'e',
+      parse: (input: string) => Promise.resolve(input.trim()),
+      required: false,
+      description: 'API host to sync to',
+    }),
+    token: Flags.string({
+      char: 't',
+      parse: (input: string) => Promise.resolve(input.trim()),
+      required: false,
+      description: 'API host token to authenticate with',
+    }),
+    delay: Flags.integer({
+      required: false,
+      default: 60000,
+      description: 'Delay (in ms) to wait before uploading stats',
+    }),
+  }
+
+  async start(): Promise<void> {
+    const { flags } = await this.parse(Stats)
+
+    const apiHost = (
+      flags.endpoint ||
+      process.env.IRONFISH_API_HOST ||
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/service/stats.ts
```diff
@@ -0,0 +1,157 @@
...
+      flags.endpoint ||
+      process.env.IRONFISH_API_HOST ||
+      'https://api.ironfish.network'
+    ).trim()
+
+    const apiToken = (flags.token || process.env.IRONFISH_API_TOKEN || '').trim()
+
+    const api = new WebApi({ host: apiHost, token: apiToken })
+
+    await this.sdk.client.connect()
```
hughy, line 51: Do we need this RPC connection here since each loop connects independently?

ironfish-cli/src/commands/service/stats.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    // metric loops
+    void this.forks(api, flags.delay)
+    void this.feeRates(api, flags.delay)
+
+    // hold program open so that metrics can be sent
+    // eslint-disable-next-line no-constant-condition
+    while (true) {
+      await PromiseUtils.sleep(1000)
+    }
```
hughy, line 61: I think an alternative to keep the process running might be to `await` the second metric loop

ironfish-cli/src/commands/service/stats.ts
jowparks: If I remove this line I get the error:
```
fish service:stats -e http://localhost:8003 -t test --datadir ~/.ironfishairdrop --delay 5000                                                                                                                                              

yarn run v1.22.19
$ yarn build && yarn start:js service:stats -e http://localhost:8003 -t test --datadir /Users/joe/.ironfishairdrop --delay 5000
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc --inspect=:0 --inspect-publish-uid=http --enable-source-maps bin/run service:stats -e http://localhost:8003 -t test --datadir /Users/joe/.ironfishairdrop --delay 5000
/Users/joe/repos/ironfish/ironfish/src/assert.ts:26
      throw new Error(message || `Expected value not to be null`)
            ^


Error: Connect first using connect()
    at Function.isNotNull (/Users/joe/repos/ironfish/ironfish/src/assert.ts:26:13)
    at RpcIpcClient.request (/Users/joe/repos/ironfish/ironfish/src/rpc/clients/socketClient.ts:141:12)
    at Object.getConsensusParameters (/Users/joe/repos/ironfish/ironfish/src/rpc/clients/client.ts:628:19)
    at Stats.forks (/Users/joe/repos/ironfish/ironfish-cli/src/commands/service/stats.ts:58:67)
    at Stats.start (/Users/joe/repos/ironfish/ironfish-cli/src/commands/service/stats.ts:52:15)
    at Stats.run (/Users/joe/repos/ironfish/ironfish-cli/src/command.ts:80:7)
    at async Stats._run (/Users/joe/repos/ironfish/ironfish-cli/node_modules/@oclif/core/lib/command.js:80:22)
    at async Config.runCommand (/Users/joe/repos/ironfish/ironfish-cli/node_modules/@oclif/core/lib/config/config.js:272:25)
    at async Object.run (/Users/joe/repos/ironfish/ironfish-cli/node_modules/@oclif/core/lib/main.js:74:5)

Node.js v18.12.1
```

hughy: Ah I think this is because there's an RPC call in `forks` on line 59 before that function tries to connect. We might be able to move this connect to happen there

hughy (approved):

PR #3882, opened 2023-05-03
feat(ifl-405): update metrics flush interval to one hour

jowparks:
## Summary
Collects metrics every 5 minutes, but flushes only once per hour. This should reduce API burden but keep our metrics granular.
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/telemetry/telemetry.ts
+++ b/ironfish/src/telemetry/telemetry.ts
@@ -19,7 +19,7 @@ import { Metric } from './interfaces/metric'
 import { Tag } from './interfaces/tag'
 
 export class Telemetry {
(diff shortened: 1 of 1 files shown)


hughy (approved):

jowparks:
> It's not clear to me that this would achieve the goal of reducing API burden. We're submitting the exact same amount of data, just on a larger interval, meaning larger payloads. There's a possibility that could make performance worse, even

I don't think payload size is a huge issue, I think frequency of request is more likely the issue. I see this as similar to hitting an API 1000 times for 1000 records vs 1 time for 1000 records. There is overhead to each request that we save on by batching requests. 

Am I incorrect in my reasoning? It is certainly possible.

PR #3883, opened 2023-05-03
makes notes from wallet RPCs uniform

hughy:
## Summary

defines a single RpcWalletNote type and schema, reuses across three wallet endpoints that return note data.

adds owner, index, nullifier, noteHash, and transactionHash fields to notes returned in transaction endpoints for consistency.

marks 'hash' field as deprecated. transaction endpoints used 'hash' and notes endpoint uses 'noteHash'.

marks 'isOwner' field as deprecated. this isn't very useful if the note contains the owner address. if clients need to compare the owner address to the public address for an account this is better done on the client side than the rpc.

## Testing Plan

existing unit tests

## Documentation

```
[X] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/transaction/index.ts
+++ b/ironfish-cli/src/commands/wallet/transaction/index.ts
@@ -80,10 +80,6 @@ export class TransactionCommand extends IronfishCommand {
         memo: {
           header: 'Memo',
         },
-        isOwner: {
-          header: 'Owner',
-          get: (note) => (note.isOwner ? `✔` : `x`),
-        },
         owner: {
           header: 'Owner Address',
           get: (note) => note.owner,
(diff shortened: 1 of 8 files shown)


ironfish-cli/src/commands/wallet/transaction/index.ts
```diff
@@ -80,10 +80,6 @@ export class TransactionCommand extends IronfishCommand {
         memo: {
           header: 'Memo',
         },
```
jowparks, line 82: I am guessing this is related to the deprecation? Just verifying.

ygao76 (approved):

jowparks (approved):
Very nice. I see there is no test coverage on `src/rpc/routes/wallet/getAccountNotesStream.ts` might be good to add a test there to verify backward compatibility.

ironfish-cli/src/commands/wallet/transaction/index.ts
hughy: Yeah, it's not totally necessary to remove this here, but I thought we should remove our own usage of the field if we're deprecating it 😁

hughy:
> Very nice. I see there is no test coverage on `src/rpc/routes/wallet/getAccountNotesStream.ts` might be good to add a test there to verify backward compatibility.

Agreed; I think this is a good time to add tests for that endpoint

PR #3885, opened 2023-05-03
Create peer/add RPC endpoint and CLI command

danield9tqh:
## Summary
Create a command to add a new peer to the node based on IP + port

## Testing Plan
Tested connecting to nodes using local node + unit tests

## Documentation

```
[X] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/peers/add.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (!connected) {
+      this.log('Could not connect to node')
+      this.exit(0)
+    }
+
+    if (!url.hostname) {
+      this.error(`Could not parse the given url`)
+    }
+
+    const reqeust = {
```
dguenther, line 46: Typo: `request`

ironfish-cli/src/commands/peers/add.ts
```diff
@@ -0,0 +1,61 @@
...
+import { DEFAULT_WEBSOCKET_PORT, parseUrl } from '@ironfish/sdk'
+import { IronfishCommand } from '../../command'
+import { RemoteFlags } from '../../flags'
+
+export class AddCommand extends IronfishCommand {
+  static description = `Add a peer to the list of candidates`
+
+  static args = [
+    {
+      name: 'url',
```
dguenther, line 13: It's not too important, but I'd lean toward calling this a `socket`, `address`, or `endpoint` instead, since a URL includes a path + protocol

ironfish/src/rpc/routes/peer/addPeerCandidate.ts
```diff
@@ -0,0 +1,51 @@
...
+    const peerManager = node.peerNetwork.peerManager
+    const { host, port, whitelist } = request.data
+
+    const peer = peerManager.getOrCreatePeer(null)
+    peer.setWebSocketAddress(host, port || DEFAULT_WEBSOCKET_PORT)
+    peer.isWhitelisted = !!whitelist
+
+    peerManager.peerCandidates.addFromPeer(peer)
+
+    const added = peerManager.connectToWebSocket(peer)
```
dguenther, line 47: This is pretty much the same as peerManager.connectToWebSocketAddress(), so I'd rather use that instead since the individual operations here are pretty important. Looks like we just need to pass through an option to force connecting even if we're above targetPeers?

ironfish/src/rpc/routes/peer/addPeerCandidate.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const AddPeerCandidateResponseSchema: yup.ObjectSchema<AddPeerCandidateResponse> = yup
+  .object({
+    added: yup.boolean().defined(),
+    error: yup.string().optional(),
+  })
+  .defined()
+
+router.register<typeof AddPeerCandidateRequestSchema, AddPeerCandidateResponse>(
+  `${ApiNamespace.peer}/addCandidate`,
```
dguenther, line 35: This PR should have the `Docs` checkmark checked on the template, since we'll need to add this route to the RPC Docs

ironfish/src/rpc/routes/peer/addPeerCandidate.ts
```diff
@@ -0,0 +1,51 @@
...
+  .defined()
+
+export const AddPeerCandidateResponseSchema: yup.ObjectSchema<AddPeerCandidateResponse> = yup
+  .object({
+    added: yup.boolean().defined(),
+    error: yup.string().optional(),
+  })
+  .defined()
+
+router.register<typeof AddPeerCandidateRequestSchema, AddPeerCandidateResponse>(
```
dguenther, line 34: Please match the filename to the route name -- the route is `peer/addCandidate`, but the file is `peer/addPeerCandidate`

ironfish/src/rpc/clients/client.ts
```diff
@@ -436,6 +438,15 @@ export abstract class RpcClient {
         },
       )
     },
+
+    addCandidate: (
```
dguenther, line 442: It'd be nice to alphabetize this based on the route name

ironfish/src/network/peers/peerManager.ts
```diff
@@ -157,31 +157,26 @@ export class PeerManager {
...
   }
 
   /**
-   * Connect to a peer using WebSockets
+   * Connect to a peer using WebSockets. This function forces a connection
+   * and doesn't check `canConnectToWebSocket`. Unless the connection needs to
+   * be forced `canConnectToWebSocket` should probably be called first
    * */
   connectToWebSocket(peer: Peer): boolean {
-    if (!this.canConnectToWebSocket(peer)) {
```
dguenther, line 181: I don't think we should move this out -- we only really want to force a connection in this one specific case we're adding, but this pattern causes you to have to remember to call `canConnectToWebSockets` in the other typical cases.

Plus, I think we might want more granularity around which checks we disable -- for example, it might make sense to not force the connection if the peer requested a disconnect, or if the peer already has an existing WebSocket connection.

ironfish/src/rpc/routes/peer/addPeerCandidate.test.ts
```diff
@@ -0,0 +1,22 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { createRouteTest } from '../../../testUtilities/routeTest'
+
+describe('Route peer/addPeerCandidate', () => {
```
dguenther, line 6:
```suggestion
describe('Route peer/addCandidate', () => {
```

ironfish/src/rpc/routes/peer/addPeerCandidate.test.ts
```diff
@@ -0,0 +1,22 @@
...
+  const routeTest = createRouteTest()
+
+  it('should add a peer with a correct address and port', async () => {
+    const reqeust = { host: 'testhost', port: 9037 }
```
dguenther, line 10:
```suggestion
    const request = { host: 'testhost', port: 9037 }
```

dguenther (changes requested):

ironfish-cli/src/commands/peers/add.ts
danield9tqh: alright yea I agree

ironfish/src/rpc/routes/peer/addPeerCandidate.ts
danield9tqh: Ok agreed to pass down a parameter for forceConnect

ironfish/src/network/peers/peerManager.ts
danield9tqh: Agreed to pass down a parameter to `forceConnect`

ironfish/src/rpc/routes/peer/addCandidate.ts
```diff
@@ -0,0 +1,49 @@
...
+  `${ApiNamespace.peer}/addCandidate`,
+  AddPeerCandidateRequestSchema,
+  (request, node): void => {
+    const peerManager = node.peerNetwork.peerManager
+    const { host, port, whitelist } = request.data
+
+    const peer = peerManager.connectToWebSocketAddress({
+      host,
+      port: port || DEFAULT_WEBSOCKET_PORT,
+      whitelist: !!whitelist,
```
dguenther, line 44: Should this pass `forceConnect`?

danield9tqh: yep, updated

ironfish-cli/src/commands/peers/add.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  static args = [
+    {
+      name: 'address',
+      parse: (
+        input: string,
+      ): Promise<{ protocol: string | null; hostname: string | null; port: number | null }> =>
+        Promise.resolve(parseUrl(input.trim())),
+      required: true,
+      description: `The url of the peer to connect to in the form {host}:{port}`,
```
dguenther, line 19:
```suggestion
      description: `The address of the peer to connect to in the form {host}:{port}`,
```

ironfish/src/network/peers/peerManager.test.ts
```diff
@@ -516,7 +532,7 @@ describe('PeerManager', () => {
       )
     })
 
-    it('Does not create a connection if Peer has disconnectUntil set', () => {
+    it('canConnectToWebSocket returns false if Peer has disconnectUntil set', () => {
```
dguenther, line 535: Can you revert changes to this test? I think it should be fine again to call `connectToWebSocket` instead of testing `canConnectToWebSocket`

ironfish/src/rpc/routes/peer/addPeer.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export type AddPeerRequest = {
+  host: string
+  port?: number
+  whitelist?: boolean
+}
+
+export type AddPeerResponse = {
+  added: boolean
+  error?: string
```
rohanjadvani, line 16: Is this used? I guess it's nice to have in the future, but seems like we can just add this field when we implement it.

danield9tqh: Yea agreed, I'll remove

PR #3893, opened 2023-05-05
adds test to reproduce invalid spends

hughy:
## Summary

creates a pending transaction with invalid spends.

creates a transaction with the chain at block A3. the spends in the transaction use witnesses created with the tree of notes as of block A3.

reorgs the chain.

spends created at block A3 are no longer valid.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/wallet.test.ts
+++ b/ironfish/src/wallet/wallet.test.ts
@@ -1040,6 +1040,53 @@ describe('Accounts', () => {
 
       expect(spentNoteHashes.has(notes[0])).toBe(true)
     })
+
+    it('should create transactions with spends valid after a reorg', async () => {
+      // create two chains to create a fork
+      const { node: nodeA } = await nodeTest.createSetup()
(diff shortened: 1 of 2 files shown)


ironfish/src/wallet/wallet.test.ts
```diff
@@ -1040,6 +1040,53 @@ describe('Accounts', () => {
 
       expect(spentNoteHashes.has(notes[0])).toBe(true)
     })
+
+    it('should create transactions with spends valid after a reorg', async () => {
```
dguenther, line 1044: Should this be `invalid`? Or are you going to use this to test changes to transaction creation?

hughy: Yeah, it is invalid now, but I'm hoping that with changes it will become valid

hughy:
Closing this PR. We can reintroduce this test when we have changes to support confirmation-range witnesses

PR #3896, opened 2023-05-07
adds support for creating witnesses at past sizes

hughy:
## Summary

modifies 'witness' to accept a size parameter to construct a witness at a past tree size.

if constructing a witness at a past size we need to recompute the hashes of all nodes in the tree that included leaves that were not in the tree at that size.

defines 'pastRightSiblingHashes' to recompute hashes along the path from the rightmost leaf of a tree at a past size to the root. returns a mapping from node index to the hash at that index for the past root.

adds merkletree unit test to verify that we can correctly recompute witnesses at past tree sizes.

## Testing Plan

adds unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


dguenther (approved):
Played around with the test a bit, and I'm satisfied this works as expected.

PR #3901, opened 2023-05-08
fix(ironfish): add additional transaction validation

mat-if:
## Summary

This is a low-cost check for transactions now that we have a transaction index for the blockchain. This would catch transaction replays in scenarios where spends might not be present. For example, a mint when the minimum fee is 0 is technically replayable. By adding this check, we ensure that even if the fee gets set to 0, we are not vulnerable to such bugs.

Closes IFL-851

## Testing Plan

Unit tests on blockchain and verifier

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/consensus/verifier.ts
```diff
@@ -519,6 +531,26 @@ export class Verifier {
...
+   * transactions that do not contain spends could be replayable in some
+   * scenarios.
+   */
+  verifyUnseenTransaction(
+    transaction: Transaction,
+    tx?: IDatabaseTransaction,
+  ): Promise<VerificationResult> {
+    return this.chain.db.withTransaction(tx, async (tx) => {
+      if (await this.chain.transactionHashToBlockHash.has(transaction.hash(), tx)) {
+        return { valid: false, reason: VerificationResultReason.DOUBLE_SPEND }
```
rohanjadvani, line 548: nit: I think with a default min fee of 1 currently in our consensus, we'll implicitly always have a spend in a transaction so the double spend reason would hold true. However, all our current double spend error messages arise from checking associated nullifiers. How do you feel about using something like `VerificationResultReason.EXISTING_TRANSACTION`? It feels cleaner to me, but I don't feel that strongly about it.

rohanjadvani (approved):

ironfish/src/consensus/verifier.ts
mat-if: Good call - I'll add a new error type.

PR #3906, opened 2023-05-09
adds rpc to follow mempool transaction stream

hughy:
## Summary

creates a mempool rpc endpoint to stream serialized transactions as they are added to the mempool.

adds an event to the mempool that emits each transaction successfully added to the mempool.

supports syncing mempool transaction to the api.

## Testing Plan

- adds unit test
- manual testing:
  - simple script to invoke the rpc endpoint using the sdk and log transactions to the console

## Documentation

```
[X] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/memPool/memPool.ts
+++ b/ironfish/src/memPool/memPool.ts
@@ -6,6 +6,7 @@ import { BufferMap } from 'buffer-map'
 import { Assert } from '../assert'
 import { Blockchain } from '../blockchain'
 import { Consensus, isExpiredSequence } from '../consensus'
+import { Event } from '../event'
 import { createRootLogger, Logger } from '../logger'
 import { MetricsMonitor } from '../metrics'
 import { getTransactionSize } from '../network/utils/serializers'
@@ -59,6 +60,8 @@ export class MemPool {
   private readonly metrics: MetricsMonitor
 
   readonly feeEstimator: FeeEstimator
+
+  onAdd = new Event<[transaction: Transaction]>()
 
   constructor(options: {
     chain: Blockchain
@@ -307,6 +310,8 @@ export class MemPool {
       const evicted = this.evictTransactions()
       this.metrics.memPoolEvictions.value += evicted.length
     }
+
+    this.onAdd.emit(transaction)
 
     return true
   }
--- a/ironfish/src/rpc/clients/client.ts
+++ b/ironfish/src/rpc/clients/client.ts
@@ -29,6 +29,8 @@ import {
   ExportChainStreamResponse,
   FollowChainStreamRequest,
   FollowChainStreamResponse,
+  FollowMempoolTransactionStreamRequest,
+  FollowMempoolTransactionStreamResponse,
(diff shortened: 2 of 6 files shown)


ironfish/src/memPool/memPool.ts
```diff
@@ -308,6 +311,8 @@ export class MemPool {
       this.metrics.memPoolEvictions.value += evicted.length
     }
```
jowparks, line 313: Will this emit messages on a valid transaction? Do we want to emit evicted/rejected transactions too?

ironfish/src/rpc/clients/client.ts
```diff
@@ -365,6 +367,15 @@ export abstract class RpcClient {
       )
     },
```
jowparks, line 369: Not in love with this "follow" naming pattern, but I see it predates this PR, so I am fine leaving or changing it.

ironfish/src/rpc/routes/mempool/followTransactionStream.ts
```diff
@@ -0,0 +1,44 @@
...
+>(
+  `${ApiNamespace.mempool}/followTransactionStream`,
+  FollowMempoolTransactionStreamRequestSchema,
+  (request, node): void => {
+    const onAdd = (transaction: Transaction) => {
+      request.stream({
+        serializedTransaction: transaction.serialize().toString('hex'),
+      })
+    }
+
```
jowparks, line 37: Dan had recommended `this.peerNetwork.onTransactionAccepted` event which already exists, but what you are doing seems fine too.

ironfish/src/memPool/memPool.ts
hughy: This will only emit for a valid transaction that is successfully added to the mempool.

We could emit for evictions and rejections, but for now I don't know what we'd do with those events

ironfish/src/rpc/clients/client.ts
hughy: I used `follow` here to distinguish from the streaming `getTransactions` endpoint that's already here for the mempool.

I'm open to renaming!

ironfish/src/rpc/routes/mempool/followTransactionStream.ts
hughy: That event will work, too! The only reason I can think of not to use `peerNetwork.onTransactionAccepted` is that it seems a little odd to follow an event emitted from the peer network as a way to sync transactions from the mempool

ironfish/src/rpc/clients/client.ts
jowparks: I like `getTransactonsStream`, we use that naming in `/chain`

jowparks (approved):
Feel free to rename or not, changes look good.

hughy:
Closing this PR. Talked with @NullSoldier and @danield9tqh about moving this logic to the `events` namespace and syncing every transaction that is gossiped to the node

PR #3915, opened 2023-05-11
perf: improve rust hex (de-)serialization

mat-if:
## Summary

While these were not very expensive, they are fairly commonly called and improving the performance here was a nice, quick win. On top of that, we removed the heap allocation for `hex_to_bytes` which should improve performance even further on lower-end machines.

I moved `serializing.rs` into a proper module since it had a sub-module in it, which I realize now was a bad idea for the diff here.

I lost my benchmarks, but `hex_to_bytes` was around 10x faster, and `bytes_to_hex` was like 15x faster

Closes IFL-952

## Testing Plan

Unit tests, running node

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-rust/src/serializing/aead.rs
```diff
@@ -0,0 +1,83 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
mat-if, line 1: This file is unchanged, it was just split out into it's own file, since before it was nested inside serializing.rs

ironfish-rust/src/serializing/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+/// all have functions for serializing, but their interface
+/// can be a bit clunky if you're just working with bytearrays.
+use ff::PrimeField;
+use group::GroupEncoding;
+
+use std::io;
+
+const HEX_CHARS: &[u8; 16] = b"0123456789abcdef";
+
+pub(crate) fn read_scalar<F: PrimeField, R: io::Read>(mut reader: R) -> Result<F, IronfishError> {
```
mat-if, line 21: `read_scalar` and `read_point` are unchanged, git just didn't catch that this was name changed from `serializing.rs`.

ironfish-rust/src/serializing/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+    let mut bytes = [0; SIZE];
+
+    let hex_iter = hex.as_bytes().chunks_exact(2);
+
+    for (i, hex) in hex_iter.enumerate() {
+        bytes[i] = hex_to_u8(hex[0])? << 4 | hex_to_u8(hex[1])?;
+    }
+
+    Ok(bytes)
+}
```
mat-if, line 62: `hex_to_bytes` now takes an array, relying on a const generic to determine the size. Removes any heap allocations via `Vec`s.

At first glance, this looks like it would be more painful to use, however, we were ALWAYS using it with an array already, but had to do a lot of extra work moving back and forth between arrays and vecs. You'll notice the call-sites are all a LOT cleaner.

ironfish-rust/src/serializing/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+pub fn bytes_to_hex(bytes: &[u8]) -> String {
+    let mut hex: Vec<u8> = vec![0; bytes.len() * 2];
+
+    for (i, b) in bytes.iter().enumerate() {
+        hex[i * 2] = HEX_CHARS[(b >> 4) as usize];
+        hex[i * 2 + 1] = HEX_CHARS[(b & 0x0f) as usize];
+    }
+
+    unsafe { String::from_utf8_unchecked(hex) }
+}
```
mat-if, line 45: `bytes_to_hex` still uses a `Vec`, due to Rust not providing any useful ways to go between arrays and Strings, which is kind of expected. We could do some mem transmuting ourselves, but no real need, this is _plenty_ fast.

You'll note that there is an `unsafe` block here to call `from_utf8_unchecked` instead of `from_utf8`. It is easy to see that it is impossible to have invalid utf8 since we only parse 4 bits at a time into an array we defined: `HEX_CHARS`.  If someone tried to pass "bad" bytes, they would still need to valid hex.

dguenther (approved):
nice, this looks good to me!

PR #3917, opened 2023-05-12
feat(ifl-934): execution context for telemetry segmentation

jowparks:
## Summary
Adds optional SDK init param `executionContext` which is propagated to `IronfishNode` and to `Telemetry`. This is defaultly written as a tag to all telemetry data so we can resolve metrics from `node-app` vs `CLI` vs `SDK`.
## Testing Plan
test passes
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/benchmarks/benches/transaction.rs
+++ b/benchmarks/benches/transaction.rs
@@ -1,69 +1,159 @@
-use benchmarks::slow_config;
-use criterion::{criterion_group, criterion_main, Criterion};
+use benchmarks::{slow_config, very_slow_config};
+use criterion::{criterion_group, criterion_main, BatchSize, Criterion};
 use ironfish_rust::{
     assets::{asset::Asset, asset_identifier::NATIVE_ASSET},
     test_util::make_fake_witness,
-    Note, ProposedTransaction, SaplingKey,
+    transaction::batch_verify_transactions,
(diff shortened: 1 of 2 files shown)


ironfish/src/node.ts
```diff
@@ -297,6 +304,7 @@ export class IronfishNode {
       privateIdentity,
       hostsStore,
       networkId: networkDefinition.id,
```
jowparks, line 306: We could put a default in the node, but this seems better so that we can know if people are accessing node via different mechanism than SDK.

dguenther (approved):

ironfish-cli/src/command.ts
```diff
@@ -181,6 +181,7 @@ export abstract class IronfishCommand extends Command {
       configName: typeof configFlag === 'string' ? configFlag : undefined,
       dataDir: typeof dataDirFlag === 'string' ? dataDirFlag : undefined,
       logger: this.logger,
+      executionContext: 'CLI',
```
NullSoldier, line 184: We already have this, it exists here. https://github.com/iron-fish/ironfish/blob/master/ironfish/src/platform.ts#L33

jowparks: TY! nice call

PR #3918, opened 2023-05-12
creates spend witnesses using confirmation range

hughy:
## Summary

supports creating spends with a confirmation range. if a spend is created with the chain head at some block A and the chain reorgs to block B, then the spends created at block A will be invalid because the commitment won't match the root hash of the notes tree at that note size.

using the note tree size from a confirmed block instead of the current chain head will reduce the likelihood that a reorg will render a spend invalid.

## Testing Plan

- updates unit tests
- manual testing: sent this transaction https://testnet.explorer.ironfish.network/transaction/018677441cd672cb30e687007c031fa65afceb1ea6a96e85cc20ef2bb05229f5

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


dguenther (approved):
Nice! This looks great!

PR #3920, opened 2023-05-13
adds confirmations to chain/getNoteWitness

hughy:
## Summary

the merkle tree now supports creating witnesses at past tree sizes. this allows us to create witnesses using a confirmation range.

adds a confirmations request parameter to the chain/getNoteWitness rpc. use confirmations to determine the confirmed tree size before creating a witness for the note at that tree size.

Closes IFL-880, IFL-881

## Testing Plan

- adds unit test

## Documentation

```
[X] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/chain/getNoteWitness.ts
+++ b/ironfish/src/rpc/routes/chain/getNoteWitness.ts
@@ -65,7 +65,9 @@ router.register<typeof GetNoteWitnessRequestSchema, GetNoteWitnessResponse>(
     const witness = await chain.notes.witness(request.data.index, maxConfirmedHeader.noteSize)
 
     if (witness === null) {
-      throw new ValidationError(`No confirmed notes exist with index ${request.data.index}`)
+      throw new ValidationError(
+        `No confirmed notes exist with index ${request.data.index} in tree of size ${maxConfirmedHeader.noteSize}`,
+      )
     }
 
     const authPath = witness.authenticationPath.map((step) => {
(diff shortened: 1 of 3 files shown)


ironfish/src/rpc/routes/chain/getNoteWitness.ts
```diff
@@ -46,10 +50,22 @@ router.register<typeof GetNoteWitnessRequestSchema, GetNoteWitnessResponse>(
...
+    const maxConfirmedHeader = await chain.getHeaderAtSequence(maxConfirmedSequence)
+
+    Assert.isNotNull(maxConfirmedHeader)
+    Assert.isNotNull(maxConfirmedHeader?.noteSize)
+
+    const witness = await chain.notes.witness(request.data.index, maxConfirmedHeader.noteSize)
 
     if (witness === null) {
-      throw new ValidationError(`No notes exist with index ${request.data.index}`)
+      throw new ValidationError(`No confirmed notes exist with index ${request.data.index}`)
```
dguenther, line 68: I wonder if it'd be worth logging the tree size here as well? So if a note is past the confirmation range, it would somehow indicate the reason why witness creation failed

dguenther (approved):

ironfish/src/rpc/routes/chain/getNoteWitness.ts
hughy: I think that's a good idea; better to include more information if it's available!

PR #3923, opened 2023-05-19
checks if submitted block is heavier when chain changed

hughy:
## Summary

when a miner submits a block to the mining manager the manager discards the block if the chain has changed. that is, if the submitted block no longer connects to the chain head.

however, if the submitted block is heavier than the curremt chain head then the manager should add this block to the chain.

- removes unnecessary check that chain head exists
- does not discard submitted block if it is heavier than the current chain head

## Testing Plan

- adds unit test

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/mining/manager.ts
+++ b/ironfish/src/mining/manager.ts
@@ -17,6 +17,7 @@ import {
 } from '../network/utils/serializers'
 import { IronfishNode } from '../node'
 import { Block } from '../primitives/block'
+import { isBlockHeavier } from '../primitives/blockheader'
 import { Transaction } from '../primitives/transaction'
 import { BlockTemplateSerde, SerializedBlockTemplate } from '../serde'
 import { BenchUtils } from '../utils/bench'
@@ -184,8 +185,8 @@ export class MiningManager {
 
     const blockDisplay = `${block.header.hash.toString('hex')} (${block.header.sequence})`
     if (
-      !this.node.chain.head ||
-      !block.header.previousBlockHash.equals(this.node.chain.head.hash)
+      !block.header.previousBlockHash.equals(this.node.chain.head.hash) &&
+      !isBlockHeavier(block.header, this.node.chain.head)
     ) {
       this.node.logger.info(
         `Discarding mined block ${blockDisplay} that no longer attaches to heaviest head`,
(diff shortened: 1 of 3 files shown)


ironfish/src/mining/manager.ts
```diff
@@ -184,8 +185,8 @@ export class MiningManager {
 
     const blockDisplay = `${block.header.hash.toString('hex')} (${block.header.sequence})`
     if (
-      !this.node.chain.head ||
-      !block.header.previousBlockHash.equals(this.node.chain.head.hash)
+      !block.header.previousBlockHash.equals(this.node.chain.head.hash) &&
+      !isBlockHeavier(block.header, this.node.chain.head)
```
dguenther, line 189: Two unrelated comments:
1. Since work should be strictly increasing, I think we could even skip the previousBlockHash equality check and just toss out all blocks that aren't heavier than the chain head.
2. This change will let through blocks that are heavier, but don't attach to the chain head. I don't think this should be common -- seems like it should only happen in edge cases like rolling back a multi-block fork. Regardless, I don't think this should be an issue, since `addBlock` will drop the block if it doesn't connect anyway.

dguenther (approved):
Great test addition!

ironfish/src/mining/manager.ts
hughy: I think we may be able to remove this check and the `CHAIN_CHANGED` case entirely.

We already check `isBlockHeavier` in the `addBlock` logic. The only benefit to returning early here is to avoid the operations for verifying the block and adding it to a fork.

hughy: I think it may be better to leave `CHAIN_CHANGED` in for now.

We still need to make a change here to compute the work for the submitted block from its previous block or `isBlockHeavier` will always be false

dguenther (approved):
This makes sense to me. 👍 Having to manually set the work is a bit precarious, but that's not in the scope of this change.

A long time ago I was working on a branch that split out blocks into two types: those with work + noteSize and those without, but I don't think we managed to agree that it was a concrete improvement.

Maybe it'd be useful to add asserts to the isHeavier check, for example, that make sure work is set? Not really necessary in this PR though

PR #3925, opened 2023-05-22
adds rpc to follow transaction gossip stream

hughy:
## Summary

creates an event rpc endpoint to stream serialized transactions as the peer network receives them in gossip messages.

adds an event to the peer network that emits each transaction that the node receives.

emits all transaction gossip without verifying  or waiting for the mempool to accept the transaction.

supports syncing pending transactions to the api.

## Testing Plan

- adds unit test

## Documentation

```
[X] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/network/peerNetwork.ts
```diff
@@ -1279,6 +1280,8 @@ export class PeerNetwork {
     const received = new Date()
     const hash = transaction.hash()
 
+    this.onTransactionGossipReceived.emit(transaction)
+
```
dguenther, line 1284: The block gossip equivalent gets called further down after we've marked the block as received and filtered for duplicates, should we align them one way or the other?

hughy: Yes I think we should move this down to after filtering for duplicates. That should remove the need for the LRU in the RPC endpoint, too

dguenther (approved):
This seems like it'll follow incoming transactions 👍

PR #3926, opened 2023-05-22
updates sync service to submit transaction gossip to api

hughy:
## Summary

moves 'syncBlocks' to a separate method

defines 'syncTransactionGossip' method to invoke the 'event/onTransactionGossip' RPC and submit transactions to the API

updates 'sync' to run both of these methods in coroutines

updates 'webApi' with a 'transactions' method. matches transaction serialization with serialization from 'followChainStream' for transactions submitted to API with blocks.

## Testing Plan

manual testing: ran syncer locally and synced transaction to testnet api: 5c1fa0c851957d46585addf64567fb6ab93f980d23d0d698a6ab49cfc2eb7e1b

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/service/sync.ts
+++ b/ironfish-cli/src/commands/service/sync.ts
@@ -1,7 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { FollowChainStreamResponse, Meter, TimeUtils, WebApi } from '@ironfish/sdk'
+import { FollowChainStreamResponse, Meter, TimeUtils, Transaction, WebApi } from '@ironfish/sdk'
 import { Flags } from '@oclif/core'
 import { IronfishCommand } from '../../command'
 import { RemoteFlags } from '../../flags'
@@ -12,7 +12,7 @@ export default class Sync extends IronfishCommand {
   static hidden = true
 
   static description = `
-    Upload blocks to an HTTP API using IronfishApi
+    Upload blocks and mempool transactions to an HTTP API using IronfishApi
   `
 
   static flags = {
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/commands/service/sync.ts
```diff
@@ -67,13 +67,19 @@ export default class Sync extends IronfishCommand {
...
 
-    let head = args.head as string | null
+    const head = args.head as string | null
+
+    void this.syncTransactionGossip(api, flags.maxUpload)
+    await this.syncBlocks(api, head, flags.maxUpload)
+  }
+
+  async syncBlocks(api: WebApi, head: string | null, maxUpload: number): Promise<void> {
+    const client = await this.sdk.connectRpc()
```
danield9tqh, line 79: Not sure here but can/should we use one connection for both of these calls?

ironfish-cli/src/commands/service/sync.ts
```diff
@@ -126,4 +132,27 @@ export default class Sync extends IronfishCommand {
...
+  async syncTransactionGossip(api: WebApi, maxUpload: number): Promise<void> {
+    const client = await this.sdk.connectRpc()
+
+    const response = client.event.onTransactionGossipStream({})
+
+    const buffer = new Array<Transaction>()
+
+    async function commit(): Promise<void> {
+      await api.transactions(buffer)
+      buffer.length = 0
```
danield9tqh, line 145: does this actually clear the buffer? Never seen this before

danield9tqh:
Looks mostly good to me. I think we can test this on the testnet first for integration testing

ironfish-cli/src/commands/service/sync.ts
hughy: yeah this looks a bit odd 🤔 

copied from the existing syncer code, but i'll take a closer look

hughy: This will clear the array without replacing it: https://stackoverflow.com/a/1232046

ironfish-cli/src/commands/service/sync.ts
hughy: updated to reuse the client

danield9tqh (approved):

PR #3928, opened 2023-05-24
Add chain database size to telemetry

ygao76:
## Summary
Create size() function for level DB
Create a new chainDatabaseSize metric

## Testing Plan
unit test
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```
@rohanjadvani @NullSoldier @dguenther

### DIFF
--- a/ironfish/src/blockchain/blockchain.test.ts
+++ b/ironfish/src/blockchain/blockchain.test.ts
@@ -46,7 +46,7 @@ describe('Blockchain', () => {
 
   it('add blocks with forks', async () => {
     const { chain } = nodeTest
-
+    const size = await chain.db.size()
     // G -> A1 -> A2
     //         -> B2 -> B3
 
@@ -103,6 +103,8 @@ describe('Blockchain', () => {
     expect((await chain.getHashAtSequence(2))?.equals(headerA1.hash)).toBe(true)
     expect((await chain.getHashAtSequence(3))?.equals(headerB2.hash)).toBe(true)
     expect((await chain.getHashAtSequence(4))?.equals(headerB3.hash)).toBe(true)
+
+    expect(await chain.db.size()).toBeGreaterThanOrEqual(size)
   })
 
   it('iterate', async () => {
--- a/ironfish/src/blockchain/blockchain.test.ts
+++ b/ironfish/src/blockchain/blockchain.test.ts
@@ -46,7 +46,6 @@ describe('Blockchain', () => {
 
   it('add blocks with forks', async () => {
     const { chain } = nodeTest
-    const size = await chain.db.size()
     // G -> A1 -> A2
     //         -> B2 -> B3
 
@@ -104,7 +103,7 @@ describe('Blockchain', () => {
     expect((await chain.getHashAtSequence(3))?.equals(headerB2.hash)).toBe(true)
     expect((await chain.getHashAtSequence(4))?.equals(headerB3.hash)).toBe(true)
 
-    expect(await chain.db.size()).toBeGreaterThanOrEqual(size)
+    expect(await chain.db.size()).toBeGreaterThan(0)
   })
 
   it('iterate', async () => {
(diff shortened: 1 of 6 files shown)


ironfish/src/blockchain/blockchain.ts
```diff
@@ -406,6 +406,8 @@ export class Blockchain {
 
         this.resolveOrphans(block)
 
+        this.metrics.chainDatabaseSize.value = await this.db.size()
```
ygao76, line 409: I am logging the chain db size when adding a new block. I am not sure if this will affect the chain performance. Let me know if there is better place to do it.

ironfish/src/metrics/metricsMonitor.ts
```diff
@@ -19,6 +19,7 @@ export class MetricsMonitor {
 
   readonly mining_newBlockTemplate: Meter
   readonly chain_newBlock: Meter
+  readonly chainDatabaseSize: Gauge
```
rohanjadvani, line 22:
```suggestion
  readonly chain_databaseSize: Gauge
```

ironfish/src/blockchain/blockchain.ts
rohanjadvani: What about disconnecting a block?

ironfish/src/metrics/metricsMonitor.ts
ygao76: I will change it, curious about the reason for this particular name convention.

ironfish/src/blockchain/blockchain.ts
ygao76: Added for removeBlock()

ironfish/src/blockchain/blockchain.test.ts
```diff
@@ -103,6 +103,8 @@ describe('Blockchain', () => {
     expect((await chain.getHashAtSequence(2))?.equals(headerA1.hash)).toBe(true)
     expect((await chain.getHashAtSequence(3))?.equals(headerB2.hash)).toBe(true)
     expect((await chain.getHashAtSequence(4))?.equals(headerB3.hash)).toBe(true)
+
+    expect(await chain.db.size()).toBeGreaterThanOrEqual(size)
```
danield9tqh, line 107: I don't think this is testing much if we're accepting it can be equal as well. They could both be 0. I think removing this would be fine or just check strictly greater than. Ideally maybe it's own test too

ironfish/src/blockchain/blockchain.ts
```diff
@@ -406,6 +406,8 @@ export class Blockchain {
 
         this.resolveOrphans(block)
 
+        this.metrics.chain_databaseSize.value = await this.db.size()
```
danield9tqh, line 409: Do we know what performance of this size check is like? Just want to make sure we're not adding extra time to addBlock

ironfish/src/metrics/metricsMonitor.ts
```diff
@@ -66,6 +67,7 @@ export class MetricsMonitor {
 
     this.mining_newBlockTemplate = this.addMeter()
     this.chain_newBlock = this.addMeter()
+    this.chain_databaseSize = new Gauge()
```
danield9tqh, line 70: We should probably find a way to initialize this to the size of the chain db. Maybe set it when the database starts up so that we don't get 0 values sent to telemetry if it's not initialized

ironfish/src/blockchain/blockchain.ts
ygao76: I just checked the operation takes 0.031458 milliseconds, it should be negligible.

ironfish/src/metrics/metricsMonitor.ts
ygao76: Added in the blockchain.open()

ironfish/src/blockchain/blockchain.test.ts
ygao76: Updated to check the chain db size > 0. The size function is the approximate size https://github.com/Level/leveldown#dbapproximatesizestart-end-callback There is disclaimer that the result might not include the recent writes. In the previous test, the size value actually does not change due to this.

danield9tqh (approved):

rohanjadvani (approved):

PR #3930, opened 2023-05-24
Perf test reporter and CI integration

ygao76:
## Summary
Add perf test in CI
Use Jest junit to produce test report
Display the test report in github workflow, see `Output test result`
Upload the test report to artifact

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```
@NullSoldier @dguenther


.github/workflows/ci.yml
```diff
@@ -99,3 +104,56 @@ jobs:
...
+          job_summary: true
+          json_file: 'tests.json'
+          report_individual_runs: true
+          report_suite_logs: 'any'
+          json_suite_details: true
+          json_test_case_results: true
+      
+      - name: Output test results
+        uses: ./.github/json_output
+        if: always()
```
danield9tqh, line 149: Do we need these `always` statements? Will it not get executed sometimes without them?

.github/json_output/action.yml
```diff
@@ -0,0 +1,31 @@
+name: 'Assert JSON output'
```
danield9tqh, line 1: Is this actions copied from somewhere? If we're modifying it I think we can change the name and author. This looks like just 2 bash calls using `jq` which is pretty generic. Maybe naming it something like `parse performance JSON output`

danield9tqh:
Could we look into using a custom Jest reporter to get this JSON file instead of JUnit? Sounds like we would want to have some custom results added to our results file that JUnit wouldn't support. Also we already use jest for code coverage reporting so I think using it for this reporting instead of bringing in another library could be good too. 

Something like this looks promising https://github.com/sholzmayer/jest-performance-reporter/tree/main
Or this https://chanonroy.medium.com/how-to-setup-a-custom-test-reporter-for-jest-f994636073a9

.github/workflows/ci.yml
ygao76: The step will not execute if previous step failed. In our cases, if any test fails it wont continue to run the following steps. Since these are perf tests, it will be better to run the whole flow and generate the final detailed results.

ironfish/src/workerPool/tasks/workerMessages.test.perf.ts
```diff
@@ -48,7 +48,11 @@ describe('WorkerMessages', () => {
     })
 
     expect(true).toBe(true)
-    printResults('createMinersFeeRequest', runs, segment)
+    const { min, max, avg } = printResults('createMinersFeeRequest', runs, segment)
+
+    expect(max).toBeLessThanOrEqual(1.191)
```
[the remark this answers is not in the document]
ygao76: These numbers are the baseline numbers, basically the number I got when I run these tests in CI jobs and locally.
If the performance is greatly improved by a certain PR, we would want to change the numbers accordingly. I imagine this will not super frequent, maybe once in a quarter.

For the flaky failures, I want to make the perf test as optional so it does not block PR. But we kept the records and can easily identify PRs that  increase the latency.

ygao76:
> Could we look into using a custom Jest reporter to get this JSON file instead of JUnit? Sounds like we would want to have some custom results added to our results file that JUnit wouldn't support. Also we already use jest for code coverage reporting so I think using it for this reporting instead of bringing in another library could be good too.
> 
> Something like this looks promising https://github.com/sholzmayer/jest-performance-reporter/tree/main Or this https://chanonroy.medium.com/how-to-setup-a-custom-test-reporter-for-jest-f994636073a9

Sounds good, I created a linear ticket for customized test report https://linear.app/if-labs/issue/IFL-1101/create-customized-test-report
I will work on that tmrw. Can I have that in a follow up PR since this PR is already getting very big?

.github/json_output/action.yml
ygao76: Updated the names

ironfish/src/blockchain/blockchain.test.perf.ts
```diff
@@ -61,104 +57,195 @@ describe('Blockchain', () => {
...
+
+    const addBlockOnForkTime = MathUtils.arrayAverage(result.fork)
+    expect(addBlockOnForkTime).toBeLessThanOrEqual(75.1)
+
+    const addHeadRewindForkBlocksTime = MathUtils.arrayAverage(result.rewind)
+    expect(addHeadRewindForkBlocksTime).toBeLessThanOrEqual(82.6)
+  })
+
+  it('Times Ran: 5, Fork Length: 10', async () => {
+    const result = await runTest(5, 10)
```
danield9tqh, line 118: Are we planning on using the timing of `it(` blocks as the final result of performance tests? If so I think we should minimize code that's called in this block besides the `runTest` function so we get accurate numbers

ygao76: Moved the node set up outside it()

ironfish/package.json
```diff
@@ -83,10 +87,12 @@
     "eslint-plugin-react-hooks": "4.2.0",
     "jest": "29.3.1",
     "jest-jasmine2": "29.3.1",
+    "jest-junit": "16.0.0",
```
danield9tqh, line 90: Prob can remove

ironfish/src/testReporter/testReporter.ts
```diff
@@ -0,0 +1,83 @@
...
+    const consoleOutputs = testResult.console?.filter((output) => output.type === 'log')
+
+    testResult.testResults.forEach((result, i) => {
+      const row: Record<string, string> = {
+        name: result.title,
+        duration: String(result.duration),
+      }
+
+      if (consoleOutputs && consoleOutputs[i]) {
+        const entries = consoleOutputs[i].message.split(',')
```
danield9tqh, line 69: It doesn't look like there are many better ways to get custom data from the tests into the test reporter. See discussion [here](https://github.com/jestjs/jest/issues/7421). It could be possible but seems like too much extra effort. I think the console buffer is a pretty good potential solution. 

We should definitely make this serialization / deserialization from the console buffer more standardized though. @ygao76 and I talked offline. We're going to merge this in so that we're able to test post-processing steps like uploading to influx but standardizing serialization / deserialization will be done in a future PR.

danield9tqh (approved):

PR #3937, opened 2023-05-29
Template generation optimisations. Caching and pre-generation.

kid-alan:
## Summary
We have identified few key bottlenecks in generating the template for mining:
1. minersFee takes up to 500-1000ms to generate.
2. A brand new template is generated for every connected RPC client. If 10 clients are connected, one of them can wait for a template for up to 5-10 seconds.

So the total time between "t1) discovering a new block" and "t2) miners starting mining a new one" was on average around 700-1000ms. During that time miners are mining an old height, resulting in a very high fork rates.

This PR provides a couple optimisations to solve this.

1. We pre-generate empty block minersFee for the N+1 sequence in advance.
2. When height changes, we are thus able to start mining the empty block almost immediately.
3. After that, we immediately generate a proper full block with transactions and push it.

In the process of doing this, we are also caching the empty and full block results for use by other RPC clients.
Every client after is getting the cached template, so the time to push new work to 10 clients is cut from 5-10 seconds down to 100-300ms.

Variations of this code is currently running on Kryptex production pool, since the height 50036 25-05-2023.
We have found 309 blocks with this code. 0 of them are forks.

## Commentaries on code
* It is not properly linted yet. Linter is very distracting for me, so I disable it during the development.
* In the process of development I relied heavily on logs to observe the chronological flow of the events. Logs make the code very hard to read, so I removed them in the last commit. Revert it to put them back and observe the way the code flows during the runtime.

## What else needs to be done
There are some things that are still need to be done:
* I wrote no new tests.
* If
  1. The default account is changed when the node is running on height N.
  2. Miner finds a solution for N+1
  3. minersFee will be rewarded to old default account, since the minersFee for N+1 was pre-generated using it's spending key. This is a low probability and can be avoided by using height + spending key as a cache key, but I am keeping it simple for now.

I hope the Iron team will be able to pick it up from here.
I will be happy to hear the feedback and help with merging the code into mainline.

## Testing Plan
No testing plan.

## Documentation
No documentation change is needed, as far as I see.
However, the logic can be better documented in code's docstrings. I will get back to this in a few days.

## Breaking Change

This is not a breaking change. It works fine with older nodes and it required no change in our pools code.


danield9tqh:
Closing this because it was completed in https://github.com/iron-fish/ironfish/pull/3942 and https://github.com/iron-fish/ironfish/pull/3965

PR #3943, opened 2023-06-01
Add the 'expiration' field to transactions returned by followChainStream

andreacorbellini:
## Summary

This change exposes the expiration information of transactions in the output of `chain/followChainStream`. This new field is always present, even with the expiration is set to 0.

## Testing Plan

Checked the output of `curl -s http://127.0.0.1:8021/chain/followChainStream` and verified that the `expiration` field is present and contains an integer value.

## Documentation

https://github.com/iron-fish/website/pull/457

## Breaking Change

No, because this only adds a new field; it does not alter/delete existing fields.

### DIFF
--- a/ironfish/src/rpc/routes/chain/followChainStream.ts
+++ b/ironfish/src/rpc/routes/chain/followChainStream.ts
@@ -35,6 +35,7 @@ export type FollowChainStreamResponse = {
       hash: string
       size: number
       fee: number
+      expiration: number
       notes: Array<{ commitment: string }>
       spends: Array<{ nullifier: string }>
       mints: Array<{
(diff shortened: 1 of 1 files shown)


ironfish/src/rpc/routes/chain/followChainStream.ts
```diff
@@ -153,6 +153,7 @@ router.register<typeof FollowChainStreamRequestSchema, FollowChainStreamResponse
             hash: BlockHashSerdeInstance.serialize(transaction.hash()),
             size: getTransactionSize(transaction),
             fee: Number(transaction.fee()),
+            expiration: transaction.expiration(),
```
NullSoldier, line 156: ok nice, thsi also needs to be added to the runtime type validator above for the response.

NullSoldier (approved):

PR #3950, opened 2023-06-06
Fix wallet:transaction decrypting at most 255 notes

dguenther:
## Summary

Fixes a bug where calling `wallet:transaction` or `wallet:transactions` on a transaction with more than 255 output notes would only display at most the first 255 results.

The decryptNotes task passes in the number of notes to decrypt as a u8, but bufio only bounds-checks if a number is a safe integer when writing numbers, and we don't do any verification that the number of results matches the number of inputs, so it's possible to overflow the value. In the wallet's transaction processing code, we decrypt notes in batches of 20, so this doesn't affect it. But when calling `wallet:transaction` or `wallet:transactions`, we attempt to decrypt output notes as a spender in one batch, and since transactions can have several hundred notes, this opens up the possibility for this bug.

The PR's change removes the length field from the serialized output -- since the worker pool is internal to the SDK and nothing external can connect to it, I don't think it's necessary to encode the number of notes as part of the input.

This doesn't affect a user's balance or available notes in their wallet, and transactions don't need to be rescanned or re-decrypted as a result of this change. Also, incoming/received notes are not affected by this, so this bug would only affect display of transactions where the decrypting account sent more than 255 notes.

## Testing Plan

* Added tests for including over 255 notes in a decryptNotes job.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/workerPool/tasks/decryptNotes.test.ts
+++ b/ironfish/src/workerPool/tasks/decryptNotes.test.ts
@@ -33,6 +33,25 @@ describe('DecryptNotesRequest', () => {
     const deserializedRequest = DecryptNotesRequest.deserializePayload(request.jobId, buffer)
     expect(deserializedRequest).toEqual(request)
   })
+
+  it('serializes over 255 notes', () => {
+    const length = 600
+
+    const request = new DecryptNotesRequest(
+      Array.from({ length }, () => ({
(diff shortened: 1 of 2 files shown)


jowparks (approved):
TY for fixing this 😅

lyonsstar:
how can i update node version?

lyonsstar:
how can i update node version? what is the commands?

PR #3954, opened 2023-06-07
Move chain db size out of telemetry class and into stats cmd

ygao76:
## Summary
Move chain db size to stats cmd
Add chain db size to node getStatus response
Remove chain db size from telemetry class and blockchain class

## Testing Plan
locally run `yarn start service:stats` 
log chain db size to console
```
yajun@Yajuns-MBP ironfish-cli % yarn start service:stats
yarn run v1.22.19
$ yarn build && yarn start:js service:stats
$ tsc -b
$ cross-env OCLIF_TS_NODE=0 IRONFISH_DEBUG=1 node --expose-gc --inspect=:0 --inspect-publish-uid=http --enable-source-maps bin/run service:stats
module: @oclif/core@1.23.1
task: toCached
plugin: ironfish
root: /Users/yajun/ironfish/ironfish/ironfish-cli
See more details with DEBUG=*
chain DB size 3486947512
```

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```
@dguenther

### DIFF
--- a/ironfish-cli/src/commands/service/stats.ts
+++ b/ironfish-cli/src/commands/service/stats.ts
@@ -50,6 +50,7 @@ export default class Stats extends IronfishCommand {
     await this.sdk.client.connect()
 
     // metric loops, must await last loop
+    await this.chainDBSize(api, flags.delay)
     void this.forks(api, flags.delay)
     await this.feeRates(api, flags.delay)
   }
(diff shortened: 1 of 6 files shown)


ironfish-cli/src/commands/service/stats.ts
```diff
@@ -50,6 +50,7 @@ export default class Stats extends IronfishCommand {
     await this.sdk.client.connect()
 
     // metric loops, must await last loop
+    await this.chainDBSize(api, flags.delay)
```
dguenther, line 53: I think this will have to be `void` so it doesn't block the other loops (since it's a `while(true)`)

ironfish/src/rpc/routes/chain/getChainInfo.ts
```diff
@@ -35,6 +36,7 @@ export const GetChainInfoResponseSchema: yup.ObjectSchema<GetChainInfoResponse>
       .object({ index: yup.string().defined(), hash: yup.string().defined() })
       .defined(),
     currentBlockTimestamp: yup.number().defined(),
+    chainDBSizeBytes: yup.number().defined(),
```
dguenther, line 39: I think this should go on `node/getStatus` since our other CPU/memory-related metrics are stored there now.

ygao76: Sure, under blockchain section right?

dguenther (approved):
Looks good! For documentation, this will require an update to https://ironfish.network/developers/documentation/rpc/node/get_status once it's released, so you could prepare a draft PR with the changes for when that happens

dguenther (approved):
Looks good! For documentation, this will require an update to https://ironfish.network/developers/documentation/rpc/node/get_status once it's released, so you could prepare a draft PR with the changes for when that happens

PR #3961, opened 2023-06-12
fix(ironfish-rust): Use rustls instead of system openssl

mat-if:
## Summary

Realized this was causing a dependency on OpenSSL for building, which is not the end of the world, but our build system won't require any changes to use rustls instead of openssl

## Testing Plan

- Built on my dev EC2 instance
- Ran the build CI workflow: (TODO add link when build finishes)

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


EvanJRichard (approved):
I like the bonus cleanup this adds.

PR #3963, opened 2023-06-13
removes extra block verification on submitted blocks

hughy:
## Summary

submitBlockTemplate calls verifyBlock and returns INVALID_BLOCK if verification fails. if verification succeeds, then submitBlockTemplate calls addBlock.

addBlock also calls verifyBlock (by way of connect, addHeadToChain, and verifyBlockAdd), so the first verification call is unnecessary.

removes the first block verification and the INVALID_BLOCK result.

## Testing Plan

- update existing unit test

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/mining/manager.ts
+++ b/ironfish/src/mining/manager.ts
@@ -29,7 +29,6 @@ import { MinersFeeCache } from './minersFeeCache'
 export enum MINED_RESULT {
   UNKNOWN_REQUEST = 'UNKNOWN_REQUEST',
   CHAIN_CHANGED = 'CHAIN_CHANGED',
-  INVALID_BLOCK = 'INVALID_BLOCK',
   ADD_FAILED = 'ADD_FAILED',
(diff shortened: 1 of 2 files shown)


ironfish/src/mining/manager.ts
```diff
@@ -380,15 +379,6 @@ export class MiningManager {
       }
     }
 
-    const validation = await this.node.chain.verifier.verifyBlock(block)
```
ygao76, line 383: Looks like the `verifyBlock` does more verification than the verify logic in `addBlock`

`addBlock` only verify the block header, but `verifyBlock` also verify the transaction proof and miner fee

ygao76: After reading the code, `addForkToChain` also calls `verifyBlock` eventually. Thanks for finding this!

ygao76 (approved):

PR #3965, opened 2023-06-13
fix: Fix mining manager not sending out empty block templates

mat-if:
## Summary

This was leading to an issue where time to send out blocks was almost 2x slower than before this change since it would only send out "full" blocks, despite creating both empty and full blocks.

A quick measurement of wall-clock time on my laptop from when the head changed to when a block template was sent out:
 - master: ~450ms
 - staging: ~450 - 1000ms
 - With bug fixed: ~17ms

While investigating this bug, I ran across a few other issues:
 - Due to a priority check, subsequent empty blocks were getting dropped instead of being sent out when they should.
   - Example: Chain changes head to block 5. Manager creates empty template 6 (priority 0), sends it out. Manager creates full template 6 (priority 1), sends it out. Chain changes head to block 6. Manager creates empty template 7 (priority 0), does _not_ send it out due to only the priority being checked, instead of priority + sequence or something.
 - The order of work being started in onConnectedBlock was not optimal.
 - The work being started in onConnectedBlock was competing with itself by not focusing on getting the empty block out first. Instead, starting several promises and letting them figure it out. In reality, this has little effect on a relatively powerful machine, but since we can optimize it, we should.
 - We were doing checks on whether to stream work out, which was happening after the computation was already done. By changing this logic to avoid doing the work in the first place, we have less wasted computation.
 - Due to pre-checking and avoiding doing extra computations, we were able to simplify the logic and remove some complexity.
 - De-duped 2 functions that were mostly identical.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/mining/manager.ts
+++ b/ironfish/src/mining/manager.ts
@@ -20,6 +20,7 @@ import { Block } from '../primitives/block'
 import { isBlockHeavier } from '../primitives/blockheader'
 import { Transaction } from '../primitives/transaction'
 import { BlockTemplateSerde, SerializedBlockTemplate } from '../serde'
+import { ErrorUtils } from '../utils'
 import { BenchUtils } from '../utils/bench'
 import { GraffitiUtils } from '../utils/graffiti'
 import { SpendingAccount } from '../wallet'
@@ -63,7 +64,14 @@ export class MiningManager {
     this.metrics = options.metrics
     this.minersFeeCache = new MinersFeeCache({ node: this.node })
 
-    this.chain.onConnectBlock.on((block) => void this.onConnectedBlock(block))
+    this.chain.onConnectBlock.on(
+      (block) =>
+        void this.onConnectedBlock(block).catch((error) => {
+          this.node.logger.info(
+            `Error creating block template: ${ErrorUtils.renderError(error)}`,
+          )
+        }),
+    )
   }
 
   get minersConnected(): number {
@@ -79,7 +87,11 @@ export class MiningManager {
       // Send an initial block template to the requester so they can begin working immediately
       void this.chain.getBlock(this.chain.head).then((currentBlock) => {
         if (currentBlock) {
-          void this.onConnectedBlock(currentBlock)
+          void this.onConnectedBlock(currentBlock).catch((error) => {
+            this.node.logger.info(
+              `Error creating block template: ${ErrorUtils.renderError(error)}`,
+            )
+          })
         }
       })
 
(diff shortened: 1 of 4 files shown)


ironfish/src/mining/manager.ts
```diff
@@ -167,26 +151,25 @@ export class MiningManager {
...
+    // Only try creating a block with transactions if there are transactions in
+    // the mempool
+    if (this.memPool.count()) {
+      const template = await this.createNewBlockTemplate(currentBlock, account)
+      this.metrics.mining_newBlockTemplate.add(BenchUtils.end(connectedAt))
+      this.streamBlockTemplate(currentBlock, template)
+    }
+
+    // Kick off job to create the next empty miners fee
+    this.minersFeeCache.startCreatingEmptyMinersFee(currentBlock.header.sequence + 2, account)
```
danield9tqh, line 172: would we still want to still kick off this job at the beginning? Or at least before we finish creating the full block?

ironfish/src/mining/manager.ts
```diff
@@ -84,15 +79,15 @@ export class MiningManager {
       // Send an initial block template to the requester so they can begin working immediately
       void this.chain.getBlock(this.chain.head).then((currentBlock) => {
         if (currentBlock) {
-          this.onConnectedBlock(currentBlock)
+          void this.onConnectedBlock(currentBlock)
```
danield9tqh, line 82: I think we might need a catch statement here and on `this.chain.onConnectBlock.on`. If there's an error in this function not positive but I think it could crash the node as an unhandled promise exception

ironfish/src/mining/manager.ts
mat-if: My assumption is that the order of priorities should be:
1. Get an empty block out ASAP - this is for the miners' benefit
2. Get a block with transactions out ASAP - this is for the network's benefit
3. Pre-empt the next miners fee. This is closely related to 1, since without this, 1. is delayed on the next block. If we assume the next block happens within 2-5s or whatever, then the next empty block suffers.

Thinking through this, since _most_ of the work for miners fee happens in a worker job, I think it is a good idea to move this above the full block kick-off, since they can happen in tandem. Although I do think getting the empty block sent out should be priority 1, and kicking off the job isnt quite free (although it's pretty close).

I'll move this up, good call.

danield9tqh: Yea I agree
1. Empty block
2. Kick off next miners fee
3. Full block

hughy (approved):
I used the attached diff for some quick and dirty testing

[test.stream.diff.txt](https://github.com/iron-fish/ironfish/files/11740332/test.stream.diff.txt)

On my machine
~50ms between block connect and the first template on this branch
~2000ms on staging

danield9tqh (approved):

PR #3967, opened 2023-06-14
verifies block template after streaming

hughy:
## Summary

optimistically sends new block templates to connected clients (miners and pools) before block verification.

we currently verify a block before sending the template out. however, if the block verification fails, then the node will crash.

block verification takes time, and miners could be mining on an old template or an empty template while verification runs. instead, we can send out the template(s) as quickly as possible and verify afterwards.

- makes block verification optional in newBlock
- does not verify block for empty block template
- verifies 'full' template after sending it out to clients
- re-sends empty template if verification of full template fails

## Testing Plan

- adds unit tests
- performance testing, time from block connect to template stream:
  - staging: ~13-18ms for empty, ~1200-1600ms for full
  - feature: ~7-8ms for empty, ~900-1200ms for full

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -1009,7 +1009,7 @@ export class Blockchain {
       )
 
       const block = new Block(header, transactions)
-      if (!previousBlockHash.equals(GENESIS_BLOCK_PREVIOUS) && verifyBlock) {
+      if (verifyBlock && !previousBlockHash.equals(GENESIS_BLOCK_PREVIOUS)) {
         // since we're creating a block that hasn't been mined yet, don't
         // verify target because it'll always fail target check here
         const verification = await this.verifier.verifyBlock(block, { verifyTarget: false })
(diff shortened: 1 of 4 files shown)


lyonsstar:
I mined some coins on the test net but when I updated to the main net I couldn't see my coins! What should I do  to back my coins

ironfish/src/blockchain/blockchain.ts
```diff
@@ -1008,7 +1009,7 @@ export class Blockchain {
       )
 
       const block = new Block(header, transactions)
-      if (!previousBlockHash.equals(GENESIS_BLOCK_PREVIOUS)) {
+      if (!previousBlockHash.equals(GENESIS_BLOCK_PREVIOUS) && verifyBlock) {
```
mat-if, line 1012: We can avoid the buffer comparison when verifyBlock is false if we move it first. Pretty minor over-optimization, but 🤷 
```suggestion
      if (verifyBlock && !previousBlockHash.equals(GENESIS_BLOCK_PREVIOUS)) {
```

lyonsstar:
Thanks but I'm not good with codes. Could you please explain it

PR #3972, opened 2023-06-15
Fix error message when chain head changes

danield9tqh:
## Summary
Miners are getting the error message `Error creating block template: Cannot create a block with a previous header that does not match` when the head changes during a template creation. This is sometimes expected behavior so when it is we don't want to surface this error to users

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -968,7 +968,7 @@ export class Blockchain {
         }
 
         if (previous && !previous.hash.equals(previousBlockHash)) {
-          throw new Error('Cannot create a block with a previous header that does not match')
+          throw new BlockCreationError(BlockCreationErrorReason.HEAD_CHANGED)
         }
 
         target = Target.calculateTarget(
@@ -1518,5 +1518,19 @@ export class VerifyError extends Error {
 
     this.reason = reason
     this.score = score
+  }
+}
+
+export enum BlockCreationErrorReason {
+  HEAD_CHANGED = 'HEAD_CHANGED',
+}
+
+export class BlockCreationError extends Error {
+  name = this.constructor.name
+  reason: BlockCreationErrorReason
+
+  constructor(reason: BlockCreationErrorReason) {
+    super()
+    this.reason = reason
   }
 }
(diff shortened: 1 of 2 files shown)


ironfish/src/blockchain/blockchain.ts
```diff
@@ -1520,3 +1520,17 @@ export class VerifyError extends Error {
     this.score = score
   }
 }
+
+export enum BlockCreationErrorReason {
+  HEAD_CHANGED = 'HEAD_CHANGED',
+}
```
hughy, line 1526: Since there's only one special case here right now, could we just have a `HeadChangedError`?

ironfish/src/mining/manager.ts
```diff
@@ -66,9 +66,21 @@ export class MiningManager {
...
-          this.node.logger.info(
-            `Error creating block template: ${ErrorUtils.renderError(error)}`,
-          )
+          const headChanged =
+            error instanceof BlockCreationError &&
+            error.reason === BlockCreationErrorReason.HEAD_CHANGED
+
+          if (headChanged) {
+            this.node.logger.info(
+              `Chain head changed while creating block template for seqeunce ${
```
hughy, line 75: `s/seqeunce/sequence/`

ironfish/src/blockchain/blockchain.ts
danield9tqh: Yea fair

ironfish/src/mining/manager.ts
dguenther: What do you think about bumping this down to `logger.debug`? Is it useful for general users to see it in their logs?

danield9tqh: Yea I changed to debug

ironfish/src/mining/manager.ts
```diff
@@ -66,9 +66,17 @@ export class MiningManager {
     this.chain.onConnectBlock.on(
       (block) =>
         void this.onConnectedBlock(block).catch((error) => {
-          this.node.logger.info(
-            `Error creating block template: ${ErrorUtils.renderError(error)}`,
-          )
+          if (error instanceof HeadChangedError) {
+            this.node.logger.debug(
+              `Chain head changed while creating block template for seqeunce ${
```
hughy:
```suggestion
              `Chain head changed while creating block template for sequence ${
```

hughy (approved):

Issue #3944, opened 2023-06-02
command line tool, sort not work


ofloxacin:
### What happened?

ironfish wallet:transactions --datadir=~/.dev0 --sort=-Submitted，sort not work。

```
    Timestamp                Status      Type    Hash                                                             Expiration Submitted Sequence Fee Paid ($IRON) Asset                     Amount          
 ── ──────────────────────── ─────────── ─────── ──────────────────────────────────────────────────────────────── ────────── ────────────────── ──────────────── ───────────────────────── ─────────────── 
    2023/6/2 GMT+8 14:47:45  unconfirmed miner   f531184081b2ce913ab1c569eff162968d4dc1a42d69822a503b80b089e418d9 0          380                                 $IRON             (51f33) 20.00000000     
    2023/6/2 GMT+8 14:47:44  unconfirmed miner   1c3fd0917eec30eb66b6298b6b28cb8b8431febe67c10a7653a6e732b74d331c 0          379                                 $IRON             (51f33) 20.00000000     
    2023/6/2 GMT+8 14:47:43  confirmed   miner   7fdf5e548597cd5cf94501069a94a65b28a60495511e5615d60f46c412f7efa9 0          378                                 $IRON             (51f33) 20.00000000     
    2023/6/2 GMT+8 14:47:40  confirmed   miner   747a7f40c3b2a6c405d7a6bf6dee8f8e25770be7756f75482b1489b7534f0fe2 0          377                                 $IRON             (51f33) 20.00000000     
    2023/6/2 GMT+8 14:47:36  confirmed   miner   16aaf9745bdab9ce07172f1a07ba7167119dad1834a72687cfd68a784e9d878d 0          376                                 $IRON             (51f33) 20.00000000     
```

### Version

1.1.0

hughy:
Unfortunately the sort flag doesn't work right now because an account may have too many transactions for the node to return to the CLI at one time.

#3975 will remove the flag to avoid confusion, and we'll look at adding support for sorting transactions in the future!

PR #3975, opened 2023-06-16
fixes usage of oclif table sort flag

hughy:
## Summary

in many of our CLI command that use oclif tables we include the 'sort' flag. however, if we are displaying streamed responses in a table then the sort flag doesn't work because each response is a separate table.

- removes 'sort' flag from commands where it can't be used
- adds message if 'sort' is used along with 'follow' and can't be supported in peers commands
- removes default setting of 'STATE' for sort in peers command: the table flags are global, so this was being applied to every command that includes the sort flag

Relates to #3944

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/peers/banned.ts
+++ b/ironfish-cli/src/commands/peers/banned.ts
@@ -7,7 +7,7 @@ import blessed from 'blessed'
 import { IronfishCommand } from '../../command'
 import { RemoteFlags } from '../../flags'
 
-const tableFlags = CliUx.ux.table.flags()
+const { sort, ...tableFlags } = CliUx.ux.table.flags()
 
 export class BannedCommand extends IronfishCommand {
   static description = `List all banned peers`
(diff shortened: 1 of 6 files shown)


ironfish-cli/src/commands/peers/banned.ts
```diff
@@ -32,6 +32,10 @@ export class BannedCommand extends IronfishCommand {
       this.exit(0)
     }
 
+    if (flags.sort !== undefined) {
+      this.log('The `sort` flag is not supported when using the `follow` flag.')
+    }
```
NullSoldier, line 37: I'm not sure if you can modify the sort flag that's returned but if you can you can use the built in feature for this which is `exclusive: ['follow']` on the sort flag.

NullSoldier (approved):

ironfish-cli/src/commands/peers/banned.ts
hughy: Good call, updated!

PR #3984, opened 2023-06-22
handles errors sending jobs to worker threads

hughy:
## Summary

the worker pool tracks the state of a job by assigning it a status and by assigning it to a worker. that assignment, placing the job in the worker's jobs map, occurs before sending the job to the worker. the job's status is also set to 'executing' before sending the job to the worker thread.

if an error occurs in sending the job to the worker, such as when serializing the job, then we do not reset these statuses: the job is left in an 'executing' state even though it is not executing, and the worker keeps the job in its map even though it never received it.

at this point the worker pool will not assign any more jobs to that worker.

- catches errors from 'worker.send' and sets the job status to 'error'
- moves assignment of the job to the worker's data structures to 'job.execute' and only assigns the job if it was successfully sent to the worker thread.
- removes 'worker.execute' since it only calls 'job.execute'

## Testing Plan

- adds unit test
- manual testing:
  - changed telemetry to include bad data that won't serialize: float value in integer field
  - changed telemetry to flush very every second
  - set telemetry to write to a local API host
  - ran the node with telemetry enabled
  - used 'workers:status' to see telemetry jobs getting stuck and other jobs filling the queue once all workers were stuck

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/workerPool/job.test.slow.ts
+++ b/ironfish/src/workerPool/job.test.slow.ts
@@ -49,6 +49,16 @@ describe('Worker Pool', () => {
 
       expect(workerPool.workers.length).toBe(1)
 
+      class ErrorWorkerMessage extends WorkerMessage {
+        serializePayload(): void {
+          throw new Error('Always throw an error during serialization.')
+        }
+
+        getSize(): number {
+          return 0
+        }
+      }
+
       const worker = workerPool.workers[0]
       const message = new ErrorWorkerMessage(WorkerMessageType.JobError)
       const job = new Job(message)
@@ -59,16 +69,7 @@ describe('Worker Pool', () => {
 
       expect(job.status).toEqual('error')
       expect(worker.jobs.size).toEqual(0)
+      expect(worker.canTakeJobs).toBe(true)
     })
   })
 })
-
-class ErrorWorkerMessage extends WorkerMessage {
-  serializePayload(): void {
-    throw new Error('Always throw an error during serialization.')
-  }
-
-  getSize(): number {
-    return 0
-  }
-}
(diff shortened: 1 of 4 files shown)


ironfish/src/workerPool/job.test.slow.ts
```diff
@@ -38,4 +40,35 @@ describe('Worker Pool', () => {
...
+
+class ErrorWorkerMessage extends WorkerMessage {
+  serializePayload(): void {
+    throw new Error('Always throw an error during serialization.')
+  }
+
+  getSize(): number {
+    return 0
+  }
+}
```
NullSoldier, line 74: You can move this class into teh test since it'll only be used once too if you want.

ironfish/src/workerPool/job.test.slow.ts
```diff
@@ -38,4 +40,35 @@ describe('Worker Pool', () => {
...
+      const worker = workerPool.workers[0]
+      const message = new ErrorWorkerMessage(WorkerMessageType.JobError)
+      const job = new Job(message)
+
+      expect(job.status).toEqual('queued')
+
+      expect(() => job.execute(worker)).toThrow()
+
+      expect(job.status).toEqual('error')
+      expect(worker.jobs.size).toEqual(0)
```
NullSoldier, line 61: should we check that the worker is still able to accept other jobs?

NullSoldier:
Ok, this PR looks good. It does mean we now need to track down send() errors and fix them because they'll now crash the node if they are unhandled but that's a good outcome. I tested with this before `this.thread.postMessage()`

```
if (Math.random() < 0.0005) {
    throw new Error()
}
```

NullSoldier (approved):

ironfish/src/workerPool/job.test.slow.ts
hughy: Added a check for `canTakeJobs`

PR #3988, opened 2023-06-27
Move perf test to separate workflow

ygao76:
## Summary
Move perf test to separate workflow, so it will only trigger on :
push to staging and master
PR on master

## Testing Plan
github workflow

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```
@danield9tqh

### DIFF
--- a/.github/workflows/perf_test.yml
+++ b/.github/workflows/perf_test.yml
@@ -1,18 +1,14 @@
 name: Performance Tests
 
 on:
-  pull_request:
-    branches:
-      - master
   push:
     branches:
-      - master
       - staging
 
 permissions:
   checks: write
   contents: write
-  
+
 jobs:
   testperf:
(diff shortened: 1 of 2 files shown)


.github/workflows/perf_test.yml
```diff
@@ -0,0 +1,66 @@
+name: Performance Tests
+
+on:
+  pull_request:
```
danield9tqh, line 4: I think we should just do on pushes to staging right now. Maybe pushes to master but pushes to staging should include that as well. Probably don't want pull_request to master

danield9tqh (approved):

.github/workflows/perf_test.yml
ygao76: updated

PR #3989, opened 2023-06-27
adds event/onReorganizeChain RPC

hughy:
## Summary

streams the block headers for the old head, the new head, and header for the point of the fork: the block at which the two forks diverged

adds an event to the blockchain reorganizeChain method to emit these headers on each reorg

supports monitoring chain reorgs over time

## Testing Plan

manual testing:
- ran two nodes with two miners to create forks
- connected to RPC endpoint and logged events

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -126,6 +126,10 @@ export class Blockchain {
   onDisconnectBlock = new Event<[block: Block, tx?: IDatabaseTransaction]>()
   // When ever a block is added to a fork
   onForkBlock = new Event<[block: Block, tx?: IDatabaseTransaction]>()
+  // When ever the blockchain is reorganized
+  onReorganizeChain = new Event<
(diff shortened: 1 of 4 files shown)


ironfish/src/blockchain/blockchain.ts
```diff
@@ -126,6 +126,10 @@ export class Blockchain {
   onDisconnectBlock = new Event<[block: Block, tx?: IDatabaseTransaction]>()
   // When ever a block is added to a fork
   onForkBlock = new Event<[block: Block, tx?: IDatabaseTransaction]>()
+  // When ever the blockchain is reorganized
+  onReorganizeChain = new Event<
```
NullSoldier, line 130: Since this is on the blockchain we can remove the chain suffix. I think `blockchain.onReorganize` is obvious that the blockchain is reorganizing.

NullSoldier (approved):

PR #3990, opened 2023-06-27
adds chain reorgs to stats command

hughy:
## Summary

adds a loop to the stats command to listen to a stream of reorganizeChain events from the node and submit reorg data to telemetry.

submits the hash and sequence of the old head, the new head, and the block at which the forks diverged.

## Testing Plan

manual testing:
- running stats command and logging telemetry submissions

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/service/stats.ts
+++ b/ironfish-cli/src/commands/service/stats.ts
@@ -52,6 +52,7 @@ export default class Stats extends IronfishCommand {
     // metric loops, must await last loop
     void this.forks(api, flags.delay)
     void this.chainDBSize(api, flags.delay)
+    void this.chainReorgs(api)
     await this.feeRates(api, flags.delay)
   }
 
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/service/stats.ts
```diff
@@ -188,4 +189,65 @@ export default class Stats extends IronfishCommand {
...
+
+      for await (const value of response.contentStream()) {
+        const { oldHead, newHead, fork } = value
+        void api.submitTelemetry({
+          points: [
+            {
+              measurement: 'chain_reorg',
+              timestamp: new Date(),
+              fields: [
+                {
```
NullSoldier, line 215: Should we also just put a field in here with the length? Or do you want to calculate that in influxdb?

hughy: I was thinking of pushing that calculation to the influx query since it can be derived from the fields here

NullSoldier: That sounds good to me.

NullSoldier (approved):

PR #3993, opened 2023-06-27
Initial implementation of the AssetsVerifier system

andreacorbellini:
## Summary

This is a new system that periodically pulls information about verified assets. The main class is `AssetsVerifier`, which can be used like this:

```typescript
const assetsVerifier = new AssetsVerifier(...options)
assetsVerifier.start() // fetch assets verification information and refresh them every 6 hours
assetsVerifier.verify(assetId) // returns one of:
                               // - { status: 'unknown' }
                               // - { status: 'verified' }
                               // - { status: 'unverified' }
```

For now this system is always enabled in the node, and cannot be configured or turned off. These features will come in future pull requests, together with support for persistent caching, local lists, better error handling with retries, and more.

## Testing Plan

- Unit tests
- Start the node and verify that the new system is running and refreshing itself at a regular interval

## Documentation

Documentation about what is a "verified asset" will be written in the future.

## Breaking Change

Not a breaking change.

### DIFF
--- a/ironfish/src/assets/assetsVerificationApi.ts
+++ b/ironfish/src/assets/assetsVerificationApi.ts
@@ -1,0 +1,74 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import axios, { AxiosError } from 'axios'
+
+type GetVerifiedAssetsResponse = {
+  data: Array<{ identifier: string }>
+}
+
+type GetVerifiedAssetsRequestHeaders = {
+  'if-modified-since'?: string
+}
+
+type GetVerifiedAssetsResponseHeaders = {
+  'last-modified'?: string
+}
+
+export class VerifiedAssets {
+  private readonly assetIds: Set<string> = new Set()
+  private lastModified?: string
+
+  isVerified(assetId: Buffer | string): boolean {
+    if (!(typeof assetId === 'string')) {
+      assetId = assetId.toString('hex')
+    }
+    return this.assetIds.has(assetId)
+  }
+}
+
+export class AssetsVerificationApi {
+  private readonly url: string
+  private readonly timeout: number
+
+  constructor(options?: { url?: string; timeout?: number }) {
+    this.url = options?.url ?? 'https://api.ironfish.network/assets?verified=true'
+    this.timeout = options?.timeout ?? 30 * 1000 // 30 seconds
+  }
+
+  async getVerifiedAssets(): Promise<VerifiedAssets> {
+    const verifiedAssets = new VerifiedAssets()
+    await this.refreshVerifiedAssets(verifiedAssets)
+    return verifiedAssets
+  }
+
+  refreshVerifiedAssets(verifiedAssets: VerifiedAssets): Promise<void> {
+    const headers: GetVerifiedAssetsRequestHeaders = {}
+    if (verifiedAssets['lastModified']) {
+      headers['if-modified-since'] = verifiedAssets['lastModified']
+    }
+    return axios
+      .get<GetVerifiedAssetsResponse>(this.url, {
+        headers: headers,
+        timeout: this.timeout,
+      })
+      .then(
+        (response: {
+          data: GetVerifiedAssetsResponse
+          headers: GetVerifiedAssetsResponseHeaders
+        }) => {
+          verifiedAssets['assetIds'].clear()
+          response.data.data.forEach(({ identifier }) => {
+            return verifiedAssets['assetIds'].add(identifier)
+          })
+          verifiedAssets['lastModified'] = response.headers['last-modified']
+        },
+      )
+      .catch((error: AxiosError) => {
+        if (error.response?.status === 304) {
+          return
+        }
+        throw error
+      })
+  }
+}
--- a/ironfish/src/assets/assetsVerifier.test.ts
+++ b/ironfish/src/assets/assetsVerifier.test.ts
@@ -1,0 +1,258 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import nock from 'nock'
+import { AssetsVerifier } from './assetsVerifier'
+
+/* eslint-disable jest/no-standalone-expect */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+describe('AssetsVerifier', () => {
+  jest.useFakeTimers()
+
+  const waitForRefreshToFinish = async (refreshSpy: jest.SpyInstance) => {
+    for (const result of refreshSpy.mock.results) {
+      await result.value
+    }
+  }
+
+  beforeEach(() => {
+    nock.cleanAll()
+    jest.clearAllTimers()
+  })
+
+  afterEach(() => {
+    expect(nock.pendingMocks()).toHaveLength(0)
+  })
+
+  it('does not refresh when not started', () => {
+    const assetsVerifier = new AssetsVerifier()
+    const refresh = jest.spyOn(assetsVerifier as any, 'refresh')
+
+    jest.runOnlyPendingTimers()
+    expect(refresh).toHaveBeenCalledTimes(0)
+  })
+
+  it('periodically refreshes once started', async () => {
+    nock('https://test')
+      .get('/assets/verified')
+      .reply(200, {
+        data: [{ identifier: '0123' }],
+      })
+      .get('/assets/verified')
+      .reply(200, {
+        data: [{ identifier: '4567' }],
+      })
+      .get('/assets/verified')
+      .reply(200, {
+        data: [{ identifier: '89ab' }],
+      })
+
+    const assetsVerifier = new AssetsVerifier({
+      apiUrl: 'https://test/assets/verified',
+    })
+    const refresh = jest.spyOn(assetsVerifier as any, 'refresh')
+
+    expect(assetsVerifier.verify('0123')).toStrictEqual({ status: 'unknown' })
+    expect(assetsVerifier.verify('4567')).toStrictEqual({ status: 'unknown' })
+    expect(assetsVerifier.verify('89ab')).toStrictEqual({ status: 'unknown' })
+
+    assetsVerifier.start()
+    expect(refresh).toHaveBeenCalledTimes(1)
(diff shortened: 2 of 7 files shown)


ironfish/src/assets/assetsVerifier.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+  }
+
+  verify(assetId: Buffer | string): AssetVerification {
+    if (!this.started || !this.verifiedAssets) {
+      return { status: 'unknown' }
+    } else if (this.verifiedAssets.isVerified(assetId)) {
+      return { status: 'verified' }
+    } else {
+      return { status: 'unverified' }
```
jowparks, line 78: Are we going to have a warning/malicious tag as well? If somehow assets are identified as intentional scam assets, seems like this service would host that info too. Not sure if this is in scope though.

ironfish/src/assets/assetsVerifier.ts
```diff
@@ -0,0 +1,81 @@
...
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { createRootLogger, Logger } from '../logger'
+import { ErrorUtils } from '../utils'
+import { SetIntervalToken } from '../utils'
+import { AssetsVerificationApi, VerifiedAssets } from './assetsVerificationApi'
+
+export type AssetVerification = {
+  status: 'verified' | 'unverified' | 'unknown'
+}
+
```
jowparks, line 12: nit: The nomenclature is a bit misleading here, this doesn't actually do any asset validation, it just retrieves verified assets, maybe rename to something like `VerifiedAssets`

ironfish/src/assets/assetsVerifier.ts
andreacorbellini: This is something that we briefly discussed among me and some other people. There is no strong consensus, so it's not in scope yet, but it may become reality in the future. By the way, this is exactly the reason why this return value is an object and not a string, so that, in a hypothetical future, we could return something like:

```js
{ status: 'flagged',
  flagReason: 'this is used for scams' }
```

ironfish/src/assets/assetsVerifier.test.ts
```diff
@@ -0,0 +1,258 @@
...
+    expect(refresh).toHaveBeenCalledTimes(1)
+    await waitForRefreshToFinish(refresh)
+
+    expect(assetsVerifier.verify('0123')).toStrictEqual({ status: 'verified' })
+    expect(assetsVerifier.verify('4567')).toStrictEqual({ status: 'unverified' })
+    expect(assetsVerifier.verify('89ab')).toStrictEqual({ status: 'unverified' })
+
+    jest.runOnlyPendingTimers()
+    expect(refresh).toHaveBeenCalledTimes(2)
+    await waitForRefreshToFinish(refresh)
```
jowparks, line 70: Really cool pattern! I don't think we have good examples of using mock timers in our codebase.

ironfish/src/assets/assetsVerifier.ts
andreacorbellini: This comment is on the `AssetVerification` type, but I think you're referring to the `AssetsVerifier` class and the module name, am I right?

jowparks: Yea I am, weird the comment looks like it is on line 12/13 to me.

ironfish/src/assets/assetsVerificationApi.ts
```diff
@@ -0,0 +1,74 @@
...
+    return this.assetIds.has(assetId)
+  }
+}
+
+export class AssetsVerificationApi {
+  private readonly url: string
+  private readonly timeout: number
+
+  constructor(options?: { url?: string; timeout?: number }) {
+    this.url = options?.url ?? 'https://api.ironfish.network/assets?verified=true'
```
jowparks, line 35: I looked at this URL and it exists, it seems like it isn't filtering out verified assets, showing them all. Just wanted to flag, but I am sure you are aware.

jowparks:
Very clean PR, looks great. I don't see a local cache for the verified assets, maybe that is to come in another PR. Also might be nice to include some simple access pattern on the `Asset` class or whatever is used elsewhere in the code base so that we don't have to remember to do all of these things to check if an asset is verified:
```
const assetsVerifier = new AssetsVerifier(...options)
assetsVerifier.start() // fetch assets verification information and refresh them every 6 hours
assetsVerifier.verify(assetId)
```

ironfish/src/assets/assetsVerifier.ts
andreacorbellini: I think you have a valid point here. I'll try to think about better names, but I think what you propose (`VerifiedAssets`) is consistent with `Telemetry`, so I might just actually go with your suggestion.

ironfish/src/assets/assetsVerificationApi.ts
andreacorbellini: Ah, yeah. API changes are complete, but they still have to be deployed to api.ironfish.network. They're deployed to testnet.api.ironfish.network FWIW.

jowparks (approved):

andreacorbellini:
Thanks! Yes, persistent caching (and other features) will come in future
PRs. I'm splitting the work in multiple PRs so that they're easier to
review.

The intended pattern is:

```
node.assetsVerifier.verify(assetId)
```

... once you have an `IronfishNode`. I think that putting an access pattern
in `Asset` is not practical because `Asset` cannot access the node, and
also wouldn't be particularly advantageous because we have some pieces of
code that pass around asset IDs without an `Asset` object.

ironfish/src/assets/assetsVerifier.ts
andreacorbellini: And also the `verify` method should change name. Maybe `classify` is a good candidate?

andreacorbellini:
Thanks! Yes, persistent caching (and other features) will come in future PRs. I'm splitting the work in multiple PRs so that they're easier to review.

The intended pattern is:

```
node.assetsVerifier.verify(assetId)
```

... once you have an `IronfishNode`. I think that putting an access pattern in `Asset` is not practical because `Asset` cannot access the node, and also wouldn't be particularly advantageous because we have some pieces of code that pass around asset IDs without an `Asset` object.

ironfish/src/assets/assetsVerifier.ts
```diff
@@ -0,0 +1,81 @@
...
+  private async refresh(): Promise<void> {
+    try {
+      if (this.verifiedAssets) {
+        await this.api.refreshVerifiedAssets(this.verifiedAssets)
+      } else {
+        this.verifiedAssets = await this.api.getVerifiedAssets()
+      }
+    } catch (error) {
+      this.logger.error(
```
NullSoldier, line 66: nit: this is non fatal to the node experience so it should be a warn. If later the source becomes available it'll just pick back up.

ironfish/src/assets/assetsVerifier.ts
```diff
@@ -0,0 +1,81 @@
...
+  private readonly logger: Logger
+  private readonly api: AssetsVerificationApi
+
+  private started: boolean
+  private refreshToken?: SetIntervalToken
+  private verifiedAssets?: VerifiedAssets
+
+  constructor(options?: { apiUrl?: string; logger?: Logger }) {
+    this.logger = options?.logger ?? createRootLogger()
+    this.api = new AssetsVerificationApi({ url: options?.apiUrl })
```
NullSoldier, line 25: We use the WebApi module for this in other places, like the faucet and telemetry.

ironfish/src/assets/assetsVerifier.ts
NullSoldier: I think both the current name, or VerifiedAssets is fine to me. Though I personally like `verify()`. What does the verifier do? verify.

ironfish/src/node.ts
```diff
@@ -351,6 +357,8 @@ export class IronfishNode {
 
     await this.memPool.start()
 
+    this.assetsVerifier.start()
```
NullSoldier, line 360: It looks like we don't call stop() below on line 376.

ironfish/src/assets/assetsVerificationApi.ts
```diff
@@ -0,0 +1,74 @@
...
+type GetVerifiedAssetsRequestHeaders = {
+  'if-modified-since'?: string
+}
+
+type GetVerifiedAssetsResponseHeaders = {
+  'last-modified'?: string
+}
+
+export class VerifiedAssets {
+  private readonly assetIds: Set<string> = new Set()
```
NullSoldier, line 19: It feels like this should be in the verifier, rather than the API since the verifier just consumes the API to hydrate its internal stores.

Also I commented on the spec, but it feels like later we should persist this with a FileStore in the data directory along with the last fetch time. If we did want to do this, we can do it as an expansion.

NullSoldier (approved):
This is the beginning of the system, so we can make some more changes as we go but I ran this locally and it does work. I would say the only bug here is that the stop() isn't called.

NullSoldier:
> Very clean PR, looks great. I don't see a local cache for the verified assets, maybe that is to come in another PR. Also might be nice to include some simple access pattern on the `Asset` class or whatever is used elsewhere in the code base so that we don't have to remember to do all of these things to check if an asset is verified:
> 
> ```
> const assetsVerifier = new AssetsVerifier(...options)
> assetsVerifier.start() // fetch assets verification information and refresh them every 6 hours
> assetsVerifier.verify(assetId)
> ```
> 
> edit: Also I might change from `assetsVerifier.verify(assetId)` to `verifiedAssets.get(assetId)`

I think the verifier verifies, so it makes sense that it has a method called verify.

Also you shouldn't need to ever create an assetsVerifier you should be using the system on the node.

As for adding Asset.verify() it would require Asset class to take the verifier, which I think it would be nice to keep Asset just data and not turn it into a system.

ironfish/src/node.ts
andreacorbellini: Uh oh, good catch

ironfish/src/assets/assetsVerificationApi.ts
andreacorbellini: Yes, this class is meant to be persisted (a future PR will take care of that). Not sure if I it makes sense to move this class to the verifier, because (1) it's the return value of the API class, and (2) it's crucial to the logic of `refreshVerifiedAssets` from the API class, so I'd say it belongs to the API module.

ironfish/src/assets/assetsVerifier.ts
andreacorbellini: I will keep the current names as there is no clear consensus. I might come back to it in a future PR

PR #3994, opened 2023-06-27
Make STUN servers configurable

danield9tqh:
## Summary
We currently use Twilio and Google as STUN servers to facilitate WebRTC. If users are not confortable with this we should give them the option to use another server (possibly in the future a team run server). Making this configurable for now

## Testing Plan
Tested locally + unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/fileStores/config.ts
+++ b/ironfish/src/fileStores/config.ts
@@ -29,7 +29,6 @@ export type ConfigOptions = {
   bootstrapNodes: string[]
   /**
    * STUN servers to use for inititating WebRTC connections.
-   * TODO: Use our own STUN servers
    */
   p2pStunServers: string[]
   databaseMigrate: boolean
--- a/ironfish/src/fileStores/config.ts
+++ b/ironfish/src/fileStores/config.ts
@@ -31,7 +31,7 @@ export type ConfigOptions = {
    * STUN servers to use for inititating WebRTC connections.
    * TODO: Use our own STUN servers
    */
-  stunServers: string[]
+  p2pStunServers: string[]
   databaseMigrate: boolean
   editor: string
   enableListenP2P: boolean
@@ -283,7 +283,7 @@ export type ConfigOptions = {
 export const ConfigOptionsSchema: yup.ObjectSchema<Partial<ConfigOptions>> = yup
   .object({
     bootstrapNodes: yup.array().of(yup.string().defined()),
-    stunServers: yup.array().of(yup.string().defined()),
+    p2pStunServers: yup.array().of(yup.string().defined()),
     databaseMigrate: yup.boolean(),
     editor: yup.string().trim(),
     enableListenP2P: yup.boolean(),
@@ -380,7 +380,7 @@ export class Config extends KeyStore<ConfigOptions> {
   static GetDefaults(files: FileSystem, dataDir: string): ConfigOptions {
     return {
       bootstrapNodes: [],
-      stunServers: ['stun:stun.l.google.com:19302', 'stun:global.stun.twilio.com:3478'],
+      p2pStunServers: ['stun:stun.l.google.com:19302', 'stun:global.stun.twilio.com:3478'],
       databaseMigrate: false,
       transactionExpirationDelta: 15,
       editor: '',
--- a/ironfish/src/fileStores/config.ts
+++ b/ironfish/src/fileStores/config.ts
@@ -27,6 +27,11 @@ const MEGABYTES = 1000 * 1000
 
 export type ConfigOptions = {
   bootstrapNodes: string[]
+  /**
+   * STUN servers to use for inititating WebRTC connections.
+   * TODO: Use our own STUN servers
+   */
+  stunServers: string[]
   databaseMigrate: boolean
   editor: string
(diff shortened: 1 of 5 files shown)


ironfish/src/fileStores/config.ts
```diff
@@ -374,6 +380,7 @@ export class Config extends KeyStore<ConfigOptions> {
   static GetDefaults(files: FileSystem, dataDir: string): ConfigOptions {
     return {
       bootstrapNodes: [],
+      stunServers: ['stun:stun.l.google.com:19302', 'stun:global.stun.twilio.com:3478'],
```
NullSoldier, line 383: can you use the p2p prefix here? It's hard to know what this config is for.

ironfish/src/network/peers/peerManager.ts
```diff
@@ -322,9 +329,15 @@ export class PeerManager {
    * @param initiator Set to true if we are initiating a connection with `peer`
    */
   private initWebRtcConnection(peer: Peer, initiator: boolean): WebRtcConnection {
-    const connection = new WebRtcConnection(initiator, this.logger, this.metrics, {
-      simulateLatency: this.localPeer.simulateLatency,
-    })
+    const connection = new WebRtcConnection(
+      initiator,
+      this.logger,
+      this.stunServers,
```
NullSoldier, line 335: Can this go into the options below?

NullSoldier (approved):
This looks good. Tested manually, and WebRTC connections still work.

ironfish/src/network/peers/peerManager.ts
```diff
@@ -138,12 +143,14 @@ export class PeerManager {
   constructor(
     localPeer: LocalPeer,
     hostsStore: HostsStore,
+    stunServers: string[],
```
andreacorbellini, line 146: What happens if I pass an empty array?

ironfish/src/network/peers/peerManager.ts
danield9tqh: Yea added it to options. Only reason I didn't before is because I didn't want to set a default value. After Andrea's comment I tested the node with and empty list of stunServers and it is still semi-functional so I think setting it to that is reasonable

NullSoldier: yes because it actually just defaults to built in servers, which is also fine

ironfish/src/fileStores/config.ts
```diff
@@ -27,6 +27,11 @@ const MEGABYTES = 1000 * 1000
 
 export type ConfigOptions = {
   bootstrapNodes: string[]
+  /**
+   * STUN servers to use for inititating WebRTC connections.
+   * TODO: Use our own STUN servers
+   */
```
NullSoldier, line 33: I think we can delete this TODO here.

ironfish/src/network/peers/connections/webRtcConnection.ts
```diff
@@ -59,7 +59,7 @@ export class WebRtcConnection extends Connection {
     initiator: boolean,
     logger: Logger,
     metrics?: MetricsMonitor,
-    options: { simulateLatency?: number } = {},
+    options: { simulateLatency?: number; stunServers: string[] } = { stunServers: [] },
```
NullSoldier, line 62: For this we should just make it optional. You shouldn't materialize these defaults unless you actually need the real value.

ironfish/src/network/peers/connections/webRtcConnection.ts
```diff
@@ -73,9 +73,8 @@ export class WebRtcConnection extends Connection {
       this.addLatencyWrapper()
     }
 
-    // TODO: Use our own STUN servers
     this.peer = new nodeDataChannel.PeerConnection('peer', {
-      iceServers: ['stun:stun.l.google.com:19302', 'stun:global.stun.twilio.com:3478'],
+      iceServers: options.stunServers,
```
NullSoldier, line 77:
```suggestion
      iceServers: options.stunServers ?? [],
```

ironfish/src/network/peers/connections/webRtcConnection.ts
danield9tqh: I thought providing a default value made it optional. I can change though

ironfish/src/fileStores/config.ts
danield9tqh: The user who mentioned this to me said they would be more comfortable with a projects' STUN server. We can discuss offline though about user perception

NullSoldier: I would keep these todos to actual technical work that needs to be done rather than future product decisions. Like, this is more of a ticket in linear than a TODO.

ironfish/src/network/peers/connections/webRtcConnection.ts
NullSoldier: yeah it does. This is just more of a nit because we don't do defaulting values like this in other places, we just materialize it where we use it, or in the constructor. To be clear, I think this code is fine, I'm just trying to massage it to be more inline with other code.

PR #3996, opened 2023-06-28
Return asset verification information from the RPC interface

andreacorbellini:
## Summary

The RPC endpoints `wallet/getAssets`, `wallet/getBalance`, and `wallet/getBalances` now return verified assets information.

## Testing Plan

Query the endpoints and check for the presence of either:
* `"verification": { "status": "unverified" }`
* `"assetVerification": { "status": "unverified" }`

Note that the native asset currently appears as "unverified". This is due to the fact that the API changes have not been deployed to `api.ironfish.network` yet. Once changes are deployed, the status will change to "verified". You may also change the Verification API URL to testnet.api.ironfish.network if you want to see "verified" in the output.

The verification status should NOT be "unknown", except very early at startup. After startup, it should always be either "verified" or "unverified".

```
$ curl -s http://localhost:8021/wallet/getAssets | jq 
{
  "data": {
    "createdTransactionHash": "4f484a5355a65be2179fc66d69e149788e78aec35c6db230b182264e3982bc15",
    "id": "51f33a2f14f92735e562dc658a5639279ddca3d5079a6d1242b2a588a9cbf44c",
    "metadata": "49726f6e2046697368204e6174697665204173736574",
    "name": "2449524f4e",
    "owner": "0000000000000000000000000000000000000000000000000000000000000000",
    "status": "pending",
    "verification": {
      "status": "unverified"
    }
  }
}
```

```
$ curl -s http://localhost:8021/wallet/getBalance | jq 
{
  "status": 200,
  "data": {
    "account": "default",
    "assetId": "51f33a2f14f92735e562dc658a5639279ddca3d5079a6d1242b2a588a9cbf44c",
    "assetVerification": {
      "status": "unverified"
    },
    "confirmed": "0",
    "unconfirmed": "0",
    "unconfirmedCount": 0,
    "pending": "0",
    "available": "0",
    "pendingCount": 0,
    "confirmations": 2,
    "blockHash": "000000000000006cae2b20f69aec1e49a7975f2b3cf87f5c23eb42e9145d2723",
    "sequence": 96523
  }
}
```

```
$ curl -s http://localhost:8021/wallet/getBalances | jq 
{
  "status": 200,
  "data": {
    "account": "default",
    "balances": [
      {
        "assetId": "51f33a2f14f92735e562dc658a5639279ddca3d5079a6d1242b2a588a9cbf44c",
        "assetName": "2449524f4e",
        "assetOwner": "49726f6e2046697368",
        "assetVerification": {
          "status": "unverified"
        },
        "blockHash": "000000000000008f614d79cd323669f91d87b6f58e09b458e6343c4f83b3174c",
        "confirmed": "0",
        "sequence": 96723,
        "unconfirmed": "0",
        "unconfirmedCount": 0,
        "pending": "0",
        "pendingCount": 0,
        "available": "0"
      }
    ]
  }
}
```

## Documentation

RPC documentation changes are required (coming soon).

## Breaking Change

- Old clients interacting with new nodes are not affected by this change (they will ignore the new field)
- New clients interacting with old nodes are affected by this change, because they will require the presence of the new field


ironfish/src/rpc/routes/wallet/getAssets.ts
```diff
@@ -38,6 +40,10 @@ export const GetAssetsResponseSchema: yup.ObjectSchema<GetAssetsResponse> = yup
     owner: yup.string().defined(),
     status: yup.string().defined(),
     supply: yup.string().optional(),
+    verification: yup
```
NullSoldier, line 43: I think this should be `yup.object({ status: yup.string() }).defined()` instead of using shape() which is used for dynamic types.

ironfish/src/rpc/routes/wallet/getBalances.test.ts
```diff
@@ -8,9 +8,9 @@ import { createRouteTest } from '../../../testUtilities/routeTest'
 describe('Route wallet/getBalances', () => {
   const routeTest = createRouteTest(true)
 
-  beforeAll(async () => {
-    await routeTest.node.wallet.createAccount('account', true)
-  })
+  //beforeAll(async () => {
+  //  await routeTest.node.wallet.createAccount('account', true)
+  //})
```
NullSoldier, line 13: 👆

ironfish/src/rpc/routes/wallet/getBalances.test.ts
```diff
@@ -85,18 +85,137 @@ describe('Route wallet/getBalances', () => {
...
+          assetName: mockBalances[1].assetName.toString('hex'),
+          assetOwner: mockBalances[1].assetOwner.toString('hex'),
+          assetVerification: { status: 'unknown' },
+          confirmed: mockBalances[1].confirmed.toString(),
+          unconfirmed: mockBalances[1].unconfirmed.toString(),
+          pending: mockBalances[1].pending.toString(),
+          available: mockBalances[1].available.toString(),
+        },
+      ])
+    })
```
NullSoldier, line 113: Can we revert this test? I'm not sure it needs to change. We can also use toMatchObject() which is better in jest because it means other tests don't need to change in the case that you want to test a new field. Otherwise, every new field has to change all tests that use toEqual() which is why it's not recommended to be used.

ironfish/src/rpc/routes/wallet/getBalances.test.ts
```diff
@@ -85,18 +85,137 @@ describe('Route wallet/getBalances', () => {
...
+          assetId: mockBalances[1].assetId.toString('hex'),
+          assetName: mockBalances[1].assetName.toString('hex'),
+          assetOwner: mockBalances[1].assetOwner.toString('hex'),
+          assetVerification: { status: 'verified' },
+          confirmed: mockBalances[1].confirmed.toString(),
+          unconfirmed: mockBalances[1].unconfirmed.toString(),
+          pending: mockBalances[1].pending.toString(),
+          available: mockBalances[1].available.toString(),
+        },
+      ])
```
NullSoldier, line 218: This is a lot of code for just testing asset verification, I think this actually can be reduced down to a few lines of code. I'm willing to pair on this to go through this test. It also duplicates the verification logic through mocking in a strange way.

andreacorbellini: Reduced the test code, and simplified the mock. I can also put the test preparation code in a common function to reduce the number of lines even further if you want.

NullSoldier: I think this is good for now. If I have any other recommendations I can just show it in a proposal patch later.

NullSoldier (approved):

NullSoldier:
I also tested this manually.

PR #4003, opened 2023-06-28
Set SDK node compatability to >=16

NullSoldier:
## Summary

We use node version 18, but our node-app is stuck using version 16. This just formalizes what we already know. We need to support at least version 16 for now in our rust and SDK package.

## Testing Plan

Run tests on Github CI.

### DIFF
--- a/ironfish-rust-nodejs/package.json
+++ b/ironfish-rust-nodejs/package.json
@@ -30,7 +30,7 @@
     }
   },
   "engines": {
-    "node": "18.x"
+    "node": ">=16"
   },
   "devDependencies": {
     "@napi-rs/cli": "2.14.3",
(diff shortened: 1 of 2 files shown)


mat-if (approved):
We'll need to update these
https://github.com/iron-fish/ironfish/blob/master/simulator/bin/run
https://github.com/iron-fish/ironfish/blob/master/ironfish-cli/bin/run
if we want to properly support multiple NodeJS versions, although doesn't necessarily need to happen here.

NullSoldier:
I discussed this with Mat. We'll keep those targeting node 18, but the SDK will be >16. The SDK will catch up once our oldest dependency upgrades which is the node-app.

PR #4005, opened 2023-06-29
implements bech32 account encoder

hughy:
## Summary

encodes account data as a buffer by concatenating fields together before encoding the hex string using bech32.

copies buffer encoding pattern from walletdb/accountValue

## Testing Plan

adds unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/account/encoder/bech32.ts
+++ b/ironfish/src/wallet/account/encoder/bech32.ts
@@ -4,35 +4,30 @@
 import { PUBLIC_ADDRESS_LENGTH } from '@ironfish/rust-nodejs'
 import bufio from 'bufio'
 import { Bech32m } from '../../../utils'
-import {
-  AccountImport,
-  KEY_LENGTH,
-  VERSION_LENGTH,
-  VIEW_KEY_LENGTH,
-} from '../../walletdb/accountValue'
+import { AccountImport, KEY_LENGTH, VIEW_KEY_LENGTH } from '../../walletdb/accountValue'
+import { ACCOUNT_SCHEMA_VERSION } from '../account'
 import { AccountEncoder } from './encoder'
 
-export const BECH32_ACCOUNT_PREFIX = 'ironfishaccount00000'
-
+export const BECH32_ACCOUNT_PREFIX = 'ifaccount'
 export class Bech32AccountEncoder implements AccountEncoder {
+  VERSION = 1
+
   encode(value: AccountImport): string {
     const bw = bufio.write(this.getSize(value))
-    bw.writeU16(value.version)
-
-    let flags = 0
-    flags |= Number(!!value.spendingKey) << 0
-    flags |= Number(!!value.createdAt) << 1
-    bw.writeU8(flags)
+    bw.writeU16(this.VERSION)
 
     bw.writeVarString(value.name, 'utf8')
-    if (value.spendingKey) {
-      bw.writeBytes(Buffer.from(value.spendingKey, 'hex'))
-    }
     bw.writeBytes(Buffer.from(value.viewKey, 'hex'))
     bw.writeBytes(Buffer.from(value.incomingViewKey, 'hex'))
     bw.writeBytes(Buffer.from(value.outgoingViewKey, 'hex'))
     bw.writeBytes(Buffer.from(value.publicAddress, 'hex'))
 
+    bw.writeU8(Number(!!value.spendingKey))
+    if (value.spendingKey) {
+      bw.writeBytes(Buffer.from(value.spendingKey, 'hex'))
+    }
+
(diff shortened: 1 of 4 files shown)


ironfish/src/wallet/account/encoder/bech32.ts
```diff
@@ -0,0 +1,114 @@
...
+  VIEW_KEY_LENGTH,
+} from '../../walletdb/accountValue'
+import { AccountEncoder } from './encoder'
+
+export const BECH32_ACCOUNT_PREFIX = 'ironfishaccount00000'
+
+export class Bech32AccountEncoder implements AccountEncoder {
+  encode(value: AccountImport): string {
+    const bw = bufio.write(this.getSize(value))
+    bw.writeU16(value.version)
```
NullSoldier, line 20: I don't think you want to use the account version, but you want to use the encoding version which I think should be hard coded. the DB schema can change independent of the exporter changing, but the version here shouldn't be incremeneted.

ironfish/src/wallet/account/encoder/bech32.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    let flags = 0
+    flags |= Number(!!value.spendingKey) << 0
+    flags |= Number(!!value.createdAt) << 1
+    bw.writeU8(flags)
```
NullSoldier, line 25: I'm considering a new way of optional behaviour which is just using a byte at the beginning of each item. I feel like we constantly run out of space in these flags. It's also easier to just see right before something is serialized, a 1 byte read. What do you think? It'll use extra bytes for each optional thing, but it might make our serialization code cleaner.

ironfish/src/wallet/account/encoder/bech32.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    bw.writeVarString(value.name, 'utf8')
+    if (value.spendingKey) {
+      bw.writeBytes(Buffer.from(value.spendingKey, 'hex'))
+    }
```
NullSoldier, line 30: can we move optional things to the end so the data / serialization looks stable in the case of a view wallet version of this? Then one is a superset of the other.

ironfish/src/wallet/account/encoder/bech32.ts
```diff
@@ -0,0 +1,114 @@
...
+import { Bech32m } from '../../../utils'
+import {
+  AccountImport,
+  KEY_LENGTH,
+  VERSION_LENGTH,
+  VIEW_KEY_LENGTH,
+} from '../../walletdb/accountValue'
+import { AccountEncoder } from './encoder'
+
+export const BECH32_ACCOUNT_PREFIX = 'ironfishaccount00000'
```
NullSoldier, line 15: If this is the new bech32 version, how about a smaller prefix? Bech32 also says a smaller prefix increases checksum quality.

ironfish/src/wallet/account/encoder/bech32.ts
```diff
@@ -0,0 +1,114 @@
...
+    if (hexEncoding === null) {
+      return null
+    }
+
+    try {
+      const buffer = Buffer.from(hexEncoding, 'hex')
+
+      const reader = bufio.read(buffer, true)
+
+      const version = reader.readU16()
```
NullSoldier, line 56: we probably want to bail here if this doesn't match our encoding version?

ironfish/src/wallet/account/encoder/bech32.ts
jowparks: Seems good to me, how about "ifaccount"?

ironfish/src/wallet/account/encoder/bech32.ts
jowparks: For optional stuff we would only need 1 bit, not a byte, but I like this idea. I thought the bitflags were kind of fragile.

jowparks (approved):

ironfish/src/wallet/walletdb/accountValue.ts
```diff
@@ -7,9 +7,9 @@ import { IDatabaseEncoding } from '../../storage'
 import { ACCOUNT_KEY_LENGTH } from '../account/account'
 import { HeadValue, NullableHeadValueEncoding } from './headValue'
 
-const KEY_LENGTH = ACCOUNT_KEY_LENGTH
+export const KEY_LENGTH = ACCOUNT_KEY_LENGTH
 export const VIEW_KEY_LENGTH = 64
-const VERSION_LENGTH = 2
+export const VERSION_LENGTH = 2
```
NullSoldier, line 12: nit: were not using this anymore

PR #4006, opened 2023-06-29
chore: Upgrade napi to latest version

mat-if:
## Summary

We've been using a fairly old napi-rs version for quite a while due to bugs in some of the newer versions. Try upgrading to the latest version and see if we run into the same issues.

This would unblock the GUI app from upgrading to a newer version of Electron, so we can officially sunset NodeJS 16 support, since 16 is going to be entering maintenance mode soon and we want to stay current with the LTS releases.

## Testing Plan

- [x] Locally ran test, test:slow, test:perf
- [x] Stress-test with wallet rescans

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):
I tested this manually and it looks good to me.

PR #4008, opened 2023-06-29
Remove creating influx config in GH Workflow

ygao76:
## Summary
The influx config only need to be created once since its running on the same machine. Remove that command in Github workflow
Solving the error in GH workflow https://github.com/iron-fish/ironfish/actions/runs/5409999241/jobs/9830857314
`Error: failed to create config "ghaction": config "ghaction" already exists` 
## Testing Plan
Run GH workflow
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```
@mat-if @danield9tqh

### DIFF
--- a/.github/workflows/perf_test.yml
+++ b/.github/workflows/perf_test.yml
@@ -47,12 +47,7 @@ jobs:
 
       - name: Import to InfluxDB
         if: contains(github.event_name, 'push')
-        env:
-          INFLUX_TOKEN: ${{ secrets.INFLUX_TOKEN }}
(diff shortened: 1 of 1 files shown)


mat-if (approved):
This is good to get it unblocked 👍 . I am curious if there's a way to have it check if it needs to create the config. I'm thinking of a case where we reformat the self-hosted runner, or bring on a 2nd one or something.

ygao76:
Pass in the authentication credentials so the influx cli can be run without a config

PR #4011, opened 2023-06-29
feat(ifl-1233): create existing encoders

jowparks:
## Summary

Adds `Encoder`s for the existing account export types:
- spendingKey
- mnemonic
- JSON
- Bech32m JSON

These encoders throw instead of return null as this removes opinion about what should be done with returned encoder value. 

These will be used when we update the CLI code to import these encoders, rather than containing the code explicitly.

## Testing Plan
tests pass
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/account/encoder/mnemonic.ts
+++ b/ironfish/src/wallet/account/encoder/mnemonic.ts
@@ -1,0 +1,59 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import {
+  generateKeyFromPrivateKey,
+  LanguageCode,
+  spendingKeyToWords,
+  wordsToSpendingKey,
+} from '@ironfish/rust-nodejs'
+import { EncodingError } from 'bufio'
+import { LanguageUtils } from '../../../utils'
+import { AccountImport } from '../../walletdb/accountValue'
+import { ACCOUNT_SCHEMA_VERSION } from '../account'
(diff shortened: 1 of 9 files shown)


ironfish/src/wallet/account/encoder/mnemonic.ts
```diff
@@ -0,0 +1,57 @@
...
+import { EncodingError } from 'bufio'
+import { LanguageUtils } from '../../../utils'
+import { AccountImport } from '../../walletdb/accountValue'
+import { ACCOUNT_SCHEMA_VERSION } from '../account'
+import { AccountEncoder } from './encoder'
+
+export class MnemonicEncoder implements AccountEncoder {
+  encode(value: AccountImport): string {
+    if (!value.spendingKey) {
+      throw new EncodingError('Spending key is required for spending key encoder')
```
hughy, line 19: nit - `for mnemonic encoder`?

ironfish/src/wallet/walletdb/accountValue.ts
```diff
@@ -26,6 +27,7 @@ export interface AccountValue {
 export type AccountImport = Omit<AccountValue, 'id' | 'createdAt' | 'name'> & {
   createdAt: { hash: string; sequence: number } | null
   name?: string
+  language?: LanguageKey
```
hughy, line 30: It feels a little odd to have `language` be part of the account

Is this to keep the `encode` method signature the same for `MnemonicEncoder`? I'm wondering if we should add `AccountEncodingOptions` or something for that

jowparks: I agree that makes sense to me

hughy (approved):

PR #4012, opened 2023-06-29
feat(ifl-1233): account name optional

jowparks:
## Summary
Makes `AccountImport.name` optional, this will enable encoders/decoders to use this type. Related to:
https://github.com/iron-fish/ironfish/pull/4011
## Testing Plan
unit tests
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/importAccount.ts
+++ b/ironfish/src/rpc/routes/wallet/importAccount.ts
@@ -6,8 +6,11 @@ import * as yup from 'yup'
 import { AccountImport } from '../../../wallet/walletdb/accountValue'
 import { ApiNamespace, router } from '../router'
 
+export class ImportError extends Error {}
+
 export type ImportAccountRequest = {
   account: AccountImport
+  name?: string
   rescan?: boolean
 }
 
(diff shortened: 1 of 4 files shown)


ironfish/src/wallet/walletdb/accountValue.ts
```diff
@@ -23,8 +23,9 @@ export interface AccountValue {
   createdAt: HeadValue | null
 }
```
jowparks, line 25: We should probably consider splitting this type...

ironfish/src/rpc/routes/wallet/importAccount.ts
```diff
@@ -52,6 +56,11 @@ router.register<typeof ImportAccountRequestSchema, ImportResponse>(
   ImportAccountRequestSchema,
   async (request, node): Promise<void> => {
     let createdAt = null
```
jowparks, line 58: We will need to update this once we support string imports. We will need to check:
```
    const accountImport = AccountEncoder.decode(account)
    const name = accountImport.name || request.data.name
    if (!name) {
      throw new ImportError('Account name is required')
    }
```

ironfish/src/wallet/walletdb/accountValue.ts
hughy: Agreed; we'll be getting to the point where only ~half the fields from `AccountValue` will be the same in this type

hughy (approved):
Looks good pending lint!

PR #4024, opened 2023-07-05
feat(ci): Add a lint check for PRs against master

mat-if:
## Summary

This will fail if the PR has a base branch of master and does not contain the word master in the PR title. This still allows release PRs and other emergency PRs, while creating a quick and easy check to make sure PRs aren't accidentally going into master.

I wanted the override keyword to be "MASTER", but the github actions expressions treat strings as case-insensitive, so good enough for now.

## Testing Plan

Tested this on a small repo: https://github.com/mat-if/libuv-playground/actions/workflows/testing.yml

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/.github/workflows/check-pr-branch.yml
+++ b/.github/workflows/check-pr-branch.yml
@@ -1,4 +1,4 @@
-name: Disallow Pull Requests to Master Branch
+name: Pull Requests should target Staging branch
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/check-pr-branch.yml
```diff
@@ -0,0 +1,18 @@
+name: Disallow Pull Requests to Master Branch
```
dguenther, line 1: What do you think about making the name something that suggests the appropriate action? Like "Pull Requests should target Staging branch", so that hopefully people can self-diagnose without having to click through to the logs

dguenther (approved):

PR #4025, opened 2023-07-05
adds separate RpcAccountImport type

hughy:
## Summary

pushes serialization of createdAt as HeadValue to the RPC; serialization of the block hash as a hex string is only needed for sending RPC responses

makes createdAt type uniform across AccountImport and AccountValue

closes IFL-1259

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/import.ts
+++ b/ironfish-cli/src/commands/wallet/import.ts
@@ -9,7 +9,7 @@ import {
   LanguageUtils,
   PromiseUtils,
 } from '@ironfish/sdk'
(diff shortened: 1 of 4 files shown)


ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -48,9 +48,9 @@ export class ImportCommand extends IronfishCommand {
 
     const client = await this.sdk.connectRpc()
```
jowparks, line 50: We will be able to get rid of a lot (if not all) of this code once we update the RPC and delete this code from the CLI.

jowparks (approved):

PR #4027, opened 2023-07-05
Verified assets: use the persisted cache when the node is not running

andreacorbellini:
## Summary

When the command line client is running but not connected to any node, use the persisted cache (if present) to get and display verified assets information.

## Testing Plan

- Start the node so that it can fetch the verified assets information and write the cache to disk (this requires https://github.com/iron-fish/ironfish/pull/4020)
- Stop the node
- Run `ironfish wallet:balances`, or similar commands

Expected behavior: verified assets should appear as green with a checkmark.

## Documentation

No doc change needed.

## Breaking Change

Not a breaking change.

### DIFF
--- a/ironfish/src/assets/assetsVerifier.test.ts
+++ b/ironfish/src/assets/assetsVerifier.test.ts
@@ -145,30 +145,6 @@ describe('AssetsVerifier', () => {
       expect(assetsVerifier.verify('4567')).toStrictEqual({ status: 'unknown' })
     })
 
-    it("returns 'unknown' after being stopped", async () => {
-      nock('https://test')
-        .get('/assets/verified')
-        .reply(200, {
-          data: [{ identifier: '0123' }],
-        })
-
-      const assetsVerifier = new AssetsVerifier({
-        apiUrl: 'https://test/assets/verified',
-      })
-      const refresh = jest.spyOn(assetsVerifier as any, 'refresh')
-
-      assetsVerifier.start()
-      await waitForRefreshToFinish(refresh)
-
-      expect(assetsVerifier.verify('0123')).toStrictEqual({ status: 'verified' })
-      expect(assetsVerifier.verify('4567')).toStrictEqual({ status: 'unverified' })
-
-      assetsVerifier.stop()
-
-      expect(assetsVerifier.verify('0123')).toStrictEqual({ status: 'unknown' })
-      expect(assetsVerifier.verify('4567')).toStrictEqual({ status: 'unknown' })
-    })
-
     it("returns 'unknown' when the API is unreachable", async () => {
       nock('https://test').get('/assets/verified').reply(500)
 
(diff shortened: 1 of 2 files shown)


ironfish/src/sdk.ts
```diff
@@ -242,6 +242,10 @@ export class IronfishSdk {
       }
     }
 
+    if (this.config.get('enableAssetVerification')) {
+      node.assetsVerifier.start({ enableRefresh: false })
+    }
```
NullSoldier, line 247: It looks like this should still happen even if the person isn't using the SDK. I would recommend starting all sub-systems in the node rather than here.

NullSoldier:
I need more of an explanation on this one because with this change, we're starting to diverge our system design into a new pattern.  Particularly because you should be able to do this without this change. In other systems, here's the flow that would make this unneccessary.

1. wallet:balances runs
2. SDK mounts and MemoryAdapter to the node
3. SDK opens the node and all DB connectiosn
4. open() is the lifecycle where all data is loaded needed for the system to run while start() is the life cycle method where all systems start
5. the keystore should have been loaded during open() allowing any calls to it in the CLI / RPC to get the data it needs
6. start should not need to be called in the SDK to make this work in an in memory adapter

ironfish/src/sdk.ts
andreacorbellini: This system is already started when you call the `IronfishNode.start()` method, and it's started with `enableRefresh: true` (the default).

I interpreted this method here (`IronfishSdk.node()`) to be a "light" version of the node, therefore I put here the lines of code to start a lightweight assets verifier. Is my understanding wrong? In order to put this code in the `IronfishNode`, I would need to put it in the `init()` method. This would be the only service that is started at `init()`. To make things more clean, perhaps I can add a `IronfishNode.lightweightStart()` (or similar) and call it here. Thoughts about that?

andreacorbellini:
> I need more of an explanation on this one because with this change, we're starting to diverge our system design into a new pattern. Particularly because we should be able to do this without this change. In other systems, here's the flow that would make this unnecessary.
> 
>     1. wallet:balances runs
> 
>     2. SDK mounts and MemoryAdapter to the node
> 
>     3. SDK opens the node and all DB connections
> 
>     4. open() is the lifecycle where all data is loaded needed for the system to run while start() is the life cycle method where all systems start
> 
>     5. the KeyStore should have been loaded during open() allowing any calls to it in the CLI / RPC to get the data it needs
> 
>     6. start should not need to be called in the SDK to make this work in an in memory adapter

Ok, I understand where you're coming from. You're right that the AssetsVerifier has all the data it needs--calling `start()` is not necessary for it to gather its data. With the current implementation, `start()` is necessary to inform the AssetsVerifier that asset verification has been enabled in the configuration. Without a call to `start()`, the verifier will return "unknown" for all assets to respect the user choice.

So, `start()` does two things currently: it enables the AssetsVerifier to give "verified"/"unverified" responses (rather than "unknown"), and it starts the refresh loop.

In order to fit the pattern you describe, perhaps we should split these two functionalities into two methods: `enable()` to make it give responses (called after the node is initialized), and `start()` just to start the refresh loop (called after the node is started).

Does that sound acceptable? Ignore my comment from before.

NullSoldier:
Here's a question, what does enable really mean? Does it mean it can give answers? I think it should always be able to give answers while enabling the system controls wether it calls home to update it's list.

That would mean that once it fetches the data, as soon as it `opens()` it's capable of giving answers even offline because it's loaded the cache into memory and can use that to respond to verified queries from other systems

andreacorbellini:
> I think it should always be able to give answers while enabling the system controls wether it calls home to update it's list.

... except if the user has specified `enableAssetVerification` to `false`, do you agree on that?

I pushed an proposed implementation that allows finer control over the periodic refresh, and enables verification as soon as possible (if the user configuration allows it). Let me know what you think!

NullSoldier:
I know it was in the spec, but I don't think `enableAssetVerification` being false needs to disable the whole system. The system can exist and run, but the main reason to disable it is because as a user I don't want a system calling home to some remote HTTP API for privacy reasons not because I don't want the system to run.

NullSoldier (approved):

andreacorbellini:
Synced with Jason: will change the code so that the AssetVerifier will provide verification information if the cache is non-empty, regardless of the value of `enableAssetVerification`. The value of `enableAssetVerification` will only control whether the periodic refreshes happen or not.

PR #4028, opened 2023-07-05
feat(ifl-1237): encodeAccount/decodeAccount generalized functions

jowparks:
## Summary
Adds `encodeAccount/decodeAccount` which can be used in place of the individual encoders to:
1. Always use recommended encoder from SDK
2. Try all decoders on string instead of trying each encoder
## Testing Plan
tests pass
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/account/encoder/bech32.test.ts
+++ b/ironfish/src/wallet/account/encoder/bech32.test.ts
@@ -5,11 +5,11 @@ import { generateKey } from '@ironfish/rust-nodejs'
 import { Bech32m } from '../../../utils'
 import { AccountImport } from '../../walletdb/accountValue'
 import { ACCOUNT_SCHEMA_VERSION } from '../account'
-import { BECH32_ACCOUNT_PREFIX, Bech32AccountEncoder } from './bech32'
+import { BECH32_ACCOUNT_PREFIX, Bech32Encoder } from './bech32'
 
 describe('Bech32AccountEncoder', () => {
   const key = generateKey()
-  const encoder = new Bech32AccountEncoder()
+  const encoder = new Bech32Encoder()
 
   it('encodes the account as a bech32 string and decodes the string', () => {
     const accountImport: AccountImport = {
--- a/ironfish/src/wallet/account/encoder/json.ts
+++ b/ironfish/src/wallet/account/encoder/json.ts
@@ -1,19 +1,36 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { RpcAccountImport } from '../../../rpc/routes/wallet/types'
 import { validateAccount } from '../../validator'
-import { AccountImport, AccountValue } from '../../walletdb/accountValue'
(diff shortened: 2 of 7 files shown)


ironfish/src/wallet/account/encoder/generic.ts
```diff
@@ -0,0 +1,55 @@
...
+  AccountDecodingOptions,
+  AccountEncoder,
+  AccountEncodingOptions,
+  Formats,
+} from './encoder'
+import { JsonEncoder } from './json'
+import { MnemonicEncoder } from './mnemonic'
+import { SpendingKeyEncoder } from './spendingKey'
+
+export class GenericEncoder implements AccountEncoder {
```
NullSoldier, line 17: This seems ok, but is this better than Oreoscans importAccount() function? Does this make sense to be a class instead of just 2 functions? https://github.com/oreoslabs/oreos/blob/master/src/wallet/account.ts#L40-L53

jowparks: Thats fine with me, we thought it was convenient that this matches the class architecture of the other encoders, but I am not married to the generic encoder being a class.

ironfish/src/wallet/account/encoder/bech32.test.ts
```diff
@@ -5,11 +5,11 @@ import { generateKey } from '@ironfish/rust-nodejs'
 import { Bech32m } from '../../../utils'
 import { AccountImport } from '../../walletdb/accountValue'
 import { ACCOUNT_SCHEMA_VERSION } from '../account'
-import { BECH32_ACCOUNT_PREFIX, Bech32AccountEncoder } from './bech32'
+import { BECH32_ACCOUNT_PREFIX, Bech32Encoder } from './bech32'
 
 describe('Bech32AccountEncoder', () => {
   const key = generateKey()
```
jowparks, line 11: Renaming to match the nomenclature of the other encoders

ironfish/src/wallet/account/encoder/json.ts
```diff
@@ -1,19 +1,36 @@
...
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { RpcAccountImport } from '../../../rpc/routes/wallet/types'
 import { validateAccount } from '../../validator'
-import { AccountImport, AccountValue } from '../../walletdb/accountValue'
+import { AccountImport } from '../../walletdb/accountValue'
 import { AccountEncoder } from './encoder'
 
 export class JsonEncoder implements AccountEncoder {
   encode(value: AccountImport): string {
```
jowparks, line 10: Fixing serialization issue

ironfish/src/wallet/account/encoder/mnemonic.ts
```diff
@@ -14,7 +14,7 @@ import { ACCOUNT_SCHEMA_VERSION } from '../account'
 import { AccountDecodingOptions, AccountEncoder, AccountEncodingOptions } from './encoder'
 
 export class MnemonicEncoder implements AccountEncoder {
```
jowparks, line 16: `options` is required for the `MnemonicEncoder` since `language` is specifically related to the mnemonic.

PR #4029, opened 2023-07-05
fix(ci): Base branch check should update to green if base changes

mat-if:
## Summary

This wasn't changing back to green when the base branch changes because it would only run if the base branch is main. Therefore, we'll need to run it on all PRs, and check if the branch is master alongside the keyword in the job if-expression.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/.github/workflows/check-pr-branch.yml
+++ b/.github/workflows/check-pr-branch.yml
@@ -4,14 +4,13 @@ on:
   pull_request:
     types: ['*']
     branches:
-      - 'master'
+      - '**'
 
 jobs:
-  check-branch-keyword:
-    name: Check Base Branch And Override Keyword
+  check:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


NullSoldier (approved):

.github/workflows/check-pr-branch.yml
```diff
@@ -4,14 +4,13 @@ on:
   pull_request:
     types: ['*']
     branches:
-      - 'master'
+      - '**'
```
NullSoldier, line 7: Is this better than limiting this to only run on PRs into master?

mat-if: Yeah, that is the problem - it wasn't re-running when the base branch changed away from master, leading to it being permanently red. See: https://github.com/iron-fish/ironfish/pull/4027

PR #4030, opened 2023-07-05
feat(ifl-1234): account import rpc updates

jowparks:
## Summary
Uses the `decodeAccount` changes that handles string types in order to clean up the RPC.

For backwards compatibility, we still support an `RpcAccountImport` type input for `account`, but moving forward we should use the `string` type instead.

## Testing Plan
tests pass
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/account/encoder/generic.ts
```diff
@@ -0,0 +1,58 @@
...
+    for (const encoder of this.versions) {
+      try {
+        decoded = new encoder().decode(value, options)
+      } catch (e) {
+        decoderErrors.push([encoder.name, e as Error])
+        continue
+      }
+      return decoded
+    }
+    const errors = decoderErrors.map(([str, err]) => `${str}: ${err.message}`)
```
jowparks, line 53: Had to add this, otherwise the error could be very cryptic. IE if the account is input is `mnemonic` and they didn't provide `name` value to RPC, this would fail cryptically even though the mnemonic value might be correct.

ironfish/src/wallet/account/encoder/mnemonic.ts
```diff
@@ -42,7 +41,7 @@ export class MnemonicEncoder implements AccountEncoder {
       language = LanguageUtils.languageCodeToKey(code)
     }
     if (language === null) {
```
jowparks, line 43: These are not `bufio` errors, casting to standard `Error`

ironfish/src/wallet/account/encoder/bech32json.ts
```diff
@@ -1,23 +1,23 @@
...
-import { EncodingError } from 'bufio'
 import { Bech32m } from '../../../utils'
 import { AccountImport } from '../../walletdb/accountValue'
 import { AccountEncoder } from './encoder'
+import { JsonEncoder } from './json'
 export class Bech32JsonEncoder implements AccountEncoder {
   /**
    * @deprecated Bech32 JSON encoding is deprecated. Use the newest version of the Bech32JSONEncoder.
    */
   encode(value: AccountImport): string {
```
jowparks, line 12: Same bug that I found for the standard `JsonEncoder`.

hughy (approved):
lgtm

The only thing I'm still on the fence about is whether to change the type of the `account` field or introduce a new field (here and in the export RPC). I think it's a bit cleaner to have only a single field, but I could be convinced otherwise

jowparks:
> lgtm
> 
> The only thing I'm still on the fence about is whether to change the type of the `account` field or introduce a new field (here and in the export RPC). I think it's a bit cleaner to have only a single field, but I could be convinced otherwise

For the record, I agree. My vote is having the same field.

PR #4031, opened 2023-07-06
updates wallet exportAccount RPC to use encoders

hughy:
## Summary

adds 'format' and 'language' fields to requests to support encoding accounts on the node side instead of the client side

updates 'account' response to return either an account object (for backwards compatibility) or an encoded string

returns account object by default if no format is passed

updates export RPC tests

updates export CLI to accommodate updated response type

## Testing Plan

adds unit tests

## Documentation

```
[X] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/wallet/export.ts
```diff
@@ -65,7 +73,21 @@ export class ExportCommand extends IronfishCommand {
...
+    )
+
+    if (error) {
+      throw error
+    }
+    Assert.isNotNull(account)
+
     let output
 
     if (flags.mnemonic) {
```
jowparks, line 93: We can get rid of all this stuff, it is handled by the `MnemonicEncoder` now.

jowparks:
These conditionals don't need to be there now, for example the `bech32` is already encoded.
https://github.com/iron-fish/ironfish/pull/4031/files#diff-03518d0bddb12c46b598fd1755ac6fd466a19695362161a07b3358cacabc3964R134

jowparks:
Discussed in slack, there will be more changes to CLI in a subsequent PR. Changes to the CLI here are only to make CLI compatible with updated RPC.

jowparks (approved):

PR #4036, opened 2023-07-06
feat(cli,ironfish): Encapsulate blockchain database interactions

rohanjadvani:
## Summary

Introduces a `BlockchainDB` component (similar to [`WalletDB`](https://github.com/iron-fish/ironfish/blob/master/ironfish/src/wallet/walletdb/walletdb.ts)) to encapsulate database interactions.

This is mostly plumbing reads/writes/updates/deletions into a new file, but a few notes:
* I opted for `undefined` instead of `null` if LevelDB returned that type by default. Callers can coerce to `null` if they want a defined variable without a value. The `WalletDB` does this too.
* Database transactions are still initiated from the `db`, but that is done for the `WalletDB` as well. I think we can improve that, but that's a separate change.

## Testing Plan

Existing unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


rohanjadvani:
I didn't realize how large this diff became, so will close and chunk it up.

PR #4039, opened 2023-07-06
fix: import failure

jowparks:
## Summary
Build was failing due to export. 

- Fixes exports
- Renames format so it is more explicit for a generic export
## Testing Plan
build passes
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/account/index.ts
```diff
@@ -2,4 +2,4 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 export * from './account'
-export { Format } from './encoder/encoder'
+export { AccountFormat as Format } from './encoder/encoder'
```
hughy, line 5: I think we can remove this file altogether with the changes in `wallet/index.ts`

hughy (approved):

PR #4041, opened 2023-07-07
Verified assets: use a retry strategy when refreshing

andreacorbellini:
## Summary

If fetching the assets fails (because e.g. of a network glitch), do not wait until the next refresh interval, but retry sooner, with an exponential backoff strategy. The delays between each retry are:

- 1st retry: 1 minute ± 20%
- 2nd retry: 2 minutes ± 20%
- 3rd retry: 4 minutes ± 20%
- 4rd retry: 8 minutes ± 20%
- ...

Up to a maximum of 1 hour.

The reason why the delays are so high (as opposed to be in the order of seconds or milliseconds) is that the default refresh interval is 6 hours, so there's no need to be too aggressive.

This PR also upgrades to the version of jest, which is needed to use `jest.advanceTimersToNextTimerAsync()`.

## Testing Plan

Simulate some network failures (e.g. by faking `api.ironfish.network` via `/etc/hosts`) and run the node with `--verbose`, observing that the node will try to refresh the list of verified assets more frequently than once every 6 hours.

```
ironfish start --verbose |& grep assets
```

## Documentation

No doc change required

## Breaking Change

Not a breaking change


ironfish/src/assets/assetsVerifier.ts
```diff
@@ -80,6 +86,7 @@ export class AssetsVerifier {
       }
     } catch (error) {
       this.logger.warn(`Error while fetching verified assets: ${ErrorUtils.renderError(error)}`)
+      throw error
```
ygao76, line 89: What type of error could happen here, besides network connection issue?  It will be good if we only retry when the error is from network issue or other retry-able error.

andreacorbellini: This could be a network error, or a server error (e.g. a 500 or 521 response), or an invalid/truncated JSON payload. The HTTP method used here (GET) is both [safe](https://www.rfc-editor.org/rfc/rfc7231#section-4.2.1) and [idempotent](https://www.rfc-editor.org/rfc/rfc7231#section-4.2.2), so this kind of request is always retry-able, regardless of the kind of error returned.

ygao76: For the internal server error 500, would it still be failing when calling `refreshVerifiedAssets` when retry again?

andreacorbellini: The next call to `refreshVerifiedAssets` may or may not fail: it all depends on whether the remote API has recovered or not.

ygao76 (approved):
LGTM, can you monitor the load of `refreshVerifiedAssets` after release. I am curious about the extra load of that due to retry and the success rate of retry. If we see a lot of retries where it keeps failing until it reaches the 1 hr limit, maybe we could work on improving the retry success rate.

PR #4064, opened 2023-07-13
adds standalone wallet config options

hughy:
## Summary

running a standalone wallet requires config options for the wallet's RPC connection to a node, for the wallet's RPC server, and for the wallet's worker pool

adds config options to the Config store and sets default values:
- sets default options for wallet node RPC connections to mirror node RPC server defaults
- sets defaults for wallet RPC server to match defaults for node RPC server
- uses separate default IPC filepath for wallet RPC server
- uses separate default TLS files for wallet RPC server
- uses distinct port for each TCP and HTTP server
- uses same defaults for wallet worker pool as node worker pool

## Testing Plan

config options are defined, but not yet used

## Documentation

```
[X] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/fileStores/config.ts
+++ b/ironfish/src/fileStores/config.ts
@@ -283,7 +283,7 @@ export type ConfigOptions = {
   /**
    * Enable the node's in-process wallet to scan blocks and mempool transactions
    */
-  enableWalletScanning: boolean
+  walletScanningEnabled: boolean
 
   /**
    * Enable standalone wallet process to connect to a node via IPC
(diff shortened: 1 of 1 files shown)


ironfish/src/fileStores/config.ts
```diff
@@ -279,6 +279,66 @@ export type ConfigOptions = {
    * Always allow incoming connections from these IPs even if the node is at maxPeers
    */
   incomingWebSocketWhitelist: string[]
+
+  /**
+   * Enable the node's in-process wallet to scan blocks and mempool transactions
+   */
+  enableWalletScanning: boolean
```
rohanjadvani, line 286: nit: I like how the other flags are prefixed with `wallet` followed by the config, so maybe something like `walletScanningEnabled` would be more consistent.

rohanjadvani (approved):

rohanjadvani (approved):

PR #4069, opened 2023-07-14
expires wallet transactions on block connect

hughy:
## Summary

moves expireTransactions from the eventLoop to the chainProcessor.onAdd handler function so that the wallet only expires transactions when a block is added to the chain

refactors expireTransactions to take a block sequence instead of reading a header from the chain thereby removing a chain dependency

## Testing Plan

updates unit tests and regenerates fixtures

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/__fixtures__/wallet.test.ts.fixture
+++ b/ironfish/src/wallet/__fixtures__/wallet.test.ts.fixture
@@ -1521,274 +1521,6 @@
       }
     }
   ],
-  "Accounts expireTransactions should not expire transactions with expiration sequence ahead of the chain": [
-    {
-      "version": 2,
-      "id": "cd8a190f-05bf-49fd-88ba-f722ef57ec1e",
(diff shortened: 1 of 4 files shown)


danield9tqh:
Do transactions ever get un-expired on disconnect events? It seems like very edge case scenario but maybe worth thinking about for the future

danield9tqh (approved):

rohanjadvani (approved):

hughy:
> Do transactions ever get un-expired on disconnect events? It seems like very edge case scenario but maybe worth thinking about for the future

It is possible! If the transaction un-expires and is included in a block during a reorg then it will be added from the block as if it were never seen in a pending state. That could still leave your wallet in a state where transactions are temporarily incorrectly marked as expired

I'm going to think about un-expiring transactions as a separate issue 😅

PR #4070, opened 2023-07-14
Allow Router and Server to be created without a node object

danield9tqh:
## Summary
To enable the RPC layer to be re-used with a standalone wallet, it needs to be able to be intsantiated without a node. Removing direct dependency on a node object here.

## Testing Plan
Unit testing / local testing

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/server.ts
+++ b/ironfish/src/rpc/server.ts
@@ -11,15 +11,12 @@ export class RpcServer {
   readonly node: IronfishNode
   readonly adapters: IRpcAdapter[] = []
 
-  private readonly router: Router
   private _isRunning = false
   private _startPromise: Promise<unknown> | null = null
   logger: Logger
 
   constructor(node: IronfishNode, logger: Logger = createRootLogger()) {
     this.node = node
-    this.router = router
-    this.router.server = this
     this.logger = logger.withTag('rpcserver')
   }
 
@@ -29,7 +26,10 @@ export class RpcServer {
 
   /** Creates a new router from this RpcServer with the attached routes filtered by namespaces */
   getRouter(namespaces: ApiNamespace[]): Router {
-    return this.router.filter(namespaces)
+    const newRouter = router.filter(namespaces)
+    newRouter.server = this
+    newRouter.context = { node: this.node }
+    return newRouter
   }
 
   /** Starts the RPC server and tells any attached adapters to starts serving requests to the routing layer */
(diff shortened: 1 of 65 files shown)


ironfish/src/rpc/server.ts
```diff
@@ -29,7 +32,10 @@ export class RpcServer {
 
   /** Creates a new router from this RpcServer with the attached routes filtered by namespaces */
   getRouter(namespaces: ApiNamespace[]): Router {
-    return this.router.filter(namespaces)
+    const newRouter = router.filter(namespaces)
+    newRouter.server = this
```
NullSoldier, line 36: Is this already in router.filter()?

ironfish/src/rpc/server.ts
```diff
@@ -2,24 +2,27 @@
...
   readonly adapters: IRpcAdapter[] = []
 
-  private readonly router: Router
   private _isRunning = false
   private _startPromise: Promise<unknown> | null = null
   logger: Logger
 
-  constructor(node: IronfishNode, logger: Logger = createRootLogger()) {
-    this.node = node
-    this.router = router
```
NullSoldier, line 21: If you pass a router in here now, and then try to get the server from the router the server will be incorrectly null.

ironfish/src/rpc/server.ts
```diff
@@ -2,24 +2,27 @@
...
-    this.router.server = this
+  constructor(
```
NullSoldier, line 19: I think the biggest change architectually here is you're now tying this to the root router in the router file. I guess that's ok, I just prefer not to have any module directly ever depend on a module level global since they are different concepts. I would prefer to still pass this in. What if I want to spawn a new RPC server from a non root level router that's already been filtered?

NullSoldier (approved):

NullSoldier:
My comment isn't a deal breaker, but I did write a comment that I don't see it necessary to hard code the RPC Server to be forced to use the root router.

ironfish/src/rpc/server.ts
danield9tqh: Not sure I understand. The previous code was setting `server.router` to the root router. This code is just doing the same thing. I agree we could separate it more and not depend on the root router but I don't see how this is changing anything from current code.

hughy: I might be misunderstanding, but it looks like they are already explicitly tied together -- the server sets its router property to the root router in the constructor

PR #4071, opened 2023-07-14
Separate global routes from Router

danield9tqh:
## Summary
We register routes globally but don't necesarily want a global Router object. The Router has a reference to the Server which should not be on a global object

## Testing Plan
Unit tests / local testing

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/router.ts
+++ b/ironfish/src/rpc/routes/router.ts
@@ -52,9 +52,58 @@ export function parseRoute(
 }
 
 export class Router {
+  routes = new Routes()
+  server: RpcServer
+
+  constructor(routes: Routes, server: RpcServer) {
+    this.routes = routes
+    this.server = server
+  }
+
+  async route(route: string, request: RpcRequest): Promise<void> {
+    const [namespace, method] = route.split('/')
+
+    const methodRoute = this.routes.get(namespace, method)
+    if (!methodRoute) {
+      throw new RouteNotFoundError(route, namespace, method)
+    }
+
+    const { handler, schema } = methodRoute
+
+    const { result, error } = await YupUtils.tryValidate(schema, request.data)
+    if (error) {
+      throw new ValidationError(error.message, 400)
+    }
+    request.data = result
+
+    try {
+      await handler(request, this.server.context)
+    } catch (e: unknown) {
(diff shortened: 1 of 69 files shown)


ironfish/src/rpc/routes/router.ts
```diff
@@ -52,42 +52,18 @@ export function parseRoute(
...
-    Assert.isNotUndefined(namespace, `Invalid namespace: ${String(namespace)}: ${route}`)
-    Assert.isNotUndefined(method, `Invalid method: ${String(namespace)}: ${route}`)
-
-    let namespaceRoutes = this.routes.get(namespace)
-
-    if (!namespaceRoutes) {
-      namespaceRoutes = new Map<string, { handler: RouteHandler; schema: YupSchema }>()
-      this.routes.set(namespace, namespaceRoutes)
-    }
+  routes = new Routes()
```
NullSoldier, line 55: this gets over-ridden in the constructor you can avoid initializing this and it works right

ironfish/src/rpc/routes/router.ts
```diff
@@ -52,42 +52,18 @@ export function parseRoute(
...
+  server: RpcServer
 
-    namespaceRoutes.set(method, {
-      handler: handler as RouteHandler<unknown, unknown>,
-      schema: requestSchema,
-    })
+  constructor(routes: Routes, server: RpcServer) {
+    this.routes = routes
```
NullSoldier, line 59: should this be a copy? Or else any modifications to this router modifies the root router permanently. Like in the tests in these files.

ironfish/src/rpc/routes/router.ts
danield9tqh: Yea good call

ironfish/src/rpc/routes/router.ts
danield9tqh: `Router` is only instantiated in one place right now (Server) and it uses filter before passing in the routes. It could be good for safety but prob not needed yet unless we start doing more complicated things with the Router. Just gonna assume whoever is instantiating it thought of that

NullSoldier (approved):

PR #4077, opened 2023-07-18
feat(ironfish): Remove node from peer network

rohanjadvani:
## Summary

For the standalone wallet, we need to support broadcasting transactions from the peer network *without* a dependency on the node. This change removes the node out of the peer network to remove a circular dependency.

Changes:
* Moves the syncer inside the peer network to avoid circular dependency 
* Pass the mining manager, worker pool, mempool directly into peer network instead of using node properties
* *Temporarily* pass wallet callback functions to the peer network
  * Add pending transaction from new messages will be removed
  * Broadcast transactions will be referenced once a wallet node client is created (but this change needs to go in first)

## Testing Plan

Covered by unit tests. No new tests needed.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[x] Yes
```

This code change updates the constructor of the `PeerNetwork` in our SDK, so this is a breaking change. We will need to ship a new major version release whenever this change is deployed.

### DIFF
--- a/ironfish/src/node.ts
+++ b/ironfish/src/node.ts
@@ -27,9 +27,9 @@ import { IsomorphicWebSocketConstructor } from './network/types'
 import { getNetworkDefinition } from './networkDefinition'
 import { Package } from './package'
 import { Platform } from './platform'
+import { Transaction } from './primitives'
 import { RpcServer } from './rpc/server'
 import { Strategy } from './strategy'
-import { Syncer } from './syncer'
 import { Telemetry } from './telemetry/telemetry'
 import { Wallet, WalletDB } from './wallet'
 import { WorkerPool } from './workerPool'
@@ -49,7 +49,6 @@ export class IronfishNode {
   files: FileSystem
   rpc: RpcServer
   peerNetwork: PeerNetwork
-  syncer: Syncer
   pkg: Package
   telemetry: Telemetry
   assetsVerifier: AssetsVerifier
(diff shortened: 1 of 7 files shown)


ironfish/src/node.ts
```diff
@@ -145,13 +144,22 @@ export class IronfishNode {
...
       logger: logger,
       telemetry: this.telemetry,
       incomingWebSocketWhitelist: config.getArray('incomingWebSocketWhitelist'),
+      miningManager: this.miningManager,
+      blocksPerMessage: config.get('blocksPerMessage'),
+      memPool,
+      workerPool,
+      // TODO(hugh, rohan): Clean up after the node clients are created for the wallet
+      // `walletAddPendingTransaction` will be removed once we can pull from the mempool
+      walletAddPendingTransaction: (transaction: Transaction) =>
```
NullSoldier, line 159: We should use events instead of passing lambdas into the constructor.

ironfish/src/node.ts
```diff
@@ -49,7 +49,6 @@ export class IronfishNode {
   files: FileSystem
   rpc: RpcServer
   peerNetwork: PeerNetwork
-  syncer: Syncer
```
NullSoldier, line 52: The syncer is a top level concept to me, so it's weird for me to see this being demoted.

ironfish/src/node.ts
rohanjadvani: I agree, but as per the TODO this is a temporary measure to make the types work / keep this code change small. It's removed in the subsequent PR #4078

rohanjadvani:
We'll fix this later

PR #4078, opened 2023-07-18
feat(ironfish): Pass node dependencies directly to mining manager

rohanjadvani:
## Summary

We're referencing a lot of fields off the node which is leading to circular dependency issues when working on the standalone wallet. Pass in dependencies directly into the `MiningManager`

## Testing Plan

Covered by existing tests.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[x] Yes
```

The constructor of the mining manager is breaking. We need to do a major version bump when this is released.


rohanjadvani:
We'll fix this later

PR #4075, opened 2023-07-17
[asset-security] refactor: rename asset owner to creator

mat-if:
## Summary

Renames asset 'owner' to 'creator' in preparation for adding the separate owner field.

Closes IFL-1324

## Documentation

Does this change require any updates to the Iron Fish Docs (ex. [the RPC API
Reference](https://ironfish.network/docs/onboarding/rpc/chain))? If yes, link a
related documentation pull request for the website.

```
[ ] Yes
```

## Breaking Change

Is this a breaking change? If yes, add notes below on why this is breaking and
what additional work is required, if any.

```
[ ] Yes
```

PR #4079, opened 2023-07-18
[asset-security] feat: add asset owner field to chain DB

mat-if:
## Summary

**Relies on #4075**

This adds an owner field to the asset DB alongside the now-renamed creator field. This is to store the current owner of an asset, whenever that changes, so we have the current owner and can verify the proof properly with the correct owner.

Closes IFL-1325

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/blockchain.test.ts
+++ b/ironfish/src/blockchain/blockchain.test.ts
@@ -1077,6 +1077,7 @@ describe('Blockchain', () => {
           name: asset.name(),
           nonce: asset.nonce(),
           creator: asset.creator(),
+          owner: asset.creator(),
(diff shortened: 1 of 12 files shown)


andreacorbellini (approved):
Looks good to me. I'm assuming we don't need an index on the new `owner` field, as all the lookups will be done from the asset ID. Am I right?

mat-if:
> Looks good to me. I'm assuming we don't need an index on the new `owner` field, as all the lookups will be done from the asset ID. Am I right?

Yeah. If we end up needing an owner index, it shouldn't be very hard to add.

PR #4081, opened 2023-07-19
Remove the node from the memory client

danield9tqh:
## Summary
The Memory Client doesn't need to know about the node directly, it should only be given the ability to route requests with a given context. Just passing in a Router object satisfies that requirement

## Testing Plan
Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/clients/memoryClient.test.ts
```diff
@@ -18,10 +18,12 @@ describe('MemoryClient', () => {
...
+    const client = new RpcMemoryClient(
+      createRootLogger(),
+      (await sdk.node()).rpc.getRouter(ALL_API_NAMESPACES),
+    )
 
-    const allowedNamespaces = ALL_API_NAMESPACES
-    const loadedNamespaces = [...(client.router?.routes.routes.keys() || [])]
-    expect([...allowedNamespaces.values()].sort()).toMatchObject(loadedNamespaces.sort())
+    const loadedNamespaces = [...(client.router.routes.routes.keys() || [])]
+    expect([...ALL_API_NAMESPACES.values()].sort()).toMatchObject(loadedNamespaces.sort())
```
NullSoldier, line 27: This test should be deleted now, because it's now a test for the router itself which doesn't belong in something that just takes a router.

ironfish/src/sdk.test.ts
```diff
@@ -131,7 +131,7 @@ describe('IronfishSdk', () => {
 
         expect(openDb).toHaveBeenCalledTimes(1)
         expect(client).toBeInstanceOf(RpcMemoryClient)
-        expect((client as RpcMemoryClient)._node).toBe(node)
+        expect((client as RpcMemoryClient).router.server.context.node).toBe(node)
```
NullSoldier, line 134: We should delete this since this field is gone. It's just a router expect / test now.

danield9tqh: It doesn't have anything to do with memory client but it is related to SDK which I think is what this is testing. It's testing that the SDK does actually attach the correct node

ironfish/src/rpc/clients/memoryClient.test.ts
danield9tqh: Yea I'll remove this

ironfish/src/rpc/clients/memoryClient.ts
```diff
@@ -3,30 +3,16 @@
...
-import { IronfishNode } from '../../node'
 import { MemoryResponse, RpcMemoryAdapter } from '../adapters'
-import { ALL_API_NAMESPACES, Router } from '../routes'
+import { Router } from '../routes'
 import { RpcClient } from './client'
 
 export class RpcMemoryClient extends RpcClient {
-  _node: IronfishNode | null
-  router: Router | null
+  router?: Router
```
NullSoldier, line 11: make this nullable

NullSoldier (approved):

PR #4084, opened 2023-07-19
feat(ironfish): Check if chain has block from node client

rohanjadvani:
## Summary

Encapsulates if a chain has a block with a call using the node client

## Testing Plan

Covered by existing tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/utils/error.ts
+++ b/ironfish/src/utils/error.ts
@@ -56,6 +56,10 @@ function isNodeError(error: unknown): error is Error & { code: string } {
   return error instanceof Error && 'code' in error && typeof error['code'] === 'string'
 }
 
+function isNotFoundError(error: unknown): error is Error & { code: 'not-found' } {
+  return isNodeError(error) && error.code === 'not-found'
+}
+
 export const ErrorUtils = {
   renderError,
   isConnectRefusedError,
@@ -63,4 +67,5 @@ export const ErrorUtils = {
   isConnectTimeOutError,
   isNoEntityError,
   isNodeError,
+  isNotFoundError,
 }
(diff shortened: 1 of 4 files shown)


ironfish/src/wallet/wallet.ts
```diff
@@ -1584,6 +1584,19 @@ export class Wallet {
...
+  private async chainHasBlock(hash: Buffer): Promise<boolean> {
+    try {
+      await this.nodeClient.chain.getBlock({ hash: hash.toString('hex') })
+      return true
+    } catch (error) {
+      if (error instanceof Error) {
+        this.logger.error(error.message)
+      }
+
+      return false
```
danield9tqh, line 1597: Should we return false here? If this errors out (one remote node for example) it will be as if the block doesn't exist in the chain. Particularly bad for the condition on `start` where if this returns false we reset the account

ironfish/src/utils/error.ts
```diff
@@ -56,11 +56,16 @@ function isNodeError(error: unknown): error is Error & { code: string } {
   return error instanceof Error && 'code' in error && typeof error['code'] === 'string'
 }
 
+function isNotFoundError(error: unknown): error is Error & { code: 'not-found' } {
+  return isNodeError(error) && error.code === 'not-found'
```
hughy, line 60: nit -- should we use `ERROR_CODES.NOT_FOUND` here instead of the string?

hughy (approved):

ironfish/src/utils/error.ts
rohanjadvani: This leads to that oclif `undefined` issue we saw yesterday.

PR #4089, opened 2023-07-19
feat(ironfish): Accept transactions into the mempool from node client

rohanjadvani:
## Summary

Use the node client when accepting transactions

## Testing Plan

Covered by existing tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


hughy (approved):

ironfish/src/wallet/wallet.ts
```diff
@@ -984,7 +979,9 @@ export class Wallet {
 
     if (broadcast) {
       await this.addPendingTransaction(transaction)
-      this.memPool.acceptTransaction(transaction)
+      await this.nodeClient.mempool.acceptTransaction({
+        transaction: transaction.serialize().toString('hex'),
+      })
```
hughy, line 984: Should we handle RPC errors on these nodeClient calls? If a request fails because of connectivity the wallet process will crash

PR #4090, opened 2023-07-19
fix memory client close handlers

danield9tqh:
## Summary

When tests are testing streaming endpoints they should close the connection at the end of the test so the node can clean up. It doesn't look like that's happening correctly currently.

## Testing Plan
Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/event/onTransactionGossip.test.ts
+++ b/ironfish/src/rpc/routes/event/onTransactionGossip.test.ts
@@ -28,7 +28,6 @@ describe('Route event/onTransactionGossip', () => {
       serializedTransaction: transaction.serialize().toString('hex'),
     })
 
-    response.end()
-    expect(response.status).toEqual(200)
+    response.close()
   })
 })
--- a/ironfish/src/rpc/routes/node/getLogStream.test.ts
+++ b/ironfish/src/rpc/routes/node/getLogStream.test.ts
@@ -18,9 +18,6 @@ describe('Route node/getLogStream', () => {
     routeTest.node.logger.info('Hello', { foo: 2 })
     const { value } = await response.contentStream().next()
 
-    response.end()
-    expect(response.status).toBe(200)
-
     expect(value).toMatchObject({
       level: LogLevel.Info.toString(),
       tag: expect.stringContaining('ironfishnode'),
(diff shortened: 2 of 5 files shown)


ironfish/src/rpc/routes/event/onTransactionGossip.test.ts
```diff
@@ -28,7 +28,6 @@ describe('Route event/onTransactionGossip', () => {
       serializedTransaction: transaction.serialize().toString('hex'),
     })
 
-    response.end()
-    expect(response.status).toEqual(200)
```
andreacorbellini, line 32: Why was this assertion removed?

ironfish/src/rpc/routes/node/getLogStream.test.ts
```diff
@@ -18,16 +18,15 @@ describe('Route node/getLogStream', () => {
     routeTest.node.logger.info('Hello', { foo: 2 })
     const { value } = await response.contentStream().next()
 
-    response.end()
-    expect(response.status).toBe(200)
```
andreacorbellini, line 22: Why was this assertion removed?

ironfish/src/rpc/routes/node/getLogStream.test.ts
```diff
@@ -41,15 +40,14 @@ describe('Route node/getLogStream', () => {
     routeTest.node.logger.info(`${BigInt(2)}`)
     const { value } = await response.contentStream().next()
 
-    response.end()
-    expect(response.status).toBe(200)
```
andreacorbellini, line 45: Ditto

ironfish/src/rpc/routes/event/onTransactionGossip.test.ts
danield9tqh: I think what this line was trying to do was tell the server that its done listening to the stream endpoint. With an HTTP or TCP connection the client can do that by just closing the connection. When the client/server is in memory however it can't just close the connection and we had no way of telling the RPC handler to stop the stream. That's what the `response.close()` function does.

andiflabs: Got it, thanks for the detailed explanation

andiflabs (approved):

PR #4094, opened 2023-07-20
Store consensus on wallet and use it for wallet/addTransaction RPC

danield9tqh:
## Summary
Standalone wallet should know whether or not its creating invalid transactions based on consensus parameters like max block size and min transaction fee. It could query this information from a full node but reasonably it seems more efficient for the wallet to store its own consensus parameters. The downside is that wallet consensus params could become out of sync due to a hardfork. This might deserve some discussion but proposing this solution because it reduces RPC calls and removes need for adding additional verifyCreatedTransaction RPC endpoint
 
## Testing Plan
Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

Makes some methods on Verifier class static
```
[X] Yes
```

### DIFF
--- a/ironfish/src/consensus/verifier.test.ts
+++ b/ironfish/src/consensus/verifier.test.ts
@@ -29,7 +29,7 @@ import {
   useTxSpendsFixture,
 } from '../testUtilities'
 import { useFixture } from '../testUtilities/fixtures/fixture'
-import { VerificationResultReason } from './verifier'
+import { VerificationResultReason, Verifier } from './verifier'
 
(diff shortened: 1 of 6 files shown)


hughy (approved):
This looks good to me

It makes me wonder whether _all_ of the methods in `Verifier` could be static; the verifier doesn't store any state of its own

danield9tqh:
Yea I think references to a chain + chain database are all that is inhibiting that. I guess that makes sense to initialize it with a chain so you don't have to pass that down all the time.

PR #4096, opened 2023-07-20
adds close method on RpcResponse

hughy:
## Summary

MemoryResponse has a 'close' method that can be used to close a stream before the end of the stream. this is necessary for the MemoryResponse because the stream won't end early even if the caller stops processing responses

for RpcResponse streams the stream ends when the server ends the stream or the client disconnects

adding 'close' to RpcResponse gives the same interface as MemoryResponse for early termination of streams

this is useful in the standalone wallet where the wallet will use streaming endpoints (e.g., chain/followChainStream) using either a MemoryClient or an RpcClient

## Testing Plan

manual testing:
- added a `response.close()` call to the `sync` command and verified that it stops the stream early

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/response.ts
+++ b/ironfish/src/rpc/response.ts
@@ -43,6 +43,14 @@ export class RpcResponse<TEnd = unknown, TStream = unknown> {
     return this as RpcResponseEnded<TEnd>
   }
 
+  close(): void {
+    this.stream.close()
+
+    if (this.timeout) {
+      clearTimeout(this.timeout)
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


danield9tqh:
I don’t think this actually sends a close event to the server side except in the case of MemoryClient. Would we also want that capability in other clients/adapters? That would be a more involved change prob have to update on the adapter side too but I think it would be worth it to keep a consistent abstraction of closing a single request

PR #4097, opened 2023-07-20
implements RemoteChainProcessor

hughy:
## Summary

the standalone wallet will depend on a remote node running a ChainProcessor and sending block events over the chain/followChainStream RPC

the RemoteChainProcessor mimics the behavior of the ChainProcessor but uses chain/getChainInfo and chain/followChainStream to update

update uses chain/followChainStream to stream block events _until_ it reaches the head of the remote chain from the initial call

the onAdd and onRemove Events on the RemoteChainProcessor emit transaction data in addition to the block header since transaction data arrives in the RPC response with the block data

updates wallet methods and tests to accommodate the change in event interface

defines WalletBlockHeader and WalletBlockTransaction types to represent the data required from the RemoteChainProcessor to sync transactions to the wallet

NOTE:
adds optional parameter to chain/followChainStream to end the stream once the chainProcessor reaches the head of the chain (one call to 'update'). this supports the RemoteChainProcessor using the same update logic as the ChainProcessor

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/remoteChainProcessor.ts
+++ b/ironfish/src/wallet/remoteChainProcessor.ts
@@ -89,7 +89,7 @@ export class RemoteChainProcessor {
     Assert.isNotNull(response.block.noteSize)
     let initialNoteIndex = response.block.noteSize
 
-    for (const rpcTransaction of response.block.transactions.reverse()) {
+    for (const rpcTransaction of response.block.transactions.slice().reverse()) {
       Assert.isNotUndefined(rpcTransaction.serialized)
       const transaction = new Transaction(Buffer.from(rpcTransaction.serialized, 'hex'))
       initialNoteIndex -= transaction.notes.length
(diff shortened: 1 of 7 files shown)


ironfish/src/wallet/remoteChainProcessor.ts
```diff
@@ -0,0 +1,105 @@
...
+    return { hashChanged: !BufferUtils.equalsNullable(this.hash, oldHash) }
+  }
+
+  getBlockTransactions(response: FollowChainStreamResponse): WalletBlockTransaction[] {
+    const transactions = []
+
+    Assert.isNotNull(response.block.noteSize)
+    let initialNoteIndex = response.block.noteSize
+
+    for (const rpcTransaction of response.block.transactions.reverse()) {
```
rohanjadvani, line 92: I think before we started this pattern of using a `slice` before `reverse` since it's mutable. It might be good to do that here and in the `return` for consistency in case we run into future issues

rohanjadvani (approved):
Looks good with one suggestion

PR #4100, opened 2023-07-25
Add the `transferOwnershipTo` field to MintDescription

andreacorbellini:
## Summary

This PR introduces a new transaction serialization version: V2. When this new version is used, `MintDescription` gets a new optional field: `transferOwnershipTo`, which will be used to allow transferring the minting permission to an arbitrary account.

## Testing Plan

Unit tests

## Documentation

No doc changes needed yet as this is a purely internal change (not exposed through RPC).

## Breaking Change

This change is backward compatible, but using the new transaction version will introduce a hard fork in the network.

### DIFF
--- a/ironfish-rust-nodejs/src/structs/transaction.rs
+++ b/ironfish-rust-nodejs/src/structs/transaction.rs
@@ -44,7 +44,7 @@ pub const TRANSACTION_EXPIRATION_LENGTH: u32 = TRANSACTION_EXPIRATION_SIZE as u3
 pub const TRANSACTION_FEE_LENGTH: u32 = TRANSACTION_FEE_SIZE as u32;
 
 #[napi]
-pub const TRANSACTION_VERSION: u8 = TX_VERSION;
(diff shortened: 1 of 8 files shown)


mat-if (approved):
🙌 

Doesn't need to be part of this PR, but as a heads-up, we'll need to update the RawTransaction size function as well. https://github.com/iron-fish/ironfish/blob/master/ironfish/src/primitives/rawTransaction.ts#L68-L70
We'll likely catch that as soon as we start writing some typescript tests to utilize this feature but figured I'd mention it

andreacorbellini:
> Doesn't need to be part of this PR, but as a heads-up, we'll need to update the RawTransaction size function as well. https://github.com/iron-fish/ironfish/blob/master/ironfish/src/primitives/rawTransaction.ts#L68-L70 We'll likely catch that as soon as we start writing some typescript tests to utilize this feature but figured I'd mention it

Thanks for pointing this out. I'll work on this, using the same trick as in this PR (i.e. adding a "transaction version" argument to all functions that need to compute the size).

PR #4102, opened 2023-07-25
refactor: remove Transaction.verify and workerPool.verify

mat-if:
## Summary

Due to upcoming mint ownership changes, we will need to pass in extra data to properly verify transactions. As such, transaction.verify() no longer really makes sense. On top of that, maintaining the multiple code-paths is not really providing any utility beyond making unit tests slightly more convenient, which will soon no longer be the case. This PR converges the multiple code-paths on `workerPool.verifyTransactions()` in Typescript and `batch_verify_transactions` in Rust.

This should technically make any previous usage of `workerPool.verify()` every so slightly faster, since it no longer deserializes the transaction inside the worker thread in Typescript before passing to Rust to be deserialized a second time. The fee check is no longer necessary, since it is covered by the min fee check defined by the consensus parameters.

## Testing Plan

Unit tests
Ran on testnet

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[x] Yes - This modifies the way transactions are verified in the SDK. Added notes to the release draft
```


ironfish/src/consensus/verifier.test.ts
```diff
@@ -55,7 +55,7 @@ describe('Verifier', () => {
       )
 
       expect(result).toEqual({
-        reason: VerificationResultReason.ERROR,
+        reason: VerificationResultReason.MINIMUM_FEE_NOT_MET,
```
mat-if, line 58: Miners fee transactions were originally being caught by pool.verify. We have replaced this functionality with the min fee check, so we get an even better error message

ironfish/src/consensus/verifier.ts
```diff
@@ -296,7 +296,7 @@ export class Verifier {
       return { valid: false, reason: VerificationResultReason.MAX_TRANSACTION_SIZE_EXCEEDED }
     }
 
-    if (!transaction.isMinersFee() && transaction.fee() < consensus.parameters.minFee) {
+    if (transaction.fee() < consensus.parameters.minFee) {
```
mat-if, line 299: This was originally only checking min feee if it wasnt a miners fee transaction, which it really should check regardless, since this is checked for new transactions as they come over the network. Miners fee transactions should not ever come over the network, so we can just remove this check.

This is what ultimately replaces the fee check in pool.verify

ironfish/src/workerPool/tasks/verifyTransactions.test.ts
```diff
@@ -0,0 +1,56 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
mat-if, line 1: This was copied from `verifyTransaction.test.ts` and slightly refactored for verifyTransactions.

dguenther (approved):
Read through to make sure it doesn't appear any consensus changes have been made. Do you want to add a notice to the release draft as part of this, since technically the worker pool is accessible to SDK consumers who might be using it to call `verify`?

PR #4103, opened 2023-07-25
Update new block timestamp

ygao76:
## Summary
Update the new block timestamp to prevent negative block mine time

## Testing Plan
unit test

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```
@mat-if @andreacorbellini


ironfish/src/blockchain/blockchain.ts
```diff
@@ -911,6 +913,10 @@ export class Blockchain {
           throw new HeadChangedError(`Can't create a block not attached to the chain head`)
         }
 
+        timestamp = new Date(
+          Math.max(currentTime, Date.parse(heaviestHead.timestamp.toUTCString()) + 1),
```
mat-if, line 917: To get the `Number` timestamp of a `Date` object, we should use `heaviestHead.timestamp.getTime()`

By using `Date.parse(heaviestHead.timestamp.toUTCString())`, you are actually losing millisecond resolution, instead only getting second resolution because you're turning a date into a _string_ that looks like `Wed, 26 Jul 2023 17:58:17 GMT`

This leaves the ability for a block to up to 999 milliseconds behind the previous block, since you could parse a time at, say, 17:58:17.999, which would parse down to 17:58:17.000, meaning we could use a `currentTime` set to 17:58:17.001, and this check would still be valid

image

ygao76: getTime() is exactly what I want. I missed it when I read the doc.

mat-if (approved):
🙌

PR #4107, opened 2023-07-25
Remove withReference from followChainStream

dguenther:
## Summary

The only field in the Transaction primitive that uses the Rust Transaction is `unsignedHash`, and we no longer return that in followChainStream, so we don't have any need to create a Rust transaction when creating block responses for the chain stream. This improves block processing time by roughly 30% (Tested by scanning first 4000 blocks of skynet).

## Testing Plan

Tests should pass, and also modified `service:sync` to run against the local chain to verify performance changes.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/chain/followChainStream.ts
+++ b/ironfish/src/rpc/routes/chain/followChainStream.ts
@@ -86,36 +86,34 @@ routes.register<typeof FollowChainStreamRequestSchema, FollowChainStreamResponse
 
     const send = (block: Block, type: 'connected' | 'disconnected' | 'fork') => {
       const transactions = block.transactions.map((transaction) => {
-        return transaction.withReference(() => {
-          return {
-            ...(request.data?.serialized
-              ? { serialized: transaction.serialize().toString('hex') }
-              : {}),
-            hash: BlockHashSerdeInstance.serialize(transaction.hash()),
-            size: getTransactionSize(transaction),
-            fee: Number(transaction.fee()),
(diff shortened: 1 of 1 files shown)


ironfish/src/rpc/routes/chain/followChainStream.ts
```diff
@@ -86,36 +86,34 @@ routes.register<typeof FollowChainStreamRequestSchema, FollowChainStreamResponse
...
-              creator: mint.asset.creator().toString('hex'),
-              value: mint.value.toString(),
-            })),
-            burns: transaction.burns.map((burn) => ({
-              id: burn.assetId.toString('hex'),
-              value: burn.value.toString(),
-            })),
-          }
-        })
+        return {
```
andreacorbellini, line 89: Nit: I think the `return` is not needed here:

```
const transactions = block.transactions.map((transaction) => {
  return { ... }
}
```

can become:


```
const transactions = block.transactions.map((transaction) => {
  ...
}
```

saving 1 more level of indentation.

andreacorbellini (approved):

ironfish/src/rpc/routes/chain/followChainStream.ts
dguenther: Ah yeah, I'll have to wrap it in parens, but otherwise that'll work. I'll update it

mat-if (approved):
🙌

PR #4112, opened 2023-07-26
Remove unused mempool and onCreateTransaction from Wallet

danield9tqh:
## Summary
Transaction is already added to the mempool as part of `broadcastTransaction` and the onTransactionCreated callback is not used anywhere currently

## Testing Plan
Unit tests + simulation tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


rohanjadvani (approved):
Seems good after lint passes; I think we need to update the constructor

PR #4113, opened 2023-07-26
[BlockTime] Update block verification to only allow positive block mine time

ygao76:
## Summary
Update block verification to only allow positive block mine time
Create a new consensus parameters for the feature activation
Set the new parameters for all envs. Testnet and Mainnet are temporary placeholder.
Add unit tests

## Testing Plan
unit tests of the before and after activation
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[x] Yes
```
@mat-if @andreacorbellini


ironfish/src/consensus/verifier.test.ts
```diff
@@ -373,6 +373,166 @@ describe('Verifier', () => {
     })
   })
 
+  describe('Before postive block mine time activation', () => {
```
mat-if, line 376: What do you think about nesting these tests in the `verifyBlockHeaderContextual` block alongside the other tests? We have an existing test there, which I think makes a good place for these to go. We should also probably just remove the old test if it isn't relevant anymore: https://github.com/iron-fish/ironfish/blob/master/ironfish/src/consensus/verifier.test.ts#L521

ironfish/src/consensus/verifier.test.ts
```diff
@@ -373,6 +373,166 @@ describe('Verifier', () => {
...
+      )
+      header = block.header
+      const previousBlock = block.header.previousBlockHash
+      const previousHeader = await chain.getHeader(previousBlock)
+      Assert.isNotNull(previousHeader)
+      prevHeader = previousHeader
+    })
+
+    it('fails validation when timestamp is too low', async () => {
+      jest
```
mat-if, line 396: Is this spy being used? I don't think it's doing anything. I'd double-check the extra spy usage in all the tests to make sure they are removed if they aren't being used

ironfish/src/consensus/verifier.test.ts
```diff
@@ -373,6 +373,166 @@ describe('Verifier', () => {
...
+          (nodeTest.chain.consensus.parameters.allowedBlockFutureSeconds + 2) * 1000,
+      )
+
+      expect(await nodeTest.verifier.verifyBlockAdd(block, prevHeader)).toMatchObject({
+        reason: VerificationResultReason.BLOCK_TOO_OLD,
+        valid: false,
+      })
+    })
+
+    it('fails validation when timestamp is too far in the future', async () => {
```
mat-if, line 410: It looks like there's tests in here already for TOO_FAR_IN_FUTURE: https://github.com/iron-fish/ironfish/blob/master/ironfish/src/consensus/verifier.test.ts#L397

Since you aren't modifying that behavior and we already have tests there, I don't think we should duplicate the tests

ironfish/src/defaultNetworkDefinitions.ts
```diff
@@ -196,6 +204,7 @@ export const DEVNET = `
         "targetBlockTimeInSeconds": 60,
         "targetBucketTimeInSeconds": 10,
         "maxBlockSizeBytes": 524288,
-        "minFee": 0
+        "minFee": 0,
+        "disallowNegativeBlockMineTime": 3
```
mat-if, line 208: Is there a specific reason for 3? I feel like 0 is the most logical choice here, otherwise we're always going to be wondering why 3 was chosen in the future (same with the strategy tests below)

If you need to modify it for the tests, you should be able to do something like `chain.consensus.parameters.negativeBlockMineTime = 3` in the beforeAll or something

ironfish/src/consensus/consensus.ts
```diff
@@ -33,6 +33,12 @@ export type ConsensusParameters = {
    * The minimum fee that a transaction must have to be accepted
    */
   minFee: number
+
+  /**
+   * Before upgrade we had negative block mine time in block headers. After this
+   * block we do a check to disallow negative block mine time.
+   */
+  disallowNegativeBlockMineTime: number
```
NullSoldier, line 41: I don't think we have a concept of block mined time. Just block timestamps. I know it seems pedantic, but I would recommend referring to concepts that can be found in our models like `block.header.timestamp`.

ygao76: I am not inclined to name is disallowNegativeBlockTimestamp since the timestamp itself is never negative, its the comparison with the previous block timestamp.
How about `disallowBlockTimestampSmallerThanPreviousBlock`? I know its pretty long but I think thats accurate if we dont want to introduce block mine time to codbase and just use block timestamp

ironfish/src/consensus/verifier.test.ts
ygao76: Sure, but I want to keep the current describe and it structures because its clear and easy to read when group the tests with before activation behavior and after activation behavior.

ironfish/src/consensus/verifier.test.ts
ygao76: Yes, the spy is used to control the Date.now() value in the verification logic. Its in `verifyBlock()` function, not all the unit tests triggered that call path tho. I will remove the ones that throw error before reach verifyBlock() function.

ironfish/src/consensus/verifier.test.ts
ygao76: Yeah I saw the existing test for TOO_FAR_IN_FUTURE case. I added them here to make sure the block sequence before and after activation are behaved as expected for TOO_FAR_IN_FUTURE case. The existing test does not test the activation flag params.

ironfish/src/defaultNetworkDefinitions.ts
ygao76: I use 3 because the sequence starts at 1. But 1 is genesis block so it has special logic and skip the time check. If we set to 1, this logic 
```
 this.chain.consensus.isActive(
        this.chain.consensus.parameters.disallowNegativeBlockMineTime,
        current.sequence,
      )
``` 
would never be called in devnet.
If I set to 2, it will be called but it will never go to the else statement since all the blocks <= the activation sequence. 
So I set to 3, so the first block is the old logic and the rest blocks are the new logics. I think this config is the most generic for devnet and unit tests. What is your thought on this?

mat-if: What is the goal of having it activate on devnet for 1 block? If the goal is ensure the code path is hit, I think that is the responsibility of unit tests. It's also ensured that its hit for both testnet and mainnet since both are guaranteed to have blocks with a negative time difference since the blockchains are immutable and we will forever have to support those blocks prior to this change.

ygao76: Updated the dev value to 1, and set the value for relevant unit tests

ironfish/src/consensus/consensus.ts
andreacorbellini: Some proposals: `disallowPastBlockTimestamp` or `enforceMonotonicBlockTimestamp`

andreacorbellini: "Monotonic block timestamp(s)" is probably the perfect term for this project too

ygao76: I like the proposal, I will rename to `enforceSequentialBlockTime`

mat-if (approved):

PR #4116, opened 2023-07-27
Fix `RawTransaction.postedSize` by considering fees when computing change notes

andreacorbellini:
## Summary

`RawTransaction.postedSize` contains logic to compute change notes, which are implicit output entries that may be included in each transaction. Currently, `postedSize` does not take into account the transaction fee, and as such, in certain circumstances, it may incorrectly add a change note even if such note won't be produced when the transaction is posted.

### Detailed explanation

Given the following proposed transaction:

```
fee: 1 $IRON
spends: [100 $IRON]
outputs: []
```

a change note will be implicitly created to take care of the remaining 99 $IRON.

Given the following proposed transaction:

```
fee: 1 $IRON
spends: [100 $IRON]
outputs: [99 $IRON]
```

no change note will be created because there is no "leftover" amount: 100 (spends) - 99 (outputs) - 1 (fee) = 0 $IRON.

`RawTrasaction.postedSize()` does not correctly handle the second case, and wrongly thinks that a change note will be always created, because it does not take the fee into account and thinks that there is 100 (spends) - 99 (outputs) = 1 $IRON remaining that needs to be included in a change note.

In this circumstance, `RawTrasaction.postedSize()` will return a number of bytes that is higher than the actual number of bytes of the final transaction.

### References

*   How `ProposedTransaction.post` computes the change note in Rust:

    https://github.com/iron-fish/ironfish/blob/master/ironfish-rust/src/transaction/mod.rs#L196-L216

    Note how the native asset is treated specially to include the fee.

*   How `RawTransactrion.postedSize` computes the change note in Typescript:

    https://github.com/iron-fish/ironfish/blob/master/ironfish/src/primitives/rawTransaction.ts#L75-L93

    Note that the native asset or the fee are not taken into consideration.

## Testing Plan

New unit tests that crate multiple variants of `RawTransaction` (with fees, without fees, with mints, without mints, ...) and compares the output of `postedSize` with the size of `post`.

Note that, because these tests run many possible variants, the new test file is the slowest of all tests (takes ~2 minutes to run on my laptop).

## Documentation

No doc change required.

## Breaking Change

Not a breaking change, but the way fees are estimated might change.


dguenther (approved):

ironfish/src/primitives/rawTransaction.test.slow.ts
```diff
@@ -0,0 +1,202 @@
...
+        'withOutputs',
+      ] as const
+
+      testOptionCombinations(flags).forEach((options) => {
+        // eslint-disable-next-line jest/valid-title
+        it(describeTestOptions(options), async () => {
+          const account = await useAccountFixture(nodeTest.wallet)
+
+          const raw = createTestRawTransaction(account, options)
+          const serialized = raw.post(account.spendingKey).serialize()
```
mat-if, line 195: What do you think of using a tx fixture? It makes the run time go from ~90s on my machine to ~12s, and on CI it looks like it takes almost 500 seconds: https://github.com/iron-fish/ironfish/actions/runs/5675255965/job/15380216885?pr=4116 `@ironfish/sdk:  PASS   @ironfish/sdk  src/primitives/rawTransaction.test.slow.ts (482.75 s)`

Kind of a goofy function for this use case, we could probably use something like a `usePostRawTxFixture` type function that literally just takes a raw tx and posts it, but this works at least for now

```suggestion
          const tx = await useTxFixture(
            nodeTest.wallet,
            account,
            account,
            () => {
              return raw.post(account.spendingKey)
            },
            undefined,
            undefined,
            false,
          )
          const serialized = tx.serialize()
```

mat-if (approved):
Minor nit about fixtures otherwise this looks great 🙌

PR #4119, opened 2023-07-27
feat(ironfish): Create `wallet/getAsset`

rohanjadvani:
## Summary

Add an RPC endpoint to fetch asset details from the wallet. We are using this to replace the call in `wallet:burn` which currently uses the chain.

## Testing Plan

Unit tests

## Documentation

```
[x] Yes
```

https://github.com/iron-fish/website/pull/501

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/wallet/getAsset.ts
```diff
@@ -0,0 +1,92 @@
...
+  id: string
+  metadata: string
+  name: string
+  nonce: number
+  status: string
+  // Populated for assets the account owns
+  supply: string | null
+  // Populated once the asset has been added to the main chain
+  blockHash: string | null
+  sequence: number | null
```
hughy, line 30: There are a few differences between this response and the one from `getAssets`, should we make them the same?

rohanjadvani: Yep, good catch. Updated

hughy (approved):

ironfish/src/rpc/routes/wallet/getAsset.ts
```diff
@@ -0,0 +1,91 @@
...
+  id: string
+}
+
+export type GetWalletAssetResponse = {
+  createdTransactionHash: string
+  creator: string
+  id: string
+  metadata: string
+  name: string
+  nonce: number
```
hughy, line 25: Should we add nonce to `getAssets` in a followup?

hughy (approved):

ironfish/src/rpc/routes/wallet/getAsset.ts
rohanjadvani: Yeah

PR #4120, opened 2023-07-27
Make remote call to node to see if its synced enough to create a tranaction

danield9tqh:
## Summary
This makes 2 changes
1) No longer check whether the remote not is connected to the peer network before attempting to send a transaction. 
   - Removing this check because the wallet can attempt rebroadcasting a transaction asynchronously so the node doesn't need to have peers for the transaction to eventually succeed
2) Checking the remote node to see if its head is synced instead of assuming there will always be a local node

## Testing Plan
Unit tests + local testing

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/sendTransaction.ts
+++ b/ironfish/src/rpc/routes/wallet/sendTransaction.ts
@@ -72,13 +72,18 @@ routes.register<typeof SendTransactionRequestSchema, SendTransactionResponse>(
 
     const account = getAccount(node.wallet, request.data.account)
 
-    if (!node.peerNetwork.isReady) {
-      throw new ValidationError(
-        `Your node must be connected to the Iron Fish network to send a transaction`,
-      )
-    }
+    const chainInfo = await node.wallet.nodeClient.chain.getChainInfo()
+    const headHash = chainInfo.content.oldestBlockIdentifier.hash
+    const headBlock = await node.wallet.nodeClient.chain.getBlock({ hash: headHash })
 
-    if (!node.chain.synced) {
+    const maxSyncedAgeMs =
+      node.wallet.config.get('maxSyncedAgeBlocks') *
+      node.wallet.consensus.parameters.targetBlockTimeInSeconds *
+      1000
+
+    const isSynced = headBlock.content.block.timestamp.valueOf() >= Date.now() - maxSyncedAgeMs
+
+    if (!isSynced) {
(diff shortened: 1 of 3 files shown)


ironfish/src/rpc/routes/wallet/sendTransaction.ts
```diff
@@ -72,13 +72,18 @@ routes.register<typeof SendTransactionRequestSchema, SendTransactionResponse>(
...
+    const chainInfo = await node.wallet.nodeClient.chain.getChainInfo()
+    const headHash = chainInfo.content.oldestBlockIdentifier.hash
+    const headBlock = await node.wallet.nodeClient.chain.getBlock({ hash: headHash })
+
+    const maxSyncedAgeMs =
+      node.wallet.config.get('maxSyncedAgeBlocks') *
+      node.wallet.consensus.parameters.targetBlockTimeInSeconds *
+      1000
+
+    const isSynced = headBlock.content.block.timestamp.valueOf() >= Date.now() - maxSyncedAgeMs
```
hughy, line 84: Why not use `blockchain.synced` from `node/getStatus`?

danield9tqh: That would def be simpler. I was thinking wallet may want to use its value for `maxSyncedAgeBlocks` to determine synced or not. Maybe that is not that important rn though

danield9tqh: Yea I agree I think the simplicity and code re-use of using synced from blockchain status probably outweighs benefits of allowing wallet to determine synced for themselves. I'll switch to that

hughy: Yeah I think it could be useful to have the wallet use its own criteria for synced, but I'm not entirely sure when it would be useful 😅

hughy (approved):

ironfish/src/rpc/routes/wallet/sendTransaction.ts
danield9tqh: I think its mostly a factor of how much are you willing to sacrifice on your note witness (anonymity set). If someone want to use the largest possible anonymity set they would have a higher criteria for how synced the chain should be. Someone else might not care. If those 2 users are connecting their wallets to the same node they could configure that separately.

PR #4121, opened 2023-07-27
Make remote call for account status

danield9tqh:
## Summary
For `wallet/getAccountsStatus` command we return whether the node contains the account head or not. Moving this call to use a node client in the case of a standalone wallet

## Testing Plan
Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/getAccountsStatus.ts
+++ b/ironfish/src/rpc/routes/wallet/getAccountsStatus.ts
@@ -3,6 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import * as yup from 'yup'
 import { Assert } from '../../../assert'
+import { HeadValue } from '../../../wallet/walletdb/headValue'
 import { ApiNamespace, routes } from '../router'
 
 export type GetAccountStatusRequest = { account?: string }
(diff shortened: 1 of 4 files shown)


ironfish/src/rpc/routes/wallet/getAccountsStatus.ts
```diff
@@ -45,22 +46,26 @@ routes.register<typeof GetAccountStatusRequestSchema, GetAccountStatusResponse>(
...
-      const headHash = headHashes.get(account.id)
-      const blockHeader = headHash ? await node.chain.getHeader(headHash) : null
-      const headInChain = !!blockHeader
-      const headSequence = blockHeader?.sequence || 'NULL'
+      const head = heads.get(account.id)
+      const blockResponse = head?.hash
+        ? await node.wallet.nodeClient.chain.getBlock({
+            hash: head.hash.toString('hex'),
+          })
+        : null
```
hughy, line 61: Could this reuse the `chainHasBlock` method that we added to the wallet?

danield9tqh: Yea good call

hughy (approved):

PR #4130, opened 2023-07-28
Use node client for wallet/rescanAccounts

danield9tqh:
## Summary
Switch over wallet rescanAccounts RPC to use remote node client

## Testing Plan
Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/rescanAccount.ts
+++ b/ironfish/src/rpc/routes/wallet/rescanAccount.ts
@@ -46,20 +46,20 @@ routes.register<typeof RescanAccountRequestSchema, RescanAccountResponse>(
 
       let fromHash = undefined
       if (request.data.from && request.data.from > GENESIS_BLOCK_SEQUENCE) {
-        const header = await node.chain.getHeaderAtSequence(request.data.from)
+        const response = await node.wallet.chainGetBlock({ sequence: request.data.from })
 
-        if (header === null) {
+        if (response === null) {
           throw new ValidationError(
             `No block header found in the chain at sequence ${request.data.from}`,
           )
         }
 
-        fromHash = header.hash
+        fromHash = Buffer.from(response.block.hash, 'hex')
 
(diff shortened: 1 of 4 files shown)


ironfish/src/wallet/wallet.ts
```diff
@@ -1571,12 +1571,15 @@ export class Wallet {
   }
 
   async chainHasBlock(hash: Buffer): Promise<boolean> {
+    return (await this.chainGetBlock({ hash: hash.toString('hex') })) === null
```
rohanjadvani, line 1574: Wouldn't === null mean the block doesn't have a block?

rohanjadvani (approved):

ironfish/src/rpc/routes/wallet/rescanAccount.ts
```diff
@@ -46,20 +46,20 @@ routes.register<typeof RescanAccountRequestSchema, RescanAccountResponse>(
 
       let fromHash = undefined
       if (request.data.from && request.data.from > GENESIS_BLOCK_SEQUENCE) {
-        const header = await node.chain.getHeaderAtSequence(request.data.from)
+        const response = await node.wallet.chainGetBlock({ sequence: request.data.from })
```
rohanjadvani, line 49: Originally we were fetching the header, but now this is going to return the full block with transactions. Should we only use the header here?

danield9tqh: Do we have an RPC call for that? I just used this because I thought it was the only call we had and didn't want to make add another endpoint for this specific use case.

danield9tqh: Discussed. This RPC call specifically is a one-off call so bandwidth is not as big of an issue so we'll use the getBlock call for now.

rohanjadvani (approved):

PR #4132, opened 2023-07-31
Fix checking result of proof verification and add tests

dguenther:
## Summary

When switching from bellman to bellperson, the batch verification function changed. Previously, the proof verification would return a 
`Result` with `Err` in every error case and `Ok(())` otherwise, but now it returns `Err` in some error cases and `Ok(bool)` in cases where the proof 
can't be verified with the given proof verification key. This PR fixes this check and adds tests to verify the change.

## Testing Plan

Adds additional Rust tests and verifies that the tests failed before and pass with this change.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/transaction/mints.rs
+++ b/ironfish-rust/src/transaction/mints.rs
@@ -355,6 +355,13 @@ mod test {
 
         let other_randomized_public_key = redjubjub::PublicKey(key.view_key.authorizing_key.into())
             .randomize(jubjub::Fr::random(thread_rng()), *SPENDING_KEY_GENERATOR);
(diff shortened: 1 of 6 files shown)


NullSoldier:
Ok, this looks good. I reviewed the code, and tested this manually.

This also leads me to another idea, that we should add a test for this in the TS side. We can generate a transaction with invalid parameters, copy the serialized version of it into JS, and keep that around for various tests. After this merges, I can start working on that.

NullSoldier (approved):

PR #4137, opened 2023-08-01
Add WalletNode interface and connect to RPC handlers

danield9tqh:
## Summary
Creating a similar interface for a full node (IronfishNode) and a standalone wallet (WalletNode) makes RPC reqeust handlers much easier to manage. This allows the standalone wallet to use the `/config` `/rpc` and `/workers` RPC endpoints without changes. The full `WalletNode` will be created in a future PR.   

## Testing Plan
Unit tests + running local node

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/router.ts
+++ b/ironfish/src/rpc/routes/router.ts
@@ -5,7 +5,7 @@ import { Assert } from '../../assert'
 import { IronfishNode } from '../../node'
 import { YupSchema, YupSchemaResult, YupUtils } from '../../utils'
 import { StrEnumUtils } from '../../utils/enums'
-import { Wallet } from '../../wallet'
+import { WalletNode } from '../../walletNode'
 import { ERROR_CODES } from '../adapters'
 import { ResponseError, ValidationError } from '../adapters/errors'
 import { RpcRequest } from '../request'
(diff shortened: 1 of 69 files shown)


ironfish/src/rpc/routes/router.ts
```diff
@@ -142,4 +142,10 @@ class Routes {
   }
 }
 
+export function isFullNode(x: unknown, message?: string): asserts x is IronfishNode {
```
rohanjadvani, line 145: Is this used right now? Should we use this in all the assertions in the RPC handlers?

danield9tqh: Not used, removing

ironfish/src/walletNode.ts
```diff
@@ -0,0 +1,19 @@
...
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { AssetsVerifier } from './assets'
+import { Config } from './fileStores'
+import { Logger } from './logger'
+import { RpcServer } from './rpc'
+import { Wallet } from './wallet/wallet'
+import { WorkerPool } from './workerPool'
+
+export type WalletNode = {
```
rohanjadvani, line 12: I think it's ok to include this as a starting point knowing that it'll be refactored later

rohanjadvani (approved):

PR #4141, opened 2023-08-02
removes wallet:repair command

hughy:
## Summary

the command is a holdover from the end of the phase 2 testnet.

it checks for common wallet bugs that we identified at the end of that testnet phase and provided users with a way to quickly fix their wallets to send transactions without having to rescan the entire chain.

now, with a standalone wallet, the command won't always work. if a user runs the command using a standalone wallet that does not have access to a local synced chain database it will make incorrect changes to tthe wallet database, like marking all notes as unspent.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ygao76 (approved):
RIP wallet:repair

rohanjadvani (approved):

PR #4148, opened 2023-08-02
feat(ironfish): Remove wallet specific config

rohanjadvani:
## Summary

Since we're moving to the concept of a wallet node and preventing a node / standalone wallet to run at the same time, we can re-use the existing config.

## Testing Plan

N/A

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/fileStores/config.ts
+++ b/ironfish/src/fileStores/config.ts
@@ -307,6 +307,11 @@ export type ConfigOptions = {
   walletNodeHttpPort: number
 
   /**
+   * Enable standalone wallet process RPC server
+   */
+  walletEnableRpc: boolean
+
+  /**
    * Enable IPC connections to a standalone wallet RPC server
    */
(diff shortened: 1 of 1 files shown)


danield9tqh (approved):

ironfish/src/fileStores/config.ts
```diff
@@ -279,66 +279,6 @@ export type ConfigOptions = {
...
-  walletNodeTcpHost: string
-  walletNodeTcpPort: number
-  walletNodeTlsEnabled: boolean
-
-  /**
-   * Enable standalone wallet process to connect to a node via HTTP
-   */
-  walletNodeHttpEnabled: boolean
-  walletNodeHttpHost: string
-  walletNodeHttpPort: number
```
hughy, line 307: Do we want to keep these to configure how the wallet connects to the node?

PR #4155, opened 2023-08-02
fix: Allow test path patterns to be passed to test commands

mat-if:
## Summary

This enables the ability to do something like `yarn test:slow rawTransaction.test.slow.ts`. Previously, this would exclude the given test file, instead of running only that test file

It seems by having `--testPathIgnorePatterns` at the end of the command, that causes it to consume the filter string and make it have the opposite of the intended effect. By moving them "inwards", we allow them to not consume the filters unexpectedly

Before: 
image

After:
image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


hughy (commented):
Does `--testPathIgnorePatterns` have any effect if it isn't followed by a pattern? Could we remove these?

mat-if:
> Does `--testPathIgnorePatterns` have any effect if it isn't followed by a pattern? Could we remove these?

Good question. It's needed to override this jest config setting for the base test command: https://github.com/iron-fish/ironfish/blob/master/config/jest.config.base.js#L7

This is what it looks like if you exclude that option from the test:slow command:
image

Realistically a refactor of these would probably be ideal to have a jest config for each test group, or something, but this has been doing a good enough job for now

hughy (approved):

andreacorbellini:
For the record, even without this change, you can still run individual test files by using `--`, example:

```
yarn test:slow -- src/primitives/rawTransaction.test.slow.ts
```

In my case, I actually have to pass `--` twice, because I'm not using the most recent version of yarn:

```
yarn test:slow -- -- src/primitives/rawTransaction.test.slow.ts
```

andreacorbellini (approved):

PR #4168, opened 2023-08-03
feat(cli,ironfish): Create a CLI helper to create a config client

rohanjadvani:
## Summary

The CLI commands under the config namespace can be run for both a standalone wallet and full node. This code change creates a client wrapper that does not open the database and uses a wallet node to instantiate an RPC client.

Some considerations as part of this change:
* We can create a parent `IronfishNode` class which the `FullNode` and `WalletNode` inherit from, and that can be used in this call. This design seems like a convenience rather than getting a benefit of a lot of shared logic (outside of just storing properties). This would also need some updates to the migration system and potentially refactoring how arguments are passed into the RPC.
* We can use the existing `connectRpc` and add some arguments. This can be a future improvement as there's probably a way to leverage generics with the RpcClient and namespaces, but the goal of this is to get something working. We will come back and investigate it.

## Testing Plan

Ran CLI commands under config namespace

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/sdk.ts
+++ b/ironfish/src/sdk.ts
@@ -246,34 +246,6 @@ export class IronfishSdk {
     return node
   }
 
-  async connectRpcConfig(
-    forceLocal = false,
-    forceRemote = false,
-  ): Promise<Pick<RpcClient, 'config'>> {
-    forceRemote = forceRemote || this.config.get('enableRpcTcp')
-
-    if (!forceLocal) {
-      if (forceRemote) {
-        await this.client.connect()
-        return this.client
-      }
-
-      const connected = await this.client.tryConnect()
-      if (connected) {
-        return this.client
-      }
-    }
-
-    // This connection uses a wallet node since that is the most granular type
-    // of node available. This can be refactored in the future if needed.
-    const node = await this.walletNode()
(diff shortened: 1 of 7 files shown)


ironfish/src/sdk.ts
```diff
@@ -246,6 +246,34 @@ export class IronfishSdk {
...
+    }
+
+    // This connection uses a wallet node since that is the most granular type
+    // of node available. This can be refactored in the future if needed.
+    const node = await this.walletNode()
+    const clientMemory = new RpcMemoryClient(
+      this.logger,
+      node.rpc.getRouter([ApiNamespace.config]),
+    )
+    return clientMemory
```
hughy, line 274: The rest of the logic outside this block looks the same as `connectRpc`. Could we reuse that if `forceLocal` is false?

ironfish/src/sdk.ts
```diff
@@ -246,6 +246,34 @@ export class IronfishSdk {
     return node
   }
 
+  async connectRpcConfig(
```
hughy, line 249: Since this logic is used only for config commands, and only when running with a local flag, should this be in the SDK or a util in the CLI?

PR #4180, opened 2023-08-08
feat(cli,ironfish): Add SDK method to connect to wallet rpc client

rohanjadvani:
## Summary

The SDK creates an RPC client when initialized. Config, rpc, wallet, and worker namespaced CLI commands connect to this RPC client but need to work for both a full node and standalone wallet. This code change adds a new SDK method to connect to the client property but default to a wallet node's memory client on failures or if local / remote is not forced.

We can move this as a utility into the CLI as well, but it seems better to offer this in the SDK for both node types.

There were alternatives that were considered (ex. writing the running context to a store), but this is a quick and functional solution without side effects that can be iterated upon in the future.

## Testing Plan

Running CLI commands under the updated namespaces

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/rpc/status.ts
+++ b/ironfish-cli/src/commands/rpc/status.ts
@@ -23,7 +23,7 @@ export default class Status extends IronfishCommand {
     const { flags } = await this.parse(Status)
 
     if (!flags.follow) {
-      const client = await this.sdk.connectRpc()
+      const client = await this.sdk.connectWalletRpc()
       const response = await client.rpc.getRpcStatus()
       this.log(renderStatus(response.content))
       this.exit(0)
--- a/ironfish-cli/src/commands/workers/status.ts
+++ b/ironfish-cli/src/commands/workers/status.ts
@@ -23,7 +23,7 @@ export default class Status extends IronfishCommand {
     const { flags } = await this.parse(Status)
 
     if (!flags.follow) {
-      const client = await this.sdk.connectRpc()
+      const client = await this.sdk.connectWalletRpc()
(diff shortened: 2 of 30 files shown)


ironfish-cli/src/commands/workers/status.ts
```diff
@@ -23,7 +23,7 @@ export default class Status extends IronfishCommand {
     const { flags } = await this.parse(Status)
 
     if (!flags.follow) {
-      const client = await this.sdk.connectRpc()
+      const client = await this.sdk.connectWalletRpc()
```
hughy, line 26: Same comment as for `rpc:status`: I'm not sure it ever makes sense for this to be a MemoryClient

hughy (approved):

ironfish-cli/src/commands/rpc/status.ts
```diff
@@ -23,7 +23,7 @@ export default class Status extends IronfishCommand {
     const { flags } = await this.parse(Status)
 
     if (!flags.follow) {
-      const client = await this.sdk.connectRpc()
+      const client = await this.sdk.connectWalletRpc()
```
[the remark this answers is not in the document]
rohanjadvani: Discussed offline, will be a future improvement since this behavior existed before

ironfish-cli/src/commands/workers/status.ts
rohanjadvani: Discussed offline, will be a future improvement since this behavior existed before

PR #4182, opened 2023-08-08
adds limit to block events in remoteChainProcessor updates

hughy:
## Summary

remoteChainProcessor uses the chain/followChainStream RPC to stream block events from a node. the endpoint can write messages to the queue much more quickly than the wallet consumes them because the wallet decrypts the notes on each transaction on each block as it processes the stream

adds a 'limit' parameter to the chain/followChainStream RPC to close the stream once that number of events have been streamed to the client

adds a wallet config, 'walletSyncingMaxQueueSize', to configure the limit to use when syncing or scanning using the RemoteChainProcessor

updates scanTransactions to make multiple calls to 'update' until the RemoteChainProcessor instance used for scanning has reached a sequence greater than or equal to the head of the chain when the scan began

## Testing Plan

- unit tests
- manual testing: rescanning on mainnet, testnet, devnet

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/chain/followChainStream.test.ts
+++ b/ironfish/src/rpc/routes/chain/followChainStream.test.ts
@@ -84,12 +84,6 @@ describe('Route chain/followChainStream', () => {
     })
 
     streamed = await chainStream.contentStream().next()
-    expect(streamed?.value).toMatchObject({
-      type: 'connected',
-      block: { hash: blockA1.header.hash.toString('hex') },
-    })
-
-    streamed = await chainStream.contentStream().next()
     expect(streamed?.value).toBeUndefined()
     expect(streamed?.done).toBe(true)
   })
--- a/ironfish/src/rpc/routes/chain/followChainStream.test.ts
+++ b/ironfish/src/rpc/routes/chain/followChainStream.test.ts
@@ -62,4 +62,35 @@ describe('Route chain/followChainStream', () => {
       },
     })
   })
+
+  it('ends the stream if the limit is reached', async () => {
+    const { chain } = routeTest
+    await chain.open()
+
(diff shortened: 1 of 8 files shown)


ironfish/src/rpc/routes/chain/followChainStream.test.ts
```diff
@@ -62,4 +62,35 @@ describe('Route chain/followChainStream', () => {
...
+    chainStream = routeTest.client.request('chain/followChainStream', { limit: 1 })
+
+    let streamed = await chainStream.contentStream().next()
+    expect(streamed?.value).toMatchObject({
+      type: 'connected',
+      block: { hash: chain.genesis.hash.toString('hex') },
+    })
+
+    streamed = await chainStream.contentStream().next()
+    expect(streamed?.value).toMatchObject({
```
rohanjadvani, line 87: This is two values returned from the iterator. Is this expected with a limit of 1?

rohanjadvani (approved):

ironfish/src/wallet/remoteChainProcessor.ts
```diff
@@ -25,23 +25,31 @@ export class RemoteChainProcessor {
...
     this.hash = options.head
+    this.maxQueueSize = options.maxQueueSize
   }
 
   async update({ signal }: { signal?: AbortSignal } = {}): Promise<{ hashChanged: boolean }> {
     const chainStream = this.nodeClient.chain.followChainStream({
       head: this.hash?.toString('hex') ?? null,
       serialized: true,
       wait: false,
+      limit: this.maxQueueSize,
```
jowparks, line 52: only change required client side

PR #4183, opened 2023-08-08
speeds up findFork for linear fast forward case

hughy:
## Summary

from the docstring of find fork: 'If one block is a linear fast forward to the other with no forks, then the earlier block will be returned.'

in the case of a linear fast forward where both blocks are on the head chain we can short-circuit the findFork logic and immediately return the earlier block without iterating from one block to the other.

this helps speed up wallet rescans that make multiple calls to the 'chain/followChainStream' RPC: each call finds the fork between the current head of the node's chain and the hash passed to the RPC

## Testing Plan

- existing unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


rohanjadvani (approved):
This is great

PR #4185, opened 2023-08-08
Add limit for queue size of pending transactions in the wallet

danield9tqh:
## Summary
When the wallet listens for incoming unconfirmed transactions, limit the, number of transactions the wallet can queue up before it starts dropping them.

## Testing Plan
Unit tests, local testing

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/wallet.ts
```diff
@@ -323,6 +323,11 @@ export class Wallet {
       this.isSyncingTransactionGossip = true
 
       for await (const content of response.contentStream()) {
+        // Start dropping trasactions if we have too many to process
+        if (response.bufferSize() > this.config.get('walletGossipTransactionsMaxQueueSize')) {
+          continue
```
andreacorbellini, line 328: Should we log a message if this occurs?

danield9tqh: Yea that's a good idea

hughy (approved):

PR #4190, opened 2023-08-09
feat(cli,ironfish): Add `wallet:start`

rohanjadvani:
## Summary

* Sets default wallet node IPC path to an empty string. If you run the standalone wallet as is, it'll connect to its own data directory and throw an error.
* Update connection logging to check against `undefined`
* Log out every 20 blocks connected
* Start the wallet after the node client is connected
* Stop the wallet when the node client is disconnected

## Testing Plan

Manually testing CLI

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/wallet/start.ts
```diff
@@ -0,0 +1,183 @@
...
+    await this.startDonePromise
+    await this.node?.shutdown()
+    await this.node?.closeDB()
+  }
+
+  private async firstRun(node: IronfishNode): Promise<void> {
+    this.log('')
+    this.log('Thank you for installing the Iron Fish Wallet Node.')
+
+    if (!node.config.get(ENABLE_TELEMETRY_CONFIG_KEY)) {
```
danield9tqh, line 155: I think we can remove telemetry stuff since it isn't used in wallet

rohanjadvani: Discussed in huddle. Alternative options included a separate first run flag for the wallet, but a concern was if we run the standalone wallet and then the full node. This will throw an error because the wallet head may not be in the chain and may require a rescan/reset. We'll leave this for now.

hughy (approved):

PR #4194, opened 2023-08-09
Rahul/ifl 1497 add disablewalletsync to config

patnir:
## Summary

Adding `disableWalletSync` to config options

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/fileStores/config.ts
+++ b/ironfish/src/fileStores/config.ts
@@ -32,7 +32,7 @@ export type ConfigOptions = {
    */
   p2pStunServers: string[]
   databaseMigrate: boolean
-  disableWalletSync: boolean
+  enableWalletSync: boolean
   editor: string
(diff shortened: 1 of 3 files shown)


ironfish/src/fileStores/config.ts
```diff
@@ -32,6 +32,7 @@ export type ConfigOptions = {
    */
   p2pStunServers: string[]
   databaseMigrate: boolean
+  disableWalletSync: boolean
```
rohanjadvani, line 35: Most config flags here are prefixed with `enable`. Can this be consistent and default to a different value?

patnir: Cool, I like that more. I'll default it to true

rohanjadvani (approved):

PR #4196, opened 2023-08-10
fixes account import if wallet not conneted to node

hughy:
## Summary

importAccount has two dependencies on a node connection: it checks to see if the account birthday is in the chain and it (by default) begins a scan following import. if there is no connection to a running node the wallet cannot do either of these things.

fixes node connection dependencies in import:
- sets account createdAt to null if it cannot be checked against the chain. this is what the wallet does during import if the account birthday is not found in the chain
- skips rescanning if there is no node connection

## Testing Plan

manual testing

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/importAccount.ts
+++ b/ironfish/src/rpc/routes/wallet/importAccount.ts
@@ -54,7 +54,7 @@ routes.register<typeof ImportAccountRequestSchema, ImportResponse>(
       ...accountImport,
     })
 
-    if (request.data.rescan) {
+    if (request.data.rescan && node.wallet.nodeClient) {
       void node.wallet.scanTransactions()
     } else {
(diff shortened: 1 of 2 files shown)


ironfish/src/rpc/routes/wallet/importAccount.ts
```diff
@@ -54,7 +54,7 @@ routes.register<typeof ImportAccountRequestSchema, ImportResponse>(
       ...accountImport,
     })
 
-    if (request.data.rescan) {
+    if (request.data.rescan && node.wallet.nodeClient) {
```
rohanjadvani, line 57: I was wondering if we should check if this client is connected, but there is retry logic so checking non-null seems good.

rohanjadvani (approved):

PR #4200, opened 2023-08-10
test partial decryption

jowparks:
Decrypting notes runtime, decrypting the first 32 bytes vs whole note:

tl;dr:
decrypt note for spender ~ 40% reduction in time
Full note  vs first 32 bytes: 300 us vs 173 us

decrypt note for owner ~75% reduction in time
Full note  vs first 32 bytes: 270 us vs 73 us

Test:
To just decrypt first 32 bytes in rust instead of full note we can change:
`let note = Note::from_spender_encrypted(transmission_key, &shared_key, &self.encrypted_note).unwrap();`
to
```
let mut truncated_encrypted_text = [0u8; 32];
truncated_encrypted_text.copy_from_slice(&self.encrypted_note[..32]);
let partial: [u8; 32] = aead::decrypt_partial(&shared_key, &truncated_encrypted_text.to_vec()).try_into().unwrap();
jubjub::Fr::from_bytes(&partial).unwrap();
```

### DIFF
--- a/benchmarks/benches/merkle_note.rs
+++ b/benchmarks/benches/merkle_note.rs
@@ -28,7 +28,7 @@ pub fn decrypt_note_for_spender(c: &mut Criterion) {
             },
             // Benchmark
             |(ovk, merkle_note)| {
-                merkle_note.decrypt_note_for_spender(&ovk).unwrap();
+                merkle_note.decrypt_note_for_spender(&ovk).unwrap_err();
             },
             BatchSize::SmallInput,
         );
@@ -59,7 +59,7 @@ pub fn decrypt_note_for_owner(c: &mut Criterion) {
             },
             // Benchmark
             |(ivk, merkle_note)| {
-                merkle_note.decrypt_note_for_owner(&ivk).unwrap();
+                merkle_note.decrypt_note_for_owner(&ivk).unwrap_err();
             },
             BatchSize::SmallInput,
         );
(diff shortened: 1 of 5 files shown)


benchmarks/benches/merkle_note.rs
```diff
@@ -28,7 +28,7 @@ pub fn decrypt_note_for_spender(c: &mut Criterion) {
             },
             // Benchmark
             |(ovk, merkle_note)| {
```
jowparks, line 30: Done to avoid having to modify return types

ironfish-rust/src/serializing/aead.rs
```diff
@@ -58,10 +60,40 @@ pub(crate) fn decrypt<const SIZE: usize>(
...
+/// will subsequently be downloaded 
+pub fn decrypt_partial(
+    key: &[u8; 32],
+    truncated_ciphertext: &Vec<u8>,
+) -> Vec<u8> {
+    let mut truncated_encrypted_text = truncated_ciphertext.clone();
+    truncated_encrypted_text.copy_from_slice(&truncated_ciphertext[..]);
+
+    let mut keystream = ChaCha20::new(key.as_ref().into(), [0u8; 12][..].into());
+    keystream.seek(64);
```
jowparks, line 81: This seek value needs to change based on the position of the plaintext (for example if you want to decrypt bytes 32-64 instead of 0-32, this would need to seek to 96. Hardcoded here for test

PR #4201, opened 2023-08-10
Rahul/ifl 1499 conditionally start wallet based on flag

patnir:
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/start.ts
+++ b/ironfish-cli/src/commands/start.ts
@@ -210,6 +210,7 @@ export default class Start extends IronfishCommand {
     const blockGraffiti = this.sdk.config.get('blockGraffiti').trim() || null
     const peerPort = this.sdk.config.get('peerPort')
     const bootstraps = this.sdk.config.getArray('bootstrapNodes')
+    const walletEnabled = this.sdk.config.get('enableWallet')
 
     this.log(`\n${ONE_FISH_IMAGE}`)
     this.log(`Version       ${node.pkg.version} @ ${node.pkg.git}`)
@@ -217,7 +218,9 @@ export default class Start extends IronfishCommand {
     this.log(`Graffiti      ${blockGraffiti || 'NONE'}`)
     this.log(`Peer Identity ${node.peerNetwork.localPeer.publicIdentity}`)
     this.log(`Peer Agent    ${node.peerNetwork.localPeer.agent}`)
+    this.log(`Peer Agent    ${node.peerNetwork.localPeer.agent}`)
     this.log(`Peer Port     ${peerPort}`)
+    this.log(`Wallet        ${walletEnabled ? 'ENABLED' : 'DISABLED'}`)
     this.log(`Bootstrap     ${bootstraps.join(',') || 'NONE'}`)
     if (inspector.url()) {
       this.log(`Inspector     ${String(inspector.url())}`)
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -151,7 +151,7 @@ export class Wallet {
   }
 
   async updateHead(): Promise<void> {
-    if (this.scan || this.updateHeadState || this.accounts.size === 0) {
+    if (this.scan || this.updateHeadState || this.accounts.size === 0 || !this.isStarted) {
       return
     }
 
@@ -657,6 +657,11 @@ export class Wallet {
   async scanTransactions(fromHash?: Buffer): Promise<void> {
     if (!this.isOpen) {
       throw new Error('Cannot start a scan if accounts are not loaded')
+    }
+
+    if (!this.isStarted) {
+      this.logger.info('Skipping Scan, wallet is not started.')
+      return
     }
 
     if (this.scan) {
--- a/ironfish/src/node.ts
+++ b/ironfish/src/node.ts
@@ -357,6 +357,8 @@ export class FullNode {
 
     if (this.config.get('enableWallet')) {
       await this.wallet.start()
+    } else {
+      this.logger.info('Wallet is disabled')
     }
 
     this.peerNetwork.start()
--- a/ironfish/src/node.ts
+++ b/ironfish/src/node.ts
@@ -355,7 +355,10 @@ export class FullNode {
       this.metrics.start()
     }
 
-    await this.wallet.start()
+    if (this.config.get('enableWallet')) {
+      await this.wallet.start()
+    }
+
     this.peerNetwork.start()
 
     if (this.config.get('enableRpc')) {


ironfish-cli/src/commands/start.ts
```diff
@@ -218,6 +219,7 @@ export default class Start extends IronfishCommand {
     this.log(`Peer Identity ${node.peerNetwork.localPeer.publicIdentity}`)
     this.log(`Peer Agent    ${node.peerNetwork.localPeer.agent}`)
     this.log(`Peer Port     ${peerPort}`)
+    this.log(`Wallet        ${walletEnabled ? 'ENABLED' : 'DISABLED'}`)
```
patnir, line 222: happy to remove this, i thought it would be useful for the user

ironfish/src/testUtilities/nodeTest.ts
```diff
@@ -80,6 +80,7 @@ export class NodeTest {
     sdk.config.setOverride('enableTelemetry', false)
     sdk.config.setOverride('enableAssetVerification', false)
     sdk.config.setOverride('confirmations', 0)
+    sdk.config.setOverride('enableWallet', true)
```
rohanjadvani, line 83: Is this necessary since the default should be `true`?

ironfish/src/wallet/wallet.test.ts
```diff
@@ -315,9 +315,21 @@ describe('Accounts', () => {
   })
 
   describe('scanTransactions', () => {
+    it('should not rescan when wallet is not started', async () => {
+      const { node } = nodeTest
```
rohanjadvani, line 319: I think you can create a new `nodeTest` in these unit tests to be a bit cleaner
```suggestion
      const nodeTest = createNodeTest(false, { config: { enableWallet: false } })
      const { node } = nodeTest
```

Alternatively, you can stop the wallet

ironfish/src/wallet/wallet.test.ts
```diff
@@ -315,9 +315,21 @@ describe('Accounts', () => {
...
+
+      await node.wallet.scanTransactions()
+
+      expect(connectSpy).not.toHaveBeenCalled()
+    })
+
     it('should update head status', async () => {
       // G -> 1 -> 2
       const { node } = nodeTest
+      node.wallet['isStarted'] = true
```
rohanjadvani, line 332: This might not be needed depending on what you do in the above test.

ironfish/src/wallet/wallet.test.ts
```diff
@@ -389,6 +402,21 @@ describe('Accounts', () => {
       expect(wallet['chainProcessor']['hash']?.equals(block2.header.hash)).toBe(true)
     })
 
+    it('should not scan if wallet is not started', async () => {
```
rohanjadvani, line 405: What's the difference between this and the test above? Is it just the `shouldRescan` check?

ironfish/src/testUtilities/nodeTest.ts
patnir: my bad, removed

ironfish/src/wallet/wallet.test.ts
patnir: that is cleaner, changed

ironfish/src/wallet/wallet.test.ts
patnir: yeah, that's the only difference. I'll remove the above one

ironfish/src/wallet/wallet.ts
```diff
@@ -659,6 +659,11 @@ export class Wallet {
       throw new Error('Cannot start a scan if accounts are not loaded')
     }
 
+    if (!this.isStarted) {
```
hughy, line 662: This will break one of the uses of `ironfish wallet:rescan`:

If you run `ironfish wallet:rescan -f` without a node running it will start a rescan even though the wallet isn't started

We could check the config value here instead of `isStarted` 🤔

hughy (commented):
Another comment on rescans -- the `rescanAccount` RPC will reset account data in the wallet before starting to scan. We might want to check if we can scan before doing that reset so that it doesn't reset without starting to scan

ironfish/src/wallet/wallet.test.ts
rohanjadvani: Yeah, this should be started by default so I don't think you need it.

ironfish/src/wallet/wallet.ts
patnir: I see different behavior: 

image

This is what the node says: 

image

patnir:
> Another comment on rescans -- the `rescanAccount` RPC will reset account data in the wallet before starting to scan. We might want to check if we can scan before doing that reset so that it doesn't reset without starting to scan

@hughy is it better to use the flag in this case too and not the `isStarted` class member?

ironfish/src/wallet/wallet.ts
patnir: I only see 1 place in the code where `isStarted` is set: 

image

ironfish/src/wallet/wallet.test.ts
patnir: image

After putting some logs in and digging deeper, this test seems to call scanTransactions and the isStarted flag is false in that case. So, I do think we need it. Lmk what i'm missing

ironfish/src/wallet/wallet.ts
hughy: The rescan case I'm talking about is a bit different:

- `enableWallet` is set to `true`
- node is _not_ running
- run `ironfish wallet:rescan -f`

The wallet should scan in this case, but I'm not sure it will because `isStarted` is false

hughy:
> > Another comment on rescans -- the `rescanAccount` RPC will reset account data in the wallet before starting to scan. We might want to check if we can scan before doing that reset so that it doesn't reset without starting to scan
> 
> @hughy is it better to use the flag in this case too and not the `isStarted` class member?

Yeah, I think it needs to be the same as the check that we use in `scanTransactions` so that it doesn't reset without a scan unexpectedly

ironfish/src/wallet/wallet.test.ts
rohanjadvani: I missed an assumption on what was changed in the wallet file, but I think @hughy's comment below on the rescan case will overlap with this. I'm not sure if the started should be checked here.

patnir: I think I addressed @hughy 's comments and added an extra test for the case he described.

patnir:
> > > Another comment on rescans -- the `rescanAccount` RPC will reset account data in the wallet before starting to scan. We might want to check if we can scan before doing that reset so that it doesn't reset without starting to scan
> > 
> > 
> > @hughy is it better to use the flag in this case too and not the `isStarted` class member?
> 
> Yeah, I think it needs to be the same as the check that we use in `scanTransactions` so that it doesn't reset without a scan unexpectedly

Gotcha! I think I resolved it and added an extra test to check this case.

rohanjadvani (approved):

PR #4217, opened 2023-08-15
Add more information to getBlockResponse

hairtail:
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/chain/getBlock.ts
+++ b/ironfish/src/rpc/routes/chain/getBlock.ts
@@ -3,12 +3,14 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import * as yup from 'yup'
 import { Assert } from '../../../assert'
+import { getBlockSize, getTransactionSize } from '../../../network/utils/serializers'
 import { FullNode } from '../../../node'
 import { BlockHeader } from '../../../primitives'
 import { GENESIS_BLOCK_SEQUENCE } from '../../../primitives/block'
 import { BufferUtils } from '../../../utils'
 import { NotFoundError, ValidationError } from '../../adapters'
 import { ApiNamespace, routes } from '../router'
+import { RpcTransaction, RpcTransactionSchema } from './types'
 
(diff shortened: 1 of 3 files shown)


ironfish/src/rpc/routes/chain/getBlock.ts
```diff
@@ -148,14 +132,31 @@ routes.register<typeof GetBlockRequestSchema, GetBlockResponse>(
     const transactions: GetBlockResponse['block']['transactions'] = []
 
     for (const tx of block.transactions) {
-      const fee = tx.fee()
-
       transactions.push({
-        signature: tx.transactionSignature().toString('hex'),
```
danield9tqh, line 154: Can we add this signature field to `RpcTransaction` for better backwards compatibility of this endpoint

hairtail: sure

danield9tqh (approved):

danield9tqh:
@hairtail can you fix this one test that's failing? We change the response type of fee from a string to a number in this PR.
Test is here: https://github.com/iron-fish/ironfish/blob/0bb7033efe9e5408ef9d0eed693261b0bf785db9/ironfish/src/rpc/routes/chain/getBlock.test.ts#L131

I've checked in our codebase and can't find anything that depends on it being a string but we will put it in the release notes that this RPC route has changed. I think ideally it should be a string not a number since it's a bigint but we can save that migration for later down the line.

hairtail:
> @hairtail can you fix this one test that's failing? We change the response type of fee from a string to a number in this PR. Test is here:
> 
> https://github.com/iron-fish/ironfish/blob/0bb7033efe9e5408ef9d0eed693261b0bf785db9/ironfish/src/rpc/routes/chain/getBlock.test.ts#L131
> 
> I've checked in our codebase and can't find anything that depends on it being a string but we will put it in the release notes that this RPC route has changed. I think ideally it should be a string not a number since it's a bigint but we can save that migration for later down the line.

Done

PR #4222, opened 2023-08-16
Add a consensus parameter for asset ownership

mat-if:
## Summary

**Merging into a feature branch `asset-security`**

Adds a consensus parameter for enabling v2 transactions which enable the transferring of ownership of assets. 

Additionally adds a helper function for getting the correct transaction version for a specific block sequence.

Closes IFL-1330

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/defaultNetworkDefinitions.ts
+++ b/ironfish/src/defaultNetworkDefinitions.ts
@@ -21,7 +21,9 @@ export function defaultNetworkName(networkId: number): string | undefined {
   }
 }
 
-export const TESTNET = `{
+// TODO(IFL-1523): Update proper activation sequence for enableAssetOwnership
+export const TESTNET = `
+{
   "id": 0,
   "bootstrapNodes": ["1.test.bn.ironfish.network", "2.test.bn.ironfish.network"],
   "genesis": ${TESTNET_GENESIS},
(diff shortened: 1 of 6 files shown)


ygao76 (approved):

ironfish/src/defaultNetworkDefinitions.ts
```diff
@@ -61,6 +67,7 @@ export const DEVNET = `
         "targetBlockTimeInSeconds": 60,
         "targetBucketTimeInSeconds": 10,
         "maxBlockSizeBytes": 524288,
-        "minFee": 0
+        "minFee": 0,
+        "enableAssetOwnership": 9999999
```
andreacorbellini, line 71: I'd say that for DEVNET we can set it to 0 from the start. Or we can update this later, up to you

andreacorbellini (approved):

ironfish/src/defaultNetworkDefinitions.ts
mat-if: Yeah - that's the plan. We can't set it that low yet because the code doesn't exist to make the tests pass, and we'll have to regenerate all the test fixtures so it'll be a PR to follow soon. Here's the commit from the queue: https://github.com/iron-fish/ironfish/commit/8d763c075c91d143e38c3b7de105e478606d9a3a

PR #4223, opened 2023-08-16
adding enable wallet flag to node start command

patnir:
## Summary

Adding an `wallet` and `no-wallet` flag to control wallet rescan and note decryption

## Testing Plan

Add these flags when running the start command. Depending on these values, when you try to rescan transactions, it should fail/ pass. 
There is also an `enableWallet` config setting. Set that value and ensure that the flag that is set in the cli command overrides the config value. 

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/start.ts
+++ b/ironfish-cli/src/commands/start.ts
@@ -110,6 +110,11 @@ export default class Start extends IronfishCommand {
       description:
         'Path to a JSON file containing the network definition of a custom network to connect to',
     }),
+    enableWallet: Flags.boolean({
+      allowNo: true,
+      default: false,
+      description: 'Enable the wallet from syncing with the blockchain and decrypting notes.',
+    }),
   }
 
   node: IronfishNode | null = null
@@ -140,6 +145,7 @@ export default class Start extends IronfishCommand {
       upgrade,
       networkId,
       customNetwork,
+      enableWallet,
     } = flags
 
     if (bootstrap !== undefined) {
@@ -200,6 +206,9 @@ export default class Start extends IronfishCommand {
     if (!this.sdk.internal.get('telemetryNodeId')) {
       this.sdk.internal.set('telemetryNodeId', uuid())
       await this.sdk.internal.save()
+    }
+    if (enableWallet !== undefined && enableWallet !== this.sdk.config.get('enableWallet')) {
+      this.sdk.config.setOverride('enableWallet', enableWallet)
     }
 
     const privateIdentity = this.getPrivateIdentity()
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/start.ts
```diff
@@ -110,6 +110,11 @@ export default class Start extends IronfishCommand {
       description:
         'Path to a JSON file containing the network definition of a custom network to connect to',
     }),
+    enableWallet: Flags.boolean({
```
rohanjadvani, line 113:
```suggestion
    wallet: Flags.boolean({
```

`--no-enableWallet` reads a bit weird. Let's make this a little cleaner

ironfish-cli/src/commands/start.ts
```diff
@@ -110,6 +110,11 @@ export default class Start extends IronfishCommand {
...
+      allowNo: true,
+      default: false,
```
rohanjadvani, line 115: Should this be true by default? If it's `false`, then the current behavior won't hold true anymore.

ironfish-cli/src/commands/start.ts
```diff
@@ -201,6 +207,9 @@ export default class Start extends IronfishCommand {
       this.sdk.internal.set('telemetryNodeId', uuid())
       await this.sdk.internal.save()
     }
+    if (enableWallet !== undefined && enableWallet !== this.sdk.config.get('enableWallet')) {
```
rohanjadvani, line 210: nit: Can this be moved above the conditional? It's a little easier to see where all the config overrides are happening together.

ironfish-cli/src/commands/start.ts
```diff
@@ -110,6 +110,11 @@ export default class Start extends IronfishCommand {
       description:
         'Path to a JSON file containing the network definition of a custom network to connect to',
     }),
+    enableWallet: Flags.boolean({
+      allowNo: true,
+      default: false,
+      description: 'Enable the wallet from syncing with the blockchain and decrypting notes.',
```
rohanjadvani, line 116:
```suggestion
      description: `Enable the node's wallet to scan transactions and decrypt notes from the blockchain`,
```

rohanjadvani (changes requested):

ironfish-cli/src/commands/start.ts
patnir: Done, sorry about that. I should have been more careful

ironfish-cli/src/commands/start.ts
patnir: Done

rohanjadvani (approved):

PR #4226, opened 2023-08-16
Rahul/ifl 1498 show enabled state in wallet status

patnir:
## Summary
Adding the wallet enabled state to the status command: 

https://github.com/iron-fish/ironfish/assets/13268167/6c76bfe9-8a81-421f-a13c-9851347e4d3d

## Testing Plan

1. Set the config enableWallet flag to false/ true 
2. Run `ironfish status` and you should see the state reflected 

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/status.ts
+++ b/ironfish-cli/src/commands/status.ts
@@ -191,6 +191,7 @@ function renderStatus(content: GetNodeStatusResponse, debugOutput: boolean): str
       accountStatus += ` - ${content.accounts.scanning.sequence} / ${content.accounts.scanning.endSequence}`
     }
   }
+  const walletEnabled: boolean = content.wallet.enabled
 
   const cores = `Cores: ${content.cpu.cores}`
   const current = `Current: ${content.cpu.percentCurrent.toFixed(1)}%`
@@ -202,6 +203,7 @@ function renderStatus(content: GetNodeStatusResponse, debugOutput: boolean): str
   return `\
 Version              ${content.node.version} @ ${content.node.git}
 Node                 ${nodeStatus}
+Wallet               ${walletEnabled ? 'ENABLED' : 'DISABLED'}  
 Node Name            ${content.node.nodeName}
 Peer ID              ${content.peerNetwork.publicIdentity}
 Block Graffiti       ${blockGraffiti}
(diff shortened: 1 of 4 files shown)


ironfish/src/rpc/routes/node/getStatus.ts
```diff
@@ -368,6 +376,9 @@ async function getStatus(node: FullNode): Promise<GetNodeStatusResponse> {
         sequence: node.wallet.chainProcessor.sequence ?? -1,
       },
     },
+    wallet: {
+      enabled: node.wallet.isEnabled(),
```
rohanjadvani, line 380: Can this just be `config.get('enableWallet')`?

ironfish-cli/src/commands/status.ts
```diff
@@ -202,6 +203,7 @@ function renderStatus(content: GetNodeStatusResponse, debugOutput: boolean): str
   return `\
 Version              ${content.node.version} @ ${content.node.git}
 Node                 ${nodeStatus}
+Wallet               ${walletEnabled ? 'ENABLED' : 'DISABLED'}
```
rohanjadvani, line 206: There is an `accountStatus` below, which is technically the wallet's chain processor state. Do you think it'd be worth to combine this somehow?

patnir: Ill think about that one

ironfish/src/rpc/routes/node/getStatus.ts
patnir: ah i thought this class didn't have access to config, i'll try that

patnir: I don't think file has access to it

ironfish-cli/src/commands/status.ts
patnir: Added the disabled marker to the account status as well

rohanjadvani: I think it's a little odd that we'll have two lines duplicate the same DISABLED information. What if we just update the `accountStatus`? Additionally, we could also rename `Accounts` -> `Wallet`

rohanjadvani (commented):
I think this will need a documentation update too

ironfish-cli/src/commands/status.ts
patnir: Sweet, that works.

rohanjadvani (approved):

PR #4227, opened 2023-08-16
`transaction_signature_hash` functions return a result instead of panic

mat-if:
## Summary

Previously, these functions would call `.unwrap()` on Results instead of bubbling up the error. These functions now properly bubble up the error instead of using unwraps and panics.

Closes IFL-1522

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


jowparks (approved):
No more unwraps 🗡️

PR #4234, opened 2023-08-22
add v2 support to RawTransaction

mat-if:
## Summary

**Merges into asset-security branch**

- Adds a parameter to `RawTransaction` constructor: `transactionVersion`. It will create a v1 or v2 transaction based on the given parameter. 
- Updates the wallet so it will create the appropriate transaction version as well.

## Testing Plan

Unit tests

## Breaking Change

```
[x] Yes
```
Changes the format of RawTransaction. Noted in the asset-security related breaking changes tracking ticket


ironfish/src/primitives/rawTransaction.ts
```diff
@@ -197,6 +202,14 @@ export class RawTransactionSerde {
       bw.writeVarString(mint.name, 'utf8')
       bw.writeVarString(mint.metadata, 'utf8')
       bw.writeBigU64(mint.value)
+      if (TransactionFeatures.hasMintTransferOwnershipTo(raw.version)) {
```
andreacorbellini, line 205: I would throw an error if `transferOwnershipTo` is set but not supported

mat-if: Added: https://github.com/iron-fish/ironfish/pull/4234/commits/bb31f2c3fe33388604fe622fa98ac3b3b1938ca7

andreacorbellini (approved):

PR #4237, opened 2023-08-22
use an optimistic transaction version when creating transactions

mat-if:
## Summary

**Builds on #4236**

This allows for transactions created around the time that a transaction version change to have an improved chance of being included in the blockchain by choosing the version if its within a range near the current head sequence instead of simply choosing the transaction version based on head sequence + 1

Closes IFL-1527

## Testing Plan

Unit tests


ironfish/src/utils/transaction.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Returns the sequence delta to use when checking which TransactionVersion to
+ * use for creating a new Transaction. Optimistically chooses a higher value to
+ * give the Transaction a better chance of being accepted when there is a
+ * version change happening soon.
+ */
+function versionSequenceDelta(expirationDelta?: number): number {
+  if (expirationDelta) {
+    // If the expirationDelta is less than the minimum, we must use that one
+    if (expirationDelta < MIN_VERSION_DELTA) {
+      return Math.max(1, expirationDelta)
```
andreacorbellini, line 18: Wouldn't this be always `1`?

`expirationDelta` (if I understood correctly) is a non-negative integer. The `if (expirationDelta)` at the top ensures it's non-zero, hence when it gets here it's a positive integer. `max(1, positiveInteger)` should always evaluate to `1`. Am I missing something?

mat-if: I may be misunderstanding, but `max(1, positiveInteger)` should always evaluate to `positiveInteger`, technically. To be overly descriptive, it would be `1` instead of `positiveInteger` if `positiveInteger` is also `1`, but I think the point remains the same. `min(1, positiveInteger)` would indeed always evaluate to `1`, though.

It is _possible_ to call this function with a negative number, though, so I felt it better to be defensive here. `expirationDelta` _should_ be a positive integer, but technically there's no validation on `createTransaction`'s `expirationDelta` parameter enforcing a positive integer. The config setting `transactionExpirationDelta` does enforce positive integer.

andreacorbellini: Right, sorry, I meant that it will always be `expirationDelta`. The core of the issue is that I don't fully understand what this method is supposed to do, and what its input should represent. Perhaps improving the comments and adding some assertions would be nice (e.g. if `expirationDelta` is supposed to be a positive integer, I'd add an assertion for that rather than using `max(1, expirationDelta)`--why would a negative `expirationDelta` be treated as 1 rather than 0?)

ironfish/src/utils/transaction.ts
```diff
@@ -0,0 +1,31 @@
...
+    } else {
+      // Otherwise, use half of the expirationDelta, clamped between
+      // MIN_VERSION_DELTA and MAX_VERSION_DELTA
```
andreacorbellini, line 21: I'm not really sure I understand the clamping. Let's suppose V2 transactions are getting activated at block 101. We're at block 90. I post a transaction with expiration = 30 (twice the default expiration, so a reasonable number in my opinion). Note that we have room for 10 blocks that accept V1 transactions (blocks 91 to 100), and room for 20 blocks that accept V2 transactions (blocks 101 to 120; after that, the transaction expires). Just from eyeballing these numbers it's apparent that the transaction has higher chances to get mined as a V2 transaction, because there are more suitable blocks for it.

But we can still try to calculate a probability for it. I'm not sure if the mining manager takes into account the age of a transaction, but I'll assume it does not. I'll also assume that the rate of arrival of transactions in the mempool equals the rate of transactions mined (i.e. for *n* transactions that get mined, on average, *n* new transactions will arrive in the mempool). With these assumptions, the probability of our transaction getting selected for an arbitrary block *b* is 1/*n*, uniformly distributed, where *n* is the average number of transactions in the mempool (which should be more or less constant as per my assumption).

If *n* = 1 (only 1 transaction in the mempool), then it will be mined immediately (this is not necessarily true, but let's assume it is for simplicity), therefore it makes sense for the transaction to be a V1 transaction.

If *n* != 1, then the probability of the transaction NOT being picked in the first 10 blocks is ((*n*−1)/*n*)^10, so the probability of the transaction getting mined if posted as a V1 transaction is 1−((*n*−1)/*n*)^10. Similarly, the probability of the transaction getting mined if posted as a V2 transaction is 1−((*n*−1)/*n*)^20. The V2 probability is always higher than the V1 probability (except in the *n* = 1 case), so it makes sense for this transaction to be a V2 transaction.

Last but not least, if we had to consider the fact that broadcasting mempool transactions has some inherent delays, the V2 probability would become even higher.

Yet, this method will tell me that I should target block 90 + (30 / 2)|_3^10 = 90 + 10 = 100, which would make this a V1 transaction.

I know this is a very long comment; the intent of it is not to criticize the logic of this method, but rather to get some clarity, because from the way I'm reasoning I cannot grasp it. It's clear that this method is making some assumptions (e.g. I think it's assuming that the number *n* of mempool transactions is low), but it's not clear to me what these assumptions are, and where numbers like `MIN/MAX_VERSION_DELTA` come from. Could you please add more comments or assertions somewhere to explain the logic behind this method?

andreacorbellini: (The actual probability model is actually much more complex than what I wrote, and in particular does not take into account how many transactions can be included in a single block, but again: my intent is not to discuss probabilities, but rather to ask clarifications about the logic behind this method.)

ironfish/src/utils/transaction.ts
```diff
@@ -0,0 +1,31 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+const MAX_VERSION_DELTA = 10
+const MIN_VERSION_DELTA = 3
```
andreacorbellini, line 6: Nit: these are called "version", but they're actually block sequence deltas.

ironfish/src/utils/transaction.ts
mat-if: I appreciate the feedback! I did feel it was too vague when I put up the PR, but having been too close to it for the past couple weeks, I was struggling to take that step back for clarity, so this is very helpful.

> It's clear that this method is making some assumptions (e.g. I think it's assuming that the number n of mempool transactions is low)

Exactly. Given the mem pool has an average of < 5 transactions right now, and the average transactions in a block right now is probably somewhere between 1-2 (https://explorer.ironfish.network/), I was optimizing for the amount of time a transaction might spend sitting in the mem pool. The assumption was being made that a transaction _will_ be picked up, so the best course of action was to minimize the amount of time it waited.

The minimum of 3 was chosen because a transaction getting picked up within 3 blocks prior to it being posted is pretty low - given time to actually post (on a fast computer and a small transaction, this can be <10 seconds, but on a more modest system and/or a larger transaction, this could be many minutes) + time to broadcast the transaction, it seems unlikely that 3 is even really enough now that I'm saying it out loud. Indeed, these values were chosen mostly off the top of my head, so I appreciate the ask for clarity.

The maximum of 10 was chosen because of the aforementioned assumptions, so that the most a transaction will wait around is roughly 10 blocks near a hard-fork activation.

While this function and all the transaction version logic should work for a potential upgrade to V3, V4, etc. it is likely that we may want to make this particular function smarter at those times, given the probabilities you outlined. We could make improvements now, too, I'm certainly not against that.

andreacorbellini (approved):

ironfish/src/utils/transaction.ts
andreacorbellini: Sounds good to me, thanks for the detailed explanation.

I would just like to point out that these assumptions are valid today, but this function is going to stay in the codebase for the longer future, and maybe when the time for V3 transaction comes, we will have plenty of traffic on the mempool, so it will be appropriate to revisit it. But LGTM with today's numbers

PR #4240, opened 2023-08-23
remove hardcoded transaction version const

mat-if:
## Summary

Removes `TRANSACTION_VERSION` const, since it no longer makes sense as a default value.

Closes IFL-1530

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust-nodejs/tests/demo.test.slow.ts
+++ b/ironfish-rust-nodejs/tests/demo.test.slow.ts
@@ -58,7 +58,7 @@ describe('Demonstrate the Sapling API', () => {
   it(`Should create a miner's fee transaction`, () => {
     const key = generateKey()
 
-    const transaction = new Transaction(key.spendingKey)
+    const transaction = new Transaction(key.spendingKey, 2)
     const note = new Note(key.publicAddress, 20n, 'test', Asset.nativeId(), key.publicAddress)
     transaction.output(note)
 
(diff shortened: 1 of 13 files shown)


ironfish-rust/src/transaction/tests.rs
```diff
@@ -71,7 +71,7 @@ fn test_transaction() {
         spender_key.public_address(),
     );
 
-    let mut transaction = ProposedTransaction::new(spender_key);
+    let mut transaction = ProposedTransaction::new(spender_key, TransactionVersion::V2);
```
andreacorbellini, line 74: I'm wondering if it would be beneficial for this kind of tests to have a `TransactionVersion::latest()` accessor? When we introduce V3 transactions, I *think* we will want these tests to use V3.

We could also parametrize these tests (making them run for all possible versions), but this would be out of scope for this PR.

ironfish-rust-nodejs/tests/demo.test.slow.ts
```diff
@@ -58,7 +58,7 @@ describe('Demonstrate the Sapling API', () => {
   it(`Should create a miner's fee transaction`, () => {
     const key = generateKey()
 
-    const transaction = new Transaction(key.spendingKey)
+    const transaction = new Transaction(key.spendingKey, 2)
```
andreacorbellini, line 61: Similar comment as the one below: would it make sense to use a `LATEST_TRANSACTION_VERSION` (or similar) for use in tests?

andreacorbellini (approved):

ironfish-rust/src/transaction/tests.rs
mat-if: Great idea - added

PR #4244, opened 2023-08-24
When an unknown asset is seen in a wallet transaction, backfill its information from the chain db before processing the transaction

andreacorbellini:
## Summary

This commit is purely a refactor of the current wallet logic, and should not alter the user-visible behavior of the wallet in any way. This refactor is foundational work needed to support the transfer of asset ownership (via `transferOwnershipTo`) in the future.

What this commit does, in a nutshell, is that it makes the backfilling code run before the transaction gets processed:

    before this change: account.addPendingTransaction(...) -> backfill assets
     after this change: backfill assets -> account.addPendingTransaction(...)

This will greatly simplify the processing of mints with `transferOwnershipTo` inside `addPendingTransaction`.

## Testing Plan

Existing tests should not break.

## Documentation

No doc change required.

## Breaking Change

Not a breaking change.

### DIFF
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -16,6 +16,7 @@ import { Witness } from '../merkletree/witness'
 import { Mutex } from '../mutex'
 import { GENESIS_BLOCK_SEQUENCE } from '../primitives'
 import { BurnDescription } from '../primitives/burnDescription'
+import { MintDescription } from '../primitives/mintDescription'
 import { Note } from '../primitives/note'
 import { NoteEncrypted } from '../primitives/noteEncrypted'
 import { MintData, RawTransaction } from '../primitives/rawTransaction'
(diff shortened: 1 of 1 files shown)


ironfish/src/wallet/account/account.ts
```diff
@@ -251,6 +251,10 @@ export class Account {
       return
     }
 
+    if (owner.toString('hex') !== this.publicAddress) {
```
mat-if, line 254: I _think_ this is incorrect, and I'm surprised this doesn't have a failing test associated with it. If I'm not mistaken, the point of it syncing from the decrypted notes is because the wallet may want to fetch the info in the event of the account receiving a note of an asset, even if it isn't the owner.

As an example, I mint 5 Matcoin which creates a note of 5 Matcoin. Later in a different transaction, I send you 1 Matcoin, so the transaction will have 1 spend: my 5 Matcoin note, and 2 outputs/notes: 1. 1 Matcoin for you and 2. 4 Matcoin for me.

In this scenario, your wallet wants to fetch the asset info so it can display e.g. the asset name (Matcoin), even though it doesn't own the asset.

ironfish/src/wallet/wallet.ts
```diff
@@ -565,35 +566,77 @@ export class Wallet {
...
+   * be found, then the information is copied from the chain db into the wallet
+   * db.
+   */
+  private async backfillAssets(
+    account: Account,
+    decryptedNotes: DecryptedNote[],
+    mints: MintDescription[],
+    tx?: IDatabaseTransaction,
+  ): Promise<void> {
+    const assetIds: Set<Buffer> = new Set()
```
mat-if, line 594: We generally use `BufferSet` (ex: https://github.com/iron-fish/ironfish/blob/master/ironfish/src/consensus/verifier.ts#L85) instead of `Set<Buffer>`, but I realize I'm not sure if its necessary or not. I suspect `Set<Buffer>` could suffer from comparison issues that Buffers generally have, though. 

To summarize, we discovered that you can't use basic equality checks with Buffers - they don't always do what one would expect.
Related:
- https://github.com/iron-fish/ironfish/pull/1094
- https://github.com/iron-fish/ironfish/pull/1114
- https://stackoverflow.com/questions/30701220/how-to-compare-buffer-objects-in-nodejs

ironfish/src/wallet/account/account.ts
andreacorbellini: Right, not sure how this escaped me, but I think this proves that our test coverage in this case is not as through as I thought. I will add more tests to catch this exact scenario

ironfish/src/wallet/wallet.ts
andreacorbellini: Also right. I actually planned to replace this set with a different thing but then I forgot. Thanks for the pointer to BufferSet--that spared me some time researching alternative. FWIW, the reason why Set is unsuitable is that Buffer is unhashable, so adding Buffers to a Set does not do any "de-duplication" work.

ironfish/src/wallet/account/account.ts
andreacorbellini: Actually, this code as-is was correct, let me explain why. First of all, let me say that there is a test covering this scenario: it's `should save assets from the chain the account does not own`.

Your explanation is correct: when you receive a note from an unknown asset, that asset should be recorded in the wallet db, backfilling it from the chain db. This job is done by `upsertAssetsFromDecryptedNotes` which does the following (both now, in staging, and with this commit):

1. checks if the asset is in the wallet db
2. if it's not, gets the asset from the chain db
3. updates the asset on the wallet db with the latest block information

In staging, `upsertAssetsFromDecryptedNotes` writes to the wallet db twice: at step 2 and 3. With the original code from this PR, the wallet db is touched only at step 3.

I have updated this PR to re-add the write at step 2, but now that I dug deeper into the issue, I'm thinking of rolling back the change, as writing twice is unnecessary and a waste of resources. I see however how this can generate confusion, so I'll make sure that the code is properly documented.

mat-if (approved):

PR #4253, opened 2023-08-28
Update the blockchain db whenever a mint with `transferOwnershipTo` is seen

andreacorbellini:
## Summary

When a `MintDescription` with `transferOwnershipTo` is seen, the relevant asset on the chain db is updated to reflect the new owner.

## Testing Plan

Unit tests. Once the RPC changes are in place, manual testing will be performed as well.

## Documentation

No doc change required

## Breaking Change

Not a breaking change


ironfish/src/assert.ts
```diff
@@ -91,4 +91,13 @@ export class Assert {
       throw new Error(`We must have forgotten a falsey value: ${String(x)}`)
     }
   }
+
+  static bufferEquals(x: Buffer, y: Buffer, message?: string): void {
```
andreacorbellini, line 95: I looked around if a similar function already exists but couldn't find any. I wouldn't be surprised though if there's already something similar

mat-if (approved):

PR #4268, opened 2023-09-01
Rahul/ifl 1587 createaccount endpoint

patnir:
## Summary

This PR is part of the RPC normalization work. This endpoint will replace the `wallet/create` endpoint
https://www.notion.so/Final-Proposal-RPC-Normalization-89c16245de2e4b4eaa4bc528e3c1f2d3?pvs=4

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/wallet/index.ts
```diff
@@ -5,14 +5,15 @@
 export * from './addTransaction'
 export * from './burnAsset'
 export * from './create'
+export * from './createAccount'
 export * from './createTransaction'
```
patnir, line 9: sorting of endpoints

ironfish/src/rpc/routes/wallet/create.ts
```diff
@@ -1,31 +1,17 @@
...
-    name: yup.string().defined(),
-    default: yup.boolean().optional(),
-  })
-  .defined()
+/**
+ * NOTE: This endpoint will be deprecated in favor of `POST /wallet/createAccount` because
+ * this endpoint does not follow the convention that all of our endpoints should follow which
+ * is the verbObject naming convention, where the verb is the HTTP verb and the object is the
+ * object being acted upon. For example, `POST /wallet/burnAsset` burns an asset.
+ */
```
hughy, line 10: nit - do we want to use the HTTP verbs in documentation since our endpoints aren't necessarily used over HTTP?

ironfish/src/rpc/routes/wallet/index.ts
hughy: a bit of an aside, but I wonder if there's a way that we can get lint to enforce this. otherwise I fear we're likely to add a new endpoint out of order at some point in the future 😅

hughy (approved):

ironfish/src/rpc/routes/wallet/create.ts
patnir: Done, changed it to remove the http language

PR #4269, opened 2023-09-01
resets account heads and birthdays in reset command

hughy:
## Summary

the 'ironfish reset' command deletes the chain database and most wallet datastores

however, it does not clear the 'heads' store. this prevents the wallet from starting the next time the node starts because the account head will reference a block that is missing from the chain

updates 'ironfish reset' to use 'wallet.reset' to reset all database stores, including the heads store

adds arguments to reset, resetAccounts to optionally reset the 'createdAt' field on each account

resets account 'createdAt' field in 'ironfish reset' if the command is used to change from one network to another because the createdAt field will reference a block on a different chain after changing networks

## Testing Plan

- manual testing:
  - used `ironfish reset --networkId=` to reset and change networks
  - used `ironfish start` to start the node

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/reset.ts
+++ b/ironfish-cli/src/commands/reset.ts
@@ -74,20 +74,17 @@ export default class Reset extends IronfishCommand {
       fsAsync.rm(hostFilePath, { recursive: true, force: true }),
     ])
 
+    let networkChanged = false
     if (flags.networkId != null && flags.networkId !== existingId) {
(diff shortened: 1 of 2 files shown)


patnir (approved):
Looks good. On staging after running reset: 

image


When I swtiched to this branch and ran reset: 

image

PR #4270, opened 2023-09-05
Track originating peer for block gossip

danield9tqh:
## Summary
Track originating peer for block gossip through telemetry

## Testing Plan
Unit tests + local testing

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/peerNetwork.ts
+++ b/ironfish/src/network/peerNetwork.ts
@@ -1198,14 +1198,13 @@ export class PeerNetwork {
       return
     }
 
-    const firstPeer = this.blockFetcher.firstSeenBy(block.header.hash)
-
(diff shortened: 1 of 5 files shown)


ironfish/src/network/peerNetwork.ts
```diff
@@ -1198,13 +1198,15 @@ export class PeerNetwork {
       return
     }
 
+    const firstPeer = this.blockFetcher.firstSeenBy(block.header.hash)
```
dguenther, line 1201: Think it would make sense to move this down to right above the telemetry line? Since for now we're only using it for that

dguenther (approved):

PR #4272, opened 2023-09-07
Rahul/ifl 1603 make asset response consistent across all endpoints

patnir:
## Summary

This PR kicks off the RPC normalization work. Here, we specifically aim to normalize the response objects for all APIs related to an `asset`. 

We're going to adhere to the following principles to create a consistent RPC experience going forward: 
1. Return a complete object or just the ID of the object, not as one off fields
2. Lean on the side of returning as few fields as possible, it’s easier to add later than to remove remove later

Going back to fix consistency issues without causing breaking changes is not feasible, but these changes demonstrate the application of the above principles to the asset object in our API responses. 

In the example below, we have a new asset response object in the burn asset response, eventually the `assetId` and `name` fields will be deprecated. 
image

## Testing Plan
Added checks for the new fields

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/wallet/burnAsset.ts
```diff
@@ -71,6 +107,18 @@ routes.register<typeof BurnAssetRequestSchema, BurnAssetResponse>(
     const burn = transaction.burns[0]
 
     request.end({
+      asset: {
```
danield9tqh, line 110: I think we would want to have this be the final version of the API. Do we want each response to be nested like
```
{
  asset: {...}
}
```

This could make sense here since its the burnAsset endpoint and there could be other fields like `success` or something. If it were `getAsset` I would say we would not want the nesting though

patnir: I agree we don't want nesting on getAsset but we do here

ironfish/src/rpc/routes/wallet/getBalance.ts
```diff
@@ -42,6 +47,7 @@ export const GetBalanceResponseSchema: yup.ObjectSchema<GetBalanceResponse> = yu
   .object({
     account: yup.string().defined(),
     assetId: yup.string().defined(),
+    asset: RpcAssetSchema.optional(),
     assetVerification: yup
       .object({ status: yup.string().oneOf(['verified', 'unverified', 'unknown']).defined() })
       .defined(),
```
patnir, line 53: can return in the RpcAsset

ironfish/src/rpc/routes/wallet/getBalances.ts
```diff
@@ -86,6 +101,7 @@ routes.register<typeof GetBalancesRequestSchema, GetBalancesResponse>(
       const asset = await account.getAsset(balance.assetId)
 
       balances.push({
+        asset: asset ? constructRpcAsset(asset) : undefined,
         assetId: balance.assetId.toString('hex'),
         assetName: asset?.name.toString('hex') ?? '',
         assetCreator: asset?.creator.toString('hex') ?? '',
```
patnir, line 107: Look for part of the code where transactions and assets get added to the wallet db

ironfish/src/rpc/routes/wallet/getAccountTransaction.ts
```diff
@@ -35,7 +40,7 @@ export type GetAccountTransactionResponse = {
     burnsCount: number
     timestamp: number
     submittedSequence: number
-    assetBalanceDeltas: Array<{ assetId: string; assetName: string; delta: string }>
+    assetBalanceDeltas: RcpAccountAssetBalanceDelta[]
     notes: RpcWalletNote[]
```
patnir, line 44: using an already defined object

ironfish/src/rpc/routes/wallet/getAccountTransactions.ts
```diff
@@ -43,7 +48,7 @@ export type GetAccountTransactionsResponse = {
   expiration: number
   timestamp: number
   submittedSequence: number
-  assetBalanceDeltas: Array<{ assetId: string; assetName: string; delta: string }>
+  assetBalanceDeltas: RcpAccountAssetBalanceDelta[]
```
patnir, line 51: using an already defined object

ironfish/src/rpc/routes/wallet/burnAsset.test.ts
```diff
@@ -74,6 +74,14 @@ describe('Route wallet/burnAsset', () => {
       })
       jest.spyOn(wallet, 'burn').mockResolvedValueOnce(burnTransaction)
 
+      const accountAsset = await account.getAsset(assetId)
+
+      expect(accountAsset).toBeDefined()
+
+      if (!accountAsset) {
+        throw new Error('accountAsset is undefined')
+      }
```
patnir, line 83: references below line to accountAsset would not respect  the `expect(accountAsset).toBeDefined()` line as an undefined check and lead to a bunch of unnecessary conditional logic. So, I added this check here. I'm happy to change it back

ironfish/src/rpc/routes/chain/getAsset.ts
```diff
@@ -7,22 +7,14 @@ import { Assert } from '../../../assert'
...
-  createdTransactionHash: string
-  id: string
-  metadata: string
-  name: string
-  nonce: number
-  creator: string
-  owner: string
-  supply: string
-}
+export type GetAssetResponse = Omit<RpcAsset, 'status'>
```
patnir, line 17: I believe that the chain doesn't have access to the status field, but lmk if I'm wrong. If that's the case, it would simplify everything to 1 asset model. Alternatively, I can return status unknown, but I figured it would be more confusing for the users.

hughy: `status` is computed based on the sequence of the block that includes the transaction that created the asset

I think we can compute it in this endpoint to simplify to a single asset model

hughy (approved):

PR #4275, opened 2023-09-07
Add (de)serialization for `Asset`

andreacorbellini:
## Summary

Add a simple test to check that the serialization/deserialization for `Asset` in Rust works correctly.

## Documentation

No doc changes required.

## Breaking Change

Not a breaking change.

### DIFF
--- a/ironfish-rust/Cargo.toml
+++ b/ironfish-rust/Cargo.toml
@@ -50,6 +50,9 @@ rand = "0.8.5"
 tiny-bip39 = "0.8"
 xxhash-rust = { version = "0.8.5", features = ["xxh3"] }
 
+[dev-dependencies]
+hex-literal = "0.4"
+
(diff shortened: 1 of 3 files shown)


ironfish-rust/Cargo.toml
```diff
@@ -42,6 +42,7 @@ chacha20poly1305 = "0.9.0"
 crypto_box = { version = "0.8", features = ["std"] }
 ff = "0.12.0"
 group = "0.12.0"
+hex-literal = "0.4"
```
mat-if, line 45: Should this be a dev-dependency?

mat-if (approved):

ironfish-rust/Cargo.toml
andreacorbellini: Yes, oops, not sure how it ended up there

PR #4277, opened 2023-09-07
Make activation flag nullable

ygao76:
## Summary
Existing consensus activation flag has type of number, the issue is before a block sequence is picked for the activation we need to have a fake placeholder for the activation block sequence. 
This PR make the flag type nullable, so we can leave the block sequence as null before we decide on the actual sequence.

## Testing Plan
unit test
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```
@mat-if @andreacorbellini

### DIFF
--- a/ironfish/src/consensus/consensus.ts
+++ b/ironfish/src/consensus/consensus.ts
@@ -39,7 +39,7 @@ export type ConsensusParameters = {
   /**
    * The block height that enables the use of V2 transactions instead of V1
    */
-  enableAssetOwnership: number
+  enableAssetOwnership: number | null
 }
 
 export class Consensus {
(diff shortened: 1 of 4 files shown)


ironfish/src/consensus/consensus.ts
```diff
@@ -39,7 +39,7 @@ export type ConsensusParameters = {
   /**
    * The block height that enables the use of V2 transactions instead of V1
    */
-  enableAssetOwnership: number
+  enableAssetOwnership: number | null
```
andreacorbellini, line 42: Nit: you can consider using a constant like `'never'` instead of `null` to make it even more explicit:

```
enableAssetOwnership: number | 'never'
```

andreacorbellini (approved):

ironfish/src/consensus/consensus.ts
ygao76: Thats good idea, the schema will be a little complicated but its doable. I will update.

ygao76: Updated PR

PR #4282, opened 2023-09-13
Use Decimal library to make feeRates deterministic

danield9tqh:
## Summary
When calculating a feeRate for a transaction we are using BigInt even though transaction feeRates can potentially be a decimal value which BigInt cannot represent. For example the fee / kb of the following transactions are the same even though one has a higher fee

Fee: 1
Bytes: 2500

FeeRate (kB): `(BigInt(1) * BigInt(1000)) / BigInt(2500) = 1 // rounded up`

Fee: 2
Bytes: 2500

FeeRate (kB): `(BigInt(2) * BigInt(1000)) / BigInt(2500) = 1 // rounded up`

This has led to any fee below a certain amount being ~4 ore being treated the same in mempool sorting

## Testing Plan
- Existing mempool unit tests + new unit tests for Decimal library
- Performance tests done on mempool insertion

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/memPool/memPool.test.ts
+++ b/ironfish/src/memPool/memPool.test.ts
@@ -11,6 +11,7 @@ import { TransactionVersion } from '../primitives/transaction'
 import {
   createNodeTest,
   useAccountFixture,
+  useBlockWithCustomTxs,
   useBlockWithTx,
   useMinerBlockFixture,
   useTxFixture,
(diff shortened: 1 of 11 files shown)


ironfish/src/memPool/memPool.test.ts
```diff
@@ -140,6 +142,37 @@ describe('MemPool', () => {
   describe('orderedTransactions', () => {
     const nodeTest = createNodeTest()
 
+    it('returns transactions sorted by fee rate determinsitically', async () => {
```
dguenther, line 145:
```suggestion
    it('returns transactions sorted by fee rate deterministically', async () => {
```

ironfish/src/memPool/memPool.test.ts
```diff
@@ -10,12 +11,13 @@ import { TransactionVersion } from '../primitives/transaction'
...
   createNodeTest,
   useAccountFixture,
+  useBlockWithCustomTxs,
   useBlockWithTx,
   useMinerBlockFixture,
   useTxFixture,
 } from '../testUtilities'
 import { Account } from '../wallet'
-import { getFeeRate } from './feeEstimator'
+import { getFeeRate, getPreciseFeeRate } from './feeEstimator'
```
dguenther, line 20: Is `getFeeRate` effectively deprecated now, or do we still need it somewhere?

dguenther (approved):

ironfish/src/memPool/memPool.test.ts
danield9tqh: We still use it in the CLI. Planning on fully removing it in future PR

PR #4286, opened 2023-09-14
Rahul/ifl 1576 normalize block header

patnir:
## Summary

Normalizing RpcBlockHeader across 5 APIs. 

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/types.ts
```diff
@@ -54,3 +56,59 @@ export const RpcEncryptedNoteSchema: yup.ObjectSchema<RpcEncryptedNote> = yup
...
+  previousBlockHash: string
+  difficulty: string
+  noteCommitment: string
+  transactionCommitment: string
+  target: string
+  randomness: string
+  timestamp: number
+  graffiti: string
+  work: string
+  noteSize?: number
```
hughy, line 76: this is a little bit inconsistent with the `BlockHeader` type, where `noteSize` has type `number | null`

ironfish/src/wallet/remoteChainProcessor.ts
```diff
@@ -97,6 +97,7 @@ export class RemoteChainProcessor {
 
     Assert.isNotNull(response.block.noteSize)
     let initialNoteIndex = response.block.noteSize
+    Assert.isNotUndefined(initialNoteIndex)
```
hughy, line 100: this assertion won't be necessary if `noteSize` is nullable instead of optional

ironfish/src/rpc/routes/chain/getBlock.ts
```diff
@@ -167,18 +154,14 @@ routes.register<typeof GetBlockRequestSchema, GetBlockResponse>(
...
-        hash: header.hash.toString('hex'),
-        previousBlockHash: header.previousBlockHash.toString('hex'),
-        sequence: Number(header.sequence),
+        ...blockHeaderResponse,
         size: getBlockSize(block),
-        timestamp: header.timestamp.valueOf(),
-        noteSize: header.noteSize,
-        noteCommitment: header.noteCommitment.toString('hex'),
-        transactions: transactions,
+        work: block.header.work.toString(),
```
hughy, line 163: is `work` needed here? it isn't in the schema definition

ironfish/src/rpc/types.ts
```diff
@@ -54,3 +56,59 @@ export const RpcEncryptedNoteSchema: yup.ObjectSchema<RpcEncryptedNote> = yup
...
+}
```
patnir, line 77: would love your thoughts on which one of these should be deprecated. we can deprecate later also

ironfish/src/rpc/routes/chain/followChainStream.ts
```diff
@@ -58,21 +51,15 @@ export const FollowChainStreamResponseSchema: yup.ObjectSchema<FollowChainStream
...
-        transactions: yup.array(RpcTransactionSchema).defined(),
-      })
-      .defined(),
+    block: RpcBlockHeaderSchema.concat(
+      yup
+        .object({
+          main: yup.boolean().defined(),
+          size: yup.number().defined(),
+          transactions: yup.array(RpcTransactionSchema).defined(),
+        })
```
hughy, line 60: do we need to include `work` in the schema?

ironfish/src/rpc/types.ts
patnir: changed it back to null

ironfish/src/rpc/types.ts
```diff
@@ -54,3 +56,59 @@ export const RpcEncryptedNoteSchema: yup.ObjectSchema<RpcEncryptedNote> = yup
...
+    previous: header.previousBlockHash.toString('hex'),
+    sequence: Number(header.sequence),
+    previousBlockHash: header.previousBlockHash.toString('hex'),
+    timestamp: header.timestamp.valueOf(),
+    difficulty: header.target.toDifficulty().toString(),
+    graffiti: header.graffiti.toString('hex'),
+    noteCommitment: header.noteCommitment.toString('hex'),
+    transactionCommitment: header.transactionCommitment.toString('hex'),
+    target: BigIntUtils.writeBigU256BE(header.target.asBigInt()).toString('hex'),
+    randomness: BigIntUtils.writeBigU64BE(header.randomness).toString('hex'),
```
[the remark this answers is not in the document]
patnir: remove the hex conversion

ironfish/src/rpc/routes/chain/getBlock.ts
patnir: you're right, don't need it here. will remove

ironfish/src/rpc/routes/chain/followChainStream.ts
patnir: i added work to the RpcBlockHeader object directly, we don't need to include it here

ironfish/src/rpc/types.ts
```diff
@@ -54,3 +56,59 @@ export const RpcEncryptedNoteSchema: yup.ObjectSchema<RpcEncryptedNote> = yup
...
+    previous: header.previousBlockHash.toString('hex'),
+    sequence: Number(header.sequence),
+    previousBlockHash: header.previousBlockHash.toString('hex'),
+    timestamp: header.timestamp.valueOf(),
+    difficulty: header.target.toDifficulty().toString(),
+    graffiti: header.graffiti.toString('hex'),
+    noteCommitment: header.noteCommitment.toString('hex'),
+    transactionCommitment: header.transactionCommitment.toString('hex'),
+    target: BigIntUtils.writeBigU256BE(header.target.asBigInt()).toString(),
+    randomness: BigIntUtils.writeBigU64BE(header.randomness).toString(),
```
hughy, line 91: I think we don't need the `BigIntUtils` here since that will convert these to Buffers. We can use `toString` directly on the BigInts like with `work`, below

patnir: Done!

hughy (approved):

PR #4289, opened 2023-09-14
Rahul/ifl 266 mintasset and burnasset do not accept a feerate in requests

patnir:
## Summary

createTransaction and sendTransaction accept an optional feeRate. Adding that option for mintAsset and burnAsset. 

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/wallet/burnAsset.ts
```diff
@@ -12,6 +12,7 @@ export interface BurnAssetRequest {
   account: string
   assetId: string
   fee: string
+  feeRate?: string
```
hughy, line 15: I think it makes sense to make `fee` optional here -- if a client passes a fee rate they shouldn't need to also pass a fee

ironfish/src/rpc/routes/wallet/mintAsset.ts
```diff
@@ -13,6 +13,7 @@ import { getAccount } from './utils'
 export interface MintAssetRequest {
   account: string
   fee: string
+  feeRate?: string
```
hughy, line 16: same re: optional fee

ironfish/src/rpc/routes/wallet/burnAsset.ts
patnir: donie

ironfish/src/rpc/routes/wallet/mintAsset.ts
patnir: done

hughy (approved):

PR #4297, opened 2023-09-18
adding seriailized field to transaction

patnir:
## Summary

Adding serialized transaction field for the javascript consumers of our RPC

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/chain/getBlock.ts
+++ b/ironfish/src/rpc/routes/chain/getBlock.ts
@@ -109,6 +109,7 @@ routes.register<typeof GetBlockRequestSchema, GetBlockResponse>(
 
     for (const tx of block.transactions) {
       transactions.push({
+        serialized: tx.serialize().toString('hex'),
         hash: tx.hash().toString('hex'),
         size: getTransactionSize(tx),
         fee: Number(tx.fee()),
--- a/ironfish/src/rpc/routes/chain/types.ts
+++ b/ironfish/src/rpc/routes/chain/types.ts
@@ -27,7 +27,7 @@ export const RpcSpendSchema: yup.ObjectSchema<RpcSpend> = yup
   .defined()
 
 export type RpcTransaction = {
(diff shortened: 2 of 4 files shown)


ironfish/src/rpc/routes/chain/followChainStream.ts
```diff
@@ -78,9 +78,7 @@ routes.register<typeof FollowChainStreamRequestSchema, FollowChainStreamResponse
 
     const send = (block: Block, type: 'connected' | 'disconnected' | 'fork') => {
       const transactions = block.transactions.map((transaction) => ({
-        ...(request.data?.serialized
-          ? { serialized: transaction.serialize().toString('hex') }
-          : {}),
+        serialized: transaction.serialize().toString('hex'),
```
patnir, line 81: does serializing a transaction have a performance impact? is that why this is optional?

danield9tqh: There are a couple use cases for this endpoint that I know of
1. Standalone wallet + internal wallet both subscribe to a chain stream and need the full serialized transaction
2. Syncer for the block explorer also subscribes to chain stream and does not use the serialized transaction

Adding it as default would increase bandwidth usage for the block explorer syncer but it would also make the endpoints more consistent. We would not run into any bandwidth issues for the syncer if we added the `serialized` field. The only case I can think of that could run into bandwith issues is something on mobile but even a mobile wallet would need the fully serialized transactions. So returning `serialized` every time although not perfectly efficient seems like a fine thing to do here

danield9tqh:
For this PR I would add into the testing plan to make sure the syncer code still works as expected. I'm pretty sure I remember it just ignores any field it doesn't use but couldn't hurt to test

danield9tqh:
I think we discussed this with @hughy and decided to keep the `transaction: {serialized: string}` format as is here for now.

PR #4299, opened 2023-09-19
Rahul/ifl 1641 add optional serialized field to transaction rpc

patnir:
## Summary

1. serialized as optional to all transaction response objects
2. required signature field
3. serialized as part of the request object

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/wallet/utils.ts
```diff
@@ -42,6 +42,7 @@ export async function serializeRpcAccountTransaction(
   account: Account,
   transaction: TransactionValue,
   _confirmations: number | undefined,
+  _returnSerialized = false,
```
hughy, line 45: nit - the underscored fields look a little odd to me

we often use an 'options' pattern for things like this: 
```
options?: {
  confirmations: number | undefined.
  serialized: boolean
}
```

hughy (approved):

ironfish/src/rpc/routes/wallet/utils.ts
patnir: i like that, will add it

ironfish/src/rpc/routes/wallet/utils.ts
```diff
@@ -41,16 +41,23 @@ export async function serializeRpcAccountTransaction(
   node: IronfishNode,
   account: Account,
   transaction: TransactionValue,
-  _confirmations: number | undefined,
+  options?: {
+    confirmations: number | undefined
+    serialized: boolean | undefined
```
hughy, line 46: could make these optional, too: `confirmations?`, `serialized?`

hughy (approved):

PR #4300, opened 2023-09-19
Rahul/ifl 1653 update gettransaction response to use the rpc transaction

patnir:
## Summary

GetTransaction now uses RpcTransaction. Depends on #4299 

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/chain/getTransaction.ts
```diff
@@ -112,8 +94,10 @@ routes.register<typeof GetTransactionRequestSchema, GetTransactionResponse>(
     const { transaction, initialNoteIndex } = foundTransaction
 
     const rawTransaction: GetTransactionResponse = {
-      fee: transaction.fee().toString(),
+      fee: Number(transaction.fee()),
```
hughy, line 97: The fee should probably be a string representation of a BigInt because fee is a BigInt rather than a Number

We'll have to look at what the impact of changing fee to a string in `RpcTransaction` is

patnir: I can take a look, added a task here: https://linear.app/if-labs/issue/IFL-1676/investigate-changing-fee-to-string-from-number

danield9tqh: I remember we standardized as a number in this PR https://github.com/iron-fish/ironfish/pull/4217#issuecomment-1690686845. String is def more precise though. I think I said to go with Number to avoid backwards incompatible change on a more commonly used endpoint (can't remember which).

TLDR though I think either is fine for now whichever is more backwards compatible on more endpoints imo. We'd need a fee that's like ¼ of the terminal IRON supply to overflow this

hughy: Good point, a fee like that would be a dream for a miner, but not likely to ever happen

danield9tqh (approved):

PR #4303, opened 2023-09-19
removes input trimming from cli commands

hughy:
## Summary

oclif automatically splits on whitespace when parsing inputs, so using 'trim' typically isn't doing anything

in cases where a user _wants_ leading or trailing whitespace, such as `ironfish
wallet:delete 'trailing   '`, trim removes the trailing whitespace that the user
has tried to include

## Testing Plan

manual testing:
- ran wallet cli commands for account name with trailing whitespace

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/chain/readd-block.ts
+++ b/ironfish-cli/src/commands/chain/readd-block.ts
@@ -18,7 +18,6 @@ export default class ReAddBlock extends IronfishCommand {
   static args = [
     {
       name: 'hash',
-      parse: (input: string): Promise<string> => Promise.resolve(input.trim()),
       required: true,
       description: 'The hash of the block in hex format',
     },
--- a/ironfish-cli/src/commands/wallet/transaction/index.ts
+++ b/ironfish-cli/src/commands/wallet/transaction/index.ts
@@ -16,13 +16,11 @@ export class TransactionCommand extends IronfishCommand {
   static args = [
     {
       name: 'hash',
-      parse: (input: string): Promise<string> => Promise.resolve(input.trim()),
       required: true,
       description: 'Hash of the transaction',
     },
     {
       name: 'account',
-      parse: (input: string): Promise<string> => Promise.resolve(input.trim()),
       required: false,
       description: 'Name of the account',
     },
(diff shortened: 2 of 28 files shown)


ygao76 (approved):

ironfish-cli/src/commands/chain/readd-block.ts
```diff
@@ -18,7 +18,6 @@ export default class ReAddBlock extends IronfishCommand {
   static args = [
     {
       name: 'hash',
-      parse: (input: string): Promise<string> => Promise.resolve(input.trim()),
```
jowparks, line 21: Won't this break the system if the user adds extra whitespace here?

ironfish-cli/src/commands/wallet/transaction/index.ts
```diff
@@ -16,13 +16,11 @@ export class TransactionCommand extends IronfishCommand {
   static args = [
     {
       name: 'hash',
-      parse: (input: string): Promise<string> => Promise.resolve(input.trim()),
```
jowparks, line 19: Same as above, won't additional whitespace mess up hash lookup?

ironfish-cli/src/commands/chain/readd-block.ts
hughy: oclif ignores whitespace in parsing unless you escape the characters or wrap them in quotes, so `ironfish chain:readd-block dead...beef` is the same as `ironfish chain:readd-block                      dead...beef`

hughy:
splitting on or ignoring whitespace might be shell-dependent rather than oclif, but even then i'm not sure we want to prevent users from being able to use leading or trailing whitespace 🤔

dguenther:
I'd maybe suggest being a bit more conservative with where we remove the trimming to start with, in case it was fixing some issue that we forgot about. I think it's okay to leave trimming in for known values or consensus-defined values that definitely shouldn't have spaces, like hashes, asset IDs, hex-encoded values, sequences. 

However, for user-set values, if we're trimming at the CLI level but not at the SDK level, this can create bugs and should definitely be fixed. For example, I think it's possible to use the SDK to create an account with a name starting with a space, which would then be inaccessible by the CLI.

PR #4308, opened 2023-09-20
Support resuming snapshot downloads if supported

dguenther:
## Summary

As the chain snapshot grows larger, it's increasingly inconvenient for the download to error midway, which requires restarting from 0. This PR checks if the server supports range requests, and if so, passes a range header to the server if a previously existing file with the same name is in the temp directory.

## Testing Plan

* [x] Downloaded part of a snapshot, canceled the download, then resumed it and verified the chain status on completion 

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/chain/download.ts
+++ b/ironfish-cli/src/commands/chain/download.ts
@@ -205,7 +205,7 @@ export default class Download extends IronfishCommand {
         url: snapshotUrl,
         cancelToken: idleCancelSource.token,
         headers: {
-          ...(canAcceptRanges
+          ...(canAcceptRanges && downloaded > 0
             ? { range: `bytes=${downloaded}-${manifest.file_size - 1}` }
             : {}),
         },
--- a/ironfish-cli/src/commands/chain/download.ts
+++ b/ironfish-cli/src/commands/chain/download.ts
@@ -186,7 +186,7 @@ export default class Download extends IronfishCommand {
           url: snapshotUrl,
           cancelToken: idleCancelSource.token,
           headers: {
-            range: `bytes=${downloaded}-${manifest.file_size - 1}`,
+            range: `bytes=${downloaded}-`,
           },
         })
 
--- a/ironfish-cli/src/commands/chain/download.ts
+++ b/ironfish-cli/src/commands/chain/download.ts
@@ -190,10 +190,11 @@ export default class Download extends IronfishCommand {
           },
         })
 
+        const resumingDownload = response.data.statusCode === 206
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/chain/download.ts
```diff
@@ -134,6 +134,23 @@ export default class Download extends IronfishCommand {
       await fsAsync.mkdir(this.sdk.config.tempDir, { recursive: true })
       snapshotPath = flags.outputPath || path.join(this.sdk.config.tempDir, manifest.file_name)
 
+      // Check if we can resume a previous download, and update the downloaded size if so
+      const headResult = await axios.head(snapshotUrl)
+      const canAcceptRanges =
+        (headResult.headers as Record<string, string>)['accept-ranges'] === 'bytes'
```
andiflabs, line 140: Nit: you can in theory send a `Range` request even if the server does not understand it. If it supports it, it will respond with a [206 Partial Content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/206), otherwise it will respond with a 200, see the [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range):

> The server can also ignore the Range header and return the whole document with a [200](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200) status code

Therefore this HEAD request should not be necessary and can be avoided to minimize the number of roundtrips.

If you really want to test for server support, then I'd invert the order here: first check if the local snapshot exists (cheap), then ask the server (expensive).

danield9tqh (approved):

ironfish-cli/src/commands/chain/download.ts
```diff
@@ -134,116 +134,142 @@ export default class Download extends IronfishCommand {
...
+          }
+        }, idleTimeout)
+
+        const response: { data: IncomingMessage } = await axios({
+          method: 'GET',
+          responseType: 'stream',
+          url: snapshotUrl,
+          cancelToken: idleCancelSource.token,
+          headers: {
+            range: `bytes=${downloaded}-${manifest.file_size - 1}`,
```
andiflabs, line 189: Nit: you could use ``range: `bytes=${downloaded}-` `` (i.e. omit the range end) to avoid calculations

andiflabs (approved):

dguenther:
Thanks for the suggestions to help clean this up! I moved the checksum verification step to the end to avoid some code duplication, and I still need to re-test this with the changes. Once this merges, I'll make the same changes to the `node-app` code.

PR #4313, opened 2023-09-20
formatting all rpc objects in the same order

patnir:
## Summary

Formatting them in the following order: 

- fields
- objects
- optional fields 
- deprecated fields

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/types.ts
+++ b/ironfish/src/rpc/routes/wallet/types.ts
@@ -4,43 +4,42 @@
 
 import * as yup from 'yup'
 import { TransactionStatus, TransactionType } from '../../../wallet'
-import { AccountImport } from '../../../wallet/walletdb/accountValue'
 import { RpcBurn, RpcBurnSchema, RpcMint, RpcMintSchema } from '../../types'
 import { RpcSpend, RpcSpendSchema } from '../chain'
 
 export type RcpAccountAssetBalanceDelta = {
   assetId: string
+  delta: string
   /**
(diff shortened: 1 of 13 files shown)


ironfish/src/rpc/routes/wallet/types.ts
```diff
@@ -4,43 +4,42 @@
 
 import * as yup from 'yup'
 import { TransactionStatus, TransactionType } from '../../../wallet'
-import { AccountImport } from '../../../wallet/walletdb/accountValue'
 import { RpcBurn, RpcBurnSchema, RpcMint, RpcMintSchema } from '../../types'
 import { RpcSpend, RpcSpendSchema } from '../chain'
 
-export type RcpAccountAssetBalanceDelta = {
+export type RpcAccountAssetBalanceDelta = {
```
patnir, line 10: type

ironfish/src/rpc/routes/wallet/types.ts
```diff
@@ -134,7 +133,14 @@ export const RpcWalletTransactionSchema: yup.ObjectSchema<RpcWalletTransaction>
   })
   .defined()
 
-export type RpcAccountImport = Omit<AccountImport, 'createdAt'> & {
+export type RpcAccountImport = {
```
patnir, line 136: just to keep things consistent. we can do this in other places, but explicitly defining the interface here like we do everywhere else

PR #4314, opened 2023-09-21
initializes service:bridge-relay command

hughy:
## Summary

adds a service command to act as a relay; decrypting deposit transactions on Iron Fish and calling the bridge contract on Ethereum

relies on remote node to decrypt transactions with a given view key via chain/getTransactionStream

waits until block confirmed before processing deposit transactions

## Testing Plan

manual testing:
- created bridge account
- ran relay with bridge account's incoming view key
- sent deposit transaction to bridge account
image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/service/bridge-relay.ts
+++ b/ironfish-cli/src/commands/service/bridge-relay.ts
@@ -108,22 +108,14 @@ export default class BridgeRelay extends IronfishCommand {
         buffer.pop()
       }
 
-      const committing = buffer.length > confirmations
-
-      this.log(
+      this.logger.debug(
         `${content.type}: ${content.block.hash} - ${content.block.sequence}${
-          committing
-            ? ' - ' +
-              TimeUtils.renderEstimate(
-                content.block.sequence,
-                content.head.sequence,
-                speed.rate5m,
-              )
-            : ''
+          ' - ' +
+          TimeUtils.renderEstimate(content.block.sequence, content.head.sequence, speed.rate5m)
         }`,
       )
 
-      if (committing) {
+      if (buffer.length > confirmations) {
         const response = buffer.shift()
         Assert.isNotUndefined(response)
         this.commit(api, response)
--- a/ironfish-cli/src/commands/service/bridge-relay.ts
+++ b/ironfish-cli/src/commands/service/bridge-relay.ts
@@ -38,18 +38,15 @@ export default class BridgeRelay extends IronfishCommand {
       description: 'Minimum number of block confirmations needed to process deposits',
       required: false,
     }),
+    fromHead: Flags.string({
+      char: 'f',
+      description: 'The block hash to start following at',
+      required: false,
+    }),
   }
 
-  static args = [
-    {
-      name: 'head',
-      required: false,
-      description: 'The block hash to start following at',
-    },
-  ]
-
   async start(): Promise<void> {
-    const { flags, args } = await this.parse(BridgeRelay)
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/commands/service/bridge-relay.ts
```diff
@@ -0,0 +1,147 @@
...
+  ): Promise<void> {
+    this.log('Connecting to node...')
+    const client = await this.sdk.connectRpc()
+
+    this.log('Watching with view key:', viewKey)
+
+    // TODO: track chain state of relay in API
+    if (!head) {
+      const chainInfo = await client.chain.getChainInfo()
+      head = chainInfo.content.genesisBlockIdentifier.hash
```
rohanjadvani, line 89: What if we provide head optionally as a flag as well? It might be useful for testing if we don't want to start from genesis

hughy: It's there as a positional arg now, but we could change to a flag

ironfish-cli/src/commands/service/bridge-relay.ts
```diff
@@ -0,0 +1,147 @@
...
+    }),
+    token: Flags.string({
+      char: 't',
+      description: 'API token to authenticate with',
+      parse: (input: string) => Promise.resolve(input.trim()),
+      env: 'IRONFISH_API_TOKEN',
+    }),
+    viewKey: Flags.string({
+      char: 'k',
+      description: 'View key to watch transactions with',
```
jowparks, line 32: We should label this as incoming view key, since we have the concept of a "view key" which has functionality of both ivk and ovk (see --export json)

ironfish-cli/src/commands/service/bridge-relay.ts
rohanjadvani: Oh, I totally missed that. Disregard lol

ironfish-cli/src/commands/service/bridge-relay.ts
```diff
@@ -0,0 +1,147 @@
...
+      if (content.type === 'connected') {
+        buffer.push(content)
+        speed.add(1)
+      } else if (content.type === 'disconnected') {
+        buffer.pop()
+      }
+
+      const committing = buffer.length > confirmations
+
+      this.log(
```
jowparks, line 113: This is gonna be a lot of log lines, currently 220k, just fyi

hughy: yeah this was just copied from the sync command

we can make this a debug-level log or remove it

ironfish-cli/src/commands/service/bridge-relay.ts
```diff
@@ -0,0 +1,147 @@
...
+  commit(api: WebApi, response: GetTransactionStreamResponse): void {
+    Assert.isNotUndefined(response)
+
+    const transactions = response.transactions
+
+    for (const transaction of transactions) {
+      for (const note of transaction.notes) {
+        this.log(`Processing deposit ${note.memo}, from transaction ${transaction.hash}`)
+        // TODO: get Eth deposit address from API
+        // TODO: call Eth bridge contract to mint
```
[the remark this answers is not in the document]
hughy: we probably need to store some status on the deposits -- if the process stops in the middle of a block then we don't want to reprocess deposits that we've already invoked the Eth contract for

PR #4315, opened 2023-09-22
Rahul/ifl 1643 update cli to stop using the deprecated fields

patnir:
## Summary

Previous: 
image

Now: 
image

Will work on `status` separately here: https://linear.app/if-labs/issue/IFL-1690/remove-dependence-on-deprecated-status-field

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/utils/asset.ts
```diff
@@ -147,3 +160,16 @@ export async function selectAsset(
 
   return response.asset
 }
+
+export async function getAssetsByIDs(
+  client: Pick<RpcClient, 'wallet'>,
+  assetIds: string[],
+): Promise<{ [key: string]: RpcAsset }> {
+  assetIds = [...new Set(assetIds)]
```
patnir, line 168: i did this to improve the performance, so we don't lookup assets twice

hughy (approved):
👍 

deprecated asset status to be handled in separate PR

PR #4317, opened 2023-09-22
Handle socket read/write errors

andiflabs:
## Summary

After a connection is established, `RpcSocketClient` does not register any error handler. If a communication error happens (for example, the connection is dropped abruptly), then Node.js will terminate the whole process with an error like this:

```
node:events:491
      throw er; // Unhandled 'error' event
      ^

Error: read ECONNRESET
    at __node_internal_captureLargerStackTrace (node:internal/errors:491:5)
    at __node_internal_errnoException (node:internal/errors:621:12)
    at TCP.onStreamRead (node:internal/stream_base_commons:217:20)
Emitted 'error' event on Socket instance at:
    at emitErrorNT (node:internal/streams/destroy:151:8)
    at emitErrorCloseNT (node:internal/streams/destroy:116:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  errno: -104,
  code: 'ECONNRESET',
  syscall: 'read'
}

Node.js v18.13.0
```

This behavior can be sometimes observed when trying to connect to an Iron Fish Node very early, right after starting it (this was observed with the new Iron Fish Tank simulator, which spawns a node and then tries to connect to it as soon as possible).

This PR introduces error handling logic so that socket errors are propagated as exceptions, and do not result in the whole process crashing.

## Testing Plan

Before this change: the process would crash with the above output.

After this change: an error is logged, and all request handlers are rejected with an exception similar to this one:

```
read ECONNRESET
Error: read ECONNRESET
    at __node_internal_captureLargerStackTrace (node:internal/errors:491:5)
    at __node_internal_errnoException (node:internal/errors:621:12)
    at TCP.onStreamRead (node:internal/stream_base_commons:217:20) {
  errno: -104,
  code: 'ECONNRESET',
  syscall: 'read'
}
```

Note that this is a JavaScript *exception*, no longer a Node.js *unhandled error*.

## Documentation

No doc change required

## Breaking Change

Not a breaking change

### DIFF
--- a/ironfish/src/rpc/clients/socketClient.ts
+++ b/ironfish/src/rpc/clients/socketClient.ts
@@ -286,10 +286,11 @@ export abstract class RpcSocketClient extends RpcClient {
     this.isConnected = true
     this.client.on('data', this.onClientData)
     this.client.on('close', this.onClientClose)
+    this.client.on('error', this.onSocketError)
   }
 
   protected onClientData = (data: Buffer): void =>
-    void this.onData(data).catch((e) => this.onError(e))
+    void this.onData(data).catch((e) => this.onError(e)) // recoverable
 
   protected onData = async (data: Buffer): Promise<void> => {
     this.messageBuffer.write(data)
@@ -314,7 +315,7 @@ export abstract class RpcSocketClient extends RpcClient {
         }
         case 'error':
         case 'malformedRequest': {
-          this.onError(data)
+          this.onError(data) // recoverable
           break
         }
       }
@@ -328,6 +329,7 @@ export abstract class RpcSocketClient extends RpcClient {
     if (this.client) {
       this.client.off('data', this.onClientData)
       this.client.off('close', this.onClientClose)
+      this.client.off('error', this.onError)
       this.client = null
     }
 
@@ -347,8 +349,13 @@ export abstract class RpcSocketClient extends RpcClient {
     this.handleStream(data).catch((e) => this.onError(e))
   }
 
-  protected onError(error: unknown): void {
+  protected onError = (error: unknown): void => {
     this.logger.error(ErrorUtils.renderError(error))
+  }
+
+  protected onSocketError = (error: unknown): void => {
+    this.logger.error(ErrorUtils.renderError(error))
+    this.onClientClose()
   }
 
   describe(): string {


ironfish/src/rpc/clients/socketClient.ts
```diff
@@ -349,6 +350,9 @@ export abstract class RpcSocketClient extends RpcClient {
 
   protected onError(error: unknown): void {
     this.logger.error(ErrorUtils.renderError(error))
+    for (const request of this.pending.values()) {
```
NullSoldier, line 353: Can you explain what this is meant to try to fix?

This is already handled in `onClientClose` in the case that the connection is closed unexpectedly. Otherwise, onError can be called if one request fails, but the other requests shouldn't be rejected in that case of one error. I think the way this is intended, an error doesn't necessarily mean the end of valid responses for all pending requests.

ironfish/src/rpc/clients/socketClient.ts
```diff
@@ -286,6 +286,7 @@ export abstract class RpcSocketClient extends RpcClient {
     this.isConnected = true
     this.client.on('data', this.onClientData)
     this.client.on('close', this.onClientClose)
+    this.client.on('error', this.onError.bind(this))
```
dguenther, line 289: Do you want to `.off` this in `onClientClose` too? I don't think it's actually necessary, but we do it for the `data` and `close` events.

ironfish/src/rpc/clients/socketClient.ts
```diff
@@ -349,6 +350,9 @@ export abstract class RpcSocketClient extends RpcClient {
 
   protected onError(error: unknown): void {
     this.logger.error(ErrorUtils.renderError(error))
+    for (const request of this.pending.values()) {
+      request.reject(error)
```
danield9tqh, line 354: Curous about Jason's question as well. But.. if we do intend to reject all requests would we also need to close connection to the server?

ironfish/src/rpc/clients/socketClient.ts
andiflabs: In the Unix world, there are only two errors that I know that may be retried with recv()/read()/send()/write() and friends, when working with sockets: 
- EAGAIN, when using nonblocking sockets (which is not the case here, I think),
- EINTR, when not using SA_RESTART when setting up signal handlers (which again I don't think can occur here).

All other errors, to my knowledge, are errors that leave the socket or the HTTP request in an invalid state that cannot be recovered. Some common scenarios of when this may occur are: the connection is closed abruptly (without receiving an RST packet from the source), routing tables change somewhere in the path and the peer becomes unreachable, the maximum number of TCP retries is hit, and so on. In all those cases, a "connection closed" event may not be fired (because technically the connection did not get closed on the TCP/IP layer), and in all those cases it won't be possible to reliably know what should be done next with the connection.

For example: suppose that the client sends the HTTP headers for a request on a TCP packet. Routing changes, the destination becomes unreachable, and an error is thrown. At that point, even if the routing is corrected, it won't be possible to recover the situation: it should in theory be possible to resend the HTTP headers in a new TCP packet, but from the Node.js exception we don't have enough information to know what TCP packet needs to be re-transmitted, nor we have the functionality from the HTTP library to do the re-transmission. Moreover, it is actually quite unlikely that such errors will recover themselves in a short period of time, and so any re-transmission logic may actually end up hanging there forever.

In case of error, it is way more reliable to just discard the connection, establish a new one, and retry all in-flight HTTP requests. This PR doesn't really do any of that: it just propagates the error to the upper layer in the hope that those layers have the correct logic to handle this situation.

So, in short, errors are (generally speaking) not retry-able, and errors may not result in a "connection closed" event.

andiflabs: Note that, in the scenario that motivated this PR, we have an ECONNRESET ("connection reset by peer") returned during a send() call, so here technically a connection was closed, but for some reason Node.js does not want treat this situation as a "connection closed", but rather wants to treat is as an error condition.

ironfish/src/rpc/clients/socketClient.ts
andiflabs: If we get a transmission error, trying to close the connection would result in another error, because transmitting the RST packet would fail too (which is fine, as long as it's properly handled). However, we do need to close the file descriptor--I think it's done automatically, am I wrong?

ironfish/src/rpc/clients/socketClient.ts
andiflabs: Just to add another example of why continuing when an error is present (if possible) may result in unintended consequences:

1. http headers for request A are sent
2. http body is sent, but an error occurs so it's never received
3. http headers for request B are sent

In this situation, the server will think that the headers for request B are the body of request A

andiflabs:
Had a discussion with Daniel about this PR. Two important points:
- The error handler should call `onClientClose` instead of just propagating errors
- `onError` is currently called in other places where the error is recoverable, so handling socket errors requires a special handler (called `onSocketError` in the next PR revision)

danield9tqh (approved):

PR #4320, opened 2023-09-25
Rahul/ifl 1690 remove dependence on deprecated status field

patnir:
## Summary

Using transaction status as single source of truth for status of an asset

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/assets.ts
+++ b/ironfish-cli/src/commands/wallet/assets.ts
@@ -59,16 +59,13 @@ export class AssetsCommand extends IronfishCommand {
 
     for await (const asset of response.contentStream()) {
       let status: string
-
       if (Asset.nativeId().toString('hex') === asset.id) {
         status = AssetStatus.CONFIRMED
       } else {
         const transaction = await client.wallet.getAccountTransaction({
           hash: asset.createdTransactionHash,
         })
-
         Assert.isNotNull(transaction.content.transaction)
-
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/commands/wallet/assets.ts
```diff
@@ -58,6 +58,17 @@ export class AssetsCommand extends IronfishCommand {
...
 
     for await (const asset of response.contentStream()) {
+      let status: string
+      if (Asset.nativeId().toString('hex') === asset.id) {
+        status = AssetStatus.CONFIRMED
+      } else {
+        const transaction = await client.wallet.getAccountTransaction({
+          hash: asset.createdTransactionHash,
+        })
+        Assert.isNotNull(transaction.content.transaction)
```
hughy, line 68: The wallet won't necessarily have this transaction -- the wallet will store a record for any assets that it has tokens for, but it won't have the transaction that created the asset unless the account also created the asset

For example, if I mint a new asset in transaction _A_ and then send you some in transaction _B_, then your account won't have a record of _A_ because it wasn't involved in the transaction

We may need to get the transaction from the chain here instead

hughy: Another option here would be to remove the status column from this command's output and give `createdTransactionHash` instead

hughy (approved):

PR #4324, opened 2023-09-29
adding string length limit to node name

patnir:
## Summary

Limiting size of node name to 32 letters. Other peers cannot connect to you if you have a long node name. 

image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/fileStores/config.ts
+++ b/ironfish/src/fileStores/config.ts
@@ -332,7 +332,7 @@ export const ConfigOptionsSchema: yup.ObjectSchema<Partial<ConfigOptions>> = yup
     // to parse logPrefix
     logPrefix: yup.string(),
     blockGraffiti: yup.string(),
-    nodeName: yup.string(),
+    nodeName: yup.string().max(32),
     nodeWorkers: yup.number().integer().min(-1),
     nodeWorkersMax: yup.number().integer().min(-1),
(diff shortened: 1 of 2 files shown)


jowparks:
Can you give context as to why other peers can't connect with > 32 char? Wondering what the underlying issue is.

dguenther (changes requested):
This looks good! One other thing I noticed, I think there's a bug in the PeerManager -- it's supposed to take the first 32 characters when the node name is too long, but it looks like it actually takes characters starting at index 32 (which could still result in more than 32 characters if the string is like 100 chars long). Would you mind adding a fix for that in this PR too?

https://github.com/iron-fish/ironfish/blob/c898a1ee23db85fdfaa549fe21f795832b400591/ironfish/src/network/peerNetwork.ts#L229

dguenther:
> Can you give context as to why other peers can't connect with > 32 char? Wondering what the underlying issue is.

We have a rule in peerManager that drops connections to peers that identify themselves with a node name greater than 32 chars: https://github.com/iron-fish/ironfish/blob/c898a1ee23db85fdfaa549fe21f795832b400591/ironfish/src/network/peers/peerManager.ts#L1094C8-L1094C8

dguenther (approved):

PR #4325, opened 2023-09-29
Adding hex flag to handle invalid hex field

patnir:
## Summary

Fixes: IFL-1714

https://linear.app/if-labs/issue/IFL-1714/cli-crashes-with-assetid-flag

Result (on Testnet)

Invalid hex value for asset id: 
image

Valid hex value: 
image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/flags.ts
+++ b/ironfish-cli/src/flags.ts
@@ -156,3 +156,14 @@ export const parseIron = (input: string, opts: IronOpts): Promise<bigint> => {
     }
   })
 }
+
+export const HexFlag = Flags.custom<string>({
+  parse: async (input, _ctx, opts) => {
+    const hexRegex = /^[0-9A-Fa-f]+$/g
+    if (!hexRegex.test(input)) {
+      throw new Error(`The hex string is invalid for field ${opts.name}`)
+    }
+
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/flags.ts
```diff
@@ -156,3 +156,14 @@ export const parseIron = (input: string, opts: IronOpts): Promise<bigint> => {
     }
   })
 }
+
+export const HexFlag = Flags.custom<string>({
+  parse: async (input, _ctx, opts) => {
+    const hexRegex = /^[0-9A-Fa-f]+$/g
+    if (!hexRegex.test(input)) {
+      throw new Error(`The hex string is invalid for field ${opts.name}`)
```
mat-if, line 164: Thinking about this error message a bit. I wonder if something like "The value provided for ${opts.name} is an invalid format. The value must only contain numbers and letters A-F" or something. I wouldn't want the expectation that a user should know or care what hex is.

mat-if (approved):

ironfish-cli/src/flags.ts
patnir: done

PR #4332, opened 2023-10-04
adds 'release' service command

hughy:
## Summary

adapted from 'faucet' service

polls bridge api for WIRON->IRON requests to process

creates a transaction with an output for each request returned, as available balance allows

updates bridge api with destination_transaction and state change after transaction creation

## Testing Plan

- manual testing: connected to local bridge api, sent transaction

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/service/bridge/release.ts
+++ b/ironfish-cli/src/commands/service/bridge/release.ts
@@ -1,0 +1,221 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Asset, isValidPublicAddress } from '@ironfish/rust-nodejs'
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/commands/service/bridge/release.ts
```diff
@@ -0,0 +1,221 @@
...
+        '   ',
+      )} ${tx.content.hash} (5m avg ${speed.rate5m.toFixed(2)})`,
+    )
+
+    const updatePayload = []
+    for (const request of requestsToProcess) {
+      updatePayload.push({
+        id: request.id,
+        destination_transaction: tx.content.hash,
+        status: 'PENDING_ON_DESTINATION_CHAIN',
```
rohanjadvani, line 215: What do you think about `PENDING_IRON_RELEASE_TRANSACTION_CONFIRMATION`? I think `PENDING_ON_DESTINATION_CHAIN` makes sense but is less explicit. I might rename or split that out

hughy: That sounds good to me!

rohanjadvani (approved):

PR #4345, opened 2023-10-09
Move snapshot download code into shared functions

danield9tqh:
## Summary
Moving some snapshot code into shared functions to be re-used by node app

## Testing Plan
Tested snapshot download locally

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


dgca (approved):

ironfish-cli/src/commands/chain/download.ts
```diff
@@ -44,11 +27,11 @@ export default class Download extends IronfishCommand {
       required: false,
       description: 'Path to a downloaded snapshot file to import',
     }),
-    outputPath: Flags.string({
+    output: Flags.string({
```
dguenther, line 30: Even though this likely isn't used often, we should add a note to the release notes that this is a breaking change

dguenther (approved):
Downloaded a snapshot and tested that the node booted up with it.

PR #4347, opened 2023-10-09
adds prompt for eth dest address to deposit

hughy:
## Summary

prompts the user to enter an address if one was not provided instead of requiring the flag be set

## Testing Plan

manual testing

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/service/bridge/deposit.ts
+++ b/ironfish-cli/src/commands/service/bridge/deposit.ts
@@ -97,7 +97,7 @@ export class Deposit extends IronfishCommand {
 
     let dest = flags.dest
     if (dest == null) {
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/service/bridge/deposit.ts
```diff
@@ -96,6 +95,11 @@ export class Deposit extends IronfishCommand {
 
     const to = flags.to ?? (await api.getBridgeAddress())
 
+    let dest = flags.dest
```
jowparks, line 98: Looks good if default is null, not sure if flags defaults to null or undefined but I am guessing you tested :)

jowparks (approved):

rohanjadvani (approved):

PR #4350, opened 2023-10-10
Use connected peers when deciding peer limits

mat-if:
## Summary

This uses connected nodes instead of non-disconnected nodes to determine if a node should continue making outbound connections and accepting inbound connections. The main change here is that we are making the assumption that most node connections will fail, so this allows a node to have more active connection attempts. The end goal should be faster peer saturation. This does introduce the possibility of a node to have more than maxPeers connected peers. The current assumption is that natural attrition will take care of this, however, it would be a low amount of effort to add additional logic to prune peers if above maxPeers if it ends up being an issue.

## Testing Plan

Manual testing: run `ironfish peers --all`

## Documentation

N/A

## Breaking Change

N/A

### DIFF
--- a/ironfish/src/network/peers/peerManager.ts
+++ b/ironfish/src/network/peers/peerManager.ts
@@ -575,10 +575,6 @@ export class PeerManager {
     peer.close()
   }
 
-  getPeersWithConnection(): ReadonlyArray<Peer> {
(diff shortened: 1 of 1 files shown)


patnir (approved):

patnir:
@mat-if if we do make this change we should remove `getPeersWithConnection`

PR #4351, opened 2023-10-10
sends eth address in deposit memo

hughy:
## Summary

eth addresses are only 20 bytes in length, so we can fit the eth address into an Iron Fish note memo instead of first creating a record with a uuid through the api

adds eth address validation to avoid sending transactions with invalid addresses by mistake

strips '0x' from start of eth addresses (addresses are valid with or without the prefix)

uses binary string encoding for eth addresses in memos

removes createDeposit from WebApi

Resolves IFL-1753

## Testing Plan

manual testing

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/service/bridge/deposit.ts
+++ b/ironfish-cli/src/commands/service/bridge/deposit.ts
@@ -4,6 +4,7 @@
 import { Asset } from '@ironfish/rust-nodejs'
 import { CurrencyUtils, SendTransactionRequest, WebApi } from '@ironfish/sdk'
 import { CliUx, Flags } from '@oclif/core'
+import { isAddress } from 'web3-validator'
 import { IronfishCommand } from '../../../command'
 import { IronFlag, RemoteFlags } from '../../../flags'
 import { promptCurrency } from '../../../utils/currency'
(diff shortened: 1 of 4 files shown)


ironfish-cli/src/commands/service/bridge/deposit.ts
```diff
@@ -60,7 +61,20 @@ export class Deposit extends IronfishCommand {
       required: false,
     }),
     dest: Flags.string({
-      description: 'Eth public address to deposit to',
+      description: 'ETH public address to deposit to',
+      parse: (input: string): Promise<string> => {
+        return new Promise((resolve, reject) => {
+          if (!isAddress(input)) {
```
rohanjadvani, line 67: Should this be
```suggestion
          if (isAddress(input)) {
```

ironfish-cli/src/commands/service/bridge/deposit.ts
```diff
@@ -175,4 +197,9 @@ export class Deposit extends IronfishCommand {
 
     this.log(`Deposit transaction hash: ${sendResponse.content.hash}`)
   }
+
+  encodeEthAddress(address: string): string {
+    const buffer = Buffer.from(address, 'hex')
+    return buffer.toString('binary')
```
hughy, line 203: This doesn't seem to work. This memo is converted to a Buffer with utf8 encoding which results in a buffer of 34 bytes. The last two bytes get truncated from the memo that ends up on the note

rohanjadvani (approved):

PR #4357, opened 2023-10-12
Fix status follow erroring out

danield9tqh:
## Summary
With previous changes in https://github.com/iron-fish/ironfish/pull/4312 the status stream breaks if a node disconnects in the middle of a status follow. Fixing that problem

## Testing Plan
Tested locally

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


danield9tqh:
Expected behavior for this PR is that it hangs and says "Node Stopped". Are you seeing something different? I'm trying to get it back to what it was originally.

mat-if (approved):

PR #4358, opened 2023-10-12
Fix custom asset fetching when passing a non-default account to wallet:transactions

dguenther:
## Summary

In `selectAsset` and `ironfish wallet:transactions`, the account flag and confirmation flag weren't being passed to the getAsset RPC call. This resulted in an error when running `ironfish wallet:transactions <nonDefaultAccount>` when the non-default account had custom assets that the default account did not:

```
No asset found with identifier 2822ed1455752040fd780d43d64f56ea624be6f17c5c96e20ae06ff3f2c7ec39
```

Fixed this by passing the flags through to `getAsset`.

## Testing Plan

* [x] Ran `ironfish wallet:transactions` on an account with hashing algo vote tokens to verify the bug, then again to verify the fix

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/transactions.ts
+++ b/ironfish-cli/src/commands/wallet/transactions.ts
@@ -91,12 +91,16 @@ export class TransactionsCommand extends IronfishCommand {
         const assetLookup = await getAssetsByIDs(
           client,
           transaction.notes.map((n) => n.assetId) || [],
+          account,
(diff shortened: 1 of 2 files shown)


hughy (approved):
Good catch!

andiflabs (approved):

PR #4365, opened 2023-10-16
updates relay service to hit /bridge/burn endpoint

hughy:
## Summary

relay services hits the '/bridge/burn' api endpoint for any transaction it receives for non-native assets

this will rely on the api to determine if the asset sent to the bridge was a valid asset that can be released on the Ethereum side

adds 'bridgeBurn' WebApi method

## Testing Plan

manual testing sending custom asset to bridge

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/service/bridge/relay.ts
+++ b/ironfish-cli/src/commands/service/bridge/relay.ts
@@ -6,6 +6,7 @@ import { Flags } from '@oclif/core'
 import { isAddress } from 'web3-validator'
 import { IronfishCommand } from '../../../command'
 import { RemoteFlags } from '../../../flags'
+import { Asset } from '@ironfish/rust-nodejs'
 
 export default class BridgeRelay extends IronfishCommand {
   static hidden = true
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/commands/service/bridge/relay.ts
```diff
@@ -213,6 +221,10 @@ export default class BridgeRelay extends IronfishCommand {
       await api.sendBridgeDeposits(sends)
     }
```
jowparks, line 223: We are not account for sends with multiple asset types, perhaps we just say this is currently not supported. It would be possible given the standard transaction construction, but would result in a weird state.

jowparks (approved):

ironfish-cli/src/commands/service/bridge/relay.ts
hughy: I think not supported for now -- we'd need a new wrapped asset on the Ethereum side for each custom asset that we want to support bridging.

PR #4367, opened 2023-10-17
Rahul/ifl 1757 address manager improvements

patnir:
## Summary

1. We save peers in the hosts.json file when their connection gets updated to CONNECTED and is a websocket + outbound connection
2. We don't overwrite the peers on disk any more, we append to the list instead of truncating it
3. We connect to those peers on startup first (and that's it)
4. Added the field `lastAddedTimestamp` that is used to sort the peers we want to keep in the address manager (by most recent)
5. We are no longer saving candidates on a timer. We just write them down as we find good candidates. We also write it down once when the `stop` command is called. On testing, when we connect to 50 peers (which took my machine 4 hours at home), we usually connect to around 10 candidates, the number of writes to disk is low

## Metric Improvements

Test: 
1. Get a fully synced node
2. Remove the hosts.json file 
3. Measure time to connect to 20 peers
4. Stop node 
6. Measure time again to connect to 20 peers
[I repeated this test manually 4 times]

Setup: Macbook Pro M2 with a 900 mb/s internet connection at my home

Number of peers in the `hosts.json` file after connecting to 20 peers: 
1. current change: ~`5`
2. old address manager: ~`20`

Time to get to 20 peers without hosts.json: around `16 minutes`

Time to get to 20 peers w/o bootstrap nodes in the hosts.json file and using the `--bootstrap = ''` flag: 
1. current change: `Low end: 3 minutes. High end: 12 minutes`
2. old address manager: `Low end: 14 minutes. High end: +30 minutes`. Note: there were some cases it would connect to a bad peer off the bat and just stay stuck at 1 peer. The likelihood of this happening with the current change is drastically reduced. 

This is not a perfect test because luck is a significant part of the equation at the moment. You might just happen to connect with a really good or bad peer off the bat and that might derail the test. However, the current change does consistently connect to good peers and also reduces the reliance on the bootstrap nodes. This test might have also been impacted by my router's settings/ internet provider.

## Testing Plan
Updating unit tests

Manual: 
1. Run the node once with this code and accumulate about 4 - 8 peers in the `hosts.json` file
7. Run the node again but with `--bootstrap ''` flag and you will notice a speed up on connecting to 3-8 peers off the bat. 
I also did the above in a wework when I had a solid set of peers. It significantly sped up connecting to peers.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/network/peerNetwork.ts
```diff
@@ -337,6 +337,22 @@ export class PeerNetwork {
...
+        `Connecting to peer with identity: ${peerAddress.identity ?? 'none'} and name: ${
+          peerAddress.name ?? 'none'
+        } at address: ${peerAddress.address}:${peerAddress.port}`,
+      )
+      this.peerManager.connectToWebSocketAddress({
+        host: peerAddress.address,
+        port: peerAddress.port,
+        whitelist: true,
+      })
+    }
```
patnir, line 354: could move this into a function in the address manager, but followed what we're doing with the bootstrap nodes below.

ironfish/src/network/peers/addressManager.ts
```diff
@@ -14,14 +14,28 @@ import { PeerManager } from './peerManager'
 export class AddressManager {
   hostsStore: HostsStore
   peerManager: PeerManager
+  peerIdentityMap: Map<string, PeerAddress>
```
patnir, line 17: core change is a new data structure where we keep track of the peer candidates and attributes associated with them

ironfish/src/network/peers/addressManager.ts
```diff
@@ -36,59 +50,73 @@ export class AddressManager {
...
-        return {
-          address: peer.address,
-          port: peer.port,
-          identity: peer.state.identity ?? null,
-          name: peer.name ?? null,
-        }
-      } else {
-        return []
-      }
+  addPeer(peer: Peer): void {
```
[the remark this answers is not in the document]
patnir: done

ironfish/src/network/peers/peerConnectionManager.ts
```diff
@@ -104,15 +104,6 @@ export class PeerConnectionManager {
...
 
-    if (connectAttempts < CONNECT_ATTEMPTS_MAX && this.peerManager.canCreateNewConnections()) {
-      const peer = this.peerManager.createRandomDisconnectedPeer()
-      if (peer && this.connectToEligiblePeers(peer)) {
-        connectAttempts++
-      } else if (peer) {
-        this.peerManager.tryDisposePeer(peer)
-      }
-    }
-
```
patnir, line 115: this is replaced with the logic in the peerNetworks. Address manager is now a store of quality candidates rather than any candidate. These candidates are connected to at node startup

ironfish/src/network/peerNetwork.ts
```diff
@@ -358,6 +358,14 @@ export class PeerNetwork {
         whitelist: true,
       })
     }
+
+    // Connect to prior websocket outbound connections that were saved to disk
+    for (const peerAddress of this.peerManager.addressManager.priorConnectedPeerAddresses) {
+      this.peerManager.connectToWebSocketAddress({
+        host: peerAddress.address,
+        port: peerAddress.port,
+      })
```
patnir, line 367: This function also adds the peers to the candidates list

patnir:
I'm also open to changing the name of the address manager to `PeerAddresStore` or something equivalent to reflect the real functionality of it

ironfish/src/network/peers/addressManager.ts
```diff
@@ -12,71 +13,113 @@ import { PeerManager } from './peerManager'
...
-    const currentPeerIdentities = new Set(peerIdentities)
+    for (const peer of priorPeers) {
+      if (peer.identity === null || peer.address === null || peer.port === null) {
+        continue
+      }
+
+      // Backwards compatible change: if lastAddedTimestamp is undefined or null,
+      // set it to the current time.
+      if (peer.lastAddedTimestamp === undefined) {
+        peer.lastAddedTimestamp = currentTime
```
danield9tqh, line 45: In `addPeer` we're using this `lastAddedTimestamp` to indicate when the last time these peers were connected to via websocket. These peers though we have no record of them connecting via websocket. Do we want to set this value here?

danield9tqh: Alternatively I think we could just rely on the calls to `peerManager.connectToWebSocketAddress` on startup to add these legacy peers into the address manager if we are able to connect to them via websocket

ironfish/src/network/peers/addressManager.ts
```diff
@@ -12,71 +13,113 @@ import { PeerManager } from './peerManager'
...
-      return null
+    // If there are more than 50 peers, we remove
+    // extra peers from the list. This should only happen during
+    // the first time the node is started after this change is implemented.
+    if (priorPeers.length > this.LIMIT) {
+      priorPeers = priorPeers.slice(0, this.LIMIT)
     }
 
-    const currentPeerIdentities = new Set(peerIdentities)
+    for (const peer of priorPeers) {
```
danield9tqh, line 37: should we filter for peers that don't have identity, address or port before we slice with the limit. That way we don't slice out all of the good options

patnir: moved the filter to after this loop

ironfish/src/network/peers/addressManager.ts
patnir: I think we can set the timestamp to 0. 

Then when the first loop executes, the peerManager.connectToWebSocketAddress will trigger the events to update the timestamp. I just tested this locally to confirm

patnir: My take is that they will get filtered out eventually as we create more connections with peers with using websocket outbound.

ironfish/src/network/peers/addressManager.ts
```diff
@@ -12,71 +13,114 @@ import { PeerManager } from './peerManager'
...
-      (address) => address.identity !== null && !currentPeerIdentities.has(address.identity),
-    )
-    if (disconnectedPriorAddresses.length) {
-      return ArrayUtils.sampleOrThrow(disconnectedPriorAddresses)
+    for (const peer of priorPeers) {
+      this.peerIdentityMap.set(peer.identity, peer)
     }
 
-    return null
+    void this.save()
```
mat-if, line 51: Maybe this is worth discussing more, but my first reaction is that these (this one and the others in this file) should not be void, but instead `return this.save()`. Places that explicitly don't want to wait for the promise can use `void` at their call-site, but in most cases we probably _do_ want to wait for the promise.

ironfish/src/network/peers/addressManager.ts
```diff
@@ -12,71 +13,114 @@ import { PeerManager } from './peerManager'
  * and provides functionality for persistence of said data.
  */
 export class AddressManager {
+  LIMIT = 50
```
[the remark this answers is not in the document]
patnir: Good idea, I'll implement those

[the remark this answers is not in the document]
patnir: What do you think about making it a static member of the class? That way to access it you would do AddressManager.MAX_PEER_ADDRESSES instead of MAX_PEER_ADDRESSES? When reading the code it would make it easier to know where the constant comes from?

ironfish/src/network/peers/addressManager.test.ts
```diff
@@ -18,94 +18,178 @@ import { PeerManager } from './peerManager'
...
+    const now = Date.now()
+    Date.now = jest.fn(() => now)
+    const hostsStore = mockHostsStore()
+
+    const pm = new PeerManager(mockLocalPeer(), hostsStore)
+    const addressManager = new AddressManager(hostsStore, pm)
+    addressManager.hostsStore = hostsStore
+    const { peer: connectedPeer } = getConnectedPeer(pm)
+    getConnectingPeer(pm)
+    getDisconnectedPeer(pm)
```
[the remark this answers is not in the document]
patnir: removed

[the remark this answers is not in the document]
patnir: I think they were left over from the previous version and i forgot to remove them

ironfish/src/network/peers/addressManager.ts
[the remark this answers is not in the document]
mat-if: Just to follow up here even though we've already discussed a bit - that seems fine to me. I think the const probably wont be referenced outside of the AddressManager and the tests for the AddressManager, so the extra clarity may be a moot point. Regardless, I don't have any issues with that approach if you want to do that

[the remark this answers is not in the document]
patnir: I made it a simple constant to keep it similar to the other patterns in our codebase

ironfish/src/network/peerNetwork.ts
```diff
@@ -358,6 +358,14 @@ export class PeerNetwork {
         whitelist: true,
       })
     }
+
+    // Connect to prior websocket outbound connections that were saved to disk
+    for (const peerAddress of this.peerManager.addressManager.priorConnectedPeerAddresses) {
```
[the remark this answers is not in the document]
patnir: Done

ironfish/src/network/peers/addressManager.ts
```diff
@@ -12,71 +13,114 @@ import { PeerManager } from './peerManager'
...
 
-  /**
-   * Returns a peer address for a disconnected peer by using current peers to
-   * filter out peer addresses. It attempts to find a previously-connected
-   * peer address that is not part of an active connection.
-   */
-  getRandomDisconnectedPeerAddress(peerIdentities: string[]): PeerAddress | null {
-    if (this.priorConnectedPeerAddresses.length === 0) {
-      return null
+    // If there are more than 50 peers, we remove
```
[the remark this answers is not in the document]
patnir: Removed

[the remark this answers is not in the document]
patnir: I like that heuristic, will pay attention to that going forward

ironfish/src/network/peers/addressManager.ts
```diff
@@ -12,71 +13,114 @@ import { PeerManager } from './peerManager'
...
-      return ArrayUtils.sampleOrThrow(disconnectedPriorAddresses)
+    for (const peer of priorPeers) {
+      this.peerIdentityMap.set(peer.identity, peer)
     }
 
-    return null
+    void this.save()
+  }
+
+  get priorConnectedPeerAddresses(): ReadonlyArray<Readonly<PeerAddress>> {
```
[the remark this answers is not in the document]
patnir: since it's only 50 items, I feel like it's not a big deal. I'll change it if feel strongly about it

ironfish/src/network/peers/addressManager.test.ts
```diff
@@ -18,94 +18,178 @@ import { PeerManager } from './peerManager'
...
+    const hostsStore = mockHostsStore()
+    const localPeer = mockLocalPeer()
+    const pm = new PeerManager(localPeer, hostsStore)
+    const addressManager = new AddressManager(hostsStore, pm)
+    expect(addressManager.priorConnectedPeerAddresses.length).toEqual(0)
+    const { peer: peer1 } = getConnectedPeer(pm)
+    addressManager.addPeer(peer1)
+    expect(addressManager.priorConnectedPeerAddresses.length).toEqual(1)
+    addressManager.removePeer(peer1)
+    expect(addressManager.priorConnectedPeerAddresses.length).toEqual(0)
```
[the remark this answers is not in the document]
patnir: done and done

ironfish/src/network/peers/addressManager.ts
```diff
@@ -12,71 +13,114 @@ import { PeerManager } from './peerManager'
...
+    this.peerIdentityMap = new Map<string, PeerAddress>()
 
-  get priorConnectedPeerAddresses(): ReadonlyArray<Readonly<PeerAddress>> {
-    return this.hostsStore.getArray('priorPeers')
-  }
+    let priorPeers = this.hostsStore.getArray('priorPeers').filter((peer) => {
+      if (peer.identity === null || peer.address === null || peer.port === null) {
+        return false
+      }
+      // Backwards compatible change: if lastAddedTimestamp is undefined or null,
```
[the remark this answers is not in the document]
patnir: Changed it to use nullish-coalescing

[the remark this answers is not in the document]
patnir: And removed the comment

ironfish/src/network/peers/addressManager.test.ts
```diff
@@ -18,94 +18,178 @@ import { PeerManager } from './peerManager'
...
-  it('removePeerAddress should remove a peer address', () => {
-    const pm = new PeerManager(mockLocalPeer(), mockHostsStore())
-    const addressManager = new AddressManager(mockHostsStore(), pm)
-    addressManager.hostsStore = mockHostsStore()
-    const { peer: peer1 } = getConnectedPeer(pm)
-    const allPeers: Peer[] = [peer1]
-    const allPeerAddresses: PeerAddress[] = []
+  it('only connected peers get added to the address manager', () => {
+    const now = Date.now()
+    Date.now = jest.fn(() => now)
```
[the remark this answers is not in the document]
patnir: very cool, didn't know this existed. I'll change it everywhere

ironfish/src/network/peers/addressManager.ts
patnir: I'm implementing this on my local and trying to think about the changes, will update what I find

patnir: I do feel like it is more explicit that these functions have been voided. So I made the change.

danield9tqh: This is in the constructor though so we can't have async code. I think maybe makes sense just to not call `save()` at all in the constructor and wait for it to be saved on the original schedule

patnir: Although, I think TS doesn't allow constructors to be async. I feel like it's not super critical to add an await here because the addPeer will be called pretty quickly after this instantiates. So, I think it's fine to leave it as void here. 

The change would be to create an `init` function and call that where the address manager is instantiated. I'm OK doing that and it fits the pattern of our other classes

patnir: @danield9tqh you're right, i'll remove it from the constructor. It's going to be called right after with addPeer

danield9tqh (approved):

mat-if (approved):

PR #4371, opened 2023-10-18
Add better peer selection logic to the syncer

mat-if:
## Summary
The Syncer's `findPeer` function will now choose 8 random peers. Of these peers, it will measure the round-trip-time of a blockHeadersRequest of the genesis block. It then chooses the fastest of these peers to sync from.

Additionally, it will check again on a growing interval (2, 4, 8... up to 60) minutes. The purpose of this is that it should check frequently on initial node startup, as the number of connected peers grows so that it has a good chance of selecting a good peer to sync from early.

It includes the current syncing peer in the checks on the interval, so that if you find a good quality peer, you are likely to continue syncing from that peer unless you find a better one.

Closes IFL-1781

## Testing Plan
- Manual testing
- Unit tests where applicable

## Documentation

N/A

## Breaking Change

N/A

### DIFF
--- a/ironfish/src/network/peers/peer.ts
+++ b/ironfish/src/network/peers/peer.ts
@@ -82,6 +82,10 @@ export class Peer {
     return this._state
   }
 
+  get identity(): Identity | null {
+    return this.state.identity
+  }
+
   get isSaturated(): boolean {
     return this.pendingRPC >= this.pendingRPCMax
   }
@@ -409,10 +413,6 @@ export class Peer {
     if (this.state.identity === null) {
       throw new Error('Called getIdentityOrThrow on an unidentified peer')
     }
-    return this.state.identity
-  }
-
-  getIdentity(): Identity | null {
     return this.state.identity
   }
 
--- a/ironfish/src/network/peers/peer.ts
+++ b/ironfish/src/network/peers/peer.ts
@@ -82,10 +82,6 @@ export class Peer {
     return this._state
   }
 
-  get identity(): Identity | null {
-    return this.state.identity
-  }
-
   get isSaturated(): boolean {
     return this.pendingRPC >= this.pendingRPCMax
   }
--- a/ironfish/src/network/peers/peer.ts
+++ b/ironfish/src/network/peers/peer.ts
@@ -412,6 +412,10 @@ export class Peer {
     return this.state.identity
   }
 
+  getIdentity(): Identity | null {
+    return this.state.identity
+  }
+
   /**
    * Get the peers connectable websocket address
    */
--- a/ironfish/src/syncer.ts
+++ b/ironfish/src/syncer.ts
@@ -13,7 +13,14 @@ import { BAN_SCORE, PeerState } from './network/peers/peer'
 import { Block, GENESIS_BLOCK_SEQUENCE } from './primitives/block'
 import { BlockHeader } from './primitives/blockheader'
 import { Telemetry } from './telemetry'
-import { BenchUtils, ErrorUtils, HashUtils, MathUtils, SetTimeoutToken } from './utils'
+import {
+  BenchUtils,
+  ErrorUtils,
+  HashUtils,
+  MathUtils,
+  SetTimeoutToken,
+  TimeUtils,
+} from './utils'
 import { ArrayUtils } from './utils/array'
 
 const SYNCER_TICK_MS = 10 * 1000
@@ -155,6 +162,9 @@ export class Syncer {
       return
     }
 
+    this.logger.debug('Syncer is beginning peer candidate measurements')
+    const measurementStart = BenchUtils.start()
+
     // Find all allowed peers that have more work than we have
     const peers = this.peerNetwork.peerManager
       .getConnectedPeers()
@@ -240,6 +250,12 @@ export class Syncer {
       return
     }
     const peer = this.peerNetwork.peerManager.getPeer(fastestCandidateIdentity)
+
+    const measurementTime = BenchUtils.end(measurementStart)
+    this.logger.debug(
+      `Syncer took ${TimeUtils.renderSpan(measurementTime)} to measure peer candidates`,
+    )
+
     if (peer) {
       this.startSync(peer)
     }
--- a/ironfish/src/syncer.ts
+++ b/ironfish/src/syncer.ts
@@ -142,7 +142,7 @@ export class Syncer {
         // in-flight requests
         this.state = 'measuring'
         if (this.loader) {
-          this.lastLoaderIdentity = this.loader.getIdentity()
+          this.lastLoaderIdentity = this.loader.identity
           await this.wait()
         }
         break
@@ -188,7 +188,7 @@ export class Syncer {
     // if we have found one.
     if (
       currentPeerIdentity &&
-      !syncCandidates.find((p) => p.getIdentity() === currentPeerIdentity)
+      !syncCandidates.find((p) => p.identity === currentPeerIdentity)
     ) {
       const currentPeer = this.peerNetwork.peerManager.getPeer(currentPeerIdentity)
       if (currentPeer && currentPeer.state.type === 'CONNECTED') {
@@ -212,10 +212,7 @@ export class Syncer {
         break
       }
 
-      const peerIdentity = peer.getIdentity()
-      if (peerIdentity == null) {
-        continue
-      }
+      const peerIdentity = peer.getIdentityOrThrow()
 
       const start = BenchUtils.start()
       try {
--- a/ironfish/src/syncer.ts
+++ b/ironfish/src/syncer.ts
@@ -142,7 +142,7 @@ export class Syncer {
         // in-flight requests
         this.state = 'measuring'
         if (this.loader) {
-          this.lastLoaderIdentity = this.loader.identity
+          this.lastLoaderIdentity = this.loader.state.identity
           await this.wait()
         }
         break
@@ -188,7 +188,7 @@ export class Syncer {
     // if we have found one.
     if (
       currentPeerIdentity &&
-      !syncCandidates.find((p) => p.identity === currentPeerIdentity)
+      !syncCandidates.find((p) => p.state.identity === currentPeerIdentity)
     ) {
       const currentPeer = this.peerNetwork.peerManager.getPeer(currentPeerIdentity)
       if (currentPeer && currentPeer.state.type === 'CONNECTED') {
--- a/ironfish/src/syncer.ts
+++ b/ironfish/src/syncer.ts
@@ -171,7 +171,7 @@ export class Syncer {
       return
     }
 
-    const syncCandidates = ArrayUtils.shuffle(peers).slice(0, 8)
+    const syncCandidates = ArrayUtils.shuffle(peers)
 
     // If we have been syncing from a peer, we want to include this peer in the
     // measurement. This will allow us to maintain a connection to a strong peer
@@ -190,9 +190,16 @@ export class Syncer {
 
     // Measure how long it takes to fetch the genesis block header from each
     // peer as an estimate of connection quality
+    let candidateCheckCount = 0
     for (const peer of syncCandidates) {
       if (this.state === 'stopped' || this.state === 'stopping') {
         return
+      }
+
+      // We only want to successfully measure so many candidates per measurement
+      // phase
+      if (candidateCheckCount > 8) {
+        break
       }
 
       const peerIdentity = peer.getIdentity()
@@ -223,6 +230,7 @@ export class Syncer {
       }
       const rtt = BenchUtils.end(start)
       peerRtt.set(peerIdentity, rtt)
+      candidateCheckCount += 1
     }
 
     // Sort the peers by the round-trip-time of the block header request and get
--- a/ironfish/src/syncer.ts
+++ b/ironfish/src/syncer.ts
@@ -212,6 +212,10 @@ export class Syncer {
         break
       }
 
+      if (peer.state.type !== 'CONNECTED') {
+        continue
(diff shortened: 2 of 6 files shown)


ironfish/src/syncer.ts
```diff
@@ -123,16 +162,90 @@ export class Syncer {
...
+      if (currentPeer && currentPeer.state.type === 'CONNECTED') {
+        syncCandidates.push(currentPeer)
+      }
+    }
+
+    const peerRtt = new Map<Identity, number>()
+
+    // Measure how long it takes to fetch the genesis block header from each
+    // peer as an estimate of connection quality
+    for (const peer of syncCandidates) {
```
dguenther, line 195: I think there's a potential case here where all of the 8 `syncCandidates` could fail to respond to the blockHeader request, and then we end up unable to start syncing even though we may have more peers we could try measuring. Have you thought about continuing to measure peers until we've tried all peers or have measured 8 successfully?

This would have the tradeoff of increasing time spent measuring

ironfish/src/syncer.ts
```diff
@@ -123,16 +162,90 @@ export class Syncer {
...
+      }
+
+      const peerIdentity = peer.getIdentity()
+      if (peerIdentity == null) {
+        continue
+      }
+
+      const start = BenchUtils.start()
+      try {
+        const response = await this.peerNetwork.getBlockHeaders(peer, 1, 1)
```
dguenther, line 207: I think this could probably be explored separately, but I wonder if we could speed up measurement by using `Promise.any` to measure two (or more) peers at once, resolving to the one that completes first. It would use a bit more bandwidth and as such make the test a bit less reliable, but I expect it would reduce the impact of peers that are unusually slow (since their response would get dropped)

ironfish/src/syncer.ts
mat-if: That's a good point. While it does seem statistically unlikely unless you have <8 peers to begin with, it is possible and it would lead to a stuck state until the next measurement time.

There's 2 potential approaches here that I'm thinking of
1. We wrap these 2 lines https://github.com/iron-fish/ironfish/pull/4371/files#diff-fde96c520c62d9d6f4af77d595f4b9775f195d846d3486a939085fe685a643d9R123-R124 in an `if (this.loader) { ... }` block. This would mean that it would try 8 more peers on the next event-loop execution
2. What you've suggested, where instead of just slicing 8 peers off the top, we just iterate until we've measured 8 peers, or exhausted all candidates. There's still the potential that we don't get a single valid response in the situation of, for example, you only having 2-3 peers and they all time out. In this case, we'll likely still need #1. 

Thinking through this, your suggestion does make it feel more robust, so I'll do that along with the check that we did indeed find a peer

ironfish/src/syncer.ts
mat-if: I did consider that, but I felt it was more correct to do them sequentially. In situations where my connection is sufficiently capable, then having simultaneous results would probably not affect the results much. However, if we assume my connection is the bottle-neck, then having multiple requests out simultaneously would definitely taint the results.

It's worth discussing further for sure, as I'm making assumptions here that I haven't _fully_ thought out / tested. During testing though, most of the time spent was the 2 seconds between stopping syncing and the next event loop execution, barring the very occasional response timeout. Rahul and I tested on a c6i.large in Australia, and the results were basically the same. Admittedly, these aren't extremely thorough tests, and the network seems to have a high percentage of powerful, capable peers to sync from right now.

ironfish/src/syncer.ts
dguenther: Sounds good, I think both 1 and 2 you proposed here make sense to me! 👍

ironfish/src/syncer.ts
dguenther: Sounds good! If this hasn't been an issue in testing, let's keep it as-is and make changes if necessary.

ironfish/src/syncer.ts
```diff
@@ -123,16 +162,90 @@ export class Syncer {
...
+      return
+    }
+    const peer = this.peerNetwork.peerManager.getPeer(fastestCandidateIdentity)
+    if (peer) {
+      this.startSync(peer)
+    }
+  }
+
+  startSyncIfIdle(peer: Peer): void {
+    if (this.state === 'idle') {
```
dguenther, line 243: I think we'll want to make sure that `this.nextMeasureTime` is updated when starting to sync directly from this function, otherwise we'll end up with `nextMeasureTime` set to whatever it was last in the event loop. Though generally the event loop should always be running and keeping it up to date.

ironfish/src/syncer.ts
```diff
@@ -123,16 +162,90 @@ export class Syncer {
...
+    }
+
+    // If there is only one valid peer to sync from, there is no point in
+    // measuring the connection so begin syncing immediately
+    if (peers.length === 1) {
+      this.startSync(peers[0])
+      return
+    }
+
+    const syncCandidates = ArrayUtils.shuffle(peers).slice(0, 8)
```
dguenther, line 176: Might be worth adding some logging to the measurement process so we could see logs on when it starts/finishes. I don't feel too strongly about this

ironfish/src/syncer.ts
```diff
@@ -33,11 +43,14 @@ export class Syncer {
...
   readonly downloadSpeed: RollingAverage
 
-  state: 'stopped' | 'idle' | 'stopping' | 'syncing'
+  state: SyncerState
   stopping: Promise<void> | null
   eventLoopTimeout: SetTimeoutToken | null
   loader: Peer | null = null
   blocksPerMessage: number
+  nextMeasureTime = 0
+  numberOfMeasurements = 0
```
patnir, line 52: could you call it simply `measurements`?

ironfish/src/syncer.ts
```diff
@@ -99,46 +111,171 @@ export class Syncer {
...
+      peerRtt.set(peerIdentity, rtt)
+      candidateCheckCount += 1
+    }
+
+    // Sort the peers by the round-trip-time of the block header request and get
+    // the fastest one to sync from
+    const fastestCandidateIdentity = [...peerRtt.entries()].sort((a, b) => a[1] - b[1])[0][0]
+    if (!fastestCandidateIdentity) {
+      return
+    }
```
patnir, line 251: I don't think we need this check. I feel like a better one might be to add a `if (peerRtt.size() < 1) return` before the sort

ironfish/src/syncer.ts
```diff
@@ -99,46 +111,171 @@ export class Syncer {
...
+      // We only want to successfully measure so many candidates per measurement
+      // phase
+      if (candidateCheckCount > 8) {
+        break
+      }
+
+      const peerIdentity = peer.getIdentity()
+      if (peerIdentity == null) {
+        continue
+      }
```
patnir, line 218: What do you think about filtering these peers out before the loop starts? That way you have a better selection anyways

ironfish/src/network/peers/peer.ts
```diff
@@ -412,6 +412,10 @@ export class Peer {
     return this.state.identity
   }
 
+  getIdentity(): Identity | null {
+    return this.state.identity
+  }
```
patnir, line 417: what do you think about changing this to `get identity` instead? that way you just have to do `peer.identity` to access it instead of `peer.getIdentity()`

ironfish/src/syncer.ts
mat-if: Hm, good call-out, technically, we _are_ filtering these out, I just didn't think about this enough. A peer, if its state is connected, has a non-nullable identity field (https://github.com/iron-fish/ironfish/blob/master/ironfish/src/network/peers/peer.ts#L60-L64), and we are calling `getConnectedPeers`. This was originally just being used to convince the compiler of that, so really this should be a `getIdentityOrThrow`

ironfish/src/syncer.ts
mat-if: `numberOfMeasurements` feels more clear at a glance in my opinion. Is `measurements` some object containing multiple types of measurements? Is it an array of a measurement object? I wouldn't expect it to be a counter of sorts if I'm just casually glancing at this file and seeing it for the first time

ironfish/src/syncer.ts
mat-if: > I don't think we need this check.

The TS compiler doesn't seem to pick up on this, but it _is_ possible for the set to be empty.

> I feel like a better one might be to add a if (peerRtt.size() < 1) return before the sort

Better how? :D

ironfish/src/network/peers/peer.ts
dguenther: I'd suggest to make this simpler we just remove `getIdentity()` and use `peer.state.identity` instead, like we do in the rest of the codebase. I'm open to adding `getIdentity` (or an `identity` property), but I think it'd be better to do a whole-code refactor to use it

ironfish/src/syncer.ts
patnir: makes sense, i'll resolve

ironfish/src/syncer.ts
patnir: If the set is empty, wouldn't the [0][0] throw an exception?

ironfish/src/syncer.ts
patnir: yeah that makes sense. ts compiler so weird sometimes.

ironfish/src/network/peers/peer.ts
patnir: I think i'd prefer to keep as is too out of all the options so things are more consistent

ironfish/src/syncer.ts
mat-if: Ah actually thinking about this _more_, there is a possibility that the connection of the peer has changed since the sampling began, we should probably also verify that the peer is still in connected state at the top of the loop

ironfish/src/syncer.ts
```diff
@@ -99,46 +111,171 @@ export class Syncer {
...
+    // If we have been syncing from a peer, we want to include this peer in the
+    // measurement. This will allow us to maintain a connection to a strong peer
+    // if we have found one.
+    if (
+      currentPeerIdentity &&
+      !syncCandidates.find((p) => p.getIdentity() === currentPeerIdentity)
+    ) {
+      const currentPeer = this.peerNetwork.peerManager.getPeer(currentPeerIdentity)
+      if (currentPeer && currentPeer.state.type === 'CONNECTED') {
+        syncCandidates.push(currentPeer)
```
dguenther, line 195: Since we're including all connected peers with more work than us, `currentPeer` is likely to be in `syncCandidates` already. If we want it to be included in one of the 8, I think we should move it to the front of the array.

ironfish/src/syncer.ts
dguenther: Actually, I think this will throw an Error. TS makes nested arrays kinda easy to cause problems. Try in the Node.js repl:

```
const set = new Set()
[...set.entries()].sort((a, b) => a[1] - b[1])[0][0]
Uncaught TypeError: Cannot read properties of undefined (reading '0')
```
You could do this to preserve the original intent:
```
[...set.entries()].sort((a, b) => a[1] - b[1])[0]?.[0]
```

ironfish/src/syncer.ts
danield9tqh: Actually wondering if it would be included. We have [this line](https://github.com/iron-fish/ironfish/blob/staging/ironfish/src/syncer.ts#L449) where we set the peer's work when we sync a block to them. So possibly after we start syncing from someone their work will always be exactly the same as ours?

dguenther: Oh yep, we update the work if the sequence is greater than the peer's sequence. On initial connect to a peer, the peer's sequence + work will likely be HEAD, but I think there are several cases where it might lag behind what blocks the peer actually has. Probably another area we could follow up on. So two cases to handle then:

1. If the peer is in `syncCandidates`, move them to the front of the array
2. If the peer is not in `syncCandidates`, insert them at the front of the array

ironfish/src/syncer.ts
```diff
@@ -99,46 +111,171 @@ export class Syncer {
...
+    // peer as an estimate of connection quality
+    let candidateCheckCount = 0
+    for (const peer of syncCandidates) {
+      if (this.state === 'stopped' || this.state === 'stopping') {
+        return
+      }
+
+      // We only want to successfully measure so many candidates per measurement
+      // phase
+      if (candidateCheckCount > 8) {
```
dguenther, line 211: You could potentially use `peerRtt.size` for this instead. I think this will allow 9 now where the previous algo allowed 8 (+1 if currentPeerIdentity was set), but I don't think the exact number is too important. Maybe it could be pulled out to a constant too?

ironfish/src/syncer.ts
```diff
@@ -99,46 +112,169 @@ export class Syncer {
...
+    if (peerRtt.size === 0) {
+      return
+    }
+
+    // Sort the peers by the round-trip-time of the block header request and get
+    // the fastest one to sync from
+    const fastestCandidateIdentity = [...peerRtt.entries()].sort((a, b) => a[1] - b[1])[0][0]
+    const peer = this.peerNetwork.peerManager.getPeer(fastestCandidateIdentity)
+
+    const measurementTime = BenchUtils.end(measurementStart)
```
dguenther, line 253: Maybe consider bumping this above `if (peerRtt.size === 0) {` so that it still logs the measurementTime even if all the attempts errored out, for some reason (or there weren't any peers). Speaking of which, adding `peerRtt.size` to that log might be helpful too

dguenther (approved):

PR #4373, opened 2023-10-18
processes multiple burns in a single transaction

hughy:
## Summary

for each iteration of the release service loop the burn process combines multiple pending burn requests into a single burn transaction

## Testing Plan

manual testing:
- sent multiple deposits
- ran relay
- ran release and created a single burn transaction for both requests:
```
Burn: [
   {
      "id": 33,
      "asset": "213c1795b756e3237a6b7bd1e30b943c079c38f12d831b870152c96a7a9b7f26",
      "amount": "1"
   },
   {
      "id": 34,
      "asset": "213c1795b756e3237a6b7bd1e30b943c079c38f12d831b870152c96a7a9b7f26",
      "amount": "2"
   }
] f2afe7e76bdbd0b3d3283050e50965a49e562984dc548c91ece31f451c6f8d8e
```

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/service/bridge/release.ts
+++ b/ironfish-cli/src/commands/service/bridge/release.ts
@@ -9,6 +9,7 @@ import {
   RpcSocketClient,
   WebApi,
 } from '@ironfish/sdk'
+import { BurnDescription } from '@ironfish/sdk/src/primitives/burnDescription'
 import { Flags } from '@oclif/core'
 import { IronfishCommand } from '../../../command'
 import { RemoteFlags } from '../../../flags'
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/service/bridge/release.ts
```diff
@@ -210,51 +211,84 @@ export default class Release extends IronfishCommand {
...
+      if (burnDescription.value + BigInt(request.amount) > availableBalance) {
+        continue
+      }
+
+      burnDescription.value += BigInt(request.amount)
+      burnDescriptions.set(assetId, burnDescription)
+      pendingRequests.push(request)
+    }
+
+    if (burnDescriptions.size === 0) {
```
jowparks, line 250: Isn't this a huge problem if this happens? Wonder if we should do anything more than log.

jowparks (approved):

ironfish-cli/src/commands/service/bridge/release.ts
```diff
@@ -210,51 +211,84 @@ export default class Release extends IronfishCommand {
...
     )
 
-    await api.updateBridgeRequests([
-      {
-        id: burnRequest.id,
+    const updatePayload = []
+    for (const request of pendingRequests) {
+      updatePayload.push({
+        id: request.id,
+        destination_transaction: tx.content.hash,
```
hughy, line 286:
```suggestion
        source_burn_transaction: tx.content.hash,
```

ironfish-cli/src/commands/service/bridge/release.ts
hughy: It might only mean that there are pending transactions that prevent us from sending another one. It is a huge problem if we consistently don't have enough balance because then we've lost some tokens 😰

PR #4374, opened 2023-10-18
relays confirmed burn transactions to bridge api

hughy:
## Summary

adds 'getBridgePendingBurnRequests' method to WebApi

for each confirmed block, fetches the list of pending burn requests from the bridge api and compares the transactions in the block to the pending burn transactions

pushes update to api setting status of confirmed burns to 'PENDING_DESTINATION_RELEASE_TRANSACTION_CREATION'

## Testing Plan

manual testing - ran relay to pick up confirmed burn transactions and submit updates to api

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/service/bridge/relay.ts
+++ b/ironfish-cli/src/commands/service/bridge/relay.ts
@@ -154,13 +154,45 @@ export default class BridgeRelay extends IronfishCommand {
   ): Promise<void> {
     Assert.isNotUndefined(response)
 
+    const { requests: pendingBurnRequests } = await api.getBridgePendingBurnRequests()
+
+    const pendingBurnTransactions: Map<string, number[]> = new Map()
+    for (const request of pendingBurnRequests) {
+      Assert.isNotUndefined(request.source_burn_transaction)
+
+      const requestIds = pendingBurnTransactions.get(request.source_burn_transaction) ?? []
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/commands/service/bridge/relay.ts
```diff
@@ -154,13 +154,45 @@ export default class BridgeRelay extends IronfishCommand {
...
+
+        this.log(
+          `Confirmed burn for request IDs [${requestIds.toString()}] in transaction ${
+            transaction.hash
+          }`,
+        )
+
+        for (const requestId of requestIds) {
+          confirmedBurns.push({
+            id: requestId,
```
jowparks, line 188: This seems like the right status to me, just wanted you to double check.

jowparks (approved):

ironfish-cli/src/commands/service/bridge/relay.ts
hughy: I think the status is right, but now that you mention it, we might need to do more here than just update the status to prompt the API to do the release on the other side

hughy: updated to use the `release` endpoint

PR #4380, opened 2023-10-20
Add sorted function to priority queue and move to utils

danield9tqh:
## Summary
Planning to re-use priority queue for other use cases besides just the mempool so moving it to a shared utils package. Also having a `sorted`function that copies the queue and generates the items in order is useful for both use cases so adding that as well

## Testing Plan
Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/memPool/memPool.ts
```diff
@@ -171,22 +171,15 @@ export class MemPool {
...
+    for (const { hash } of this.feeRateQueue.sorted()) {
+      const transaction = this.transactions.get(hash)
 
       // The queue is cloned above, but this.transactions is not, so the
       // transaction may be removed from this.transactions while iterating.
-      if (transaction === undefined) {
-        continue
+      if (transaction !== undefined) {
+        yield transaction
       }
```
patnir, line 182: could you do `if (transaction)` ?

danield9tqh: yea I can change to that

patnir (approved):
Looks good I left a comment

ironfish/src/memPool/index.ts
```diff
@@ -3,4 +3,4 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 export * from './memPool'
 export * from './feeEstimator'
-export * from './priorityQueue'
+export { PriorityQueue } from '../utils/priorityQueue'
```
mat-if, line 6: Is it intentional that we're exporting this from a higher directory here down in the mempool index? This seems kind of strange

danield9tqh: Yea was just about to change this. Also added SDK change to next draft release

PR #4383, opened 2023-10-20
adds sdk support for sending minted assets in same tx

hughy:
## Summary

subtracts mint amounts from amount of each asset needed for funding transactions

if the amount of an asset in the outputs of the transaction is less than the amount minted in the transaction then minted tokens are used directly in the outputs instead of funding the outputs with existing notes

Same as #4382, but for the `bridge-prototype` branch

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


jowparks (commented):
LGTM

PR #4384, opened 2023-10-23
changes Note.memo() to return Buffer

hughy:
## Summary

the 'memo' method on the Note class returns the memo as a human-readable string by using 'BufferUtils.toHuman' on a utf8 decoded string of the memo

this human-readable representation may lose some information from the memo

returning the buffer instead allows sdk clients to choose how to interpret the data encoded in the note memo

for instance, a memo might contain a base64-encoded Ethereum public address. if this string is treated as a human-readable utf8 string, then the encoded address cannot be recovered accurately

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[X] Yes
```

This is a breaking change to the SDK. Clients that use the `Note.memo()` method will need to use `BufferUtils.toHuman(note.memo())` to produce the same value

### DIFF
--- a/ironfish/src/primitives/note.ts
+++ b/ironfish/src/primitives/note.ts
@@ -10,7 +10,6 @@ import {
   RANDOMNESS_LENGTH,
 } from '@ironfish/rust-nodejs'
 import bufio from 'bufio'
-import { BufferUtils } from '../utils/buffer'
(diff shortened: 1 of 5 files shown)


dguenther (approved):
Good change to make. If you could add the breaking change to the next release's draft, that'd be great!

PR #4386, opened 2023-10-24
removes account name trimming in wallet cli inputs

hughy:
## Summary

trimming the account name makes any account with leading or trailing whitespace unusable with the cli

we don't trim whitespace in the rpc or sdk when creating accounts, so it's possible for users to have accounts that have leading or trailing whitespace

limits scope from #4303 to account name inputs in the wallet only

## Testing Plan

manual testing

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/address.ts
+++ b/ironfish-cli/src/commands/wallet/address.ts
@@ -16,7 +16,6 @@ export class AddressCommand extends IronfishCommand {
   static args = [
     {
       name: 'account',
-      parse: (input: string): Promise<string> => Promise.resolve(input.trim()),
       required: false,
       description: 'Name of the account to get the address for',
     },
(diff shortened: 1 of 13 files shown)


mat-if (approved):

NullSoldier:
I want to add some clarification to this PR.

I think the reason this code started to become pervasive is because it's not clear that oclif already trims all arguments.

It was trying to avoid an issue where you would run `ironfish wallet:address     foo` and get an error that `an account with the name     foo` does not exist. However, if you look the argument already comes trimmed as `foo`.

So the trimming really only comes into play when users run `ironfish wallet:address "     foo"` where they explicitly quoted their arg name and have explicitly stated they probably want whitespace before the wallet name.

So not only is this really annoying, but stopping users from doing something they explicitly want to do.

NullSoldier (approved):

PR #4390, opened 2023-10-25
updates getTransactionStream for relay integration

hughy:
## Summary

- optionally decrypts notes for spender - adds optional outgoingViewKey to request - attempts to decrypt using outgoingViewKey if decryption fails using incomingViewKey
- adds sender address to note responses
- returns memo as hex string instead of human-readable utf8 string

supports relay applications (e.g., bridges) that require information on outgoing transactions and may encode non-human-readable data in memos

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[X] Yes
```

Changes the `memo` field in the returned notes to a hex string instead of human-readable utf8

### DIFF
--- a/ironfish/src/rpc/routes/chain/getTransactionStream.ts
+++ b/ironfish/src/rpc/routes/chain/getTransactionStream.ts
@@ -7,9 +7,9 @@ import { ChainProcessor } from '../../../chainProcessor'
 import { FullNode } from '../../../node'
 import { Block } from '../../../primitives/block'
 import { BlockHeader } from '../../../primitives/blockheader'
-import { BufferUtils, CurrencyUtils } from '../../../utils'
+import { CurrencyUtils } from '../../../utils'
 import { PromiseUtils } from '../../../utils/promise'
-import { isValidIncomingViewKey } from '../../../wallet/validator'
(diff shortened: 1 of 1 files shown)


ironfish/src/rpc/routes/chain/getTransactionStream.ts
```diff
@@ -132,16 +144,21 @@ routes.register<typeof GetTransactionStreamRequestSchema, GetTransactionStreamRe
...
+          if (!decryptedNote && request.data.outgoingViewKey) {
+            decryptedNote = note.decryptNoteForSpender(request.data.outgoingViewKey)
+          }
 
           if (decryptedNote) {
             const assetValue = await node.chain.getAssetById(decryptedNote.assetId())
             notes.push({
               value: CurrencyUtils.encode(decryptedNote.value()),
-              memo: BufferUtils.toHuman(decryptedNote.memo()),
+              memo: decryptedNote.memo().toString('hex'),
```
hughy, line 157: We could make this `utf8` to preserve some backwards compatibility and still avoid removing non-human-readable characters.

On the other hand, `hex` is consistent with how we treat all other buffers in the RPC

jowparks (approved):

PR #4391, opened 2023-10-25
Rahul fixing void calls to file system in address manager

patnir:
## Summary

According to https://nodejs.org/api/fs.html#filehandlewritefiledata-options 

image

This adds a basic mutex functionality to prevent fileWrite from being called multiple times without the promise completing. 

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/peers/peerStoreManager.test.ts
+++ b/ironfish/src/network/peers/peerStoreManager.test.ts
@@ -3,6 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 jest.mock('ws')
 
+import { PromiseUtils } from '../../utils'
 import {
   getConnectedPeer,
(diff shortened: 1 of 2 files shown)


danield9tqh:
I think this should probably be implemented at a lower level like where we call `fsPromises.writeFile()`. Also instead of returning if save is already being called I would maybe look at how we're doing it on the [database lock](https://github.com/iron-fish/ironfish/blob/staging/ironfish/src/storage/levelup/database.ts#L53). It might be more correct for subsequent calls to save to queue up after the current call is finished instead of just returning early

patnir:
Going to fix it in the root class we inherit from https://github.com/iron-fish/ironfish/blob/staging/ironfish/src/fileStores/fileStore.ts#L34

PR #4392, opened 2023-10-25
Move compute state from connections into setState

danield9tqh:
## Summary
Move compute state from connections into setState

## Testing Plan
Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


mat-if (approved):
awyis look at that code, much cleaner

PR #4393, opened 2023-10-26
Move PeerManager start/stop functions near constructor

mat-if:
## Summary

We maintain this pattern pretty consistently throughout the codebase, and it makes it hard to find the start/stop functions easily because they're buried deep in the file compared to other classes.

## Testing Plan

Existing unit tests

## Documentation

N/A

## Breaking Change

N/A


patnir (approved):
way better

PR #4398, opened 2023-10-26
Test multiple NodeJS versions in CI

mat-if:
## Summary

Since we officially support NodeJS v18 and v20 (https://github.com/iron-fish/ironfish/pull/4395), we should test them in CI as well.

## Testing Plan

CI

## Documentation

N/A

## Breaking Change

N/A


jowparks (approved):
Do we only support rust build with node 18? Wondering why you didn't put a matrix on that too.

mat-if:
> Do we only support rust build with node 18? Wondering why you didn't put a matrix on that too.

Building the binaries with node 18 appears to work on 20, so I'm cautiously going forward without adding that additional work unless we need it.

PR #4401, opened 2023-10-26
returning list of brokering peers to try a variety of brokering peers for web rtc connections

patnir:
## Summary

Removes the logical inconsistency that would allow trying the same peer as a brokering peer multiple times even if we didn't necessarily need it more than once.

This tackles that by changing the logic in the `initWebRtcConnection` function

## Testing Plan

In draft state at the moment, I'll update unit tests if this is viable solution

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/peers/peerManager.ts
+++ b/ironfish/src/network/peers/peerManager.ts
@@ -639,12 +639,12 @@ export class PeerManager {
 
     for (const neighbor of val.neighbors) {
       const neighborPeer = this.identifiedPeers.get(neighbor)
-      if (!neighborPeer || neighborPeer.state.type !== 'CONNECTED') {
-        val.neighbors.delete(neighbor)
-        continue
-      }
 
-      candidates.push(neighborPeer)
+      if (neighborPeer && neighborPeer.state.type === 'CONNECTED') {
+        candidates.push(neighborPeer)
+      } else {
+        val.neighbors.delete(neighbor)
+      }
     }
 
     return candidates
--- a/ironfish/src/network/peers/peerManager.ts
+++ b/ironfish/src/network/peers/peerManager.ts
@@ -258,9 +258,9 @@ export class PeerManager {
     }
 
     // Make sure we can find at least one brokering peer before we create the connection
-    const brokeringPeers = this.getBrokeringPeers(peer)
+    const hasBrokeringPeers = this.hasBrokeringPeers(peer)
 
-    if (brokeringPeers.length === 0) {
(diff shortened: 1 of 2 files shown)


ironfish/src/network/peers/peerManager.ts
```diff
@@ -609,37 +617,37 @@ export class PeerManager {
...
+      } else {
         val.neighbors.delete(neighbor)
-        continue
       }
-
-      candidates.push(neighborPeer)
     }
 
-    return ArrayUtils.sample(candidates)
+    return candidates
```
mat-if, line 650: What do you think about adding the shuffle here, since the callsites (that are actually interested in the results) call shuffle anyway? I'd also suggest adding a `hasBrokeringPeer` function, for the cases where we don't want the overhead of materializing the array. Specifically these spots, really: https://github.com/iron-fish/ironfish/pull/4401/files#diff-52ccbf1a59c642f18891176e72f472cab95d1439d6c9e458741facc0394ac8e7R357
https://github.com/iron-fish/ironfish/pull/4401/files#diff-52ccbf1a59c642f18891176e72f472cab95d1439d6c9e458741facc0394ac8e7R261

patnir: Moved the shuffling operation into that function. 

I think the problem with adding a hasBrokeringPeer function is that to determine if there are brokering peers, you have to go through the entire logic of the getBrokeringPeers function because the last loop in that function can still return an empty array. 

```
 const candidates = []

    for (const neighbor of val.neighbors) {
      const neighborPeer = this.identifiedPeers.get(neighbor)

      if (neighborPeer && neighborPeer.state.type === 'CONNECTED') {
        candidates.push(neighborPeer)
      } else {
        val.neighbors.delete(neighbor)
      }
    }

    return ArrayUtils.shuffle(candidates)
```

mat-if: It could still be a lot lighter, if I'm not misunderstanding. Something like:
``` typescript
for (const neighbor of val.neighbors) {
  const neighborPeer = this.identifiedPeers.get(neighbor)
  if (neighborPeer && neighborPeer.state.type === 'CONNECTED') {
    return true
  }
}
return false
```

You're still iterating but you're not creating the new array nor doing the shuffle

patnir: I'm not sure what the unintended consequence of what you have above which eliminates the `val.neighbors.delete(neighbor)` call

mat-if (approved):

PR #4402, opened 2023-10-27
adding save and create file mutex

patnir:
## Summary

According to the documentation: "It is unsafe to use filehandle.writeFile() multiple
times on the same file without waiting for the promise to be fulfilled (or rejected)."
https://nodejs.org/api/fs.html#filehandlewritefiledata-options
We have had several complaints especially from Windows users about the json file being corrupted.
We call this function several places in our codebase without waiting for the promise to complete.
This mutex is used to prevent multiple writes to the same file.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


danield9tqh (approved):

patnir:
I'm re-writing the test to test the mutex more thoroughly

PR #4403, opened 2023-10-27
Normalize WebSocket fields on Peer

danield9tqh:
## Summary
A WebSocket address with a port but no host is not a valid address. Catch these cases earlier so we can have better type safety in the rest of the code

## Testing Plan
Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/peers/peer.ts
+++ b/ironfish/src/network/peers/peer.ts
@@ -68,6 +68,11 @@ export enum KnownBlockHashesValue {
   Sent = 2,
 }
 
+export type WebSocketAddress = {
+  host: string
+  port: number | null
+}
+
 export class Peer {
   readonly pendingRPCMax: number
   readonly logger: Logger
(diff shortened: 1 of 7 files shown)


ironfish/src/network/peers/peer.ts
```diff
@@ -439,22 +444,8 @@ export class Peer {
...
-    }
-
-    // Don't do anything if the address and port stay the same
-    if (address === this._address && port === this._port) {
-      return
-    }
-
-    this._address = address
-    this._port = port
+  setWebSocketAddress(wsAddress: WebSocketAddress | null): void {
```
mat-if, line 447: Still sorta thinking through this, but I wonder if it makes more sense to have this function signature stay how it was (`setWebSocketAddress(address: string | null, port: number | null)`).

It feels like we're creating a less-safe function by making it the responsibility of the function-caller to do the error check that this function was doing. By keeping the function signature the same, we encapsulate this a bit better: function callers don't have to care about the specifics, and internally we get to utilize this type

It would be nice if we could have a constructor on a type definition, so we could contain the logic there, but alas. Making it a full-fledged class seems like the wrong approach, as well.

mat-if (approved):

PR #4404, opened 2023-10-30
Normalize WS Address on Peer Candidates

danield9tqh:
## Summary
For Peer candidates remove the unrepresentable state of having a null address but a non-null port. Also in this PR changing the way we process PeerList messages. If a peer sends us a PeerList containing a null `address` but non-null `port` just set the `wsAddress` to null instead of throwing an error. This would cause the node to crash if a peer sent us a message like that

## Testing Plan
Running local node + unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/peers/peerManager.ts
+++ b/ironfish/src/network/peers/peerManager.ts
@@ -507,7 +507,7 @@ export class PeerManager {
       disconnectOk &&
       hasNoConnection &&
       retryOk &&
-      peer.address !== null
+      peer.wsAddress !== null
     )
   }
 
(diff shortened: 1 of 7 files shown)


ironfish/src/network/peers/peerManager.ts
```diff
@@ -1445,10 +1444,6 @@ export class PeerManager {
         continue
       }
 
-      if (connectedPeer.port !== null && connectedPeer.address === null) {
```
danield9tqh, line 1448: Changing handling of PeerList messages here. If the peer candidate does not contain an `address` then ignore the port as well

mat-if (approved):

PR #4405, opened 2023-10-30
adding peer direction to rpc peers command

patnir:
## Summary

Adding connection direction field to the peers command. Required an RPC change for the GetPeer and GetPeers endpoints

image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/peers/index.ts
+++ b/ironfish-cli/src/commands/peers/index.ts
@@ -47,6 +47,11 @@ export class ListCommand extends IronfishCommand {
       default: false,
       description: 'Display node names',
     }),
+    direction: Flags.boolean({
+      char: 'c',
+      default: false,
+      description: 'Display connection direction',
+    }),
     features: Flags.boolean({
       default: false,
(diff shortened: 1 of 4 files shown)


mat-if (approved):
Excellent improving for debugging and working on peer network stuff, thanks!

ironfish-cli/src/commands/peers/index.ts
```diff
@@ -47,6 +47,11 @@ export class ListCommand extends IronfishCommand {
       default: false,
       description: 'Display node names',
     }),
+    direction: Flags.boolean({
```
danield9tqh, line 50: I think we have an extended flag already we could just hide it behind that? Not sure it needs its own flag we might be getting into too many custom flags

patnir: Ah good call out, I didn't know that's how the extended field works. Yeah totally agree, too many flags. I made the change.

danield9tqh (approved):

PR #4407, opened 2023-10-31
simplifying logic in peer connection manager - connectToEligiblePeers and eventLoop

patnir:
## Summary

Easier to read/ understand code. Some improvements to double checking the peer candidate object exists

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/peers/peerCandidates.ts
+++ b/ironfish/src/network/peers/peerCandidates.ts
@@ -76,8 +76,8 @@ export class PeerCandidates {
     }
   }
 
-  shufflePeerCandidates(): string[] {
-    return ArrayUtils.shuffle([...this.map.keys()])
+  shufflePeerCandidates(): [string, PeerCandidate][] {
+    return ArrayUtils.shuffle([...this.map.entries()])
   }
 
   get(identity: Identity): PeerCandidate | undefined {
--- a/ironfish/src/network/peers/peerCandidates.ts
+++ b/ironfish/src/network/peers/peerCandidates.ts
@@ -76,8 +76,8 @@ export class PeerCandidates {
     }
   }
 
-  shufflePeerCandidates(): [Identity, PeerCandidate][] {
-    return ArrayUtils.shuffle([...this.map.entries()])
+  shufflePeerCandidates(): Identity[] {
+    return ArrayUtils.shuffle([...this.map.keys()])
   }
 
   get(identity: Identity): PeerCandidate | undefined {
--- a/ironfish/src/network/peers/peerCandidates.ts
+++ b/ironfish/src/network/peers/peerCandidates.ts
@@ -76,7 +76,7 @@ export class PeerCandidates {
     }
   }
 
-  shufflePeerCandidates(): [string, PeerCandidate][] {
+  shufflePeerCandidates(): [Identity, PeerCandidate][] {
     return ArrayUtils.shuffle([...this.map.entries()])
   }
 
--- a/ironfish/src/network/peers/peerConnectionManager.ts
+++ b/ironfish/src/network/peers/peerConnectionManager.ts
@@ -114,11 +114,7 @@ export class PeerConnectionManager {
       return false
     }
 
-    return (
-      (this.peerManager.canConnectToWebRTC(peer) && this.peerManager.connectToWebRTC(peer)) ||
-      (this.peerManager.canConnectToWebSocket(peer) &&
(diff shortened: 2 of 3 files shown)


ironfish/src/network/peers/peerConnectionManager.ts
```diff
@@ -84,24 +84,24 @@ export class PeerConnectionManager {
...
+      peerCandidateIdentity,
+      peerCandidate,
+    ] of this.peerManager.peerCandidates.shufflePeerCandidates()) {
       if (connectAttempts >= CONNECT_ATTEMPTS_MAX) {
         break
       }
 
       if (!this.peerManager.identifiedPeers.has(peerCandidateIdentity)) {
-        const val = this.peerManager.peerCandidates.get(peerCandidateIdentity)
-        if (val) {
```
patnir, line 94: this check is basically not required because the peercandidateidentity is in the peer candidate list (which we are iterating over)

ironfish/src/network/peers/peerConnectionManager.ts
```diff
@@ -110,21 +110,11 @@ export class PeerConnectionManager {
   }
 
   private connectToEligiblePeers(peer: Peer): boolean {
-    if (peer.state.type !== 'CONNECTED') {
-      if (this.peerManager.canConnectToWebRTC(peer)) {
```
patnir, line 114: canConnectToWebRTC is called in connectToWebRTC

ironfish/src/network/peers/peerConnectionManager.ts
```diff
@@ -110,21 +110,11 @@ export class PeerConnectionManager {
...
-        if (this.peerManager.connectToWebRTC(peer)) {
-          return true
-        }
-      }
-
-      if (this.peerManager.canConnectToWebSocket(peer)) {
```
patnir, line 120: canConnectToWebSocket is called in connectToWebSocket

ironfish/src/network/peers/peerConnectionManager.ts
```diff
@@ -110,21 +110,11 @@ export class PeerConnectionManager {
...
-        if (this.peerManager.connectToWebSocket(peer)) {
-          return true
-        }
-      }
+    if (peer.state.type === 'CONNECTED') {
+      return false
     }
 
-    return false
+    return this.peerManager.connectToWebRTC(peer) || this.peerManager.connectToWebSocket(peer)
```
danield9tqh, line 117: I think on this I still would rather prefer 
```javascript
if(this.peerManager.connectToWebRTC(peer)) {
  return true
}

return this.peerManager.connectToWebSocket(peer)
```

Just makes it a little more explicit that an action is being taken and `connectToWebRTC(peer)` is not just a boolean

ironfish/src/network/peers/peerConnectionManager.ts
patnir: image

ironfish/src/network/peers/peerConnectionManager.ts
patnir: image

ironfish/src/network/peers/peerConnectionManager.ts
patnir: done

danield9tqh (approved):

danield9tqh (approved):

ironfish/src/network/peers/peerCandidates.ts
```diff
@@ -76,8 +76,8 @@ export class PeerCandidates {
     }
   }
 
-  shufflePeerCandidates(): string[] {
-    return ArrayUtils.shuffle([...this.map.keys()])
+  shufflePeerCandidates(): [Identity, PeerCandidate][] {
```
mat-if, line 79: I don't think this is a good change - previously we were allocating an array of size this.map.size * identity length, now we're allocating an array of this.map.size * (identity length + peer candidate size) which is substantially more, just to avoid looking the candidate up from a hashmap we already have. Readability is good, but not at the cost of performance

mat-if (changes requested):

ironfish/src/network/peers/peerCandidates.ts
patnir: That's fair, I changed it back. Now that specific change just has a few variable name changes

dguenther (approved):

PR #4410, opened 2023-11-01
Add keepOpenPeerSlot flag; disconnect from peers if above limit

mat-if:
## Summary

The peer connection manager will now disconnect from peers if there are more than maxPeers connected peers. Additionally, if keepOpenPeerSlot is set, it will disconnect if there are more than maxPeers - 1 connected peers. This can be used by nodes who want to ensure they are allowing incoming connections regardless of them being near/at capacity

## Testing Plan

Unit tests
Manual testing

## Documentation

```
[x] Yes
```

https://github.com/iron-fish/website/pull/564

## Breaking Change

N/A

### DIFF
--- a/ironfish/src/network/peers/peerConnectionManager.ts
+++ b/ironfish/src/network/peers/peerConnectionManager.ts
@@ -3,8 +3,10 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 import type { Peer } from './peer'
+import { DEFAULT_KEEP_OPEN_PEER_SLOT } from '../../fileStores/config'
 import { createRootLogger, Logger } from '../../logger'
-import { SetTimeoutToken } from '../../utils'
+import { ArrayUtils, SetTimeoutToken } from '../../utils'
+import { DisconnectingReason } from '../messages/disconnecting'
 import { PeerManager } from './peerManager'
 
 /**
@@ -30,6 +32,7 @@ export class PeerConnectionManager {
   private readonly logger: Logger
   private readonly peerManager: PeerManager
   readonly maxPeers: number
+  readonly keepOpenPeerSlot: boolean
 
   private started = false
   private eventLoopTimer?: SetTimeoutToken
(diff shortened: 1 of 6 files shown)


ironfish/src/network/peers/peerConnectionManager.ts
```diff
@@ -106,9 +111,43 @@ export class PeerConnectionManager {
...
     this.eventLoopTimer = setTimeout(() => this.eventLoop(), EVENT_LOOP_MS)
   }
 
+  /**
+   * Maintain a maximum number of peers by disconnecting from peers if we are
+   * connected to more than we should be
+   */
+  private maintainMaxPeerCount(): void {
+    const connectedPeers = this.peerManager.getConnectedPeers()
+    const maxPeerCount = this.maxPeers - Number(this.keepOpenPeerSlot)
```
patnir, line 125: is this like controlling whether this feature is on or not? if so, would it be better to just add an if statement and exit the function if the flag is false? More readable that way?

ironfish/src/network/peers/peerConnectionManager.ts
```diff
@@ -39,11 +42,13 @@ export class PeerConnectionManager {
     logger: Logger = createRootLogger(),
     options: {
       maxPeers: number
+      keepOpenPeerSlot?: boolean
```
patnir, line 45: Alternative: `keepOpenPeerSlot: boolean = DEFAULT_KEEP_OPEN_PEER_SLOT`

mat-if: Type literals can't have an initializer: 
image


You can kinda emulate this behavior but it's a lot more verbose than the existing pattern. We do it somewhere else but I can't remember where off the top of my head

ironfish/src/network/peers/peerConnectionManager.ts
mat-if: > is this controlling whether this feature is on or not?

No, this explicitly adds the behavior to shed peers if above maxPeers regardless of the flag being enabled. This is helpful in combination with #4350 where we open the door of being above maxPeers. This line just determines whether we're going to shed if peers are _above_ maxPeers (default behavior), or if we are _at_ maxPeers (to maintain an open slot, if this flag is enabled.

patnir (approved):

PR #4413, opened 2023-11-01
Use WSAddress type on incoming WS connections

danield9tqh:
## Summary
Parse a WebSocketAddress type from incoming websocket connections to use consistently instead of host + port

## Testing Plan
Running local node + unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/network/peerNetwork.ts
```diff
@@ -309,7 +309,13 @@ export class PeerNetwork {
           return
         }
 
-        this.peerManager.createPeerFromInboundWebSocketConnection(connection, address)
+        let wsAddress: WebSocketAddress | null = null
```
mat-if, line 312: What do you think about creating a function for going from address/port -> WebSocketAddress in the util file?

ironfish/src/network/utils/parseUrl.ts
```diff
@@ -2,6 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+export type WebSocketAddress = {
```
mat-if, line 5: We should consider changing the filename since it is no longer specific to the parseUrl function

mat-if (approved):
Couple minor things but otherwise looks good 👍

ironfish/src/network/peerNetwork.ts
danield9tqh: yea I think we could def change the parseUrl a bit to output that. Because that function should probably never return `null` for the `hostname`. If there's no `hostname` it's prob invalid. I'll try to do that in a separate PR

ironfish/src/network/utils/parseUrl.ts
danield9tqh: agree, I'll make that change in this PR

PR #4419, opened 2023-11-03
Implement a safer file write function

mat-if:
## Summary

When writing a file, we now do a couple extra things:
- Write to a temp file first, and rename if the write was successful. Usually, renames are considered atomic operations, so if it fails for some reason, we should keep the original file intact.
- After writing the contents of the file, we call sync to try to force a buffer flush and ensure the file is written before we close it. NodeJS implements this functionality by adding a `flush` option to the `writeFile` utility function starting in NodeJS 21, but it will be a long time before we can use that.

Closes IFL-1795

## Testing Plan

Manual testing, keyStore and fileStore unit tests

## Documentation

N/A

## Breaking Change

N/A


patnir (approved):

patnir:
Do we want to replace write file everywhere? We can do it in a separate PR?

PR #4420, opened 2023-11-03
chore(IFL-1355): removes reference to transmission key

jowparks:
## Summary
We do not use a diversifier in ironfish. This decision was made to allow auditability of fund sources when someone is receiving funds and uses their incoming view key to decrypt incoming funds.

Because of this transparency of sender, there is no longer a concept of a transmission key, which was an opaque version of the public address. This changes removes the vestigial naming.
## Testing Plan

Tests pass

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/keys/public_address.rs
+++ b/ironfish-rust/src/keys/public_address.rs
@@ -10,7 +10,7 @@ use group::GroupEncoding;
 use ironfish_zkp::constants::PUBLIC_KEY_GENERATOR;
 use jubjub::SubgroupPoint;
 
-use std::{convert::TryInto, io};
+use std::io;
 
 use super::{IncomingViewKey, SaplingKey};
 pub const PUBLIC_ADDRESS_SIZE: usize = 32;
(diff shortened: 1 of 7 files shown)


ironfish-rust/src/keys/public_address.rs
```diff
@@ -19,19 +19,14 @@ pub const PUBLIC_ADDRESS_SIZE: usize = 32;
...
-    /// form a public address to which payments can be sent.
-    pub(crate) transmission_key: SubgroupPoint,
-}
+pub struct PublicAddress(pub(crate) SubgroupPoint);
 
 impl PublicAddress {
     /// Initialize a public address from its 32 byte representation.
     pub fn new(address_bytes: &[u8; PUBLIC_ADDRESS_SIZE]) -> Result<PublicAddress, IronfishError> {
-        let transmission_key = PublicAddress::load_transmission_key(&address_bytes[0..])?;
+        let public_address = PublicAddress::load_public_address(&address_bytes[0..])?;
```
mat-if, line 27: Bit of a nit but what do you think about just making `new` do the stuff in `load_public_address` and removing the `load_public_address` function? There used to be more stuff here but now it feels unnecessary to keep that extra layer of indirection

mat-if (approved):
thanks for tackling this 🙌

PR #4422, opened 2023-11-06
Only loop through peer candidates if we can create new connections

mat-if:
## Summary

Previously, this loop would get executed every event loop iteration regardless of whether it could even create new connections. Now, it will only run if we can actually create new connections, saving a bit of overhead when it is not useful.

Also includes a light refactor to break this chunk of code into a function for easier reading and testing.

## Testing Plan

Unit tests, manually tested

## Documentation

N/A

## Breaking Change

N/A

### DIFF
--- a/ironfish/src/network/peers/peerConnectionManager.ts
+++ b/ironfish/src/network/peers/peerConnectionManager.ts
@@ -87,6 +87,21 @@ export class PeerConnectionManager {
       }
     }
 
+    this.attemptNewConnections()
+    this.maintainMaxPeerCount()
+
+    this.eventLoopTimer = setTimeout(() => this.eventLoop(), EVENT_LOOP_MS)
+  }
+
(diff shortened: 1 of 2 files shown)


ironfish/src/network/peers/peerConnectionManager.ts
```diff
@@ -87,35 +87,48 @@ export class PeerConnectionManager {
...
+
+    this.eventLoopTimer = setTimeout(() => this.eventLoop(), EVENT_LOOP_MS)
+  }
+
+  /**
+   * Attempts to connect to a number of peer candidates if it is eligible to
+   * create new connections
+   */
+  private attemptNewConnections(): void {
+    if (!this.peerManager.canCreateNewConnections()) {
```
mat-if, line 101: This is the only new thing

PR #4433, opened 2023-11-06
Fix an RPC timing attack

NullSoldier:
## Summary

This fixes a timing attack where the comparater for strings is not constant time. You can find the password over time by trying different bytes which allow you to get closer and closer to the actual password.

See https://en.wikipedia.org/wiki/Timing_attack

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/server.ts
+++ b/ironfish/src/rpc/server.ts
@@ -96,8 +96,14 @@ export class RpcServer {
       return false
     }
 
-    const userTokenBuffer = Buffer.from(requestAuthToken)
-    const authTokenBuffer = Buffer.from(rpcAuthToken)
+    const maxLength = Math.max(requestAuthToken.length, rpcAuthToken.length)
+
+    const userTokenBuffer = Buffer.alloc(maxLength)
+    userTokenBuffer.write(requestAuthToken)
+
+    const authTokenBuffer = Buffer.alloc(maxLength)
+    authTokenBuffer.write(rpcAuthToken)
+
     return timingSafeEqual(userTokenBuffer, authTokenBuffer)
   }
 
--- a/ironfish/src/rpc/server.ts
+++ b/ironfish/src/rpc/server.ts
@@ -4,7 +4,7 @@
 
 import { randomBytes, timingSafeEqual } from 'crypto'
 import { Assert } from '../assert'
-import { InternalOptions, InternalStore } from '../fileStores'
+import { InternalStore } from '../fileStores'
 import { createRootLogger, Logger } from '../logger'
 import { IRpcAdapter } from './adapters'
 import { ApiNamespace, RequestContext, Router, routes } from './routes'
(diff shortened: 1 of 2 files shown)


jowparks (approved):

ironfish/src/rpc/server.ts
```diff
@@ -90,10 +91,11 @@ export class RpcServer {
...
 
     if (!rpcAuthToken) {
-      this.logger.debug(`Missing RPC Auth token in internal.json config.`)
       return false
     }
 
-    return requestAuthToken === rpcAuthToken
+    const userTokenBuffer = Buffer.from(requestAuthToken)
+    const authTokenBuffer = Buffer.from(rpcAuthToken)
+    return timingSafeEqual(userTokenBuffer, authTokenBuffer)
```
andiflabs, line 99: Note that this will throw an error if the two buffers/tokens have a different length. The current version of this function would return `false` instead

NullSoldier: Good point. This didn't fail locally, so I'll fix the code and also add a test that will fail in this case.

ironfish/src/rpc/server.ts
```diff
@@ -86,30 +94,51 @@ export class RpcServer {
...
     }
 
-    const rpcAuthToken = this.internal.get('rpcAuthToken')
+    if (!this.authTokenBuffer.byteLength) {
+      return false
+    }
 
-    if (!rpcAuthToken) {
-      this.logger.debug(`Missing RPC Auth token in internal.json config.`)
+    if (requestAuthToken.length >= AUTH_MAX_LENGTH) {
```
andiflabs, line 104: Nit: I think this should be a `>`, not a `>=`

ironfish/src/rpc/server.ts
```diff
@@ -86,30 +94,51 @@ export class RpcServer {
...
-      this.internal.set('rpcAuthToken', newPassword)
-      await this.internal.save()
+      this.loadAuth(this.internal.get('rpcAuthToken'))
+      return
     }
+
+    this.logger.debug(
+      `Missing RPC Auth token in internal.json config. Automatically generating auth token.`,
+    )
+    const newPassword = randomBytes(AUTH_MAX_LENGTH).toString('hex')
```
andiflabs, line 124: This will produce a string that is `2 * AUTH_MAX_LENGTH` bytes long. Passing this string to `authenticate()` won't be possible because it will fail the length check. This should perhaps be:

```
randomBytes(AUTH_MAX_LENGTH / 2).toString('hex')
```

ironfish/src/rpc/server.ts
NullSoldier: Fixed

ironfish/src/rpc/server.ts
NullSoldier: Fixed.

andiflabs (approved):

PR #4435, opened 2023-11-07
Add WSAddress type to connection

danield9tqh:
## Summary
Changing WebSocketConnection to use address type instead of host and port

## Testing Plan
Running locally + unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/peers/connections/webSocketConnection.ts
+++ b/ironfish/src/network/peers/connections/webSocketConnection.ts
@@ -24,7 +24,9 @@ export class WebSocketConnection extends Connection {
   }
 
   set port(port: number) {
-    this._address = this._address ? { host: this._address.host, port } : null
+    if (this._address) {
+      this._address.port = port
+    }
   }
 
   constructor(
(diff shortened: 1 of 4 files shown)


ironfish/src/network/peers/connections/webSocketConnection.ts
```diff
@@ -17,24 +18,26 @@ import { NetworkError } from './errors'
...
-  readonly hostname?: string
+  private _address: WebSocketAddress | null
+  get address(): WebSocketAddress | null {
+    return this._address
+  }
 
-  // The port of the address that was used to establish the WebSocket connection, if any
-  port?: number
+  set port(port: number) {
+    this._address = this._address ? { host: this._address.host, port } : null
```
mat-if, line 27: What do you think about a more traditional if check? I don't think it's necessary to create a new object just to set the port

```suggestion
    if (this._address) {
      this._address.port = port
    }
```

mat-if (approved):
Minor thing but otherwise looks good

ironfish/src/network/peers/connections/webSocketConnection.ts
danield9tqh: yea sounds good to me

PR #4441, opened 2023-11-08
Change RpcServer to use RpcContext

NullSoldier:
## Summary

This allows us to isolate the requirements down to what each route needs. This is a form of dependency injection where the route needs to satisfy it's own dependencies, rather than have a DI system do this. A DI system would be fine too, I just didn't want to pull one into the code base.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


andiflabs (approved):

ironfish/src/assert.ts
```diff
@@ -100,4 +100,16 @@ export class Assert {
       throw new Error(message || 'Expected buffers to have the same contents')
     }
   }
+
+  static hasKeys<TObj extends Record<string, unknown>, TKeys extends keyof TObj>(
+    obj: TObj,
+    keys: TKeys[],
+    message?: string,
+  ): asserts obj is TKeys extends keyof TObj ? Required<Pick<TObj, TKeys>> : never {
```
andiflabs, line 108: Is it possible that `TKeys extends keyof TObj` is `false`? I'm wondering if this ternary is needed at all

NullSoldier: It's more to convince the type system. Here is the error without the ternary.
```
A type predicate's type must be assignable to its parameter's type.
  Type 'Required<Pick<TObj, TKeys>>' is not assignable to type 'TObj'.
    'Required<Pick<TObj, TKeys>>' is assignable to the constraint of type 'TObj', but 'TObj' could be instantiated with a different subtype of constraint 'Record<string, unknown>'
``` 
I did a lot of debugging and the closest I can get is that TObj cannot be asserted to Required<Pick<TObj, TKeys>> because we need to convince TS that TKeys truly is Keyof TObj. I'm not really sure why.

PR #4442, opened 2023-11-08
Sorting for peers command 

patnir:
## Summary

Since we are displaying a single table, we can sort the output for the follow state.

image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/peers/index.ts
+++ b/ironfish-cli/src/commands/peers/index.ts
@@ -19,10 +19,7 @@ export class ListCommand extends IronfishCommand {
   static flags = {
     ...RemoteFlags,
     ...tableFlags,
-    sort: {
-      ...sort,
-      exclusive: ['follow'],
-    },
+    sort,
     follow: Flags.boolean({
       char: 'f',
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/peers/index.ts
```diff
@@ -19,10 +19,7 @@ export class ListCommand extends IronfishCommand {
   static flags = {
     ...RemoteFlags,
     ...tableFlags,
-    sort: {
-      ...sort,
-      exclusive: ['follow'],
-    },
+    sort,
```
mat-if, line 22: I think you can just replace L21 and 22 with `...CliUx.ux.table.flags()` and can remove the de-structure on L14 since we're not modifying the `sort` key anymore

mat-if (approved):
minor nit, but good usability win

ironfish-cli/src/commands/peers/index.ts
patnir: Done

PR #4447, opened 2023-11-10
Update TS to 5.0.4 and node-datachannel to 0.5.1

dguenther:
## Summary

The current version of node-datachannel, 0.4.3, causes a crash in the Electron app, so we'd like to update to 0.5.0, but it's an ES Module, which doesn't work great with our CommonJS setup.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/jest.setup.env.js
```diff
@@ -4,28 +4,6 @@
 const consola = require('consola')
 const { initializeSapling } = require('@ironfish/rust-nodejs')
 
-jest.mock('node-datachannel', () => {
```
danield9tqh, line 7: Just kinda curious what the reason for removing was? Assuming its because the new version of `node-datachannel` doesn't match this shape but then does it need to be mocked everywhere still?

danield9tqh: Ok nvm I think I understand now because we are passing it down from the SDK

dguenther: Main reason is actually you can't do module-level mocks in `jest` of ES Module packages. I think we could use something like `esmock`, but didn't chase it down too far since I found the current solution.

ironfish-cli/package.json
```diff
@@ -26,18 +26,19 @@
     "@oclif/test": "2.1.0",
     "@types/blessed": "0.1.17",
     "@types/inquirer": "8.2.5",
-    "@types/node": "18.11.16",
+    "@types/node": "20.9.0",
```
danield9tqh, line 29: Does this assume that host machine will be running node v20?

ironfish/src/network/peers/localPeer.ts
```diff
@@ -24,6 +24,8 @@ export class LocalPeer {
   readonly version: number
   // constructor for either a Node WebSocket or a browser WebSocket
   readonly webSocket: IsomorphicWebSocketConstructor
+  // asynchronously imported WebRTC datachannel library
+  readonly nodeDataChannel: NodeDataChannelType
```
danield9tqh, line 28: Was there a reason we had this on the localPeer? Guessing it's because the `webSocket` is also here but I couldn't figure out why that was here in the first place either

ironfish-cli/package.json
dguenther: Let me see if it still works if I revert this actually -- I'm not sure if we needed to bump that or not

ironfish/src/network/peers/localPeer.ts
dguenther: I put it here because `webSocket` was here. I don't necessarily care where it goes, but if we're going to move both of them, I'd like to do it in a separate PR

danield9tqh: Gotcha yea was just curious. Have no preference for doing it in this PR

ironfish-cli/package.json
dguenther: Nice, looks like this still works when I roll back the change. Going to resolve this

ironfish/src/network/peers/connections/webRtcConnection.ts
```diff
@@ -193,15 +196,18 @@ export class WebRtcConnection extends Connection {
       return false
     }
 
-    if (!this.datachannel.isOpen()) {
-      this.logger.debug('Datachannel no longer open, closing connection')
-      this.close()
-      return false
+    let sent = false
```
danield9tqh, line 199: Curious was there another error you ran into here beside the channel being closed?

dguenther: I think we don't actually need this now, I'll try to remove it. I started tinkering around with the code when I found that you can't mock functions in ES Module packages, and we were mocking the implementation of `isOpen`. But I think the `connection` class already handles try/catching errors and logging them, so we probably don't need either the `try/catch` here, or `isOpen`.

danield9tqh: Ah yep you're right, good call

danield9tqh (approved):
Looks good, I tested locally running both Node 18 + 20. I was getting some WebRTC errors but nothing that I wasn’t also getting on staging and still was able to get up to max peers

ironfish/src/network/peers/connections/webRtcConnection.ts
dguenther: All right, I rolled this back, but I did leave the change that uses the return value of `sendMessageBinary`, since I think we should be handling that value instead of ignoring it.

PR #4457, opened 2023-11-22
Adding note hash to view transaction output

patnir:
## Summary

Debugging + UX improvement for what the output for the user looks like after a transaction 

image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/transaction/index.ts
+++ b/ironfish-cli/src/commands/wallet/transaction/index.ts
@@ -92,6 +92,10 @@ export class TransactionCommand extends IronfishCommand {
       }
 
       CliUx.ux.table(noteAssetPairs, {
+        noteHash: {
+          header: 'Note Hash',
+          get: ({ note }) => note.noteHash,
+        },
         amount: {
           header: 'Amount',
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/wallet/transaction/index.ts
```diff
@@ -92,6 +92,10 @@ export class TransactionCommand extends IronfishCommand {
       }
 
       CliUx.ux.table(noteAssetPairs, {
+        noteHash: {
```
danield9tqh, line 95: Probably would be better if note hash wasn't the first in the list. It seems more like a power user item. Maybe even hiding this behind an `--extended` flag like we do in other Oclif CLI tables would be good

NullSoldier: agreed, we could hide this behind extended. Also you don't need to say "Note Hash" when this is listing notes. If it says, "Hash" it's the hash of the resource being displayed.

patnir:
Yeah, I agree it's not required. I'll close

PR #4465, opened 2023-11-29
adding ability to select notes in the transaction send command

patnir:
## Summary

Add ability to send notes in the transaction command.

After a lot of back and forth with these PRs (https://github.com/iron-fish/ironfish/pull/4464 and https://github.com/iron-fish/ironfish/pull/4459), I decided it was best to at least get this out there for our users and work on more UX improvements including warnings for larger transactions in subsequent PRs. 

## Testing Plan

Manual testing. 

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/send.ts
+++ b/ironfish-cli/src/commands/wallet/send.ts
@@ -90,6 +90,11 @@ export class Send extends IronfishCommand {
       default: false,
       description: 'Allow offline transaction creation',
     }),
+    notes: Flags.string({
+      char: 'n',
+      description: 'The notes to include in the transaction',
+      multiple: true,
+    }),
   }
 
   async start(): Promise<void> {
@@ -189,6 +194,7 @@ export class Send extends IronfishCommand {
       feeRate: flags.feeRate ? CurrencyUtils.encode(flags.feeRate) : null,
       expiration: flags.expiration,
       confirmations: flags.confirmations,
+      notes: flags.notes,
     }
 
     let raw: RawTransaction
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/wallet/send.ts
```diff
@@ -90,6 +90,11 @@ export class Send extends IronfishCommand {
       default: false,
       description: 'Allow offline transaction creation',
     }),
+    notes: Flags.string({
+      char: 'n',
+      description: 'The notes to include in the transaction',
```
hughy, line 95: nit - maybe something like `note hashes to include`? just to be a little more explicit about what needs to be passed here

ironfish-cli/src/commands/wallet/send.ts
```diff
@@ -90,6 +90,11 @@ export class Send extends IronfishCommand {
       default: false,
       description: 'Allow offline transaction creation',
     }),
+    notes: Flags.string({
```
NullSoldier, line 93: If you are using the multiple flag, this should be `note`

patnir: you can also `--notes 72e91d6672d79f9006f128a16cf52bbb585037fee2aea8385b56919c3ac9570b 4a5feb193d425fa73abfb108027a643eb027e46684c50e2bef5d66ea48ec3d2d f8b91f81caf4489c83734c2b29ad4587c9ba0c1080854a3f6da19f6fad49265e` which I think is slightly simpler

patnir: I looked at the docs and I think it seems singular naming conventions are standard, so I'll change it to `--note`

NullSoldier (approved):

ironfish-cli/src/commands/wallet/send.ts
NullSoldier: yeah note hashes would be more explicit

hughy (approved):

PR #4471, opened 2023-12-04
Add warnings about a change to the RedDSA spec

andiflabs:
## Summary

The RedDSA specification (and the Rust crates that implement it) has slightly deviated from the code we're using. The changes make the signatures backward compatible, but the code we have written needs to change in order to be compatible. This commit adds some comments to the relevant code so that if in the future someone attempts to update the RedDSA crates, they won't have to spend hours to figure out what's wrong.

## Testing Plan

Not a code change

## Documentation

No doc change required

## Breaking Change

Not a breaking change

### DIFF
--- a/ironfish-rust/src/transaction/mints.rs
+++ b/ironfish-rust/src/transaction/mints.rs
@@ -127,6 +127,11 @@ impl UnsignedMintDescription {
             return Err(IronfishError::new(IronfishErrorKind::InvalidSigningKey));
         }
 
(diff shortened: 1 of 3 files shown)


jowparks (approved):
LGTM, we just need to link to changes that we end up making for the multisig `partial_post` changes once we PR for that, since that will be a likely breadcrumb

hughy (approved):

PR #4472, opened 2023-12-05
Rahul/ifl 1867 order by field in getnotes rpc

patnir:
## Summary

Adding a sortByValue field to the getNotes RPC. Some design tradeoffs I made:

1. I used getNotes RPC instead of creating a new one because the filtering benefits we get will likely be useful in this context as well 
2. Two additional fields need to be explicitly defined `spent` and `assetId` because in order to use the existing `unspentNotes` index we have in the DB. This allows us to only load the noteHash and value (64 bytes) into memory for all notes. If we were to allow this sorting for all assets and regardless of spends would be inefficient and slow because we would have to load the entire set of notes into memory. 

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/getNotes.ts
+++ b/ironfish/src/rpc/routes/wallet/getNotes.ts
@@ -89,7 +89,9 @@ routes.register<typeof GetNotesRequestSchema, GetNotesResponse>(
 
     if (
       request.data.sortByValue &&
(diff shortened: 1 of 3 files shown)


patnir:
Decided that it's better to add an index/ more efficient way to sort here instead of loading into memory.

PR #4475, opened 2023-12-07
Add ability to extend Config

NullSoldier:
https://github.com/iron-fish/ironfish/pull/4475/files?diff=split&w=1

This allows you to extend Config with custom extended config types. It
does now require you to pass the extended defeults into the Config()
constructor even if you are not extending it for type reasons.

This makes the Config API slightly more burdensome to use but it's a
small price to pay. This also leads to a breaking compatability change in
the SDK in the case that third parties are creating a Config. I'll document
this in the release notes.

### DIFF
--- a/ironfish/src/fileStores/config.ts
+++ b/ironfish/src/fileStores/config.ts
@@ -382,18 +382,30 @@ export const ConfigOptionsSchema: yup.ObjectSchema<Partial<ConfigOptions>> = yup
   })
   .defined()
 
-export class Config extends KeyStore<ConfigOptions> {
+export class Config<
+  TExtend extends Record<string, unknown> = Record<string, unknown>,
+> extends KeyStore<ConfigOptions & TExtend> {
   readonly chainDatabasePath: string
   readonly walletDatabasePath: string
   readonly tempDir: string
 
-  constructor(files: FileSystem, dataDir: string, configName?: string) {
+  constructor(
+    files: FileSystem,
+    dataDir: string,
+    extend: TExtend,
+    configName?: string,
+    extendSchema?: yup.ObjectSchema<Partial<TExtend>>,
+  ) {
+    const schema = (
+      extendSchema ? YupUtils.union(ConfigOptionsSchema, extendSchema) : ConfigOptionsSchema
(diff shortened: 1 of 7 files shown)


ironfish/src/node.ts
```diff
@@ -214,7 +215,7 @@ export class FullNode {
     dataDir = dataDir || DEFAULT_DATA_DIR
 
     if (!config) {
-      config = new Config(files, dataDir)
+      config = WalletConfig.NewConfig(files, dataDir)
```
NullSoldier, line 218: This is demonstrating that an extended config could be passed into the SDK components.

ironfish/src/mining/poolDatabase/database.test.ts
```diff
@@ -21,7 +21,7 @@ describe('poolDatabase', () => {
     const fileSystem = new NodeFileProvider()
     await fileSystem.init()
     // TODO(mat): It would be convenient if we didn't need a filesystem for Config for tests
-    const config = new Config(fileSystem, dataDir)
+    const config = Config.Init(fileSystem, dataDir)
```
hughy, line 24: should this be `Config.NewConfig`? FWIW I think I prefer the syntax of something like `Config.new` or `Config.init`

NullSoldier: Just fixed this, but it actually revealed that the solution doesn't work without a cast.

ironfish/src/fileStores/config.ts
```diff
@@ -310,109 +310,116 @@ export type ConfigOptions = {
...
+    configName?: string,
+  ) {
+    super(files, configName || DEFAULT_CONFIG_NAME, defaults, dataDir, schema)
 
     this.chainDatabasePath = this.files.join(this.storage.dataDir, 'databases', 'chain')
     this.walletDatabasePath = this.files.join(this.storage.dataDir, 'databases', 'wallet')
     this.tempDir = this.files.join(this.storage.dataDir, 'temp')
   }
 
+  static Init(files: FileSystem, dataDir: string, configName?: string): Config {
```
andiflabs, line 418: How about `fromFile` or `fromFilesystem`?

dguenther (approved):

PR #4476, opened 2023-12-08
removes spender_key from ProposedTransaction

hughy:
## Summary

when signing transactions with FROST we will not have a full, valid SaplingKey to pass as the 'spender_key'. FROST will also require that public key randomness is shared between signers with randomized parameters used as an input to signing.

spender_key is only used when posting ProposedTransactions, so these changes remove it from the struct and instead pass it as an input to each of the post functions

~~public_key_randomness is only used in the context of _partial_post, so these changes remove public_key_randomness from the struct and initialize a randomized public address from randomness only when needed~~

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust-nodejs/src/structs/transaction.rs
+++ b/ironfish-rust-nodejs/src/structs/transaction.rs
@@ -248,7 +248,7 @@ impl NativeTransaction {
         let spender_key = SaplingKey::from_hex(&spender_hex_key).map_err(to_napi_err)?;
         let transaction = self
             .transaction
-            .post_miners_fee(spender_key)
+            .post_miners_fee(&spender_key)
             .map_err(to_napi_err)?;
 
         let mut vec: Vec<u8> = vec![];
@@ -263,7 +263,7 @@ impl NativeTransaction {
         let spender_key = SaplingKey::from_hex(&spender_hex_key).map_err(to_napi_err)?;
         let transaction = self
             .transaction
-            .post_miners_fee_unchecked(spender_key)
+            .post_miners_fee_unchecked(&spender_key)
             .map_err(to_napi_err)?;
 
         let mut vec: Vec<u8> = vec![];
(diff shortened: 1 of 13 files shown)


ironfish-rust-nodejs/src/structs/transaction.rs
```diff
@@ -245,8 +244,12 @@ impl NativeTransaction {
     /// a miner would not accept such a transaction unless it was explicitly set
     /// as the miners fee.
     #[napi(js_name = "post_miners_fee")]
-    pub fn post_miners_fee(&mut self) -> Result<Buffer> {
-        let transaction = self.transaction.post_miners_fee().map_err(to_napi_err)?;
+    pub fn post_miners_fee(&mut self, spender_hex_key: String) -> Result<Buffer> {
```
andiflabs, line 247: Consider changing `spender_hex_key: String` to `spender_hex_key: &str` if possible (same for `_post_miners_fee_unchecked`)

ironfish-rust/src/transaction/mod.rs
```diff
@@ -204,6 +190,7 @@ impl ProposedTransaction {
     /// aka: self.value_balance - intended_transaction_fee - change = 0
     pub fn post(
         &mut self,
+        spender_key: SaplingKey,
```
andiflabs, line 193: Wondering if this should be a `&SaplingKey`, to avoid excessive copying to the stack

hughy: I'll update!

ironfish-rust-nodejs/src/structs/transaction.rs
hughy: I think the type here needs to be `String` for napi support

jowparks (approved):
LGTM

PR #4482, opened 2023-12-13
Remove unused account parameter in getAccountsStatus and add getAccountStatus

dguenther:
## Summary

getAccountsStatus has a few types named getAccountStatus, so adjusted them for consistency. Also, the docs specify that getAccountsStatus takes an account as a parameter, but that parameter isn't actually used anywhere, so removed it and added a separate getAccountStatus RPC for fetching status for a single account.

## Documentation

```
[x] Yes
```

https://github.com/iron-fish/website/pull/573

## Breaking Change

```
[x] Yes
```

* `getAccountsStatus` no longer accepts an account parameter (it did nothing)
* `getAccountsStatus` now returns accountStatus objects with shape:
```
{
  name: string
  id: string
  head: {
    hash: string
    sequence: number
    inChain: boolean | null
  } | null
  viewOnly: boolean
}
```


jowparks (approved):

NullSoldier:
I think this is fine, though we've been opting to have more simpler routes if possible. That would mean creating a `getAccountStatus` and doing some code share on serialization.

dguenther:
Yeah, as I looked at the wallet app a bit more, I realized we'd probably want an RPC for just fetching status for a single account as well. I'll see if I can add that here

hughy (approved):

NullSoldier (approved):

PR #4484, opened 2023-12-14
Feature/support wasm

peinlcy:
## Summary

I want to compile ironfish-rust and ironfish-zkp to WebAssembly (wasm). Currently, ironfish-rust uses libc::c_void and libc::c_int, but there is no libc library in wasm. Therefore, these parts need to be replaced with ffi::c_void and ffi::c_int, which have equivalent implementations in wasm. Additionally, in ironfish-zkp, the rand library is used for generating random numbers, and this library relies on the getrandom library at its core. To support this functionality in wasm, it is necessary to enable js feature for the getrandom library, as described in https://docs.rs/getrandom/#webassembly-support.

## Documentation

None

## Breaking Change

None

### DIFF
--- a/ironfish-rust/src/signal_catcher.rs
+++ b/ironfish-rust/src/signal_catcher.rs
@@ -2,9 +2,18 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+ #[cfg(target_family = "wasm")]
+use core::ffi::c_void;
+#[cfg(target_family = "wasm")]
+use core::ffi::c_int;
+#[cfg(not(target_family = "wasm"))]
+use libc::c_void;
+#[cfg(not(target_family = "wasm"))]
+use libc::c_int;
+
 extern "C" {
     // This is present in libc on unix, but not on linux
-    fn backtrace_symbols_fd(buffer: *const *mut libc::c_void, size: libc::c_int, fd: libc::c_int);
+    fn backtrace_symbols_fd(buffer: *const *mut c_void, size: c_int, fd: c_int);
 }
 
 /// # Safety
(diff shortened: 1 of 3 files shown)


NullSoldier:
Sorry for the delay, we're working on reviewing this soon.

andiflabs:
`ironfish-rust` depends on some crates that I believe do not have WASM support, namely the blstrs crate, which contains some native ASM code. I might be wrong, but I do not think it will be possible to ever compile `ironfish-rust` to WASM unless the blst dependency is replaced or made compatible with WASM.

Could you please elaborate on how this change was tested? Were you able to get a fully functional WASM build?

ironfish-zkp/Cargo.toml
```diff
@@ -29,6 +29,8 @@ ff = "0.12.0"
 group = "0.12.0"
 jubjub = { git = "https://github.com/iron-fish/jubjub.git", branch = "blstrs" }
 lazy_static = "1.4.0"
+getrandom = { version = "0.2", features = ["js"] }
```
andiflabs, line 32: This feature should not be generally available for anyone, see the [getrandom documentation](https://docs.rs/getrandom/latest/getrandom/#webassembly-support):

> This feature should only be enabled for binary, test, or benchmark crates. Library crates should generally not enable this feature, leaving such a decision to *users* of their library. Also, libraries should not introduce their own `js` features just to enable `getrandom`’s `js` feature.

One of the reasons for that is that WASM may be used in environments other than a web browser and Node.js, and this feature restricts usage.

ironfish-rust/src/signal_catcher.rs
```diff
@@ -2,9 +2,18 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+ #[cfg(target_family = "wasm")]
```
andiflabs, line 5: This whole Rust file won't be usable in WASM environments, so I'd rather skip the compilation of this file altogether rather than tweaking imports.

Besides that, `ffi::c_int` and `libc::c_int` are interchangable. In fact, they're defined in the same way. I would just change everything to `use core::ffi::c_int` and remove the `use libc::c_int`. Same thing for `c_void`.

ironfish-zkp/Cargo.toml
peinlcy: Yes, but it is specially prepared for wasm, if we don't enable it, there will be compilation problems

peinlcy:
> `ironfish-rust` depends on some crates that I believe do not have WASM support, namely the blstrs crate, which contains some native ASM code. I might be wrong, but I do not think it will be possible to ever compile `ironfish-rust` to WASM unless the blst dependency is replaced or made compatible with WASM.
> 
> Could you please elaborate on how this change was tested? Were you able to get a fully functional WASM build?

PR #4495, opened 2023-12-19
Remove unused field from block fetcher

danield9tqh:
## Summary
It doesn't look like this field on the block fetcher is ever being accessed. It is making BlockHeader refactor for the hashing algorithm more compliacated so removing it

## Testing Plan
Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/blockFetcher.ts
+++ b/ironfish/src/network/blockFetcher.ts
@@ -3,7 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 import { BufferMap } from 'buffer-map'
-import { Block, CompactBlock } from '../primitives/block'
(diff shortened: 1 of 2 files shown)


dguenther:
My guess is that this is somewhat relevant to the case where we've received a block and have sent it out, but haven't added it to the store -- we might want to respond to requests for the block by hash. I think this is more relevant to the transaction case though, since this one is not used.

I do like the style of having `receivedCompactBlock` take a compact block though, it helps make sure you actually do have a compact block at that point.

mat-if (approved):

danield9tqh:
Closing in favor of https://github.com/iron-fish/ironfish/pull/4496

PR #4498, opened 2023-12-20
Rahul/ifl 2006 combine command produces a single note as output

patnir:
## Summary

Changing the amount the transaction so that the combine command produces 1 note as the output: 

image

image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/notes/combine.ts
+++ b/ironfish-cli/src/commands/wallet/notes/combine.ts
@@ -469,8 +469,7 @@ export class CombineNotesCommand extends IronfishCommand {
       raw = RawTransactionSerde.deserialize(bytes)
     }
 
-    // This allows us to have a single note as an output.
-
+    // This allows for a single note output.
     const amount = amountIncludingFees - raw.fee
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/wallet/notes/combine.ts
```diff
@@ -469,6 +469,18 @@ export class CombineNotesCommand extends IronfishCommand {
       raw = RawTransactionSerde.deserialize(bytes)
     }
 
+    // This allows for a single note output.
+    const amount = amountIncludingFees - raw.fee
+    params.outputs[0].amount = CurrencyUtils.encode(amount)
+    params.fee = raw.fee ? CurrencyUtils.encode(raw.fee) : null
```
hughy, line 475: nit - do we need the ternary operator here? I think `fee` is always non-null on a `RawTransaction`

hughy (approved):

ironfish-cli/src/commands/wallet/notes/combine.ts
patnir: Removed! Thank you!

PR #4499, opened 2023-12-21
Rahul/ifl 1867 sorted notes by value db index

patnir:
## Summary

New index that keeps decrypted notes sorted by value. 

This has the effect of giving the user the best chance to include the fewest notes in a transaction. 

## Testing Plan

Added unit tests + manual testing

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/wallet.test.ts
```diff
@@ -915,6 +916,39 @@ describe('Wallet', () => {
     })
   })
 
+  describe('addSpendsForAsset', () => {
+    it('should select notes in order of largest to smallest', async () => {
```
patnir, line 920: double checked and this fails on staging/ master

patnir:
Feedback from Jason: 

- Change the index to just be unspent notes. This way the index stays small and performance does not degrade over time. 
- async *loadNotesSortedByValue should return noteHashes instead of the decrytedNote.

patnir:
@NullSoldier I'm working on the updates and will close this PR

PR #4505, opened 2024-01-04
fix: binary gha bug with zip filename

jowparks:
## Summary

This PR addresses an issue with the binary github action (gha) builder/uploader.

Somehow there was no error when running when the `github.event*` were not populated (when testing on a feature branch). But when they were (during the context of a release) and error was occurring as zip file extention wasn't being included by the zip step. This explicitly adds the `.zip` extention and resolves the error.

For release v`1.15.0` I manually created a branch off master, retrieved the upload url for the release via github actions:
```bash
curl -H "Authorization: token :token" \
     -H "Accept: application/vnd.github.v3+json" \
     https://api.github.com/repos/iron-fish/ironfish/releases/tags/v1.15.0
```
And manually updated the `${{github.event.*}}` references to point to that release. The manually run GHA off master then uploaded the files to the release page.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/.github/workflows/publish-binaries.yml
+++ b/.github/workflows/publish-binaries.yml
@@ -42,18 +42,12 @@ jobs:
     runs-on: ${{ matrix.settings.host }}
     steps:
 
-      - name: clean selfhosted node_modules
-        if: matrix.settings.system == 'apple' && matrix.settings.arch == 'arm64'
-        run: |
-          cd $GITHUB_WORKSPACE
-          find . -name . -o -prune -exec rm -rf -- {} +
-
-      - name: Use Node.js
+      - name: Use node.js
         uses: actions/setup-node@v4
         with:
           node-version: 18
       
-      - name: Use Go
+      - name: Use go
         uses: actions/setup-go@v4
         with:
           go-version: '1.20.6'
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-binaries.yml
```diff
@@ -163,10 +157,10 @@ jobs:
         with:
           directory: tools
           type: 'zip'
```
jowparks, line 159: Bug was here,

.github/workflows/publish-binaries.yml
```diff
@@ -187,13 +181,13 @@ jobs:
...
           if-no-files-found: error
 
-      - name: Upload Release Asset
+      - name: Upload release asset
         id: upload-release-asset
         uses: actions/upload-release-asset@v1
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           upload_url: "${{ github.event.release.upload_url }}?name=${{ steps.set_paths.outputs.zip }}"
```
jowparks, line 190: This would have been a bug too.

.github/workflows/publish-binaries.yml
```diff
@@ -42,18 +42,12 @@ jobs:
     runs-on: ${{ matrix.settings.host }}
     steps:
```
jowparks, line 44: No longer needed since we aren't on self hosted runners for mac arm anymore.

dguenther (approved):

PR #4508, opened 2024-01-06
Use readonly BlockHeader fields in verifier tests

danield9tqh:
## Summary
BlockHeader class fields will be changed to be readonly so that a BlockHeader fields cannot be accidentally updated without updating the hash. This requires updating many of the test suites to not directly modify block header fields

## Testing Plan
Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/consensus/verifier.test.ts
```diff
@@ -484,16 +519,40 @@ describe('Verifier', () => {
...
-      header.graffiti = Buffer.alloc(33)
+      const invalidHeader33Byte = new BlockHeader(
+        header.sequence,
+        header.previousBlockHash,
+        header.noteCommitment,
+        header.transactionCommitment,
+        header.target,
+        header.randomness,
+        header.timestamp,
+        Buffer.alloc(31),
```
mat-if, line 549:
```suggestion
        Buffer.alloc(33),
```

mat-if (approved):
1 little thing, and it looks good 👍

PR #4510, opened 2024-01-06
Move block explorer urls to CLI client side and out of the SDK

danield9tqh:
## Summary
This PR accomplishes two tasks:
1. The CLI now shows the correct link for transactions whether they are made on mainnet or testnet
2. Any logic relating to ironfish's block explorer is removed from the SDK

The second change is related to an ongoing effort to only have information that is necessary for the blockchain operation in the SDK. Finding blocks and transactions on the block explorer is a layer on top of the network but not part of the network itself. For this it seems to make sense to remove this information from the SDK to the CLI.

## Testing Plan
- Tested running a pool on mainnet + testnet to see if urls match
- Tested transaction send on testnet to see if URL is correct

## Documentation

```
[X] Yes
```
https://github.com/iron-fish/website/pull/581

## Breaking Change

```
[X] Yes
```

This removes two fields from the SDK config `explorerBlocksUrl` and `explorerTransactionsUrl`

### DIFF
--- a/ironfish-cli/src/utils/explorer.ts
+++ b/ironfish-cli/src/utils/explorer.ts
@@ -16,12 +16,12 @@ export const getExplorerUrl = (networkId: number): string | null => {
   }
 }
 
-export const getTransactionUrl = (networkId: number, txId?: string): string | null => {
+export const getExplorerTransactionUrl = (networkId: number, txId: string): string | null => {
   const explorerUrl = getExplorerUrl(networkId)
-  return explorerUrl ? `${explorerUrl}/transaction${txId ? `/${txId}` : ''}` : null
+  return explorerUrl ? `${explorerUrl}/transaction/${txId}` : null
 }
 
-export const getBlockUrl = (networkId: number, blockHash?: string): string | null => {
+export const getExplorerBlockUrl = (networkId: number, blockHash: string): string | null => {
   const explorerUrl = getExplorerUrl(networkId)
-  return explorerUrl ? `${explorerUrl}/blocks${blockHash ? `/${blockHash}` : ''}` : null
+  return explorerUrl ? `${explorerUrl}/blocks/${blockHash}` : null
 }
--- a/ironfish-cli/src/utils/explorer.ts
+++ b/ironfish-cli/src/utils/explorer.ts
@@ -3,8 +3,8 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 export const EXPLORER_URLS = {
-  0: 'https://explorer.ironfish.network',
-  1: 'https://testnet.explorer.ironfish.network',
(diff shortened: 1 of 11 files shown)


ironfish-cli/src/utils/explorer.ts
```diff
@@ -0,0 +1,27 @@
...
+  }
+}
+
+export const getTransactionUrl = (networkId: number, txId?: string): string | null => {
+  const explorerUrl = getExlorerUrl(networkId)
+  return explorerUrl ? `${explorerUrl}/transaction${txId ? `/${txId}` : ''}` : null
+}
+
+export const getBlockUrl = (networkId: number, blockHash?: string): string | null => {
+  const explorerUrl = getExlorerUrl(networkId)
```
dguenther, line 25:
```suggestion
export const getExplorerUrl = (networkId: number): string | null => {
  switch (networkId) {
    case 0:
      return EXPLORER_URL_TESTNET
    case 1:
      return EXPLORER_URL_MAINNET
    default:
      return null
  }
}

export const getTransactionUrl = (networkId: number, txId?: string): string | null => {
  const explorerUrl = getExplorerUrl(networkId)
  return explorerUrl ? `${explorerUrl}/transaction${txId ? `/${txId}` : ''}` : null
}

export const getBlockUrl = (networkId: number, blockHash?: string): string | null => {
  const explorerUrl = getExplorerUrl(networkId)
```

ironfish/src/fileStores/config.ts
```diff
@@ -473,8 +461,6 @@ export class Config extends KeyStore<ConfigOptions> {
       poolMaxConnectionsPerIp: 0,
       poolLarkWebhook: '',
       jsonLogs: false,
-      explorerBlocksUrl: 'https://explorer.ironfish.network/blocks/',
-      explorerTransactionsUrl: 'https://explorer.ironfish.network/transaction/',
```
dguenther, line 477: I feel okay with removing this from the SDK config, since it's just used as a UI feature. Though this makes me think that it'd be nice to have a way to have config settings that are CLI-specific, and also a way to have mapped configs...

Had you thought about including them in the network definition? Not really sure whether we'd want that since it again is more SDK-focused

danield9tqh: Yea, I've thought this before too having a separate CLI config. Another config value that should be CLI specific for example is `dataDir` or node `rpcTcpHost` and `rpcTcpPort`. Really those should have separate values for the CLI that tell the CLI which node to connect to. We just happen to re-use the node (server side) config values for connection which works most of the time but also results in some weird edge cases. So I definitely agree with CLI config but maybe we save for a separate PR.

Network definition is interesting yea but ultimately I agree I think its too CLI specific to add there

dguenther (approved):

PR #4525, opened 2024-01-10
Rahul/ifl 2084 unspent note hashes sorted by value

patnir:
## Summary

New index that keeps UNSPENT decrypted note hashes sorted by value. 

This has the effect of giving the user the best chance to include the fewest notes in a transaction. 

## Testing Plan

Unit tests + Integration tests

Manual tests: 
1. use the wallet combine notes command 
2. it should create a large note
3. when sending your next transaction that large note will be used first

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/wallet.test.ts
```diff
@@ -257,6 +258,39 @@ describe('Wallet', () => {
     expect(await accountA.getNoteHash(forkSpendNullifier)).toBeUndefined()
   })
 
+  describe('addSpendsForAsset', () => {
```
patnir, line 261: this test fails on staging

ironfish/src/migrations/data/030-value-to-unspent-note.ts
```diff
@@ -0,0 +1,72 @@
...
+        await stores.new.valueToUnspentNoteHash.put(
+          [account.prefix, [note.note.assetId(), [note.note.value(), noteHash]]],
+          null,
+        )
+        decryptedNotes++
+      }
+
+      logger.info(
+        ` Indexed ${decryptedNotes} notes and sorted by value for account ${account.name}`,
+      )
```
hughy, line 63: nit - may be worth clarifying that this is the number of unspent notes

ironfish/src/wallet/account/account.test.ts
```diff
@@ -2179,6 +2179,171 @@ describe('Accounts', () => {
...
+        account,
+        node.wallet,
+      )
+
+      await node.chain.addBlock(minerBlockB)
+      await node.wallet.updateHead()
+
+      const block = await useMinerBlockFixture(node.chain, undefined, account, node.wallet, [
+        await useTxFixture(node.wallet, account, account),
+        await useTxFixture(node.wallet, account, account),
```
hughy, line 2209: nit - might be helpful to set the value of these transactions so that there are multiple notes with distinct values

hughy (approved):

ironfish/src/wallet/wallet.ts
```diff
@@ -852,22 +852,6 @@ export class Wallet {
     return account.getBalance(assetId, confirmations)
   }
 
-  private async *getUnspentNotes(
```
NullSoldier, line 855: How will people know how to get unspent notes now in our SDK since this is going away? Is there a reason we're deleting this? It seems like this will make our SDK more confusing.

patnir: it is a private function, so i'm assuming it won't be exported from our sdk. so i don't think they would be using it? or are you saying that they would use this as an example?

ironfish/src/wallet/walletdb/walletdb.ts
```diff
@@ -604,6 +660,61 @@ export class WalletDB {
       [account.prefix, [assetId, [sequence, [value, noteHash]]]],
       tx,
     )
+    await this.deleteValueToUnspentNoteHash(account, decryptedNote, tx)
+  }
+
+  async *loadUnspentNoteHashesByValue(
```
NullSoldier, line 666: This function says it's loading note hashes but it's not.

Should it be, `loadUnspentNotesByValue`? It's the mirror function of `loadUnspentNotes`.

ironfish/src/wallet/walletdb/walletdb.ts
```diff
@@ -604,6 +660,61 @@ export class WalletDB {
...
+    )) {
+      const decryptedNote = await this.decryptedNotes.get([account.prefix, noteHash], tx)
+
+      if (decryptedNote !== undefined) {
+        yield decryptedNote
+      }
+    }
+  }
+
+  async *loadSortedUnspentNoteHashes(
```
NullSoldier, line 690: This function is actually what the above function name is doing. The precedence is to use `loadIndexName` so I think this should be `loadValueToUnspentNoteHashes` since all this is is an index loading function.

ironfish/src/wallet/walletdb/walletdb.ts
patnir: yeah that is a more consistent, i changed it.

ironfish/src/wallet/walletdb/walletdb.ts
patnir: I ended up removing this function and updating the one in `ironfish/src/wallet/account/account.ts` to include the logic of checking whether the decrypted note is undefined.

ironfish/src/wallet/walletdb/walletdb.test.ts
```diff
@@ -47,6 +55,126 @@ describe('WalletDB', () => {
     })
   })
 
+  describe('loadUnspentNotesByValue', () => {
+    it('notes should be returned in order of value', async () => {
+      const { node } = await nodeTest.createSetup()
```
NullSoldier, line 60: this should not be called in a test, it needs to be called in the describe or bugs can occur because it hooks into jests's beforeEach and afterAll.

ironfish/src/wallet/walletdb/walletdb.test.ts
```diff
@@ -47,6 +55,126 @@ describe('WalletDB', () => {
...
+        null,
+      )
+
+      const allNotes = await walletDb.valueToUnspentNoteHashes.getAll()
+
+      const correctOrder = [1n, 10n, 100n, 1000n, 10000n]
+
+      const values = allNotes.map((note) => note[0][1][1][0])
+
+      expect(values).toEqual(correctOrder)
```
NullSoldier, line 132: This could be a bit more concise and still pretty clean I think

```suggestion
      const { wallet } = await nodeTest.createSetup()
      const account = await useAccountFixture(wallet)

      const values = [100n, 1n, 10n, 1000n, 10000n]
      const expected = [1n, 10n, 100n, 1000n, 10000n]

      await Promise.all(
        values.map((value) =>
          wallet.walletDb.valueToUnspentNoteHashes.put(
            [account.prefix, [Asset.nativeId(), [value, Buffer.alloc(32, Math.random())]]],
            null,
          ),
        ),
      )

      const allNotes = await wallet.walletDb.valueToUnspentNoteHashes.getAll()
      expect(allNotes.map((index) => index[0][1][1][0])).toEqual(expected)
```

ironfish/src/migrations/data/030-value-to-unspent-note/new/index.ts
```diff
@@ -0,0 +1,42 @@
...
+          8,
+        ),
+        32,
+      ),
+      4,
+    ),
+    valueEncoding: NULL_ENCODING,
+  })
+
+  return { valueToUnspentNoteHash: valueToUnspentNoteHash }
```
NullSoldier, line 41: This should be in a file called `AccountValue`, but it looks like the standard for migrations broke with migration 24. I'll go in and fix them in another PR.

ironfish/src/migrations/data/030-value-to-unspent-note.ts
```diff
@@ -0,0 +1,74 @@
...
+        )
+        unspentNotes++
+      }
+
+      logger.info(
+        ` Indexed ${unspentNotes} unspent notes and sorted by value for account ${account.name}`,
+      )
+    }
+  }
+
```
NullSoldier, line 68: I feel like this could be a lot simpler with something like this...

```ts
async forward(
  context: MigrationContext,
  db: IDatabase,
  tx: IDatabaseTransaction | undefined,
  logger: Logger,
): Promise<void> {
  const stores = GetStores(db)

  let unspentNotes = 0

  for await (const [account, [assetId, [sequence, [value, noteHash]]]] of stores.old.unspentNoteHashes.getAllKeysIter()) {
    await stores.new.valueToUnspentNoteHash.put([account, [assetId, [value, noteHash]]], null)

    unspentNotes++
  }

  logger.info(` Indexed ${unspentNotes} unspent notes into valueToUnspentNoteHash`)
}
```

ironfish/src/wallet/walletdb/walletdb.test.ts
patnir: Done!

ironfish/src/migrations/data/030-value-to-unspent-note.ts
patnir: Done!

NullSoldier:
Testing the migration

```
Applying 1 migrations:
  030-value-to-unspent-note

Running 030-value-to-unspent-note...
Indexing all unspent notes and sorting them by value
 Indexed 9 unspent notes and sorted by value.
```

ironfish/src/wallet/account/account.ts
```diff
@@ -122,22 +122,30 @@ export class Account {
...
     const confirmations = options?.confirmations ?? 0
-
     const maxConfirmedSequence = Math.max(head.sequence - confirmations, GENESIS_BLOCK_SEQUENCE)
 
-    for await (const decryptedNote of this.walletDb.loadUnspentNotes(
+    for await (const unspentNoteHash of this.walletDb.loadValueToUnspentNoteHashes(
       this,
       assetId,
-      maxConfirmedSequence,
+      options?.reverse ?? false,
```
NullSoldier, line 138: why are you using `?? false`? It looks like `loadValueToUnspentNoteHashes` will default to false if it's undefined.

NullSoldier (approved):

NullSoldier:
I've combed over this PR with a fine comb and have nothing left to find. I've approved it with one last comment.

ironfish/src/wallet/account/account.ts
patnir: Made this change

PR #4529, opened 2024-01-10
feat: Add build method proposed transaction

jowparks:
## Summary
This allows us to take the necessary keys and go from a `ProposedTransaction` (raw transaction data), to an `UnsignedTransaction`, which will be used for frost multisig.

This introduces a bit of code duplication in the `post()` and `build()` code paths of `ProposedTransaction`, I suggest we move the `post` and `_partial_post` methods to `UnsignedTransaction` so that we can deduplicate the logic. Then on `ProposedTransaction`.
## Testing Plan
New test passes
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/transaction/mod.rs
+++ b/ironfish-rust/src/transaction/mod.rs
@@ -4,6 +4,7 @@
 
 use blstrs::Bls12;
 use ff::Field;
+use ironfish_frost::frost::keys::PublicKeyPackage;
 use outputs::OutputBuilder;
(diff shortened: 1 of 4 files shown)


hughy (approved):
this looks good to me

another possibility for reducing code duplication: pull the `change_notes` and the loops to build the unsigned descriptions out into separate methods and call those in `build`, `post`, and `_partial_post`

PR #4532, opened 2024-01-10
Rahul/ifl 2093 add frost signature to unsigned spends and mints

patnir:
## Summary

Add a method to assign the authorizing signature. This will be used in the signature aggregate method when we construct the transaction for multi signature wallets. 

## Testing Plan

Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/transaction/mints.rs
+++ b/ironfish-rust/src/transaction/mints.rs
@@ -663,7 +663,6 @@ mod test {
     fn test_add_signature() {
         let key = SaplingKey::generate_key();
         let public_address = key.public_address();
-        let sender_key = SaplingKey::generate_key();
 
         let asset = Asset::new(public_address, "name", "").expect("should be able to create asset");
         let public_key_randomness = jubjub::Fr::random(thread_rng());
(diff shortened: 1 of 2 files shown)


ironfish-rust/src/transaction/mints.rs
```diff
@@ -147,6 +147,11 @@ impl UnsignedMintDescription {
         Ok(self.description)
     }
 
+    pub fn add_signature(mut self, signature: Signature) -> Result<MintDescription, IronfishError> {
+        self.description.authorizing_signature = signature;
+        Ok(self.description)
+    }
+
```
jowparks, line 154: We don't need the result type, can just be a plain `MintDescription` return type

ironfish-rust/src/transaction/spends.rs
```diff
@@ -194,6 +194,14 @@ impl UnsignedSpendDescription {
         Ok(self.description)
     }
 
+    pub fn add_signature(
+        mut self,
+        signature: Signature,
+    ) -> Result<SpendDescription, IronfishError> {
```
jowparks, line 200: same here

ironfish-rust/src/transaction/mints.rs
patnir: Done

jowparks (approved):

PR #4537, opened 2024-01-12
Rahul/ifl 2101 proofgenerationkey serialization deserialization

patnir:
## Summary

Serialize + Deserialize Proof Generation Key.

## Testing Plan

Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-zkp/src/primitives/proof_generation_key.rs
+++ b/ironfish-zkp/src/primitives/proof_generation_key.rs
@@ -3,12 +3,12 @@ use jubjub::{Fr, SubgroupPoint};
 use zcash_primitives::sapling::ProofGenerationKey;
 
 pub trait ProofGenerationKeySerializable {
-    fn serialize(self) -> [u8; 64];
+    fn serialize(&self) -> [u8; 64];
     fn deserialize(bytes: [u8; 64]) -> Self;
 }
 
 impl ProofGenerationKeySerializable for ProofGenerationKey {
-    fn serialize(self) -> [u8; 64] {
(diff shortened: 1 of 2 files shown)


ironfish-zkp/src/primitives/proof_generation_key.rs
```diff
@@ -0,0 +1,153 @@
...
+            deserialized_proof_generation_key.ak
+        );
+        assert_eq!(
+            proof_generation_key.nsk,
+            deserialized_proof_generation_key.nsk
+        );
+    }
+
+    #[test]
+    fn test_proof_generation_key_serializable_deserializable() {
```
hughy, line 132: does this test cover anything different than the one above it?

hughy (commented):
small comment on possibly redundant test, but looks good otherwise!

ironfish-zkp/src/primitives/proof_generation_key.rs
jowparks: Looks like same test, you can remove one. Otherwise looks good.

jowparks (approved):
Note that we checked the underlying library, there is no serde for this type. Also, we decided to put it in the `ironfish-zkp` because both zkp/rust sub projects rely on this. Since we don't have a lib, and `ironfish-zkp` is imported into `ironfish-rust` seemed like the only option.

ironfish-zkp/src/primitives/proof_generation_key.rs
patnir: Removed!

PR #4538, opened 2024-01-12
adds optional multiSigKeys to AccountValue

hughy:
## Summary

multiSigKeys includes three fields needed for an account to participate in a FROST signing group: identifier, keyPackage, and proofGenerationKey

the identifier field is still to be determined as either a public key identifier for the participant, or a private key from which we can derive the public key identifier

the keyPackage is the key package generated for the participant during key generation and contains the participant's authorizing key shard

the proofGenerationKey is a key shared by the signing group that is needed for any participant to generate spend or mint proofs

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/walletdb/accountValue.test.ts
+++ b/ironfish/src/wallet/walletdb/accountValue.test.ts
@@ -39,7 +39,8 @@ describe('AccountValueEncoding', () => {
       incomingViewKey: key.incomingViewKey,
       outgoingViewKey: key.outgoingViewKey,
       publicAddress: key.publicAddress,
-      spendingKey: key.spendingKey,
+      // NOTE: accounts with multiSigKeys should not have spendingKey
+      spendingKey: null,
       viewKey: key.viewKey,
       version: 1,
       createdAt: null,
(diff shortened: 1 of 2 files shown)


ironfish/src/wallet/walletdb/accountValue.test.ts
```diff
@@ -22,6 +22,32 @@ describe('AccountValueEncoding', () => {
...
+    const encoder = new AccountValueEncoding()
+
+    const key = generateKey()
+    const value: AccountValue = {
+      id: 'id',
+      name: 'foobar👁️🏃🐟',
+      incomingViewKey: key.incomingViewKey,
+      outgoingViewKey: key.outgoingViewKey,
+      publicAddress: key.publicAddress,
+      spendingKey: key.spendingKey,
```
jowparks, line 42: I thought we decided we would have an additional field in `multiSigKeys` for `groupSpendingKey` and make this null? That way these accounts will be mostly treated as view only accounts.

hughy: this spending key should always be null if `multiSigKeys` is defined. i'll change that for the test and look into whether there's a straightforward way to formalize it

i think it's simpler to store the `proofGenerationKey` in the multisig keys instead of the group spending key since we'd only need that key to derive the proof generation key

jowparks: Oh yea that makes sense, and outgoing view key will be stored in the main account structure. SGTM

jowparks (approved):

PR #4539, opened 2024-01-13
refactor posting of proposed/unsigned transaction

jowparks:
## Summary
1. This refactor removes code duplication for `_partial_post` and `build` method of `ProposedTransaction`. 

The original code path was 
Standard transaction:
`ProposedTransaction -> Transaction`
Multisig
`ProposedTransaction -> UnsignedTransaction -> Transaction` 

Now the code path is the same for both.
`ProposedTransaction -> UnsignedTransaction -> Transaction` 

2. Splits `build/_partial_post` logic into two separate methods where `add_change_notes` is added. This allows us to only add change notes to the `post`ed transactions, and not the miners fee.

NOTE: The `post` method bound to `ProposedTransaction` remains the same, so any higher level code should remain unchanged
## Testing Plan
tests pass
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/transaction/mod.rs
+++ b/ironfish-rust/src/transaction/mod.rs
@@ -190,28 +190,19 @@ impl ProposedTransaction {
         Ok(())
     }
 
-    pub fn build(
+    pub fn add_change_notes(
         &mut self,
-        proof_generation_key: ProofGenerationKey,
-        view_key: ViewKey,
-        outgoing_view_key: OutgoingViewKey,
-        public_address: PublicAddress,
         change_goes_to: Option<PublicAddress>,
-        intended_transaction_fee: u64,
-    ) -> Result<UnsignedTransaction, IronfishError> {
-        // The public key after randomization has been applied. This is used
-        // during signature verification. Referred to as `rk` in the literature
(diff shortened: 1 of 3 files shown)


ironfish-rust/src/transaction/mod.rs
```diff
@@ -319,38 +326,19 @@ impl ProposedTransaction {
...
-            if change_amount > 0 {
-                let change_address = change_goes_to.unwrap_or_else(|| spender_key.public_address());
-                let change_note = Note::new(
-                    change_address,
-                    change_amount as u64, // we checked it was positive
-                    "",
-                    *asset_id,
-                    spender_key.public_address(),
-                );
+        let signed_fee = i64::try_from(intended_transaction_fee)?;
```
hughy, line 331: Multiple meanings of signed and unsigned!

I can't think of a name that I like better, but this could cause some confusion

hughy (approved):

ironfish-rust/src/transaction/mod.rs
jowparks: Yea thats fair, what about `i64_fee`? I will update

PR #4540, opened 2024-01-14
Load or generate networkIdentity if none is passed

dguenther:
## Summary

Currently, we manage the network identity in the `start` CLI command, which requires some code duplication for everyone instantiating the node (The CLI and the node app, at least). My goal with this PR was to bring the saving and loading into the SDK without changing any behavior when `generateNewIdentity` is used or an identity is passed in to `sdk.node()`. See below for the current vs new behaviors:

### Current Behavior

Saving and loading `networkIdentity` is delegated to the CLI start command.

* If no identity is set in internal.json, one will be generated when `sdk.node()` is called. It is then saved to `internal.json` in the `start` command.
* If an identity is set in internal.json, a new one will be generated when `sdk.node()` is called. Instead, the `start` command should load the identity and pass it in via `sdk.node({ privateIdentity })`.
* If an identity is set or unset in internal.json and `generateNewIdentity` config is true, a new identity will be generated when `sdk.node()` is called. It is then saved to `internal.json` in the `start` command.
* If an identity is set in internal.json and `sdk.node({ privateIdentity })` is called, the passed-in identity will be used. It is then saved to `internal.json` in the `start` command.
* If an identity is set in internal.json, `sdk.node({ privateIdentity })` is called, and `generateNewIdentity` config is `true`, a new identity will be generated and used. It is then saved to `internal.json` in the `start` command.

### New Behavior

* If no identity is set in internal.json, one will be generated **and saved** when `sdk.node()` is called.
* If an identity is set in internal.json, **it will be loaded** when `sdk.node()` is called.
* If an identity is set or unset in internal.json and `generateNewIdentity` config is `true`, a new identity will be generated **and saved** when `sdk.node()` is called.
* If an identity is set in internal.json and `sdk.node({ privateIdentity })` is called, the passed-in identity will be used **and saved** to `internal.json`.
* If an identity is set in internal.json, `sdk.node({ privateIdentity })` is called, and `generateNewIdentity` config is `true`, a new identity will be generated **and saved**.

## Testing Plan

* [x] Added yarn tests to verify networkIdentity is saved as expected.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[x] Yes
```

If a user was calling `sdk.node()` without passing in a private identity, previously it would generate a new identity on each call to `sdk.node()`, but now it loads the identity from `internal.json` unless overridden with `sdk.node({ privateIdentity })` or the `generateNewIdentity` config.

### DIFF
--- a/ironfish/src/node.ts
+++ b/ironfish/src/node.ts
@@ -205,8 +205,8 @@ export class FullNode {
     metrics?: MetricsMonitor
     files: FileSystem
     strategyClass: typeof Strategy | null
-    webSocket: IsomorphicWebSocketConstructor,
-    privateIdentity?: PrivateIdentity,
+    webSocket: IsomorphicWebSocketConstructor
+    privateIdentity?: PrivateIdentity
   }): Promise<FullNode> {
     logger = logger.withTag('ironfishnode')
     dataDir = dataDir || DEFAULT_DATA_DIR
@@ -249,7 +249,9 @@ export class FullNode {
       privateIdentity = new BoxKeyPair()
     } else if (!privateIdentity) {
       const internalNetworkIdentity = internal.get('networkIdentity')
-      privateIdentity = isHexSecretKey(internalNetworkIdentity) ? BoxKeyPair.fromHex(internalNetworkIdentity) : new BoxKeyPair()
+      privateIdentity = isHexSecretKey(internalNetworkIdentity)
+        ? BoxKeyPair.fromHex(internalNetworkIdentity)
+        : new BoxKeyPair()
     }
     internal.set('networkIdentity', privateIdentity.secretKey.toString('hex'))
     await internal.save()
--- a/ironfish/src/node.ts
+++ b/ironfish/src/node.ts
@@ -21,6 +21,7 @@ import { MetricsMonitor } from './metrics'
 import { Migrator } from './migrations'
 import { MiningManager } from './mining'
 import { PeerNetwork, PrivateIdentity, privateIdentityToIdentity } from './network'
+import { isHexSecretKey } from './network/identity'
 import { IsomorphicWebSocketConstructor, NodeDataChannelType } from './network/types'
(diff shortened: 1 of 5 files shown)


mat-if (approved):

ironfish/src/node.ts
```diff
@@ -246,6 +245,17 @@ export class FullNode {
       config.setOverride('bootstrapNodes', networkDefinition.bootstrapNodes)
     }
 
+    if (config.get('generateNewIdentity')) {
```
danield9tqh, line 248: nit suggestion you can either take it or leave it

```javascript
    const createNewIdentity =
      config.get('generateNewIdentity') || !isHexSecretKey(internal.get('networkIdentity'))

    if (createNewIdentity) {
      const newId = new BoxKeyPair()
      internal.set('networkIdentity', newId.secretKey.toString('hex'))
      await internal.save()
    }

    privateIdentity = BoxKeyPair.fromHex(internal.get('networkIdentity'))
```

danield9tqh: actually don't know the implications here of not using `save()` if we already have an identity (possibly overrides wouldn't get saved 🤷🏻 ). so could be safer just to keep it the same

danield9tqh (approved):

ironfish/src/node.ts
dguenther: It's a bit undefined whether passing in a private identity to a node should save to the internal.json or not, but I felt like it would be confusing if the identity in the node doesn't match the internal.json file (that file should basically be "per-node" anyway). So I think for now I'd like to save in all cases. But I'll see about pulling out the `internal.get` function

dguenther: Actually, I think the logic is okay as-is, so I'm going to merge. We want to use `privateIdentity` regardless of `internalIdentity` if it's set, unless generateNewIdentity is also set. Otherwise we fall back to internalIdentity (if it's valid).

PR #4547, opened 2024-01-17
adds optional multiSigKeys to Account type

hughy:
## Summary

includes multiSigKeys (identifier, keyPackage, proofGenerationKey) in Account constructor

throws an error if an account is created that includes both multiSigKeys and a non-null spendingKey

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/account/account.ts
+++ b/ironfish/src/wallet/account/account.ts
@@ -82,14 +82,6 @@ export class Account {
     this.walletDb = walletDb
     this.version = version ?? 1
     this.createdAt = createdAt
-
-    if (multiSigKeys !== undefined) {
-      Assert.isNull(
-        this.spendingKey,
-        'An account cannot have its own spending key if it is part of a signing group.',
-      )
-    }
-
     this.multiSigKeys = multiSigKeys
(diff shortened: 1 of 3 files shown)


ironfish/src/wallet/account/account.ts
```diff
@@ -76,6 +82,15 @@ export class Account {
     this.walletDb = walletDb
     this.version = version ?? 1
     this.createdAt = createdAt
+
+    if (multiSigKeys !== undefined) {
+      Assert.isNull(
+        this.spendingKey,
+        'An account cannot have its own spending key if it is part of a signing group.',
```
patnir, line 89: nit: shorten the error message "Multi sig wallet cannot have a spending key"?

patnir (approved):
2 nits. approved

ironfish/src/wallet/account/account.ts
hughy: I'm actually thinking that it may be better not to assert in the constructor and to handle this logic at import/account creation 🤔

PR #4549, opened 2024-01-17
Use createdAt as head of imported account

hairtail:
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -1544,7 +1544,7 @@ export class Wallet {
 
     await this.walletDb.db.transaction(async (tx) => {
       await this.walletDb.setAccount(account, tx)
-      await account.updateHead(null, tx)
+      await account.updateHead(createdAt, tx)
     })
 
     this.accounts.set(account.id, account)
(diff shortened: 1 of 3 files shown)


dguenther:
Let me take a look at this -- I was just running into something similar while testing imports on the Node App

ironfish/src/wallet/wallet.ts
```diff
@@ -1544,7 +1544,7 @@ export class Wallet {
 
     await this.walletDb.db.transaction(async (tx) => {
       await this.walletDb.setAccount(account, tx)
-      await account.updateHead(null, tx)
+      await account.updateHead(createdAt, tx)
```
hughy, line 1547: I think this is a good idea, but it will take some work to go through all of the existing tests and patch them for the new behavior and make sure that this doesn't introduce any regressions in edge cases

hughy: I spent some time debugging the test failures, and I think the issue is that we need to use the block _before_ `createdAt` here so that the the `createdAt` block is still scanned for the account.

For example, if an account receives tokens in the genesis block then it has a `createdAt` that refers to the genesis block. It will need to scan the genesis block in order to have a correct balance.

hairtail: Thanks for fixing tests!

dguenther (approved):
Tested this out a bit and doesn't seem like there are any functionality regressions, thanks for the improvement!

PR #4551, opened 2024-01-17
Rahul/ifl 2044 1 add rust method to split a sapling key into shares for FROST

patnir:
## Summary

Adding a method to split a sapling key. 
Defining the ShareSecretConfig

These two items are dependencies for splitting the spender key which will be next

## Testing Plan
Tests pass
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/transaction/mod.rs
+++ b/ironfish-rust/src/transaction/mod.rs
@@ -27,9 +27,7 @@ use rand::{
     rngs::{OsRng, ThreadRng},
     thread_rng,
 };
-use std::collections::BTreeMap;
 
-use ironfish_frost::frost::keys::SecretShare;
 use ironfish_frost::frost::{
     keys::{IdentifierList, KeyPackage, PublicKeyPackage},
     Identifier, SigningKey,
@@ -72,9 +70,6 @@ pub mod unsigned;
 mod utils;
 mod value_balances;
 mod version;
-
-#[cfg(test)]
-mod tests;
 
 pub use version::TransactionVersion;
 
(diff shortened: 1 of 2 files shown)


ironfish-rust/src/transaction/mod.rs
```diff
@@ -59,9 +71,6 @@ mod utils;
 mod value_balances;
 mod version;
```
jowparks, line 73: What was this line removed for?

ironfish-rust/src/transaction/mod.rs
```diff
@@ -935,3 +947,44 @@ pub fn batch_verify_transactions<'a>(
...
+        config.min_signers,
+        identifiers,
+        rng,
+    )?;
+
+    for (_k, v) in shares.clone() {
+        frost::keys::KeyPackage::try_from(v)?;
+    }
+
+    let mut key_packages: HashMap<_, _> = HashMap::new();
```
hughy, line 982: should this `HashMap` use the specific `Identifier, KeyPackage` type? does it make a difference?

jowparks (approved):

hughy (commented):
this looks good!

wondering if this should go into `keys/mod.rs` instead of `transaction/mod.rs`?

ironfish-rust/src/transaction/mod.rs
jowparks: Yea you can remove the `: HashMap<_, _>` and it will do type inference

PR #4554, opened 2024-01-17
Frost round one rust

jowparks:
## Summary
Introduces a wrapper method for `frost::round1::commit` that provides a `u64` as seed. This is a convenience wrapper that allows us to avoid importing the `rand` library into in our `ironfish-rust-nodejs` bindings library. 

## Testing Plan
tests pass
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/frost_utils/round_one.rs
+++ b/ironfish-rust/src/frost_utils/round_one.rs
@@ -1,0 +1,56 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+use ironfish_frost::frost::{
+    self,
+    keys::KeyPackage,
+    round1::{SigningCommitments, SigningNonces},
+};
+use rand::{rngs::StdRng, SeedableRng};
+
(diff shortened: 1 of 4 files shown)


ironfish-rust/src/frost_utils/round_one.rs
```diff
@@ -0,0 +1,56 @@
...
+        )
+        .expect("key shares to be created");
+        let key_package = key_packages
+            .0
+            .into_iter()
+            .next()
+            .expect("key package to be created")
+            .1;
+        let (nonces, commitments) = super::round_one(&key_package, seed);
+        let (nonces2, commitments2) = super::round_one(&key_package, seed);
```
jowparks, line 51: For whatever reason, the comparison method is not implemented for nonces, but it is for commitments 🤷

ironfish-rust/src/lib.rs
```diff
@@ -6,6 +6,7 @@ use blstrs::Bls12;
 
 pub mod assets;
 pub mod errors;
```
jowparks, line 8: Would have liked to call this `frost`, but there is a namespace clash with line 26 below.
```
pub use {
    ironfish_frost::frost,
```

patnir (approved):

PR #4555, opened 2024-01-18
Rahul/ifl 2044 split spender key

patnir:
## Summary

Creates the trusted dealer key package when giving a list of participant identifiers and a coordinator key

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/errors.rs
+++ b/ironfish-rust/src/errors.rs
@@ -129,3 +129,9 @@ impl From<num::TryFromIntError> for IronfishError {
         IronfishError::new_with_source(IronfishErrorKind::TryFromInt, e)
     }
 }
+
+impl From<ironfish_frost::frost::Error> for IronfishError {
+    fn from(e: ironfish_frost::frost::Error) -> IronfishError {
+        IronfishError::new_with_source(IronfishErrorKind::Frost, e)
+    }
+}
--- a/ironfish-rust/src/errors.rs
+++ b/ironfish-rust/src/errors.rs
@@ -26,7 +26,7 @@ pub struct IronfishError {
 pub enum IronfishErrorKind {
     BellpersonSynthesis,
     CryptoBox,
-    Frost,
+    FrostLibError,
     IllegalValue,
     InconsistentWitness,
     InvalidAssetIdentifier,
@@ -132,6 +132,6 @@ impl From<num::TryFromIntError> for IronfishError {
 
 impl From<ironfish_frost::frost::Error> for IronfishError {
     fn from(e: ironfish_frost::frost::Error) -> IronfishError {
-        IronfishError::new_with_source(IronfishErrorKind::Frost, e)
+        IronfishError::new_with_source(IronfishErrorKind::FrostLibError, e)
     }
 }
--- a/ironfish-rust/src/errors.rs
+++ b/ironfish-rust/src/errors.rs
@@ -22,10 +22,11 @@ pub struct IronfishError {
 /// in the code to reduce the cognitive load needed for using Result and Error
 /// types. The second is to give a singular type to convert into NAPI errors to
 /// be raised on the Javascript side.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub enum IronfishErrorKind {
     BellpersonSynthesis,
     CryptoBox,
+    Frost,
     IllegalValue,
     InconsistentWitness,
     InvalidAssetIdentifier,
--- a/ironfish-rust/src/errors.rs
+++ b/ironfish-rust/src/errors.rs
@@ -47,6 +47,7 @@ pub enum IronfishErrorKind {
     InvalidOutputProof,
     InvalidPaymentAddress,
     InvalidPublicAddress,
+    InvalidSecret,
     InvalidSignature,
     InvalidSigningKey,
     InvalidSpendProof,
--- a/ironfish-rust/src/frost_utils/split_secret.rs
+++ b/ironfish-rust/src/frost_utils/split_secret.rs
@@ -79,6 +79,11 @@ mod test {
         let vec = vec![1; 31];
         let array = vec_to_array(vec);
         assert!(array.is_err());
+        // assert error type
+        assert!(
+            matches!(array.unwrap_err().kind, IronfishErrorKind::InvalidSecret),
+            "expected InvalidSecret error"
+        );
     }
 
     #[test]
--- a/ironfish-rust/src/frost_utils/split_secret.rs
+++ b/ironfish-rust/src/frost_utils/split_secret.rs
@@ -19,7 +19,7 @@ pub struct SecretShareConfig {
     pub secret: Vec<u8>,
 }
 
-pub fn split_secret(
+pub(crate) fn split_secret(
     config: &SecretShareConfig,
     identifiers: IdentifierList,
     rng: &mut ThreadRng,
--- a/ironfish-rust/src/frost_utils/split_secret.rs
+++ b/ironfish-rust/src/frost_utils/split_secret.rs
@@ -18,24 +18,16 @@ pub struct SecretShareConfig {
     pub secret: Vec<u8>,
 }
 
-fn vec_to_array(vec: Vec<u8>) -> Result<[u8; 32], IronfishError> {
-    if vec.len() != 32 {
-        return Err(IronfishError::new(IronfishErrorKind::InvalidSecret));
-    }
-
-    let array: [u8; 32] = vec
-        .try_into()
-        .map_err(|_| IronfishError::new(IronfishErrorKind::InvalidSecret))?;
-
-    Ok(array)
-}
-
 pub(crate) fn split_secret(
     config: &SecretShareConfig,
     identifiers: IdentifierList,
     rng: &mut ThreadRng,
 ) -> Result<(HashMap<Identifier, KeyPackage>, PublicKeyPackage), IronfishError> {
-    let secret_bytes = vec_to_array(config.secret.clone())?;
+    let secret_bytes: [u8; 32] = config
+        .secret
+        .clone()
+        .try_into()
+        .map_err(|_| IronfishError::new(IronfishErrorKind::InvalidSecret))?;
 
     let secret_key = SigningKey::deserialize(secret_bytes)?;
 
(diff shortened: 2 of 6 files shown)


ironfish-rust/src/frost_utils/split_spender_key.rs
```diff
@@ -0,0 +1,105 @@
...
+use jubjub::SubgroupPoint;
+use rand::thread_rng;
+use std::collections::HashMap;
+
+use crate::{IncomingViewKey, OutgoingViewKey, PublicAddress, SaplingKey, ViewKey};
+
+use super::split_secret::{split_secret, SecretShareConfig};
+
+pub struct SpenderKeyConfig {
+    pub ak: [u8; 32],
```
jowparks, line 21: We could probably create a type alias for this authorizing key.

ironfish-rust/src/frost_utils/split_spender_key.rs
```diff
@@ -0,0 +1,105 @@
...
+) -> SpenderKeyConfig {
+    let secret_config = SecretShareConfig {
+        min_signers,
+        max_signers,
+        secret,
+    };
+
+    let mut rng = thread_rng();
+    let (key_packages, pubkeys) =
+        split_secret(&secret_config, IdentifierList::Default, &mut rng).unwrap();
```
jowparks, line 45: We are not going to be using the default identifier list, we are going to use Andreas implementation of Identifiers from `ironfish-frost`

ironfish-rust/src/frost_utils/split_spender_key.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    let nullifier_deriving_key = *PROOF_GENERATION_KEY_GENERATOR
+        * coordinator_sapling_key.sapling_proof_generation_key().nsk;
+
+    let view_key = ViewKey {
+        authorizing_key,
+        nullifier_deriving_key,
+    };
+
+    let incoming_viewing_key = IncomingViewKey {
```
jowparks, line 64: you don't need to recreate this, it exists on the `SaplingKey`

ironfish-rust/src/frost_utils/split_spender_key.rs
```diff
@@ -0,0 +1,105 @@
...
+        max_signers,
+        secret,
+    };
+
+    let mut rng = thread_rng();
+    let (key_packages, pubkeys) =
+        split_secret(&secret_config, IdentifierList::Default, &mut rng).unwrap();
+
+    let authorizing_key_bytes = pubkeys.verifying_key().serialize();
+    let authorizing_key = Option::from(SubgroupPoint::from_bytes(&authorizing_key_bytes))
```
jowparks, line 48: Why option? You should error if can't be generated.

ironfish-rust/src/frost_utils/split_spender_key.rs
```diff
@@ -0,0 +1,105 @@
...
+        .expect("should be able to deserialize the verifying key into a SubgroupPoint");
```
jowparks, line 49: expect should only be used in tests.

ironfish-rust/src/frost_utils/split_spender_key.rs
```diff
@@ -0,0 +1,105 @@
...
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_split_spender_key() {
+        let key = SaplingKey::generate_key();
+        let secret = key.spend_authorizing_key.to_bytes().to_vec();
+
```
jowparks, line 102: You could just do some basic assertions here for ak, vk, ivk, against the input sapling key. Then you can also check the public_address from the pubkeys output and verify it matches the input sapling key address, I think that test should be sufficient.

ironfish-rust/src/keys/mod.rs
```diff
@@ -241,7 +241,7 @@ impl SaplingKey {
     ///
     /// This method is only called once, but it's kind of messy, so I pulled it
     /// out of the constructor for easier maintenance.
-    fn hash_viewing_key(
+    pub fn hash_viewing_key(
```
jowparks, line 244: You will be able to remove this `pub` once you use the method I mentioned above for retrieveing ivk.

ironfish-rust/src/frost_utils/split_spender_key.rs
```diff
@@ -0,0 +1,175 @@
...
+                    .to_identity()
+                    .to_frost_identifier(),
+            );
+        }
+
+        let sapling_key = SaplingKey::generate_key();
+
+        let result = std::panic::catch_unwind(|| {
+            split_spender_key(sapling_key, 5, 11, identifiers.clone());
+        });
```
patnir, line 115: the fact that this is how to catch the error makes me think that I should right the function differently with a Result<> return type?

ironfish-rust/src/frost_utils/split_secret.rs
```diff
@@ -21,7 +23,7 @@ pub fn split_secret(
     config: &SecretShareConfig,
     identifiers: IdentifierList,
     rng: &mut ThreadRng,
-) -> Result<(HashMap<Identifier, KeyPackage>, PublicKeyPackage), Error> {
+) -> Result<(HashMap<Identifier, KeyPackage>, PublicKeyPackage), IronfishError> {
     let secret_key = SigningKey::deserialize(
         config
             .secret
```
jowparks, line 29: Down on line 50 I missed an `unwrap`, that should be a `.map_err` to an `IronfishError`. Can you make that change too?

jowparks: Line 32 you should also convert to an `IronfishError`

jowparks: Also line 44

ironfish-rust/src/frost_utils/split_secret.rs
```diff
@@ -21,7 +23,7 @@ pub fn split_secret(
     config: &SecretShareConfig,
```
jowparks, line 23: can we change this to `pub(crate)` instead of `pub`? No one should call this outside of `ironfish_rust` right?

ironfish-rust/src/frost_utils/split_spender_key.rs
```diff
@@ -0,0 +1,181 @@
...
+        let err = result.err().unwrap();
+        assert_eq!(err.kind, IronfishErrorKind::Frost);
+        assert!(err.to_string().contains("Incorrect number of identifiers."));
+
+        let sapling_key2 = SaplingKey::generate_key();
+
+        let result = split_spender_key(sapling_key2, 5, 9, identifiers.clone());
+
+        assert!(result.is_err());
+        let err = result.err().unwrap();
```
jowparks, line 128: Another way of doing this:
https://github.com/iron-fish/ironfish/blob/4f44aae6d07c702c6664bc42adaadcbca763cf81/ironfish-rust/src/keys/proof_generation_key.rs#L100

ironfish-rust/src/errors.rs
```diff
@@ -22,10 +22,11 @@ pub struct IronfishError {
 /// in the code to reduce the cognitive load needed for using Result and Error
 /// types. The second is to give a singular type to convert into NAPI errors to
 /// be raised on the Javascript side.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub enum IronfishErrorKind {
     BellpersonSynthesis,
     CryptoBox,
+    Frost,
```
jowparks, line 29: Maybe call this `FrostCoreError` or `FrostLibError`, Frost is kind of generic.

ironfish-rust/src/errors.rs
```diff
@@ -128,3 +129,9 @@ impl From<num::TryFromIntError> for IronfishError {
         IronfishError::new_with_source(IronfishErrorKind::TryFromInt, e)
     }
 }
+
```
jowparks, line 132: Nice!

ironfish-rust/src/frost_utils/split_spender_key.rs
```diff
@@ -0,0 +1,181 @@
...
+
+        let result = split_spender_key(sapling_key, 5, 11, identifiers.clone());
+        assert!(result.is_err());
+        let err = result.err().unwrap();
+        assert_eq!(err.kind, IronfishErrorKind::Frost);
+        assert!(err.to_string().contains("Incorrect number of identifiers."));
+
+        let sapling_key2 = SaplingKey::generate_key();
+
+        let result = split_spender_key(sapling_key2, 5, 9, identifiers.clone());
```
jowparks, line 125: Why do you have the sapling_key2 part of the test? Can't you remove this isn't it basically the same as the above test component?

ironfish-rust/src/frost_utils/split_spender_key.rs
```diff
@@ -0,0 +1,181 @@
...
+
+        assert!(result.is_err());
+        let err = result.err().unwrap();
+        assert_eq!(err.kind, IronfishErrorKind::Frost);
+        assert!(err.to_string().contains("Incorrect number of identifiers."));
+    }
+
+    #[test]
+    fn test_split_spender_key() {
```
jowparks, line 134:
```suggestion
    fn test_split_spender_key_success() {
```

ironfish-rust/src/frost_utils/split_spender_key.rs
patnir: i was checking less than + greater than cases. it gives me the following error when i use the same sapling key twice: `use of moved value: `sapling_key`` should i remove the second case?

ironfish-rust/src/frost_utils/split_secret.rs
patnir: Done

ironfish-rust/src/errors.rs
patnir: ty!

ironfish-rust/src/frost_utils/split_spender_key.rs
patnir: Done

ironfish-rust/src/frost_utils/split_secret.rs
```diff
@@ -2,35 +2,44 @@
...
+    if vec.len() != 32 {
+        return Err(IronfishError::new(IronfishErrorKind::InvalidSecret));
+    }
+
+    let array: [u8; 32] = vec
+        .try_into()
+        .map_err(|_| IronfishError::new(IronfishErrorKind::InvalidSecret))?;
+
+    Ok(array)
+}
```
jowparks, line 31: If you provide a type parameter to the method and move to serializing this would be a generally useful method. I would return a generic serialization error from this method and then make it more refined where you call it in `split_key`

see this method for an example of the generic.
https://github.com/iron-fish/ironfish/blob/99763bbdc0238efff677fc11776c00d3e77f2d49/ironfish-rust/src/serializing/mod.rs#L50-L64

jowparks: As a matter of fact, this might already exist. In the code that you deleted on line 29, I think there may be an attempt to coerce the `vec` into an `array`, so this method won't be needed. You just will need to map the error to an ironfish error.

ironfish-rust/src/frost_utils/split_spender_key.rs
jowparks: Oh I see no this is fine. I just missed the fact that you were checking greater vs less than, just add some code comments

ironfish-rust/src/frost_utils/split_spender_key.rs
patnir: why `matches!` this better than `assert_eq!(err.kind, IronfishErrorKind::FrostLibError);`

ironfish-rust/src/frost_utils/split_secret.rs
patnir: Done!

ironfish-rust/src/frost_utils/split_spender_key.rs
jowparks: It's not, you just need partial equal to be able to use assert equality.

jowparks (approved):

PR #4561, opened 2024-01-18
Add config value for building the full FishHash context

danield9tqh:
## Summary
For power users like mining pools they may want to enable the full 4GB FishHash cache to increase performance of verifying block headers. This gives that option as a config value

## Testing Plan
Unit tests + running node locally

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/node.ts
+++ b/ironfish/src/node.ts
@@ -250,12 +250,16 @@ export class FullNode {
     }
 
     const consensus = new TestnetConsensus(networkDefinition.consensus)
-    // TODO: config value for fullContext
-    const blockHasher = new BlockHasher({
-      consensus,
-      context: fishHashContext,
-      fullContext: false,
-    })
+
+    if (consensus.isNeverActive('enableFishHash')) {
+      fishHashContext = undefined
+    } else if (!fishHashContext) {
+      const isFull = config.get('fishHashFullContext')
+      fishHashContext = new FishHashContext(isFull)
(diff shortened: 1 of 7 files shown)


ironfish/src/node.ts
```diff
@@ -250,15 +249,17 @@ export class FullNode {
...
-      context: fishHashContext,
-      fullContext: false,
-    })
+
+    if (consensus.isNeverActive('enableFishHash')) {
+      fishHashContext = undefined
+    } else if (!fishHashContext) {
+      const isFull = config.get('fishHashFullContext')
+      fishHashContext = new FishHashContext(isFull)
+      fishHashContext.prebuildDataset(numWorkers)
```
mat-if, line 258: I don't think this should be here yet - not until we have data on whether this actually provides real world benefits. My suspicion is that it won't, and the drastic startup time required to prebuild is not worth the headache of trying to communicate to users why their node takes 10+ minutes to start. It also makes it difficult to test/develop, as we now have to modify code to not have a 10 minute iteration loop when testing here

danield9tqh: `prebuildDataset` is a noop if its not a full context. And it's not a full context by default so this won't do anything unless users explicitly set their config value to true

danield9tqh: Talked about this, the node does not verify blocks often enough to warrant pre-building the cache at startup

mat-if (approved):

PR #4563, opened 2024-01-18
Unsigned transaction napi

jowparks:
## Summary
Adds binding for `UnsignedTransaction`. 
This intermediary form of transaction contains unsigned spends/mints and will likely be used mostly for FROST multisig transactions.

Currently only supports ser/de but will eventually support signing.

## Testing Plan
Tests pass
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-rust-nodejs/src/structs/unsigned.rs
```diff
@@ -0,0 +1,38 @@
...
+use napi_derive::napi;
+
+use ironfish::transaction::unsigned::UnsignedTransaction as UnsignedTransactionRust;
+
+use crate::to_napi_err;
+
+#[napi]
+pub struct UnsignedTransaction {
+    transaction: UnsignedTransactionRust,
+}
```
hughy, line 18: nit - we use `Native` elsewhere (`NativeProposedTransaction`, `NativeTransaction`), should we do the same here?

ironfish-rust-nodejs/src/structs/mod.rs
```diff
@@ -22,3 +22,6 @@ pub use witness::*;
 
 mod key_packages;
 pub use key_packages::*;
+
+mod unsigned;
+pub use unsigned::*;
```
hughy, line 27: should we define a new module for UnsignedTransaction or use the existing transaction module?

ironfish-rust-nodejs/src/structs/unsigned.rs
jowparks: We do use Native, but then we remap to `Transaction` via JS. 
```
#[napi(js_name = "Transaction")]
pub struct NativeTransaction {
    transaction: ProposedTransaction,
}
```
My guess is that this was originally done because whoever wrote it didn't think to use an alias for the import.
IMO this is better from searchability perspective, open to converting to other mechanism though

hughy (approved):

ironfish-rust-nodejs/src/structs/mod.rs
jowparks: Yea probably should since that is the current status quo, although I would like to refactor to make them submodules

ironfish-rust-nodejs/src/structs/unsigned.rs
jowparks: Talked with Hugh, I renamed back to our original semantics, we can change all these if we want to with a future refactor.

PR #4564, opened 2024-01-18
Output proof generation key

jowparks:
## Summary
Add `proof_generation_key` to output for sapling key in napi bindings. 

I needed this to write a test for `UnsignedTransaction`, but I am guessing we will need this when creating a new account for multisig.

Note that there are changes in here for the NAPI bindings from previous PR that weren't picked up. I will add a build check in a subsequent PR to make sure they are present.
## Testing Plan
N/A
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-rust-nodejs/index.d.ts
```diff
@@ -3,6 +3,13 @@
 
 /* auto-generated by NAPI-RS */
```
jowparks, line 5: Missing from a previous PR.

hughy (approved):

PR #4568, opened 2024-01-19
Remove customNetwork config value

danield9tqh:
## Summary
The `customNetwork` config value was originally added to be able to pass in a path to a custom network definition when first initializing a node in a data directory. After the data directory has been initialized the `customNetwork` config value is useless because the network cannot be changed.

Passing along the `customNetworkPath` string accomplishes this same goal of being able to use a custom JSON file and removes the unecessary config option.

## Testing Plan
Started up a new node with the customNetwork flag on startup pointing to a JSON file

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[X] Yes
```
Removes the `customNetwork` value of config. This is technically a breaking change but I can't see how it would actually break anyone's builds since the config value can only be used at initial startup

### DIFF
--- a/ironfish-cli/src/commands/start.ts
+++ b/ironfish-cli/src/commands/start.ts
@@ -204,16 +204,13 @@ export default class Start extends IronfishCommand {
     if (networkId !== undefined && networkId !== this.sdk.config.get('networkId')) {
       this.sdk.config.setOverride('networkId', networkId)
     }
-    if (customNetwork !== undefined && customNetwork !== this.sdk.config.get('customNetwork')) {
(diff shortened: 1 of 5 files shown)


NullSoldier:
I think this looks fine to me, but we should also get Derek to review this as well.

I tested this manually by making a new custom network definition and passing the path in via the CLI.

Everything still seems to work as long as were ok removing the ability to specify this by the config.

dguenther:
I'll take a look at this on Monday, ran out of time for review today.

NullSoldier (approved):

dguenther (approved):

PR #4571, opened 2024-01-19
Add RpcHttpClient

NullSoldier:
## Summary

This adds a new RpcHttpClient to the SDK which allows you to make HTTP requests against the RpcHttpAdapter.

**HttpAdapter Bugs**
- Crash if you send a chunk and then send a response because headers try to send twice
- Crash if you send an RpcValidationError after a chunk was sent because headers try to send twice
- Fixed not sending delimter before rendering an error to the client which would cause client to crash when parsing 2 messages as 1 JSON payload.

## Testing Plan

Still need to add tests and check on postman that some of the requests still work.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/adapters/httpAdapter.test.ts
+++ b/ironfish/src/rpc/adapters/httpAdapter.test.ts
@@ -1,0 +1,117 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+/* eslint-disable jest/no-conditional-expect */
+import Mitm from 'mitm'
+import * as yup from 'yup'
+import { Assert } from '../../assert'
+import { createNodeTest } from '../../testUtilities'
+import { RpcHttpClient, RpcRequestError } from '../clients'
+import { ALL_API_NAMESPACES } from '../routes'
+import { RPC_ERROR_CODES, RpcValidationError } from './errors'
+import { RpcHttpAdapter } from './httpAdapter'
+
+describe('HttpAdapter', () => {
+  let httpAdapter: RpcHttpAdapter | undefined
+  let mitm: ReturnType<typeof Mitm>
+
+  const nodeTest = createNodeTest(false, {
+    config: {
+      enableRpc: false,
+      enableRpcIpc: false,
+      enableRpcTcp: false,
+      enableRpcTls: false,
+      rpcHttpPort: 0,
+    },
+  })
+
+  beforeEach(async () => {
+    httpAdapter = new RpcHttpAdapter('localhost', 0, undefined, ALL_API_NAMESPACES)
+
+    mitm = Mitm()
+    mitm.on('request', (req, res) => httpAdapter?.onRequest(req, res))
+
+    await nodeTest.node.rpc.mount(httpAdapter)
+  }, 20000)
+
+  afterEach(() => {
+    mitm.disable()
+  })
+
+  it('should send and receive message', async () => {
+    Assert.isNotUndefined(httpAdapter)
+    Assert.isNotNull(httpAdapter.router)
+
+    httpAdapter.router.routes.register('foo/bar', yup.string(), (request) => {
+      request.end(request.data)
+    })
+
+    const client = new RpcHttpClient('http://localhost')
+
+    const response = await client.request('foo/bar', 'hello world').waitForEnd()
+    expect(response.content).toBe('hello world')
+  }, 20000)
+
+  it('should stream message', async () => {
+    Assert.isNotUndefined(httpAdapter)
+    Assert.isNotNull(httpAdapter?.router)
+
+    httpAdapter.router.routes.register('foo/bar', yup.object({}), (request) => {
+      request.stream('hello 1')
+      request.stream('hello 2')
+      request.end()
+    })
+
+    const client = new RpcHttpClient('http://localhost')
+
+    const response = client.request('foo/bar')
+    expect((await response.contentStream().next()).value).toBe('hello 1')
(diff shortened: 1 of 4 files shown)


ironfish/src/rpc/adapters/httpAdapter.ts
```diff
@@ -132,6 +117,24 @@ export class RpcHttpAdapter implements IRpcAdapter {
...
+    })
+
+    this.requests.set(requestId, { req, waitForClose })
+
+    // All response bodies should be application/json
+    res.setHeader('Content-Type', 'application/json')
+
+    void this.handleRequest(req, res, requestId)
+  }
+
```
NullSoldier, line 137: Moved this to its own function so we can call it in tests

ironfish/src/rpc/adapters/httpAdapter.ts
```diff
@@ -164,86 +167,123 @@ export class RpcHttpAdapter implements IRpcAdapter {
...
-
-    if (request.url === undefined) {
-      throw new RpcResponseError('No request url provided')
-    }
+    let chunkStreamed = false
 
-    this.logger.debug(
-      `Call HTTP RPC: ${request.method || 'undefined'} ${request.url || 'undefined'}`,
-    )
+    try {
```
NullSoldier, line 172: Wrapped this in try/catch from a .catch() on the outside so when rendering an error we can check chunkStreamed.

ironfish/src/rpc/adapters/httpAdapter.ts
```diff
@@ -164,86 +167,123 @@ export class RpcHttpAdapter implements IRpcAdapter {
...
 
-    const currRequest = this.requests.get(requestId)
-    currRequest && this.requests.set(requestId, { ...currRequest, rpcRequest })
+      const responseData = JSON.stringify(responseMessage)
+      const responseSize = Buffer.byteLength(responseData, 'utf-8')
+      this.outboundTraffic.add(responseSize)
+      this.outboundBytes.value += responseSize
 
-    await router.route(route, rpcRequest)
+      response.end(delimeter + responseData)
```
NullSoldier, line 284: Fixed not sending a delimeter when sending an error if a chunk has been streamed which would crash the client.

jowparks: nit: delimiter, not delimeter

ironfish/src/rpc/adapters/httpAdapter.ts
```diff
@@ -164,86 +167,123 @@ export class RpcHttpAdapter implements IRpcAdapter {
...
+        data.push(chunk)
 
-    let chunkStreamed = false
-    const rpcRequest = new RpcRequest(
-      body === undefined ? undefined : JSON.parse(body),
-      route,
-      (status: number, data?: unknown) => {
-        response.statusCode = status
-        const delimeter = chunkStreamed ? MESSAGE_DELIMITER : ''
+        if (size >= MAX_REQUEST_SIZE) {
```
jowparks, line 201: Should this be configurable by runner? Wondering why we put arbitrary 5MB limit on this.

ironfish/src/rpc/adapters/httpAdapter.test.ts
```diff
@@ -0,0 +1,117 @@
...
+    const client = new RpcHttpClient('http://localhost')
+
+    const response = client.request('foo/bar', body)
+
+    await expect(response.waitForEnd()).rejects.toThrow(RpcRequestError)
+    await expect(response.waitForEnd()).rejects.toMatchObject({
+      status: 400,
+      code: ERROR_CODES.VALIDATION,
+      codeMessage: expect.stringContaining('this must be defined'),
+    })
```
jowparks, line 115: Side quest: 
Add test case for timeout.
Add test case for middle of stream failure.

jowparks (approved):

ironfish/src/rpc/adapters/httpAdapter.ts
NullSoldier: https://github.com/iron-fish/ironfish/pull/4579

NullSoldier: Fixed

ironfish/src/rpc/adapters/httpAdapter.test.ts
```diff
@@ -0,0 +1,117 @@
...
+import Mitm from 'mitm'
+import * as yup from 'yup'
+import { Assert } from '../../assert'
+import { createNodeTest } from '../../testUtilities'
+import { RpcHttpClient, RpcRequestError } from '../clients'
+import { ALL_API_NAMESPACES } from '../routes'
+import { ERROR_CODES, RpcValidationError } from './errors'
+import { RpcHttpAdapter } from './httpAdapter'
+
+describe('TcpAdapter', () => {
```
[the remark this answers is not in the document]
NullSoldier: Fixed

ironfish/src/rpc/adapters/httpAdapter.test.ts
NullSoldier: I have a good idea. This file was copy pasted from TlsAdapter tests which also don't have those tests. However, all the RPC clients inherit from `RpcClient` base class. We could just make one set of tests that test every adapter and client using a normalized test suite. I can do that in another PR afterwards.

If I write those tests here, they'll only cover the HTTP adapter.

jowparks: SGTM!

ironfish/src/rpc/adapters/httpAdapter.ts
NullSoldier: This is already in the code from the existing HTTP adapter.

PR #4576, opened 2024-01-20
Query the full consensus parameters in the pool and solo miners

danield9tqh:
## Summary
Add the consensus parameters to pool and solo miner to be able to switch at the FishHash activation sequence.

## Testing Plan
Unit test + running pool miner / solo miner on devnet

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/mining/pool.ts
+++ b/ironfish/src/mining/pool.ts
@@ -354,7 +354,7 @@ export class MiningPool {
       return
     }
 
-    // Get block explorer URLs based on netowrk ID
+    // Get block explorer URLs based on network ID
     const networkResponse = await this.rpc.chain.getNetworkInfo()
     const explorer = this.getExplorer(networkResponse.content.networkId)
     this.webhooks.map((w) => w.poolConnected(explorer ?? undefined))
--- a/ironfish/src/mining/pool.ts
+++ b/ironfish/src/mining/pool.ts
@@ -5,6 +5,7 @@ import { blake3 } from '@napi-rs/blake-hash'
 import LeastRecentlyUsed from 'blru'
 import tls from 'tls'
 import { Assert } from '../assert'
+import { ConsensusParameters } from '../consensus'
 import { Config } from '../fileStores/config'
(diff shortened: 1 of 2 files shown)


ironfish/src/serde/BlockTemplateSerde.ts
```diff
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-import { Block } from '../primitives/block'
+import { Block, RawBlock } from '../primitives/block'
```
mat-if, line 5: Are these changes intended to be in this PR?

ironfish/src/rpc/clients/client.ts
```diff
@@ -733,6 +734,24 @@ export abstract class RpcClient {
       ).waitForEnd()
     },
 
+    getParsedConsensusParameters: (
```
mat-if, line 737: I'm not sure this is the best approach

1. This is duplicating code of the `getConsensusParameters`
2. `number | null` is a better type, especially for consumers of the SDK (modifying ConsensusParameters itself is a topic for another day..)
3. This is a new pattern that deviates from the expectations of the existing pattern - RPC client functions return RPC responses - leading to an inconsistent API for consumers of the SDK

If we do need this, my initial feeling is that this should be a util function

danield9tqh: Ok yea I can just duplicate it in the 2 places we call it. I was back and forth on this

ironfish/src/serde/BlockTemplateSerde.ts
danield9tqh: They are. The hashing code deals with `RawBlock` and `RawBlock` header a lot so its very helpful to be able to go from `SerializedBlockTemplate` to `RawBlock` and back. I can do it in another PR though

mat-if (approved):

PR #4592, opened 2024-01-23
Rahul/ifl 2014 create napi function to split a secret

patnir:
## Summary

Napi binding to split spender key

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust-nodejs/Cargo.toml
+++ b/ironfish-rust-nodejs/Cargo.toml
@@ -30,7 +30,6 @@ base64 = "0.13.0"
 fish_hash = "0.1.0"
 ironfish = { path = "../ironfish-rust" }
 ironfish_mpc = { path = "../ironfish-mpc" }
-ironfish-frost = { git = "https://github.com/iron-fish/ironfish-frost.git" }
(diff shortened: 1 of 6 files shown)


ironfish-rust-nodejs/Cargo.toml
```diff
@@ -30,6 +30,7 @@ base64 = "0.13.0"
 fish_hash = "0.1.0"
 ironfish = { path = "../ironfish-rust" }
 ironfish_mpc = { path = "../ironfish-mpc" }
+ironfish-frost = { git = "https://github.com/iron-fish/ironfish-frost.git" }
```
hughy, line 33: do we need this dependency?

patnir: I use `use ironfish::frost::Identifier;`

patnir: I tried removing it and it works without it. I'll push the change. I could have sworn that i wasn't able to import it

hughy (approved):

PR #4599, opened 2024-01-24
adds frost participant napi bindings

hughy:
## Summary

implements ParticipantSecret and ParticipantIdentity to wrap the Secret and Identity types for participants defined in ironfish-frost

uses 'random' as ParticipantSecret constructor because we haven't implemented ser/de for Secret

serializes frost Identifier as a hex string rather than defining a napi type for Identifier

## Testing Plan

- adds unit test

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


patnir (approved):

ironfish-rust-nodejs/src/frost.rs
```diff
@@ -52,3 +54,60 @@ pub fn round_two(
...
 }
+
+#[napi]
+pub struct ParticipantSecret {
+    secret: Secret,
+}
+
+#[napi]
+impl ParticipantSecret {
+    // TODO(hughy): implement Secret ser/de
```
jowparks, line 65: new + serialize methods?

ironfish-rust-nodejs/tests/frost.test.slow.ts
```diff
@@ -0,0 +1,21 @@
...
+      const identity = secret.toIdentity()
+
+      const serialized = identity.serialize()
+
+      const deserialized = new ParticipantIdentity(serialized)
+
+      expect(identity).toEqual(deserialized)
+    })
+  })
+})
```
jowparks, line 21: nl

ironfish-rust-nodejs/src/frost.rs
```diff
@@ -52,3 +54,60 @@ pub fn round_two(
...
+
+        Ok(ParticipantIdentity { identity })
+    }
+}
+
+#[napi]
+pub struct ParticipantIdentity {
+    identity: Identity,
+}
+
```
jowparks, line 85: Will we need an `Identifer` object at all in these bindings? What about when we are doing encryption/validation via the public keys (identifier). Maybe we don't, we also can probably just wait until we need it.

ironfish-rust-nodejs/src/frost.rs
hughy: The `Secret` type in `ironfish-frost` doesn't implement ser/de yet, so new and serialize aren't yet possible

ironfish-rust-nodejs/src/frost.rs
```diff
@@ -52,3 +54,60 @@ pub fn round_two(
 
     Ok(bytes_to_hex(&signature_share.serialize()))
 }
+
+#[napi]
+pub struct ParticipantSecret {
```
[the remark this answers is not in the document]
hughy: I'm open to renaming, but went with `ParticipantSecret` so that it will be clear what this is on the TypeScript side

jowparks (approved):

PR #4600, opened 2024-01-25
fixes flaky filestore test

patnir:
## Summary

I think what is happening here is that the node js event loop needs to run 1 more time to confirm promises are settled

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):
you know what I always say, don't let perfect be the enemy of good. I'll take a look after this to try to find a better way to resolve this flakyness. Thanks for doing this.

PR #4602, opened 2024-01-25
Feat/ifl 2132/multisig integration

patnir:
## Summary

The new test case is for the frost functionality and specifically tests the ability to perform a multisig transaction.

1. Generates a set of participant identifiers and secrets for a multisig transaction.
2. Imports participant accounts imports the accounts into the wallet (these accounts all have the same view key)
3. Creates a transaction from the multisig account back to a recipient account.
4. Signs the transaction using the multisig keys and generates signature shares.
5. Signs the transaction using Frost signature scheme.
6. Adds the Frost transaction to a new block and verifies its addition to the chain.
7. Updates the wallet's head and verifies the balance of the recipient account.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/wallet.test.slow.ts
```diff
@@ -1128,4 +1141,193 @@ describe('Wallet', () => {
...
+
+      const unsignedTransaction = rawTransaction.build(
+        trustedDealerPackage.proofGenerationKey,
+        trustedDealerPackage.viewKey,
+        trustedDealerPackage.outgoingViewKey,
+        trustedDealerPackage.publicAddress,
+      )
+
+      const signingPackage = unsignedTransaction.signingPackage(signingCommitments)
+      const publicKeyRandomnes = unsignedTransaction.publicKeyRandomness()
```
hughy, line 1288: nit - dropped an `s` from `Randomness` :trollface:

hughy (approved):

ironfish/src/wallet/wallet.test.slow.ts
```diff
@@ -1128,4 +1141,193 @@ describe('Wallet', () => {
...
+          },
+        ],
+        fee: BigInt(0),
+      })
+
+      // Create a block with a miner's fee and the transaction to send IRON to the multisig account
+      const minersfee2 = await nodeTest.strategy.createMinersFee(
+        transaction.fee(),
+        block.header.sequence + 1,
+        generateKey().spendingKey,
```
hughy, line 1252: nit - I think we could reuse the `miner` account fixture here and below instead of generating new keys

patnir: done

PR #4611, opened 2024-01-25
feat: create trusted dealer key packages rpc

jowparks:
## Summary
1. Sets up `multisig` RPC namespace
2. Adds RPC for creating trusted dealer key package for multisig accounts using the FROST protocol. Input requires that trusted dealer has retrieved identifiers from participants before calling RPC.
## Testing Plan
Test passes

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/multisig/createTrustedDealerKeyPackage.ts
```diff
@@ -0,0 +1,85 @@
...
+import { generateKey, splitSecret } from '@ironfish/rust-nodejs'
+import * as yup from 'yup'
+import { ApiNamespace } from '../namespaces'
+import { routes } from '../router'
+
+export type CreateTrustedDealerKeyPackageRequest = {
+  minSigners: number
+  maxSigners: number
+  participants: Array<{
+    identifier: string
```
hughy, line 13: I was thinking about whether we'd want this to be `Identifier` or `Identity`

I think that since the response uses `Identifier` it probably makes more sense to use `Identifier` as the input rather than forcing the wallet or RPC to transform between the two

hughy (approved):

ironfish/src/rpc/routes/multisig/createTrustedDealerKeyPackage.ts
jowparks: Hmm, it is a good point. If we imagine the case where the identities are provided to the trusted dealer before key generation, they will have the ids in identity format, which will need to be stored. At some point we will have to do the transform from identity to identifier though

PR #4616, opened 2024-01-26
adds version to RawTransactionSerde

hughy:
## Summary

if and when we change the serialization of a RawTransaction we risk breaking offline signing on transactions that were created on prior versions of the code. we may have already done this once

including a version in the serialization of the RawTransaction will clarify errors deserializing old serialized transactions and enable us to support backwards-compatibility by version

- renames version field used for transaction version to 'transactionVersion'
- adds version field for serde version
- sets RawTransactionSerde version to 3: this is the second time we've changed it, and 3 avoids collision with any transaction versions
- throws error if version does not match current version when deserializing

## Testing Plan

- updates unit tests
- adds unit test

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/primitives/rawTransaction.ts
+++ b/ironfish/src/primitives/rawTransaction.ts
@@ -42,7 +42,7 @@ export interface MintData {
 }
 
 export class RawTransaction {
-  version: TransactionVersion
+  transactionVersion: TransactionVersion
   expiration: number | null = null
   fee = 0n
   mints: MintData[] = []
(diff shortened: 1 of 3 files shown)


ironfish/src/primitives/rawTransaction.ts
```diff
@@ -264,7 +270,15 @@ export class RawTransactionSerde {
 
     const version = reader.readU8()
 
-    const raw = new RawTransaction(version)
+    if (version !== RawTransactionSerde.VERSION) {
+      throw new RawTransactionSerdeError(
+        `RawTransaction serialization version ${version} does not match current version ${RawTransactionSerde.VERSION}`,
+      )
+    }
```
hughy, line 277: Alternatively, we could still try to deserialize if `version` is lower than the minimum version and use the version we read here in place of `transactionVersion`

PR #4620, opened 2024-01-26
adds nullable serialized UnsignedTransaction to RawTransaction

hughy:
## Summary

supports serializing a RawTransaction that includes the transaction with generated proofs. this might be of use in multisig when participants need the serialized unsigned transaction to generate signature shares and in offline signing when the offline device cannot generate proofs

caches the built serialized UnsignedTransaction in the RawTransaction's 'build' method

updates RawTransactionSerde
- increments serde version
- sets unsignedTransactionSerialized to null if the version is earlier than these changes

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


hughy:
Closing this in favor of creating a separate RPC to generate UnsignedTransactions

PR #4629, opened 2024-01-29
Changes frost test to use participantB address

patnir:
The goal is to make it explicit in the test that the participants all have the same address.

This is solely a stylistic change. 

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/wallet.test.slow.ts
+++ b/ironfish/src/wallet/wallet.test.slow.ts
@@ -1230,6 +1230,9 @@ describe('Wallet', () => {
       await expect(node.chain).toAddBlock(block)
       await node.wallet.updateHead()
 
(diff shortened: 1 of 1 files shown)


ironfish/src/wallet/wallet.test.slow.ts
```diff
@@ -1234,7 +1234,7 @@ describe('Wallet', () => {
         account: miner,
         outputs: [
           {
-            publicAddress: participantA.publicAddress,
+            publicAddress: participantB.publicAddress,
```
hughy, line 1237: nit - you could add a comment stating why we can use `participantB` here

hughy (approved):

PR #4636, opened 2024-01-30
Rahul/ifl 2136 add proof authorizing key to account value

patnir:
## Summary

Adds the proof authorization key to the account schema. This is a field that will be populated for both view only accounts and full spend accounts. Every account type will now be able to create unsigned transactions. 

This also still supports older view only account types. Those accounts will not be able to generate the proofs required for creating an unsigned transaction. The user will have to export the account again.  

Also updates the import account and export account RPCs. The changes there are minimal so I included it in this PR. I am happy to split it out if it's necessary

## Testing Plan

1. Exported a view only and a full account before running the migration 
2. Imported those accounts after migrating 

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


patnir:
```
Applying 1 migrations:
  031-add-pak-to-account

Running 031-add-pak-to-account...
Migrating account data to store proof authorization key
 Migrating account newaccount
 Migrating account default
```

ironfish/src/wallet/walletdb/accountValue.ts
```diff
@@ -37,13 +38,18 @@ export class AccountValueEncoding implements IDatabaseEncoding<AccountValue> {
...
     bw.writeU16(value.version)
     bw.writeVarString(value.id, 'utf8')
     bw.writeVarString(value.name, 'utf8')
     if (value.spendingKey) {
       bw.writeBytes(Buffer.from(value.spendingKey, 'hex'))
     }
+
+    if (value.proofAuthorizingKey) {
+      bw.writeBytes(Buffer.from(value.proofAuthorizingKey, 'hex'))
+    }
```
hughy, line 52: I think this is fine, but I was initially uneasy about adding this in the middle of the serialization.

Since the migration will be reading the old version and overwriting with the new version I don't think we'll see any problems

hughy (approved):
could we add another test case for this version of bech32 export to `ironfish/src/rpc/routes/wallet/__importTestCases__`?

patnir:
> ironfish/src/rpc/routes/wallet/__importTestCases__

Done

patnir:
> could we add another test case for this version of bech32 export to `ironfish/src/rpc/routes/wallet/__importTestCases__`?

Done

PR #4637, opened 2024-01-30
adds BuildTransaction worker job

hughy:
## Summary

we offload PostTransaction to the worker pool because proof generation is compute intensive and could otherwise bog down the node process

creating an UnsignedTransaction using 'build' generates proofs and would create the same problem

the new job allows us to offload compute to the worker pool in RPC calls that build UnsignedTransactions

## Testing Plan

- adds unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


hughy:
> LGTM as long as we accept `build` for now 😅

A better term has yet to emerge, so we decided to go with `build`

PR #4640, opened 2024-01-30
feat: experimental flag

jowparks:
## Summary
WIP:

Adds experimental flag in config that can be used to gate code in staging, currently this would be used in RPC but could be used anywhere config is available.
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/router.ts
+++ b/ironfish/src/rpc/routes/router.ts
@@ -52,7 +52,15 @@ export class Router {
       throw new RouteNotFoundError(route, namespace, method)
     }
 
-    const { handler, schema } = methodRoute
+    const { handler, schema, experimental } = methodRoute
+
+    if (experimental && !this.server.context.config.experimental) {
+      throw new RpcResponseError(
+        `Experimental features for route ${route} is not enabled, please run server with --experimental flag`,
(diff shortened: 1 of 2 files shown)


ironfish/src/rpc/routes/router.ts
```diff
@@ -52,7 +52,15 @@ export class Router {
       throw new RouteNotFoundError(route, namespace, method)
     }
 
-    const { handler, schema } = methodRoute
+    const { handler, schema, experimental } = methodRoute
+
+    if (experimental && !this.server.context.config.experimental) {
```
NullSoldier, line 57: I would prefer not to add a blanket feature like this. It should be feature based. I also don't think we need to gate anything with this right now, do we? What would we need to gate?

jowparks: We don't necessarily need it right now, but we do have RPCs exposed that people technically shouldn't be using. It could be helpful in this case, lets say if a bridge integrator starts operating with this before we complete the feature.


This PR is not fully baked at all, just using as a placeholder for discussion.

jowparks: FWIW I agree, namespacing the features seems like a good idea.

jowparks:
Closing as won't do for now, but we can keep discussion here and I will add task to backlog for when we need this

PR #4641, opened 2024-01-30
Rahul/ifl 2144 refactor key object to return nsk instead of pgk

patnir:
## Summary

We will allow view only accounts now to construct proofs for creating unsigned transactions. 

As a part of this, we need the view only accounts to contain the proof authorizing key. 

Currently we return all the view keys and the proof generation key (which was added in the last release). The authorizing key can be derived from the generation key.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust-nodejs/src/structs/transaction.rs
+++ b/ironfish-rust-nodejs/src/structs/transaction.rs
@@ -14,6 +14,7 @@ use ironfish::frost::round1::SigningCommitments;
 use ironfish::frost::round2::SignatureShare;
 use ironfish::frost::Identifier;
 use ironfish::frost::SigningPackage;
+use ironfish::keys::{ProofAuthorizingKey, ProofAuthorizingKeySerializable};
 use ironfish::serializing::hex_to_vec_bytes;
 use ironfish::serializing::{bytes_to_hex, hex_to_bytes};
 use ironfish::transaction::unsigned::UnsignedTransaction;
@@ -22,7 +23,6 @@ use ironfish::transaction::{
     TRANSACTION_FEE_SIZE, TRANSACTION_PUBLIC_KEY_SIZE, TRANSACTION_SIGNATURE_SIZE,
 };
 use ironfish::{
-    keys::proof_generation_key::{ProofGenerationKey, ProofGenerationKeySerializable},
     MerkleNoteHash, OutgoingViewKey, ProposedTransaction, PublicAddress, SaplingKey, Transaction,
     ViewKey,
 };
@@ -327,7 +327,7 @@ impl NativeTransaction {
     #[napi]
     pub fn build(
         &mut self,
-        proof_generation_key_str: String,
+        proof_authorizing_key: String,
         view_key_str: String,
         outgoing_view_key_str: String,
         intended_transaction_fee: BigInt,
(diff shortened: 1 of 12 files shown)


ironfish-rust/src/keys/proof_authorizing_key.rs
```diff
@@ -0,0 +1,112 @@
...
+
+use jubjub::Fr;
+
+use crate::{
+    errors::{IronfishError, IronfishErrorKind},
+    serializing::{bytes_to_hex, hex_to_bytes},
+};
+
+pub type ProofAuthorizingKey = jubjub::Fr;
+
```
patnir, line 13: Is there a better way to do this/ better location for this?

patnir: i could do ProofAuthorizingKey { nsk: jubjub::Fr } but i think it unnecessarily complicates things

hughy: It seems like the purpose of this is to be able to call `proof_authorizing_key.hex_key()`?

I'm a Rust novice, but it seems like it might be simpler to add something like this to our `serializing` mod:
```
pub trait HexKey {
    fn hex_key(&self) -> String;
}

impl HexKey for jubjub::Fr {
    fn hex_key(&self) -> String {
        bytes_to_hex(&self.to_bytes())
    }
}
```

hughy: You could also call `bytes_to_hex(proof_authorizing_key.to_bytes())` directly

patnir: That is 1 reason + having `ProofAuthorizingKey` as the type everywhere would make it easier to read. Otherwise you have to do some mental math to confirm indeed that `jubjub::Fr` is a proof_authorizing_key by looking at the Sapling key. Don't feel strongly about this thought. A trait would be useful regardless

ironfish-rust-nodejs/src/structs/transaction.rs
```diff
@@ -327,7 +327,7 @@ impl NativeTransaction {
     #[napi]
     pub fn build(
         &mut self,
-        proof_generation_key_str: String,
+        proof_authorizing_key: String,
```
hughy, line 330: It might be tricky to change this interface before changing the account structure -- right now the multisig integration test depends on using the `proofGenerationKey` that's returned in the trusted dealer key package

ironfish-rust/src/keys/proof_authorizing_key.rs
hughy: The type alias and trait for `ProofAuthorizingKey` seem unnecessary to me for now

ironfish-rust-nodejs/src/structs/transaction.rs
patnir: i was working through changing the account structure and kept running into performing hacks to get the authorizing key. I figured it might make more sense to fix this first because changing the account structure that has larger implications: https://github.com/iron-fish/ironfish/pull/4636

ironfish-rust/src/keys/proof_authorizing_key.rs
patnir: Ok I'll change it to just the `bytes_to_hex(proof_authorizing_key.to_bytes())`

ironfish-rust-nodejs/src/structs/transaction.rs
hughy: Yeah changing the account structure is non-trivial. I meant more that we can make the change to the key object to return nsk instead of pgk without changing this interface

patnir:
In an upcoming PR, I'm going to change TrustedDealerPackage to return nsk instead of pgk

ironfish-rust/src/keys/proof_authorizing_key.rs
patnir: I tried the above and I think I still prefer to have it as a separate type/ alias. The code is much cleaner, fits the patterns with everything else, and doesn't require and additional dependency to be added to the rust-nodejs project

patnir:
Got some feedback that this change is

PR #4646, opened 2024-01-31
Move Consensus into Network

NullSoldier:
## Summary
Now the network creates the consensus from the definition.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/networks/network.ts
+++ b/ironfish/src/networks/network.ts
@@ -3,23 +3,30 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { Assert } from '../assert'
 import { Consensus } from '../consensus'
+import { SerializedBlock } from '../primitives/block'
 import { MathUtils } from '../utils'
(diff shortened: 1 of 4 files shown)


ironfish/src/networks/network.ts
```diff
@@ -15,11 +15,11 @@ export class Network {
 
   private miningRewardCachedByYear = new Map<number, number>()
 
-  constructor(definition: NetworkDefinition, consensus: Consensus) {
+  constructor(definition: NetworkDefinition) {
     this.id = definition.id
     this.default = isDefaultNetworkId(definition.id)
     this.definition = definition
```
danield9tqh, line 21: Doesn't seem like this is an issue rn but we do create a new object from consensus params in the `Consensus` constructor. Wondering if it would make sense to do the same here?

NullSoldier: Fixed

danield9tqh (approved):

NullSoldier:
LGTM too

PR #4649, opened 2024-01-31
adding serializable trait to jubjub::Fr

patnir:
## Summary

The type of the proof authorizing key is jubjub::Fr. We don't have a clean way to convert to and from bytes/ string for this type. 

This will be useful when we start to use proof authorizing key instead of the proof generation key in our napi bindings and account structure

## Testing Plan

Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-rust/src/serializing/fr.rs
```diff
@@ -0,0 +1,109 @@
...
+pub const MAC_SIZE: usize = 16;
+
+pub trait FrSerializable {
+    fn serialize(&self) -> [u8; 32];
+    fn deserialize(bytes: [u8; 32]) -> Result<Fr, IronfishError>;
+    fn hex_key(&self) -> String;
+    fn from_hex(hex_key: &str) -> Result<Fr, IronfishError>;
+}
+
+impl FrSerializable for Fr {
```
jowparks, line 19: Can you provide context why we need this, with example code? Can't we just call `bytes_to_hex` on the `to_bytes` when we need it, This is not a key so we don't need to match the implementations like we did with proof generation key.

jowparks: This may be useful at some point, but I am not sure all this code merits one call, could be wrong

ironfish-rust/src/serializing/fr.rs
```diff
@@ -0,0 +1,109 @@
...
+    fn serialize(&self) -> [u8; 32] {
+        self.to_bytes()
+    }
+
+    fn deserialize(bytes: [u8; 32]) -> Result<Self, IronfishError> {
+        let nsk = match Fr::from_bytes(&bytes).into() {
+            Some(nsk) => nsk,
+            None => {
+                return Err(IronfishError::new(
+                    IronfishErrorKind::InvalidProofAuthorizingKey,
```
jowparks, line 29: `Fr` isn not just a proof authorizing key, it is a scalar field element, the proof authorizing key just happens to be a scalar field element.

ironfish-rust/src/serializing/fr.rs
patnir: The alternative would be to use the following logic in our napi layer everywhere we want to deserialize: 

image

(idk if I'm explaining this right)

Writing tests will also be much easier

jowparks: Ok you convinced me :)

hughy (approved):

PR #4654, opened 2024-02-01
fixes Bech32Encoder backwards compatibility

hughy:
## Summary

we recently released a change that added an optional field to the end of bech32 account encodings. however, we did not update the encoder to correctly handle older encodings that are missing that optional field

if a user attempts to import an older bech32 account they will see an error when the Bech32Encoder attempts to read a byte after the end of their import string

updates the Bech32Encoder to support decoding multiple versions:
- defines 'Bech32Decoder' as a function that takes a reader and decoding options and returns an AccountImport
- adds a mapping to Bech32Encoder that maps supported versions to decode functions for those versions

## Testing Plan

adds a test case for a bech32 account from v1.16.0

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):

patnir (approved):
Confirmed that this works: 

image

PR #4672, opened 2024-02-05
feat(IFL-2174): create signing share cli

jowparks:
## Summary
Takes signing package and transaction in order to create signing share for a given participants account when using a frost multisig account.
## Testing Plan
UNTESTED, need to ship `create-signing-package` first
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/wallet/multisig/create-signature-share.ts
```diff
@@ -0,0 +1,49 @@
...
+    signingPackage: Flags.string({
+      char: 's',
+      description: 'The signing package for which the signature share will be created',
+      required: true,
+    }),
+  }
+
+  async start(): Promise<void> {
+    const { flags } = await this.parse(CreateSignatureShareCommand)
+    const signingPackage = flags.signingPackage
```
hughy, line 34: nit - could we pass this directly from `flags` below like we do with other args?

hughy (approved):

PR #4674, opened 2024-02-05
Rahul/ifl 2178 remove max signers input from split spender key

patnir:
## Summary

Max signers is always the number of identifiers. There is even a check in frost that confirms this. I believe removing it would be a better user experience and less confusing for the user. 

## Testing Plan

Updated unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/frost_utils/split_secret.rs
+++ b/ironfish-rust/src/frost_utils/split_secret.rs
@@ -62,7 +62,7 @@ mod test {
 
     #[test]
     fn test_invalid_secret() {
-        let mut identifiers = create_identifiers(10);
+        let identifiers = create_identifiers(10);
 
         let vec = vec![1; 31];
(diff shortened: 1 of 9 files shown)


ironfish-rust/src/frost_utils/split_secret.rs
```diff
@@ -57,22 +58,33 @@ pub(crate) fn split_secret(
...
+        let mut identifiers = Vec::new();
+
+        for _ in 0..10 {
+            identifiers.push(
+                Secret::random(thread_rng())
+                    .to_identity()
+                    .to_frost_identifier(),
+            );
+        }
+
```
patnir, line 78: is there a good way to create a test utility for generating identifiers?

hughy (approved):

PR #4677, opened 2024-02-05
adds multisig/aggregateSigningShares RPC

hughy:
## Summary

takes a publicKeyPackage, an unsignedTransaction, a signingPackage for that transaction, and a list of signatureShares for that signingPackage

uses UnsignedTransaction.signFrost to aggregate the signatureShares, sign the transaction, and produce a signed transaction

returns a signed transaction serialized as hex

## Testing Plan

RPC integration test will be added in a followup PR

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/clients/client.ts
+++ b/ironfish/src/rpc/clients/client.ts
@@ -823,7 +823,7 @@ export abstract class RpcClient {
     },
   }
   multisig = {
-    aggregateSignatureShares: (
+    aggregateSigningShares: (
       params: AggregateSigningSharesRequest,
     ): Promise<RpcResponseEnded<AggregateSigningSharesResponse>> => {
       return this.request<AggregateSigningSharesResponse>(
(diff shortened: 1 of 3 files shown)


ironfish/src/rpc/clients/client.ts
```diff
@@ -821,6 +823,15 @@ export abstract class RpcClient {
     },
   }
   multisig = {
+    aggregateSignatureShares: (
```
jowparks, line 826: We should either rename this to match `signFrost` or rename `signFrost` to match this.

jowparks (approved):
Looks good, we just need to rename this (or signFrost) and in a subsequent PR write integration test like you suggested offline.

ironfish/src/rpc/clients/client.ts
```diff
@@ -821,6 +823,15 @@ export abstract class RpcClient {
...
+      params: AggregateSigningSharesRequest,
+    ): Promise<RpcResponseEnded<AggregateSigningSharesResponse>> => {
+      return this.request<AggregateSigningSharesResponse>(
+        `${ApiNamespace.multisig}/aggregateSignatureShares`,
```
hughy, line 830: typo: `s/Signature/Signing/`

ironfish/src/rpc/clients/client.ts
hughy: Merging as is for now, we'll settle on naming in a followup

PR #4682, opened 2024-02-05
chore: move multisig rpc into wallet namespace

jowparks:
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/router.ts
```diff
@@ -31,7 +31,8 @@ export class RouteNotFoundError extends RpcResponseError {
 export function parseRoute(
   route: string,
 ): [namespace: string | undefined, method: string | undefined] {
```
jowparks, line 33: Had to change this as it was originally returning:
`['wallet','multisig','route']`
and using the first two elements.

This peels out the first section for the namespace instead.

patnir (approved):

PR #4683, opened 2024-02-06
Add checkpoints to blockchain

danield9tqh:
## Summary
Code implementation of [this discussion](https://discourse.ironfish.network/t/91)

## Testing Plan
Writing unit tests and simulations

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -100,8 +100,6 @@ export class Blockchain {
 
   private _head: BlockHeader | null = null
 
-  private latestCheckpoint: BlockHeader | null = null
-
   get head(): BlockHeader {
     Assert.isNotNull(
       this._head,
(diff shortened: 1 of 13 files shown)


ironfish/src/consensus/consensus.ts
```diff
@@ -61,18 +62,29 @@ export type ConsensusParameters = {
...
+  checkpoints: Checkpoint[]
 }
 
 export class Consensus {
   readonly parameters: ConsensusParameters
+  readonly checkpoints: Map<number, Buffer>
 
   constructor(parameters: ConsensusParameters) {
     this.parameters = parameters
+    this.checkpoints = new Map<number, Buffer>()
```
mat-if, line 79: Can you add some tests here around the new logic in the constructor and also the isActive function? Doesn't need to be too crazy, but its finnicky enough that it is liable to silently break in the future

ironfish/src/blockchain/blockchain.ts
```diff
@@ -100,6 +100,8 @@ export class Blockchain {
 
   private _head: BlockHeader | null = null
 
+  private latestCheckpoint: BlockHeader | null = null
```
mat-if, line 103: Can you add some tests around this property and when it is set for sanity sake? I don't feel strongly about whether they need to be whole new tests or just added expects to the existing tests, up to you.

mat-if (commented):
Looks good, just a few things around tests

ironfish/src/blockchain/blockchain.ts
danield9tqh: Added asserts to existing tests for this

mat-if (approved):

PR #4689, opened 2024-02-06
Rahul/ifl 2187 replace keypackage with account name input for

patnir:
## Summary

Changes create signing commitment RPC to accept account name as input instead of the keyPackage

## Testing Plan

Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


hughy (approved):
looks good, but don't think we need the `wallet.nodeClient` dependency

PR #4691, opened 2024-02-06
refactor: serialization signature share

jowparks:
## Summary
Updates serialization of signature share to include identifier. This allows us to have a single string output for the identifier/signingShare.

Note: We now have to pass in identifier to this method just so the serialization can include identifier in the hex string.
## Testing Plan

Tests pass

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust-nodejs/src/frost.rs
+++ b/ironfish-rust-nodejs/src/frost.rs
@@ -8,10 +8,10 @@ use crate::{
 };
 use ironfish::{
     frost::{keys::KeyPackage, round2::Randomizer, Identifier, SigningPackage},
-    frost_utils::split_spender_key::split_spender_key,
     frost_utils::{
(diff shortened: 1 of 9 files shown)


ironfish-rust-nodejs/src/frost.rs
```diff
@@ -28,6 +32,35 @@ pub struct NativeCommitment {
     pub binding: String,
 }
 
+pub struct IdentifierSignatureShare {
+    pub identifier: Identifier,
+    pub signature_share: SignatureShare,
+}
+
```
jowparks, line 39: Starting the migration to `SignatureShare` from `SigningShare`, we can fixup the rest in a following PR.

andiflabs (approved):

PR #4692, opened 2024-02-06
Rahul/adds error message to multi sig signer assert

patnir:
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/account/account.ts
```diff
@@ -42,6 +43,12 @@ export function AssertMultiSig(account: Account): asserts account is MultiSigAcc
   )
 }
```
jowparks, line 45: This should look like `AssertMultiSig` above, if we are just going to be using it with accounts we might as well put it on the account.

jowparks: Also we might as well rename this to `AssertSignerMultiSig`

jowparks (approved):

PR #4704, opened 2024-02-07
Prefer `ironfish_rust::participant::Identity` over `frost::Identifier`

andiflabs:
## Summary

`Identity` holds more information than `Identifier` and hence should be the preferred data structure. An `Identity` can always be converted to `Identifier` but the other way around is not possible. 

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-rust/src/test_util.rs
```diff
@@ -60,17 +60,10 @@ pub(crate) fn auth_path_to_root_hash(
...
-            Secret::random(thread_rng())
-                .to_identity()
-                .to_frost_identifier(),
-        );
-    }
-
-    identifiers
+/// Helper function to create a list of random identifiers for multisig participants.
+pub fn create_multisig_identifiers(num_identifiers: usize) -> Vec<Identifier> {
+    (0..10)
```
rohanjadvani, line 65: Should this use `num_identifiers`?

andiflabs: Oops, yes, fixed

rohanjadvani (approved):

ironfish-rust/src/frost_utils/split_secret.rs
```diff
@@ -30,25 +31,35 @@ pub(crate) fn split_secret<R: RngCore + CryptoRng>(
...
-        KeyPackage::try_from(v)?;
-    }
-
     let mut key_packages: HashMap<_, _> = HashMap::new();
 
-    for (identifier, secret_share) in shares {
+    for (frost_id, secret_share) in shares {
+        let identity = frost_id_map
+            .remove(&frost_id)
+            .expect("frost returned an identifier that was not passed as an input");
```
jowparks, line 60: Should we return an `IronfishError` instead of expect?

patnir (approved):

ironfish-rust/src/frost_utils/split_secret.rs
```diff
@@ -5,23 +5,24 @@
...
+use ironfish_frost::participant::Identity;
 use rand::{CryptoRng, RngCore};
 use std::collections::HashMap;
 
 use crate::errors::{IronfishError, IronfishErrorKind};
 
 pub struct SecretShareConfig {
     pub min_signers: u16,
-    pub identifiers: Vec<Identifier>,
+    pub identifiers: Vec<Identity>,
```
jowparks, line 18: rename `identities`

ironfish-rust/src/frost_utils/split_secret.rs
andiflabs: This would be a symptom of a bug in the `frost` crate, not something that depends on user input, so I'd say that a panic here is fine. The rule I'm following is "panic for internal bugs, results for input and system errors".

ironfish-rust/src/test_util.rs
```diff
@@ -60,17 +60,10 @@ pub(crate) fn auth_path_to_root_hash(
...
-        identifiers.push(
-            Secret::random(thread_rng())
-                .to_identity()
-                .to_frost_identifier(),
-        );
-    }
-
-    identifiers
+/// Helper function to create a list of random identifiers for multisig participants.
+pub fn create_multisig_identifiers(num_identifiers: usize) -> Vec<Identifier> {
```
jowparks, line 64: rename `create_multisig_identities`

Also should the return type here be `Vec<Identity>`?

jowparks: Oh cool, I see the collect is so that you can gfet back the `Identifiers`

ironfish/src/rpc/routes/wallet/multisig/createSigningCommitment.test.ts
```diff
@@ -73,7 +73,7 @@ describe('Route wallet/multisig/createSigningCommitment', () => {
 
   it('should create signing commitment', async () => {
     const participants = Array.from({ length: 3 }, () => ({
-      identifier: ParticipantSecret.random().toIdentity().toFrostIdentifier(),
+      identifier: ParticipantSecret.random().toIdentity().serialize().toString('hex'),
```
jowparks, line 76: If these are `Identity` we should rename the arguments.

PR #4715, opened 2024-02-09
Add a one-time difficulty adjustment when FishHash activates

mat-if:
## Summary

Introduces a one-time difficulty adjustment of `(difficulty / 100)` at the sequence that FishHash activates. The reasoning here is that FishHash has dramatically lower hashrate than the current blake3 algorithm. Initial testing shows a slowdown of 250-500x on GPUs. The reasoning for only adjusting by 100 is because we don't want to overshoot and make things too easy, but this still keeps the blocktime in the realm of usable.

If we say the current blake3 hashrate is 1,000,000, and the FishHash hashrate is 250x lower, that's 4,000. If it's 500x, it's 2,000.

Based on that blake3 hashrate, and a 60 second blocktime, the given difficulty would be 60,000,000. Given these FIshHash hashrates, we would see blocktimes in the 4-8 hour range. This is not good. If we simply divide by 100, we land on a new difficulty of 600,000. This gives us blocktimes in the 2.5 to 5 minute range. This would naturally adjust to the 60 second blocktime quite quickly, leading to a smooth transition. 

## Testing Plan

Unit testing, manual testing

## Documentation

N/A

## Breaking Change

```
[x] Yes
```

### DIFF
--- a/ironfish/src/primitives/target.ts
+++ b/ironfish/src/primitives/target.ts
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+import { ActivationSequence } from '../consensus'
 import { BigIntUtils } from '../utils/bigint'
 
 /**
@@ -67,6 +68,8 @@ export class Target {
     targetBlockTimeInSeconds: number,
     targetBucketTimeInSeconds: number,
     maxBuckets: number,
+    enableFishHashSequence: ActivationSequence,
+    sequence: number,
   ): Target {
     const parentDifficulty = previousBlockTarget.toDifficulty()
 
@@ -77,6 +80,8 @@ export class Target {
       targetBlockTimeInSeconds,
       targetBucketTimeInSeconds,
       maxBuckets,
+      enableFishHashSequence,
+      sequence,
     )
 
     return Target.fromDifficulty(difficulty)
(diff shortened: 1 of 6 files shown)


ironfish/src/primitives/target.ts
```diff
@@ -110,6 +115,8 @@ export class Target {
     targetBlockTimeInSeconds: number,
     targetBucketTimeInSeconds: number,
     maxBuckets: number,
+    enableFishHashSequence: ActivationSequence,
```
danield9tqh, line 118: could combine these into one param if we want to check sequence outside of this function

danield9tqh (approved):

ironfish/src/primitives/target.ts
```diff
@@ -104,26 +103,35 @@ export class Target {
...
     bucket = Math.min(bucket, maxBuckets)
 
-    const difficulty =
+    let difficulty =
       previousBlockDifficulty - (previousBlockDifficulty / 2048n) * BigInt(bucket)
 
+    // A one-time difficulty adjustment when the mining algorithm changes from
+    // blake3 to fishhash, since the fishhash hashrate is much lower than blake3
+    if (sequence === enableFishHashSequence) {
+      difficulty /= 100n
```
danield9tqh, line 132: Should maybe have this in a constant. Was also thinking about if we want this number in the consensus params too. Just in case we wanted to make it different on different networks. Maybe a consensus param called `oneTimeFishHashAdjustment` or something like that

PR #4716, opened 2024-02-12
feat: large prompt signature share

jowparks:
## Summary
Prompting mechanism for long text input truncates with `Cli.ux.prompt` and is very laggy when using `inquirer` library. 
When using the `CliUx` of `oclif/core` I see truncation of input.

This creates a simple wrapper around the node native input system for prompting long text.

## Testing Plan

https://github.com/iron-fish/ironfish/assets/26990067/42d8b8c6-2e25-480b-a34c-fd3eea1bd64a

Tested on macOS and linux, same behavior exists on both systems.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/utils/longPrompt.ts
+++ b/ironfish-cli/src/utils/longPrompt.ts
@@ -1,0 +1,19 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import readline from 'readline'
+
+// Most effective way to take in a large textual prompt input without affecting UX
+export function largePrompt(question: string): Promise<string> {
+  const rl = readline.createInterface({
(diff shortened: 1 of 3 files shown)


patnir (approved):
Looks great!

ironfish-cli/src/utils/longPrompt.ts
```diff
@@ -0,0 +1,19 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import readline from 'readline'
+
+// Most effective way to take in a large textual prompt input without affecting UX
```
andiflabs, line 6: I'm wondering if this is fixing a MacOS-specific problem? If so, I would write it in the comment, otherwise someone like me might come around and remove this function if they cannot reproduce the lag problem

PR #4721, opened 2024-02-12
Rahul/long prompt for account import

patnir:
## Summary

Uses large prompt for account import

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/utils/longPrompt.ts
```diff
@@ -17,3 +17,18 @@ export function largePrompt(question: string): Promise<string> {
...
+  },
+): Promise<string> {
+  let userInput = (await fetchResponse(question)).trim()
+
+  while (userInput.length === 0 && options?.required) {
+    userInput = (await fetchResponse(question)).trim()
+  }
+
+  return userInput
+}
```
patnir, line 34: we'll need this pattern in multiple places to replace 
```
CliUx.ux.prompt('',
      {
        required: true,
      },
    )
```

ironfish-cli/src/utils/longPrompt.ts
```diff
@@ -4,7 +4,7 @@
 import readline from 'readline'
 
 // Most effective way to take in a large textual prompt input without affecting UX
```
jowparks, line 6: Instead of separate methods here, lets just use recursion.

jowparks (approved):

PR #4723, opened 2024-02-13
Fix flaky fileStore test

danield9tqh:
## Summary
This test is sometimes failing on the second expect `expect(writeFileSpy).toHaveBeenCalledTimes(2)`. Changing this to wait for everything to execute until checking that the file is saved for a second time. This still tests the desired functionality because the test is testing that "The 2nd save IS NOT called until the first save resolves". That is still being tested with these changes and we can make the test less flaky without reducing coverage.

## Testing Plan
Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


patnir (approved):
TY for fixing this!

PR #4724, opened 2024-02-13
Change multisig implementation to use deterministic nonces and commitments

andiflabs:
## Summary

The nonces/commitments are built from the secret share, the transaction hash, and the list of signers, so they are unique per transaction and they cannot be accidentally reused to sign multiple transactions.

## Testing Plan

Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/wallet/multisig/createSigningPackage.test.ts
```diff
@@ -52,23 +45,29 @@ describe('Route multisig/createSigningPackage', () => {
...
+    )
+      .serialize()
+      .toString('hex')
+
+    const commitments = await Promise.all(
+      participants.map(async (_) => {
+        const signingCommitment =
+          await routeTest.client.wallet.multisig.createSigningCommitment({
+            unsignedTransaction,
+            signers: [], // TODO
```
rohanjadvani, line 62: Does this need to be set?

andiflabs: That needed to be set in some places (to make the tests more realistic). I forgot to resolve the TODOs before pushing; fixed

ironfish-cli/src/commands/wallet/multisig/create-signature-share.ts
```diff
@@ -28,6 +28,13 @@ export class CreateSignatureShareCommand extends IronfishCommand {
       description: 'The signing package for which the signature share will be created',
       required: false,
     }),
+    signerIdentity: Flags.string({
```
hughy, line 31: will this always need to include all of the identities in the group? or all of the identities that were passed when creating the commitment?

andiflabs: Only the ones that were passed when creating the commitment (i.e. the actual signers).

This argument can go away once we place this information into the signing package.

ironfish-rust-nodejs/src/frost.rs
```diff
@@ -7,45 +7,75 @@ use crate::{
...
+use ironfish_frost::nonces::deterministic_signing_nonces;
 use napi::{bindgen_prelude::*, JsBuffer};
 use napi_derive::napi;
 use rand::thread_rng;
+use std::ops::Deref;
+
+fn try_deserialize_signers<I, S>(signers: I) -> Result<Vec<Identity>>
+where
+    I: IntoIterator<Item = S>,
+    S: Deref<Target = str>,
```
patnir, line 30: should we have some unit tests for this function + edgecases?

andiflabs: So... I tried to add some tests, but unfortunately it looks like `ironfish-rust-nodejs` is untestable. Running `cargo test` on the crate results in linking errors (which is understandable, given the purpose of the crate, but this makes testing impossible with the current state of things). Maybe in the long term we can move this function somewhere else where it can be tested.

FWIW, the underlying serialization code (`Identity::deserialize_from`) does have tests.

patnir (approved):

PR #4727, opened 2024-02-13
Decrypt notes in parallel

dguenther:
## Summary

Scanning 40+ accounts on one node right now is almost impossible due to only executing one decryptNotes batch at a time, and splitting up the decryptNotes batches per transaction and per account. In the past, this was more of a concern that the worker pool job queue would get blocked, but now that we process tasks round robin from every task queue, we should be able to keep the queue full.

With nodeWorkersMax and nodeWorkers set to 12, this increases decryption time from about 1 block per second to about 10, depending on the number of notes per block. A secondary benefit is the decryption no longer happens inside the DB lock -- the account status seems to update more frequently on this branch, but it could be due to the scan speed.

This is still WIP, I don't have time to clean up the code and tests right now, and also needs testing to make sure P2P and syncing performance aren't affected. I ran a perf test to verify that the main thread isn't blocked.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/utils.ts
+++ b/ironfish/src/rpc/routes/wallet/utils.ts
@@ -176,6 +176,7 @@ export async function getTransactionNotes(
     }
 
     decryptNotesPayloads.push({
+      accountId: account.id,
       serializedNote: note.serialize(),
       incomingViewKey: account.incomingViewKey,
       outgoingViewKey: account.outgoingViewKey,
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -403,11 +403,10 @@ export class Wallet {
         async (a) => await this.isAccountUpToDate(a),
       ))
 
-    const decryptedNotesByAccountId = new Map<string, Array<DecryptedNote>>()
-
     const batchSize = 20
+    const notePromisesByAccountId: Map<string, Array<Promise<Array<DecryptedNote>>>> = new Map()
     for (const account of accountsToCheck) {
-      const decryptedNotes = []
+      const notePromises = []
       let decryptNotesPayloads = []
       let currentNoteIndex = initialNoteIndex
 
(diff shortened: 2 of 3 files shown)


ironfish/src/wallet/wallet.ts
```diff
@@ -484,9 +483,47 @@ export class Wallet {
...
+    const decryptedNotesMap: Map<string, Map<string, Array<DecryptedNote>>> = new Map()
+    for (const { transaction, result } of decryptedTransactions) {
+      for (const [accountId, decryptedNotes] of result) {
+        if (!decryptedNotes.length) {
+          continue
+        }
+
+        const accountTxnsMap =
+          decryptedNotesMap.get(accountId) ?? new Map<string, Array<DecryptedNote>>()
+        accountTxnsMap.set(transaction.hash().toString('base64'), decryptedNotes)
```
hughy, line 521: why base64 instead of hex?

ironfish/src/workerPool/tasks/decryptNotes.ts
```diff
@@ -19,6 +20,7 @@ export interface DecryptNoteOptions {
 }
 
 export interface DecryptedNote {
+  accountId: string
```
danield9tqh, line 23: Seems like we can get away with not passing the `accountId` through here as its not used in the worker pool job

ironfish/src/wallet/wallet.ts
```diff
@@ -484,9 +483,47 @@ export class Wallet {
...
+      await Promise.all(
+        accounts.map((a) =>
+          this.shouldDecryptForAccount(blockHeader, a).then(
+            (shouldDecrypt): [Account, boolean] => [a, shouldDecrypt],
+          ),
+        ),
+      )
+    )
+      .filter((v) => v[1])
+      .map((v) => v[0])
```
hughy, line 496: I think this can be simplified with `AsyncUtils.filter`

ironfish/src/wallet/wallet.ts
hughy: not sure if it would be more or less efficient, but we could also use a `BufferMap` to avoid converting the hashes to strings

ironfish/src/wallet/wallet.ts
```diff
@@ -484,9 +483,47 @@ export class Wallet {
...
+      ),
+    )
 
+    // account id -> transaction hash -> Array<DecryptedNote>
+    const decryptedNotesMap: Map<string, Map<string, Array<DecryptedNote>>> = new Map()
+    for (const { transaction, result } of decryptedTransactions) {
+      for (const [accountId, decryptedNotes] of result) {
+        if (!decryptedNotes.length) {
+          continue
+        }
```
hughy, line 517: should we insert the empty list anyway since we default to that below?

hughy (approved):
Awesome!

A couple of minor comments, but this will be a huge improvement

ironfish/src/wallet/wallet.ts
danield9tqh: Yea I was thinking the same thing

ironfish/src/wallet/wallet.ts
dguenther: I'll switch to using BufferMap here, it's more clear than making sure the encoding is same on both ends. FWIW under the hood BufferMap just converts Buffer keys using `Buffer.toString('binary')`

ironfish/src/wallet/wallet.ts
dguenther: Sure, I updated this

ironfish/src/rpc/routes/wallet/utils.ts
```diff
@@ -187,6 +187,7 @@ export async function getTransactionNotes(
     }
 
     decryptNotesPayloads.push({
+      accountId: account.id,
```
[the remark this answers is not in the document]
dguenther: Fixed this

PR #4728, opened 2024-02-13
Rahul/fix frost slow test

patnir:
## Summary

**note:** does not solve underlying issue with wallet scanning. I'll leave this in draft as hopefully a short term fix we don't need

This job https://github.com/iron-fish/ironfish/actions/runs/7888494359/job/21526149802 has been failing since this test was introduced. 

1. Adding a scanTransactions call to ensure the balance of the account was updated

## Testing Plan

Remove the fixture and it should pass on the first try

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/wallet.test.slow.ts
```diff
@@ -1251,6 +1251,7 @@ describe('Wallet', () => {
       expect(addResult2.isAdded).toBeTruthy()
 
       await node.wallet.updateHead()
+      await node.wallet.scanTransactions()
```
hughy, line 1254: nit - I think we should be able to move this call to right after all of the account imports

one thing we're running into here is that importing an account through the sdk doesn't automatically trigger a scan (but importing through rpc does)

hughy (approved):

jowparks (approved):

PR #4729, opened 2024-02-13
feat(cli): Ask for confirmation when creating signature share

rohanjadvani:
## Summary

Small UX improvement to ask for user confirmation before creating signature share

## Testing Plan

Ran the command

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/multisig/create-signature-share.ts
+++ b/ironfish-cli/src/commands/wallet/multisig/create-signature-share.ts
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-import { Flags } from '@oclif/core'
+import { CliUx, Flags } from '@oclif/core'
 import { IronfishCommand } from '../../../command'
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/wallet/multisig/create-signature-share.ts
```diff
@@ -43,6 +47,13 @@ export class CreateSignatureShareCommand extends IronfishCommand {
       signingPackage = await longPrompt('Enter the signing package: ')
     }
 
+    if (!flags.confirm) {
+      const confirmed = await CliUx.ux.confirm('Confirm new signature share creation (Y/N)')
```
hughy, line 51: will we come back later to add some display of what's in the transaction?

hughy (approved):

ironfish-cli/src/commands/wallet/multisig/create-signature-share.ts
rohanjadvani: Yeah, I think that requires a lot of changes. I would rather get some sort of confirmation in right now and revisit lalter.

PR #4732, opened 2024-02-14
removes duplicate check for duplicate name in createAccount

hughy:
## Summary

wallet.createAccount already checks for duplicate names

instead of checking in both RPC and wallet sdk, the RPC catches the error that the wallet sdk throws and handles it with the appropriate RPC response

## Testing Plan

- ran `wallet:create` to create accounts, verified that behaviour unchanged for duplicate and unique names

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


jowparks (approved):

jowparks:
Need an update of test assertion, looks good though. Can you link to original PR by Jason for posterity?

PR #4734, opened 2024-02-14
Rahul/unsigned transaction in signing package

patnir:
## Summary

Creates a new signing package struct. 
Adds `unsignedTransaction` to the signing package. 
This new struct is passed between the rust/ typescript layer instead of the frost SigningPackage. 
This change has also led to simplification on the RPC and CLI side for creating signature shares. 
Removes the need to pass the transaction hash of the unsigned transaction

Next step is to remove `unsignedTransaction` from the `aggregateSigningShare` RPC

## Testing Plan

Updated unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-rust/src/frost_utils/signing_package.rs
```diff
@@ -0,0 +1,51 @@
...
+        Ok(vec)
+    }
+
+    pub fn deserialize(bytes: &[u8]) -> Result<SigningPackage, IronfishError> {
+        let mut signing_package_length_bytes = [0u8; 4];
+        signing_package_length_bytes.copy_from_slice(&bytes[..4]);
+
+        let length = u32::from_le_bytes(signing_package_length_bytes);
+
+        let frost_signing_package_bytes = &bytes[4..(4 + length as usize)];
```
hughy, line 40: the byte offsets in here look a bit odd

i wonder if it might be cleaner to follow the reader/writer pattern that we've used in other serializations like `SigningCommitment`

patnir: This is more consistent with Signature Share. I don't think doing read/ write will help with the "oddness". I think the core issue is that we don't know how large the unsigned transaction is going to be

patnir: I'm not opposed to changing it to the read/ write pattern.

hughy: > I don't think doing read/ write will help with the "oddness".

I think using a reader will let us rely on a cursor for reading through the byte array rather than slicing the array with indices.  It might make it a bit easier to follow since that's the pattern that most of our encodings use

patnir: I'll give that a go

ironfish-rust/src/frost_utils/signing_package.rs
```diff
@@ -0,0 +1,51 @@
...
+        // write length
+        vec.extend_from_slice(&length_bytes);
+        // write frost signing package
+        vec.extend_from_slice(&serialized_frost_signing_package);
+
+        let mut transaction_vec = vec![];
+        // write unsigned transaction
+        self.unsigned_transaction.write(&mut transaction_vec)?;
+
+        vec.extend_from_slice(&transaction_vec);
```
andiflabs, line 29: This could be written directly into the `vec` vector:

```rust
self.unsigned_ransaction.write(&mut vec);
```

ironfish-rust/src/frost_utils/signing_package.rs
```diff
@@ -0,0 +1,51 @@
...
+    pub unsigned_transaction: UnsignedTransaction,
+    pub frost_signing_package: FrostSigningPackage,
+}
+
+impl SigningPackage {
+    pub fn serialize(&self) -> Result<Vec<u8>, IronfishError> {
+        let serialized_frost_signing_package = self.frost_signing_package.serialize()?;
+        let length_bytes = (serialized_frost_signing_package.len() as u32).to_le_bytes();
+
+        let mut vec: Vec<u8> = vec![];
```
andiflabs, line 19: Nit: `let mut vec = Vec::<u8>::new()` (or even `with_capacity`)

ironfish-rust/src/frost_utils/signing_package.rs
```diff
@@ -0,0 +1,51 @@
...
+#[derive(Clone)]
+pub struct SigningPackage {
+    pub unsigned_transaction: UnsignedTransaction,
+    pub frost_signing_package: FrostSigningPackage,
+}
+
+impl SigningPackage {
+    pub fn serialize(&self) -> Result<Vec<u8>, IronfishError> {
+        let serialized_frost_signing_package = self.frost_signing_package.serialize()?;
+        let length_bytes = (serialized_frost_signing_package.len() as u32).to_le_bytes();
```
andiflabs, line 17: Nit: although unlikely, this cast can truncate the length. I'd use `TryFrom` or `TryInto` for safety:

```rust
u32::try_from(serialized_frost_signing_package.len())
```

hughy (approved):

ironfish/src/rpc/routes/wallet/createAccount.ts
```diff
@@ -46,23 +45,24 @@ routes.register<typeof CreateAccountRequestSchema, CreateAccountResponse>(
...
-      }
-      throw e
+    const name = request.data.name
+
+    if (context.wallet.accountExists(name)) {
+      throw new RpcValidationError(
+        `There is already an account with the name ${name}`,
+        400,
+        RPC_ERROR_CODES.ACCOUNT_EXISTS,
+      )
```
andiflabs, line 55: Was this change intentional? This introduces a concurrency problem (two concurrent calls with the same `name` may both pass the `accountExists` test and then both call `createAccount`, resulting in one call overriding the other without giving any hint of what occurred to the user)

patnir: unintentional, reverting it. merge got messed up somehow

PR #4735, opened 2024-02-14
adds walletdb multisigSecrets

hughy:
## Summary

adds a datastore to the walletDb to store each 'ParticipantSecret' that a multisig user generates

uses a string 'name' as the key and stores the secret as a buffer as the datastore value

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


jowparks (approved):

jowparks:
Seems low risk since this is unused. Will be interested to see if this store is suffcient or if we will need more info. To be safe we could add an encoder where the first value is the secret, otherwise this store would need a migration to be updated.

hughy:
> Seems low risk since this is unused. Will be interested to see if this store is suffcient or if we will need more info. To be safe we could add an encoder where the first value is the secret, otherwise this store would need a migration to be updated.

since we'll need a migration either way I think I'll leave this as just a string for now 😁

PR #4736, opened 2024-02-14
standardizing how the account flag is used in our CLI

patnir:
## Summary

These are the ways we enter the account in our CLI: 

1. (most common) Using the `-f` flag
2. As an argument (has it's place, for example when creating an account or changing the name of an account)
3. `-a` 
4. `--account`

As a user want the experience to be as consistent as possible. Provide the account with the same flag. If the account is not provided, use the default account.

This change goes through our CLI commands and attempts to make the way we enter the account consistent. 

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/post.ts
+++ b/ironfish-cli/src/commands/wallet/post.ts
@@ -25,8 +25,8 @@ export class PostCommand extends IronfishCommand {
   static flags = {
     ...RemoteFlags,
     account: Flags.string({
+      char: 'a',
       description: 'The account that created the raw transaction',
-      required: false,
     }),
     confirm: Flags.boolean({
       default: false,
--- a/ironfish-cli/src/commands/wallet/multisig/create-signature-share.ts
+++ b/ironfish-cli/src/commands/wallet/multisig/create-signature-share.ts
@@ -14,7 +14,7 @@ export class CreateSignatureShareCommand extends IronfishCommand {
   static flags = {
     ...RemoteFlags,
     account: Flags.string({
-      char: 'f',
+      char: 'a',
       description: 'The account from which the signature share will be created',
       required: false,
     }),
--- a/ironfish-cli/src/commands/wallet/multisig/create-signing-commitment.ts
+++ b/ironfish-cli/src/commands/wallet/multisig/create-signing-commitment.ts
@@ -12,7 +12,7 @@ export class CreateSigningCommitmentCommand extends IronfishCommand {
   static flags = {
     ...RemoteFlags,
     account: Flags.string({
-      char: 'f',
+      char: 'a',
       description:
         'The account to use for generating the commitment, must be a multisig participant account',
       required: false,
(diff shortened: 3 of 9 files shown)


ironfish-cli/src/commands/chain/genesisadd.ts
```diff
@@ -21,7 +21,7 @@ export default class GenesisAddCommand extends IronfishCommand {
   static flags = {
     ...LocalFlags,
     account: Flags.string({
-      char: 'a',
+      char: 'f',
```
patnir, line 24: is this a breaking change for another service we own?

ironfish-cli/src/commands/chain/genesisblock.ts
```diff
@@ -25,7 +25,7 @@ export default class GenesisBlockCommand extends IronfishCommand {
   static flags = {
     ...LocalFlags,
     account: Flags.string({
-      char: 'a',
+      char: 'f',
```
patnir, line 28: is this a breaking change for a service we use?

ironfish-cli/src/commands/wallet/send.ts
```diff
@@ -32,7 +32,7 @@ export class Send extends IronfishCommand {
     ...RemoteFlags,
     account: Flags.string({
       char: 'f',
-      description: 'The account to send money from',
+      description: 'The account to send coins from',
```
patnir, line 35: coins is more consistent with the description of the command and of other fields in the command

ironfish-cli/src/commands/wallet/post.ts
```diff
@@ -26,7 +26,7 @@ export class PostCommand extends IronfishCommand {
     ...RemoteFlags,
     account: Flags.string({
       description: 'The account that created the raw transaction',
-      required: false,
```
patnir, line 29: by default not required

danield9tqh:
What does `f` stand for? Is that short for `from`? Seems like that makes sense in some contexts but not others when passing an account. `-a` seems more universal. was there a reason for going with `f` instead?

hughy (commented):
I think I might actually prefer to go the opposite way and move towards deprecating single character flags like `-f` and use the full `--account`: it's only intuitive to me that `f` would be an account for `send -f(rom)`

patnir:
> What does `f` stand for? Is that short for `from`? Seems like that makes sense in some contexts but not others when passing an account. `-a` seems more universal. was there a reason for going with `f` instead?

I wanted to keep `send` as is since it's probably the most used command and conform everything to that. I agree `a` is more intuitive

patnir:
> I think I might actually prefer to go the opposite way and move towards deprecating single character flags like `-f` and use the full `--account`: it's only intuitive to me that `f` would be an account for `send -f(rom)`

@hughy i feel like single letter is a common pattern for many CLI interfaces, so i personally don't like the idea of deprecating single characters. I agree that `--account` is better and more intuitive. Maybe changing everything to `-a` would be better like @danield9tqh mentioned above?

jowparks:
My vote is for keeping -f with deprecation warning, and making -a the universal. I think having single letter flags is useful for power users, having `--account` is great for newer users.

patnir:
I reverted the change to `send`, `mint`, and `burn` because the `-a` flag conflicts with the `-a` flag for amount. I do believe the remaining changes are still relevant and I'll think of what the best approach for the other commands is.

PR #4737, opened 2024-02-15
adds wallet multisig rpc to create secret

hughy:
## Summary

provides an interface for generating a random ParticipantSecret and storing it in the walletDb with a given name

updates wallet:multisig:identity:create to use the new RPC. perhaps we should rename to 'wallet:multisig:secret:create'?

## Testing Plan

- ran `wallet:multisig:identity:create`

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/wallet/multisig/identity/create.ts
```diff
@@ -2,17 +2,32 @@
...
   static hidden = true
 
-  start(): void {
-    // TODO: generate secret over RPC, persist in walletDb
-    const secret = ParticipantSecret.random()
+  static flags = {
+    ...RemoteFlags,
+    name: Flags.string({
+      char: 'n',
+      description: 'Name to use for the coordinator',
```
andiflabs, line 17: Nit: I'm not sure how the coordinator is relevant here. This name is just a human-friendly identifier that can be shared with anyone, not just the coordinator

ironfish/src/rpc/routes/wallet/multisig/createSecret.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+export const CreateSecretRequestSchema: yup.ObjectSchema<CreateSecretRequest> = yup
+  .object({
+    name: yup.string().defined(),
+  })
+  .defined()
+
+export const CreateSecretResponseSchema: yup.ObjectSchema<CreateSecretResponse> = yup
+  .object({
+    secret: yup.string().defined(),
```
andiflabs, line 25: Why returning the secrets? Ideally secrets (in any shape or form) should never escape the wallet. I'd return the public identity only

ironfish-cli/src/commands/wallet/multisig/identity/create.ts
hughy: oops, this was a copy-paste

ironfish/src/rpc/routes/wallet/multisig/createSecret.ts
hughy: I'll change to return only the identity. On that note it probably makes sense to rename to 'createIdentity'

ironfish-cli/src/commands/wallet/multisig/identity/create.ts
```diff
@@ -1,18 +1,56 @@
...
+          this.log()
+          this.log(e.codeMessage)
+        }
+
+        const name = await CliUx.ux.prompt('Enter a name for the identity', {
+          required: true,
+        })
+        if (name === flags.name) {
+          this.error(`Entered the same name: '${name}'`)
+        }
```
andiflabs, line 45: Nit: this is a bit redundant because, after displaying this message, it will still retry to call `createIdentity`, which will result in another error message to be printed. Maybe you meant to insert a `continue` after `this.error()`? (I don't have any strong opinion for any particular behavior)

ironfish-cli/src/commands/wallet/multisig/identity/create.ts
```diff
@@ -1,18 +1,56 @@
...
+        flags.name = name
+        continue
+      }
+    }
+
+    const identity = new ParticipantIdentity(Buffer.from(response.content.identity, 'hex'))
 
-    const identity = secret.toIdentity()
     this.log('Identity:')
     this.log(identity.serialize().toString('hex'))
```
andiflabs, line 55: Nit: this is converting a hex string to a `ParticipantIdentity` then back to hex. I think you can print `response.content.identity` directly (unless the two hex strings are different, but I don't see an obvious reason why it'd be the case)

ironfish/src/rpc/routes/wallet/multisig/createIdentity.ts
```diff
@@ -0,0 +1,53 @@
...
+
+routes.register<typeof CreateIdentityRequestSchema, CreateIdentityResponse>(
+  `${ApiNamespace.wallet}/multisig/createIdentity`,
+  CreateIdentityRequestSchema,
+  async (request, context): Promise<void> => {
+    AssertHasRpcContext(request, context, 'wallet')
+
+    const { name } = request.data
+
+    if (await context.wallet.walletDb.hasMultisigSecret(name)) {
```
andiflabs, line 38: Is this thread-safe (serializable consistency)? If I make two RPC calls to `createIdentity` with the same name at the same time, is it possible that both calls to `hasMultisigSecret` return false, hence resulting in the creation of 2 secrets, one of which will override the other?

hughy: these calls should be wrapped in a database transaction

andiflabs (approved):

PR #4743, opened 2024-02-16
Change Note memos from string to 32 byte buffer

NullSoldier:
## Summary

This changes a few places to allow memo to be passed as a Buffer instead of a string. The reasoning for this is that underneath, memo is stored as a 32 bit buffer. By accepting strings it's unclear what's happening to string based memos. This makes it more clear that it's got a byte limit, not a character limit when the user has to pass in a Buffer.

- RPC route `wallet/createTransaction` now has memoHex which takes a hex encoded buffer
- RPC route `wallet/sendTransaction` now has memoHex which takes a hex encoded buffer
- RPC routes `createTransaction()` and `sendTransaction()` no longer require a memo
- `Wallet.send()` now requires output's memo to be a Buffer
- `Wallet.createTransaction()` now requires output's memo to be a Buffer
- NativeNote::new now allows passing in a Buffer to memo instead just a string

## Testing Plan

Created new tests

## Documentation

```
[X] Yes
```

- [x] Update RPC wallet/createTransaction
- [x] Update RPC wallet/sendTransaction

## Breaking Change

```
[X] Yes
```

Anyone using the SDK now will need to update to pass in buffers.

### DIFF
--- a/ironfish/src/primitives/rawTransaction.test.slow.ts
+++ b/ironfish/src/primitives/rawTransaction.test.slow.ts
@@ -46,7 +46,7 @@ function createTestRawTransaction(
     new NativeNote(
       account.publicAddress,
       123456789n,
-      'some memo',
+      Buffer.from('some memo'),
       Asset.nativeId(),
       generateKey().publicAddress,
     ).serialize(),
@@ -102,7 +102,7 @@ function createTestRawTransaction(
       new NativeNote(
         account.publicAddress,
         123456789n,
-        'some memo',
+        Buffer.from('some memo'),
         TEST_ASSET_ID_1,
         generateKey().publicAddress,
       ).serialize(),
@@ -112,7 +112,7 @@ function createTestRawTransaction(
       new NativeNote(
         account.publicAddress,
         123456789n,
-        'some memo',
+        Buffer.from('some memo'),
         TEST_ASSET_ID_2,
         generateKey().publicAddress,
       ).serialize(),
@@ -130,7 +130,7 @@ function createTestRawTransaction(
       new NativeNote(
         generateKey().publicAddress,
         123456789n - raw.fee,
-        'some memo',
+        Buffer.from('some memo'),
         Asset.nativeId(),
         account.publicAddress,
       ).serialize(),
--- a/ironfish/src/rpc/routes/wallet/createTransaction.ts
+++ b/ironfish/src/rpc/routes/wallet/createTransaction.ts
@@ -24,7 +24,8 @@ export type CreateTransactionRequest = {
   outputs: {
     publicAddress: string
     amount: string
-    memo: string
+    memo?: string
+    memoHex?: string
     assetId?: string
   }[]
   mints?: {
@@ -59,7 +60,8 @@ export const CreateTransactionRequestSchema: yup.ObjectSchema<CreateTransactionR
           .object({
             publicAddress: yup.string().defined(),
             amount: YupUtils.currency({ min: 1n }).defined(),
-            memo: yup.string().defined().max(MEMO_LENGTH),
+            memo: yup.string().optional().max(MEMO_LENGTH),
+            memoHex: yup.string().length(MEMO_LENGTH * 2, 'Must be 32 byte hex encoded'),
             assetId: yup.string().optional(),
           })
           .defined(),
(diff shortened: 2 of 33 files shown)


ironfish/src/rpc/routes/wallet/createTransaction.ts
```diff
@@ -122,10 +124,16 @@ routes.register<typeof CreateTransactionRequestSchema, CreateTransactionResponse
       params.outputs = []
 
       for (const output of request.data.outputs) {
+        const memo = output.memo
+          ? Buffer.from(output.memo)
+          : output.memoHex
+          ? Buffer.from(output.memoHex, 'hex')
+          : Buffer.alloc(32, 0)
```
jowparks, line 131: We should throw if both are null right? This Buffer.alloc would lead to a cryptic error in the case of both being null.

ironfish/src/rpc/routes/wallet/createTransaction.ts
```diff
@@ -59,7 +60,8 @@ export const CreateTransactionRequestSchema: yup.ObjectSchema<CreateTransactionR
           .object({
             publicAddress: yup.string().defined(),
             amount: YupUtils.currency({ min: 1n }).defined(),
-            memo: yup.string().defined().max(MEMO_LENGTH),
+            memo: yup.string().optional().max(MEMO_LENGTH),
+            memoHex: yup.string().max(MEMO_LENGTH * 2, 'Must be 32 byte hex encoded'),
```
jowparks, line 64: missing `.optional`

ironfish/src/rpc/routes/wallet/createTransaction.ts
hughy: I think the empty Buffer should be fine here -- the resulting memo would be the same as if the user submitted an empty string for the memo.

I'll add a test to cover this case

andiflabs: Related: what if both are non-null/non-empty? Shouldn't we throw in that case?

hughy: > So it is acceptable to send no memo at all? Because previously memo was a required RPC field (albeit could be sent with empty string).

I think it makes for a nicer interface not to have to provide an empty string, but I'm open to reverting it back to a required field

hughy: > Related: what if both are non-null/non-empty? Shouldn't we throw in that case?

I think we should throw in that case -- the client shouldn't set both

NullSoldier: The change here is to make the RPC not required to take a memo. It's not a necessary input to creating a transaction and was really just rude to require users to put as an input.

jowparks: Ok that seems reasonable to me, this is just a change in behavior that isn’t listed in PR summary

ironfish/src/primitives/rawTransaction.test.slow.ts
```diff
@@ -46,7 +46,7 @@ function createTestRawTransaction(
     new NativeNote(
       account.publicAddress,
       123456789n,
-      'some memo',
+      Buffer.from('some memo'),
```
jowparks, line 49: This is sort of misleading right? Shouldn’t the buffer be from a hex encoded string?

jowparks: Never mind I misread this, it will be a satisfactory buffer.

ironfish/src/rpc/routes/wallet/sendTransaction.ts
```diff
@@ -85,12 +90,27 @@ routes.register<typeof SendTransactionRequestSchema, SendTransactionResponse>(
...
 
-    const outputs = request.data.outputs.map((output) => ({
-      publicAddress: output.publicAddress,
-      amount: CurrencyUtils.decode(output.amount),
-      memo: output.memo,
-      assetId: output.assetId ? Buffer.from(output.assetId, 'hex') : Asset.nativeId(),
-    }))
+    const outputs = request.data.outputs.map((output) => {
+      if (output.memo && output.memoHex) {
+        throw new RpcValidationError('Only one of memo or memoHex may be set for each output')
```
jowparks, line 95: Nit: more descriptive error about what was set for both?

jowparks (approved):

PR #4745, opened 2024-02-16
feat(ironfish,rust,rust-nodejs): Return descriptions from unsigned transaction

rohanjadvani:
## Summary

* Adds getter methods to the unsigned transaction struct
* Adds a `descriptions` method on the Napi layer to fetch parts of an unsigned transaction
* Adds a method to the JS primitive to fetch descriptions

## Testing Plan

Unit test

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/primitives/unsignedTransaction.ts
+++ b/ironfish/src/primitives/unsignedTransaction.ts
@@ -2,7 +2,13 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-import { UnsignedTransaction as NativeUnsignedTransaction } from '@ironfish/rust-nodejs'
+import {
+  NativeBurnDescription,
+  NativeMintDescription,
+  UnsignedTransaction as NativeUnsignedTransaction,
+} from '@ironfish/rust-nodejs'
+import { Note } from './note'
+import { NoteEncrypted } from './noteEncrypted'
 
 export class UnsignedTransaction {
   private readonly unsignedTransactionSerialized: Buffer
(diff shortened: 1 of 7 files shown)


ironfish/src/primitives/unsignedTransaction.ts
```diff
@@ -55,4 +61,40 @@ export class UnsignedTransaction {
...
+
+      const receivedNote = note.decryptNoteForOwner(incomingViewKey)
+      if (receivedNote) {
+        receivedNotes.push(receivedNote)
+      }
+
+      const spentNote = note.decryptNoteForSpender(outgoingViewKey)
+      if (spentNote) {
+        spentNotes.push(spentNote)
+      }
```
hughy, line 90: Wondering if we'll want/need to handle notes showing up in both received and spent when we render

ironfish/src/primitives/unsignedTransaction.ts
```diff
@@ -55,4 +61,40 @@ export class UnsignedTransaction {
 
     return result
   }
+
+  descriptions(
+    incomingViewKey: string,
+    outgoingViewKey: string,
+  ): {
+    receivedNotes: Note[]
+    spentNotes: Note[]
```
hughy, line 70: nit - should this be `sent` instead of `spent`?

hughy (approved):
This looks to me like a good way of getting the information that we need from the existing fields and data on an UnsignedTransaction

ironfish/src/primitives/unsignedTransaction.ts
rohanjadvani: Yeah. I thought I'd just return everything here for now and then the caller can decide what to do.

PR #4753, opened 2024-02-20
feat(ironfish): Add `wallet/getAccountTransactionDescriptions`

rohanjadvani:
## Summary

Add an RPC endpoint that gets the descriptions from an unsigned transaction

## Testing Plan

Unit test

## Documentation

```
[x] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/wallet/getAccountTransactionDescriptions.ts
```diff
@@ -0,0 +1,130 @@
...
+    sender: string
+    value: string
+  }[]
+  sentNotes: {
+    assetId: string
+    memo: string
+    owner: string
+    sender: string
+    value: string
+  }[]
```
patnir, line 40: would it be possible to reuse one or more of our existing RPC objects like this https://ironfish.network/developers/documentation/rpc/objects/rpcWalletTransaction

ironfish/src/rpc/routes/wallet/getAccountTransactionDescriptions.ts
```diff
@@ -0,0 +1,130 @@
...
+}
+
+export const GetAccountTransactionDescriptionsRequestSchema: yup.ObjectSchema<GetAccountTransactionDescriptionsRequest> =
```
jowparks, line 43: This isn't getting the descriptions, this is decrypting the unsigned transaction. Technically I think it should be able to take in a signed or unsigned transaction as well.

jowparks: Aren't descriptions supposed to be the publically viewable contents of a transaction? For mints/burns they just happen to be public in general. The notes are private

rohanjadvani: It's a mix of both I guess. It's returning the mint/burn descriptions but also decrypting the notes associated with the output descriptions on the unsigned transaction.

The primary use case here is just to be for the unsigned transaction to show more details. But if you think it's useful to expand this, I can rework it or rename it so it's more explicit for unsigned transactions.

PR #4759, opened 2024-02-21
Rahul/use public key package with identities

patnir:
## Summary

Uses the public key package in the ironfish-frost library instead of the one provided by frost. 
This new public key package contains the identities + frost public key package

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust-nodejs/src/structs/transaction.rs
+++ b/ironfish-rust-nodejs/src/structs/transaction.rs
@@ -8,7 +8,6 @@ use std::collections::BTreeMap;
 use std::convert::TryInto;
 
 use ironfish::assets::asset_identifier::AssetIdentifier;
-use ironfish::frost::keys::PublicKeyPackage;
 use ironfish::frost::round1::SigningCommitments;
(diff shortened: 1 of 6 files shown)


ironfish-rust-nodejs/src/structs/transaction.rs
```diff
@@ -521,8 +521,8 @@ pub fn aggregate_signature_shares(
     signing_package_str: String,
     signature_shares_arr: Vec<String>,
 ) -> Result<Buffer> {
-    let public_key_package = PublicKeyPackage::deserialize(
-        &hex_to_vec_bytes(&public_key_package_str).map_err(to_napi_err)?,
+    let public_key_package = PublicKeyPackage::read(
+        &hex_to_vec_bytes(&public_key_package_str).map_err(to_napi_err)?[..],
```
hughy, line 525: I think this implies that each account will store the public key package as a full ironfish-frost `PublicKeyPackage` rather than a frost `PublicKeyPackage`

That seems fine to me, but wanted to note it

hughy (approved):

PR #4761, opened 2024-02-21
saves participant identities on account import

hughy:
## Summary

adds each participant identity from a PublicKeyPackage to the participantIdentities store when importing a multisig account

adds napi bindings to deserialize a PublicKeyPackage and return the list of identities as buffers

## Testing Plan

adds slow unit test to verify that identities are saved when importing multisig accounts

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-rust-nodejs/src/frost.rs
```diff
@@ -237,3 +237,34 @@ pub fn split_secret(
...
+        let mut identities: Vec<Buffer> = vec![];
+
+        for identity in &self.public_key_package.identities {
+            let mut vec: Vec<u8> = vec![];
+            identity.serialize_into(&mut vec).map_err(to_napi_err)?;
+
+            identities.push(Buffer::from(vec))
+        }
+
+        Ok(identities)
```
andiflabs, line 268: I believe this can be rewritten in a way that is infallible and reduces the number of allocations (untested):

```rust
self.public_key_packages.identities.iter().map(|id| Buffer:from(&identity.serialize()[..])).collect()
```

andiflabs (approved):

ironfish-rust-nodejs/src/frost.rs
```diff
@@ -237,3 +237,30 @@ pub fn split_secret(
...
+        let bytes = hex_to_vec_bytes(&value).map_err(to_napi_err)?;
+
+        let public_key_package = PublicKeyPackage::read(&bytes[..]).map_err(to_napi_err)?;
+
+        Ok(NativePublicKeyPackage { public_key_package })
+    }
+
+    #[napi]
+    pub fn identities(&self) -> Result<Vec<Buffer>> {
+        Ok(self
```
andiflabs, line 259: Because this is now infallible, `Result<Vec<Buffer>>` can be changed to `Vec<Buffer>`, and the `Ok` can be dropped

PR #4770, opened 2024-02-23
Add SDK exports to repl

danield9tqh:
## Summary
For faster debugging in the repl command expose all the SDK exported functions

## Testing Plan
Tested running repl locally

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/repl.ts
+++ b/ironfish-cli/src/commands/repl.ts
@@ -1,6 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import * as IronfishSDK from '@ironfish/sdk'
 import { ALL_API_NAMESPACES, NodeUtils, RpcMemoryClient } from '@ironfish/sdk'
 import { Flags } from '@oclif/core'
 import fs from 'fs/promises'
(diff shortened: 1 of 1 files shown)


hughy (approved):

ironfish-cli/src/commands/repl.ts
```diff
@@ -1,6 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import * as IronfishSDK from '@ironfish/sdk'
```
jowparks, line 4: Can we just import as `sdk`?

jowparks: Ironfish repl implies it is ironfish, no? Also this is consistent with how it is used in the node code.

ironfish-cli/src/commands/repl.ts
```diff
@@ -54,13 +55,17 @@ export default class Repl extends IronfishCommand {
...
+    this.log('')
 
     const historyPath = path.join(node.config.tempDir, 'repl_history.txt')
     await fs.mkdir(node.config.tempDir, { recursive: true })
     this.log(`Storing repl history at ${historyPath}`)
     this.log('Type .exit or press CTRL+C to quit')
 
     const server = repl.start('> ')
+    server.context.IronfishSDK = IronfishSDK
     server.context.sdk = this.sdk
```
jowparks, line 69: Why is this not accessible via `this.sdk`?

danield9tqh: `this.sdk` is an instance of the sdk not all the exported functions

jowparks: Why do the two have to be different? Why can't we access the exported functions from an instance?

danield9tqh: [This](https://github.com/iron-fish/ironfish/blob/master/ironfish/src/sdk.ts) is the SDK class. [This](https://www.npmjs.com/package/@ironfish/sdk) is the SDK library. Different things

PR #4777, opened 2024-02-26
fix: throw on max buffer length exceeded

jowparks:
## Summary
Max buffer length is currently not enforced in `createTransaction`. I ran into this when passing in a string that is longer than 32 bytes. The code happily truncated the memo without error or warning.
## Testing Plan
Tests pass
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/wallet.test.ts
+++ b/ironfish/src/wallet/wallet.test.ts
@@ -1015,6 +1015,41 @@ describe('Wallet', () => {
       )
     })
 
+    it.only('should throw error if memo is too long', async () => {
+      const { node } = nodeTest
+
+      const accountA = await useAccountFixture(node.wallet, 'a')
+
+      const blockA1 = await useMinerBlockFixture(node.chain, undefined, accountA, node.wallet)
+      await expect(node.chain).toAddBlock(blockA1)
+
+      await node.wallet.updateHead()
+
+      const transaction = blockA1.minersFee
+
+      const transactionValue = await accountA.getTransaction(transaction.hash())
+      Assert.isNotUndefined(transactionValue)
+      Assert.isNotNull(transactionValue.sequence)
+
+      const rawTransaction = node.wallet.createTransaction({
+        account: accountA,
+        fee: 1n,
+        outputs: [
+          {
+            publicAddress: '0d804ea639b2547d1cd612682bf99f7cad7aad6d59fd5457f61272defcd4bf5b',
+            amount: 10n,
+            memo: Buffer.from('This is a string with 33 bytes!!!'),
+            assetId: Asset.nativeId(),
+          },
+        ],
+        expiration: 0,
+      })
+
+      await expect(rawTransaction).rejects.toThrow(
+        'Memo is too long, max byte length is 32, provided memo is 33 bytes. Memo: This is a string with 33 bytes!!!',
+      )
+    })
+
     it('should create raw transaction with fee rate', async () => {
       const { node } = nodeTest
 
--- a/ironfish/src/wallet/wallet.test.ts
+++ b/ironfish/src/wallet/wallet.test.ts
@@ -1,7 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { Asset, generateKey } from '@ironfish/rust-nodejs'
+import { Asset, generateKey, MEMO_LENGTH } from '@ironfish/rust-nodejs'
 import { BufferMap, BufferSet } from 'buffer-map'
 import { v4 as uuid } from 'uuid'
 import { Assert } from '../assert'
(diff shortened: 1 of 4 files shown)


ironfish/src/wallet/wallet.ts
```diff
@@ -991,6 +992,16 @@ export class Wallet {
       )
     }
 
+    for (const output of options.outputs ?? []) {
+      if (output.memo.byteLength > MEMO_LENGTH) {
+        throw new Error(
+          `Memo is too long, max byte length is ${MEMO_LENGTH}, provided memo is ${
+            output.memo.byteLength
+          } bytes. Memo: ${output.memo.toString('utf8')}`,
```
dguenther, line 1000: Since we're taking memos as buffers in places now, I wonder if we should display it as something other than utf8, like hex... Or what do you think about instead saying something like `The memo fields on the following outputs are more than 32 bytes: 1, 2, ....`

Also we should probably throw this as a breaking change in the release notes draft

ironfish/src/wallet/wallet.ts
```diff
@@ -991,6 +992,16 @@ export class Wallet {
       )
     }
 
+    for (const output of options.outputs ?? []) {
```
NullSoldier, line 995: wrap this in an if statement for options.outputs don't allocate an empty array

ironfish/src/wallet/wallet.ts
```diff
@@ -991,6 +992,16 @@ export class Wallet {
...
+      if (output.memo.byteLength > MEMO_LENGTH) {
+        throw new Error(
```
NullSoldier, line 997: Make a new error type for this so it can be caught in the RPC in the future. Don't use `Error`

ironfish/src/wallet/wallet.ts
andiflabs: You could display the both the hex and the decoded string (if it is a valid utf-8 string). Examples

valid utf-8 string:

```
Memo: 74686973206d656d6f206973206c6f6e676572207468616e2033322063686172616374657273 ("this memo is longer than 32 characters")
```

invalid utf-8 string:

```
Memo: aa006cd60a64ae8b3ab4c80c579f47ed6648a2dcfa7b2fcee56d0ac7c9b229006b4b27a9daa5
```

We could also truncate long strings.

Regardless, if we plan to display user-inputted strings, I suggest we encode it to JSON first, so that special characters (like new line characters or ANSI escape sequences) do not result in weird output.

jowparks: We are currently truncating, but I don't really like it because it is easy for someone to not catch it is happening if they aren't paying attention to logs/ouputs.

I will with Derek's option, because if there are multiple error outputs then the message could get long and noisy. I think I would rather know which are the problematic outputs rather than have to recreate transaction having modified one of the outputs just to find that another of my outputs was a problem.

ironfish/src/wallet/wallet.ts
```diff
@@ -991,6 +992,22 @@ export class Wallet {
...
+    if (options.outputs) {
+      const errorMemos: number[] = []
+      for (let i = 0; i < options.outputs.length; i++) {
+        if (options.outputs[i].memo.byteLength > MEMO_LENGTH) {
+          errorMemos.push(i)
+        }
+      }
+      if (errorMemos.length > 0) {
+        throw new MaxMemoLengthError(
+          `The memo fields on the following outputs are more than the max of ${MEMO_LENGTH} bytes: ${errorMemos.join(
```
NullSoldier, line 1004: Generate this message inside of MaxMemoLengthError

ironfish/src/wallet/wallet.ts
```diff
@@ -991,6 +992,22 @@ export class Wallet {
       )
     }
 
+    if (options.outputs) {
+      const errorMemos: number[] = []
```
NullSoldier, line 996: This looks fine, but I personally wouldn't accumulate errors. One error is enough to make the caller realize they aren't validating the input of byte lengths which is a fix they need to make in their code.

jowparks: I am fine with that route too, will allow us to output message too

ironfish/src/wallet/wallet.test.ts
```diff
@@ -1015,6 +1015,41 @@ describe('Wallet', () => {
...
+            amount: 10n,
+            memo: Buffer.from('This is a string with 33 bytes!!!'),
+            assetId: Asset.nativeId(),
+          },
+        ],
+        expiration: 0,
+      })
+
+      await expect(rawTransaction).rejects.toThrow(
+        'The memo fields on the following outputs are more than the max of 32 bytes: 0',
```
NullSoldier, line 1049: Check the error type here with `toThrow(MaxMemoLengthError)` and don't check the message

ironfish/src/wallet/wallet.test.ts
```diff
@@ -1015,6 +1015,41 @@ describe('Wallet', () => {
...
+      )
+    })
```
NullSoldier, line 1051: I think this whole test can be made a lot shorter. There's some cruft from copy pasting tests.

```suggestion
    it('should throw error if memo is too long', async () => {
      const account = await useAccountFixture(nodeTest.node.wallet, 'a')

      const promise = nodeTest.wallet.createTransaction({
        account: account,
        fee: 1n,
        outputs: [
          {
            publicAddress: account.publicAddress,
            amount: 1n,
            memo: Buffer.alloc(MEMO_LENGTH + 1),
            assetId: Asset.nativeId(),
          },
        ],
        expiration: 0,
      })

      await expect(promise).rejects.toThrow(MaxMemoLengthError)
    })
```

NullSoldier (approved):

PR #4784, opened 2024-02-28
Multisig wallet: generate TDK accounts on the server side

andiflabs:
## Summary

The `wallet/multisig/createTrustedDealerKeyPackage` RPC now creates base64 accounts directly, instead of delegating this operation to the CLI. This change is a pre-requisite for multisig account encryption.

## Testing Plan

Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/multisig/createTrustedDealerKeyPackage.ts
+++ b/ironfish/src/rpc/routes/wallet/multisig/createTrustedDealerKeyPackage.ts
@@ -3,6 +3,9 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { generateKey, splitSecret } from '@ironfish/rust-nodejs'
 import * as yup from 'yup'
+import { Assert } from '../../../../assert'
+import { FullNode } from '../../../../node'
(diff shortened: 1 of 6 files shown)


ironfish/src/rpc/routes/wallet/multisig/createTrustedDealerKeyPackage.ts
```diff
@@ -12,53 +15,47 @@ export type CreateTrustedDealerKeyPackageRequest = {
...
+  publicAddress: string
+  publicKeyPackage: string
   viewKey: string
   incomingViewKey: string
   outgoingViewKey: string
-  publicAddress: string
-  keyPackages: Array<{ identity: string; keyPackage: string }>
-  publicKeyPackage: string
+  proofAuthorizingKey: string
+  accounts: Array<{ identity: string; account: string }>
```
patnir, line 26: would participants be a better name?

andiflabs: Instead of `accounts`? I initially named it `participantAccounts` but then renamed it to `accounts` because the former looked too long and verbose, but I can revert it back if you think it's clearer. Naming it just `participants` is too ambiguous IMO. What do you think?

PR #4786, opened 2024-02-28
Add functionality to Base64JsonEncoder to encrypt multisig accounts

andiflabs:
## Summary

Multisig accounts can be encrypted as follows:

```ts
const encoder = new Base64JsonEncoder()
encoder.encode(account, {
  encryptWith: { kind: 'MultisigIdentity', identity: ... })
```

and decrypted as follows:

```ts
decodeAccount(encodedAccount, { multisigSecret: ... })
```

Base64 multisig encrypted accounts differ from regular base64 encoded accounts from the prefix: `ifaccount` vs `ifmsaccount`.

## Testing Plan

Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/account/encoder/base64json.test.ts
+++ b/ironfish/src/wallet/account/encoder/base64json.test.ts
@@ -1,14 +1,41 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { generateKey } from '@ironfish/rust-nodejs'
+import { generateKey, ParticipantSecret } from '@ironfish/rust-nodejs'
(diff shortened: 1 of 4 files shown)


rohanjadvani (commented):
I think we should probably add some test cases for this new format under `/rpc/routes/wallet/__importTestCases__`, either in this PR or a follow-up

andiflabs:
> I think we should probably add some test cases for this new format under `/rpc/routes/wallet/__importTestCases__`, either in this PR or a follow-up

I was actually wondering if we had this kind of tests somewhere, thanks for pointing me to them

andiflabs:
Those tests are for the RPC endpoint. The RPC endpoint does not support decryption yet, that'll come in my next PR. The tests will be added in that PR

rohanjadvani (approved):

PR #4787, opened 2024-02-28
Quit scanning without throwing if the wallet is closed during a scan

andiflabs:
## Summary

If the `Wallet` is closed while `scanTransactions` is running, it may fail with an exception. This problem has been causing unit tests to fail spuriously with `ReadError: Database is not open`.

This commit changes the behavior of `scanTransactions` to catch errors during scan and check whether the wallet is still open: if closed, the exception is suppressed, otherwise the behavior is unchanged.

## Testing Plan

Unit tests. In particular, checked that failures often seen in PR #4784 no longer occur.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -771,39 +771,47 @@ export class Wallet {
     // but setting this.scan is our lock so updating the head doesn't run again
     await this.updateHeadState?.wait()
 
-    const startHash = fromHash ?? (await this.getEarliestHeadHash())
+    try {
+      const startHash = fromHash ?? (await this.getEarliestHeadHash())
 
-    // Fetch current chain head sequence
-    const chainHead = await this.getChainHead()
-    scan.endSequence = chainHead.sequence
+      // Fetch current chain head sequence
+      const chainHead = await this.getChainHead()
+      scan.endSequence = chainHead.sequence
 
-    this.logger.info(`Scan starting from block ${startHash?.toString('hex') ?? 'null'}`)
+      this.logger.info(`Scan starting from block ${startHash?.toString('hex') ?? 'null'}`)
 
-    const scanProcessor = new RemoteChainProcessor({
-      logger: this.logger,
-      nodeClient: this.nodeClient,
-      head: startHash,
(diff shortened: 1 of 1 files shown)


NullSoldier:
https://github.com/iron-fish/ironfish/pull/4787/files?diff=split&w=1

ironfish/src/wallet/wallet.ts
```diff
@@ -771,39 +771,47 @@ export class Wallet {
...
+      // Update chainProcessor following scan
+      this.chainProcessor.hash = scanProcessor.hash
+      this.chainProcessor.sequence = scanProcessor.sequence
+    } catch (e) {
+      if (!this.isOpen) {
+        this.logger.info('Scan aborted (wallet closed while scanning)')
+        return
+      }
+      throw e
+    }
```
NullSoldier, line 814: I just need a little more info about this change. How can the wallet be closed without `wallet.stop()` being called which should abort the scan and prevent errors?

I guess the fact that errors are occuring here because the wallet is closed is a bug I'd like to track down as well and could lead to this code being deleted later unless it's just too hard to correlate the two.

I think this is a good pragmatic fix in the mean time but I am interested in finding information about the root bug too.

NullSoldier:
I just need a little more info about this change. How can the wallet be closed without wallet.stop() being called which should abort the scan and prevent errors?

I guess the fact that errors are occuring here because the wallet is closed is a bug I'd like to track down as well and could lead to this code being deleted later unless it's just too hard to correlate the two.

I think this is a good pragmatic fix in the mean time but I am interested in finding information about the root bug too.

andiflabs:
Calling stop() inside close() does not fix the issue. I guess the problem is that stop() just sets a flag that says "you need to stop" and that needs to be explicitly checked, but does not actually abort the execution of the coroutine.

andiflabs:
Another piece of information: start() actually never gets called, so calling stop() has no effect because `isStarted` is never set to true.

NullSoldier:
I added a fix for some of these issues here. https://github.com/iron-fish/ironfish/pull/4791/files

PR #4788, opened 2024-02-28
creating new signing key when splitting secret from dealer

patnir:
## Summary

Problem: The trusted dealer provides the spending key to create the package. This means that the trusted dealer has the capability to sign the transaction on their own terms without the participants being involved. 

Solution: We generate a separate key that is discarded while partially creating the package (it generates the nullifier key and outgoing view key). This means that the trusted dealer no longer will be able to sign the transaction and only all the participants can do so. 

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/frost_utils/split_secret.rs
+++ b/ironfish-rust/src/frost_utils/split_secret.rs
@@ -14,26 +14,19 @@ use ironfish_frost::{
 use rand::{CryptoRng, RngCore};
 use std::collections::HashMap;
 
-use crate::errors::{IronfishError, IronfishErrorKind};
+use crate::errors::IronfishError;
+use crate::SaplingKey;
 
 pub struct SecretShareConfig {
     pub min_signers: u16,
     pub identities: Vec<Identity>,
-    pub secret: Vec<u8>,
+    pub spender_key: SaplingKey,
 }
 
 pub(crate) fn split_secret<R: RngCore + CryptoRng>(
     config: &SecretShareConfig,
     mut rng: R,
 ) -> Result<(HashMap<Identity, KeyPackage>, PublicKeyPackage), IronfishError> {
-    let secret_bytes: [u8; 32] = config
-        .secret
-        .clone()
-        .try_into()
-        .map_err(|_| IronfishError::new(IronfishErrorKind::InvalidSecret))?;
-
-    let secret_key = SigningKey::deserialize(secret_bytes)?;
-
     let mut frost_id_map = config
         .identities
         .iter()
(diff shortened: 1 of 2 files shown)


rohanjadvani (approved):
This looks good to me. I'd wait for an additional approval from @andiflabs

ironfish-rust/src/frost_utils/split_secret.rs
```diff
@@ -14,26 +14,19 @@ use ironfish_frost::{
...
 
-use crate::errors::{IronfishError, IronfishErrorKind};
+use crate::errors::IronfishError;
+use crate::SaplingKey;
 
 pub struct SecretShareConfig {
     pub min_signers: u16,
     pub identities: Vec<Identity>,
-    pub secret: Vec<u8>,
+    pub coordinator_key: SaplingKey,
```
andiflabs, line 23: Nit: let's get rid of the name "coordinator"

ironfish-rust/src/frost_utils/split_spender_key.rs
```diff
@@ -89,21 +93,11 @@ mod test {
         let trusted_dealer_key_packages =
             split_spender_key(&sapling_key, 5, identities).expect("spender key split failed");
 
-        assert_eq!(
-            trusted_dealer_key_packages.key_packages.len(),
-            10,
-            "should have 10 key packages"
-        );
```
andiflabs, line 96: Why was this assertion removed?

ironfish-rust/src/frost_utils/split_secret.rs
```diff
@@ -77,39 +70,18 @@ mod test {
...
-        };
-
-        let rng = rand::thread_rng();
-        let result = split_secret(&config, rng);
-        assert!(result.is_err());
-        assert!(
-            matches!(result.unwrap_err().kind, IronfishErrorKind::InvalidSecret),
-            "expected InvalidSecret error"
-        );
-    }
```
andiflabs, line 98: Why was this removed?

patnir: because this was removed: 
```
let secret_bytes: [u8; 32] = config
        .secret
        .clone()
        .try_into()
        .map_err(|_| IronfishError::new(IronfishErrorKind::InvalidSecret))?;
```

we can keep the map_err if that is preferred

ironfish-rust/src/frost_utils/split_spender_key.rs
patnir: Added it back, not sure how it got removed. must have made a mistake

ironfish-rust/src/frost_utils/split_secret.rs
patnir: we're no longer passing in a vector. the earlier conversion to the vector was unnecessary. This edge case handled at the napi layer i believe

patnir: https://github.com/iron-fish/ironfish/blob/rahul/ifl-2307-create-new-gsk-when-splitting-secret-from-dealer/ironfish-rust-nodejs/src/frost.rs#L204

andiflabs (approved):

PR #4792, opened 2024-02-29
Ensure that the release binaries are built using the right version

mat-if:
## Summary

We got reports that the latest binaries (v1.20.0) were actually of v1.19.0. Looking into this, I suspect what happened was that the github version was tagged prior to the npm release finishing. This change ensures that ONLY the correct version will be pulled from npm, and the job will fail if the specified version has not yet been published.

Fixes IFL-2308

## Testing Plan

I set up a test repo to emulate the release steps. The github action on release simply tries to download the specified version of ironfish from npm and then run `ironfish status`.
v1.20.0 which succeeded: https://github.com/mat-if/foo/actions/runs/8100633100
v.1.21.0 which does not exist yet, so it errored out: https://github.com/mat-if/foo/actions/runs/8100653355

## Documentation

N/A

## Breaking Change

N/A

### DIFF
--- a/.github/workflows/publish-binaries.yml
+++ b/.github/workflows/publish-binaries.yml
@@ -46,12 +46,12 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: 18
-      
+
       - name: Use go
         uses: actions/setup-go@v4
         with:
           go-version: '1.20.6'
-      
+
       - name: Checkout repository
         uses: actions/checkout@v4
 
@@ -60,7 +60,7 @@ jobs:
         shell: bash
         run: |
           identifier=$(awk 'BEGIN {
-            srand(); 
+            srand();
             chars = "abcdefghijklmnopqrstuvwxyz0123456789";
             for (i = 1; i <= 10; i++) {
                 printf "%s", substr(chars, int(rand() * length(chars)) + 1, 1);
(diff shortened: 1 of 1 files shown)


jowparks (approved):

.github/workflows/publish-binaries.yml
```diff
@@ -76,13 +76,13 @@ jobs:
           cd build
           cp $(node -e "console.log(process.execPath)") ${{ matrix.settings.system != 'windows' && 'node' || 'node.exe' }}
           npm init -y
-          npm install ironfish
+          npm install ironfish@{{ github.event.release.tag_name }}
```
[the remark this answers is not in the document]
jowparks: `npm install ironfish@${{ github.event.release.tag_name }}`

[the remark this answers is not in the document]
mat-if: WOOPS it sure does, good catch

.github/workflows/publish-binaries.yml
```diff
@@ -76,13 +76,13 @@ jobs:
           cd build
           cp $(node -e "console.log(process.execPath)") ${{ matrix.settings.system != 'windows' && 'node' || 'node.exe' }}
           npm init -y
-          npm install ironfish
+          npm install ironfish@${{ github.event.release.tag_name }}
```
danield9tqh, line 79: Will this be `ironfish@v1.20.0` instead of `ironfish@1.20.0`?

danield9tqh: Also if the command didn't work previously because the latest version wasn't published yet would it even be able to find that latest version? Thinking maybe we could fail the job if the latest version on npm doesn't match the tag. `npm view ironfish version` would get latest version looks like

mat-if: Both are valid, surprisingly. I was expecting to have to do some string manipulation, but it ended up working just fine. https://github.com/mat-if/foo/blob/main/.github/workflows/foo.yml#L25

danield9tqh (approved):

PR #4796, opened 2024-02-29
feat(ironfish): Default to identity name when importing

rohanjadvani:
## Summary

Importing an account via the output from the TDK defaults the name to the identity. (see [here](https://github.com/iron-fish/ironfish/blob/staging/ironfish/src/rpc/routes/wallet/multisig/createTrustedDealerKeyPackage.ts#L94)). This is probably not ideal and will lead to confusion.

This updates the behavior so the default behavior is to use the identity name and requires an existing secret to be present.

In a subsequent PR, there will be a more explicit link between the stores instead of just assuming the name is the same.

## Testing Plan

Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -1521,9 +1521,24 @@ export class Wallet {
   }
 
   async importAccount(accountValue: AccountValue): Promise<Account> {
-    if (accountValue.name && this.getAccountByName(accountValue.name)) {
-      throw new DuplicateAccountNameError(accountValue.name)
+    let name = accountValue.name
+    if (accountValue.multisigKeys && isSignerMultisig(accountValue.multisigKeys)) {
+      const identityName = await this.getIdentityName(accountValue.multisigKeys.identity)
+      if (!identityName) {
+        throw new Error('Cannot import identity without an existing secret')
+      }
+
+      if (this.findAccountByIdentity(accountValue.multisigKeys.identity)) {
+        throw new Error('Account already exists with the provided identity')
+      }
+
+      name = identityName
     }
+
+    if (name && this.getAccountByName(name)) {
+      throw new DuplicateAccountNameError(name)
+    }
+
     const accounts = this.listAccounts()
     if (
       accountValue.spendingKey &&
@@ -1535,7 +1550,6 @@ export class Wallet {
     validateAccount(accountValue)
 
     let createdAt = accountValue.createdAt
-    let name = accountValue.name
 
     if (createdAt && !this.nodeClient) {
       this.logger.debug(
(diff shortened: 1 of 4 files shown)


ironfish/src/wallet/wallet.ts
```diff
@@ -1599,6 +1617,31 @@ export class Wallet {
     return this.getAccountByName(name) !== null
   }
 
+  async getIdentityName(identity: string): Promise<string | undefined> {
```
andiflabs, line 1620: Nit: `getMultisigIdentityName`?

Also, I always thought that this "name" was the name of a (partial) account, am I wrong? Is this thing totally separated from the account?

ironfish/src/wallet/wallet.ts
```diff
@@ -1599,6 +1617,31 @@ export class Wallet {
...
+      const secret = new ParticipantSecret(value)
+      if (identity === secret.toIdentity().serialize().toString('hex')) {
+        return name
+      }
+    }
+
+    return undefined
+  }
+
+  findAccountByIdentity(identity: string): Account | undefined {
```
andiflabs, line 1631: Nit: `MultisigIdentity`

ironfish/src/wallet/wallet.ts
```diff
@@ -1519,9 +1521,24 @@ export class Wallet {
   }
 
   async importAccount(accountValue: AccountValue): Promise<Account> {
-    if (accountValue.name && this.getAccountByName(accountValue.name)) {
-      throw new DuplicateAccountNameError(accountValue.name)
+    let name = accountValue.name
+    if (accountValue.multisigKeys && isSignerMultisig(accountValue.multisigKeys)) {
+      const identityName = await this.getIdentityName(accountValue.multisigKeys.identity)
+      if (!identityName) {
+        throw new Error('Cannot import identity without an existing secret')
```
andiflabs, line 1528: "Cannot import _account_"?

ironfish/src/wallet/wallet.ts
```diff
@@ -1519,9 +1521,24 @@ export class Wallet {
...
+      }
+
+      if (this.findAccountByIdentity(accountValue.multisigKeys.identity)) {
```
andiflabs, line 1531: I think this should be guarded by a transaction to avoid concurrency issues?

ironfish/src/wallet/wallet.ts
```diff
@@ -1519,9 +1521,24 @@ export class Wallet {
...
+        throw new Error('Account already exists with the provided identity')
```
andiflabs, line 1532: Is it really a problem to have two different accounts sharing the same multisig identity?

andiflabs:
General comment: I would change all user-visible messages to say "multisig identity" (or similar) instead of just "identity". It's clear to us what these identities are because we've been working on this project for a while, but a user who's not very familiar with it might need more context

ironfish/src/wallet/wallet.ts
rohanjadvani: There's no explicit linking between the two, and there's no guarantee in the code that they are the same. I think that was the intention, but we discovered via live testing it can be broken fairly easily. Will need some sort of join index/store between the two I think.

PR #4814, opened 2024-03-06
preserves order of identities when returning key packages

patnir:
## Summary

Preserves order of identities when returning key packages for each identity

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust-nodejs/src/frost.rs
+++ b/ironfish-rust-nodejs/src/frost.rs
@@ -203,13 +203,19 @@ pub fn generate_and_split_key(
     let identities = try_deserialize_identities(identities)?;
 
     let packages =
-        split_spender_key(&spending_key, min_signers, identities).map_err(to_napi_err)?;
+        split_spender_key(&spending_key, min_signers, identities.clone()).map_err(to_napi_err)?;
 
     let mut key_packages = Vec::with_capacity(packages.key_packages.len());
-    for (identity, key_package) in packages.key_packages.iter() {
+
+    // preserves the order of the identities
+    for identity in identities {
         key_packages.push(ParticipantKeyPackage {
             identity: bytes_to_hex(&identity.serialize()),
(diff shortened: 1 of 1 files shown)


ironfish-rust-nodejs/src/frost.rs
```diff
@@ -203,13 +203,19 @@ pub fn generate_and_split_key(
...
     let mut key_packages = Vec::with_capacity(packages.key_packages.len());
-    for (identity, key_package) in packages.key_packages.iter() {
+
+    // preserves the order of the identities
+    for identity in identities {
         key_packages.push(ParticipantKeyPackage {
             identity: bytes_to_hex(&identity.serialize()),
-            key_package: bytes_to_hex(&key_package.serialize().map_err(to_napi_err)?),
+            key_package: bytes_to_hex(
+                &packages.key_packages[&identity]
```
andiflabs, line 215: Note that this will panic if `identity` is not contained in `key_packages`. If you want a more robust solution, use `.get`

andiflabs (approved):

ironfish-rust-nodejs/src/frost.rs
```diff
@@ -203,13 +203,19 @@ pub fn generate_and_split_key(
     let identities = try_deserialize_identities(identities)?;
 
     let packages =
-        split_spender_key(&spending_key, min_signers, identities).map_err(to_napi_err)?;
+        split_spender_key(&spending_key, min_signers, identities.clone()).map_err(to_napi_err)?;
```
andiflabs, line 206: We should change all these functions that expect a Vec as an input to accept a slice or an iter, so that cloning can be avoided here and in other places

patnir: I created a ticket for this: https://linear.app/if-labs/issue/IFL-2358/replace-vec-in-function-inputs-with-iter-or-arrays

andiflabs (approved):

PR #4823, opened 2024-03-08
verifies signing package signers in createSignatureShare

hughy:
## Summary

checks each signer identity in the input signing package against the set of participants for the account

throws an error upon finding a signer that doesn't belong to the set

## Testing Plan

manual testing with `wallet:multisig:signature:create`

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


hughy:
We verify the list of signers when creating a signing package, so this check shouldn't be necessary if the coordinator is using our wallet and RPC implementation

rohanjadvani:
There's a minor conflict to be resolved

PR #4828, opened 2024-03-08
Rahul/ifl 2350 add information to cli output to indicate next command to

patnir:
## Summary

Adding information to cli output to indicate next step in multi sig signing process. 

I leaned on the side of providing less information

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


rohanjadvani (commented):
I wonder if we should include the next command the next party should run too. Ex. `Send the commitment to the coordinator. They will need to run 'ironfish wallet:multisig:commitment:aggregate'.` What do you think?

patnir:
> I wonder if we should include the next command the next party should run too. Ex. `Send the commitment to the coordinator. They will need to run 'ironfish wallet:multisig:commitment:aggregate'.` What do you think?

I thought wanted doing it this way, but I felt like it isn't super relevant to the participant what the coordinator runs next. That information can be found in the docs, I thought of just focusing it to the immediate next step.

rohanjadvani (approved):

hughy (approved):

PR #4832, opened 2024-03-11
checks coordinator name against accounts and identities

hughy:
## Summary

prompts for new name until the user inputs a unique name

checks names before trusted dealer so that import doesn't fail after accounts have been created

includes small fix to RPC method to provide default undefined params

## Testing Plan

image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


rohanjadvani (approved):
CI might need a retry but looks good otherwise

patnir (approved):

PR #4837, opened 2024-03-12
Replace uses of `request<T>(...)` in tests with the appropriate method from `RpcClient`

andiflabs:
## Summary

This replaces code like the following:

    await routeTest.client.request<FooResponse>('foo', ...).waitForEnd()

with:

    await routeTest.client.foo(...)

Not all the uses of `request` have been replaced in this PR: only the trivial ones. The remaining ones require a bit more involved changes.

This PR also fixes some typing errors in the request parameters, and fixes some tests that were using `try`/`catch` instead of `expect` (the try/catch approach had the problem that if the RPC did not throw any exception, the test would have not failed).

## Testing Plan

Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):

NullSoldier:
Excellent cleanup, thank you.

rohanjadvani (approved):

PR #4843, opened 2024-03-14
Clarify when users are entering iron denominations

NullSoldier:
## Summary

This makes it more clear, however it also reveals a confusing issue we have. When using custom assets, why are we entering iron? What should we refer to the coin denomination for that asset? These issues will need to be fixed goign forward from a UX perspective.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/utils/currency.ts
+++ b/ironfish-cli/src/utils/currency.ts
@@ -6,6 +6,10 @@ import { Asset } from '@ironfish/rust-nodejs'
 import { Assert, CurrencyUtils, Logger, RpcClient } from '@ironfish/sdk'
 import { CliUx } from '@oclif/core'
 
+/**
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/utils/currency.ts
```diff
@@ -6,6 +6,10 @@ import { Asset } from '@ironfish/rust-nodejs'
 import { Assert, CurrencyUtils, Logger, RpcClient } from '@ironfish/sdk'
 import { CliUx } from '@oclif/core'
 
+/**
+ * This promps the user to enter an ammount of currency in IRON denominations
```
mat-if, line 10:
```suggestion
 * This prompts the user to enter an amount of currency in IRON denominations
```

mat-if (approved):

andiflabs (approved):

PR #4851, opened 2024-03-18
add additional fields to asset verification code

mat-if:
## Summary

Adds the additional fields outlined in the spec. There's more renaming to be done, but I didn't want to clobber the git diff too much in one PR.

## Testing Plan

Unit tests

## Documentation

N/A

## Breaking Change

No

### DIFF
--- a/ironfish/src/assets/assetsVerificationApi.ts
+++ b/ironfish/src/assets/assetsVerificationApi.ts
@@ -7,7 +7,6 @@ import { FileSystem } from '../fileSystems'
 
 type AssetData = {
   identifier: string
-  name: string
   symbol: string
   decimals?: number
   logoURI?: string
@@ -157,7 +156,6 @@ const axiosFileAdapter =
 const sanitizedAssetData = (assetData: AssetData): AssetData => {
   return {
     identifier: assetData.identifier,
-    name: assetData.name,
     symbol: assetData.symbol,
     decimals: assetData.decimals,
     logoURI: assetData.logoURI,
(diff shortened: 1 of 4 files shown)


ironfish/src/assets/assetsVerificationApi.ts
```diff
@@ -5,8 +5,17 @@ import axios, { AxiosAdapter, AxiosError, AxiosRequestConfig, AxiosResponse } fr
 import url, { URL } from 'url'
 import { FileSystem } from '../fileSystems'
 
-type GetVerifiedAssetsResponse = {
-  assets: Array<{ identifier: string }>
```
andiflabs, line 9: Note that changing `identifier` to `assetId` is a breaking change

ironfish/src/assets/assetsVerificationApi.ts
```diff
@@ -5,8 +5,17 @@ import axios, { AxiosAdapter, AxiosError, AxiosRequestConfig, AxiosResponse } fr
...
+type AssetData = {
+  assetId: string
+  name: string
+  symbol: string
+  decimals: number
+  logoURI: string
+}
```
andiflabs, line 14: While we're at this, it might be worth adding other fields, in particular: an URI/URL to point at a website describing asset. I'm not sure if this has been requested yet, but I certainly see some use cases for this

ironfish/src/assets/assetsVerificationApi.ts
```diff
@@ -5,8 +5,17 @@ import axios, { AxiosAdapter, AxiosError, AxiosRequestConfig, AxiosResponse } fr
...
 import { FileSystem } from '../fileSystems'
 
-type GetVerifiedAssetsResponse = {
-  assets: Array<{ identifier: string }>
+type AssetData = {
+  assetId: string
+  name: string
+  symbol: string
+  decimals: number
+  logoURI: string
```
andiflabs, line 13: Should `logoURI` be required or optional?

PR #4852, opened 2024-03-18
Adds a transaction timer class to reuse

patnir:
## Summary

Adds a transaction timer class that abstracts the timer logic away from combine notes. 
This is in setup for adding the timer to `wallet:send`

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/utils/transaction.ts
+++ b/ironfish-cli/src/utils/transaction.ts
@@ -23,22 +23,12 @@ export class TransactionTimer {
   private estimateInMs: number
   private timer: NodeJS.Timer | undefined
 
-  constructor(
-    spendPostTime: number,
-    raw: RawTransaction,
-    logger?: Logger,
-    options: {
-      displayEstimate?: boolean
-    } = {
-      displayEstimate: true,
-    },
-  ) {
+  constructor(spendPostTime: number, raw: RawTransaction, logger?: Logger) {
     this.logger = logger ?? createRootLogger()
     this.estimateInMs = Math.max(Math.ceil(spendPostTime * raw.spends.length), 1000)
-    options.displayEstimate && this.displayEstimate()
   }
 
-  private displayEstimate() {
+  displayEstimate() {
     this.logger.log(
       `Time to send: ${TimeUtils.renderSpan(this.estimateInMs, {
         hideMilliseconds: true,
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/utils/transaction.ts
```diff
@@ -14,6 +14,80 @@ import {
...
+    logger?: Logger,
+    options: {
+      displayEstimate?: boolean
+    } = {
+      displayEstimate: true,
+    },
+  ) {
+    this.logger = logger ?? createRootLogger()
+    this.estimateInMs = Math.max(Math.ceil(spendPostTime * raw.spends.length), 1000)
+    options.displayEstimate && this.displayEstimate()
```
andiflabs, line 38: Nit: it is unusual for constructors to have side-effects (in this case, printing something). I would make this behavior more explicit and require callers to call this method explicitly

ironfish-cli/src/utils/transaction.ts
```diff
@@ -14,6 +14,80 @@ import {
...
+    this.timer = setInterval(() => {
+      if (!this.progressBar || !this.startTime) {
+        return
+      }
+      const durationInMs = Date.now() - this.startTime
+      const timeRemaining = this.estimateInMs - durationInMs
+      const progress = Math.round((durationInMs / this.estimateInMs) * 100)
+
+      this.progressBar.update(progress, {
+        estimate: TimeUtils.renderSpan(timeRemaining, { hideMilliseconds: true }),
```
andiflabs, line 70: If `timeRemaining` is negative, this will show a negative estimate. Is there a way to show "unknown" or maybe "soon" instead?

andiflabs (approved):

ironfish-cli/src/utils/transaction.ts
patnir: Great feedback, I had it that way originally and I'll change it back

ironfish-cli/src/utils/transaction.ts
patnir: The behavior currently involves "counting up" after the countdown is complete. I talked through some options with Jason when initially implementing this and this seemed like the best one. This PR is only a refactor, so I'll leave as is but I'll continue to think through more options.

PR #4854, opened 2024-03-19
Rahul/ifl 1881 adds transaction progress bar to wallet send

patnir:
## Summary

- Adds transaction timer to wallet:send 
- when spendPostTime is not available, transaction timer displays the "normal" output 
- after posting a transaction, the spend post time is recalculated  
- The spendPostTime is averaged to smooth out anomalies over time
- We no longer need the `spendPostTimeAt` in the internal settings because of the above features

Example: 

image

## Testing Plan
 
1. With no spend post time in internal.json [you can manually remove it] run `wallet:send` - this will show you `Sending the transaction...` without an estimate. The next time you run the command, an estimate will be displayed
2. With a time estimate, run `wallet:send` you should get time estimated for both 

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


hughy (approved):
Looks good to me!

Tested with and without existing estimates 👍

PR #4864, opened 2024-04-01
allows deploying Docker images for staging branch

hughy:
## Summary

adds deploy step to deploy to image to github for branch refs

adds rule to deploy docker image on push to staging

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/.github/workflows/deploy-node-docker-image.yml
+++ b/.github/workflows/deploy-node-docker-image.yml
@@ -1,5 +1,8 @@
 name: Deploy Node Docker Image
 on:
+  push:
+    branches:
+      'staging'
   workflow_dispatch:
     inputs:
       github_tag_mainnet:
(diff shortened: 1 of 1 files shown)


andiflabs:
This will create _a lot_ of Docker images. Will we incur in costs due to this? Do we need a retention policy?

andiflabs (approved):

hughy:
> This will create _a lot_ of Docker images. Will we incur in costs due to this? Do we need a retention policy?

I think as an open source project we do not incur costs for these builds. Since all of the images use the same tag I don't think a retention policy is necessary.

That said, we could leave 'staging' builds as ad hoc or look at limiting automated builds to only when we do a release.

dguenther (approved):

PR #4872, opened 2024-04-04
Rahul/ifl 2455 refactor explicitly call benchmark spend post time

patnir:
## Summary

Refactors combine to explicitly call benchmarkSpendPostTime

Also removes `spendPostTimeAt` from internal store. In an upcoming PR this will be replaced with a different variable to create more accurate measurements for spendPostTime. details here: https://github.com/iron-fish/ironfish/pull/4854/files
This refactor is to make it easier to understand where `spendPostTime` is being called and how the value is being calculated.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/utils/spendPostTime.ts
+++ b/ironfish-cli/src/utils/spendPostTime.ts
@@ -15,16 +15,11 @@ import {
 import { CliUx } from '@oclif/core'
 import { fetchNotes } from './note'
 
-export async function getSpendPostTimeInMs(
-  sdk: IronfishSdk,
-  client: RpcClient,
-  account: string,
-  forceBenchmark: boolean,
-): Promise<number> {
+export function getSpendPostTimeInMs(sdk: IronfishSdk): number {
   return sdk.internal.get('spendPostTime')
 }
 
-async function benchmarkSpendPostTime(
+export async function benchmarkSpendPostTime(
   sdk: IronfishSdk,
   client: RpcClient,
   account: string,
--- a/ironfish-cli/src/utils/spendPostTime.ts
+++ b/ironfish-cli/src/utils/spendPostTime.ts
@@ -24,6 +24,8 @@ export async function benchmarkSpendPostTime(
   client: RpcClient,
   account: string,
 ): Promise<number> {
+  CliUx.ux.action.start('Measuring time to combine 1 note')
+
   const publicKey = (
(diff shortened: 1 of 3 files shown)


ironfish-cli/src/utils/spendPostTime.ts
```diff
@@ -50,6 +32,11 @@ async function benchmarkSpendPostTime(client: RpcClient, account: string): Promi
 
   const notes = await fetchNotes(client, account, 10)
 
+  // Not enough notes in the account to measure the time to combine a note
+  if (notes.length < 3) {
+    return 0
```
dguenther, line 37: I didn't look through the code to see where this is used, but do we handle this new 0 case somewhere up the stack so it doesn't look like your post time is instant?

Also, unrelated to this PR, but you could just make fake notes to do the benchmarking rather than requiring the client to have a certain number of notes.

dguenther (changes requested):

ironfish-cli/src/utils/spendPostTime.ts
```diff
@@ -97,15 +84,24 @@ async function benchmarkSpendPostTime(client: RpcClient, account: string): Promi
...
 
-  const delta = Math.ceil(
+  const spendPostTime = Math.ceil(
     (resultTxn2.reduce((acc, curr) => acc + curr, 0) -
       resultTxn1.reduce((acc, curr) => acc + curr, 0)) /
       3,
   )
 
-  CliUx.ux.action.stop(TimeUtils.renderSpan(delta))
+  const test = true
```
dguenther, line 93: Does this need to be removed?

ironfish-cli/src/utils/spendPostTime.ts
patnir: This check happens upstream in `fetchNotes` in the `wallet:notes:combine` command

ironfish-cli/src/utils/spendPostTime.ts
patnir: yep, removed

ironfish-cli/src/utils/spendPostTime.ts
patnir: I added an exception to be extra sure that this case doesn't get hit.

dguenther (approved):

PR #4873, opened 2024-04-04
removes migration logic from before network reset

hughy:
## Summary

any node that has asset, transaction, or note data from before the network reset in commit `f483d9aeb87eda3101ae2c602e19d3ebb88897b6` will need to clear out this data before the node can start on any version after the network reset.

since this data must be deleted it is unnecessary to execute any migrations that operate on this data. some of these migrations depend directly on the walletdb and may need to be updated if and when we make changes to the walletdb in the future. removing the logic reduces our code maintenance burden.

removes migration logic that reads or writes from/to chain or wallet datastores that store assets, transactions, or notes from before the network reset.

maintains migration logic that clears old datastores that no longer exist.

maintains migration logic that affects accounts. account data will not be deleted, so we still need to execute this migration logic.

## Testing Plan

manual testing running migrations from v0.1.63, v0.1.75

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


mat-if (approved):
💥

PR #4877, opened 2024-04-08
lazy deserialization in levelup db iter methods

hughy:
## Summary

our LevelupStore implementation has several methods for iterating over keys and values in the store: 'getAllIter', 'getAll', 'getAllValuesIter', 'getAllValues', 'getAllKeysIter', and 'getAllKeys'. these methods all use 'getAllIter', which always deserializes all keys and all values. this results in unnecessary deserialization of values when using 'getAllKeysIter' or 'getAllKeys' and likewise for keys when using 'getAllValuesIter' or 'getAllValues'.

updates LevelupStore to use lazy deserialization: only deserializing keys or values during iteration if required.

- changes LevelupTransaction to cache using database-encoded keys as Buffers instead of converting those Buffers to strings
- defines LevelupStore._getAllIter to yield encoded keys. yields encoded values where deserialized value is not in transaction cache
- defines resolveValue to deserialize value from Buffer if necessary

fixes bug in WalletDb.cleanupDeletedAccounts:
- if db migrations run after deleting an account, then data for deleted accounts are not migrated
- this may also occur if a rrescan starts before running db migrations because rescanning resets accounts by marking asset, transaction, and note data for deletion
- when cleanupDeletedAccounts runs it iterates over keys, but errors because deserialization of unmigrated values fails
- fixes error by preventing deserialization of unmigrated values

## Testing Plan

manual testing with cleanupDeletedAccounts bug:
1. checked out old version of Iron Fish (v1.3.0) from before most recent migration to walletDb assets store
2. started node and miner
3. minted asset on dev network
4. stopped node
5. deleted account that minted asset
6. checked out branch
7. ran migrations ('ironfish start --upgrade')
8. reset chain db due to genesis block mismatch ('ironfish reset')
9. started node without error

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/storage/levelup/store.ts
+++ b/ironfish/src/storage/levelup/store.ts
@@ -18,7 +18,6 @@ import {
   SchemaKey,
   SchemaValue,
 } from '../database'
-import { BufferToStringEncoding } from '../database/encoding'
 import { StorageUtils } from '../database/utils'
 import { LevelupTransaction } from './transaction'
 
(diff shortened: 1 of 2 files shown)


patnir:
I was able to reproduce the error: 
image

patnir:
I was able to restart the node using Hugh's fixes:

patnir (approved):
I tested the changes and everything works! Thanks @hughy for fixing this

PR #4878, opened 2024-04-08
Add new verified asset metadata fields to RpcAsset

mat-if:
## Documentation

Yes

https://github.com/iron-fish/website/pull/663

## Breaking Change

Yes - Add new fields to some RPC endpoints

### DIFF
--- a/ironfish/src/assets/assetsVerifier.ts
+++ b/ironfish/src/assets/assetsVerifier.ts
@@ -8,15 +8,14 @@ import { ErrorUtils } from '../utils'
 import { SetIntervalToken } from '../utils'
 import { Retry } from '../utils'
 import {
-  AdditionalAssetData,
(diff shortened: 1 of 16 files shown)


ironfish/src/assets/assetsVerifier.ts
```diff
@@ -114,4 +119,32 @@ export class AssetsVerifier {
       return { status: 'unverified' }
     }
   }
+
+  getAssetData(assetId: Buffer | string): VerifiedAssetMetadata | undefined {
+    return this.verifiedAssets?.getAssetData(assetId)
+  }
+
+  verifyWithMetadata(
```
danield9tqh, line 127: I think we could combine this with the `verify` function and use the same one for all instances

danield9tqh (approved):

PR #4879, opened 2024-04-08
Add new currency render function

mat-if:
## Summary

Adds a new render function for currency that handles custom decimal places for verified assets. Additionally, it will use the verified asset symbol when including the name instead of the asset id, which will improve readability a lot for verified assets.

## Testing Plan

Unit tests

## Documentation

N/A

## Breaking Change

N/A

### DIFF
--- a/ironfish/src/rpc/routes/wallet/createTransaction.ts
+++ b/ironfish/src/rpc/routes/wallet/createTransaction.ts
@@ -229,8 +229,13 @@ routes.register<typeof CreateTransactionRequestSchema, CreateTransactionResponse
       if (e instanceof NotEnoughFundsError) {
         // We are overriding the error message from the node to include verified assets in their appropriate denomination.
         const assetData = node.assetsVerifier.getAssetData(e.assetId)
-        const renderedAmountNeeded = CurrencyUtils.render(e.amountNeeded, true, assetData)
-        const renderedAmount = CurrencyUtils.render(e.amount, false, assetData)
+        const renderedAmountNeeded = CurrencyUtils.render(
+          e.amountNeeded,
+          true,
+          e.assetId,
+          assetData,
+        )
+        const renderedAmount = CurrencyUtils.render(e.amount, false, e.assetId, assetData)
         const message = `Insufficient funds: Needed ${renderedAmountNeeded} but have ${renderedAmount} available to spend. Please fund your account and/or wait for any pending transactions to be confirmed.'`
         throw new RpcValidationError(message, 400, RPC_ERROR_CODES.INSUFFICIENT_BALANCE)
       }
--- a/ironfish/src/rpc/routes/wallet/createTransaction.ts
+++ b/ironfish/src/rpc/routes/wallet/createTransaction.ts
@@ -112,7 +112,7 @@ routes.register<typeof CreateTransactionRequestSchema, CreateTransactionResponse
   `${ApiNamespace.wallet}/createTransaction`,
   CreateTransactionRequestSchema,
   async (request, node): Promise<void> => {
-    AssertHasRpcContext(request, node, 'wallet')
+    AssertHasRpcContext(request, node, 'wallet', 'assetsVerifier')
 
     const account = getAccount(node.wallet, request.data.account)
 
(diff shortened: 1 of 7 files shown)


ironfish/src/utils/currency.ts
```diff
@@ -58,6 +58,54 @@ export class CurrencyUtils {
...
 
+  /**
+   * Renders values for human-readable purposes:
+   * - Renders $IRON in the major denomination, with 8 decimal places
+   * - If a custom asset, and `decimals` is provided, it will render the custom
+   *     asset in the major denomination with the decimal places
+   * - If a custom asset, and `decimals` is not provided, it will render the
+   *     custom asset in the minor denomination with no decimal places
+   */
+  static render(
```
danield9tqh, line 69: Thoughts?
```typescript
static render(
    amount: bigint | string,
    includeSymbol: boolean = false,
    assetId?: string,
    verifiedAssetMetadata?: {
      decimals?: number
      symbol?: string
    },
  ): string {
    if (typeof amount === 'string') {
      amount = this.decode(amount)
    }

    let symbol: string
    let decimals: number
    // If an asset ID was provided, check if it is the native asset. Otherwise,
    // we can only assume that the amount is in the native asset
    if (!assetId || isNativeIdentifier(assetId)) {
      // Hard-code the amount of decimals in the native asset
      decimals = IRON_DECIMAL_PLACES
      symbol = '$IRON'
    } else {
      // Default to displaying 0 decimal places for custom assets
      decimals = verifiedAssetMetadata?.decimals || 0
      symbol = verifiedAssetMetadata?.symbol || assetId
    }

    const majorDenominationAmount = FixedNumberUtils.render(amount, decimals)

    if (includeSymbol) {
      return `${symbol} ${majorDenominationAmount}`
    }

    return majorDenominationAmount
  }
```

ironfish/src/rpc/routes/wallet/createTransaction.ts
```diff
@@ -227,7 +227,17 @@ routes.register<typeof CreateTransactionRequestSchema, CreateTransactionResponse
       })
     } catch (e) {
       if (e instanceof NotEnoughFundsError) {
-        throw new RpcValidationError(e.message, 400, RPC_ERROR_CODES.INSUFFICIENT_BALANCE)
+        // We are overriding the error message from the node to include verified assets in their appropriate denomination.
```
danield9tqh, line 230: nit: Maybe write a TODO to return these fields from RPC in error form?

ironfish/src/wallet/errors.ts
```diff
@@ -7,16 +7,19 @@ import { CurrencyUtils } from '../utils'
...
-      true,
-      assetId.toString('hex'),
-    )} but have ${CurrencyUtils.renderIron(
-      amount,
-    )} available to spend. Please fund your account and/or wait for any pending transactions to be confirmed.'`
+    this.assetId = assetId.toString('hex')
+    this.amount = amount
+    this.amountNeeded = amountNeeded
+
+    const renderedAmountNeeded = CurrencyUtils.render(amountNeeded, true, this.assetId)
```
danield9tqh, line 20: Do we want to support this in /sendTransaction also? If we're forcing the caller to already constructing their own message do we even need to construct a message here?

ironfish/src/utils/currency.test.ts
```diff
@@ -42,6 +43,62 @@ describe('CurrencyUtils', () => {
     expect(CurrencyUtils.decodeIron('0.00000999')).toBe(999n)
   })
 
+  describe('render', () => {
+    // Randomly generated custom asset ID
+    const assetId = '1a75bf033c1c1925cfcd1a77461364e77c6e861c2a3acabaf9e398e980146651'
+
+    it('should render iron with no extra parameters with 8 decimal places', () => {
+      expect(CurrencyUtils.render(0n)).toEqual('0.00000000')
```
danield9tqh, line 51: Do we actually want to render trailing zeros?

danield9tqh: Will discuss in future PR

danield9tqh (approved):

PR #4880, opened 2024-04-08
Use new render function in the CLI

mat-if:
## Summary

Use the new currency render function in the CLI. See #4879 for more information about that function

## Testing Plan

Manual testing

## Documentation

N/A

## Breaking Change

N/A

### DIFF
--- a/ironfish-cli/src/commands/wallet/balance.ts
+++ b/ironfish-cli/src/commands/wallet/balance.ts
@@ -68,7 +68,7 @@ export class BalanceCommand extends IronfishCommand {
     if (isNativeIdentifier(assetId)) {
       nameToRender = '$IRON'
     } else {
-      nameToRender = asset.symbol || assetId
+      nameToRender = asset.verification.symbol || assetId
     }
     const assetName = renderAssetWithVerificationStatus(nameToRender, asset)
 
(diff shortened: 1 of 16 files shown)


ironfish-cli/src/commands/wallet/balance.ts
```diff
@@ -54,60 +54,56 @@ export class BalanceCommand extends IronfishCommand {
...
         account,
-        id: response.content.assetId,
+        id: assetId,
         confirmations: flags.confirmations,
       })
     ).content
 
-    const assetId = response.content.assetId
-    const assetName = renderAssetName(isNativeIdentifier(assetId) ? '$IRON' : assetId, {
+    let nameToRender
```
danield9tqh, line 67: Do you think we should re-use this functionality in `CurrencyUtils`?

ironfish-cli/src/utils/currency.ts
```diff
@@ -40,11 +40,17 @@ export async function promptCurrency(options: {
   if (options.balance) {
```
danield9tqh, line 40: We should also be changing `CurrencyUtils.decodeIronTry(input)`

ironfish-cli/src/utils/asset.ts
```diff
@@ -124,9 +124,13 @@ export async function selectAsset(
   const choices = balances.map((balance) => {
```
danield9tqh, line 124: We should
- give verified status when selecting an asset
- put verified assets at top of list

(can do in future PR)

ironfish-cli/src/utils/currency.ts
```diff
@@ -31,7 +31,7 @@ export async function promptCurrency(options: {
   minimum?: bigint
   balance?: {
     account?: string
-    assetId?: string
+    asset?: RpcAsset
```
danield9tqh, line 34: I think we should move this to top level params

danield9tqh (approved):

PR #4884, opened 2024-04-08
Parse values in the proper denomination for CLI commands

mat-if:
## Summary

Updates `decodeIronTry` with `tryMajorToMinor` for converting a value from major denomination to minor denomination. Previously, it assumed all assets would have 8 decimal places, but this is no longer the case.

## Testing Plan

Unit tests, manual tests

## Documentation

N/A

## Breaking Change

N/A


ironfish-cli/src/commands/wallet/burn.ts
```diff
@@ -46,9 +46,9 @@ export class Burn extends IronfishCommand {
       minimum: 1n,
       flagName: 'fee rate',
     }),
-    amount: IronFlag({
+    amount: ValueFlag({
```
danield9tqh, line 49: Should handle parsing decimal places here and in other commands. We prob have to keep this as a string flag and parse it later

ironfish-cli/src/flags.ts
```diff
@@ -157,6 +157,17 @@ export const parseIron = (input: string, opts: IronOpts): Promise<bigint> => {
...
 }
 
+/**
+ * A flag used for a value when we do not yet know how to treat the conversion
+ * from major to minor denomination. Parses the value simply as a valid number,
+ * to be converted later manually.
+ */
+export const ValueFlag = Flags.custom<string>({
+  parse: async (input, _ctx, _opts) => {
+    return Promise.resolve(CurrencyUtils.parseFixed(input))
```
danield9tqh, line 167: Should we reject with a custom message here?

PR #4885, opened 2024-04-08
Implement `wallet/multisig/dkg/round1` RPC

andiflabs:
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-rust-nodejs/src/multisig.rs
```diff
@@ -353,3 +357,54 @@ impl NativeSigningPackage {
...
+    }
+
+    let mut rng = thread_rng();
+    let (secret_package, public_package) = dkg::part1(
+        self_identity.to_frost_identifier(),
+        max_signers,
+        min_signers,
+        &mut rng,
+    )
+    .map_err(to_napi_err)?;
```
hughy, line 393: wondering if we should move logic like this into `ironfish-rust` or `ironfish-frost` as we add more validation, checksums, etc.

I'm not sure if having it in `ironfish-rust-nodejs` will make it difficult to use if someone were trying to build on top of it outside of the TypeScript SDK

ironfish/src/rpc/routes/wallet/multisig/dkg/round1.ts
```diff
@@ -0,0 +1,89 @@
...
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { multisig } from '@ironfish/rust-nodejs'
+import * as yup from 'yup'
+import { RPC_ERROR_CODES, RpcValidationError } from '../../../../adapters'
+import { ApiNamespace } from '../../../namespaces'
+import { routes } from '../../../router'
+import { AssertHasRpcContext } from '../../../rpcContext'
+
+export type DkgRound1Request = {
+  // TODO `secretName` or `multisigSecretName`?
```
hughy, line 12: I think either of these is a good option over `name` 😁

ironfish/src/rpc/routes/wallet/multisig/dkg/round1.ts
```diff
@@ -0,0 +1,89 @@
...
+      )
+    }
+
+    if (minSigners > participantIdentities.length) {
+      throw new RpcValidationError(
+        `minSigners (${minSigners}) exceeds the number of participants (${participantIdentities.length})`,
+        400,
+        RPC_ERROR_CODES.VALIDATION,
+      )
+    }
```
hughy, line 83: Should we also verify that `minSigners >= 2`?

ironfish/src/rpc/routes/wallet/multisig/dkg/round1.ts
```diff
@@ -0,0 +1,89 @@
...
+  DkgRound1RequestSchema,
+  async (request, node): Promise<void> => {
+    AssertHasRpcContext(request, node, 'wallet')
+
+    const { name, minSigners, participants } = request.data
+
+    // TODO check if this secret is already being used for an account (there's
+    // nothing inherently wrong with using the same secret for multiple
+    // multisig groups, but I think our code assumes a 1:1 relationship between
+    // secret names and account names)
```
hughy, line 52: I don't think we have a way to prevent this at all with TDK 🤔

I think we should avoid relying on a 1:1 relationship between secret names and account names

andiflabs: Good to know, thanks. I must have misread some code I found earlier.

ironfish/src/rpc/routes/wallet/multisig/dkg/round1.ts
andiflabs: Good idea, thanks for spotting that

ironfish/src/rpc/routes/wallet/multisig/dkg/round1.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    const otherParticipants = Array.from({ length: 2 }, () => ({
+      identity: multisig.ParticipantSecret.random().toIdentity().serialize().toString('hex'),
+    }))
+    const participants = [{ identity }, ...otherParticipants]
+
+    const request = { secretName, minSigners: 2, participants }
+    const response = await routeTest.client.wallet.multisig.dkg.round1(request)
+
+    expect(response.content).toMatchObject({
+      encryptedSecretPackage: expect.any(String),
```
rohanjadvani, line 25: Wonder if it'd be useful to add that this can be decrypted in the test

rohanjadvani (approved):

ironfish/src/rpc/routes/wallet/multisig/dkg/round1.test.ts
andiflabs: Good idea, will do

rohanjadvani (approved):

PR #4892, opened 2024-04-15
Fix an issue with the mint command where the id was not properly being passed

mat-if:
## Summary

This was causing an issue where minting a new asset would default to the iron default of 8 decimal places instead of the custom asset default of 0

## Testing Plan

Manually tested

## Documentation

No

## Breaking Change

No

### DIFF
--- a/ironfish-cli/src/commands/wallet/mint.ts
+++ b/ironfish-cli/src/commands/wallet/mint.ts
@@ -10,7 +10,9 @@ import {
   isValidPublicAddress,
   RawTransaction,
   RawTransactionSerde,
+  RPC_ERROR_CODES,
   RpcAsset,
+  RpcRequestError,
   Transaction,
 } from '@ironfish/sdk'
 import { CliUx, Flags } from '@oclif/core'
@@ -186,10 +188,20 @@ export class Mint extends IronfishCommand {
 
     let assetData
     if (assetId) {
-      assetData = (await client.chain.getAsset({ id: assetId })).content
-      const isAssetOwner = this.isAssetOwner(assetData, accountPublicKey)
-      if (!isAssetOwner) {
-        this.error(`The account '${account}' does not own this asset.`)
+      try {
+        const assetRequest = await client.chain.getAsset({ id: assetId })
+        assetData = assetRequest.content
+        const isAssetOwner = this.isAssetOwner(assetData, accountPublicKey)
+        if (!isAssetOwner) {
+          this.error(`The account '${account}' does not own this asset.`)
+        }
+      } catch (e) {
+        if (e instanceof RpcRequestError && e.code === RPC_ERROR_CODES.NOT_FOUND.valueOf()) {
+          // Do nothing, not finding an asset is acceptable since we're likely
+          // to be minting one for the first time
+        } else {
+          throw e
+        }
       }
     }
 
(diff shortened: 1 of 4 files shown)


ironfish-cli/src/commands/wallet/mint.ts
```diff
@@ -230,7 +243,8 @@ export class Mint extends IronfishCommand {
       outputs: [],
       mints: [
         {
-          assetId,
+          // Only provide the asset id if we are not minting an asset for the first time
+          ...(assetData != null ? { assetId } : {}),
```
mat-if, line 247: This isn't .. ideal, but the create transaction RPC throws if you pass an asset id for a new asset and I didn't want to muddy this PR up too much

ironfish-cli/src/commands/wallet/mint.ts
```diff
@@ -186,10 +188,20 @@ export class Mint extends IronfishCommand {
...
-        this.error(`The account '${account}' does not own this asset.`)
+      try {
+        const assetRequest = await client.chain.getAsset({ id: assetId })
+        assetData = assetRequest.content
+        const isAssetOwner = this.isAssetOwner(assetData, accountPublicKey)
+        if (!isAssetOwner) {
+          this.error(`The account '${account}' does not own this asset.`)
+        }
+      } catch (e) {
+        if (e instanceof RpcRequestError && e.code === RPC_ERROR_CODES.NOT_FOUND.valueOf()) {
```
danield9tqh, line 199: This one I could potentially see putting in the client SDK and returning null. nit tho

danield9tqh: Also don't think we need the `valueOf`

ironfish-cli/src/commands/wallet/mint.ts
```diff
@@ -186,10 +188,20 @@ export class Mint extends IronfishCommand {
...
     let assetData
     if (assetId) {
-      assetData = (await client.chain.getAsset({ id: assetId })).content
-      const isAssetOwner = this.isAssetOwner(assetData, accountPublicKey)
-      if (!isAssetOwner) {
-        this.error(`The account '${account}' does not own this asset.`)
+      try {
+        const assetRequest = await client.chain.getAsset({ id: assetId })
+        assetData = assetRequest.content
+        const isAssetOwner = this.isAssetOwner(assetData, accountPublicKey)
```
danield9tqh, line 194: I think this `isAssetOwner` function can be removed now too

ironfish-cli/src/commands/wallet/mint.ts
```diff
@@ -186,10 +188,20 @@ export class Mint extends IronfishCommand {
 
     let assetData
     if (assetId) {
-      assetData = (await client.chain.getAsset({ id: assetId })).content
-      const isAssetOwner = this.isAssetOwner(assetData, accountPublicKey)
-      if (!isAssetOwner) {
-        this.error(`The account '${account}' does not own this asset.`)
+      try {
```
danield9tqh, line 191: Feel like maybe this `client.chain.getAsset({ id: assetId })` can be under a branch where `isMintingNewAsset` is `false`. In that case we prob would want to throw this not found error because you're trying to mint an existing asset that doesn't exist. Currently I don't think this command handles that case

danield9tqh (approved):

PR #4894, opened 2024-04-15
updating level db version to 6.1.1

patnir:
## Summary

Core reason for this upgrade: bug fixes that come with this version. Here is a description from their docs: 

```
This is a rewrite to N-API - which is a huge milestone, achieved without an impact on write performance - and an upgrade to abstract-leveldown v6, which solves long-standing issues around serialization and type support.
```

More details: https://github.com/Level/leveldown/blob/master/UPGRADING.md

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/package.json
+++ b/ironfish/package.json
@@ -38,7 +38,7 @@
     "fastpriorityqueue": "0.7.1",
     "imurmurhash": "0.1.4",
     "level-errors": "2.0.1",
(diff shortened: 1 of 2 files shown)


dguenther (approved):
One other idea I had was cross-platform -- I synced up a node on ARM Mac and x64 Windows and didn't run into any issues.

PR #4895, opened 2024-04-16
Avoid an infinite loop in `SaplingKey::generate_key`

andiflabs:
## Summary

If the random scalar generated is zero (which is a rare occurrence, almost impossible, and has the same probability to occur as guessing somebody's private key), `SaplingKey::generate_key` is supposed to generate a new random scalar and try again. The current code however, does not generate a new random scalar if that occurs, resulting in an infinite loop.

## Testing Plan

Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


hughy (approved):

mat-if (approved):
Ooh good catch

PR #4900, opened 2024-04-16
Implement `wallet/multisig/dkg/round3` RPC

andiflabs:
## Testing Plan

Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust-nodejs/src/multisig.rs
+++ b/ironfish-rust-nodejs/src/multisig.rs
@@ -5,19 +5,22 @@
 use crate::{structs::NativeUnsignedTransaction, to_napi_err};
 use ironfish::{
     frost::{keys::KeyPackage, round2, Randomizer},
-    frost_utils::{signing_package::SigningPackage, split_spender_key::split_spender_key},
+    frost_utils::{
+        account_keys::derive_account_keys, signing_package::SigningPackage,
+        split_spender_key::split_spender_key,
+    },
     participant::{Identity, Secret},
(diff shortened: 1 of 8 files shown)


ironfish-rust-nodejs/src/multisig.rs
```diff
@@ -447,3 +451,64 @@ pub struct DkgRound2Packages {
...
+    let round1_public_packages =
+        round1_public_packages
+            .into_iter()
+            .try_fold(Vec::new(), |mut vec, pkg| {
+                let pkg = hex_to_vec_bytes(&pkg).map_err(to_napi_err)?;
+                let pkg =
+                    dkg::round1::PublicPackage::deserialize_from(&pkg[..]).map_err(to_napi_err)?;
+                vec.push(pkg);
+                Ok::<_, napi::Error>(vec)
+            })?;
```
hughy, line 472: nit - this is what `try_deserialize_public_packages` does. should we inline this logic in both places?

hughy (approved):

ironfish-rust-nodejs/src/multisig.rs
andiflabs: Forgot about that method; added a more generic `try_deserialize` that can be reused in all places

andiflabs: I did not remove `try_deserialize_identities` because that's used in relatively many places and I think it's nice to keep it

PR #4905, opened 2024-04-17
feat(cli,ironfish): Add `wallet:multisig:dkg:round3`

rohanjadvani:
## Summary

* Returns name and public address from round3 rpc endpoint
* Adds CLI command for dkg round 3

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


hughy (approved):
for a followup task -- we should support a way to choose a different name other than the secretName for the account

andiflabs:
> for a followup task -- we should support a way to choose a different name other than the secretName for the account

Yes, I initially included this functionality in the RPC, but then discovered that the name was being overridden in the SDK, but now that #4903 is merged we can add this functionality back

PR #4912, opened 2024-04-18
feat(cli,ironfish): Add flag to override name in round3

rohanjadvani:
## Summary

Allow an override of the secret name for the account name during round 3

## Testing Plan

Updated unit test

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/multisig/dkg/round3.ts
+++ b/ironfish-cli/src/commands/wallet/multisig/dkg/round3.ts
@@ -17,7 +17,7 @@ export class DkgRound3Command extends IronfishCommand {
       description: 'The name of the secret to use for decryption during DKG',
       required: true,
     }),
-    name: Flags.string({
+    accountName: Flags.string({
       char: 'n',
       description: 'The name to set for the imported account',
     }),
@@ -92,7 +92,7 @@ export class DkgRound3Command extends IronfishCommand {
 
     const response = await client.wallet.multisig.dkg.round3({
       secretName: flags.secretName,
-      accountName: flags.name,
+      accountName: flags.accountName,
       round2SecretPackage,
       round1PublicPackages,
       round2PublicPackages,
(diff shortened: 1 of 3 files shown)


ironfish/src/rpc/routes/wallet/multisig/dkg/round3.ts
```diff
@@ -28,6 +29,7 @@ export const DkgRound3RequestSchema: yup.ObjectSchema<DkgRound3Request> = yup
     round2SecretPackage: yup.string().defined(),
     round1PublicPackages: yup.array().of(yup.string().defined()).defined(),
     round2PublicPackages: yup.array().of(yup.string().defined()).defined(),
+    name: yup.string().optional(),
```
andiflabs, line 32: Nit: can we call this `accountName`?

ironfish-cli/src/commands/wallet/multisig/dkg/round3.ts
```diff
@@ -17,6 +17,10 @@ export class DkgRound3Command extends IronfishCommand {
       description: 'The name of the secret to use for decryption during DKG',
       required: true,
     }),
+    name: Flags.string({
```
andiflabs, line 20: Would `accountName` be a better name for the flag?

andiflabs (approved):

ironfish-cli/src/commands/wallet/multisig/dkg/round3.ts
rohanjadvani: Yeah I'm ok with that. Updated

ironfish/src/rpc/routes/wallet/multisig/dkg/round3.ts
rohanjadvani: Yeah I'm ok with that. Updated

PR #4915, opened 2024-04-19
adds raw transaction flag to combine notes

patnir:
## Summary

Adds raw transaction flag to notes:combine

Usecase: Create a transaction and post it separately using `wallet:post`. This is a common pattern we use with with other transaction types as well

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/notes/combine.ts
+++ b/ironfish-cli/src/commands/wallet/notes/combine.ts
@@ -326,7 +326,7 @@ export class CombineNotesCommand extends IronfishCommand {
 
     if (flags.rawTransaction) {
       this.log('Raw Transaction')
-      this.log(RawTransactionSerde.serialize(raw).toString('hex'))
+      this.log(createTransactionBytes.toString('hex'))
       this.log(`Run "ironfish wallet:post" to post the raw transaction. `)
       this.exit(0)
     }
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/wallet/notes/combine.ts
```diff
@@ -76,6 +76,11 @@ export class CombineNotesCommand extends IronfishCommand {
       default: false,
       description: 'Force run the benchmark to measure the time to combine 1 note',
     }),
+    rawTransaction: Flags.boolean({
+      default: false,
+      description:
+        'Return raw transaction. Use it to create a transaction but not post to the network',
+    }),
   }
```
jowparks, line 84: What is the use case for this, can you include in PR description?

patnir: Updated

andiflabs (approved):

ironfish-cli/src/commands/wallet/notes/combine.ts
```diff
@@ -319,6 +324,13 @@ export class CombineNotesCommand extends IronfishCommand {
 
     displayTransactionSummary(raw, assetData, amount, from, to, memo)
 
+    if (flags.rawTransaction) {
+      this.log('Raw Transaction')
+      this.log(RawTransactionSerde.serialize(raw).toString('hex'))
```
hughy, line 329: nit - can use `createdTransactionBytes` here to avoid re-serializing

hughy (approved):

PR #4920, opened 2024-04-22
Multisig: fetch participant identities from the public key package

andiflabs:
## Summary

When importing a multisig account, participant identities are stored in two places:

1. the public key package (part of the account record in the walletdb)
2. on dedicated records in the walletdb

The \#2 records can sometimes get lost. This is the case currently of account rescans.

Because there is no future development scheduled that will benefit from #2, it's easier to just ignore those records and instead rely only on the public key package.

## Testing Plan

Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/walletdb/walletdb.ts
+++ b/ironfish/src/wallet/walletdb/walletdb.ts
@@ -1297,32 +1297,4 @@ export class WalletDB {
       yield value
     }
   }
-
-  async addParticipantIdentity(
-    account: Account,
-    identity: Buffer,
-    tx?: IDatabaseTransaction,
-  ): Promise<void> {
-    await this.participantIdentities.put([account.prefix, identity], { identity }, tx)
-  }
-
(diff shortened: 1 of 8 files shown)


ironfish/src/wallet/walletdb/walletdb.ts
```diff
@@ -1297,32 +1297,4 @@ export class WalletDB {
       yield value
     }
   }
-
-  async addParticipantIdentity(
-    account: Account,
-    identity: Buffer,
-    tx?: IDatabaseTransaction,
-  ): Promise<void> {
-    await this.participantIdentities.put([account.prefix, identity], { identity }, tx)
```
hughy, line 1306: should we remove this datastore from the walletdb altogether?

maybe in a separate pr

hughy (approved):

PR #4927, opened 2024-04-23
renames secretName to participantName

hughy:
## Summary

the term 'secretName' does not make it clear that the name should be the same as the one used with 'participant:create', that is, the name of a participant

renames the flag and rpc parameter 'secretName' to 'participantName' across all dkg cli commands and rpc endpoints

changes flag character from s to n. in round3, this also requires changing the 'accountName' flag character from n to a

adds an alias 'name' for 'participantName' in all dkg commands

## Testing Plan

manual testing: created an account by running all three dkg rounds

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/multisig/dkg/round1.ts
+++ b/ironfish-cli/src/commands/wallet/multisig/dkg/round1.ts
@@ -21,7 +21,7 @@ export class DkgRound1Command extends IronfishCommand {
     identity: Flags.string({
       char: 'i',
       description:
-        'The identity of the participants will generate the group keys (may be specified multiple times to add multiple participants). Must include the identity for participantName',
+        'The identity of the participants will generate the group keys (may be specified multiple times to add multiple participants)',
       multiple: true,
     }),
     minSigners: Flags.integer({
--- a/ironfish-cli/src/commands/wallet/multisig/dkg/round1.ts
+++ b/ironfish-cli/src/commands/wallet/multisig/dkg/round1.ts
@@ -13,14 +13,15 @@ export class DkgRound1Command extends IronfishCommand {
 
   static flags = {
     ...RemoteFlags,
-    secretName: Flags.string({
-      char: 's',
+    participantName: Flags.string({
+      char: 'n',
       description: 'The name of the secret to use for encryption during DKG',
+      aliases: ['name'],
     }),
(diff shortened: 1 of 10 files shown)


ironfish-cli/src/commands/wallet/multisig/dkg/round2.ts
```diff
@@ -34,15 +35,15 @@ export class DkgRound2Command extends IronfishCommand {
...
+    let participantName = flags.participantName
+    if (!participantName) {
+      participantName = await selectSecret(client)
     }
 
     let encryptedSecretPackage = flags.encryptedSecretPackage
     if (!encryptedSecretPackage) {
       encryptedSecretPackage = await CliUx.ux.prompt(
-        `Enter the encrypted secret package for secret ${secretName}`,
+        `Enter the encrypted secret package for secret ${participantName}`,
```
rohanjadvani, line 46:
```suggestion
        `Enter the encrypted secret package for participant ${participantName}`,
```

ironfish-cli/src/commands/wallet/multisig/dkg/round1.ts
```diff
@@ -13,14 +13,15 @@ export class DkgRound1Command extends IronfishCommand {
...
+    participantName: Flags.string({
+      char: 'n',
       description: 'The name of the secret to use for encryption during DKG',
+      aliases: ['name'],
     }),
     identity: Flags.string({
       char: 'i',
       description:
-        'The identity of the participants will generate the group keys (may be specified multiple times to add multiple participants). Must include the identity for secretName',
+        'The identity of the participants will generate the group keys (may be specified multiple times to add multiple participants). Must include the identity for participantName',
```
andiflabs, line 24: (Does not need to be addressed here)

"Must include hte identity for participantName" << is this really true? Don't we add it automatically if omitted?

andiflabs (approved):

ironfish-cli/src/commands/wallet/multisig/dkg/round1.ts
hughy: Good catch! It's a small change, so I'll make it here

rohanjadvani (approved):

PR #4930, opened 2024-04-24
updates text of identities prompts

hughy:
## Summary

clarifies which participant identities are expected when prompting for a list of identities on the command line in 'commitment:create', 'dealer:create', and 'dkg:round1'

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/multisig/commitment/create.ts
+++ b/ironfish-cli/src/commands/wallet/multisig/commitment/create.ts
@@ -48,7 +48,7 @@ export class CreateSigningCommitmentCommand extends IronfishCommand {
     let identities = options.identity
     if (!identities || identities.length < 2) {
       const input = await longPrompt(
-        'Enter the identities of all participants who will sign the transaction separated by commas',
+        'Enter the identities of all participants who will sign the transaction, separated by commas',
         {
           required: true,
         },
--- a/ironfish-cli/src/commands/wallet/multisig/commitment/create.ts
+++ b/ironfish-cli/src/commands/wallet/multisig/commitment/create.ts
@@ -47,9 +47,12 @@ export class CreateSigningCommitmentCommand extends IronfishCommand {
 
     let identities = options.identity
     if (!identities || identities.length < 2) {
-      const input = await longPrompt('Enter the identities separated by commas', {
-        required: true,
-      })
+      const input = await longPrompt(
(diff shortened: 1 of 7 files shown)


ironfish-cli/src/commands/wallet/multisig/commitment/create.ts
```diff
@@ -47,9 +47,12 @@ export class CreateSigningCommitmentCommand extends IronfishCommand {
 
     let identities = options.identity
     if (!identities || identities.length < 2) {
-      const input = await longPrompt('Enter the identities separated by commas', {
-        required: true,
-      })
+      const input = await longPrompt(
+        'Enter the identities of all participants who will sign the transaction separated by commas',
```
andiflabs, line 51: (Not a native speaker/writer) Should there be a comma between "transaction" and "separated"?

ironfish-cli/src/commands/wallet/multisig/dealer/create.ts
```diff
@@ -38,9 +38,12 @@ export class MultisigCreateDealer extends IronfishCommand {
 
     let identities = flags.identity
     if (!identities || identities.length < 2) {
-      const input = await longPrompt('Enter the identities separated by commas', {
-        required: true,
-      })
+      const input = await longPrompt(
+        'Enter the identities of all participants separated by commas',
```
andiflabs, line 42: Same question as above

ironfish-cli/src/commands/wallet/multisig/dkg/round1.ts
```diff
@@ -42,9 +42,12 @@ export class DkgRound1Command extends IronfishCommand {
 
     let identities = flags.identity
     if (!identities || identities.length < 2) {
-      const input = await longPrompt('Enter the identities separated by commas', {
-        required: true,
-      })
+      const input = await longPrompt(
+        'Enter the identities of all participants separated by commas',
```
andiflabs, line 46: Same

andiflabs (approved):

ironfish-cli/src/commands/wallet/multisig/commitment/create.ts
hughy: I think there should be. I had one in there but removed it because we had other uses without the comma 😅

PR #4938, opened 2024-04-26
Add minorToMajor conversion function

danield9tqh:
## Summary
This PR
- Removes some dependencies on `@ethersproject/bignumber`
- Creates new method `minorToMajor` which converts a minor currency denomination to a major currency denomination without rendering it as a string. This is useful for re-using in places like the node app where we want to render in different ways

## Testing Plan
Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change
- Removes `FixedNumberUtils` helper class


ironfish-cli/src/flags.ts
```diff
@@ -167,7 +167,7 @@ export const ValueFlag = Flags.custom<string>({
   parse: async (input, _ctx, opts) => {
     return new Promise((resolve, reject) => {
       try {
-        FixedNumberUtils.tryDecodeDecimal(input)
+        Exponent.tryParse(input)
```
danield9tqh, line 170: This seems a little weird exposing Exponent like this

ironfish/src/utils/currency.ts
```diff
@@ -80,6 +75,28 @@ export class CurrencyUtils {
...
 
+  /**
+   * Renders values for human-readable purposes:
+   * - Renders $IRON in the major denomination, with 8 decimal places
+   * - If a custom asset, and `decimals` is provided, it will render the custom
+   *     asset in the major denomination with the decimal places
+   * - If a custom asset, and `decimals` is not provided, it will render the
+   *     custom asset in the minor denomination with no decimal places
+   */
+  static minorToMajor(
```
mat-if, line 86: Would you mind adding a few tests for this? Doesn't have to be crazy since this is a fairly light function that mostly rides on other tested functions - mostly just a way to provide some better visibility into SDK changes if they have to be altered etc

mat-if (approved):

ironfish/src/utils/currency.ts
danield9tqh: Yep, will add

mat-if (approved):

PR #4941, opened 2024-05-03
Upgrade node-datachannel to 0.8.0

dguenther:
## Summary

Upgrades node-datachannel to 0.8.0. In https://github.com/murat-dogan/node-datachannel/releases/tag/v0.6.0, OpenSSL was changed to link statically on arm64, so this upgrade will let us build Linux arm64 Docker images without having to add in OpenSSL.

They also resumed supporting CommonJS modules, so we can revert all of the changes to plumb through an ESModule implementation.

## Testing Plan

Make sure you can run start and connect to other peers via WebRTC on the following platforms:

* [x] Test on Windows x64
* [x] Test on Mac arm64
* [x] Test on Linux x64

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


Bernmos22:
The Site needs revamping and a good wallet protocol

danield9tqh (approved):

PR #4942, opened 2024-05-06
Feat: support chain/getBlocks

hairtail:
## Summary
This PR adds `chain/getBlocks` to support get Ironfish blocks in batch.
## Testing Plan
To do: add tests.
## Documentation

```
[✓] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/chain/getBlocks.ts
```diff
@@ -0,0 +1,132 @@
...
+        assetName: mint.asset.name().toString('hex'),
+      })),
+      burns: tx.burns.map((burn) => ({
+        id: burn.assetId.toString('hex'),
+        value: burn.value.toString(),
+        assetId: burn.assetId.toString('hex'),
+        assetName: '',
+      })),
+      signature: tx.transactionSignature().toString('hex'),
+    })
```
dguenther, line 124: Could you refactor this out of `getBlock` and `getBlocks` into a function in `rpc/routes/chain/utils.ts` called `serializeRpcTransaction` that takes a `Transaction` and returns a `RpcTransaction`? Since we're using the same code in both places.

ironfish/src/rpc/routes/chain/getBlocks.ts
```diff
@@ -0,0 +1,132 @@
...
+  .defined()
+
+export const GetBlocksResponseSchema: yup.ObjectSchema<GetBlocksResponse> = yup
+  .object({
+    blocks: yup.array(RpcBlockSchema).defined(),
+  })
+  .defined()
+
+routes.register<typeof GetBlocksRequestSchema, GetBlocksResponse>(
+  `${ApiNamespace.chain}/getBlocks`,
```
dguenther, line 48: Could you add a test file for this, since the other RPC routes in this folder have them? You should be able to copy and modify the tests in `getBlock.test.ts`, since it's almost the same route as those

ironfish/src/rpc/routes/chain/getBlocks.ts
```diff
@@ -0,0 +1,132 @@
...
+import { FullNode } from '../../../node'
+import { GENESIS_BLOCK_SEQUENCE } from '../../../primitives/block'
+import { BufferUtils } from '../../../utils'
+import { RpcNotFoundError, RpcValidationError } from '../../adapters'
+import { ApiNamespace } from '../namespaces'
+import { routes } from '../router'
+import { RpcBlock, RpcBlockSchema, serializeRpcBlockHeader } from '../types'
+import { RpcTransaction } from './types'
+
+const MAX_BLOCKS_RANGE = 30
```
dguenther, line 16: Marking this to think about a bit more, but no action needed yet.

After looking around the RPC code, I don't think we have max limits for RPC calls anywhere else yet, so I want to think a bit about what our pattern should be.

dguenther (changes requested):

ironfish/src/rpc/routes/chain/getBlock.ts
```diff
@@ -109,40 +109,7 @@ routes.register<typeof GetBlockRequestSchema, GetBlockResponse>(
...
-        burns: tx.burns.map((burn) => ({
-          id: burn.assetId.toString('hex'),
-          value: burn.value.toString(),
-          assetId: burn.assetId.toString('hex'),
-          assetName: '',
-        })),
-        signature: tx.transactionSignature().toString('hex'),
-        ...(request.data?.serialized ? { serialized: tx.serialize().toString('hex') } : {}),
-      })
+      transactions.push(serializeRpcTransaction(tx))
```
dguenther, line 112: `getBlock` will still need to support the `serialized` flag, like this:

```suggestion
      transactions.push({
        ...serializeRpcTransaction(tx),
        ...(request.data?.serialized ? { serialized: tx.serialize().toString('hex') } : {}),
      })
```

ironfish/src/rpc/routes/chain/getBlocks.test.ts
```diff
@@ -0,0 +1,94 @@
...
+        hash: hash1,
+        sequence: 2,
+      },
+    })
+    expect(blocks[0]).toMatchObject({
+      block: {
+        hash: hash2,
+        sequence: 3,
+      },
+    })
```
dguenther, line 48: This looks like it's checking blocks[0] multiple times. I'm guessing it should be:

```suggestion
    expect(blocks[0]).toMatchObject({
      block: {
        hash: hash0,
        sequence: 1,
      },
    })
    expect(blocks[1]).toMatchObject({
      block: {
        hash: hash1,
        sequence: 2,
      },
    })
    expect(blocks[2]).toMatchObject({
      block: {
        hash: hash2,
        sequence: 3,
      },
    })
```

ironfish/src/rpc/routes/chain/getBlocks.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export type GetBlocksRequest = {
+  /**
+   * The starting block height (inclusive).
+   */
+  start: number
+  /**
+   * The ending block height (exclusive).
+   */
+  end: number
```
NullSoldier, line 26: Our other API's that stream ranges returns the end inclusive, so I think we should keep in line with that.

ironfish/src/rpc/routes/chain/getBlocks.ts
```diff
@@ -0,0 +1,100 @@
...
+import { RpcNotFoundError, RpcValidationError } from '../../adapters'
+import { ApiNamespace } from '../namespaces'
+import { routes } from '../router'
+import { RpcBlock, RpcBlockSchema, serializeRpcBlockHeader } from '../types'
+import { RpcTransaction } from './types'
+import { serializeRpcTransaction } from './utils'
+
+const MAX_BLOCKS_RANGE = 30
+
+export type GetBlocksRequest = {
```
NullSoldier, line 18: It looks like we're missing the serialized flag here to be consistent with getBlock().

https://github.com/iron-fish/ironfish/blob/master/ironfish/src/rpc/routes/chain/getBlock.ts#L21

ironfish/src/rpc/routes/chain/getBlocks.ts
```diff
@@ -0,0 +1,100 @@
...
+    for (let seq = request.data.start; seq < request.data.end; seq++) {
+      const block = await getBlockWithSequence(context, seq)
+      blocks.push(block)
+    }
+
+    request.end({ blocks })
+  },
+)
+
+const getBlockWithSequence = async (node: FullNode, sequence: number): Promise<RpcBlock> => {
```
NullSoldier, line 72: nit: we use the word `getBlockAtSequence` like the function you called on line 73 here https://github.com/iron-fish/ironfish/blob/8b26bfead96c577399e4cccd5879dcb785194307/ironfish/src/blockchain/blockchain.ts#L1062

So this should be named `getBlockAtSequence`

ironfish/src/rpc/routes/chain/getBlocks.ts
```diff
@@ -0,0 +1,100 @@
...
+    Assert.isInstanceOf(context, FullNode)
+
+    // Use negative numbers to start from the genesis block of the chain
+    if (request.data.start <= 0) {
+      request.data.start = Math.max(request.data.start, GENESIS_BLOCK_SEQUENCE)
+    }
+
+    if (request.data.end - request.data.start > MAX_BLOCKS_RANGE) {
+      request.data.end = request.data.start + MAX_BLOCKS_RANGE
+    }
```
NullSoldier, line 60: If you you add the max support to this function you can actually use, https://github.com/iron-fish/ironfish/blob/master/ironfish/src/utils/blockchain.ts#L9

ironfish/src/rpc/routes/chain/getBlocks.ts
NullSoldier: I think it's ok if we remove it, since you are the only one that can hit your own RPC anyway.

No one else can DDOS your node.

dguenther (approved):
Will wait for @NullSoldier to take another peek before merging.

NullSoldier (approved):

PR #4949, opened 2024-05-07
Rahul/ifl 2559 import ledger view only account

patnir:
## Summary

Adds the `--ledger` flag to `ironfish wallet:import` to import a view-only account from a connected ledger device

## Testing Plan

1. Connect your ledger device to your machine. 
2. Download the latest Ironfish ledger app v0.0.7 
3. Use `ironfish wallet:import --ledger` to import the view only keys from the device 
4. Try breaking the flow by doing the following: unplugging the device during import, not opening the ironfish app, doing this operation when the device is locked, try using the `--name` flag with the name of an account for an existing app, etc.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -20,6 +27,12 @@ export class ImportCommand extends IronfishCommand {
     path: Flags.string({
       description: 'the path to the file containing the account to import',
     }),
```
jowparks, line 29: Is it possible to check the USB devices automatically  instead of requiring a flag?  We could download the public address and see if it is already imported, if not then prompt whether they want to import from ledger. If it is already imported, we could skip this check. Could be fine to have a `--ledger` flag for forcing, but would be nice if the user doesn't need to know this exists.

ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -34,24 +47,103 @@ export class ImportCommand extends IronfishCommand {
     },
   ]
 
+  async importLedger(): Promise<RpcAccountImport> {
+    const PATH = "m/44'/1338'/0"
```
jowparks, line 51: Won't this path change depending on how many accounts are present for a ledger device? For example will last bit be `../1` for a second account?

ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -34,24 +47,103 @@ export class ImportCommand extends IronfishCommand {
...
+
+    const transport = await TransportNodeHid.create(3000, 3000)
+
+    const app = new IronfishApp(transport)
+
+    const appInfo = await app.appInfo()
+    if (appInfo.appName !== 'Ironfish') {
+      this.error('Please open the Iron Fish app on your ledger device')
```
jowparks, line 59: Why do we need to error out here? Can't we just wait for them to open the app and then continue?

PR #4950, opened 2024-05-07
Rahul/validate correct public address on import

patnir:
## Summary

Validates public address is accurate for corresponding keys on import 

Adds a public to the rust node js bindings that returns a public address from an IVK.

## Testing Plan

Manual testing + unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/validator.ts
+++ b/ironfish/src/wallet/validator.ts
@@ -2,7 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-import { isValidPublicAddress as nativeIsValidPublicAddress } from '@ironfish/rust-nodejs'
+import {
+  generatePublicAddressFromIncomingViewKey,
+  isValidPublicAddress as nativeIsValidPublicAddress,
+} from '@ironfish/rust-nodejs'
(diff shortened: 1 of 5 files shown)


ironfish/src/wallet/validator.ts
```diff
@@ -76,6 +83,12 @@ export function validateAccount(toImport: Partial<AccountValue>): void {
   if (toImport.spendingKey && !isValidSpendingKey(toImport.spendingKey)) {
     throw new Error(`Provided spending key ${toImport.spendingKey} is invalid`)
   }
+
+  if (isValidIVKAndPublicAddressPair(toImport.incomingViewKey, toImport.publicAddress)) {
+    throw new Error(
+      `Public address ${toImport.publicAddress} does not match public address generated from incoming view key ${toImport.incomingViewKey}`,
```
dguenther, line 89: It might be nice to also log the address generated from the incoming view key -- might give someone a clue as to why they don't match.

dguenther (approved):
Left one suggestion but otherwise seems like a worthwhile improvement to me!

ironfish/src/wallet/validator.ts
patnir: Done!

PR #4951, opened 2024-05-08
Add wallet/setAccountHead RPC

dguenther:
## Summary

Implements an RPC for manually updating an account's head. The user can pass in hashes of transactions known to be owned by the account, and the wallet will decrypt and apply only those transactions to the account's state.

## Needs Review

* [x] ~Should the name of the RPC route be different?~ Renamed from wallet/addKnownTransactions to wallet/setAccountHead
* [x] ~Should this logic live somewhere other than an RPC route, e.g. integrated with the wallet?~ We might move this out into the Scanner once it's refactored out of the wallet.
* [x] ~Should the route have wallet context instead of fullnode?~ Decided this is fine for now since it's similar to a service command anyway.
* [x] ~What should we do if someone cancels an RPC request before updateHead finishes?~ Added early-exit to the loops by checking `request.closed`.

## Testing Plan

Added automated tests for the RPC.

#### Manual testing

1. Run `wallet:scanning:off` to disable scanning for an account
1. Run `wallet:transactions` on the account. Save all transaction hashes.
1. Run `wallet:status`. Save the account's head hash as `end`
1. Run `wallet:rescan` to reset the account (#4985 will allow you to do this for one account). Cancel the scan
1. Run `wallet:status` again. Save the account's head hash as `start`
1. Run `start` on the directory.
1. In a separate prompt, run `repl` on the directory. Execute the command with the values you previously saved.
```
client = await sdk.connectRpc()
(await client.wallet.addKnownTransactions({ account: "account", start: "startHash", end: "endHash", transactions: [{ hash: "ceddfc8c0978db964f8fb3ad66f4d127235dcd25385401e2cd092a7c088c8992" }] })).content
```

After running `wallet:status` again, you should see the account restored to its former state.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/wallet/addKnownTransactions.ts
```diff
@@ -0,0 +1,249 @@
...
+import { ApiNamespace } from '../namespaces'
+import { routes } from '../router'
+import { getAccount } from './utils'
+
+export type AddKnownTransactionsRequest = {
+  /**
+   * Name of the account to update.
+   */
+  account: string
+  start: string
```
NullSoldier, line 19: Are these hashes?

ironfish/src/wallet/wallet.ts
```diff
@@ -485,42 +485,59 @@ export class Wallet {
...
-            tx,
-          )
-        }
+      const accountTxnsMap = decryptedNotesMap.get(account.id)
+      const txns = transactions.map((t) => ({
+        transaction: t.transaction,
+        decryptedNotes: accountTxnsMap?.get(t.transaction.hash()) ?? [],
+      }))
+
+      await this.connectBlockForAccount(
```
NullSoldier, line 494: Note: the lock is moving down into this function now

ironfish/src/rpc/routes/wallet/addKnownTransactions.ts
dguenther: Yep they are

ironfish/src/rpc/routes/wallet/addKnownTransactions.ts
```diff
@@ -0,0 +1,249 @@
...
+      //    rewind accountHead to startHeader.previousBlockHash. Note that we must rewind, we cannot
+      //    ignore blocks in this case because the accountHead is on a different chain.
+      //  - Otherwise there's a gap between accountHead and startHeader, so abort
+      if (!fork.equals(startHeader) && !fork.hash.equals(startHeader.previousBlockHash)) {
+        const nextHash = (await context.chain.getNextHash(fork.hash)) ?? fork.hash
+        throw new RpcValidationError(`Start must be ${nextHash?.toString('hex')} or earlier.`)
+      }
+
+      while (accountHead !== null && !accountHead.hash.equals(startHeader.previousBlockHash)) {
+        const header: BlockHeader | null = await context.chain.getHeader(accountHead.hash)
```
NullSoldier, line 116: Is this a typescript bug? I cannot figure out why TS interprets this type as `any` without this type annotation.

NullSoldier:
This PR looks extremely well written to me. The tests are very easy to verify. My only comment here would be, we should abort eagerly when we can if the request is closed so we don't continue using up CPU if the client disconnects.

dguenther:
I added early-exit to both of the loops. Tested it by curling `wallet/addKnownTransactions` with start at block 1 and an account with head at block 500k, then closed the request after a few seconds. Verified that the account head was no longer rolling back to block 1.

ironfish/src/rpc/routes/wallet/addKnownTransactions.ts
```diff
@@ -113,6 +121,11 @@ routes.register<typeof AddKnownTransactionsRequestSchema, AddKnownTransactionsRe
       }
 
       while (accountHead !== null && !accountHead.hash.equals(startHeader.previousBlockHash)) {
+        if (request.closed) {
+          request.end()
```
NullSoldier, line 125: you don't need to call request.end() here when the request is already closed

NullSoldier (approved):

PR #4953, opened 2024-05-10
add public account description for multisig bridge integration

jowparks:
## Summary
- Creates a separate description for creation of these public accounts. It is possible to combine them all into a single description, but it seemed cleaner to have them separated, particularly for the verification step.
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/public_account/description.rs
+++ b/ironfish-rust/src/public_account/description.rs
@@ -3,14 +3,15 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 use ed25519_dalek::{Signature, Verifier, VerifyingKey};
-use std::io;
+use std::{collections::HashMap, io};
 
 use crate::{
     errors::{IronfishError, IronfishErrorKind},
     PublicAddress,
 };
 
-use super::transfer::{Transfer};
+use super::transfer::Transfer;
+use std::collections::HashSet;
 
 #[derive(Clone)]
 pub struct PublicAccountDescription {
@@ -33,17 +34,17 @@ impl PublicAccountDescription {
     pub fn new(
         version: u8,
         min_signers: u16,
-        signers: Vec<VerifyingKey>,
-        transfers: Vec<Transfer>,
+        signers:impl IntoIterator<Item = VerifyingKey>,
+        transfers: impl IntoIterator<Item = Transfer>,
         address: PublicAddress,
     ) -> Result<PublicAccountDescription, IronfishError> {
         let description = Self {
             version,
             min_signers,
-            signers,
+            signers: signers.into_iter().collect(),
             signatures: vec![],
             address,
-            transfers,
+            transfers: transfers.into_iter().collect(),
         };
 
         description.valid()?;
@@ -148,33 +149,32 @@ impl PublicAccountDescription {
             return Err(IronfishError::new(IronfishErrorKind::InvalidThreshold));
         }
 
-        let unique_signers: Vec<_> = self.signers.iter().collect();
+        let unique_signers: HashSet<_> = self.signers.iter().collect();
         if unique_signers.len() != self.signers.len() {
             return Err(IronfishError::new(IronfishErrorKind::DuplicateSigner));
-        }
-
-        let unique_signatures: Vec<_> = self.signatures.iter().collect();
-        if unique_signatures.len() != self.signatures.len() {
-            return Err(IronfishError::new(IronfishErrorKind::DuplicateSignature));
         }
 
         Ok(())
     }
 
-    fn verify(&self) -> Result<(), IronfishError> {
+    fn verify(&self, signatures: &Vec<Signature>) -> Result<(), IronfishError> {
         self.valid()?;
 
-        let hash = &self.hash()?;
-        for signature in &self.signatures {
-            let is_valid = self
+        let mut signers = HashMap::new();
+        let hash = self.hash()?;
+        for signature in signatures {
+            let signer = self
                 .signers
                 .iter()
-                .any(|signer| signer.verify(hash, signature).is_ok());
-
-            if !is_valid {
-                return Err(IronfishError::new(IronfishErrorKind::InvalidSignature));
-            }
+                .filter(|signer| signer.verify(&hash, &signature).is_ok())
+                .next();
+            match signer {
+                None => return Err(IronfishError::new(IronfishErrorKind::InvalidSignature)),
+                Some(signer) => signers.entry(signer.clone()).and_modify(|count| *count += 1).or_insert(1),
+            };
         }
+        if signers.len() < self.min_signers.into() { return Err(IronfishError::new(IronfishErrorKind::SignatureThresholdNotMet)) }
+        if signers.values().any(|count| *count > 1) { return Err(IronfishError::new(IronfishErrorKind::DuplicateSigner)) }
         Ok(())
     }
 
@@ -191,15 +191,16 @@ impl PublicAccountDescription {
         let transfer_len = self.transfers.len() as u16;
         hasher.update(&transfer_len.to_le_bytes());
         for transfer in &self.transfers {
-            hasher.update(&transfer.as_bytes()?);
+            hasher.update(&transfer.to_bytes()?);
         }
         hasher.update(&self.address.public_address());
         Ok(hasher.finalize().into())
     }
 
-    pub fn add_signatures(&mut self, signatures: Vec<Signature>) -> Result<(), IronfishError> {
+    pub fn sign(&mut self, signatures: &Vec<Signature>) -> Result<(), IronfishError> {
+        self.verify(signatures)?;
         self.signatures.extend(signatures);
-        self.verify()
+        Ok(())
     }
 
     pub fn version(&self) -> u8 {
@@ -260,8 +261,10 @@ mod tests {
         let signature = signing_key.sign(&hash);
 
         original
-            .add_signatures(vec![signature])
+            .sign(&vec![signature])
             .expect("Should be valid/verified creation");
+        
+        assert_eq!(original.signatures.len(), 1);
 
         let mut buffer = Vec::new();
         original.write(&mut buffer).unwrap();
--- a/ironfish-rust/src/public_account/description.rs
+++ b/ironfish-rust/src/public_account/description.rs
@@ -197,8 +197,9 @@ impl PublicAccountDescription {
         Ok(hasher.finalize().into())
     }
 
-    pub fn sign(&mut self, signatures: &Vec<Signature>) -> Result<(), IronfishError> {
-        self.verify(signatures)?;
+    pub fn sign(&mut self, signatures: impl IntoIterator<Item = Signature>) -> Result<(), IronfishError> {
+        let signatures = signatures.into_iter().collect::<Vec<_>>();
+        self.verify(&signatures)?;
         self.signatures.extend(signatures);
         Ok(())
     }
@@ -211,15 +212,15 @@ impl PublicAccountDescription {
         self.min_signers
     }
 
-    pub fn signers(&self) -> &Vec<VerifyingKey> {
+    pub fn signers(&self) -> &[VerifyingKey] {
         &self.signers
     }
 
-    pub fn signatures(&self) -> &Vec<Signature> {
+    pub fn signatures(&self) -> &[Signature] {
         &self.signatures
     }
 
-    pub fn transfers(&self) -> &Vec<Transfer> {
+    pub fn transfers(&self) -> &[Transfer] {
         &self.transfers
     }
 
@@ -261,7 +262,7 @@ mod tests {
         let signature = signing_key.sign(&hash);
 
         original
-            .sign(&vec![signature])
+            .sign(vec![signature])
             .expect("Should be valid/verified creation");
         
         assert_eq!(original.signatures.len(), 1);
--- a/ironfish-rust/src/public_account/description.rs
+++ b/ironfish-rust/src/public_account/description.rs
@@ -13,6 +13,7 @@ use super::transfer::Transfer;
 pub struct PublicAccountDescription {
     /// TODO(jwp): do we need a separate account address for later migration
     /// pub(crate) account_address: EthereumAddress
+    pub(crate) version: u8,
     // Minimum number of required signers
     pub(crate) threshold: i16,
     // Signers of the public account
@@ -25,6 +26,10 @@ pub struct PublicAccountDescription {
 
 impl PublicAccountDescription {
     pub fn read<R: io::Read>(mut reader: R) -> Result<Self, IronfishError> {
+        let mut version_buf = [0; 1];
+        reader.read_exact(&mut version_buf)?;
+        let version = version_buf[0];
+        
         let mut threshold_buf = [0; 2];
         reader.read_exact(&mut threshold_buf)?;
         let threshold = i16::from_le_bytes(threshold_buf);
@@ -65,6 +70,7 @@ impl PublicAccountDescription {
         }
 
         Ok(Self {
+            version,
             threshold,
             signers,
             signatures,
@@ -73,6 +79,8 @@ impl PublicAccountDescription {
     }
 
     pub fn write<W: io::Write>(&self, mut writer: W) -> Result<(), IronfishError> {
+        writer.write_all(&self.version.to_le_bytes())?;
+        
         writer.write_all(&self.threshold.to_le_bytes())?;
 
         let signers_len = self.signers.len() as i16;
@@ -119,7 +127,7 @@ impl PublicAccountDescription {
         self.valid()?;
 
         let hash =
-            &PublicAccountDescription::hash(&self.threshold, &self.signers, &self.transfers)?;
+            &PublicAccountDescription::hash(&self.version, &self.threshold, &self.signers, &self.transfers)?;
         let is_valid = self
             .signers
             .iter()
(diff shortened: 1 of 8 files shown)


ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,218 @@
...
+
+use super::transfer::Transfer;
+
+#[derive(Clone)]
+pub struct PublicAccountDescription {
+    /// TODO(jwp): do we need a separate account address for later migration
+    /// pub(crate) account_address: EthereumAddress
+    pub(crate) version: u8,
+    // Minimum number of required signers
+    pub(crate) threshold: i16,
```
andiflabs, line 18: Shouldn't this be a `u16`? What would be the meaning of a negative threshold?

ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,218 @@
...
+        reader.read_exact(&mut version_buf)?;
+        let version = version_buf[0];
+        
+        let mut threshold_buf = [0; 2];
+        reader.read_exact(&mut threshold_buf)?;
+        let threshold = i16::from_le_bytes(threshold_buf);
+
+        let mut signers_len_buf = [0; 2];
+        reader.read_exact(&mut signers_len_buf)?;
+        let signers_len = i16::from_le_bytes(signers_len_buf) as usize;
```
andiflabs, line 41: `u16`?

ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,218 @@
...
+            let mut signer_buf = [0; 32];
+            reader.read_exact(&mut signer_buf)?;
+            let signer = VerifyingKey::from_bytes(&signer_buf)
+                .map_err(|_| IronfishError::new(IronfishErrorKind::InvalidData))?;
+            signers.push(signer);
+        }
+
+        let mut signatures_len_buf = [0; 2];
+        reader.read_exact(&mut signatures_len_buf)?;
+        let signatures_len = i16::from_le_bytes(signatures_len_buf) as usize;
```
andiflabs, line 54: `u16`?

ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,218 @@
...
+        for _ in 0..signatures_len {
+            let mut signature_buf = [0; 64];
+            reader.read_exact(&mut signature_buf)?;
+            let signature = Signature::from_bytes(&signature_buf);
+            signatures.push(signature);
+        }
+
+        let mut transfers_len_buf = [0; 2];
+        reader.read_exact(&mut transfers_len_buf)?;
+        let transfers_len = i16::from_le_bytes(transfers_len_buf) as usize;
```
andiflabs, line 66: `u16`?

ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,218 @@
...
+        })
+    }
+
+    pub fn write<W: io::Write>(&self, mut writer: W) -> Result<(), IronfishError> {
+        // TODO: think about ideal ordering here
+        writer.write_all(&self.version.to_le_bytes())?;
+        
+        writer.write_all(&self.threshold.to_le_bytes())?;
+
+        let signers_len = self.signers.len() as i16;
```
andiflabs, line 92: Note that the result of `.len()` may not fit into a `u16`. It'd be best to use `.len().try_into()` and return an error in the unlikely event that the vector is too long

ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,218 @@
...
+        for transfer in &self.transfers {
+            transfer.write(&mut writer)?;
+        }
+
+        self.address.write(&mut writer)?;
+
+        Ok(())
+    }
+
+    pub fn valid(&self) -> Result<(), IronfishError> {
```
andiflabs, line 118: We shouldn't allow to instantiate an invalid struct. This method should be private and should be automatically called by all other methods that create a new `PublicAccountDescription` (such as `read()`)

ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        if self.signatures.len() < self.threshold as usize {
+            return Err(IronfishError::new(IronfishErrorKind::InvalidData));
+        }
+
+        Ok(())
+    }
+
+    pub fn verify(&self) -> Result<(), IronfishError> {
```
andiflabs, line 134: I'd also call this implicitly on instantiation

ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,218 @@
...
+        }
+        Ok(())
+    }
+
+    pub fn hash(
+        version: &u8,
+        threshold: &i16,
+        signers: &Vec<VerifyingKey>,
+        address: &PublicAddress,
+        transfers: &Vec<Transfer>,
```
andiflabs, line 155: Why not make this a `&self`?

ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,218 @@
...
+        // TODO(jwp): verify which hashers supported by axelar
+        let mut hasher = blake3::Hasher::new();
+        hasher.update(&version.to_le_bytes());
+        hasher.update(&threshold.to_le_bytes());
+        for signer in signers {
+            hasher.update(signer.as_bytes());
+        }
+        for transfer in transfers {
+            hasher.update(&transfer.as_bytes()?);
+        }
```
andiflabs, line 166: Variable-length items should be prefixed by their length. This makes the input to the hash "prefix free".

Not including a length prefix can potentially allow somebody to construct two distinct inputs that hash the same. For example, input 1 could be `signer 1 || signer 2 || transfer 1 || transfer 2`, and input 2 could be `singer 1 || signer 2 || signer 3 || transfer 2`, where `transfer 1` and `signer 3` have the same contents. The contents are semantically different, but would result in the same hash.

While this specific example might not result in a successful attack in this context, it's best to entirely close the door to this class of attacks.

ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,218 @@
...
+mod tests {
+    use crate::{assets::asset_identifier, public_account::transfer::PublicMemo, SaplingKey};
+
+    use super::*;
+    use ed25519_dalek::{ed25519::signature::SignerMut, SigningKey};
+    use rand::rngs::OsRng;
+
+    #[test]
+    fn test_public_account_create_description() {
+        let mut csprng = OsRng {};
```
andiflabs, line 182: Nit: use `thread_rng()` instead

ironfish-rust/src/public_account/transfer.rs
```diff
@@ -0,0 +1,53 @@
...
+
+use crate::{assets::asset_identifier::AssetIdentifier, errors::IronfishError, PublicAddress};
+
+#[derive(Clone, Copy)]
+pub struct PublicMemo(pub [u8; 256]);
+
+#[derive(Clone, Copy)]
+pub struct Transfer {
+    pub(crate) asset_id: AssetIdentifier,
+    pub(crate) amount: u64,
```
andiflabs, line 11: Wondering if this should be signed (`i64`)? How do we distinguish transfers in and out? Is there going to be another struct?

ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,218 @@
...
+        Ok(hasher.finalize().into())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::{assets::asset_identifier, public_account::transfer::PublicMemo, SaplingKey};
+
+    use super::*;
+    use ed25519_dalek::{ed25519::signature::SignerMut, SigningKey};
```
andiflabs, line 177: Use `Signer` instead of `SignerMut`

ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,218 @@
...
+    pub fn verify(&self) -> Result<(), IronfishError> {
+        self.valid()?;
+
+        let hash =
+            &PublicAccountDescription::hash(&self.version, &self.threshold, &self.signers, &self.address, &self.transfers)?;
+        let is_valid = self
+            .signers
+            .iter()
+            .zip(&self.signatures)
+            .any(|(signer, signature)| signer.verify(hash, signature).is_ok());
```
andiflabs, line 143: This should be `.all()`, not `.any()`

ironfish-rust/src/public_account/description.rs
jowparks: Good call, I would have expected rust to catch this, seems like it is better to avoid explicit type casting for this reason.

ironfish-rust/src/public_account/description.rs
jowparks: Verify will not be applicable to transfer cases where you are sending funds to the public account. We don't need signatures for that to happen, that is why I split this into a separate method.

ironfish-rust/src/public_account/description.rs
jowparks: I think my method was actually just wrong here, I don't know what order the signatures are going to be in, so I reworked to use a loop instead.

ironfish-rust/src/public_account/description.rs
jowparks: Sg, I can do that as long as I instantiate with an empty vec for signatures (instead of requiring them to be present on instantiation). I updated to do that.

ironfish-rust/src/public_account/description.rs
jowparks: Great call!

ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,275 @@
...
+        min_signers: u16,
+        signers: Vec<VerifyingKey>,
+        transfers: Vec<Transfer>,
+        address: PublicAddress,
+    ) -> Result<PublicAccountDescription, IronfishError> {
+        let description = Self {
+            version,
+            min_signers,
+            signers,
+            signatures: vec![],
```
jowparks, line 44: No signatures when instantiating (need to do this if we want to be able to use `hash` with `&self`)

ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,275 @@
...
+    errors::{IronfishError, IronfishErrorKind},
+    PublicAddress,
+};
+
+use super::transfer::{Transfer};
+
+#[derive(Clone)]
+pub struct PublicAccountDescription {
+    /// TODO(jwp): do we need a separate account address for later migration
+    /// pub(crate) account_address: EthereumAddress
```
jowparks, line 18: Made these private so we have to use constructor

ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,275 @@
...
+    // address of the account
+    address: PublicAddress,
+}
+
+impl PublicAccountDescription {
+    pub fn new(
+        version: u8,
+        min_signers: u16,
+        signers: Vec<VerifyingKey>,
+        transfers: Vec<Transfer>,
```
andiflabs, line 37: Nit: `Vec<T>` -> `impl IntoIterator<Item = T>`

ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,275 @@
...
+    fn valid(&self) -> Result<(), IronfishError> {
+        if self.min_signers < 1 {
+            return Err(IronfishError::new(IronfishErrorKind::InvalidThreshold));
+        }
+
+        if self.signers.len() < self.min_signers as usize {
+            return Err(IronfishError::new(IronfishErrorKind::InvalidThreshold));
+        }
+
+        let unique_signers: Vec<_> = self.signers.iter().collect();
```
andiflabs, line 151: This is not removing any duplicates, it's only creating a copy of `self.signers` in a slower way than `.clone()`.

I think you meant something like:

```rust
let mut unique_signers = self.signers.clone();
unique_signers.sort_unstable();
unique_signers.dedup();
```

ironfish-rust/src/public_account/transfer.rs
```diff
@@ -0,0 +1,53 @@
...
+    pub fn write<W: io::Write>(&self, mut writer: W) -> Result<(), IronfishError> {
+        self.asset_id.write(&mut writer)?;
+        writer.write_all(&self.amount.to_le_bytes())?;
+        self.to.write(&mut writer)?;
+        writer.write_all(&self.memo.0)?;
+
+        Ok(())
+    }
+
+    pub fn as_bytes(&self) -> Result<Vec<u8>, IronfishError> {
```
andiflabs, line 48: Nit: by convention, `as_*` methods perform zero-cost conversion. This method is performing an allocation + a serialization, so it's quite expensive. It should be better called `to_bytes` instead. See:

https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

ironfish-rust/src/public_account/transfer.rs
```diff
@@ -0,0 +1,53 @@
...
+        let mut bytes = Vec::new();
```
andiflabs, line 49: It looks to me like the `write()` method is only writing fixed-length bytes. If that's the case, I would consider returning an array instead of a `Vec`

ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,275 @@
...
+
+    pub fn version(&self) -> u8 {
+        self.version
+    }
+
+    pub fn min_signers(&self) -> u16 {
+        self.min_signers
+    }
+
+    pub fn signers(&self) -> &Vec<VerifyingKey> {
```
andiflabs, line 213: Nit: `-> &[VerifyingKey]`

ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,275 @@
...
+        &self.signers
+    }
+
+    pub fn signatures(&self) -> &Vec<Signature> {
```
andiflabs, line 217: Nit: `-> &[Signature]`

ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,275 @@
...
+        &self.signatures
+    }
+
+    pub fn transfers(&self) -> &Vec<Transfer> {
```
andiflabs, line 221: Nit: `-> &[Transfer]`

ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,275 @@
...
+        let transfer_len = self.transfers.len() as u16;
+        hasher.update(&transfer_len.to_le_bytes());
+        for transfer in &self.transfers {
+            hasher.update(&transfer.as_bytes()?);
+        }
+        hasher.update(&self.address.public_address());
+        Ok(hasher.finalize().into())
+    }
+
+    pub fn add_signatures(&mut self, signatures: Vec<Signature>) -> Result<(), IronfishError> {
```
andiflabs, line 200: Nit: `Vec` -> `IntoIterator`

ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,275 @@
...
+        self.signatures.extend(signatures);
+        self.verify()
```
andiflabs, line 202: Note that if the verification fails, `signatures` will still include the invalid signature. We should make sure that `signatures` does not get modified if it's invalid.

I suggest changing `verify(&self)` to `verify_signature(&self, signature: &Signature)` (or similar) and calling it before updating `self.signatures`

ironfish-rust/src/public_account/description.rs
```diff
@@ -0,0 +1,275 @@
...
+            return Err(IronfishError::new(IronfishErrorKind::InvalidThreshold));
+        }
+
+        let unique_signers: Vec<_> = self.signers.iter().collect();
+        if unique_signers.len() != self.signers.len() {
+            return Err(IronfishError::new(IronfishErrorKind::DuplicateSigner));
+        }
+
+        let unique_signatures: Vec<_> = self.signatures.iter().collect();
+        if unique_signatures.len() != self.signatures.len() {
```
andiflabs, line 157: Note that this works only because Ed25519 signatures are deterministic. If we were using a different signature scheme (e.g. ECDSA), then this wouldn't guarantee that a single signer didn't add multiple signatures, which would make the `min_signatures` test pass even though the threshold wasn't met.

To prevent this kind of problems, I would count the number of signers that contributed a signature instead (here or in the `verify` method):

```rust
let signers = HashMap::new();
for signature in &self.signatures {
    let signer = self
        .signers
        .iter()
        .filter(|signer| signer.verify(...))
        .next();
    match signer {
        None => return Err(...),
        Some(signer) => signers.entry(signer.clone()).and_modify(|count| count += 1).or_insert(1),
    }
}
```

ironfish-rust/src/public_account/description.rs
andiflabs: Given that there are multiple use cases supported in this struct, I wonder if we should restructure it by using an `enum`. For example we could have an `enum` variant with signers, and another without.

ironfish-rust/src/public_account/transfer.rs
jowparks: Oh wow, thanks for the advice, I didn't know this convention!

ironfish-rust/src/public_account/description.rs
jowparks: Ty master jedi

ironfish-rust/src/public_account/description.rs
jowparks: I meant to write:
```
let unique_signers: HashSet<_> = self.signers.iter().collect();
```
But your solution below involving the signer check is clearly better, I will go with that.

ironfish-rust/src/public_account/description.rs
jowparks: What would this look like? Do we have an example of this in our codebase?

jowparks:
won't do, closing for now

PR #4955, opened 2024-05-10
Remove version account version default

NullSoldier:
## Summary

This code would never execute.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/types.ts
+++ b/ironfish/src/rpc/routes/wallet/types.ts
@@ -163,14 +163,14 @@ export type RpcMultisigKeys = {
 }
 
 export type RpcAccountImport = {
-  version: number
+  version?: number
   name: string
-  viewKey: string
+  viewKey?: string
   incomingViewKey: string
   outgoingViewKey: string
   publicAddress: string
   spendingKey: string | null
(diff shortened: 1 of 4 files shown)


dguenther (approved):

ironfish/src/rpc/routes/wallet/types.ts
```diff
@@ -163,14 +163,14 @@ export type RpcMultisigKeys = {
...
+  version?: number
   name: string
-  viewKey: string
+  viewKey?: string
   incomingViewKey: string
   outgoingViewKey: string
   publicAddress: string
   spendingKey: string | null
-  createdAt: { hash: string; sequence: number } | null
+  createdAt: { hash: string; sequence: number } | string | null
```
hughy, line 173: Is this to support an empty string? It doesn't look like we do anything to parse a string passed here

hughy (approved):

NullSoldier:
After much discussion, we're going to treat this as a one part of a cleanup fix the issue where RpcAccountImport is used as the baes JSON serialized type in the account exporting system which should not happen. More PR's coming soon.

PR #4957, opened 2024-05-11
Update description of wallet:rescan

dguenther:
## Summary

I'm open to opinions on whether this is more clear or not, but I was a bit confused by what was meant by a full scan. It looks like it only deletes createdAt from all accounts in the wallet before resetting them.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/rescan.ts
+++ b/ironfish-cli/src/commands/wallet/rescan.ts
@@ -29,7 +29,7 @@ export class RescanCommand extends IronfishCommand {
       hidden: true,
     }),
     full: Flags.boolean({
-      description: 'Force a full rescan of the chain starting from the genesis block',
+      description: 'Remove `createdAt` from accounts before rescanning',
     }),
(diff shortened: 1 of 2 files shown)


hughy (commented):
I think that deleting createdAt more closely matches _what_ this flag does, but scanning the chain from the genesis block (a "full scan") is _why_ it does that.

I'm not sure which would be better or more informative for CLI users

andiflabs:
How about something like this?

> Forces a rescan of the chain starting from the genesis block, ignoring the `createdAt` field of the account

dguenther:
Nice, I like that. I was confused by the difference between `from` and `full` (`from` is hidden). I think generally CLI users will want to know "why would I want to use this flag", and Hugh explained a potential case in multisig for this, so I tried to add a sentence as to why you'd want to pass `full` as well.

hughy (approved):

PR #4963, opened 2024-05-14
Separate out block getter from RPC serialization

NullSoldier:
## Summary
This adds a new method to the block chain by adding a new matching
method getBlockAtSequence which matches getHeaderAtSequence. It also
does not conflate data fetching with serialization so getBlocks is a lot
cleaner now.

## Testing Plan
Added a new test

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/chain/utils.ts
+++ b/ironfish/src/rpc/routes/chain/utils.ts
@@ -180,6 +180,7 @@ export const serializeRpcBlock = (block: Block, serialized?: boolean): RpcBlock 
     transactions,
   }
 }
+
 export const serializeRpcTransaction = (
   tx: Transaction,
   serialized?: boolean,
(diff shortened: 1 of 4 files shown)


ironfish/src/rpc/routes/chain/utils.ts
```diff
@@ -165,6 +166,20 @@ export async function renderGraph(
...
+  for (const tx of block.transactions) {
+    transactions.push(serializeRpcTransaction(tx, serialized))
+  }
+
+  return {
+    ...blockHeaderResponse,
+    size: getBlockSize(block),
+    transactions,
+  }
+}
```
dguenther, line 182: Newline here might be nice:

```suggestion
}

```

dguenther (approved):

PR #4970, opened 2024-05-20
determines raw tx sender from tx fields

hughy:
## Summary

uses the spends, outputs, and/or mints in a raw transaction to determine which account to use to post the transaction.

deprecates the account flag from the 'wallet:post' CLI command

deprecates the account request parameter from the 'wallet/postTransaction' RPC

## Testing Plan

manual testing

## Documentation

```
[X] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/primitives/rawTransaction.ts
```diff
@@ -187,6 +187,16 @@ export class RawTransaction {
...
     return posted
   }
+
+  sender(): string | undefined {
+    if (this.spends.length > 0) {
+      return this.spends[0].note.owner()
+    } else if (this.outputs.length > 0) {
+      return this.outputs[0].note.sender()
+    } else if (this.mints.length > 0) {
+      return this.mints[0].creator
```
mat-if, line 197: This should _technically_ be `.owner`, but I see that `RawTransaction` lacks that information right now. We'll need to tackle that if/when ownership v2 comes around. No action needed here, just mentioning it

mat-if (approved):

PR #4973, opened 2024-05-22
Remove AccountValue from account importer/exporter

NullSoldier:
## Summary

This changes all exporter / importer infrastructure to operate off of `AccountImport` and not `AccountValue`.

You will need to convert an Account to an AccountImport to serialize it now. So the flow is `AccountValue` -> `Account` -> `AccountImport` -> `string` -> `AccountImport` -> `Account` -> `AccountValue`.

The exporter and importer infrastructure for accounts should not use `AccountValue`. If you re-use `AccountValue` for exporting this will change the export format and break compatability with accounts if you add new fields. If you change the `AccountValue` fields, nothing should change about exported accounts automatically.

## Testing Plan
Existing tests are adequate.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/wallet/utils.ts
```diff
@@ -101,6 +101,28 @@ export async function serializeRpcWalletTransaction(
   }
 }
 
+export const serializeRpcImportAccount = (encoded: AccountImport): RpcAccountImport => {
```
dguenther, line 104: Maybe name this `accountImport` instead of `encoded`?

ironfish/src/rpc/routes/wallet/exportAccount.ts
```diff
@@ -43,31 +48,29 @@ routes.register<typeof ExportAccountRequestSchema, ExportAccountResponse>(
...
+
       request.end({ account: encoded })
+      return
     }
+
+    // Backwards compatability path we must send back an RpcAccountImport
+
+    const serialized = serializeRpcImportAccount(
+      decodeAccountImport(
+        encodeAccountImport(toAccountImport(account), AccountFormat.JSON, {
```
dguenther, line 68: Does this need to encode/decode the account import? Looks like this would allow for adding support for viewOnly, but the old codepath didn't have that. I'm okay either way, just noticed the nested functions

ironfish/src/wallet/exporter/accountImport.ts
```diff
@@ -0,0 +1,34 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Account } from '../account/account'
+import { MultisigKeysImport } from '../interfaces/multisigKeys'
+import { HeadValue } from '../walletdb/headValue'
+
+export type AccountImport = {
```
dguenther, line 8: Might consider adding a comment with some guidance for how this type should be used, for example if `AccountImport` should only be used as an input/output for the encoders.

ironfish/src/wallet/exporter/encoders/base64json.ts
```diff
@@ -19,31 +19,34 @@ export class Base64JsonEncoder implements AccountEncoder {
...
-    } else {
-      throw new Error('Unknown encryption method requested')
+    if (options?.encryptWith) {
+      if (isMultisigIdentityEncryption(options.encryptWith)) {
+        const encrypted = encodeEncryptedMultisigAccount(binary, options.encryptWith)
+        const encoded = encrypted.toString('base64')
+        return `${BASE64_JSON_MULTISIG_ENCRYPTED_ACCOUNT_PREFIX}${encoded}`
+      } else {
+        throw new Error('Unknown encryption method requested')
+      }
```
dguenther, line 29: I think this `else` can be dropped:

```suggestion
      }
      
      throw new Error('Unknown encryption method requested')
```

ironfish/src/rpc/routes/wallet/utils.ts
NullSoldier: done

ironfish/src/rpc/routes/wallet/exportAccount.ts
NullSoldier: I think the old code path supported viewOnly here. https://github.com/iron-fish/ironfish/blob/8b26bfead96c577399e4cccd5879dcb785194307/ironfish/src/rpc/routes/wallet/exportAccount.ts#L47-L54

The problem I'm trying to solve is to remove all the duplicated code for how to encode twice. If I move support for viewOnly to `toAccountImport` it would remove duplication. I was thinking of doing that in the next PR to avoid this.

ironfish/src/wallet/exporter/accountImport.ts
NullSoldier: I'm planning to do that in a file `wallet/exporter/README.md` in another PR but I can also add that here too. Are you  OK if that's in another PR? I'm going to rename this type in another PR next to `AccountExported`

ironfish/src/wallet/exporter/encoders/base64json.ts
NullSoldier: Fixed

dguenther (approved):

ironfish/src/rpc/routes/wallet/exportAccount.ts
NullSoldier: Fixed by moving this into toAccountImport

PR #4978, opened 2024-05-22
Fix migration 031 to use proper multisig encoder

NullSoldier:
## Summary

031 is not using the correct multisig encoder that matches AccountValue. This leads to errors when you try to use the new migration for 31 in any future migration such as 32.

## Testing Plan

Run the migration on an old database.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/migrations/data/031-add-pak-to-account/new/AccountValue.ts
+++ b/ironfish/src/migrations/data/031-add-pak-to-account/new/AccountValue.ts
@@ -5,6 +5,7 @@ import { PUBLIC_ADDRESS_LENGTH } from '@ironfish/rust-nodejs'
 import bufio from 'bufio'
 import { IDatabaseEncoding } from '../../../../storage'
 import { HeadValue, NullableHeadValueEncoding } from './HeadValue'
+import { MultisigKeys, MultisigKeysEncoding } from './MultisigKeys'
 
 const KEY_LENGTH = 32
 export const VIEW_KEY_LENGTH = 64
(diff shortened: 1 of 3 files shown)


ironfish/src/migrations/data/031-add-pak-to-account/new/AccountValue.ts
```diff
@@ -53,8 +51,9 @@ export class AccountValueEncoding implements IDatabaseEncoding<AccountValue> {
     }
 
     if (value.multisigKeys) {
-      bw.writeVarBytes(Buffer.from(value.multisigKeys.secret, 'hex'))
-      bw.writeVarBytes(Buffer.from(value.multisigKeys.keyPackage, 'hex'))
+      const encoding = new MultisigKeysEncoding()
+      bw.writeU64(encoding.getSize(value.multisigKeys))
+      bw.writeBytes(encoding.serialize(value.multisigKeys))
     }
```
patnir, line 57: Does this break the current multisig accounts because of the size also being written?

patnir: Maybe this change doesn't matter because there were no multisig accounts before 31?

NullSoldier: Yes that's correct. If this ran for anyone in the first place, their accounts would not be de-codable. It would crash when they try to start the node.

patnir (approved):
Looks good, thanks Jason for fixing this. Good learning for me

dguenther (approved):

PR #4979, opened 2024-05-23
defaults cli tables to no-truncate

hughy:
## Summary

we typically do not want to truncate data that we display, especially hashes

- wraps oclif 'CliUx.ux.table.flags()' and adds a 'truncate' flag which defaults to false
- includes 'no-truncate' as a hidden flag that defaults to the inverse of the 'truncate' flag. this allows passing table flags to the oclif table constructor without handling our own truncate flag

- replaces all uses of oclif table flags with our own constant

## Testing Plan

manual testing with '--truncate' and '--no-truncate'

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/utils/table.ts
+++ b/ironfish-cli/src/utils/table.ts
@@ -148,16 +148,5 @@ export const TableFlags = {
     default(context) {
       return Promise.resolve(!context.flags['truncate'])
     },
-    relationships: [
-      {
-        type: 'all',
-        flags: [
-          {
-            name: 'truncate',
-            when: async (flags) => Promise.resolve(flags['truncate'] === true),
-          },
-        ],
-      },
-    ],
   }),
 }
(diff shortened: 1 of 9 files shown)


ironfish-cli/src/utils/table.ts
```diff
@@ -132,3 +133,31 @@ export enum Format {
...
+const { 'no-truncate': _, ...tableFlags } = CliUx.ux.table.flags()
+
+export const TableFlags = {
+  ...tableFlags,
+  truncate: Flags.boolean({
+    description: 'truncate output to fit screen',
+    default: false,
+    allowNo: true,
+  }),
+  'no-truncate': Flags.boolean({
```
jowparks, line 146: Looks good, the relationships section seems sort of funny but I am guessing that is some necessary magic for oclif.

jowparks (approved):

PR #4983, opened 2024-05-23
Move wallet scanning code into WalletScanner

NullSoldier:
## Summary

This new wallet scanner tries to move some of the code (excluding
decryption) into a new class called the WalletScanner. This will
act as the holding ground for new scanning related code. This doesn't
exactly clean up a lot of the code, but does start as the beginning of
an attempt to simplify the scanning code.

- Most scanning code now exists in WalletScanner
- Merged Wallet.scanTransactions() into Wallet.scan()
- Renamed Wallet.updateHead() to Wallet.scan()
- The ChainProcessor is no longer exposed publically
- The wallet head now no longer returns anything meaningful in the getStatus()
RPC.
- Effectively deleted `wallet/rescanAccount` in favor of `wallet/rescan`. It rescans all accounts and doesn't apply to one account. Also changed request input.
- `wallet/rescan` nowalways aborts a previous scan no matter what. This is to make the purpose of this command more simple. If you call rescan it _WILL_ rescan.

Fixes IFL-2570

## Testing Plan

 1. Start scanning
 2. Cause rescan
 3. Disable an account
 4. Start another rescan

## Documentation
```
[X] Yes
```

Documentation added here https://github.com/iron-fish/website/pull/704

## Breaking Change
```
[x] Yes
```
- Replaced `wallet/rescanAccount` RPC with `wallet/rescan`
- RPC `node/getStatus` accounts.head.hash` and `accounts.head.sequence` are always `""` and `-1` respectively. The head and sequence is no longer returned because there is no longer a single wallet head. Each account has their own head.


ironfish-cli/src/commands/chain/rewind.ts
```diff
@@ -136,15 +136,17 @@ async function rewindWalletHead(
...
+          speed.add(1)
+          bar.update(++rewound, {
+            speed: speed.rate1s.toFixed(2),
+            estimate: TimeUtils.renderEstimate(rewound, toRewind, speed.rate1m),
+          })
         })
-      })
-
-      await wallet.updateHead()
+      }
```
NullSoldier, line 149: May need to rethink this

ironfish-cli/src/commands/wallet/rescan.ts
```diff
@@ -28,14 +28,11 @@ export class RescanCommand extends IronfishCommand {
       description: 'Sequence to start account rescan from',
       hidden: true,
     }),
-    full: Flags.boolean({
```
NullSoldier, line 31: This flag is deleted in favor of wallet/reset which should be used. Rescan should never delete wallet data like the birthday, only cause a chain rescan.

dguenther: I think we'll want a `reset` CLI command as part of this, so multisig users can easily access it -- maybe it'll be easier to pull it out into a CLI command then

ironfish/src/rpc/routes/wallet/rescan.ts
```diff
@@ -0,0 +1,64 @@
...
+export const RescanResponseSchema: yup.ObjectSchema<RescanResponse> = yup
+  .object({
+    sequence: yup.number().defined(),
+    endSequence: yup.number().defined(),
+    startedAt: yup.number().defined(),
+  })
+  .defined()
+
+routes.register<typeof RescanRequestSchema, RescanResponse>(
+  `${ApiNamespace.wallet}/rescan`,
```
dguenther, line 32: Noting that this is a breaking change renaming `rescanAccount` to `rescan`.

ironfish/src/rpc/routes/wallet/rescan.ts
```diff
@@ -0,0 +1,64 @@
...
+  async (request, context): Promise<void> => {
+    AssertHasRpcContext(request, context, 'wallet', 'logger')
+
+    while (context.wallet.scanner.running) {
+      context.logger.warn('Aborting scanning to start a full wallet rescan.')
+      await context.wallet.scanner.abort()
+    }
+
+    await context.wallet.resetAccounts({
+      resetScanningEnabled: true,
```
dguenther, line 43: I'm not sure if we want to reset `scanningEnabled` here by default, maybe a param for it? At least if we do, we should document that this will resume scanning for all accounts with scanning off.

That said, I feel like this RPC route is near deprecation anyway, since I feel like usually users will only want to rescan one particular account, and you can effectively do that with `wallet/resetAccount`

ironfish/src/rpc/routes/node/getStatus.ts
```diff
@@ -366,12 +364,21 @@ async function getStatus(node: FullNode): Promise<GetNodeStatusResponse> {
       speed: MathUtils.round(node.workerPool.speed?.rate5s ?? 0, 2),
     },
     accounts: {
-      scanning: accountsScanning,
       enabled: node.config.get('enableWallet'),
       head: {
-        hash: node.wallet.chainProcessor.hash?.toString('hex') ?? '',
-        sequence: node.wallet.chainProcessor.sequence ?? -1,
+        hash: '',
+        sequence: -1,
```
dguenther, line 370: Noting that this is a breaking change (though more to what's expected. and it might have been broken already by #4964).

ironfish-cli/src/commands/chain/rewind.ts
dguenther: Maybe this should use the sequence, similar to what `rescan` does?

ironfish/src/wallet/scanner/walletScanner.ts
```diff
@@ -0,0 +1,247 @@
...
+      })
+
+      if (start === null) {
+        start = await this.wallet.getChainGenesis()
+      }
+
+      this.logger.info(`Scan starting from block ${start.sequence} to ${end.sequence}`)
+
+      const scan = new ScanState(start, end)
+      this.state = scan
```
dguenther, line 118: Maybe just assign directly to `this.state`? I don't feel too strongly about this

ironfish/src/wallet/scanner/walletScanner.ts
```diff
@@ -0,0 +1,247 @@
...
+      return null
+    }
+
+    if (this.running && force) {
+      this.logger.debug('Aborting scan in progress and starting new scan.')
+      await this.abort()
+    }
+
+    await this.state?.wait()
+    const unlock = await this.lock.lock()
```
dguenther, line 83: This lock plus the void async code below feels a bit confusing -- it might be nice to either have a test that demonstrates why this is here, or I wonder if we could rework it somehow so `this.scan` acts as a lock.

Kind of thinking of something like this, but it might be pushing too much logic into ScanState
```
await this.state?.wait()
this.state = ScanState.init(async () => {
  // set up scanning
})
```

dguenther (commented):
This definitely feels like an overall improvement, reducing the code in the wallet and de-duping some of the code from `updateHead` and `scanTransaction`.

I have the impression that we could find ways to simplify it further, but I'd have to think about it more, and since this is an overall improvement, I don't think it needs to become a giga-PR 👍 Let me know when tests are fixed up and I'll re-review

ironfish-cli/src/commands/wallet/rescan.ts
NullSoldier: That sounds good, want me to add the reset command to this PR?

ironfish/src/rpc/routes/wallet/rescan.ts
NullSoldier: Added a note for this.

ironfish/src/rpc/routes/node/getStatus.ts
NullSoldier: Added a note for this.

ironfish/src/rpc/routes/wallet/rescan.ts
NullSoldier: I was just keeping backwards compatability with the previous code. I can remove this. I think it should not be reset personally.

ironfish/src/wallet/scanner/walletScanner.ts
NullSoldier: I need to guarantee a reference to this inside the async handler. Maybe if the async piece is cleaned up we can simplify this too.

ironfish-cli/src/commands/wallet/rescan.ts
dguenther: I think it should either be in this PR, or we should pull the removal out into a separate PR, so we're not potentially leaving staging in a state where doing a release would leave a gap in functionality. You could pull the removal out into a separate PR and just temporarily shim it in this PR with `resetAccount({ resetCreatedAt: true })`

ironfish/src/rpc/routes/wallet/rescan.ts
dguenther: I'd lean toward removing it. I don't see where it reset `scanningEnabled` in the previous code, but maybe I'm missing it

ironfish/src/rpc/routes/node/getStatus.ts
dguenther: I'd be okay with removing it. I know it's used in the node app but I'm not sure if any external consumers use it. I almost feel like it'd be better to either remove it, or just mark it as deprecated and replace it with `getEarliestHeadHash` (should be pretty much the same thing)

ironfish/src/wallet/scanner/walletScanner.ts
```diff
@@ -0,0 +1,247 @@
...
+      this.logger.debug(`Scan starting from block ${start.sequence} to ${end.sequence}`)
+
+      const scan = new ScanState(start, end)
+      this.state = scan
+
+      void (async () => {
+        let hashChanged = true
+        while (hashChanged) {
+          const result = await chainProcessor.update({ signal: scan.abortController.signal })
+          hashChanged = result.hashChanged
```
dguenther, line 124: The old version of the scan updated `chainProcessor.hash` to `wallet.getEarliestHeadHash` on every loop. I think we may want to do this here or else accounts will be left behind when stopping, then starting scanning.

ironfish-cli/src/commands/wallet/rescan.ts
NullSoldier: I'll create the `ironfish wallet:reset` command in this PR.

ironfish/src/rpc/routes/wallet/rescan.ts
NullSoldier: Done

ironfish/src/wallet/scanner/walletScanner.ts
NullSoldier: Fixed another bug related to this

ironfish/src/wallet/scanner/walletScanner.ts
NullSoldier: I'll do this immediately in the next PR. We'll most likely move all this logic into ScanState and rename it to WalletScan.

ironfish/src/wallet/scanner/walletScanner.ts
NullSoldier: Made this change

dguenther (approved):

dguenther:
This needs doc changes at least queued up before it goes in I think

PR #4985, opened 2024-05-24
Add wallet/resetAccount RPC

dguenther:
## Summary

Adds an RPC for resetting an account. This can be used to trigger a rescan on a single account, or while scanning is disabled to quickly reset the account to the state it would have after importing the account.

Fixes IFL-2569

## Testing Plan

Added tests for the RPC, as well as the new `resetScanningEnabled` option.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/wallet.ts
```diff
@@ -1606,13 +1606,15 @@ export class Wallet {
...
       resetCreatedAt?: boolean
+      resetScanningEnabled?: boolean
       tx?: IDatabaseTransaction
     },
   ): Promise<void> {
     const newAccount = new Account({
       accountValue: {
         ...account,
         createdAt: options?.resetCreatedAt ? null : account.createdAt,
+        scanningEnabled: options?.resetScanningEnabled ? true : account.scanningEnabled,
```
NullSoldier, line 1617: Should this change the flag? It's probably fine just pointing out this implication.

NullSoldier (approved):

PR #4988, opened 2024-05-28
Prompt for expiration sequence when creating raw or unsigned transactions in CLI

dguenther:
## Summary

Users will typically want to specify an expiration sequence when creating a raw or unsigned transaction, as otherwise by default they'll have 15 blocks to submit the transaction, which may not be long enough to sign it and submit it.

The check is optional, so if users leave the prompt blank, it'll use the default.

## Testing Plan

Manually tested: 
image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/burn.ts
+++ b/ironfish-cli/src/commands/wallet/burn.ts
@@ -16,6 +16,7 @@ import { IronFlag, RemoteFlags, ValueFlag } from '../../flags'
 import { confirmOperation } from '../../utils'
 import { selectAsset } from '../../utils/asset'
 import { promptCurrency } from '../../utils/currency'
+import { promptExpiration } from '../../utils/expiration'
 import { getExplorer } from '../../utils/explorer'
 import { selectFee } from '../../utils/fees'
 import { watchTransaction } from '../../utils/transaction'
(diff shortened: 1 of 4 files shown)


ironfish-cli/src/utils/expiration.ts
```diff
@@ -0,0 +1,27 @@
...
+    const number = parseInt(input, 10)
+
+    if (Number.isNaN(number) || number < 0) {
+      options.logger.error(
+        'Error: Expiration sequence must be a number greater than or equal to 0.',
+      )
+      continue
+    }
+
+    return number
```
hughy, line 25: Should we query the chain head to make sure that the number entered is ahead of the chain head?

hughy (commented):
Should we query the chain head

ironfish-cli/src/utils/expiration.ts
dguenther: Added a check for this

hughy (approved):

ironfish-cli/src/commands/wallet/burn.ts
```diff
@@ -171,6 +172,16 @@ export class Burn extends IronfishCommand {
...
 
+    let expiration = flags.expiration
+    if (flags.rawTransaction && expiration === undefined) {
+      expiration = await promptExpiration({ logger: this.logger, client: client })
+    }
+
+    if (expiration !== undefined && expiration < 0) {
+      this.log('Expiration sequence must be non-negative')
+      this.exit(1)
+    }
```
NullSoldier, line 183: We usually put this into promptExpiration in a while loop

NullSoldier (approved):

PR #4994, opened 2024-05-29
adds account flag to wallet commands

hughy:
## Summary

some wallet commands take 'account' as a positional argument when it may make more sense as a flag.

adds account flag to commands that list things for an account: assets, balances, notes, transactions

adds account flag where commands are oriented around a transaction rather than an account: wallet:transaction, wallet:transaction:watch

parse flags and args so that account flag is preferred, but account arg still accepted

updates arg description to display that argument is deprecated

## Testing Plan

manual testing

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/assets.ts
+++ b/ironfish-cli/src/commands/wallet/assets.ts
@@ -9,7 +9,7 @@ import {
   PUBLIC_ADDRESS_LENGTH,
 } from '@ironfish/rust-nodejs'
 import { BufferUtils } from '@ironfish/sdk'
-import { CliUx } from '@oclif/core'
+import { CliUx, Flags } from '@oclif/core'
 import { IronfishCommand } from '../../command'
 import { RemoteFlags } from '../../flags'
 import { renderAssetWithVerificationStatus } from '../../utils'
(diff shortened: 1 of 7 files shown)


patnir (approved):

ironfish-cli/src/commands/wallet/assets.ts
```diff
@@ -27,19 +27,24 @@ export class AssetsCommand extends IronfishCommand {
...
-      description: 'Name of the account',
+      description: 'Name of the account. DEPRECATED: use --account flag',
     },
   ]
 
   async start(): Promise<void> {
     const { flags, args } = await this.parse(AssetsCommand)
-    const account = args.account as string | undefined
+    // TODO: remove account arg
+    const account = (flags.account ? flags.account : args.account) as string | undefined
```
NullSoldier, line 47: For this to be right, the cast needs to be on args.account rather than the whole expression. It's the same total result, but before it gets cast it's actually `string | any` -> string | undefined.

```suggestion
    const account = (flags.account ? flags.account : args.account as string | undefined)
```

NullSoldier (approved):

PR #5005, opened 2024-06-03
supports combining notes for custom assets

hughy:
## Summary

adds an assetId flag to the 'wallet:notes:combine' command

command prompts user to select asset if no asset is specified

displays output notes (including any change note) from the resulting transaction

does not truncate output in table displaying combined notes so that asset information isn't truncated

## Testing Plan

manual testing:
image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/utils/note.ts
+++ b/ironfish-cli/src/utils/note.ts
@@ -1,10 +1,14 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { Asset } from '@ironfish/rust-nodejs'
 import { Assert, RpcClient } from '@ironfish/sdk'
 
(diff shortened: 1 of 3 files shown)


ironfish-cli/src/utils/note.ts
```diff
@@ -1,18 +1,22 @@
...
-export async function fetchNotes(client: RpcClient, account: string, notesToCombine: number) {
+export async function fetchNotes(
+  client: RpcClient,
+  account: string,
+  assetId: string,
+  notesToCombine: number,
+) {
   const noteSize = await getNoteTreeSize(client)
 
   // TODO(mat): We need to add asset support here for bridges, etc.
```
patnir, line 14: we can remove this right?

patnir (approved):

patnir:
Tested it out by combining a custom asset and it worked great

ironfish-cli/src/utils/note.ts
hughy: good catch, removed!

PR #5006, opened 2024-06-04
Feat: abort scanning asap if needed during reset account

hairtail:
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


dguenther:
We talked through this in Telegram -- this isn't the intended use of `resetScanningEnabled`, rather it's supposed to set scanningEnabled back to its default value (true). But I'll talk with @NullSoldier about the idea of changing it to `setScanningEnabled`

PR #5007, opened 2024-06-04
Rahul/ifl 2638 chainport add utils

patnir:
## Summary

Adds a utils file to test whether a transaction is a chainport transaction or not. 

## Testing Plan

Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/utils/chainport/utils.ts
+++ b/ironfish-cli/src/utils/chainport/utils.ts
@@ -1,0 +1,86 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { RpcWalletTransaction, TransactionType } from '@ironfish/sdk'
+import { getConfig } from './config'
+import { ChainportMemoMetadata } from './metadata'
+
+type ChainportTransactionData =
+  | {
+      type: TransactionType.SEND | TransactionType.RECEIVE
+      chainportNetworkId: number
+      address: string
+    }
+  | undefined
+
+export const extractChainportDataFromTransaction = (
+  networkId: number,
+  transaction: RpcWalletTransaction,
+): ChainportTransactionData => {
+  const config = getConfig(networkId)
+
+  if (transaction.type === TransactionType.SEND) {
+    return getOutgoingChainportTransactionData(transaction, config)
+  }
+  if (transaction.type === TransactionType.RECEIVE) {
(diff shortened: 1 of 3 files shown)


ironfish-cli/src/utils/chainport/utils.ts
```diff
@@ -0,0 +1,111 @@
...
+import { ChainportMemoMetadata } from './metadata'
+
+type ChainportTransactionDetails = {
+  chainportNetworkId: number
+  address: string
+}
+
+type NotChainportTransaction = {
+  isChainportTransaction: false
+}
```
hughy, line 16: nit -- this looks a little funny to wrap `false` this way

Do you think it would be simpler to have a function that returns null or undefined if it's not a Chainport transaction and returns an object with details otherwise?

ironfish-cli/src/utils/chainport/utils.ts
```diff
@@ -0,0 +1,111 @@
...
+    details: {
+      chainportNetworkId: sourceNetwork,
+      address: address,
+    },
+  }
+}
+
+const isAddressInList = (address: string, addressList: string[]): boolean => {
+  return addressList.some((addr) => addr.toLowerCase() === address.toLowerCase())
+}
```
hughy, line 111: are the `outgoingAddresses` and `incomingAddresses` configs used as lists elsewhere? it might be better to use sets

ironfish-cli/src/utils/chainport/utils.ts
```diff
@@ -0,0 +1,111 @@
...
+  )
+
+  if (!bridgeNote) {
+    return {
+      isChainportTransaction: false,
+    }
+  }
+
+  const [sourceNetwork, address, _] = ChainportMemoMetadata.decode(
+    Buffer.from(bridgeNote.memoHex).toString(),
```
hughy, line 96: why do we need to convert string -> Buffer -> string here?

ironfish-cli/src/utils/chainport/utils.ts
patnir: Sounds good, I changed it to be undefined if not a chainport transaction

ironfish-cli/src/utils/chainport/utils.ts
patnir: Ah good idea, I changed it to a set.

ironfish-cli/src/utils/chainport/utils.ts
patnir: fixed

hughy (approved):

PR #5015, opened 2024-06-06
Adds wallet:chainport:send CLI command

patnir:
Also adds the web3-validator dependency to the ironfish-cli package.json to validate ethereum addresses

## Testing Plan

Send a bridge transaction using different assets to separate ethereum addresses that stake (eg: 0x5DF170F118753CaE92aaC2868A2C25Ccb6528f9a)/ don't stake(0x7A68B1Cf1F16Ef89A566F5606C01BA49F4Eb420A) PortX tokens. 

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/chainport/send.ts
+++ b/ironfish-cli/src/commands/wallet/chainport/send.ts
@@ -76,6 +76,12 @@ export class BridgeCommand extends IronfishCommand {
     const client = await this.sdk.connectRpc()
     const { flags } = await this.parse(BridgeCommand)
 
+    const networkId = (await client.chain.getNetworkInfo()).content.networkId
+
+    if (networkId !== TESTNET.id) {
+      this.error(`Chainport transactions are only available on testnet.`)
+    }
+
     if (!flags.offline) {
       const status = await client.wallet.getNodeStatus()
 
@@ -84,12 +90,6 @@ export class BridgeCommand extends IronfishCommand {
           `Your node must be synced with the Iron Fish network to send a transaction. Please try again later`,
         )
       }
-    }
-
-    const networkId = (await client.chain.getNetworkInfo()).content.networkId
-
-    if (networkId !== TESTNET.id) {
-      this.error(`Chainport transactions are only available on testnet.`)
     }
 
     const { targetNetwork, from, to, amount, asset, assetData } =
--- a/ironfish-cli/src/commands/wallet/chainport/send.ts
+++ b/ironfish-cli/src/commands/wallet/chainport/send.ts
@@ -1,0 +1,421 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { Asset } from '@ironfish/rust-nodejs'
+import {
+  CreateTransactionRequest,
+  CurrencyUtils,
+  RawTransaction,
+  RawTransactionSerde,
+  RpcAsset,
+  RpcClient,
+  TESTNET,
+  Transaction,
+} from '@ironfish/sdk'
+import { CliUx, Flags } from '@oclif/core'
+import inquirer from 'inquirer'
+import * as validator from 'web3-validator'
+import { IronfishCommand } from '../../../command'
+import { HexFlag, IronFlag, RemoteFlags, ValueFlag } from '../../../flags'
+import { confirmOperation, selectAsset } from '../../../utils'
+import {
+  ChainportBridgeTransaction,
+  ChainportNetwork,
+  ChainportVerifiedToken,
+  fetchChainportBridgeTransaction,
+  fetchChainportNetworkMap,
+  fetchChainportVerifiedTokens,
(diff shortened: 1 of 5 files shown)


ironfish-cli/src/commands/wallet/chainport/send.ts
```diff
@@ -0,0 +1,421 @@
...
+      })
+    }
+
+    if (!validator.isAddress(to)) {
+      this.error('Invalid to ethereum address')
+    }
+
+    if (flags.expiration !== undefined && flags.expiration < 0) {
+      this.error('Expiration sequence must be non-negative')
+    }
```
hughy, line 179: We added `promptExpiration` last week, could be useful here

ironfish-cli/src/commands/wallet/chainport/send.ts
```diff
@@ -0,0 +1,421 @@
...
+      const names = tokens.map(
+        (t, index) => `${index + 1}. ${t.name} (${t.symbol}) - ${t.web3_address}`,
+      )
+
+      this.error(
+        `Asset ${assetId} not supported by Chainport. Here are the supported tokens: \n\n${names.join(
+          '\n',
+        )}\n`,
+      )
+    }
```
hughy, line 215: Do we need this check now that we pass the filter to `selectAsset`?

ironfish-cli/src/commands/wallet/chainport/send.ts
```diff
@@ -0,0 +1,421 @@
...
+        account: from,
+        id: assetId,
+      })
+    ).content
+
+    if (assetData.verification.status === 'unverified') {
+      assetData.verification.decimals = asset.decimals
+      assetData.verification.symbol = asset.symbol
+      assetData.verification.status = 'verified'
+    }
```
hughy, line 230: Is this because tokens could be in the Chainport verified token list, but not in our verified token list? Is it possible that decimals or symbol could be different between the two lists?

ironfish-cli/src/commands/wallet/transaction/watch.ts
```diff
@@ -7,7 +7,7 @@ import { RemoteFlags } from '../../../flags'
 import { watchTransaction } from '../../../utils/transaction'
 
 export class WatchTxCommand extends IronfishCommand {
-  static description = `Display an account transaction`
+  static description = `Wait for the status of an account transaction to confirm or expire`
```
hughy, line 10: Nice, this description has been hiding out for a while 😅

ironfish-cli/src/commands/wallet/chainport/send.ts
patnir: Yes, this is still useful in case the user uses `--assetId` and passes an id for an asset not supported by chainport

ironfish-cli/src/commands/wallet/chainport/send.ts
patnir: Yes, that is why I added it. 

I guess it is possible. Chainport could send us an incorrect decimal place but I'm assuming that the chance of that is very low, but we would default to the decimals in the verified token list rather than the one provided by chainport. 

Not sure if there is a better approach to handle this case, lmk what you think.

ironfish-cli/src/commands/wallet/chainport/send.ts
```diff
@@ -0,0 +1,421 @@
...
+
+    if (!to) {
+      to = await CliUx.ux.prompt('Enter the public address of the recipient', {
+        required: true,
+      })
+    }
+
+    if (!validator.isAddress(to)) {
+      this.error('Invalid to ethereum address')
+    }
```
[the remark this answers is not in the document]
patnir: I don't see it as super critical to do at the moment, but I'll add it if you think otherwise.

ironfish-cli/src/commands/wallet/chainport/send.ts
patnir: I think to use this prompt, we would have to add support for the `--rawTransaction` flag. Should we add this now?

ironfish-cli/src/commands/wallet/chainport/send.ts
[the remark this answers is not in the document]
hughy: Sounds good, we can follow up on this later if needed.

ironfish-cli/src/commands/wallet/chainport/send.ts
hughy: Yeah I think that it makes more sense to do it when there's support for `rawTransaction` since this will just use the default expiration delta. I don't think we need to add support for `rawTransaction` before this goes in

ironfish-cli/src/commands/wallet/chainport/send.ts
hughy: I'm not sure which decimals we should use if there's a mismatch 🤔 

I think it makes sense to treat our own list as the source of truth, but it wouldn't be good for us to display values one way and then have Chainport interpret the values differently on the other side

We should try to resolve this, but we can do that separately

hughy (approved):

PR #5024, opened 2024-06-07
oclif 1.23.1 -> 3.27.0

mat-if:
## Summary

Upgrades `@oclif/core` from 1.23.1 to 3.27.0. The latest version is 4.0.3, however the transition to 4.0 is going to be very substantial, and will basically require us to build our own CLI UX library so I am going to do this piece first.

There is a good amount of cleanup that can be done now also, but I wanted to keep this PR relatively small, and can followup with more cleanup PRs.

The biggest win here is the typed args. We can treat them like Flags now, which means we can make them more usable, more re-usable, and with better type safety.

## Testing Plan

Manual 💀 

## Documentation

N/A

## Breaking Change

N/A

### DIFF
--- a/ironfish-cli/src/commands/peers/show.ts
+++ b/ironfish-cli/src/commands/peers/show.ts
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { GetPeerMessagesResponse, GetPeerResponse, TimeUtils } from '@ironfish/sdk'
+import { Args } from '@oclif/core'
 import colors from 'colors/safe'
 import { IronfishCommand } from '../../command'
 import { RemoteFlags } from '../../flags'
@@ -12,13 +13,12 @@ type GetPeerMessagesResponseMessages = GetPeerMessagesResponse['messages'][0]
 export class ShowCommand extends IronfishCommand {
   static description = `Display info about a peer`
 
-  static args = [
-    {
-      name: 'identity',
+  static args = {
+    identity: Args.string({
       required: true,
       description: 'Identity of the peer',
-    },
-  ]
+    }),
+  }
 
   static flags = {
     ...RemoteFlags,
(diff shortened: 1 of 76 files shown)


ironfish-cli/src/commands/peers/show.ts
```diff
@@ -27,7 +27,8 @@ export class ShowCommand extends IronfishCommand {
   async start(): Promise<void> {
     const { args } = await this.parse(ShowCommand)
 
-    const identity = (args.identity as string).trim()
+    // TODO(mat): Move trim into parse function like the others
+    const identity = args.identity.trim()
```
hughy, line 31: Do we need to trim args at all? Maybe it's shell dependent, but I imagine oclif needs to separate on whitespace to parse args earlier on

ironfish-cli/src/commands/wallet/rename.ts
```diff
@@ -1,33 +1,33 @@
...
-      name: 'account',
+  static args = {
+    account: Args.string({
       required: true,
       description: 'Name of the account to rename',
-    },
-    {
-      name: 'new-name',
+    }),
+    // TODO(mat): Do we need to do 'new-name', or does oclif do some magic like flags now?
```
hughy, line 16: It looks like args accept a 'name' field, but it doesn't seem to do anything 🤔 

FWIW the name of the arg only affects the help text and the name of the variable in the code, so I think we could rename this to `newName` if we wanted to without meaningfully impacting users

hughy (approved):
Looks good to me!

I haven't tested _every_ command, but everything that I have tested has worked!

ironfish-cli/src/commands/peers/show.ts
mat-if: We could _probably_ get away without it, but it is possible to do something like `peers:show "     0x89a73b38a99c9    "` (note the enclosing quotes). Admittedly kind of a silly edge-case, but an edge-case nonetheless.

What I'm thinking of doing in a follow-up PR is introducing a `TrimmedStringArg` to avoid all the duplicated boiler plate and make it something we don't have to think about

ironfish-cli/src/commands/wallet/rename.ts
mat-if: My guess is the name field is a derelict from the old version of args that they kept around for.. backwards compatibility? That is kind of strange since it's now decided by the key name. 

It did completely slip my mind that these are _args_ and thus the name is sort of arbitrary - thanks for pointing that out. 😂

PR #5027, opened 2024-06-07
Remove RemoteChainProcessor calculating noteIndex

NullSoldier:
## Summary

This was only used in a few places, and it's better for people down the callstack to calculate this since they are also given the header as well.

This is easier to review with https://github.com/iron-fish/ironfish/pull/5027/files?diff=split&w=1

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/setAccountHead.ts
+++ b/ironfish/src/rpc/routes/wallet/setAccountHead.ts
@@ -138,7 +138,8 @@ routes.register<typeof SetAccountHeadRequestSchema, SetAccountHeadResponse>(
 
         const header: BlockHeader | null = await context.chain.getHeader(accountHead.hash)
         Assert.isNotNull(header, 'Account head must be in chain')
-        const transactions = await context.chain.getBlockTransactions(header)
+        const blockTransactions = await context.chain.getBlockTransactions(header)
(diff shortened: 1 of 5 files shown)


ironfish/src/rpc/routes/wallet/setAccountHead.ts
```diff
@@ -138,7 +138,8 @@ routes.register<typeof SetAccountHeadRequestSchema, SetAccountHeadResponse>(
 
         const header: BlockHeader | null = await context.chain.getHeader(accountHead.hash)
         Assert.isNotNull(header, 'Account head must be in chain')
-        const transactions = await context.chain.getBlockTransactions(header)
+        const blockTransactions = await context.chain.getBlockTransactions(header)
+        const transactions = blockTransactions.map((txn) => txn.transaction)
```
hughy, line 142: nit -- we talked about this on slack, but we probably don't need to use `getBlockTransactions` here if we only need the transactions from the block

hughy (approved):

ironfish/src/rpc/routes/wallet/setAccountHead.ts
NullSoldier: I'll open up a new PR to remove usage of this right after.

PR #5030, opened 2024-06-10
update parsing of chainport status API to new interface

patnir:
## Summary

Chainport just sent us some new information about how to use the transaction status API. These changes reflect the new interaction. 

There is also a difference in types. Chainport either sends us an empty object or a body with all keys that either have a type of string or null. 

Examples below: 

Pending transaction: 

image

Confirmed transaction: 

image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/transaction/index.ts
+++ b/ironfish-cli/src/commands/wallet/transaction/index.ts
@@ -56,16 +56,18 @@ export class TransactionCommand extends IronfishCommand {
       hash,
     })
 
-    if (!response.content.transaction) {
+    const transaction = response.content.transaction
+
+    if (!transaction) {
       this.log(`No transaction found by hash ${hash}`)
       return
     }
 
     // by default the notes and spends should be returned
-    Assert.isNotUndefined(response.content.transaction.notes)
-    Assert.isNotUndefined(response.content.transaction.spends)
+    Assert.isNotUndefined(transaction.notes)
+    Assert.isNotUndefined(transaction.spends)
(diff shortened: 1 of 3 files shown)


ironfish-cli/src/commands/wallet/transaction/index.ts
```diff
@@ -57,41 +57,40 @@ export class TransactionCommand extends IronfishCommand {
       hash,
     })
 
-    if (!response.content.transaction) {
+    const transaction = response.content.transaction
+
```
patnir, line 61: this change is just to replace `response.content.transaction` everywhere

ironfish-cli/src/utils/chainport/utils.ts
```diff
@@ -113,38 +114,78 @@ export const displayChainportTransactionSummary = async (
...
+  }
+
+  // If the source bridge transaction has reached the minimum confirmation threshold,
+  // but the target bridge transaction hasn't been sent yet -> base_tx_hash is set and
+  // base_tx_status is 1, but target_tx_hash and target_tx_status are null
+  if (
+    transactionStatus.target_tx_hash === null &&
+    transactionStatus.base_tx_status === 1 &&
+    transactionStatus.base_tx_hash
+  ) {
```
hughy, line 165: Is there a status that we should display if only one or two of these is true?

hughy (approved):

hughy (approved):

PR #5032, opened 2024-06-10
Removes the ability to import/export AccountImport

NullSoldier:
## Summary

For backwards compatability, we used to allow users to import/export AccountImport types instead of strings. Now we only allow importing / exporting encoded accounts.

## Documentation

```
[x] Yes
```

https://github.com/iron-fish/website/pull/709

## Breaking Change

```
[x] Yes
```

Added to the release draft

### DIFF
--- a/ironfish/src/rpc/routes/wallet/importAccount.ts
+++ b/ironfish/src/rpc/routes/wallet/importAccount.ts
@@ -10,14 +10,12 @@ import { RPC_ERROR_CODES, RpcValidationError } from '../../adapters'
 import { ApiNamespace } from '../namespaces'
 import { routes } from '../router'
 import { AssertHasRpcContext } from '../rpcContext'
-import { deserializeRpcAccountImport } from './serializers'
-import { RpcAccountImport } from './types'
 import { tryDecodeAccountWithMultisigSecrets } from './utils'
 
 export class ImportError extends Error {}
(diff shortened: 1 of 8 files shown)


ironfish/src/rpc/routes/wallet/importAccount.ts
```diff
@@ -31,7 +29,7 @@ export const ImportAccountRequestSchema: yup.ObjectSchema<ImportAccountRequest>
   .object({
     rescan: yup.boolean().optional().default(true),
     name: yup.string().optional(),
-    account: yup.mixed<RpcAccountImport | string>().defined(),
+    account: yup.string().defined(),
   })
```
jowparks, line 33: This is backwards incompatible, but I am guessing no one uses it. We should probably make sure the CLI isn’t using it anywhere

jowparks (approved):
Can we delete the RpcAccountImport type?

ironfish/src/rpc/routes/wallet/importAccount.ts
NullSoldier: We cannot use it because it's used in the account export flow. It actually represents the basic type of JSON encoding in the JSON encoder. I'm renaming the type in another PR and moving it into the exporter system to represent that.

PR #5034, opened 2024-06-11
Move usage of RpcAccountImport to exporter system

NullSoldier:
## Summary

This old type was only used by the RPC system, but now is moving into the JSON encoder to show that it's only used by the JSON encoding system to export an AccountImport into JSON.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/utils/json.ts
```diff
@@ -30,7 +30,7 @@ function parse<T = unknown>(data: string, fileName?: string): T {
 function tryParse<T = unknown>(
   data: string,
   fileName?: string,
-): [T, null] | [null, ParseJsonError] {
+): [result: T, error: null] | [result: null, error: ParseJsonError] {
```
NullSoldier, line 33: I made these named to make it easier to understand in intellisense.

ironfish/src/utils/yup.ts
```diff
@@ -83,7 +79,7 @@ export class YupUtils {
     value: unknown,
     options?: yup.ValidateOptions<unknown>,
   ):
-    | { result: YupSchemaResultSync<S>; error: null }
+    | { result: YupSchemaResult<S>; error: null }
```
NullSoldier, line 82: `YupSchemaResult` is already synchronous but oddly `YupSchemaResultSync` was not.

ironfish/src/utils/yup.ts
```diff
@@ -94,7 +90,7 @@ export class YupUtils {
     }
 
     try {
-      const result = schema.validateSync(value, options) as YupSchemaResultSync<S>
+      const result = schema.validateSync(value, options) as YupSchemaResult<S>
```
NullSoldier, line 93: YupSchemaResult is already synchronous but oddly YupSchemaResultSync was not.

ironfish/src/wallet/exporter/accountImport.ts
```diff
@@ -13,7 +21,10 @@ export type AccountImport = {
   incomingViewKey: string
   outgoingViewKey: string
   publicAddress: string
-  createdAt: HeadValue | null
+  createdAt: {
+    hash: Buffer
+    sequence: number
+  } | null
```
NullSoldier, line 27: No DB types in exporting code!

ironfish/src/wallet/exporter/accountImport.ts
```diff
@@ -44,3 +55,54 @@ export function toAccountImport(account: Account, viewOnly: boolean): AccountImp
 
   return value
 }
+
+export function validateAccountImport(toImport: AccountImport): void {
```
NullSoldier, line 59: This was the old `validators/validateAccount` function. However, It's only used to validate an AccountImport and so I moved it in here. There's no need to ever validate an account because the import is verified before an account is constructed. So you only want to verify an AccountImport.

ironfish/src/wallet/exporter/encoders/json.ts
```diff
@@ -1,38 +1,148 @@
...
+  multisigKeys?: {
+    identity?: string
+    secret?: string
+    keyPackage?: string
+    publicKeyPackage: string
+  }
+  proofAuthorizingKey: string | null
+}
+
+const AccountEncodedJSONSchema: yup.ObjectSchema<AccountEncodedJSON> = yup
```
dguenther, line 71: This schema doesn't quite match the type AccountEncodedJSON, I'm guessing that's where some of the test failures are from

dguenther (approved):
Once tests pass this looks good to me!

ironfish/src/wallet/exporter/encoders/json.ts
NullSoldier: Interestingly enough, I didn't actually change this. This is the old schema from RpcAccountImport. That means this is now catching failures that always existed now that we're actually checking the schema on import. That's a good thing. I'll investigate and make the schema match!

PR #5037, opened 2024-06-12
Replaces isAddress depdenency with our own implementation

patnir:
Reduces our dependency on web3-validator which is a large package and is not used in many places. Only dependency used now is keccak which is used in the checksum validation function.

## Testing Plan

Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1586,18 +1586,6 @@
   resolved "https://registry.yarnpkg.com/@napi-rs/cli/-/cli-2.16.1.tgz#912e1169be6ff8bb5e1e22bb702adcc5e73e232b"
   integrity sha512-L0Gr5iEQIDEbvWdDr1HUaBOxBSHL1VZhWSk1oryawoT8qJIY+KGfLFelU+Qma64ivCPbxYpkfPoKYVG3rcoGIA==
 
-"@noble/curves@1.3.0", "@noble/curves@~1.3.0":
-  version "1.3.0"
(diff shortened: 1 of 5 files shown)


yarn.lock
```diff
@@ -1586,18 +1586,6 @@
   resolved "https://registry.yarnpkg.com/@napi-rs/cli/-/cli-2.16.1.tgz#912e1169be6ff8bb5e1e22bb702adcc5e73e232b"
   integrity sha512-L0Gr5iEQIDEbvWdDr1HUaBOxBSHL1VZhWSk1oryawoT8qJIY+KGfLFelU+Qma64ivCPbxYpkfPoKYVG3rcoGIA==
 
-"@noble/curves@1.3.0", "@noble/curves@~1.3.0":
```
mat-if, line 1589: (We already discussed on slack but putting here so it's clear this PR is being looked at)

This PR seems to only remove about 1/3 of the dependencies that the `web3-validator` library added. It's not clear to me if these are now being used by keccak but I would certainly hope not. Might be easier to have this PR also revert that original commit that added the dependency or something, not sure, requires a bit more investigation. If these are indeed all being used by keccak, we should probably just surface some bindings to the rust keccak library instead of using a JS one, since we already introduced that transitively for FishHash

mat-if (changes requested):

yarn.lock
mat-if: Per conversation in slack - these extra deps are from keccak, so we'll just merge this for now as it is an improvement, and then surface rust bindings to keccak in a following PR later

mat-if (approved):

PR #5038, opened 2024-06-12
Separate out account decoding from encrypting

NullSoldier:
## Summary

This fixes the fact that we attempted to add encryption to our encoder
itself, but it was nerver fully inserted into the encoding because the
decoding flow never used the wallet to try all the secrets. This means
the encryption of an encoded account was only half in the encoding
system, and half out.

This introduces a very clear separate module at
Wallet/Exporter/Encryption which has all the methods to handle
encryption.

## Testing Plan
- Added new tests
- Revised old tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/multisig/createTrustedDealerKeyPackage.ts
+++ b/ironfish/src/rpc/routes/wallet/multisig/createTrustedDealerKeyPackage.ts
@@ -5,7 +5,9 @@ import { multisig } from '@ironfish/rust-nodejs'
 import * as yup from 'yup'
 import { Assert } from '../../../../assert'
 import { FullNode } from '../../../../node'
-import { ACCOUNT_SCHEMA_VERSION, Base64JsonEncoder } from '../../../../wallet'
+import { ACCOUNT_SCHEMA_VERSION, JsonEncoder } from '../../../../wallet'
+import { AccountImport } from '../../../../wallet/exporter'
+import { encryptEncodedAccount } from '../../../../wallet/exporter/encryption'
 import { ApiNamespace } from '../../namespaces'
 import { routes } from '../../router'
 
@@ -66,8 +68,8 @@ routes.register<
   (request, node): void => {
     Assert.isInstanceOf(node, FullNode)
 
-    const { minSigners, participants } = request.data
-    const identities = participants.map((p) => p.identity)
+    const identities = request.data.participants.map((p) => p.identity)
+
     const {
       publicAddress,
       publicKeyPackage,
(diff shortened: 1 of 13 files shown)


ironfish/src/testUtilities/matchers/string.ts
```diff
@@ -0,0 +1,30 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { makeResult } from './utils'
+
+// const isBase64 = (s: string): boolean => {
+//   const s === Buffer.from(s, 'base64').toString('base64')
+//   return s === rebuilt
+// }
```
hughy: nit -- leftover code ✂️

ironfish/src/testUtilities/matchers/string.ts
```diff
@@ -0,0 +1,30 @@
...
+
```
dguenther, line 11: Could delete this commented code

```suggestion
```

ironfish/src/rpc/routes/wallet/multisig/createTrustedDealerKeyPackage.ts
```diff
@@ -76,42 +78,46 @@ routes.register<
...
+        multisigKeys: {
+          identity,
+          keyPackage,
+          publicKeyPackage,
         },
-      ),
-    }))
+      }
+
+      const encoder = new JsonEncoder()
```
dguenther, line 106: This was `Base64JsonEncoder` originally, double-checking that you intended to change it to `JsonEncoder`

ironfish/src/wallet/exporter/encryption.ts
```diff
@@ -0,0 +1,81 @@
...
+  method: AccountEncryptionMethod,
+): method is MultisigIdentityEncryption {
+  return 'kind' in method && method.kind === 'MultisigIdentity'
+}
+
+export const BASE64_JSON_MULTISIG_ENCRYPTED_ACCOUNT_PREFIX = 'ifmsaccount'
+
+/**
+ * This returns the decrypted account if decryption was successful
+ * or returns the original unnecrypted input if encryption was not
```
dguenther, line 25:
```suggestion
 * or returns the original unencrypted input if encryption was not
```

ironfish/src/testUtilities/matchers/string.ts
NullSoldier: Removed

hughy (approved):
tested this out by creating and importing a couple of encrypted multisig accounts 👍

ironfish/src/rpc/routes/wallet/multisig/createTrustedDealerKeyPackage.ts
NullSoldier: yeah, it's a little tricky but changing this to JsonEncoder prevents it from being double base64 encoded.

The encryption base64 encodes and adds the prefix to the encrypted result. It means here we use the JsonEncoder to just avoid double base64 which is unnecessary.

ironfish/src/testUtilities/matchers/string.ts
NullSoldier: Fixed

PR #5039, opened 2024-06-12
Automatically start the worker pool if a job is submitted

andiflabs:
## Summary

The worker pool is a thread pool that allows multiple jobs to run concurrently. The worker pool is normally started only when a full node is initialized. If a full node is not initialized, then no separate thread is started, and jobs are not run concurrently but sequentially.

This means that the performance of CLI commands that use the worker pool, like `wallet:rescan`, can vary drastically depending on whether a full node is running or not. This behavior is not documented and far from obvious from a user prospective.

To always offer the best performance, this commit changes the worker pool to ensure that threads are always spawned before executing any job.

## Testing Plan

with a node NOT running:
- Run `wallet:rescan` before this change. Observe that the CPU utilization does not exceed 100%
- Run `wallet:rescan` after this change. Observe that the CPU utilization goes well beyond 100%

Alternatively: observe the time that `wallet:rescan` takes when running with/without a node.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/testUtilities/nodeTest.ts
+++ b/ironfish/src/testUtilities/nodeTest.ts
@@ -85,6 +85,7 @@ export class NodeTest {
     sdk.config.setOverride('enableTelemetry', false)
     sdk.config.setOverride('enableAssetVerification', false)
     sdk.config.setOverride('confirmations', 0)
+    sdk.config.setOverride('nodeWorkers', 0)
 
     // Allow tests to override default settings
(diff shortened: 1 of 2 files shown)


mat-if (approved):
This is a good improvement - it doesn't appear to be negatively affecting CI run times or anything either, so that's good 👍

PR #5041, opened 2024-06-13
Improve the calculation for the number of workers in the worker pool

andiflabs:
## Summary

This commit changes the default value for `nodeWorkersMax` to the number of CPU cores, minus one, capped at 16. Before this change, `nodeWorkersMax` was hardcoded to 6.

As a consequence of this change, most users with empty configuration will observe a larger number of workers spawned. Users who run on systems with SMT enabled and wi.

For example: on a system with 2 cores, with SMT enabled (4 logical CPUs in total), and empty configuration: before this change, the number of workers would have been .

The rationale behind using the number of CPU cores rather than the number of logical CPUs is that the worker pool should mostly be used for CPU-bound and CPU-intensiv.

This commit also changes the default value for `nodeWorkers` to be based on the number of processing units available to the process, rather than the number of logical.

Some examples to understand the impact of this change:

- users with `nodeWorkers` set in their configuration: no change
- users with `nodeWorkers`/`nodeWorkersMax` not set, 2 CPU cores, SMT enabled (4 CPU threads):
  number of workers before this change = min(4 - 1, 6) = 3
  number of workers after this change = min(4 - 1, 2 - 1, 16) = 1
- users with `nodeWorkers`/`nodeWorkersMax` not set, 8 CPU cores, SMT enabled (16 CPU threads):
  number of workers before this change = min(16 - 1, 6) = 6
  number of workers after this change = min(16 - 1, 8 - 1, 16) = 7
- users with `nodeWorkers`/`nodeWorkersMax` not set, 8 CPU cores, SMT disabled (8 CPU threads):
  number of workers before this change = min(16 - 1, 6) = 6
  number of workers after this change = min(8 - 1, 8 - 1, 16) = 7
- users with `nodeWorkers`/`nodeWorkersMax` not set, 32 CPU cores, SMT enabled (64 CPU threads):
  number of workers before this change = min(64 - 1, 6) = 6
  number of workers after this change = min(64 - 1, 32 - 1, 16) = 16
- users with `nodeWorkers`/`nodeWorkersMax` not set, 32 CPU cores, SMT enabled (64 CPU threads), scheduler affinity set to 4 CPUs:
  number of workers before this change = min(64 - 1, 6) = 6
  number of workers after this change = min(4 - 1, 32 - 1, 16) = 3

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/fileStores/config.ts
+++ b/ironfish/src/fileStores/config.ts
@@ -90,13 +90,20 @@ export type ConfigOptions = {
   /**
    * The number of CPU workers to use for long-running node operations, like creating
    * transactions and verifying blocks. 0 disables workers (this is likely to cause
-   * performance issues), and -1 auto-detects based on the number of CPU cores.
+   * performance issues), and -1 auto-detects an optimal value based on the number of
+   * CPU cores (without exceeding `nodeWorkersMax`).
+   *
    * Each worker uses several hundred MB of memory, so try a lower value to reduce memory
    * consumption.
    */
   nodeWorkers: number
   /**
-   * The max number of node workers. See config "nodeWorkers"
+   * The max number of node workers. See also `nodeWorkers". If -1, it defaults to the
+   * number of CPU cores available to the system (capped at 16).
+   *
+   * Note that on system with Simultaneous Multithreading (SMT, also known as
+   * Hyper-Threading on Intel processors), the number of CPU cores is usually half of the
+   * number of virtual CPUs.
    */
   nodeWorkersMax: number
   peerPort: number
@@ -448,7 +455,7 @@ export class Config<
       blockGraffiti: '',
       nodeName: '',
       nodeWorkers: -1,
-      nodeWorkersMax: 6,
+      nodeWorkersMax: -1,
       peerPort: DEFAULT_WEBSOCKET_PORT,
       rpcTcpHost: 'localhost',
       rpcTcpPort: 8020,
--- a/ironfish/src/workerPool/pool.ts
+++ b/ironfish/src/workerPool/pool.ts
@@ -2,9 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-import { UnsignedTransaction } from '@ironfish/rust-nodejs'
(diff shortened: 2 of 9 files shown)


ironfish/src/workerPool/pool.ts
```diff
@@ -316,15 +316,15 @@ export class WorkerPool {
...
+  // user's system responsiveness
 
-    const maxWorkers = nodeWorkersMax
-    if (maxWorkers !== -1) {
-      workers = Math.min(workers, maxWorkers)
-    }
-  }
+  // TODO: ideally we should use `os.availableParallelism()` here, instead of
+  // `os.cpus().length`, but it seems like Typescript doesn't have the type
+  // bindings for it
```
dguenther, line 325: I thought this too, but I think it was only added in 18.14.0: https://nodejs.org/api/os.html#osavailableparallelism and we technically support all versions of 18. You could check if it exists at runtime and call it, else fall back to os.cpus().length

andiflabs: Somehow I was convinced we jumped to Node 20, but no, you're right: we're still on Node 18

ironfish/src/fileStores/config.ts
```diff
@@ -90,13 +90,24 @@ export type ConfigOptions = {
   /**
    * The number of CPU workers to use for long-running node operations, like creating
    * transactions and verifying blocks. 0 disables workers (this is likely to cause
-   * performance issues), and -1 auto-detects based on the number of CPU cores.
+   * performance issues), and a negative value resolves to the amount of parallelism units
+   * available to the process (as returned by `os.availableParallelism()`) minus one.
```
NullSoldier, line 94: While correct, node runners are often dev-ops / normal technical people, while this documentation is engineering focused. I think this is fine but I try not to refer to node documentation or something dev-ops people may struggle with.

andiflabs: I didn't realize this was user-visible documentation; I'll correct that

NullSoldier: We usually copy these docs into the website, but it's manual and we don't need too. I think this is fine if we don't copy them.

NullSoldier (approved):

PR #5042, opened 2024-06-13
adds networkId to wallet, accountImport

hughy:
## Summary

adds networkId to wallet and wallet constructor. node wallets and standalone wallets can only run on a single network without reset. the networkId is stored in the 'internal' file, so it is not necessary to persist networkId in wallet metadata

adds networkId field to 'AccountImport' type to indicate which networkId an account was running on at export time. the network that an account was running on is only important when importing the account to another node

uses networkId to determine if imported account birthday matches network instead of looking up block from chain. this will avoid resetting the account birthday if imported to a node that hasn't synced to the birthday yet

## Testing Plan

updates existing unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/exporter/encoders/json.ts
```diff
@@ -66,6 +66,7 @@ type AccountEncodedJSON = {
     publicKeyPackage: string
   }
   proofAuthorizingKey?: string | null
+  networkId?: number
```
NullSoldier, line 69: I think accounts are pretty translatable across networks, but I was kind of imagining that the networkId is more a function of the createdAt information. I think this is fine, but you might wonder why an exported account has a networkId. Does it not work across networks? Maybe they don't so this is fine.

In my head I was imaginging
```ts
type AccountImport = {
    createdAt?: {
        networkId?: number
        sequence?: number
    }
}
```

hughy: I think it makes sense to do it this way -- I agree that the networkId is only relevant to the information in createdAt.

I'll update to nest it in that field.

ironfish/src/wallet/wallet.ts
```diff
@@ -1355,21 +1359,9 @@ export class Wallet {
...
-    }
-
-    if (createdAt !== null && !(await this.chainHasBlock(createdAt.hash))) {
-      this.logger.debug(
-        `Account ${accountValue.name} createdAt block ${createdAt.hash.toString('hex')} (${
-          createdAt.sequence
-        }) not found in the chain. Setting createdAt to null.`,
+    if (createdAt?.networkId !== this.networkId) {
+      this.logger.warn(
+        `Account ${accountValue.name} networkId ${createdAt?.networkId} does not match wallet networkId ${this.networkId}. Setting createdAt to null.`,
```
NullSoldier, line 1364: I think we only want this message when there is a mismatch but it'll show this when there is no birthday. I don't think we need a warning for importing accounts with no birthdays, which will be all older accounts.

hughy: updated to warn only when networkId is defined

NullSoldier (approved):

PR #5049, opened 2024-06-17
Change `DecryptNotesResponse` to use a sparse serialization format

andiflabs:
## Summary

In the vast majority of the cases, a `DecryptNotesResponse` will contain zero or close-to-zero decrypted notes. It makes sense to have the serialization of `DecryptNotesResponse` optimize for those cases.

This changes the serialization size of `DecryptNotesResponse` from `O(null_notes + non_null_notes)` to `O(non_null_notes)`.

This introduces a small size overhead in the case where the are a lot of decrypted notes. We could in theory have `DecryptNotesResponse` choose between sparse and dense serialization based on how many decrypted notes there are, but that case is so rare, and the overhead so small, that it's not worth optimizing for it.

## Testing Plan

Unit tests

## Documentation

N/A

## Breaking Change

N/A

### DIFF
--- a/ironfish/src/workerPool/tasks/decryptNotes.ts
+++ b/ironfish/src/workerPool/tasks/decryptNotes.ts
@@ -120,42 +120,60 @@ export class DecryptNotesResponse extends WorkerMessage {
   }
 
   serializePayload(bw: bufio.StaticWriter | bufio.BufferWriter): void {
-    // TODO: the majority of responses will have 0 decrypted notes. It make sense to
-    // return a more compact serialization in that case.
-    for (const note of this.notes) {
-      const hasDecryptedNote = Number(!!note)
-      bw.writeU8(hasDecryptedNote)
+    // The majority of responses will have 0 decrypted notes. A small
+    // percentage of responses will have a few decrypted notes. It's
+    // practically rare that a response will contain a large number of
+    // decrypted notes. For this reason, it makes sense to optimize for the
+    // case where the decrypted notes are 0 or close to 0.
+    //
+    // Here we use a sparse serialization: we write the length of the array at
+    // the beginning, and then we serialize only the non-null notes, prefixing
+    // them with their position in the array.
+    //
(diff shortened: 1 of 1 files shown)


ironfish/src/workerPool/tasks/decryptNotes.ts
```diff
@@ -120,42 +120,60 @@ export class DecryptNotesResponse extends WorkerMessage {
...
       }
     }
   }
 
   static deserializePayload(jobId: number, buffer: Buffer): DecryptNotesResponse {
     const reader = bufio.read(buffer)
-    const notes = []
+
+    const arrayLength = reader.readU32()
+    const notes = Array(arrayLength).fill(null) as Array<DecryptedNote | null>
```
hughy, line 173: do we need to return the `null` notes at all?

maybe i'm missing something but it looks like we always filter them out

andiflabs: Yes: upper-level code needs to know which nodes exactly were successfully decrypted. Here's an example: suppose you want to decrypt all notes in this block:

```
block:
  transaction A: [note 1, note 2, note 3, note 4]
  transaction B: [note 5, note 6]
  transaction C: [note 7, note 8]
```

You can submit a decrypt request with all the notes:

```
[note 1, note 2, note 3, note 4, note 5, note 6, note 7, note 8]
```

Once you get a response, you need to know which notes were successfully decrypted, so that you can match them with the corresponding transactions.

If this response returned only non-null values, it wouldn't be possible to decrypt an entire block at once (or even multiple blocks).

andiflabs: An alternative approach that I considered is to change the input from just `note` to `(block, transaction, note)`, and repeat the block/transaction information in the response. This however would greatly increase the size of requests, and that's why I didn't go that route.

hughy (approved):

PR #5051, opened 2024-06-17
WalletScanner: do not update the chain processor head unless accounts have changed

andiflabs:
## Summary

Instead of updating the chain processor head at every loop, do it only if the accounts in the wallet have changed (new accounts have been added, or existing accounts have been removed).

This change is a prerequisite to implement decryption parallelism.

## Testing Plan

Unit tests

## Documentation

N/A

## Breaking Change

N/A

### DIFF
--- a/ironfish/src/wallet/scanner/walletScanner.ts
+++ b/ironfish/src/wallet/scanner/walletScanner.ts
@@ -24,6 +24,13 @@ export class WalletScanner {
   state: ScanState | null = null
   lock: Mutex = new Mutex()
 
+  /**
+   * A snapshot of the account IDs that have `scanningEnabled` set to true.
+   * Used to tell whether accounts were added/removed from the wallet during
+   * scanning (in which case the scanning head is readjusted).
+   */
+  private scanningAccountIds = new Set<string>()
+
   constructor(options: {
     logger: Logger
     nodeClient: RpcClient | null
@@ -53,6 +60,27 @@ export class WalletScanner {
     }
   }
 
+  /**
+   * Replace `scanningAccountIds` with fresh values from the wallet. Returns a
+   * boolean indicating whether `scanningAccountIds` has changed (meaning that
+   * some accounts were added/removed during the scan).
+   */
+  private refreshScanningAccounts(): boolean {
+    const accountIdsFromWallet = this.wallet
+      .listAccounts()
+      .filter((account) => account.scanningEnabled)
+      .map((account) => account.id)
+
+    if (accountIdsFromWallet.length === this.scanningAccountIds.size) {
+      if (accountIdsFromWallet.every((accountId) => this.scanningAccountIds.has(accountId))) {
+        return false
+      }
+    }
+
+    this.scanningAccountIds = new Set(accountIdsFromWallet)
+    return true
+  }
+
   async scan(): Promise<ScanState> {
     const unlock = await this.lock.lock()
 
@@ -61,6 +89,8 @@ export class WalletScanner {
     }
 
     try {
+      this.refreshScanningAccounts()
+
       const start = await this.wallet.getEarliestHead()
       const end = await this.wallet.getChainHead()
 
@@ -96,8 +126,10 @@ export class WalletScanner {
       void (async () => {
         let hashChanged = true
         while (hashChanged) {
-          const head = await this.wallet.getEarliestHead()
-          chainProcessor.hash = head?.hash ?? null
+          if (this.refreshScanningAccounts()) {
+            const head = await this.wallet.getEarliestHead()
+            chainProcessor.hash = head?.hash ?? null
+          }
 
           const result = await chainProcessor.update({ signal: scan.abortController.signal })
           hashChanged = result.hashChanged


ironfish/src/wallet/scanner/walletScanner.ts
```diff
@@ -96,8 +126,10 @@ export class WalletScanner {
       void (async () => {
         let hashChanged = true
         while (hashChanged) {
-          const head = await this.wallet.getEarliestHead()
-          chainProcessor.hash = head?.hash ?? null
+          if (this.refreshScanningAccounts()) {
```
dguenther, line 129: I think it'd be better to change `getEarliestHead` to return block + 1 instead, and maybe move it into a scanner-specific function if we want the logic to be more complex. `refreshScanningAccounts` still iterates over every account to check scanningEnabled like `getEarliestHead`, so I don't think there's a performance improvement here. And `refreshScanningAccounts` adds a cache that's more susceptible to missing relevant account changes (e.g. if resetAccount didn't create a new account, this would cause a bug, or if an account paused scanning, updated account head, and resumed scanning in between calls to `refreshScanningAccounts`, this would not pick up the change)

dguenther (changes requested):

ironfish/src/wallet/scanner/walletScanner.ts
dguenther: Ah, it avoids fetching the account head from the DB each iteration, but I think if that was the goal, I'd prefer to try caching the head itself

andiflabs: (Sorry for the long response; there's some context missing that I should have put in the PR description.)

> I think it'd be better to change `getEarliestHead` to return block + 1 instead, and maybe move it into a scanner-specific function if we want the logic to be more complex

This requires the wallet scanner to have a way to peek into the next block, or changes to the chain processor to skip one block. Is it ok with you if I implement that in a separate PR?

> so I don't think there's a performance improvement here
> [...]
> Ah, it avoids fetching the account head from the DB each iteration, but I think if that was the goal, I'd prefer to try caching the head itself

To be clear: this PR is not a performance improvement. It's a prerequisite for a performance improvement. The improvement I'm talking about is making block processing and note decryption + account updates happen concurrently. The theory is that block processing can be a lot faster than note decryption, so a lot more blocks can be read in a single thread than notes can be decrypted, and therefore note decryption can happen over multiple parallel threads (as opposed to now, where note decryption makes almost no use of parallel threads).

So, in short, the upcoming change is: read as much blocks as you can, decrypt their notes in parallel, write changes to the database.

This means that, at any given point of time, the situation might look like this:

```
ChainProcessor: at block 200
Wallet head: at block 100
Background decryption threads: handling blocks 101..199
```

In this situation, you can see that the wallet head (1) is a moving target, and (2) is always behind the chain processor and the decryption threads. **The wallet head cannot be trusted**.

The code that I'm trying to write is this:

```
if (accountsHaveChanged) {
  getWalletHead
  updateChainProcessorHead
}
```

and after the parallel decryption is implemented is going to become like this:

```
if (accountsHaveChanged) {
  waitForBackgroundDecryptionToBeDone
  getWalletHead
  updateChainProcessorHead
}
```

Only after a call to `waitForBackgroundDecryptionToBeDone`, the wallet head can be trusted. By doing that at every iteration would severely limit performance, and this is why fencing it behind `if (accountsHaveChanged)` is necessary.

So the question for me is: **how do we detect when accounts have changed?** I thought the problem was just about detecting new/removed accounts, but you raise interesting points: there are other ways that accounts can be changed and affect scanning.

My initial idea was to add some code to the `Wallet` class such that, every time an account is added/removed/modified in a way that affects scanning, the `WalletScanner` is notified. That would be the best approach IMO if it wasn't for the fact that there's the possibility that, in the future, people will write code that should trigger those notifications, and if they forget about it, they would be introducing hard-to-debug bugs. However, I'm now reconsidering going with this "notification" approach, as it can catch all corner cases without too much complexity.

dguenther: Thanks for the explanation of the planned changes! I think I might be conflating two separate issues here:

1. Chain iteration methods can fetch the same block multiple times. aka setting chainProcessor.hash to `getEarliestHead` fetching the same block twice
1. We want to allow the chain processor to run ahead of the account heads, so we want the chain processor head to be `next block that needs fetching` rather than the earliest account head.

For 1, I only really care about it if it seems like it's impacting performance, and if you're planning to move away from using `getEarliestHead` anyway, I don't think it's important to address now.

For 2, I think it'd be helpful to have this change attached to the change that stops regarding the account heads as the source of the chain processor state (or introduces `waitForBackgroundDecryptionToBeDone`). By itself, this change doesn't feel like it provides a clear benefit to scanning logic as it stands right now, and function's logic will probably end up changing depending on the implementation of background block fetching + decryption.

Aside from that, I wonder if rather than considering the problem as determining whether `accountsHaveChanged`, maybe `shouldQueueBlocksForFetching` would be easier to solve instead. But it might depend on how multiple-block-fetching is implemented.

PR #5058, opened 2024-06-19
Maximize decryption parallelism during wallet scans

andiflabs:
## Summary

A wallet scan currently works like this:

```mermaid
graph LR;
  r[read block]-->d[decrypt notes]
  d-->u[update accounts]
  u-.->r
```

*read block* and *update accounts* are primarily IO-bound operations, and they run in the main thread only. This means that while *read block* and *update accounts* are running, the CPUs are under-utilized.

To maximize CPU utilization (and therefore improve scan performance), the scan has been changed to look like this:

```mermaid
graph LR;
  r[read block]-.->r
  r-->q[queue]
  q-->d[decrypt notes]
  d-->u[update accounts]
```

With this new algorithm, *read block* will try to read as many blocks as possible, without stopping, and place them on a queue. The queue is consumed and notes are decrypted in parallel inside the worker threads. Because *read block* is significantly faster than *decrypt notes*, the queue will normally be almost full, and so the worker threads are fully utilized.

## Testing Plan

* Unit tests
* `wallet:rescan`

## Documentation

N/A

## Breaking Change

N/A

### DIFF
--- a/ironfish/src/wallet/scanner/noteDecryptor.ts
+++ b/ironfish/src/wallet/scanner/noteDecryptor.ts
@@ -1,0 +1,221 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Assert } from '../../assert'
+import { Config } from '../../fileStores'
+import { BlockHeader } from '../../primitives'
+import { Transaction } from '../../primitives/transaction'
+import { AsyncQueue } from '../../utils/asyncQueue'
+import { WorkerPool } from '../../workerPool'
+import { Job } from '../../workerPool/job'
+import {
+  DecryptedNote,
+  DecryptNotesOptions,
+  DecryptNotesRequest,
+  DecryptNotesResponse,
+} from '../../workerPool/tasks/decryptNotes'
+import { JobAbortedError } from '../../workerPool/tasks/jobAbort'
+import { Account } from '../account/account'
+
+export type DecryptNotesFromTransactionsCallback = (
+  account: Account,
+  blockHeader: BlockHeader,
+  transactions: Array<{ transaction: Transaction; decryptedNotes: Array<DecryptedNote> }>,
+) => Promise<void>
+
+export class BackgroundNoteDecryptor {
+  private isStarted = false
+
+  private triggerFlushed: (() => void) | null = null
+  private triggerStopped: (() => void) | null = null
+
+  private onFlushed: Promise<void> = Promise.resolve()
+  private onStopped: Promise<void> = Promise.resolve()
+
+  private readonly workerPool: WorkerPool
+  private readonly options: DecryptNotesOptions
+  private readonly decryptQueue: AsyncQueue<{
+    job: Job
+    accounts: ReadonlyArray<Account>
+    blockHeader: BlockHeader
+    transactions: ReadonlyArray<Transaction>
+    callback: DecryptNotesFromTransactionsCallback
+  }>
+
+  constructor(workerPool: WorkerPool, config: Config, options: DecryptNotesOptions) {
+    this.workerPool = workerPool
+    this.options = options
+
+    let queueSize = 8 * workerPool.numWorkers
+    const maxQueueSize = config.get('walletSyncingMaxConcurrency')
+    if (maxQueueSize > 0) {
+      queueSize = Math.min(queueSize, maxQueueSize)
+    }
+    queueSize = Math.max(queueSize, 1)
+    this.decryptQueue = new AsyncQueue(queueSize)
+  }
+
+  start(abort?: AbortController) {
+    if (!this.isStarted) {
+      this.isStarted = true
+      this.onStopped = new Promise((resolve) => (this.triggerStopped = resolve))
+      void this.decryptLoop()
+
+      if (abort) {
+        abort.signal.addEventListener('abort', this.stop.bind(this))
+      }
+    }
+  }
+
+  stop() {
+    if (this.isStarted) {
+      this.isStarted = false
+      for (const { job } of this.decryptQueue) {
+        job.abort()
+      }
+      this.decryptQueue.clear()
+      if (this.triggerStopped) {
+        this.triggerStopped()
+      }
+    }
+  }
+
+  private async decryptLoop(): Promise<void> {
+    while (this.isStarted) {
+      if (this.decryptQueue.isEmpty() && this.triggerFlushed) {
+        this.triggerFlushed()
+        this.triggerFlushed = null
+      }
+
+      const item = await Promise.race([this.decryptQueue.pop(), this.onStopped])
+      if (!item) {
+        break
+      }
+
+      const { job, accounts, blockHeader, transactions, callback } = item
+
+      let decryptNotesResponse
+      try {
+        decryptNotesResponse = await job.result()
+      } catch (e) {
+        if (e instanceof JobAbortedError) {
+          break
+        }
+        throw e
+      }
+
+      if (!this.isStarted) {
+        break
+      }
+
+      Assert.isInstanceOf(decryptNotesResponse, DecryptNotesResponse)
+      const decryptedNotes = decryptNotesResponse.mapToAccounts(
+        accounts.map((account) => ({ accountId: account.id })),
+      )
+
+      for (const { account, decryptedTransactions } of regroupNotes(
+        accounts,
+        transactions,
+        decryptedNotes,
+      )) {
+        if (!this.isStarted) {
+          break
+        }
+        await callback(account, blockHeader, decryptedTransactions)
+      }
+    }
+  }
+
+  /**
+   * Waits for all the in flight decrypt requests to be fully processed.
+   */
+  async flush(): Promise<void> {
+    if (!this.isStarted) {
+      return
+    }
+    await this.onFlushed
+  }
+
+  decryptNotesFromBlock(
+    blockHeader: BlockHeader,
+    transactions: ReadonlyArray<Transaction>,
+    accounts: ReadonlyArray<Account>,
+    callback: DecryptNotesFromTransactionsCallback,
+  ): Promise<void> {
+    if (!this.isStarted) {
+      throw new Error('decryptor was not started')
+    }
+
+    if (!this.triggerFlushed) {
+      this.onFlushed = new Promise((resolve) => (this.triggerFlushed = resolve))
+    }
+
+    const accountKeys = accounts.map((account) => ({
+      incomingViewKey: account.incomingViewKey,
+      outgoingViewKey: account.outgoingViewKey,
+      viewKey: account.viewKey,
+    }))
+    Assert.isNotNull(blockHeader.noteSize)
+
+    const encryptedNotes = []
+    let currentNoteIndex = transactions
+      .map((transaction) => transaction.notes.length)
+      .reduce((accumulator, numNotes) => accumulator - numNotes, blockHeader.noteSize)
+
+    for (const transaction of transactions) {
+      for (const note of transaction.notes) {
+        encryptedNotes.push({ serializedNote: note.serialize(), currentNoteIndex })
+        currentNoteIndex++
+      }
+    }
+
+    const decryptNotesRequest = new DecryptNotesRequest(
+      accountKeys,
+      encryptedNotes,
+      this.options,
+    )
+    const job = this.workerPool.execute(decryptNotesRequest)
+
+    return this.decryptQueue.push({
+      job,
+      accounts,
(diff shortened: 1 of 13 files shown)


ironfish/src/wallet/scanner/walletScanner.ts
```diff
@@ -137,71 +167,43 @@ export class WalletScanner {
...
+          }))
 
-      await this.wallet.connectBlockForAccount(
-        account,
-        blockHeader,
-        txns,
-        shouldDecryptAccountIds.has(account.id),
-      )
-    }
+          const shouldDecrypt = this.wallet.shouldDecryptForAccount(blockHeader, account)
```
hughy, line 198: Do we check `shouldDecryptForAccount` before doing decryption? We compare the `account.createdAt` to the current block to avoid doing any decryption if the block was added to the chain before the account's `createdAt`

ironfish/src/wallet/scanner/queue.ts
```diff
@@ -0,0 +1,124 @@
...
+function assertNotEmpty<T>(x: T | typeof EMPTY): asserts x is T {
+  if (x === EMPTY) {
+    throw new Error('Expected item not to be empty')
+  }
+}
+
+/**
+ * A fixed-size, async queue, implemented on top of a circular buffer.
+ */
+export class Queue<Item> {
```
rohanjadvani, line 22: We have a priority queue in `/src/utils`, wonder if this should be in the same directory

ironfish/src/wallet/scanner/walletScanner.ts
andiflabs: Good point, not sure how I missed this. I wonder if it's even possible for an account to have their head set before their `createdAt`? That would allow simplifying the logic a bit

ironfish/src/wallet/scanner/queue.ts
andiflabs: Sounds good to me, I can move it there

ironfish/src/wallet/scanner/walletScanner.ts
```diff
@@ -252,21 +254,48 @@ export class WalletScanner {
...
+        })),
+    )
   }
 
-  return result.slice().reverse()
+  private getEarliestHead(): HeadValue | null {
+    let earliestHead = null
+    for (const { scanFrom: head } of this.scanningAccounts) {
+      if (!head) {
+        return null
```
NullSoldier, line 293: Note to look into this: but a null head just means the head is the block before the genesis block. Wouldn't this be considered the earliest head? I need to read the rest of the code to see how this is handled now.

ironfish/src/wallet/scanner/walletScanner.ts
```diff
@@ -93,15 +108,27 @@ export class WalletScanner {
...
-          const head = await this.wallet.getEarliestHead()
-          chainProcessor.hash = head?.hash ?? null
+          if (this.haveWalletAccountsChanged()) {
+            // Accounts have changed in the wallet. Wait for all pending
+            // decrypt requests to be completed, then update the head of the
+            // chain processor.
+            await decryptor.flush()
+            await this.refreshScanningAccounts()
+            const head = this.getEarliestHead()
+            chainProcessor.hash = head?.hash ?? null
```
NullSoldier, line 124: Since this is a long running operation should we check the abort controller here in the scan state?

ironfish/src/wallet/scanner/walletScanner.ts
NullSoldier, line 198: I think it's time to move this method into the WalletScanner soon as well.

NullSoldier: The head can be set after the birthday if you set it to the birthday, then immediately rewind because you were on a fork.

andiflabs: I understand the fork case, but I don't think that checking for the sequence (like `shouldDecryptForAccount` is doing) is the right approach. Consider this chain:

```mermaid
graph LR;
  a1["A1 (seq: 100)"]
  a1-->b1["B1 (seq: 101)"]
  b1-->b2["B2 (seq: 102)"]
  b2-->b3["B3 (seq: 103)"]
  b3-->b4["B4 (seq: 104)"]
  a1-->c1["C1 (seq: 101)"]
  c1-->c2["C2 (seq: 102)"]
  c2-->c3["C3 (seq: 103)"]
```

Here we have a chain A that forks and is split into two chains: B and C. Suppose an account is initially created at block B3, therefore at sequence 103. The main chain then becomes C, the blocks are rolled back and a transaction involving the account is created at block C2, sequence 102.

In this case, the `createdAt` has sequence 103, but a transaction exists at sequence 102. `shouldDecryptForAccount` will return `false` for C2, and will cause the scanner to miss the transaction.

Anyway, this problem is not strictly relevant to this PR, but if the intent of `shouldDecryptForAccount` is to handle forks, then we probably need to improve on that.

ironfish/src/wallet/scanner/queue.ts
NullSoldier: My only feedback is that I would rename this to AsyncQueue since Javascript already has a built in queue (Array).

When I saw this I didn't understand why this is re-implementing the Javascript array since the default array in javascript is a queue that works well. I expected this to be an Async wrapper on top of array. I think the circular buffer is a fine optimization, but not necessary because most of the time you aren't dealing with throughput and allocations but waiting on async operations to clear the queue.

ironfish/src/wallet/scanner/noteDecryptor.ts
```diff
@@ -0,0 +1,144 @@
...
+  private readonly decryptQueue: Queue<{
+    promise: Promise<ReadonlyMap<string, ReadonlyArray<DecryptedNote | null>>>
+    transactions: ReadonlyArray<Transaction>
+    callback: DecryptNotesFromTransactionsCallback
+  }>
+
+  constructor(workerPool: WorkerPool, options: DecryptNotesOptions) {
+    this.workerPool = workerPool
+    this.options = options
+    this.decryptQueue = new Queue(Math.max(32 * workerPool.numWorkers, 1))
```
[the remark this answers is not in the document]
NullSoldier: This will be roughly 524 MB for a 32 core machine worst case with full blocks.

ironfish/src/wallet/scanner/queue.ts
andiflabs: Ok to rename this to `AsyncQueue`.

I'd like to point out though that this is a FIFO queue; an array could only be used efficiently for a LIFO queue. I've seen in our code base that we sometimes use arrays as FIFO queues using `.shift()`. For large queues of N elements, this has the side effect that N-1 elements have to be read and re-written for every `pop()` operation. This has a huge impact on the performance of the queue and anything using it.

ironfish/src/wallet/scanner/walletScanner.ts
NullSoldier: The documented above behaviour is expected. It's why the birthday is not actually 103 for B3.. The createdAt sequence should correctly be 103 - confirmation range which is defaulted to 2. The birthday should be calculated at 101. That being said, it's still possible to miss if at the point of creation you are on a re-org that's longer than the confirmation time. However, the idea of the confirmation time is to make the unlikely. This is a change that Hugh is working on.

The other note is that in general, birthdays are heuristics and if they are wrong you are expected to miss transactions.

ironfish/src/wallet/scanner/walletScanner.ts
```diff
@@ -137,71 +167,43 @@ export class WalletScanner {
...
 
-    for (const account of accounts) {
-      if (abort?.signal.aborted) {
-        return
-      }
+    return decryptor.decryptNotesFromBlock(
+      blockHeader,
+      transactions,
+      accounts,
+      async (decryptedNotesMap) => {
```
NullSoldier, line 185: Random thought. This is sort of an old school NodeJS approach to use a callback like this now that we have promises. Only one place is really going to use this API but generally the preferred way would probably have this method return Promise<AccountTransactionNoteMap | null> in the case of cancelation.

That being said, I don't want to recommend a change to this because it works great.

ironfish/src/wallet/scanner/noteDecryptor.ts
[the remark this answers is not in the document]
andiflabs: > This will be roughly 524 MB for a 32 core machine worst case with full blocks.

The maximum number of workers is capped at 16 by default, so this would be (32 + 1) * 16 * 0.5 MiB = 264 MiB worst case.

ironfish/src/wallet/scanner/noteDecryptor.ts
```diff
@@ -0,0 +1,144 @@
...
+      void this.decryptLoop()
+    }
+  }
+
+  stop() {
+    this.isStarted = false
+    if (this.triggerStopped) {
+      this.triggerStopped()
+    }
+  }
```
NullSoldier, line 56: We did talk about this offline but I want to repost my final thoughts her as well. After looking at the code, I think this code works as expected. The primary issue I have is that this code doesn't manage it's promises very well causing the node to hault up when you're doing things like... telling the node to stop, and aborting a scan.

I can see why. If the abort controller for a scan is triggered, it waits for the decryptor to flush and then it stops it. I think this is an issue with the current version but I would definitely want it fixed in a future PR.

Things that have queue's of async work in general should manage those items, and cancel them appropriately when the system is told to stop. Abort's should be propagated through the sub-modules and ideally cancel async operations if they are abled to and fully wait on them to be aborted so no hanging promises are left which will keep a NodeJS process alive after the main loop has exited.

ironfish/src/wallet/scanner/walletScanner.ts
andiflabs: I'm aware of those patterns, however returning a `Promise` (and, supposedly, awaiting it) in this case can't be done, because it would prevent the next block from being read. The callback is there to make sure that its code gets executed in a different unit without affecting the reading of blocks.

NullSoldier:
After the latest change, I'm starting to see these errors emmitted from the logs.

```
Could not parse payload from request: (jobId: 66, type: JobAborted, body: '')
```

Reproduction
- Run `ironfish start`
- Run `ironfish wallet:rescan`

NullSoldier:
When running the CLI command `ironfish wallet:rescan` I got this error on my node

```
/Users/jasonspafford/projects/fish/ironfish/src/assert.ts:20
      throw new Error(message || `Expected value not to be undefined`)
            ^


Error: Expected value not to be undefined
    at Function.isNotUndefined (/Users/jasonspafford/projects/fish/ironfish/src/assert.ts:20:13)
    at WalletDB.getHead (/Users/jasonspafford/projects/fish/ironfish/src/wallet/walletdb/walletdb.ts:399:12)
    at <anonymous> (/Users/jasonspafford/projects/fish/ironfish/src/wallet/scanner/walletScanner.ts:293:21)
    at async Promise.all (index 69)
    at WalletScanner.refreshScanningAccounts (/Users/jasonspafford/projects/fish/ironfish/src/wallet/scanner/walletScanner.ts:287:29)
    at WalletScanner.scan (/Users/jasonspafford/projects/fish/ironfish/src/wallet/scanner/walletScanner.ts:78:7)
    at Wallet.scan (/Users/jasonspafford/projects/fish/ironfish/src/wallet/wallet.ts:190:18)
```

andiflabs:
> After the latest change, I'm starting to see these errors emmitted from the logs.
> 
> ```
> Could not parse payload from request: (jobId: 66, type: JobAborted, body: '')
> ```

It happens rarely for me, but I was able to observe it. It's caused by [`Job.abort()` sending a `JobAbortedMessage`](https://github.com/iron-fish/ironfish/blob/e57b1a0e371cc2faadf68ce773221d038fc5b6b0/ironfish/src/workerPool/job.ts#L58):

```ts
this.worker.send(new JobAbortedMessage(this.id))
```

which is a [message of type `JobAborted`](https://github.com/iron-fish/ironfish/blob/e57b1a0e371cc2faadf68ce773221d038fc5b6b0/ironfish/src/workerPool/tasks/jobAbort.ts#L8):

```ts
super(WorkerMessageType.JobAborted, jobId)
```

which, according to `Worker`, [should never be sent](https://github.com/iron-fish/ironfish/blob/master/ironfish/src/workerPool/worker.ts#L225-L226):

```ts
case WorkerMessageType.JobAborted:
  throw new Error('JobAbort should not be sent as a request')
```

My guess is that this bug has been hidden for a while because perhaps this is the first use of `.abort()` that we have. In any case, I'll work on getting it fixed, but if you have some historical context on the intentions of this code, please share.

andiflabs:
https://github.com/iron-fish/ironfish/pull/5093 should fix the "Could not parse payload from request" problem

NullSoldier (approved):

PR #5060, opened 2024-06-20
removes references to account createdAt hash

hughy:
## Summary

now that we use networkId in account imports to determine whether the createdAt sequence refers to a block on the same chain the hash is no longer needed. it is mostly vestigial, like the appendix or the wings of a kiwi bird

these changes remove remaining references to createdAt.hash

## Testing Plan

existing unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -22,7 +22,7 @@ import { Side } from '../merkletree/merkletree'
 import { Witness } from '../merkletree/witness'
 import { Mutex } from '../mutex'
 import { BlockHeader } from '../primitives'
-import { GENESIS_BLOCK_PREVIOUS, GENESIS_BLOCK_SEQUENCE } from '../primitives/block'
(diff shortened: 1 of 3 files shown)


ironfish/src/wallet/wallet.ts
```diff
@@ -1372,11 +1372,11 @@ export class Wallet {
...
       if (createdAt !== null) {
-        const previousBlockHash = await this.chainGetPreviousBlockHash(createdAt.hash)
+        const previousBlock = await this.chainGetBlock({ sequence: createdAt.sequence - 1 })
 
-        const head = previousBlockHash
+        const head = previousBlock
           ? {
-              hash: previousBlockHash,
+              hash: Buffer.from(previousBlock.block.hash, 'hex'),
               sequence: createdAt.sequence - 1,
```
NullSoldier, line 1380: nit

```suggestion
              sequence: previousBlock.block.sequence,
```

NullSoldier (approved):

PR #5061, opened 2024-06-20
Use `inquirer` for confirm prompt

mat-if:
## Summary

This also creates an abstraction for the confirm prompt as well as a confirm-or-quit function so it is easier to use confirm logic in commands

This is part of oclif 4.x upgrade prep

## Trade-offs

We're trading our manually-added "Do you confirm (Y)es/(N)o" for the library appended "(Y/n)". There is no easy way to override this, short of upgrading the version of inquirer we use. However, the new version of inquirer is an ECMAScript Module, which on its own is solvable, but on top of that, the `@types/inquirer` library is quite outdated, which would require us to write our own types to use it, which is more work than it is worth right now.

image

## Testing Plan

Manual testing

## Documentation

N/A

## Breaking Change

N/A


ironfish-cli/src/ui.ts
```diff
@@ -109,3 +111,28 @@ export class ProgressBar {
...
+    // add a space before the message, which is unwanted.
+    message: `\n${message}`,
+    name: 'prompt',
+    prefix: '',
+  })
+  return result.prompt
+}
+
+export async function confirmOrQuit(message?: string, confirm?: boolean): Promise<void> {
+  if (confirm) {
```
NullSoldier, line 128: nit: confirm should be past tense since if it's true it's already confirmed

NullSoldier (approved):

PR #5067, opened 2024-06-24
wip: simple evm transaction execution

hughy:
## Summary

adds blockchain interface for evm

adds example for generating evm addresses from ironfish private keys and sending a simple transfer between two accounts

refactors evm 'state' module to wrap the evm 'vm' and state manager

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/test-evm.ts
```diff
@@ -15,27 +17,49 @@ export class TestEvmCommand extends IronfishCommand {
...
-    await evmState.manager.commit()
-    await evmState.manager.flush()
+    let senderBalance = (await evm.stateManager.getAccount(senderAddress))?.balance ?? 0n
+    this.log(
+      `Sender account at address ${senderAddress.toString()} has balance ${senderBalance}`,
+    )
 
-    balance = (await evmState.manager.getAccount(address))?.balance ?? 0n
-    this.log(`Account at address ${address.toString()} has balance ${balance}`)
+    const tx = new LegacyTransaction({
```
jowparks, line 42: What is the difference between LegacyTransaction and Transaction? I am guessing you chose the right one, just want to verify.

jowparks (approved):

ironfish-cli/src/commands/test-evm.ts
hughy: There are a few different types of transactions based on EIPs: https://github.com/ethereumjs/ethereumjs-monorepo/blob/a8a89732efdbb2ceb57d80aa46afcfc294c84c0c/packages/tx/src/types.ts#L114

`LegacyTransaction` seemed like the simplest for testing 😅

PR #5068, opened 2024-06-24
Introduce optional pagination to ChainProcessor

NullSoldier:
## Summary

This adds similar optional pagination that the RemoteChainProcessor has to the ChainProcessor. This is useful if you want to stop after a certain amount of blocks to perform other work.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/chainProcessor.ts
+++ b/ironfish/src/chainProcessor.ts
@@ -35,11 +35,18 @@ export class ChainProcessor {
   logger: Logger
   onAdd = new Event<[block: BlockHeader]>()
   onRemove = new Event<[block: BlockHeader]>()
+  maxQueueSize: number | null
 
-  constructor(options: { logger?: Logger; chain: Blockchain; head: Buffer | null }) {
+  constructor(options: {
+    logger?: Logger
+    chain: Blockchain
+    head: Buffer | null
+    maxQueueSize?: number | null
+  }) {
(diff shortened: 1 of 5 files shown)


ironfish/src/chainProcessor.ts
```diff
@@ -92,6 +101,11 @@ export class ChainProcessor {
       await this.remove(remove)
       this.hash = remove.previousBlockHash
       this.sequence = remove.sequence - 1
+
+      blockCount++
+      if (this.maxQueueSize && blockCount >= this.maxQueueSize) {
+        return { hashChanged: !oldHash || !this.hash.equals(oldHash) }
```
andiflabs, line 107: Nit: given that this is repeated in multiple places, we could have an anonymous fn like this:

```ts
const returnValue = () => ({ hashChanged: !oldHash || !this.hash.equals(oldHash) })
```

and then change all the `return { ... }` statements to:

```ts
return returnValue()
```

andiflabs (approved):

ironfish/src/chainProcessor.ts
NullSoldier: I have one more recommendation for this that I put in another PR here, https://github.com/iron-fish/ironfish/pull/5073.

andiflabs (approved):

PR #5073, opened 2024-06-25
Simplify ChainProcessor return logic

NullSoldier:
## Summary

We only ever cared if work was done, not that the specific hash was changed from the original. This does change the behaviour, but I looked at everywhere it's used and it should not affect the overall behaviour.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/chainProcessor.ts
+++ b/ironfish/src/chainProcessor.ts
@@ -58,8 +58,6 @@ export class ChainProcessor {
   }
 
   async update({ signal }: { signal?: AbortSignal } = {}): Promise<{ hashChanged: boolean }> {
-    const oldHash = this.hash
-
     if (!this.hash) {
       await this.add(this.chain.genesis)
       this.hash = this.chain.genesis.hash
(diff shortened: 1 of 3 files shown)


hughy (approved):

ironfish/src/chainProcessor.ts
```diff
@@ -122,13 +122,14 @@ export class ChainProcessor {
       await this.add(add)
       this.hash = add.hash
       this.sequence = add.sequence
-
+      hashChanged = true
```
andiflabs, line 125: If I'm reading `findFork` and `iterateTo` correctly, it is possible that `add.hash` may be the same as the initial `this.hash`. In that case, wouldn't setting `hashChanged = true` potentially get consumers stuck in an infinite loop (because they would keep calling `update()`)?

Maybe I'm wrong, because if what I'm describing is right, then some tests would fail. But by looking at the code I'm not convinced that this change is totally harmless.

andiflabs (approved):

ironfish/src/chainProcessor.ts
NullSoldier: I added a test to verify this, so it should be more verifiable now via a test.

PR #5074, opened 2024-06-25
Add a development-only feature to print statistics related to notes encryption

andiflabs:
## Summary

This commit adds an optional feature (that can be enabled only during local development) to print the number of cryptographic operations performed for note decryption and encryption.

This feature is enabled by building `ironfish-rust-nodejs` through `yarn build:stats`. After doing that, and after starting `ironfish-cli` through `yarn start <command>`, it is possible to trigger the printing of the statistics by sending `SIGUSR1` to the CLI process. The statistics will look like this:

```
Elliptic Curve Point Multiplication Stats:
• affine muls: 0
• extended muls: 752
Note Encryption Stats:
• total: 3
Note Decryption Stats:
• for owner: 107 (17 [15.89%] successful)
• for spender: 3 (3 [100.00%] successful)
```

All the code related to this feature is in Rust only, and hidden behind a feature flag. The JS/TS code is untouched. As such, this commit has no impact on release builds of Iron Fish.

## Testing Plan

* `cd ironfish-rust-nodejs && yarn build:stats`
* `cd ironfish-cli && yarn start start`
* Send SIGUSR1 to the Node process (e.g. `pkill -USR1 -f '^/usr/bin/node --expose-gc .* bin/run .*'`)
* Observe the stats being printed out

## Documentation

N/A

## Breaking Change

N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1517,6 +1517,7 @@ dependencies = [
  "napi-derive",
  "num_cpus",
  "rand",
+ "signal-hook",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


jowparks (approved):
Very cool!

I would love to see this integrated into a GHA somehow so that a warning is triggered on build if you see an increase in this number for some example tests (like sending a transaction, with mint/burn etc).

andiflabs:
> Very cool!
> 
> I would love to see this integrated into a GHA somehow so that a warning is triggered on build if you see an increase in this number for some example TS tests (like sending a transaction, with mint/burn etc).

Yeah, definitely it would be helpful. We could have a variant of this feature that writes stuff to a JSON file, and compare the file contents before/after each test run. Having assertions around the counters would ensure that in the future nobody accidentally introduces performance penalties without realizing it.

PR #5075, opened 2024-06-25
Fast construction of `NoteEncrypted` during wallet scans

andiflabs:
## Summary

Wallet scans get the blocks/transactions/notes from the chain db. The chain db contains data that has already been validated in the past (or that otherwise is assumed to be valid). For this reason, it should be possible to speed up scans by skipping expensive validation performed when parsing notes.

More specifically, `NoteEncrypted` parses some elliptic curve points in compressed form. The regular parsing done through `jubjub::SubgroupPoint::from_bytes()` involves an expensive check: it checks whether the uncompressed point belongs to the prime-order subgroup. This check is perfromed using elliptic curve point multiplication, which is the most predominant and slowest operation happening during wallet scans.

Due to this check, for each note, the wallet currently performs two elliptic curve point multiplications: one to parse the note, another to attempt the actual decryption. By using `from_bytes_unchecked()` instead of `from_bytes()` we can skip the first multiplication, effectively improving note decryption performance by a factor of *almost* 2x.

This change only affects the performance of wallet scans. Note parsing and decryption outside of a scan is unaffected at the moment. In the future, we can disable validation in other contexts as we see fit.

## Testing Plan

Observe the number of elliptic curve point multiplications happening during a scan before and after this change (using the feature introduced in GH-5074).

## Documentation

N/A

## Breaking Change

N/A

### DIFF
--- a/ironfish-rust/src/merkle_note.rs
+++ b/ironfish-rust/src/merkle_note.rs
@@ -2,7 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-use crate::{errors::IronfishError, keys::EphemeralKeyPair, serializing::read_point};
+use crate::{
+    errors::IronfishError,
+    keys::EphemeralKeyPair,
+    serializing::{read_point, read_point_unchecked},
+};
 
 /// Implement a merkle note to store all the values that need to go into a merkle tree.
 /// A tree containing these values can serve as a snapshot of the entire chain.
@@ -39,7 +43,7 @@ pub const NOTE_ENCRYPTION_MINER_KEYS: &[u8; NOTE_ENCRYPTION_KEY_SIZE] =
     b"Iron Fish note encryption miner key000000000000000000000000000000000000000000000";
 const SHARED_KEY_PERSONALIZATION: &[u8; 16] = b"Iron Fish Keyenc";
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct MerkleNote {
     /// Randomized value commitment. Sometimes referred to as
     /// `cv` in the literature. It's calculated by multiplying a value by a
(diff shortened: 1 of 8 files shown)


ironfish-rust/src/merkle_note.rs
```diff
@@ -162,6 +166,30 @@ impl MerkleNote {
         })
     }
 
+    /// Load a MerkleNote from the given reader, skipping some expensive validity checks on the
+    /// input.
+    ///
+    /// This method is faster than [`read()`], but it requires trusted or pre-validated input. It
+    /// might panic if the input is not properly encoded.
```
jowparks, line 173: By panic do you mean return error result?

andiflabs: I recall there was some panic behavior somewhere, but I can't find it anymore. I guess that, when I wrote this line, the code was a bit different. I'm going to reword this comment.

hughy (approved):

ironfish/src/primitives/noteEncrypted.ts
```diff
@@ -22,9 +22,16 @@ export class NoteEncrypted {
...
   private noteEncrypted: NativeNoteEncrypted | null = null
   private referenceCount = 0
-
-  constructor(noteEncryptedSerialized: Buffer) {
+  /**
+   * Whether the note has already been previously validated, and thus does not
+   * need to be checked anymore after parsing. Used to speed up construction of
+   * `NativeNoteEncrypted` in `takeReference`.
+   */
+  private validated: boolean
```
NullSoldier, line 30: nit: I sorta wish this name matched the underlying rust name which is `skipValidation` but I'm not strongly opinionated about this.

NullSoldier:
I've reviewed this but mostly from a learning perspective, not a bug perspective so take my approve with a grain of salt.

I did ultimately track down the difference in checked vs unchecked to here, if I'm correct. https://github.com/iron-fish/jubjub/blob/47dfe5181ccf39166c0c479c35c0644d708f4294/src/lib.rs#L1424-L1433

NullSoldier (approved):

andiflabs:
> I've reviewed this but mostly from a learning perspective, not a bug perspective so take my approve with a grain of salt.
> 
> I did ultimately track down the difference in checked vs unchecked to here, if I'm correct. https://github.com/iron-fish/jubjub/blob/47dfe5181ccf39166c0c479c35c0644d708f4294/src/lib.rs#L1424-L1433

Yes, that's correct. The checked function calls `into_subgroup()`, the unchecked variant doesn't. `into_subgroup()` [calls `is_torsion_free()`](https://github.com/iron-fish/jubjub/blob/47dfe5181ccf39166c0c479c35c0644d708f4294/src/lib.rs#L1348), which in turn [calls `multiply()`](https://github.com/iron-fish/jubjub/blob/47dfe5181ccf39166c0c479c35c0644d708f4294/src/lib.rs#L710), which is the expensive operation that we want to get rid of.

ironfish/src/primitives/noteEncrypted.ts
andiflabs: Changed every name to be in line with `skipValidation`

PR #5080, opened 2024-06-26
adds minimal verification to ironfish transactions for evm

jowparks:
## Summary
Simple `EvmDescription` verification via the EVM. This is a roundtrip transaction that passes through rust, to node typescript, and then to EVM.

Simply verifies the transaction is executable. Only form of validation explicitly done is signature for the unit test.
## Testing Plan
tests pass
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/test-evm.ts
+++ b/ironfish-cli/src/commands/test-evm.ts
@@ -4,7 +4,7 @@
 import { LegacyTransaction } from '@ethereumjs/tx'
 import { Account, Address } from '@ethereumjs/util'
 import { generateKey } from '@ironfish/rust-nodejs'
-import { IronfishEvm } from '@ironfish/sdk'
+import { IronfishEvm } from '@ironfish/sdk/src/evm'
 import { IronfishCommand } from '../command'
 import { LocalFlags } from '../flags'
 
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -10,6 +10,7 @@ import { BlockHasher } from '../blockHasher'
 import { Consensus } from '../consensus'
 import { VerificationResultReason, Verifier } from '../consensus/verifier'
 import { Event } from '../event'
+import { IronfishEvm } from '../evm'
 import { Config } from '../fileStores'
 import { FileSystem } from '../fileSystems'
 import { createRootLogger, Logger } from '../logger'
@@ -61,6 +62,8 @@ export class Blockchain {
   location: string
   files: FileSystem
   consensus: Consensus
+  // TODO: the EVM creation is async, so this has to be optional for the constructor
+  evm?: IronfishEvm
   seedGenesisBlock: SerializedBlock
   config: Config
   blockHasher: BlockHasher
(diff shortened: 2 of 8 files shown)


ironfish/src/primitives/evmDescription.ts
```diff
@@ -11,3 +13,30 @@ export interface EvmDescription {
...
+export function evmDescriptionToLegacyTransaction(desc: EvmDescription): LegacyTransaction {
+  return new LegacyTransaction({
+    nonce: desc.nonce,
+    to: desc.to.length > 0 ? desc.to : undefined,
+    value: desc.value,
+    data: desc.data,
+    v: BigInt(desc.v),
+    r: desc.r.length > 0 ? bytesToBigInt(desc.r) : undefined,
+    s: desc.s.length > 0 ? bytesToBigInt(desc.s) : undefined,
+    // TODO(jwp) gas constants are hardcoded
```
jowparks, line 38: I have this hardcoded, we could potentially just add to the description for now, wdyt?

ironfish/src/wallet/wallet.ts
```diff
@@ -98,7 +99,7 @@ export class Wallet {
   readonly scanner: WalletScanner
   readonly nodeClient: RpcClient | null
   private readonly config: Config
```
jowparks, line 101: Needs to be readonly to mock.

ironfish-cli/src/commands/test-evm.ts
```diff
@@ -4,7 +4,7 @@
 import { LegacyTransaction } from '@ethereumjs/tx'
 import { Account, Address } from '@ethereumjs/util'
 import { generateKey } from '@ironfish/rust-nodejs'
```
jowparks, line 6: This import was failing for me, not sure why

hughy (approved):

ironfish/src/primitives/evmDescription.ts
jowparks: Is there a way to generically disable gas int eh EVM?  If so we should do it.

ironfish/src/consensus/verifier.ts
```diff
@@ -548,6 +562,36 @@ export class Verifier {
...
+      return { valid: true }
+    }
+    if (!this.chain.evm) {
+      return { valid: false, reason: VerificationResultReason.EVM_NOT_INITIALIZED }
+    }
+    // verify signature
+
+    const tx = evmDescriptionToLegacyTransaction(evmDescription)
+    try {
+      const result = await this.chain.evm.runTx({ tx })
```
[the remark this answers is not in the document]
jowparks: Good call, whatever db transaction passthrough mechanism we add we can revert in this method. I will add a TODO in here.

ironfish/src/blockchain/blockchain.ts
```diff
@@ -61,6 +62,8 @@ export class Blockchain {
   location: string
   files: FileSystem
   consensus: Consensus
+  // TODO: the EVM creation is async, so this has to be optional for the constructor
+  evm?: IronfishEvm
```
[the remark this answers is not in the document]
jowparks: Is it possible to make in non-async with this library? I don't see a standard method for it in the lib.

PR #5083, opened 2024-06-27
implements BlockchainDBTransaction to wrap EVM state changes

hughy:
## Summary

the ethereumjs packages implement their own systems for committing changes to the underlying database

plumbing our own database transaction objects down through several layers will require significant refactoring to evm caching and database access

BlockchainDBTransaction wraps an IDatabaseTransaction and inserts logic to use the evm stateManager's checkpoint logic for committing and reverting state changes

implements transaction and withTransaction methods on the BlockchainDB and replaces access of the underlying db instance with these transaction methods to ensure that BlockchainDBTransactions are used

moves stateManager directly onto BlockchainDB to support access in transaction context

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/database/blockchaindb.test.ts
+++ b/ironfish/src/blockchain/database/blockchaindb.test.ts
@@ -1,0 +1,92 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Account, Address, hexToBytes } from '@ethereumjs/util'
+import { createNodeTest } from '../../testUtilities'
+
+describe('BlockchainDBTransaction', () => {
+  const nodeTest = createNodeTest()
+
+  it('should update state if tx succeeds', async () => {
+    const { node } = nodeTest
+    const stateManager = node.chain.blockchainDb.stateManager
+
+    const address = new Address(hexToBytes('0x0e5069fd80d59b92e359dade34f6a66f0bf8dcc5'))
+    const account = new Account(BigInt(0), 100000n)
+
+    await expect(
+      node.chain.blockchainDb.stateManager.getAccount(address),
+    ).resolves.toBeUndefined()
+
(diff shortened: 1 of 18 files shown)


ironfish/src/blockchain/database/blockchaindb.test.ts
```diff
@@ -0,0 +1,92 @@
...
+      await stateManager.commit()
+    })
+
+    const newAccount = await node.chain.blockchainDb.stateManager.getAccount(address)
+
+    expect(newAccount).not.toBeUndefined()
+    expect(newAccount?.balance).toEqual(100000n)
+  })
+
+  it('should not update the stateManager state root if tx succeeds', async () => {
```
jowparks, line 33: I think this test maybe should be `should update root`

ironfish/src/blockchain/database/blockchaindb.ts
```diff
@@ -344,11 +358,91 @@ export class BlockchainDB {
...
+    if (this.checkpoint) {
+      await this.stateManager.commit()
+      this.checkpoint = false
+    }
+    await this.tx.commit()
+  }
+
+  async abort(): Promise<void> {
+    if (this.checkpoint) {
+      await this.stateManager.revert()
```
jowparks, line 439: nice

ironfish/src/blockchain/database/blockchaindb.test.ts
hughy: good catch, updated!

ironfish/src/storage/database/transaction.ts
```diff
@@ -18,6 +22,8 @@
  * or write is performed on it.
  */
 export interface IDatabaseTransaction {
+  cache: BufferMap<unknown>
```
jowparks, line 25: Should this be a generic?

jowparks (approved):

ironfish/src/storage/database/transaction.ts
hughy: updated to be a generic `Map<Buffer, unknown>` interface

PR #5085, opened 2024-06-27
Evm verification checks

jowparks:
## Summary
- Adds "No Mint" check for global account when no EVM description is present

TODO: Add checks for evm shield/unshield
```

    const shieldBalance: BufferMap<bigint> = new BufferMap()
    const mintBalance: BufferMap<bigint> = new BufferMap()
    const unshieldBalance: BufferMap<bigint> = new BufferMap()
    const burnBalance: BufferMap<bigint> = new BufferMap()

    for (const shield of result.shields) {
      shieldBalance.set(shield.assetId, shield.amount)
    }
    for (const unshield of result.unshields) {
      unshieldBalance.set(unshield.assetId, unshield.amount)
    }

    for (const mint of transaction.mints) {
      mintBalance.set(mint.asset.id(), mint.value)
    }

    for (const burn of transaction.burns) {
      burnBalance.set(burn.assetId, burn.value)
    }

    // verify mints equals shields and burns equals unshields
    if (mintBalance.size !== shieldBalance.size) {
      return {
        valid: false,
        reason: VerificationResultReason.EVM_MINT_LENGTH_MISMATCH,
      }
    }
    for (const [assetId, value] of mintBalance) {
      if (shieldBalance.get(assetId) !== value) {
        return {
          valid: false,
          reason: VerificationResultReason.EVM_MINT_BALANCE_MISMATCH,
        }
      }
    }

    if (burnBalance.size !== unshieldBalance.size) {
      return {
        valid: false,
        reason: VerificationResultReason.EVM_BURN_LENGTH_MISMATCH,
      }
    }
    for (const [assetId, value] of burnBalance) {
      if (unshieldBalance.get(assetId) !== value) {
        return {
          valid: false,
          reason: VerificationResultReason.EVM_BURN_BALANCE_MISMATCH,
        }
      }
    }
```

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/evm/evm.ts
```diff
@@ -39,4 +42,34 @@ export class IronfishEvm {
   async runTx(opts: RunTxOpts): Promise<RunTxResult> {
     return this.vm.runTx(opts)
   }
+
+  async verifyTx(opts: RunTxOpts): Promise<EvmResult> {
+    // TODO(jwp) add db transaction and roll back
+    const result = await this.runTx(opts)
+
+    // TODO(jwp) from custom opcodes populate shields and unshields
```
jowparks, line 50: Not sure how we will capture the opcode events during this call, but this is where it will go

ironfish/src/evm/evm.ts
```diff
@@ -39,4 +42,34 @@ export class IronfishEvm {
...
+    return {
+      result,
+      shields: [],
+      unshields: [],
+    }
+  }
+}
+
+type EvmShield = {
```
jowparks, line 59: Should we call it `mint` on the EVM side? I wonder if that will confuse people writing the contract. Maybe not, I just think shield/unshield is more intuitive

ironfish/src/evm/evm.ts
```diff
@@ -13,6 +13,9 @@ import { EvmStateDB } from './database'
 export class IronfishEvm {
   stateManager: DefaultStateManager
   private vm: VM
+  // spending key
+  static spendingKey = '3333fff66a72eff15588d48d9b5404e3a20a4ab90e0dfcf4aa999f1855788da6'
+  static publicAddress = '2edbd0f3296a925ce373598294102c7179902bb5323b8bdcec54feabd10658d4'
```
hughy, line 18: How will we use this key and address? Will the global account be signing with this key?

jowparks: I imagined that the mint descriptions will be generated with this key and the public address will own all the shielded assets. This way anyone can generate the mints. I suspect we will need to sign with this key too since the outputs will be generated from this account.

patnir (approved):

PR #5086, opened 2024-06-27
Add Opcode and handlers

danield9tqh:
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/evm/decodeOpcodes.ts
```diff
@@ -0,0 +1,56 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
```
jowparks, line 4: What is this file for? Doesn't look like it is doing anything relative to the custom opcodes. I also don't see where it is being used, maybe I am missing something though.

jowparks (approved):
Generally the integration looks good, not sure what the decode file is for though.

PR #5087, opened 2024-06-27
Make unsignedTransaction flag visible in send command.

patnir:
Make unsignedTransaction flag visible in send command. Will be required for adding signatures using ledger devices to this transaction. 

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/send.ts
+++ b/ironfish-cli/src/commands/wallet/send.ts
@@ -96,7 +96,6 @@ export class Send extends IronfishCommand {
         'Return raw transaction. Use it to create a transaction but not post to the network',
     }),
     unsignedTransaction: Flags.boolean({
-      hidden: true,
       default: false,
       description:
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/wallet/send.ts
```diff
@@ -255,6 +254,8 @@ export class Send extends IronfishCommand {
       raw = RawTransactionSerde.deserialize(bytes)
     }
 
+    displayTransactionSummary(raw, assetData, amount, from, to, memo)
+
```
patnir, line 258: Makes sens to show the summary before exiting for the rawTransaction/ unsignedTransaction flags

mat-if: Agreed - big improvement. Small nit though, can you add a newline to the end of the transaction summary output to improve the readability between the summary and the raw/unsigned tx output?

mat-if (approved):
1 lil newline guy and I think this is 👍

PR #5089, opened 2024-06-27
DecryptNotes: move the code that groups notes by account from WorkerPool to the response class

andiflabs:
## Summary

The goal of this PR is to move that chunk of code to a location where it's more easily reusable. The plan is to use the new method introduced here in PR-5058: that PR will be changed to not call `WorkerPool.decryptNotes()` anymore, but rather call `WorkerPool.execute()` to obtain the individual jobs.

## Testing Plan

Unit tests

## Documentation

N/A

## Breaking Change

N/A

### DIFF
--- a/ironfish/src/workerPool/tasks/decryptNotes.ts
+++ b/ironfish/src/workerPool/tasks/decryptNotes.ts
@@ -3,6 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { DECRYPTED_NOTE_LENGTH, ENCRYPTED_NOTE_LENGTH } from '@ironfish/rust-nodejs'
 import bufio from 'bufio'
+import { Assert } from '../../assert'
 import { NoteEncrypted } from '../../primitives/noteEncrypted'
 import { ACCOUNT_KEY_LENGTH } from '../../wallet'
 import { VIEW_KEY_LENGTH } from '../../wallet/walletdb/accountValue'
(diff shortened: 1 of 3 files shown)


ironfish/src/workerPool/tasks/decryptNotes.ts
```diff
@@ -225,6 +226,31 @@ export class DecryptNotesResponse extends WorkerMessage {
...
+  /**
+   * Groups each note in the response by the account it belongs to. The
+   * `accounts` passed must be in the same order as the `accountKeys` in the
+   * `DecryptNotesRequest` that generated this response.
+   */
+  mapToAccounts(
+    accounts: ReadonlyArray<{ accountId: string }>,
+  ): Map<string, Array<DecryptedNote | null>> {
+    const decryptedNotesByAccount: Array<
+      [accountId: string, notes: Array<DecryptedNote | null>]
```
jowparks, line 239: `.map` is much slower in some cases than `for..of`. Not sure if this is one of those cases just checking to see if you are aware

jowparks (approved):

ironfish/src/workerPool/tasks/decryptNotes.ts
andiflabs: It's true that a for loop can outperform `.map()`, but in this very specific case I'm trying to get an array out of another array, so I'd expect the performance difference to be minimal. I would be more concerned if I was doing chains of `.map`/`.filter`, because those create a new array each time, and that would be better written as a for loop (at least until Node.js gains the ability to optimize these methods). Thanks for pointing it out though.

PR #5092, opened 2024-06-28
ensures that postTransaction uses a spending account

hughy:
## Summary

replaces 'getAccountByPublicAddress' with a more generic 'findAccount' that takes a predicate

uses findAccount in postTransaction to find the account that has the same public address and is a spending account

fixes issue where wallet has a spending account and a view only account with the same public address and cannot post a transaction

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


patnir (approved):
Looks good!

PR #1155, opened 2022-03-17
feat(ironfish): Add binary serialization to the worker pool

rohanjadvani:
## Summary

Creates a serializable interface and migrates existing worker pool jobs to implement the new behavior.

## Testing Plan

Manually run node and added unit tests.

## Breaking Change

Is this a breaking change? If yes, add notes below on why this is breaking and
what additional work is required, if any.

```
[ ] Yes
[x] No
```

PR #5093, opened 2024-06-28
Fix `Worker` handling of `JobAborted` requests

andiflabs:
## Summary

`Job.abort()` sends a `JobAborted` request to the worker. `Worker.onMessageFromParent()` is able to handle such requests, but `Worker.parseRequest()` rejects them. Fixed `Worker.parseRequest()` so that such requests are no longer rejected.

Also fixed `JobAbortedMessage.deserializePayload()` to return the correct type.

## Documentation

N/A

## Breaking Change

N/A

### DIFF
--- a/ironfish/src/workerPool/tasks/jobAbort.ts
+++ b/ironfish/src/workerPool/tasks/jobAbort.ts
@@ -12,8 +12,8 @@ export class JobAbortedMessage extends WorkerMessage {
     return
   }
 
-  static deserializePayload(): JobAbortedError {
-    return new JobAbortedError()
+  static deserializePayload(): JobAbortedMessage {
+    return new JobAbortedMessage()
   }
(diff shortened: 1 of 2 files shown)


NullSoldier (approved):

NullSoldier:
This was introduced here, https://github.com/iron-fish/ironfish/pull/1155/files#diff-37121b9a047d787288b096b47cdc558dfc452d21e069ef2045b87a49e13394a8L113

It looks like it might have been a copy paste error that prevented this from getting plumbed through and handled here. https://github.com/iron-fish/ironfish/pull/1155/files#diff-37121b9a047d787288b096b47cdc558dfc452d21e069ef2045b87a49e13394a8R147 It worked fine in the JSON messages because we never validated the incoming message types.

PR #5094, opened 2024-06-28
Prevent race conditions when getting list of accounts during scanning

andiflabs:
## Summary

Calling `Wallet.reset()` while a scanning is running can sometimes result in race conditions. In particular, `WalletScanner` may attempt to query account heads at the same time while accounts are being reset, resulting in errors.

This commit fixes the problem by using database transactions (1) while `WalletScanner` read accounts and (2) while `Wallet.reset()` updates accounts.

This commit also adds transactions in `Wallet.getEarliestHead()` and `Wallet.getLatestHead()`. These methods are not used during scanning, but they suffer from the same kind of problem.

Note that this commit fixes specific race conditions related to getting the list of accounts to scan, but other race conditions that affect scanning may still exist.

## Documentation

N/A

## Breaking Change

N/A

### DIFF
--- a/ironfish/src/wallet/scanner/walletScanner.ts
+++ b/ironfish/src/wallet/scanner/walletScanner.ts
@@ -10,7 +10,7 @@ import { Config } from '../../fileStores'
 import { Logger } from '../../logger'
 import { Mutex } from '../../mutex'
 import { BlockHeader, Transaction } from '../../primitives'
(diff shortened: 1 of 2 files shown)


hughy (approved):

NullSoldier (approved):

NullSoldier:
one day we'll remove the need for this TX and just use a lock that you can take when you are modifying or using accounts.

PR #5095, opened 2024-06-28
adds store to map contract addresses to the assets they own

jowparks:
## Summary

Adds store so that we can have contracts that own assets

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


hughy (approved):

hughy:
Will any EVM execution logic need access to this mapping?

jowparks:
> Will any EVM execution logic need access to this mapping?

No it shouldn't. It is just used to determine which contract is the owner of the asset.

PR #5096, opened 2024-06-28
add_signature to unsigned transaction

patnir:
This feature will be used to attach signatures to unsigned transactions. The main usecase is to allow the user to sign the transaction offline and then submit it to the network. The offline signing will happen on ledger devices.

## Testing Plan

Unit tests added

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-rust/src/transaction/tests.rs
```diff
@@ -837,3 +839,31 @@ fn test_aggregate_signature_shares() {
...
+    let bytes = hex_to_vec_bytes(unsigned).expect("invalid hex");
```
jowparks, line 847: expects should be the positive "should be valid hex"

ironfish-rust/src/transaction/tests.rs
```diff
@@ -837,3 +839,31 @@ fn test_aggregate_signature_shares() {
     // verify transaction
     verify_transaction(&signed_transaction).expect("should be able to verify transaction");
 }
+
+#[test]
+fn test_add_signature() {
```
jowparks, line 844: Why are you using hex repr here? What you are really trying to test here is adding signature to a unsigned transaction, irrespective of ledger. You should be constructing the transaction in code in the test.

ironfish-rust-nodejs/src/structs/transaction.rs
```diff
@@ -460,6 +460,19 @@ impl NativeUnsignedTransaction {
 
         Ok(Buffer::from(vec))
     }
+
+    #[napi]
+    pub fn add_signature(&mut self, signatures: String) -> Result<Buffer> {
```
jowparks, line 465: Why are you using a `String` here? I think you should be using a `JsBuffer` to avoid ser/de. Not sure why we use string in some places and not others:
ironfish-rust-nodejs/src/structs/transaction.rs

Unless of course the data is coming from ledger as a string, then it doesn't matter if you do serde at this layer.

jowparks (commented):
Left some comments.

ironfish-rust-nodejs/src/structs/transaction.rs
patnir: Data from the ledger is coming in as a buffer, so I'll change it to JsBuffer

ironfish-rust/src/transaction/unsigned.rs
```diff
@@ -256,6 +256,38 @@ impl UnsignedTransaction {
...
     }
 
+    pub fn add_signature(
+        &mut self,
+        signature_string: String,
+    ) -> Result<Transaction, IronfishError> {
+        let signature_bytes: [u8; 64] = hex_to_bytes(&signature_string)?;
+        let signature = Signature::read(&signature_bytes[..])?;
+
+        let mut spend_descriptions = Vec::with_capacity(self.spends.len());
```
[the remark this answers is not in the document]
hughy: I think the only other use of `drain` right now is in `aggregate_signature_shares`. That's carried on from a former implementation of `ProposedTransaction.post`.

I think the benefit of `drain` in these cases is that it avoids cloning and allocating a new object.

ironfish-rust/src/transaction/unsigned.rs
```diff
@@ -256,6 +256,36 @@ impl UnsignedTransaction {
         Ok(transaction)
     }
 
+    pub fn add_signature(&mut self, signature: [u8; 64]) -> Result<Transaction, IronfishError> {
```
hughy, line 259: We might be able to refactor `aggregate_signature_shares` to call `add_signature`

ironfish-rust/src/transaction/unsigned.rs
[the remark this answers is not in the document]
patnir: Sounds good, I am going to change aggregate_signature_shares to call add_signatures and change the implementation to use `drain`

ironfish-rust/src/transaction/tests.rs
patnir: aggregate_signature_shares tests should cover this now

jowparks (approved):

PR #5102, opened 2024-07-02
Add test for ArrayUtils

NullSoldier:
## Summary

These didn't have tests

## Testing Plan
I added new tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/utils/array.test.ts
+++ b/ironfish/src/utils/array.test.ts
@@ -1,6 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Assert } from '../assert'
 import { ArrayUtils } from './array'
 
 describe('ArrayUtils', () => {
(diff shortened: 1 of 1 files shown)


ironfish/src/utils/array.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    let items: number[] = []
+
+    for (let i = 0; i <= 10000; ++i) {
+      items.push(i)
+    }
+
+    items = ArrayUtils.shuffle(items)
+
+    const shuffled = items[0] !== 0 && items[500] !== 500 && items[1000] !== 1000
+    expect(shuffled).toBe(true)
```
hughy, line 17: this will randomly fail sometimes (maybe ~3/1000 times?)

i think any assertions we make about random outputs might fail randomly since even very unlikely outcomes are possible

ironfish/src/utils/array.test.ts
```diff
@@ -0,0 +1,43 @@
...
+  it('shuffles array', () => {
+    let items: number[] = []
+
+    for (let i = 0; i <= 10000; ++i) {
+      items.push(i)
+    }
+
+    items = ArrayUtils.shuffle(items)
+
+    const shuffled = items[0] !== 0 && items[500] !== 500 && items[1000] !== 1000
```
andiflabs, line 16: Nit: this test would pass if all elements are set to 99. In that case, the array wouldn't be shuffled, but filled with copies of a single element.

Probably a better test would be to check that it contains all the original items, but in a different order, like this (untested):

```ts
const shuffled = ArrayUtils.shuffle(items)
expect(shuffled).not.toEqual(items)
expect(shuffled.sort()).toEqual(items)
```

Still not perfect, but a bit more robust.

andiflabs (approved):

ironfish/src/utils/array.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    expect(shuffled).toBe(true)
+  })
+
+  it('sample a random item', () => {
+    let sample = ArrayUtils.sample([0, 1, 2])
+    expect(sample).toBeGreaterThanOrEqual(0)
+    expect(sample).toBeLessThanOrEqual(2)
```
andiflabs, line 23: Similar problem to above: this test can pass if `ArrayUtils.sample()` returns always the first value. A better test would be to sample many values from the same array, and check that after ~100 or ~1000 samples, all possible samples have been returned.

ironfish/src/utils/array.test.ts
NullSoldier: oh yeah I did make this better now

ironfish/src/utils/array.test.ts
NullSoldier: The test doesn't set all elements to 99, the test relies on every element being different.

However I still think your test is better so I'm gonna use that.

NullSoldier: Fixed

ironfish/src/utils/array.test.ts
NullSoldier: Improved this with the suggestion

andiflabs (approved):

hughy (approved):

ironfish/src/utils/array.test.ts
```diff
@@ -0,0 +1,50 @@
...
+describe('ArrayUtils', () => {
+  it('shuffles array', () => {
+    const items: number[] = []
+    for (let i = 0; i <= 10000; ++i) {
+      items.push(i)
+    }
+
+    const shuffled = ArrayUtils.shuffle(items)
+    expect(shuffled).not.toEqual(items)
+    expect(shuffled.sort()).toEqual(items)
```
hughy, line 16: JS sorting uses string representations by default

```suggestion
    expect(shuffled.sort((a, b) => a - b).toEqual(items)
```

PR #5103, opened 2024-07-02
Add lint rule to not use Promise.race

dguenther:
## Summary

@andiflabs did some great investigation to discover that Promise.race leaks memory, and the TL;DR is it'll be easiest to generally avoid using Promise.race. I added a lint rule to catch uses of this in the future, and added workarounds for our current use.

## Testing Plan

Run `status -f` and watch the memory usage while rescanning to observe it doesn't rapidly increase.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


andiflabs (approved):
LGTM, thanks for fixing it, and great idea to add the lint rule!

PR #5109, opened 2024-07-03
adds IronfishEvm methods to simulate tx execution

hughy:
## Summary

adds 'withCopy', inspired by withTransaction, to use a shallow copy of the underlying vm for execution

using challowCopy creates a vm instance with empty caches over the committed database state. use of shallowCopy follows the example of gas estimation https://github.com/ethereumjs/ethereumjs-monorepo/blob/cc2848a1f3fc0ebb8c04cf6e9f76b0992e506c2c/packages/client/src/rpc/modules/eth.ts#L535

wraps handler execution with checkpoint/revert logic over the evm journal to ensure that cached state is not committed to the database

updates EvmStateDB to return itself in shallowCopy so that it does not create a duplicate datastore

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/test-evm.ts
+++ b/ironfish-cli/src/commands/test-evm.ts
@@ -4,7 +4,7 @@
 import { LegacyTransaction } from '@ethereumjs/tx'
 import { Account, Address } from '@ethereumjs/util'
 import { generateKey } from '@ironfish/rust-nodejs'
-import { IronfishEvm } from '@ironfish/sdk'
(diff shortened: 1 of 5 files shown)


jowparks (approved):
Nice, not sure how performant this is, but should be sufficient for MVP.

hughy:
> Nice, not sure how performant this is, but should be sufficient for MVP.

It shouldn't be much less performant than using `runTx` without a copy of the VM: `shallowCopy` is mostly just instantiating a second object with the same parameters and empty caches.

PR #5111, opened 2024-07-03
Rahul/ifl 2746 wallet sign command

patnir:
## Summary

Adds the `wallet:sign` CLI command along with the `wallet/signTransaction` RPC.

We already have a `sign` function on an UnsignedTransaction on the Rust side. This basically exposes that feature. 

Also included is the ledger signing option for this CLI command. 

## Testing Plan

Manually tested signing an unsigned transaction and adding (wallet:transaction:add) the signed transaction using both an account with a spending key and an account that is a ledger device. 

## Documentation

To follow in a separate PR on the website

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/wallet/signTransaction.ts
```diff
@@ -0,0 +1,54 @@
...
+import { UnsignedTransaction } from '@ironfish/rust-nodejs'
+import * as yup from 'yup'
+import { ApiNamespace } from '../namespaces'
+import { routes } from '../router'
+import { AssertHasRpcContext } from '../rpcContext'
+import { getAccount } from './utils'
+
+export type SignTransactionRequest = {
+  unsignedTransaction: string
+  account?: string
```
hughy, line 13: Should `wallet/signTransaction` take an account?

We've deprecated the account field from `postTransaction` because you can determine the sender account from the raw transaction. Is it possible to determine the sender from an unsigned transaction?

ironfish-cli/src/commands/wallet/sign.ts
```diff
@@ -0,0 +1,158 @@
...
+import { Flags } from '@oclif/core'
+import { IronfishCommand } from '../../command'
+import { RemoteFlags } from '../../flags'
+import { longPrompt } from '../../utils/input'
+import { Ledger } from '../../utils/ledger'
+import { watchTransaction } from '../../utils/transaction'
+
+export class SignTransaction extends IronfishCommand {
+  static description = `Sign an unsigned transaction`
+  static hidden = true
```
hughy, line 15: Would this be hidden because of the ledger feature?

ironfish/src/rpc/routes/wallet/signTransaction.ts
patnir: Done!

ironfish/src/rpc/routes/wallet/signTransaction.ts
```diff
@@ -0,0 +1,63 @@
...
+        randomizedPublicKey,
+    )
+
+    if (!account) {
+      throw new Error('Wallet does not contain sender account for this transaction.')
+    }
+
+    if (!account.spendingKey) {
+      throw new Error('Account does not have a spending key')
+    }
```
hughy, line 55: nit - we could include this check in the predicate function for `findAccount`, above

ironfish-cli/src/commands/wallet/sign.ts
```diff
@@ -0,0 +1,134 @@
...
+
+      if (!response.content.account) {
+        this.error(
+          `No account is currently active.
+             Use ironfish wallet:create <name> to first create an account`,
+        )
+      }
+
+      account = response.content.account.name
+    }
```
hughy, line 57: `account` isn't used now that we can infer it from the transaction

ironfish-cli/src/commands/wallet/sign.ts
```diff
@@ -0,0 +1,134 @@
...
+    }),
+    broadcast: Flags.boolean({
+      default: false,
+      allowNo: true,
+      description: 'Broadcast the transaction to the network after signing',
+    }),
+    watch: Flags.boolean({
+      default: false,
+      description: 'Wait for the transaction to be confirmed',
+    }),
```
hughy, line 38: Should this only work if `broadcast` is set?

ironfish/src/rpc/routes/wallet/signTransaction.ts
patnir: I think distinct messages are better. If the user has the account but it is a view only account, that would be a different error than just not having the account

hughy (approved):

PR #5113, opened 2024-07-08
Remove `@types/jest`, and use Jest's own type annotations instead

andiflabs:
## Summary

This removes `@types/jest` in favor of `@jest/types` and friends. This allows using jest and expect to their full potential, and allow using new features like this: https://github.com/iron-fish/ironfish/blob/d8dffe54b17a3555efb013230726c3bb19eef784/ironfish/src/testUtilities/matchers/transaction.ts.

`@types/jest` was probably added at a time when Jest did not have any type annotations, but now that jest does have the annotations, it makes sense (in my opinion) to switch to using those. Some drawbacks:

- jest requires more verbose type annotations
- jest does not declare any global objects, so we have to do it by ourselves
- jest's types are more strict, and slightly incorrect code that used to compile before no longer compiles (this is not necessarily a drawback, but it's the reason why the delta is so large)

In addition to removing `@types/jest`, this PR also:

- removes some unused test functions
- changes incorrect uses of `mockResolvedValue(Promise.resolved(x))` with `mockResolvedValue(x)`
- changes incorrect uses of `toMatchObject(x)` with `toEqual(x)`

## Testing Plan

* yarn test

## Documentation

N/A

## Breaking Change

N/A


ironfish-cli/src/testSetup.ts
```diff
@@ -0,0 +1,23 @@
...
+  const fit: Global.GlobalAdditions['fit']
+  const xit: Global.GlobalAdditions['xit']
+  const xtest: Global.GlobalAdditions['xtest']
+  const describe: Global.GlobalAdditions['describe']
+  const xdescribe: Global.GlobalAdditions['xdescribe']
+  const fdescribe: Global.GlobalAdditions['fdescribe']
+  const beforeAll: Global.GlobalAdditions['beforeAll']
+  const beforeEach: Global.GlobalAdditions['beforeEach']
+  const afterEach: Global.GlobalAdditions['afterEach']
+  const afterAll: Global.GlobalAdditions['afterAll']
```
andiflabs, line 20: If any TypeScript expert knows how to put everything from `Global.GlobalAdditions` into the global namespace with 1 line, please let me know!

rohanjadvani:
```ts
const {
  it, 
  test, 
  fit, 
  xit, 
  xtest, 
  describe, 
  xdescribe, 
  fdescribe, 
  beforeAll, 
  beforeEach, 
  afterEach, 
  afterAll,
}: Global.GlobalAdditions;
```

Does this work for what you want?

rohanjadvani (approved):

ironfish-cli/src/testSetup.ts
andiflabs: That's better, thanks. What I would really like to do is avoiding writing `it`, `test`, `fit`, `xit`, ... I would like to do something like `import * from Global.GlobalAdditions`

rohanjadvani: I don't think TS has a way to import all properties from a type into the global namespace, but maybe someone else knows a better way.

PR #5118, opened 2024-07-08
Implement custom CLI table logic

mat-if:
## Summary

Since oclif 4.x removes their table feature, lets implement our own. The logic
is straight-forward enough that doing this was a reasonable task. The
alternative CLI table libraries didn't have the specific features we use, and
were generally over-engineered for our use.

This does remove some functionality that we were not using, as well as
potentially removing some extra usability that users may miss. We can re-add
these things as needed.

## Testing Plan

Manually tested the commands side-by-side

## Documentation

N/A

## Breaking Change

N/A

### DIFF
--- a/ironfish-cli/package.json
+++ b/ironfish-cli/package.json
@@ -77,6 +77,8 @@
     "inquirer": "8.2.5",
     "json-colorizer": "2.2.2",
     "keccak": "3.0.4",
+    "natural-orderby": "3.0.2",
+    "string-width": "4.2.3",
     "supports-hyperlinks": "2.2.0",
     "tar": "6.1.11",
     "uuid": "8.3.2"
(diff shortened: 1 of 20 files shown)


ironfish-cli/src/ui/confirm.ts
```diff
@@ -0,0 +1,31 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
mat-if, line 1: This is unchanged. It was moved out of the `ui.ts` file, which was broken into smaller pieces

ironfish-cli/package.json
```diff
@@ -77,6 +77,8 @@
     "inquirer": "8.2.5",
     "json-colorizer": "2.2.2",
     "keccak": "3.0.4",
+    "natural-orderby": "3.0.2",
```
mat-if, line 80: Neither of these are "new" - they were both used by oclif's table. They have been promoted to direct dependencies

hughy (commented):
This looks good to me and works as expected on all of the affected commands

Just one comment re: flags for features that we don't support (columns, truncation)

hughy (approved):

PR #5122, opened 2024-07-09
add ledger flag to send command

patnir:
## Summary

Add `--ledger` to `wallet:send` command. 

## Testing Plan

Send transactions with a ledger device using the `--ledger` flag with the `--watch` and `--confirm` flags. Send a transaction with another account that isn't the ledger account. 

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/send.ts
+++ b/ironfish-cli/src/commands/wallet/send.ts
@@ -8,6 +8,7 @@ import {
   isValidPublicAddress,
   RawTransaction,
   RawTransactionSerde,
+  RpcClient,
   TimeUtils,
   Transaction,
 } from '@ironfish/sdk'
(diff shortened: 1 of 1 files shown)


hughy (approved):
Looks good to me!

The only thing I noticed is that the transaction summary that gets printed in the CLI doesn't account for change notes, so the number of outputs there may differ from the number of outputs that you confirm on the Ledger

patnir:
> Looks good to me!
> 
> The only thing I noticed is that the transaction summary that gets printed in the CLI doesn't account for change notes, so the number of outputs there may differ from the number of outputs that you confirm on the Ledger

Sweet, I'll merge and look into the change notes in a separate issue

PR #5130, opened 2024-07-11
Upgrade to oclif 4.x

mat-if:
## Summary

The final oclif 4.x upgrade PR

## Documentation

N/A

## Breaking Change

N/A


ironfish-cli/bin/run
```diff
@@ -24,7 +24,4 @@ if (Number(process.versions.node.split('.')[0]) < MIN_NODE_VERSION) {
   process.exit(1)
 }
 
-require('@oclif/core')
-  .run()
-  .then(require('@oclif/core/flush'))
-  .catch(require('@oclif/core/handle'))
+require('@oclif/core').execute({ dir: __dirname })
```
mat-if, line 27: This change makes me a *little* nervous, but is the new? default according to the docs. Doing it the way we had was making it throw stack traces for even non-error things which was very strange.

ironfish-cli/src/commands/peers/banned.ts
```diff
@@ -19,7 +19,7 @@ export class BannedCommand extends IronfishCommand {
     sort: {
       ...sort,
       exclusive: ['follow'],
-    },
+    } as typeof sort,
```
mat-if, line 22: :(

andiflabs (approved):

PR #5132, opened 2024-07-11
add stateCommitment to block header

hughy:
## Summary

adds an optional stateCommitment field to the block header

updates serialization and deserialization of block headers to include the stateCommitment, if present

adds logic to verifyConnectedBlock to verify that the stateCommitment from the header is equal to the stateRoot from the state manager after all transactions have been executed

## Testing Plan

set activation sequence for EVM on devnet and started mining blocks that include stateCommitment after activation sequence

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/utils/serializers.ts
+++ b/ironfish/src/network/utils/serializers.ts
@@ -11,9 +11,7 @@ import {
 } from '../../primitives/block'
 import { RawBlockHeader } from '../../primitives/blockheader'
 import { Target } from '../../primitives/target'
-import { Transaction } from '../../primitives/transaction'
-
-export const MINERS_FEE_TRANSACTION_SIZE_BYTES = 664
+import { Transaction, TransactionVersion } from '../../primitives/transaction'
 
 const BLOCK_TRANSACTIONS_LENGTH_BYTES = 2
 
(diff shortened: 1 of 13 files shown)


ironfish/src/primitives/blockheader.ts
```diff
@@ -203,6 +206,9 @@ export class BlockHeader {
     bw.writeBigU256BE(this.target.asBigInt())
     bw.writeU64(this.timestamp.getTime())
     bw.writeBytes(this.graffiti)
+    if (this.stateRoot) {
```
dguenther, line 209: There's a constant `180` in this function that this will probably run afoul of -- might need to search around and see if we have the block header size hardcoded anywhere 😬

ironfish/src/primitives/blockheader.ts
```diff
@@ -150,6 +150,8 @@ export class BlockHeader {
    */
   public readonly graffiti: Buffer
 
+  public readonly stateRoot?: Buffer
```
dguenther, line 153: The actual name of this is not super-important, since we could theoretically rename it without a fork, but I feel like using `commitment` instead of `root` might better match the other fields on the block header

ironfish/src/primitives/blockheader.ts
hughy: Good call! I noticed that we have this `180` in the `blockHasher` code, but I

ironfish/src/primitives/blockheader.ts
hughy: That sounds good to me, I'm not afraid of `commitment`

ironfish/src/network/utils/serializers.ts
```diff
@@ -106,12 +104,22 @@ export function getBlockSize(block: RawBlock): number {
...
+    getBlockHeaderSize() +
+    BLOCK_TRANSACTIONS_LENGTH_BYTES +
+    getMinersFeeTransactionSize(transactionVersion)
   )
 }
 
+export function getMinersFeeTransactionSize(transactionVersion: TransactionVersion): number {
+  let size = 664
+  if (transactionVersion >= TransactionVersion.V3) {
+    size += 1
```
jowparks, line 118: This is a preexisting bug that is fixed here

jowparks (approved):

PR #5134, opened 2024-07-15
Rahul/ifl 2763 add solidity contract to ironfish

patnir:
## Summary

Sets up hardhat for local development so we don't have to rely on online IDEs and can have version control. 
These contracts will become a core component of EVM development so we should include it in our code base and iterate over it. 
Sets up a basic contract that emits a shield event. Adds a test to confirm this. 
Updates the test-evm command to call this smart contract and confirm the logs from this event. 
We will then take the logs from this event and add the appropriate mints/ burns to the transaction object. 

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/test-evm.ts
+++ b/ironfish-cli/src/commands/test-evm.ts
@@ -3,9 +3,8 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { LegacyTransaction } from '@ethereumjs/tx'
 import { Account, Address } from '@ethereumjs/util'
-import { Asset, generateKey } from '@ironfish/rust-nodejs'
-import { IronfishEvm } from '@ironfish/sdk'
-import { ethers } from 'ethers'
+import { generateKey } from '@ironfish/rust-nodejs'
+import { IronfishEvm } from '@ironfish/sdk/src/evm'
 import { IronfishCommand } from '../command'
 import { LocalFlags } from '../flags'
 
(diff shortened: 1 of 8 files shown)


ironfish-contracts/contracts/Ironfish.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.18;
+
+contract Ironfish {
+    event Shield(string assetId, uint amount);
+
+    constructor() {}
+
+    function shield(string calldata assetId, uint amount) public {
+        emit Shield(assetId, amount);
```
danield9tqh, line 10: Can we change this to the `bytes` event

ironfish-cli/src/commands/test-evm.ts
```diff
@@ -28,40 +29,108 @@ export class TestEvmCommand extends IronfishCommand {
     const senderAddress = Address.fromPrivateKey(senderPrivateKey)
     const recipientAddress = Address.fromPrivateKey(recipientPrivateKey)
 
-    const senderAccount = new Account(BigInt(0), 500000n)
+    const senderAccount = new Account(BigInt(0), 500_000_000n)
```
danield9tqh, line 32: Wanna create a new command for testing this?

ironfish-contracts/contracts/Ironfish.sol
jowparks: We decided to keep it this way for now

jowparks (approved):

PR #5135, opened 2024-07-16
Optimize `WalletDB.loadNoteHash()` to avoid hitting the database if the note nullifier is known to be absent

andiflabs:
## Summary

This commit adds an in-memory bloom filter to remember which nullifiers are stored in the database. The bloom filter is used by `WalletDB.loadNoteHash()` to avoid querying the database if the given nullifier is known to be not present.

While at it, `WalletDB.getTransactionHashFromNullifier()` was optimized in the same way.

## Testing Plan

* unit tests
* `wallet:rescan`

## Documentation

N/A

## Breaking Change

N/A

### DIFF
--- a/ironfish/src/utils/bloomFilter.ts
+++ b/ironfish/src/utils/bloomFilter.ts
@@ -1,0 +1,77 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+function isPowerOf2(n: number): boolean {
+  // This checks that n is non-zero, and that n and n-1 have no bits in common
+  // (taken from https://stackoverflow.com/a/30924333)
+  return !!(n && !(n & (n - 1)))
+}
+
+/**
+ * Simple implementation of a bloom filter. This is a set-like data structure
+ * that can be used to check whether certain items may or may not be present in
+ * the set. This is a probabilistic data structure that does not return false
+ * negatives, but may return false positives.
+ *
+ * Usually, bloom filters work by computing the hash of each item added to the
+ * set. This implementation differs in that it does not compute any hash for
+ * efficiency, and instead assumes that each item contains random data. Items
+ * that share the same prefix will result in collisions/false positives.
+ */
+export class BloomFilter {
+  private readonly bitMask: number
+  private readonly bitArray: Buffer
+
+  constructor(bits: number) {
+    if (bits < 8) {
+      // Need to have at least 1 byte
+      throw new Error(`bits must be at least 8 (got ${bits})`)
+    }
+    if (bits >= 1 << 30) {
+      // `bits` cannot exceed 2**30 because `indexOf` relies on
+      // `Buffer.readUInt32LE` (also, the array would be larger than 128 MiB,
+      // which may not be desiderable). This is 30 and not 32 because `1 << 32
+      // === 1` and `1 << 31 === -2147483648`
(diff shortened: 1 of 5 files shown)


ironfish/src/wallet/walletdb/walletdb.ts
```diff
@@ -740,12 +743,40 @@ export class WalletDB {
...
+    }
+  }
+
   async loadNoteHash(
     account: Account,
     nullifier: Buffer,
     tx?: IDatabaseTransaction,
   ): Promise<Buffer | undefined> {
-    return this.nullifierToNoteHash.get([account.prefix, nullifier], tx)
+    const nullifierFilter = await this.loadNullifierBloomFilter(tx)
```
hughy, line 774: should we call `loadNullifierBloomFilter` when the wallet loads?

ironfish/src/wallet/walletdb/walletdb.ts
```diff
@@ -1243,13 +1284,18 @@ export class WalletDB {
...
+    }
   }
 
   async deleteNullifierToTransactionHash(
     account: Account,
     nullifier: Buffer,
     tx?: IDatabaseTransaction,
   ): Promise<void> {
+    // TODO: maybe keep a count of how many times this method was called, and
+    // reset the bloom filter after a certain threshold
```
hughy, line 1298: I don't think we need to worry about nullifiers here. Entries in `nullifierToTransactionHash` are only deleted if a pending transaction expires. The nullifier is still valid for that account, but that transaction isn't.

ironfish/src/wallet/walletdb/walletdb.ts
```diff
@@ -780,6 +814,8 @@ export class WalletDB {
...
```
hughy, line 818: I think that nullifiers are deleted so infrequently that it might not be worth the added complexity to reset the bloom filter

hughy (approved):
👍 👍

ironfish/src/wallet/walletdb/walletdb.ts
andiflabs: Not all usages of the wallet/walletdb may require access to the notes/nullifiers, so I preferred to go with this lazy loading route. If someone has a lot of notes, this can *potentially* add a noticeable performance latency to every command/rpc call.

ironfish/src/wallet/walletdb/walletdb.ts
andiflabs: Agreed, that's why I did not implement it. I might turn this "TODO" into a simple comment saying that the bloom filter is unchanged after a delete, and that's by design, because \<explanation\>.

ironfish/src/utils/bloomFilter.ts
```diff
@@ -0,0 +1,58 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+function isPowerOf2(n: number): boolean {
+  // This checks that n is non-zero, and that n and n-1 have no bits in common
+  // (taken from https://stackoverflow.com/a/30924333)
+  return !!(n && !(n & (n - 1)))
+}
+
+export class BloomFilter {
```
andiflabs, line 11: Forgot to add a doc comment here explaining the design choices & the assumptions. I might also add a `hashFn` parameter to the constructor that equals the identity function by default.

ironfish/src/wallet/walletdb/walletdb.ts
hughy: That makes sense. Now that I think about it, calling this during `load` would be a major performance drag on every wallet command that a user runs without a running node.

PR #5136, opened 2024-07-17
Decode Shield Events in Verifier

patnir:
## Summary

Decodes the shield event when processing the evm transaction
Slight tweaks to the ironfish.sol contract 

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/evm/evm.ts
```diff
@@ -33,17 +35,41 @@ export class IronfishEvm {
     const result = await this.runTx(opts)
 
     // TODO(jwp) from custom opcodes populate shields and unshields
+
+    const shields = []
+
+    for (const log of result.receipt.logs) {
+      if (Buffer.from(log[0]).toString('hex') !== 'globalContractAddress') {
```
hughy, line 42: is `'globalContractAddress'` a placeholder here until we determine the hex address?

patnir: yeah

ironfish-contracts/contracts/Ironfish.sol
```diff
@@ -2,11 +2,12 @@
 pragma solidity ^0.8.18;
 
 contract Ironfish {
```
jowparks, line 4: Can we move this to the sdk? If we do this we will be able to export and use directly in the rest of the sdk and cli

patnir: Yeah I can move this to a separate PR

ironfish/src/evm/evm.ts
```diff
@@ -33,17 +35,42 @@ export class IronfishEvm {
     const result = await this.runTx(opts)
 
     // TODO(jwp) from custom opcodes populate shields and unshields
+
```
jowparks, line 38: Can we add a test for shields here? One where shield comes from non-globaladdress and one that does? We will probably need to define a global address for this.

jowparks (approved):

danield9tqh (approved):

PR #5138, opened 2024-07-17
updating bidirectional mapping for asset to contract/token

jowparks:
## Summary
Updates mapping of Asset ID (ironfish) to EVM (contract token) to be bidirectional. Uses a prefix on the storage to prevent collisions.

## Testing Plan
tests pass
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -552,15 +552,15 @@ export class Blockchain {
     return this.consensus.checkpoints.get(header.sequence)?.equals(header.hash) ?? false
   }
 
-  isEvmAsset(assetId: Buffer, tx?: BlockchainDBTransaction): Promise<boolean> {
-    return this.blockchainDb.assetIdToContract.has(assetId, tx)
+  async isEvmAsset(assetId: Buffer, tx?: BlockchainDBTransaction): Promise<boolean> {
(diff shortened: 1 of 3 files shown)


hughy (commented):
I'm not sure that this is better than having two separate stores for `assetId -> contract` and `contract -> assetId`. Each 'store' is really just an abstraction with a shared schema and a specific shared key prefix (the store name) in the database.

hughy (approved):

ironfish/src/blockchain/blockchain.ts
```diff
@@ -552,15 +552,15 @@ export class Blockchain {
...
+  async isEvmAsset(assetId: Buffer, tx?: BlockchainDBTransaction): Promise<boolean> {
+    return (await this.getEvmContractByAssetId(assetId, tx)) !== undefined
   }
 
   getEvmContractByAssetId(
     assetId: Buffer,
     tx?: BlockchainDBTransaction,
   ): Promise<Buffer | undefined> {
-    return this.blockchainDb.assetIdToContract.get(assetId, tx)
+    return this.blockchainDb.getContractTokenByAssetId(assetId, tx)
```
hughy, line 563: do we need to extract the contract address from the returned buffer here?

jowparks: Good call, I will return extraced values

PR #5140, opened 2024-07-18
executes evm transactions on blocks

hughy:
## Summary

executes evm transactions in Blockchain.newBlock so that mining manager executes evm transactions during block template creation

executes evm transactions when adding blocks to chain

changes 'verifyTx' to use 'simulateTx' to avoid modifying state, updates 'simulateTx' to turn off gas fees

updates BlockchainDBTransaction to reset the stateManager stateRoot when aborting transactions

adds 'abort' method to IronfishStateManager to revert all checkpoints when db transactions abort

adds evm cli commands for testing sending transaction with evm description

## Testing Plan

manual testing:
1. set activation sequence for evm transactions on devnet
2. start devnet node
3. stop devnet node
4. use a spending key with evm:putAccount to initialize an an account with public funds
5. start devnet node with '--forceMining'
6. start miner
7. use evm:sendTransaction to create and submit a transaction with an evm description (use the spending key and account nonce from step 4)
8. wait for transaction to be mined
9. stop devnet node, use evm:getAccount to check account balances and verify transaction was executed

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/evm/sendTransaction.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    const rawTransaction = new RawTransaction(TransactionVersion.V3)
+    rawTransaction.evm = legacyTransactionToEvmDescription(tx.sign(senderKey))
+
+    const transaction = rawTransaction.post(senderKey.toString('hex'))
+
+    const response = await client.mempool.acceptTransaction({
+      transaction: transaction.serialize().toString('hex'),
+    })
+
```
jowparks, line 64: Can we make these unit tests too, just so we verify nothing breaks when updating evm branch?

ironfish/src/blockchain/database/blockchaindb.ts
```diff
@@ -438,33 +438,34 @@ export class BlockchainDBTransaction implements IDatabaseTransaction {
...
     }
     await this.tx.update()
   }
 
   async commit(): Promise<void> {
-    if (this.checkpoint) {
+    if (this.stateRoot) {
       await this.stateManager.commit()
-      this.checkpoint = false
+      this.stateRoot = null
```
jowparks, line 459: Why set these to null instead of leaving them as the value? Is there a situation where the `stateRoot` will be different at the start of the next operation?

jowparks (approved):

ironfish/src/blockchain/database/blockchaindb.ts
hughy: Setting to null makes it so that we can't/won't roll back to the previous stateRoot after committing database changes. If we do something like this
```
withTransaction((tx) => {
  // add some values
  tx.update()
  
  // add some more values
  
  // throw an error
}
```
we want the stateRoot to include the first set of changes to be consistent with the other database stores

patnir:
Went through all the steps and it successfully worked: 

```
Account balance for address 0x6746b1bb362e3bb24355547da4a275f29901d8aa: 9999000
Account nonce for address 0x6746b1bb362e3bb24355547da4a275f29901d8aa: 1
```

patnir (approved):

PR #5142, opened 2024-07-18
Shield verification of mints 

jowparks:
## Summary
Verifies mints are valid, does not verify native asset.
## Testing Plan
test passes
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/consensus/verifier.evm.test.slow.ts
+++ b/ironfish/src/consensus/verifier.evm.test.slow.ts
@@ -137,9 +137,16 @@ describe('Verifier', () => {
 
       const globalContract = new ethers.Interface(ContractArtifact.abi)
 
+      const assetMetadata = {
+        creator: senderAccountIf.publicAddress,
+        name: 'foo',
+        metadata: '',
+      }
+      const asset = new Asset(assetMetadata.creator, assetMetadata.name, assetMetadata.metadata)
+
       const encodedFunctionData = globalContract.encodeFunctionData('shield', [
         Buffer.from(senderAccountIf.publicAddress, 'hex'),
-        Asset.nativeId(),
+        asset.id(),
         100n,
       ])
 
@@ -161,6 +168,12 @@ describe('Verifier', () => {
         fee: 0n,
         expiration: 0,
         expirationDelta: 0,
+        mints: [
+          {
+            ...assetMetadata,
+            value: 100n,
+          },
+        ],
         evm: description,
       })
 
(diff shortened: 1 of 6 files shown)


ironfish/src/consensus/verifier.evm.test.slow.ts
```diff
@@ -192,67 +205,9 @@ describe('Verifier', () => {
         senderAccountIf.publicAddress,
       )
       expect(shieldEvents[0].caller).toEqual(evmSenderAddress)
-      expect(shieldEvents[0].assetId).toEqual(Asset.nativeId())
-    })
-
```
jowparks, line 197: No longer relevant, will fail due to length mismatch of mint/shield

ironfish/src/consensus/verifier.ts
```diff
@@ -654,6 +654,41 @@ export class Verifier {
     // }
 
     // TODO(jwp): verify shielding/mints balance
+
+    const shieldBalance: BufferMap<bigint> = new BufferMap()
+    const mintBalance: BufferMap<bigint> = new BufferMap()
```
hughy, line 659: we could also do this with a single `BufferMap` or `AssetBalances` instance if we kept track of deltas/balances for each asset

ironfish/src/consensus/verifier.ts
```diff
@@ -654,6 +654,41 @@ export class Verifier {
...
+
+    for (const event of result.events) {
+      // TODO(jwp): handle native asset as output balance (will require decrypting notes and comparing)
```
hughy, line 662: I think we may need another description with public values for shielding the native asset -- both to get values to balance internally and to avoid decryption

ironfish/src/consensus/verifier.ts
```diff
@@ -654,6 +654,41 @@ export class Verifier {
...
+      if (event.assetId.equals(Asset.nativeId())) {
+        continue
+      }
+      if (event.name === 'shield') {
+        shieldBalance.set(event.assetId, event.amount)
+      }
+    }
+    for (const mint of transaction.mints) {
+      if (mint.asset.id().equals(Asset.nativeId())) {
```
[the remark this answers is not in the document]
jowparks: True it shouldn't be possible

ironfish/src/consensus/verifier.ts
```diff
@@ -654,6 +654,41 @@ export class Verifier {
...
+    const shieldBalance: BufferMap<bigint> = new BufferMap()
+    const mintBalance: BufferMap<bigint> = new BufferMap()
+
+    for (const event of result.events) {
+      // TODO(jwp): handle native asset as output balance (will require decrypting notes and comparing)
+      if (event.assetId.equals(Asset.nativeId())) {
+        continue
+      }
+      if (event.name === 'shield') {
+        shieldBalance.set(event.assetId, event.amount)
```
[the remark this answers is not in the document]
jowparks: Technically it is possible, you could call shield twice from a single smart contract call, resulting in two events.

ironfish/src/consensus/verifier.ts
jowparks: Yea maybe, although I don't think this is blocking us since we know that is possible. I think we should kick that down the road, wdyt?

hughy: Yeah I think we can definitely defer that to later

hughy (approved):

PR #5155, opened 2024-07-19
manager verifies evm transactions during template construction

hughy:
## Summary

as the mining manager takes transactions from the mempool during block template construction it performs some basic verification and skips any invalid transactions. this prevents the mining manager from getting stuck producing templates that include invalid transactions

each evm transaction in the mempool is verified individually, but may be invalid when sequenced after another evm transaction. for instance, if a later transaction uses the same nonce as an earlier transaction, or if it causes an insufficient funds error after an earlier transaction is executed.

adds logic to getNewBlockTransactions to call verifyEvm on each evm transaction taken from the mempool to avoid invalid evm transactions

uses withCopy to prevent database commits from evm transaction execution. transactions are still executed in Blockchain.newBlock, so we should avoid state updates to prevent transactions being invalid on the second execution

in the future we might look at refactoring block template construction and Blockchain.newBlock to avoid unnecessarily executing evm transactions multiple times, but for the spike it's not necessary to optimize

## Testing Plan

adds unit test

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/mining/manager.test.slow.ts
+++ b/ironfish/src/mining/manager.test.slow.ts
@@ -546,6 +546,94 @@ describe('Mining manager', () => {
       expect(node.memPool.count()).toBe(1)
     })
 
+    it('should not create block templates with invalid evm transactions', async () => {
+      const { node, chain } = nodeTest
+      const { miningManager } = node
+
+      jest
+        .spyOn(node.chain.consensus, 'getActiveTransactionVersion')
+        .mockImplementation(() => TransactionVersion.V3)
+
+      const account = await useAccountFixture(nodeTest.node.wallet, 'account')
+      await nodeTest.node.wallet.setDefaultAccount(account.name)
+
+      const previous = await useMinerBlockFixture(chain, 2, account, node.wallet)
+      await expect(chain).toAddBlock(previous)
+      await node.wallet.scan()
+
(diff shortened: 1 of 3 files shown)


ironfish/src/mining/manager.test.slow.ts
```diff
@@ -546,6 +546,94 @@ describe('Mining manager', () => {
...
+      const ethAccount = new EthAccount(0n, 1000n)
+
+      const evmRecipientAddress = Address.fromString(
+        '0x3c6c6d51f71a0f146cf79843e888feb20258f567',
+      )
+
+      await node.chain.blockchainDb.transaction(async (_) => {
+        await node.chain.blockchainDb.stateManager.putAccount(ethAddress, ethAccount)
+      })
+
```
jowparks, line 574: Interesting question, what if a new account submits two transactions at once, but the mining manager tries to add the one with a higher nonce first?

jowparks (approved):

ironfish/src/mining/manager.test.slow.ts
hughy: In that case I think the one with the higher nonce would be skipped as invalid (assuming the higher nonce doesn't match the account nonce)

jowparks: Ahh yes you are right!

PR #5156, opened 2024-07-19
add all functions to global contract

danield9tqh:
## Summary
- Add all functions to global contract 
- calculate asset id based on global constants

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-contracts/contracts/Ironfish.sol
```diff
@@ -4,19 +4,38 @@ pragma solidity ^0.8.18;
...
-        emit Shield(ironfishAddress, ironfishAssetId, msg.sender, amount);
+        emit Shield(ironfishAddress, tokenId, msg.sender, amount);
+    }
+
+    function shield_iron(bytes32 ironfishAddress) public payable {
+        emit Shield(ironfishAddress, 0, address(this), msg.value);
+    }
+
+    function unshield(uint tokenId, uint amount) public {
+        emit UnShield(msg.sender, tokenId, amount);
```
jowparks, line 34: For unshields, we generally need a `to address` in all, right?

ironfish/src/evm/evm.ts
```diff
@@ -96,6 +71,54 @@ export class IronfishEvm {
...
+
+        events.push({
+          name: 'shield',
+          ironfishAddress: Buffer.from((ironfishAddress as string).slice(2), 'hex'),
+          caller: Address.fromString(caller as string),
+          assetId: this.getAssetId(caller as string, tokenId as bigint),
+          amount: amount as bigint,
+        })
+      } catch (e) {
+        try {
```
jowparks, line 100: Maybe separate try/catches and add a continue statement at end of each try?

ironfish-contracts/contracts/Ironfish.sol
```diff
@@ -4,19 +4,38 @@ pragma solidity ^0.8.18;
 contract Ironfish {
     event Shield(
         bytes32 ironfishAddress,
-        bytes32 ironfishAssetId,
+        uint tokenId,
         address contractAddress,
         uint amount
     );
```
jowparks, line 11: nit: not sure we need camel casing here.

ironfish-contracts/contracts/Ironfish.sol
```diff
@@ -4,19 +4,38 @@ pragma solidity ^0.8.18;
...
         bytes32 ironfishAddress,
-        bytes32 ironfishAssetId,
+        uint tokenId,
         uint amount
     ) public {
         // msg.sender is the contract address of the ERC20 token
-        emit Shield(ironfishAddress, ironfishAssetId, msg.sender, amount);
+        emit Shield(ironfishAddress, tokenId, msg.sender, amount);
+    }
+
```
jowparks, line 28: If we are making Iron the native currency, rather than ERC20, then instead we would just be looking for sends to the global contract address and get the amount from that.

jowparks (approved):

ironfish-contracts/contracts/Ironfish.sol
```diff
@@ -4,19 +4,38 @@ pragma solidity ^0.8.18;
...
+    function shield_iron(bytes32 ironfishAddress) public payable {
```
hughy, line 29: nit - since IRON is the native asset I tink that `shield` and `unshield` should be for IRON, and something like `shield_token` would be for custom tokens

hughy (approved):

ironfish-contracts/contracts/Ironfish.sol
jowparks: Hmm I dunno, I think you could argue either way. I think people will generally interact with shield token more, so from a usability perspective might make more sense to have it the way Daniel has it.

ironfish/src/evm/evm.ts
danield9tqh: not sure what you mean here. if we `continue` then we just skip that log

PR #5157, opened 2024-07-19
Changes how EVM is instantiated.

patnir:
Previously, we needed to check whether the EVM on the blockchain object was instantiated. There shouldn't be a case where EVM is not instantiated. We should always have an EVM instance.

That change was made because instantiating EVM requires async code because of the VM instantiation.

This change changes that pattern to a synchronous instantiation. The VM instantiation now happens in the open function that is no longer static. We check whether the object has been instantiated in the sub functions of the EVM class.

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -63,7 +63,7 @@ export class Blockchain {
   files: FileSystem
   consensus: Consensus
   // TODO: the EVM creation is async, so this has to be optional for the constructor
-  evm?: IronfishEvm
+  evm: IronfishEvm
   seedGenesisBlock: SerializedBlock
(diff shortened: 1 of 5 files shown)


ironfish/src/blockchain/blockchain.ts
```diff
@@ -63,7 +63,7 @@ export class Blockchain {
   files: FileSystem
   consensus: Consensus
   // TODO: the EVM creation is async, so this has to be optional for the constructor
-  evm?: IronfishEvm
+  evm: IronfishEvm
```
patnir, line 66: core behavior change

hughy (approved):

PR #5162, opened 2024-07-22
Update style guide to document command output info

NullSoldier:
## Summary

This should help users figure out how to design the output of their commands.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/STYLE_GUIDE.md
+++ b/ironfish-cli/STYLE_GUIDE.md
@@ -69,7 +69,7 @@ However, if prompting is required to complete a command, this means the user wil
 
 ## Output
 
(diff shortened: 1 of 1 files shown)


ironfish-cli/STYLE_GUIDE.md
```diff
@@ -69,6 +69,20 @@ However, if prompting is required to complete a command, this means the user wil
 
 ## Output
 
+When designing the output for a command, commands should output human readable output and not machine readable output. This means you should use components under the `ui` module such as `card`, `table`, or normal logs. It's fine if you only display a simplified version of the output.If the user needs the full data in machine readable format they can use the `--json` flag.
```
mat-if, line 72:
```suggestion
When designing the output for a command, commands should output human readable output and not machine readable output. This means you should use components under the `ui` module such as `card`, `table`, or normal logs. It's fine if you only display a simplified version of the output. If the user needs the full data in machine readable format they can use the `--json` flag.
```

mat-if (approved):

PR #5168, opened 2024-07-23
add global contract on startup

danield9tqh:
## Summary
Initialize global contract on startup and test shielding IRON

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/evm/deploy-contract.ts
```diff
@@ -32,26 +32,15 @@ export class TestEvmCommand extends IronfishCommand {
     await node.chain.blockchainDb.stateManager.putAccount(senderAddress, senderAccount)
     await node.chain.blockchainDb.stateManager.commit()
 
-    const globalContractAddress = Address.fromString(
-      '0xffffffffffffffffffffffffffffffffffffffff',
-    )
-
-    await node.chain.blockchainDb.stateManager.checkpoint()
-    await node.chain.blockchainDb.stateManager.putContractCode(
```
hughy, line 40: should we remove this command altogether since it doesn't deploy a contract anymore? or leave it for now and reuse it for deploying a test ERC20?

ironfish/src/evm/evm.ts
```diff
@@ -35,6 +39,19 @@ export class IronfishEvm {
     this.blockchainDb = blockchainDb
   }
 
+  async load(): Promise<void> {
+    const stateRoot = await this.blockchainDb.stateManager.getStateRoot()
+
+    if (Buffer.from(stateRoot).equals(NULL_STATE_ROOT)) {
```
hughy, line 45: in the future do you think this should be a db migration or always checked on load? it _should_ only need to run once

hughy (approved):

ironfish/src/evm/evm.ts
danield9tqh: Feel like the same way we do genesis block makes sense. Maybe both should have been a migration? but want to keep it in the same place as genesis for now

PR #5172, opened 2024-07-24
Add the ability for CLI commands to track and close an RPC client

mat-if:
## Summary

With the addition of the JSON changes, which requires us to await on the start command in the IronfishCommand, commands that used `this.sdk.connectRpc()` were hanging because the clients were not being closed. This change allows us to fetch a client from the IronfishCommand directly, which gives the IronfishCommand the ability to track the client, and close it when the code is done being ran/executed.

Closes IFL-2823

## Documentation

N/A

## Breaking Change

N/A

### DIFF
--- a/ironfish-cli/src/command.ts
+++ b/ironfish-cli/src/command.ts
@@ -9,6 +9,7 @@ import {
   InternalOptions,
   IronfishSdk,
   Logger,
+  RpcClient,
   RpcConnectionError,
 } from '@ironfish/sdk'
 import { Command, Config, ux } from '@oclif/core'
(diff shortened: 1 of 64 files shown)


ironfish-cli/src/command.ts
```diff
@@ -78,7 +81,9 @@ export abstract class IronfishCommand extends Command {
 
   async run(): Promise<unknown> {
     try {
-      return await this.start()
+      const json = await this.start()
+      this.client?.close()
```
NullSoldier, line 85: Should this be in a finally() below?

ironfish/src/rpc/clients/memoryClient.ts
```diff
@@ -29,4 +29,8 @@ export class RpcMemoryClient extends RpcClient {
 
     return RpcMemoryAdapter.requestStream<TEnd, TStream>(this.router, route, data)
   }
+
+  close(): void {
+    return
```
NullSoldier, line 34:
```suggestion
```

NullSoldier (approved):

NullSoldier:
This looks good, and should prevent issues where commands don't clean up their clients and cause commands to hang.

PR #5191, opened 2024-07-31
moves evm transaction signing into rust

hughy:
## Summary

signs evm transaction descriptions at signing time for other transaction descriptions

defines UnsignedEvmDescription to exclude signature fields (v, r, s) from EvmDescription

replaces EvmDescription with UnsignedEvmDescription on ProposedTransaction and UnsignedTransaction

uses reth_primitives crate to sign transactions in rust

upgrades napi and other dependencies as required to add reth_primitives dependency

## Testing Plan

- adds new tests in ironfish-rust to test signing
- modifies verifier.evm.test.slow.ts to accommodate new signing logic

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/transaction/evm.rs
+++ b/ironfish-rust/src/transaction/evm.rs
@@ -7,9 +7,92 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
+use reth_primitives::{
+    revm_primitives::FixedBytes, sign_message, Address, Transaction, TxKind, TxLegacy, U256,
+};
 use std::io;
 
-use crate::errors::IronfishError;
+use crate::{errors::IronfishError, SaplingKey};
+#[derive(Clone, PartialEq, Debug)]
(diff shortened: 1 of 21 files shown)


ironfish-rust/src/transaction/evm.rs
```diff
@@ -35,26 +27,89 @@ impl EvmDescription {
...
+            nonce: self.description.nonce,
+            gas_price: self.description.gas_price.into(),
+            gas_limit: self.description.gas_limit,
+            to: tx_kind,
+            value: U256::from(self.description.value),
+            input: self.description.data.clone().into(),
+            chain_id: None,
+        });
+
+        // TODO(hughy): add appropriate error handling for signing errors from reth
```
patnir, line 82: does this need to be tackled here?

patnir (approved):

ironfish-rust/src/transaction/evm.rs
hughy: I think I'm going to leave the unwrap for now 😅

PR #5206, opened 2024-08-05
Add verification for native asset shield and unshield

danield9tqh:
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/consensus/verifier.ts
+++ b/ironfish/src/consensus/verifier.ts
@@ -665,10 +665,6 @@ export class Verifier {
     const assetBalanceDeltas = new AssetBalances()
 
     for (const event of result.events) {
-      // TODO(jwp): handle native asset as output balance (will require decrypting notes and comparing)
-      if (event.assetId.equals(Asset.nativeId())) {
-        continue
-      }
       if (event.name === 'shield') {
         assetBalanceDeltas.increment(event.assetId, -event.amount)
       }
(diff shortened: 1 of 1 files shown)


ironfish/src/consensus/verifier.ts
```diff
@@ -689,27 +694,26 @@ export class Verifier {
...
+      }
+    }
+
     const assetBalanceDeltas = new AssetBalances()
 
     for (const event of result.events) {
-      if (event.assetId.equals(Asset.nativeId())) {
-        return {
-          valid: false,
-          reason: VerificationResultReason.EVM_UNSHIELD_EVENT_NATIVE_ASSET,
```
hughy, line 699: can we get rid of this `VerificationResultReason`?

hughy (approved):

ironfish/src/consensus/verifier.ts
danield9tqh: The way I understand the verifying we have two steps:
1. Transaction balancing in rust which incorporates `evm.publicIron` and `evm.privateIron` with the native asset spends and outputs.
2. In the verifier check that `evm.publicIron` matches the number of native asset unshields and `evm.privateIron` matches the number of native asset shields

These checks should be sufficient. Joe said this check was just there because the checks had not yet been implemented yet so we shouldn't need it

jowparks: Yup both right. We can get rid of it.

PR #5211, opened 2024-08-06
moves evm tx execution to verifyBlock

hughy:
## Summary

we will need to use the result of evm transaction execution to calculate miner fees (including gas from evm transactions) in verifyBlock

executes evm transactions in verifyBlock instead of saveConnect and newBlock

makes evm transaction verification in verifyBlock optional so that mining manager does not execute evm transactions more than once when creating block templates and verifying them

makes evm transaction execution in block template creation commit state to the stateManager (i.e., removes use of withCopy)

## Testing Plan

- manual testing to ensure that we can still mine blocks with evm transactions

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/blockchain.ts
+++ b/ironfish/src/blockchain/blockchain.ts
@@ -686,6 +686,11 @@ export class Blockchain {
       throw new VerifyError(reason, BAN_SCORE.MAX)
     }
 
+    // set state root back to head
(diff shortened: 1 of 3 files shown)


hughy:
Closing this and rethinking the approach.

We do need to execute EVM transaction during block verification, but we should avoid mutating the state trie since block verification may happen more than once.

PR #5214, opened 2024-08-07
pays gas fees to miners in minersFee transactions

hughy:
## Summary

moves evm tx execution out of verifyEvm

consolidates IronfishEvm methods runTx and verifyTx into a single method that returns a result and events decoded from logs or an error if exeuction fails

removes redundant simulateTx method

refactors verifyEvm to take an EvmResult instead of executing the transaction itself. allows us to use evm transaction execution results outside of verification, e.g., for summing miner fees from gas

adds runDesc method for convenience to wrap conversion from EvmDescription to evm transaction

## Testing Plan

manual testing to see miners fee includes gas:

- apply attached diff: [miner_fee_gas_test.diff.txt](https://github.com/user-attachments/files/16532152/miner_fee_gas_test.diff.txt)
- start node with `--networkId=2` and `--forceMining`
- start miner
- use `evm:sendTransaction` with the spending key from the diff to send an evm transaction
- use `wallet:transactions` to see miners fee transaction that includes 21000 ORE from gas

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/evm/shield-iron.ts
+++ b/ironfish-cli/src/commands/evm/shield-iron.ts
@@ -64,6 +64,7 @@ export class TestEvmCommand extends IronfishCommand {
 
     const { events } = await evm.runTx({ tx: tx.sign(senderPrivateKey) })
 
+    Assert.isNotUndefined(events)
     Assert.isEqual(events.length, 1)
     const log = events[0] as EvmShield
 
--- a/ironfish-cli/src/commands/evm/shield-iron.ts
+++ b/ironfish-cli/src/commands/evm/shield-iron.ts
@@ -62,12 +62,10 @@ export class TestEvmCommand extends IronfishCommand {
       data: data,
     })
 
-    const result = await evm.runTx({ tx: tx.sign(senderPrivateKey) })
+    const { events } = await evm.runTx({ tx: tx.sign(senderPrivateKey) })
 
-    const logEvents = evm.decodeLogs(result.receipt.logs)
-
-    Assert.isEqual(logEvents.length, 1)
-    const log = logEvents[0] as EvmShield
+    Assert.isEqual(events.length, 1)
+    const log = events[0] as EvmShield
 
     this.log('Contract Address')
     this.log(log.caller.toString())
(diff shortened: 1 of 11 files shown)


ironfish-cli/src/commands/evm/shield-iron.ts
```diff
@@ -62,12 +62,10 @@ export class TestEvmCommand extends IronfishCommand {
       data: data,
     })
```
jowparks, line 64: Nice

jowparks (approved):

ironfish/src/evm/evm.ts
```diff
@@ -183,7 +190,18 @@ export type EvmUnshield = {
...
+export type EvmResult =
+  | {
+      result: RunTxResult
+      error: undefined
+      events: UTXOEvent[]
+    }
+  | {
+      result: undefined
+      error: EvmError
+      events: UTXOEvent[]
```
danield9tqh, line 202: should events be undefined here?

ironfish/src/consensus/verifier.ts
```diff
@@ -179,11 +179,13 @@ export class Verifier {
       totalTransactionFees += transaction.fee()
     }
 
-    // minersFee should be (negative) miningReward + totalTransactionFees
-    const miningReward = this.chain.network.miningReward(block.header.sequence)
-
-    if (minersFeeTransaction.fee() !== -1n * (BigInt(miningReward) + totalTransactionFees)) {
-      return { valid: false, reason: VerificationResultReason.INVALID_MINERS_FEE }
+    // only verify mining reward if evm gas verified
+    if (options.verifyEvm) {
```
danield9tqh, line 183: Not fully understanding the reasoning for the conditional here

hughy: Updated so that `verifyBlock` always executes the evm transactions, but it does so without affecting state.

ironfish/src/evm/evm.ts
hughy: Yeah I think that it's probably more correct for it to be undefined

PR #5222, opened 2024-08-07
update migrations descriptions, migrations list supports json

mat-if:
## Summary

image

image

image
...
image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/migrations/index.ts
```diff
@@ -1,15 +1,47 @@
...
     await this.parse(StatusCommand)
 
     const node = await this.sdk.node()
-    await node.migrator.check()
+
+    let migrationsStatus
+    try {
+      migrationsStatus = await node.migrator.status()
+    } catch (e) {
+      if (e instanceof DatabaseIsLockedError || e instanceof DatabaseOpenError) {
```
mat-if, line 26: We can use `NodeUtils.waitForOpen` here instead of a try/catch

NullSoldier (approved):
Tested this manually, it works well.

PR #5228, opened 2024-08-09
eth_sendRawTransaction RPC

jowparks:
## Summary
Send Raw Transaction that will take standard eth transactions and submit them to the network.
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/eth/sendRawTransaction.ts
+++ b/ironfish/src/rpc/routes/eth/sendRawTransaction.ts
@@ -1,0 +1,66 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { LegacyTransaction } from '@ethereumjs/tx'
+import * as yup from 'yup'
+import { Assert } from '../../../assert'
+import { GLOBAL_IF_ACCOUNT } from '../../../evm'
+import { FullNode } from '../../../node'
+import { legacyTransactionToEvmDescription } from '../../../primitives'
+import { ApiNamespace } from '../namespaces'
+import { routes } from '../router'
+
+export type SendRawTransactionRequest = {
+  transaction: string
+}
+
+export type SendRawTransactionResponse = {
+  hash: string
+  ifHash: string
+  accepted?: boolean
+  broadcasted?: boolean
+}
+
+export const SendRawTransactionRequestSchema: yup.ObjectSchema<SendRawTransactionRequest> = yup
+  .object({
+    transaction: yup.string().defined().trim(),
+  })
+  .defined()
+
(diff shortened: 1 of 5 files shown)


ironfish/src/rpc/routes/eth/sendRawTransaction.ts
```diff
@@ -0,0 +1,71 @@
...
+    const ethTransaction = LegacyTransaction.fromSerializedTx(bytes)
+    const evmDescription = legacyTransactionToEvmDescription(ethTransaction)
+    const heaviestHead = await node.wallet.getChainHead()
+    const transactionVersion = node.wallet.consensus.getActiveTransactionVersion(
+      heaviestHead.sequence,
+    )
+
+    const raw = new RawTransaction(transactionVersion)
+    // TODO(jwp): hardcoded a expiration, none exists in ethereum legacy transaction
+    raw.expiration = heaviestHead.sequence + 100
```
hughy, line 56: nit - should we use the default expirationDelta from config?

ironfish/src/rpc/routes/eth/sendRawTransaction.ts
```diff
@@ -0,0 +1,71 @@
...
+  async (request, node): Promise<void> => {
+    Assert.isInstanceOf(node, FullNode)
+
+    const bytes = Buffer.from(request.data.transaction, 'hex')
+    const ethTransaction = LegacyTransaction.fromSerializedTx(bytes)
+    const evmDescription = legacyTransactionToEvmDescription(ethTransaction)
+    const heaviestHead = await node.wallet.getChainHead()
+    const transactionVersion = node.wallet.consensus.getActiveTransactionVersion(
+      heaviestHead.sequence,
+    )
```
hughy, line 52: nit - since we have a full node for this RPC we can use the chain instead of the wallet

ironfish/src/wallet/wallet.ts
```diff
@@ -814,6 +814,7 @@ export class Wallet {
     expiration?: number
     expirationDelta?: number
     confirmations?: number
+    fund?: boolean
```
hughy, line 817: do we need the optional funding? if there aren't any outputs or burns i don't think there's anything to fund.

ironfish/src/rpc/routes/eth/sendRawTransaction.ts
```diff
@@ -0,0 +1,71 @@
...
+import { FullNode } from '../../../node'
+import { legacyTransactionToEvmDescription, RawTransaction } from '../../../primitives'
+import { ApiNamespace } from '../namespaces'
+import { routes } from '../router'
+
+export type SendRawTransactionRequest = {
+  transaction: string
+}
+
+export type SendRawTransactionResponse = {
```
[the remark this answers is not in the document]
jowparks: Lets leave it this way for testing, I would guess extra fields shouldn't matter since we are using JSON and not some length specific serde.

hughy (approved):

PR #5233, opened 2024-08-09
merge verify block logic into one entry point

patnir:
## Summary

Creates 1 entry point for verifying blocks. Addresses the comment left in our codebase over 3 years ago. 

Side note: This refactor will be helpful in optimizing performance for EVM. It seems like we verify blocks twice when we receive them, once when we get it and once when we add it to the chain. This should help consolidate that logic. 

## Testing Plan

Ensure all of our tests pass

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/consensus/verifier.ts
+++ b/ironfish/src/consensus/verifier.ts
@@ -44,7 +44,7 @@ export class Verifier {
    *  *  All transaction proofs are valid
    *  *  Miner's fee is transaction list fees + miner's reward
    */
-  async verifyBlock(
+  private async _verifyBlock(
     block: Block,
     options: { verifyTarget?: boolean } = { verifyTarget: true },
   ): Promise<VerificationResult> {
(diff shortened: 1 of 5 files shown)


ironfish/src/consensus/verifier.ts
```diff
@@ -392,22 +392,29 @@ export class Verifier {
     return header.target.targetValue === expectedTarget.targetValue
   }
 
-  // TODO: Rename to verifyBlock but merge verifyBlock into this
```
dguenther, line 395: I feel like I'd rather either:

* Remove the TODO and rename `verifyBlockAdd` to something a bit more descriptive, like `verifyBlockWithPreviousHeader` (I know that's a bit long)
* Or attempt to refactor `verifyBlock` to make `prev` non-optional. This would require a bit more testing to make sure nothing accidentally breaks, but I think we should have access to the previous block in the cases where we call verifyBlock specifically (maybe not in tests?)


The things that make me a bit unsure how I feel about this PR are:
* We now have two `verifyBlock` functions in `verifier` that aren't clearly distinguished from each other
* The structure of `verifyBlockAdd` didn't really change here -- we just added a new optional param that branches instead of calling `verifyBlockAdd` with a required param.

There's a slight functionality change where previously calling `verifyBlock` on the genesis block would actually verify all of the fields on it, but with this change, it short-circuits and returns `valid: true`. I think this isn't too important though -- we skipped verification for the genesis block in `chain.newBlock` anyway, so I don't think it should get called on the genesis block unless we were doing it in tests.

NullSoldier: I agree, I'd rather keep them separate as Derek has alluded too. By combining these, you can only make the callsite more confusing. Overall, having an optional param and merging these together does not make the code easier to understand that you're distinguishing between `Contextual Verification (blockchain state needed)` and `Contextless Verification (No blockchain state needed)`

patnir: appreciate the feedback, i'll close this for now

PR #5234, opened 2024-08-12
update miners command descriptions, add json to pool status

mat-if:
## Summary

Required some changes to the stratum client to properly end. Before this PR, the 'close' immediately kicked off the reconnect logic, so there was no actual way to close a client. This adds the ability to manually close a stratum client connection.

## Documentation

N/A

## Breaking Change

N/A

### DIFF
--- a/ironfish/src/mining/stratum/clients/client.ts
+++ b/ironfish/src/mining/stratum/clients/client.ts
@@ -37,6 +37,7 @@ export abstract class StratumClient {
   readonly version: number
 
   private started: boolean
+  private isClosing = false
   private id: number | null
   private connected: boolean
   private connectWarned: boolean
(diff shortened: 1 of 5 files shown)


ironfish/src/mining/stratum/clients/client.ts
```diff
@@ -191,11 +197,13 @@ export abstract class StratumClient {
       }
 
       this.logger.info(message)
-    } else {
+    } else if (!this.isClosing) {
       this.logger.info('Disconnected from pool unexpectedly. Reconnecting.')
```
NullSoldier, line 201: should we just move this message into the if statement below so the 2 are tied together?

ironfish/src/mining/stratum/clients/tcpClient.ts
```diff
@@ -19,7 +19,7 @@ export class StratumTcpClient extends StratumClient {
 
   protected onSocketDisconnect = (): void => {
     this.client?.off('error', this.onError)
-    this.client?.off('close', this.onDisconnect)
+    this.client?.off('close', this.onSocketDisconnect)
```
NullSoldier, line 22: Nice bug find

NullSoldier (approved):

PR #5236, opened 2024-08-12
Optimize block verification by adding skipVerification flag

patnir:
- Add skipVerification parameter to Syncer.addBlock
- Propagate skipVerification flag through Blockchain.addBlock and connect methods
- Prevent redundant block verification when flag is set
- Update relevant method signatures and calls

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/peerNetwork.ts
+++ b/ironfish/src/network/peerNetwork.ts
@@ -1242,8 +1242,8 @@ export class PeerNetwork {
       return
     }
 
-    // add the block to the chain
-    const result = await this.node.syncer.addBlock(peer, block)
+    // add the block to the chain, skipping verification
+    const result = await this.node.syncer.addBlock(peer, block, true)
 
     // We should have checked if the block is an orphan or duplicate already, so we
     // don't have to handle those cases here. If there was a verification error, the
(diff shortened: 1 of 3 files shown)


ironfish/src/network/peerNetwork.ts
```diff
@@ -1242,8 +1242,8 @@ export class PeerNetwork {
       return
     }
 
-    // add the block to the chain
-    const result = await this.node.syncer.addBlock(peer, block)
+    // add the block to the chain, skipping verification
+    const result = await this.node.syncer.addBlock(peer, block, true)
```
patnir, line 1246: verifyBlock is called above. When `syncer.addBlock` is called, then the following code is run: 

```
    const { isAdded, reason, score } = await this.chain.addBlock(block)
``` 

then `connect` is called in the Blockchain class when then calls: 

```
    const verifyBlockAdd = this.verifier.verifyBlock(block, { prev }).catch((_) => {
      return { valid: false, reason: VerificationResultReason.ERROR }
    })
```

NullSoldier:
I know this is in draft, but can we get an explanation as to the purpose of this PR?

PR #5240, opened 2024-08-13
Shield cli

jowparks:
## Summary

Requires user to input ethereum spending key during shielding, alternative is that we store as account, pass account name, retrieve account in RPC and then sign in RPC and submit transaction.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/evm/shield.ts
+++ b/ironfish-cli/src/commands/evm/shield.ts
@@ -1,0 +1,99 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { GoldTokenJson } from '@ironfish/ironfish-contracts'
+import { ContractArtifact, GLOBAL_CONTRACT_ADDRESS } from '@ironfish/sdk'
+import { Flags, ux } from '@oclif/core'
+import { ethers } from 'ethers'
+import { IronfishCommand } from '../../command'
+import { LocalFlags } from '../../flags'
+import { promptCurrency } from '../../utils/currency'
+
+export class ShieldCommand extends IronfishCommand {
+  static description = `Shield token from EVM to private asset`
+
+  static flags = {
+    ...LocalFlags,
+    amount: Flags.integer({
+      char: 'a',
+      description: 'The amount of the asset to shield',
+    }),
+    contractAddress: Flags.string({
+      char: 'c',
+      description: 'The EVM contract address of the asset to shield',
+    }),
+    nonce: Flags.integer({
+      char: 'n',
+      description: 'The nonce of the EVM transaction',
+    }),
+    to: Flags.string({
+      char: 't',
+      description: 'The Ironfish public address of the recipient',
+    }),
+    from: Flags.string({
+      char: 'f',
+      description: 'The account name of the sender',
+    }),
+  }
+
+  async start(): Promise<void> {
+    const { flags } = await this.parse(ShieldCommand)
+    const { nonce } = flags
+    let { amount, contractAddress, to, from } = flags
+    const client = await this.sdk.connectRpc()
+
+    if (!amount) {
+      amount = Number(
+        await promptCurrency({
+          client: client,
+          required: true,
+          text: 'Enter the amount of the token you want to shield',
+          minimum: 1n,
+          logger: this.logger,
+        }),
+      )
+    }
+
+    if (!to) {
+      const input = await ux.prompt('Enter recipient address: ', {
+        required: true,
+      })
+      to = input
+    }
+
+    if (!from) {
+      const defaultAccount = await client.wallet.getDefaultAccount()
+      from = defaultAccount.content.account?.name
+    }
+
+    if (!contractAddress) {
+      contractAddress = GLOBAL_CONTRACT_ADDRESS.toString()
+    }
+    const isIron = contractAddress === GLOBAL_CONTRACT_ADDRESS.toString()
+    const response = await client.wallet.getAccountPublicKey({ account: from })
+    const publicAddress = response.content.evmPublicAddress
+    if (!publicAddress) {
+      this.error(`Account ${from} does not have an EVM public address`)
+    }
+
+    let data: string
+    if (isIron) {
+      const globalContract = new ethers.Interface(ContractArtifact.abi)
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/commands/evm/shield.ts
```diff
@@ -0,0 +1,122 @@
...
+      })
+      from = input
+    }
+
+    if (!privateKey) {
+      const input = await ux.prompt('Enter private key of the sender account: ', {
+        required: true,
+      })
+      privateKey = EthUtils.prefix0x(input)
+    }
```
hughy, line 83: Do we still need this?

ironfish-cli/src/commands/evm/shield.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    if (!to) {
+      const input = await ux.prompt('Enter recipient address: ', {
+        required: true,
+      })
+      to = input
+    }
+
+    if (!from) {
```
hughy, line 71: should we default to the default account if one isn't set?

ironfish-cli/src/commands/evm/shield.ts
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    if (!privateKey) {
+      const input = await ux.prompt('Enter private key of the sender account: ', {
+        required: true,
+      })
+      privateKey = EthUtils.prefix0x(input)
+    }
+
+    if (!contractAddress) {
```
hughy, line 85: should we default to the global contract address?

ironfish-cli/src/commands/evm/shield.ts
```diff
@@ -0,0 +1,122 @@
...
+    if (!publicAddress) {
+      this.error(`Account ${from} does not have an EVM public address`)
+    }
+
+    let data: string
+    if (isIron) {
+      const globalContract = new ethers.Interface(ContractArtifact.abi)
+      data = globalContract.encodeFunctionData('shield_iron', [Buffer.from(to, 'hex')])
+    } else {
+      const contract = new ethers.Interface(GoldTokenJson.abi)
```
hughy, line 106: we don't need to address this now, but I'm wondering if we could use a generic ERC20 abi here?

ironfish-cli/src/commands/evm/shield.ts
```diff
@@ -0,0 +1,122 @@
...
+    const client = await this.sdk.connectRpc()
+
+    if (!amount) {
+      const input = await ux.prompt('Enter amount to send: ', {
+        required: true,
+      })
+      amount = parseInt(input)
+    }
+
+    if (!nonce) {
```
hughy, line 50: I think this can be optional since `eth/sendTransaction` will use the nonce of the account by default

ironfish-cli/src/commands/evm/shield.ts
```diff
@@ -0,0 +1,122 @@
...
+      const input = await ux.prompt('Enter nonce for EVM transaction: ', {
+        required: true,
+      })
+      nonce = parseInt(input)
+    }
+
+    if (!fee) {
```
hughy, line 57: doesn't look like we use the fee

ironfish-cli/src/commands/evm/shield.ts
```diff
@@ -0,0 +1,122 @@
...
+    }),
+  }
+
+  async start(): Promise<void> {
+    const { flags } = await this.parse(ShieldCommand)
+    let { amount, contractAddress, nonce, fee, to, privateKey, from } = flags
+    const client = await this.sdk.connectRpc()
+
+    if (!amount) {
+      const input = await ux.prompt('Enter amount to send: ', {
```
hughy, line 44: can we use `promptCurrency` here? that will make it so that the amount prompt works the same way as the send command with decimal input

ironfish-cli/src/commands/evm/shield.ts
```diff
@@ -0,0 +1,122 @@
...
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { GoldTokenJson } from '@ironfish/ironfish-contracts'
+import { ContractArtifact, EthUtils, GLOBAL_CONTRACT_ADDRESS } from '@ironfish/sdk'
+import { Flags, ux } from '@oclif/core'
+import { ethers } from 'ethers'
+import { IronfishCommand } from '../../command'
+import { LocalFlags } from '../../flags'
+
+export class ShieldCommand extends IronfishCommand {
+  static description = `Shield asset from EVM to private token`
```
patnir, line 12:
```suggestion
  static description = `Shield token from EVM to private asset`
```

ironfish-cli/src/commands/evm/shield.ts
jowparks: I wasn't sure how to do this, that is why I created an example contract. As long as any ERC20 uses the same shield interface, the encoded data should be fine since the function signature is the same. But yea I agree long term that has to be something that is slightly cleaner.

ironfish-cli/src/commands/evm/shield.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export class ShieldCommand extends IronfishCommand {
+  static description = `Shield token from EVM to private asset`
+
+  static flags = {
+    ...LocalFlags,
+    amount: Flags.integer({
+      char: 'a',
+      description: 'The amount of the asset to shield',
+    }),
```
patnir, line 20:
```suggestion
    amount: ValueFlag({
      char: 'a',
      description: 'The amount to send in the major denomination',
      flagName: 'amount',
    }),
```

hughy (approved):

ironfish-cli/src/commands/evm/shield.ts
```diff
@@ -0,0 +1,100 @@
...
+      data = globalContract.encodeFunctionData('shield_iron', [Buffer.from(to, 'hex')])
+    } else {
+      const contract = new ethers.Interface(GoldTokenJson.abi)
+      data = contract.encodeFunctionData('shield', [Buffer.from(to, 'hex'), amount])
+    }
+    // TODO: should we unhardcode gas limit and gas price?
+    const txDetails = {
+      to: contractAddress,
+      from: publicAddress,
+      value: String(amount),
```
patnir, line 91: amount is only valid for $IRON right? I think in the shield token case we are not calling payable function

jowparks: Yea you are right, good call!

danield9tqh (approved):

PR #5268, opened 2024-08-14
eth/sendTransaction RPC (WIP)

danield9tqh:
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/eth/sendTransaction.ts
+++ b/ironfish/src/rpc/routes/eth/sendTransaction.ts
@@ -2,8 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { LegacyTransaction } from '@ethereumjs/tx'
-import { Account, Address } from '@ethereumjs/util'
-import { Wallet } from 'ethers/wallet'
+import { Address } from '@ethereumjs/util'
 import * as yup from 'yup'
 import { Assert } from '../../../assert'
 import { GLOBAL_IF_ACCOUNT } from '../../../evm'
--- a/ironfish/src/rpc/routes/eth/sendTransaction.ts
+++ b/ironfish/src/rpc/routes/eth/sendTransaction.ts
@@ -1,0 +1,110 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { LegacyTransaction } from '@ethereumjs/tx'
+import { Account, Address } from '@ethereumjs/util'
+import { Wallet } from 'ethers/wallet'
+import * as yup from 'yup'
+import { Assert } from '../../../assert'
+import { GLOBAL_IF_ACCOUNT } from '../../../evm'
+import { FullNode } from '../../../node'
+import { legacyTransactionToEvmDescription } from '../../../primitives'
+import { EthUtils } from '../../../utils/eth'
+import { AssertSpending } from '../../../wallet/account/account'
+import { ApiNamespace } from '../namespaces'
+import { routes } from '../router'
+
+export type EthSendTransactionRequest = {
+  from: string
+  to?: string
+  gas?: string
+  gasPrice?: string
+  value?: string
+  data?: string
+  nonce?: string
+}
+
+export type EthSendTransactionResponse = {
+  result: string
+}
+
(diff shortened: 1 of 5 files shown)


ironfish/src/rpc/routes/eth/sendTransaction.ts
```diff
@@ -0,0 +1,107 @@
...
+import { ApiNamespace } from '../namespaces'
+import { routes } from '../router'
+
+export type EthSendTransactionRequest = {
+  from: string
+  to?: string
+  gas?: string
+  gasPrice?: string
+  value?: string
+  data: string
```
jowparks, line 22: Does data have to be defined? I thought you could send nil for data

ironfish/src/rpc/routes/eth/sendTransaction.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    const ethAccount = account.ethAddress
+      ? await node.chain.evm.getAccount(Address.fromString(account.ethAddress))
+      : undefined
+
+    const nonce = request.data.nonce
+      ? BigInt(request.data.nonce)
+      : ethAccount?.nonce ?? BigInt(0)
+
+    const gas = request.data.gas ? BigInt(request.data.gas) : undefined
```
jowparks, line 69: if you set these as undefined, will the transaction still succeed? I thought the default values won't work.

ironfish/src/rpc/routes/eth/sendTransaction.ts
```diff
@@ -0,0 +1,107 @@
...
+    Assert.isInstanceOf(node, FullNode)
+
+    const account = node.wallet
+      .listAccounts()
+      .find((a) => a.ethAddress === EthUtils.remove0x(request.data.from))
+    Assert.isNotUndefined(account, 'Account not found')
+    AssertSpending(account)
+
+    const ethAccount = account.ethAddress
+      ? await node.chain.evm.getAccount(Address.fromString(account.ethAddress))
```
jowparks, line 62:
```suggestion
      ? await node.chain.evm.getAccount(Address.fromString(EthUtils.prefix0x(account.ethAddress)))
```

jowparks (approved):
LGTM, up to you if you want to add test before shipping or yolo in evm branch

ironfish/src/rpc/routes/eth/sendTransaction.ts
danield9tqh: yea I'll check this

ironfish/src/rpc/routes/eth/sendTransaction.ts
danield9tqh: made it optional

ironfish/src/rpc/routes/eth/sendTransaction.ts
```diff
@@ -0,0 +1,110 @@
...
+    const ethTransaction = new LegacyTransaction({
+      nonce: nonce,
+      to: request.data.to,
+      gasLimit: gas,
+      gasPrice: gasPrice,
+      value: value,
+      data: request.data.data,
+    })
+
+    const signed = ethTransaction.sign(Buffer.from(account.spendingKey, 'hex'))
```
hughy, line 83: Do we only need to sign here so that `simulateTx` can run? I think it's possible to bypass signature verification when running transactions

ironfish/src/rpc/routes/eth/sendTransaction.ts
```diff
@@ -0,0 +1,110 @@
...
+    Assert.isNotUndefined(events)
+
+    const raw = await node.wallet.createEvmTransaction({
+      evm: evmDescription,
+      evmEvents: events,
+      account: account,
+    })
+
+    // TODO: This is pretty hacky to figure out which key to post with
+    const unshields = events.filter((e) => e.name === 'unshield')
```
hughy, line 97: Do you think it's less hacky to check if the transaction has any spends instead?

hughy (approved):

ironfish/src/rpc/routes/eth/sendTransaction.ts
danield9tqh: we prob need to sign before we post it too

ironfish/src/rpc/routes/eth/sendTransaction.ts
jowparks: We only need to sign with global account because of the mints right? Could be signed with personal key if we use the balancing approach maybe? Thoughts for later

danield9tqh: Yea checking spends might be slightly less hacky but I think I'd still keep the comment there bc we want to figure out a better way in the future

PR #5276, opened 2024-08-14
Add `wallet:transactions:delete` command

mat-if:
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/wallet/transactions/delete.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  static args = {
+    transaction: Args.string({
+      required: true,
+      description: 'Hash of the transaction to delete from the wallet',
+    }),
+  }
+
+  static flags = {
+    ...RemoteFlags,
```
NullSoldier, line 20: nit: we should not use RemoteFlags for this command since it's local only at the moment. We should convert it to use an RPC later, though.

NullSoldier (approved):

PR #5281, opened 2024-08-15
feat(ironfish): Add unlock to wallet

rohanjadvani:
## Summary

* Adds an unlock method to the wallet which:
  * Does nothing if the wallet is decrypted
  * Decrypts all encrypted accounts in memory
  * Sets `locked` to false and starts a timeout to automatically lock the wallet
  * If the provided `timeout` is `-1`, leave the wallet unlocked indefinitely
  * Use a default timeout of 5 minutes if no argument is passed
* Clears any lock timeout when `lock` is called
* Clears the lock timeout when stopping the wallet 

## Testing Plan

Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -1844,7 +1844,7 @@ export class Wallet {
       this.locked = false
     } catch (e) {
       this.logger.debug('Wallet unlock failed')
-      await this.lock()
+      this.accountById.clear()
 
       throw e
     } finally {
--- a/ironfish/src/wallet/wallet.ts
+++ b/ironfish/src/wallet/wallet.ts
@@ -1844,7 +1844,9 @@ export class Wallet {
       this.locked = false
     } catch (e) {
       this.logger.debug('Wallet unlock failed')
+      this.stopUnlockTimeout()
       this.accountById.clear()
+      this.locked = true
 
       throw e
     } finally {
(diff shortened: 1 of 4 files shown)


ironfish/src/wallet/wallet.ts
```diff
@@ -1813,10 +1819,56 @@ export class Wallet {
...
+      if (!encrypted) {
+        return
+      }
+
+      for (const [id, account] of this.encryptedAccountById.entries()) {
+        this.accountById.set(id, account.decrypt(passphrase))
+      }
+
+      this.startUnlockTimeout(timeout)
+      this.locked = false
```
mat-if, line 1844: Might be a bit of an overkill defensive thought but, does this get stuck in a weird state if for some unknown/bug reason one of the decrypts fails, or perhaps even a failure to read the encrypted accounts from disk? 

Thinking about it a little, should the `catch` try to re-lock things up or recover in some manner? Or maybe start the unlock timeout in the finally?

mat-if (approved):
1 comment that can be a future PR if anything actionable comes from it, but looks good!

ironfish/src/wallet/wallet.ts
rohanjadvani: I don't think reading encrypted accounts from disk should apply since this is using cached values. But if at least one decryption fails, that's true this could end up in a weird state where some subset of accounts would be decrypted.

I don't think starting the timeout should happen in the `finally` since that will always execute, and the re-lock should be dependent on all decryptions being successful. But I think re-locking in the `catch` is a good idea - that will guard if the first decryptions worked but subsequent ones did not.

PR #5285, opened 2024-08-15
Add EVM unshield command

patnir:
## Summary

Implements a new CLI command to unshield private assets to public tokens on the EVM chain. Features include:
- Support for both IRON and custom ERC20 tokens
- Uses the Gold Token as an example interface for ERC20 contracts
- User prompts for missing required information
- Validation of Ethereum addresses and node sync status
- Automatic use of default account if not specified

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/evm/unshield.ts
+++ b/ironfish-cli/src/commands/evm/unshield.ts
@@ -1,0 +1,135 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { isValidAddress } from '@ethereumjs/util'
+import { GoldTokenJson } from '@ironfish/ironfish-contracts'
+import {
+  ContractArtifact,
+  EthSendTransactionRequest,
+  GLOBAL_CONTRACT_ADDRESS,
+} from '@ironfish/sdk'
+import { Flags, ux } from '@oclif/core'
+import { ethers } from 'ethers'
+import { IronfishCommand } from '../../command'
+import { LocalFlags } from '../../flags'
+import { promptCurrency } from '../../utils/currency'
+
+export class UnshieldCommand extends IronfishCommand {
+  static description = `Unshield private asset to public token`
+
+  static flags = {
+    ...LocalFlags,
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/evm/unshield.ts
```diff
@@ -0,0 +1,138 @@
...
+import { promptCurrency } from '../../utils/currency'
+
+export class UnshieldCommand extends IronfishCommand {
+  static description = `Unshield private asset to public token`
+
+  static flags = {
+    ...LocalFlags,
+    amount: Flags.integer({
+      char: 'a',
+      description: 'The amount to send in the major denomination',
```
jowparks, line 25: Are we sending in major or minor?

ironfish-cli/src/commands/evm/unshield.ts
```diff
@@ -0,0 +1,138 @@
...
+      from: publicAddress,
+      value: '0',
+      gas: String(1000000),
+      gasPrice: String(0),
+      data,
+    }
+
+    if (nonce) {
+      txDetails.nonce = String(nonce)
+    }
```
jowparks, line 133:
```suggestion
    const txDetails: EthSendTransactionRequest = {
      nonce: nonce ? String(nonce) : undefined,
      to: contractAddress,
      from: publicAddress,
      value: '0',
      gas: String(1000000),
      gasPrice: String(0),
      data,
    }
```

ironfish-cli/src/commands/evm/unshield.ts
hughy: I think it will be minor -- I think the global contract treats the values in minor denomination

ironfish-cli/src/commands/evm/unshield.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    if (!to) {
+      to = await ux.prompt('Enter the public address of the recipient', {
+        required: true,
+      })
+    }
+
+    if (!isValidAddress(to)) {
+      this.error('Invalid Ethereum address')
+    }
```
hughy, line 83: later on we could integrate this into the prompt for the address

hughy (approved):

jowparks (approved):

PR #5291, opened 2024-08-15
add ability to send evm transaction that has no events

jowparks:
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/eth/sendTransaction.ts
```diff
@@ -84,7 +84,6 @@ routes.register<typeof EthSendTransactionRequestSchema, EthSendTransactionRespon
     const evmDescription = legacyTransactionToEvmDescription(signed)
     const result = await node.chain.evm.simulateTx({ tx: signed })
     const events = result.events
-    Assert.isNotUndefined(events)
```
hughy, line 87: it might still be worthwhile to assert here that `result.error` is undefined

hughy (approved):

PR #5292, opened 2024-08-15
adds cli command for unshielded evm balances

hughy:
## Summary

uses the 'eth/getAccount' RPC to fetch account state at the current chain head and the latest confirmed block

displays unconfirmed and confirmed balances

only supports the native asset for now

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/evm/balance.ts
+++ b/ironfish-cli/src/commands/evm/balance.ts
@@ -1,0 +1,61 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { isValidAddress } from '@ethereumjs/util'
+import { Flags } from '@oclif/core'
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/commands/evm/balance.ts
```diff
@@ -0,0 +1,61 @@
...
+    confirmations: Flags.integer({
+      required: false,
+      description: 'Minimum number of blocks confirmations for a transaction',
+    }),
+    // TODO(hughy): add support for custom tokens
+  }
+
+  async start(): Promise<void> {
+    const { flags } = await this.parse(EvmBalanceCommand)
+
```
jowparks, line 28: Could use default account if not provided

jowparks (approved):

ironfish-cli/src/commands/evm/balance.ts
hughy: I wasn't sure whether to use accounts or addresses for this command. I went with addresses so that we'd be able to use it to check balances for accounts that aren't in our wallets

PR #5317, opened 2024-08-20
eth rpc wrapper

jowparks:
## Summary
For registering RPCs for integration with external systems, but maintain our own api specification, I created this wrapper. It will make the endpoints registered available via the standardized JSON rpc system of eth nodes, but also with our typesafe client.

For example, an endpoint can hit:

`rpc/ethRouter` with request body:
```
{
  jsonrpc: '2.0'
  id: 1
  method: 'eth_sendRawTransaction`
  params: [{transaction: '0x....']
}
```

Which is the same as using our client at
`client.eth.sendRawTransaction({transaction: '0x....'})`
or
`rpc/eth/sendTransaction`

## Testing Plan
Test Passes.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/eth/ethRouter.test.ts
+++ b/ironfish/src/rpc/routes/eth/ethRouter.test.ts
@@ -1,0 +1,36 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Address } from '@ethereumjs/util'
+import { useAccountFixture } from '../../../testUtilities'
+import { createRouteTest } from '../../../testUtilities/routeTest'
+
+describe('Route eth/ethRouter', () => {
+  const routeTest = createRouteTest(false)
+
+  it('should handle eth_sendTransaction successfully', async () => {
+    const { node } = routeTest
+    const account = await useAccountFixture(node.wallet, 'test')
+    const address = Address.fromPrivateKey(Buffer.from(account.spendingKey, 'hex'))
+
+    const transactionRequest = {
+      jsonrpc: '2.0' as const,
+      id: 1,
+      method: 'eth_sendTransaction' as const,
+      params: [
+        {
+          from: address.toString(),
+          to: address.toString(),
+          gas: '0x5208',
+          gasPrice: '0x0',
+          value: '0x0',
+          data: '0x',
+        },
+      ],
+    }
+
+    const response = await routeTest.client.eth.ethRouter(transactionRequest)
+
+    expect(response.status).toEqual(200)
+  })
+})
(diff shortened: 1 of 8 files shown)


ironfish/src/rpc/routes/eth/ethRouter.ts
```diff
@@ -0,0 +1,88 @@
...
+        request.data.params[0],
+        '',
+        request.onEnd as (status: number, data?: unknown) => void,
+        request.onStream as (data: unknown) => void,
+      )
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-explicit-any
+      await handlerEntry.handler(req as unknown as any, node)
+    } else {
+      throw new Error(`Unsupported method: ${request.data.method}`)
+    }
```
patnir, line 86: Here is the spec for returning errors: 

https://www.quicknode.com/docs/ethereum/error-references

This can be done in a separate PR, but I think the expectation is a similar object to the one that we send for successful queries.

patnir: Maybe worth leaving a comment here while it is not implemented

ironfish/src/rpc/routes/eth/ethRouter.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+} = {}
+
+export function registerEthRoute<TRequestSchema extends YupSchema, TResponse>(
+  ethRoute: string,
+  route: string,
+  requestSchema: TRequestSchema,
+  handler: RouteHandler<YupSchemaResult<TRequestSchema>, TResponse>,
+): void {
+  ethRoutes[ethRoute] = {
```
patnir, line 45: very cool

patnir:
Are the responses from this API in this format: 

```
{
    "jsonrpc": "2.0",
    "result": "0x13a1d15",
    "id": 0
}
```

jowparks:
> Are the responses from this API in this format:
> 
> ```
> {
>     "jsonrpc": "2.0",
>     "result": "0x13a1d15",
>     "id": 0
> }
> ```

They do not have the jsonrpc/id in the response, we could add that though

ironfish/src/rpc/routes/eth/ethRouter.test.ts
```diff
@@ -0,0 +1,36 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Address } from '@ethereumjs/util'
+import { useAccountFixture } from '../../../testUtilities'
+import { createRouteTest } from '../../../testUtilities/routeTest'
+
+describe('Route eth/ethRouter', () => {
+  const routeTest = createRouteTest(false)
+
+  it('should handle eth_sendTransaction successfully', async () => {
```
patnir, line 11: maybe also worth doing the eth_getAccount test and asserting the format of the response?

patnir:
eth spec seems to allow an empty params array, but it doesn't work currently. also i might be because i messed something up when implementing blockNumber?

image

works with an empty object in it: 

image

jowparks:
> eth spec seems to allow an empty params array, but it doesn't work currently. also i might be because i messed something up when implementing blockNumber?
> 
> image
> 
> works with an empty object in it:
> 
> image

Are you using the new register method for `blockNumber`?

danield9tqh (approved):

PR #5325, opened 2024-08-21
Implement eth_blockNumber RPC method

patnir:
## Summary

- Add blockNumber route to return current block number in hexadecimal
- Include tests

Spec: https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_blocknumber

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/eth/blockNumber.ts
+++ b/ironfish/src/rpc/routes/eth/blockNumber.ts
@@ -1,0 +1,39 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import * as yup from 'yup'
+import { Assert } from '../../../assert'
+import { FullNode } from '../../../node'
+import { ApiNamespace } from '../namespaces'
+import { routes } from '../router'
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+export type BlockNumberRequest = {} | undefined
+
+export const BlockNumberRequestSchema: yup.ObjectSchema<BlockNumberRequest> = yup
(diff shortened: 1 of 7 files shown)


ironfish/src/rpc/routes/eth/blockNumber.ts
```diff
@@ -0,0 +1,39 @@
...
+export type BlockNumberResponse = {
+  number: string
+}
+
+export const BlockNumberResponseSchema: yup.ObjectSchema<BlockNumberResponse> = yup
+  .object({
+    number: yup.string().defined(),
+  })
+  .defined()
+
```
jowparks, line 28: Should use the `registerEthRoute` method.

ironfish/src/rpc/routes/eth/blockNumber.ts
```diff
@@ -0,0 +1,39 @@
...
+import { Assert } from '../../../assert'
+import { FullNode } from '../../../node'
+import { ApiNamespace } from '../namespaces'
+import { routes } from '../router'
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+export type BlockNumberRequest = {} | undefined
+
+export const BlockNumberRequestSchema: yup.ObjectSchema<BlockNumberRequest> = yup
+  .object({})
```
jowparks, line 15: Will this ever take a parameter? Should never need one right? Maybe something like:
`yup.MixedSchema<void>`

jowparks: https://github.com/iron-fish/ironfish/blob/3c70456b30903f21bfdfe0c49d4e5f3d96fe9800/ironfish/src/rpc/routes/chain/getChainInfo.ts#L24-L26

jowparks (approved):

PR #5332, opened 2024-08-26
eth_getBalance RPC

jowparks:
## Summary
`eth` RPC can accept other statuses, but not sure they are well defined in our chain:
https://www.quicknode.com/docs/ethereum/eth_getBalance
`The block number in hexadecimal format or the string latest, earliest, pending, safe or finalized (safe and finalized tags are only supported on Ethereum, Gnosis, Arbitrum, Arbitrum Nova and Avalanche C-chain), see the default block parameter description in the official Ethereum documentation`
## Testing Plan
tests pass
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/eth/getBalance.ts
+++ b/ironfish/src/rpc/routes/eth/getBalance.ts
@@ -1,0 +1,60 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Address } from '@ethereumjs/util'
+import * as yup from 'yup'
+import { Assert } from '../../../assert'
+import { FullNode } from '../../../node'
+import { BlockHeader } from '../../../primitives'
+import { RpcNotFoundError } from '../../adapters'
+import { ApiNamespace } from '../namespaces'
+import { registerEthRoute } from './ethRouter'
+
+export type EthGetBalanceRequest = {
+  address: string
+  blockNumber: string
+}
+
+export type EthGetBalanceResponse = string
+
+export const EthGetBalanceRequestSchema: yup.ObjectSchema<EthGetBalanceRequest> = yup
+  .object({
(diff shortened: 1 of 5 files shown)


ironfish/src/rpc/routes/eth/getBalance.ts
```diff
@@ -0,0 +1,60 @@
...
+import { Assert } from '../../../assert'
+import { FullNode } from '../../../node'
+import { BlockHeader } from '../../../primitives'
+import { RpcNotFoundError } from '../../adapters'
+import { ApiNamespace } from '../namespaces'
+import { registerEthRoute } from './ethRouter'
+
+export type EthGetBalanceRequest = {
+  address: string
+  blockNumber: string
```
hughy, line 15: I was thinking it was a bit odd to use 'Number' since this might be a tag like 'latest', but I see that some other providers (e.g., QuickNode) call it `blockNumber`

ironfish/src/rpc/routes/eth/getBalance.ts
```diff
@@ -0,0 +1,60 @@
...
+  `${ApiNamespace.eth}/getBalance`,
+  EthGetBalanceRequestSchema,
+  async (request, node): Promise<void> => {
+    Assert.isInstanceOf(node, FullNode)
+
+    const address = Address.fromString(request.data.address)
+    // TODO: latest, earliest, pending, safe or finalized in other chains, stubbing for now
+    let header: BlockHeader | null
+    if (request.data.blockNumber === 'latest') {
+      header = await node.chain.getHeader(node.chain.latest.hash)
```
hughy, line 42: nit - `node.chain.latest` is already a header

hughy (approved):

ironfish/src/rpc/routes/eth/getBalance.ts
jowparks: I agree, I am just following the eth spec, it is indeed a hex block sequence number I believe

danield9tqh (approved):

PR #5335, opened 2024-08-26
adds `eth_getTransactionByHash` rpc

jowparks:
## Summary
Returns eth transaction info based on eth transaction hash
Uses eth transaction hash -> utxo hash mapping in blockchain
## Testing Plan
tests pass
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/eth/getTransactionByHash.ts
```diff
@@ -0,0 +1,143 @@
...
+      throw new RpcNotFoundError(`Transaction ${request.data} not found`)
+    }
+    if (!transaction.transaction.evm) {
+      throw new RpcNotFoundError(`Transaction ${request.data} does not have EVM description`)
+    }
+    const ethTransaction = evmDescriptionToLegacyTransaction(transaction.transaction.evm)
+    // TODO deal with items that are mocked
+    request.end({
+      blockHash: EthUtils.prefix0x(transaction.blockHash.toString('hex')),
+      blockNumber: EthUtils.prefix0x(transaction.sequence.toString(16)),
```
danield9tqh, line 122: Could prob use a number / bigint to hex string util since we're using this a lot

danield9tqh (approved):

ironfish/src/rpc/routes/eth/getTransactionByHash.ts
```diff
@@ -0,0 +1,143 @@
...
+      maxFeePerGas: '0x',
+      maxPriorityFeePerGas: '0x',
+      hash: EthUtils.prefix0x(Buffer.from(ethTransaction.hash()).toString('hex')),
+      input: EthUtils.prefix0x(Buffer.from(ethTransaction.data).toString('hex')),
+      nonce: EthUtils.prefix0x(ethTransaction.nonce.toString(16)),
+      to: ethTransaction.to === undefined ? null : ethTransaction.to.toString(),
+      value: EthUtils.prefix0x(ethTransaction.value.toString(16)),
+      type: EthUtils.prefix0x(ethTransaction.type.toString(16)),
+      accessList: [],
+      chainId: '0x42069',
```
hughy, line 136: 😂 

at some point we might need to figure out what to use for the chainId. I think it should ideally be a parameter to signing, too

ironfish/src/rpc/routes/eth/getTransactionByHash.ts
```diff
@@ -0,0 +1,143 @@
...
+      | {
+          transaction: Transaction
+          initialNoteIndex: number
+          sequence: number
+          blockHash: Buffer
+          previousBlockHash: Buffer
+          timestamp: Date
+        }
+      | undefined = undefined
+    let transactionNum = 0
```
hughy, line 104: I think we'll end up creating a util for this because there are multiple eth endpoints that return `transactionIndex`

hughy (approved):

ironfish/src/rpc/routes/eth/getTransactionByHash.ts
jowparks: Added

PR #5337, opened 2024-08-27
adds db store for evm receipts

hughy:
## Summary

stores most of the same fields as geth, but excludes fields that we can obtain from the transaction or block and excludes blob-related fields because we haven't addressed blobs

defines a separate interface for the values stored in the db store since the RunTxResult has many nested fields

updates the store in tandem with the ethTransactionHashToTransactionHash store

TODO: we don't accurately track cumulativeGasUsed right yet

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/blockchain/database/blockchaindb.ts
+++ b/ironfish/src/blockchain/database/blockchaindb.ts
@@ -25,6 +25,7 @@ import {
   AssetToContractTokenSchema,
   ContractTokenToAssetSchema,
   ethTransactionHashToTransactionHashSchema,
+  EvmReceiptSchema,
   HashToNextSchema,
   HeadersSchema,
   MetaSchema,
@@ -34,6 +35,7 @@ import {
   TransactionsSchema,
 } from '../schema'
 import { AssetValue, AssetValueEncoding } from './assetValue'
+import { EvmReceiptValue, EvmReceiptValueEncoding } from './evmReceiptValue'
 import { HeaderEncoding, HeaderValue } from './headers'
 import { SequenceToHashesValueEncoding } from './sequenceToHashes'
 import { TransactionsValue, TransactionsValueEncoding } from './transactions'
@@ -67,6 +69,8 @@ export class BlockchainDB extends TransactionalDatabase {
   transactionHashToBlockHash: IDatabaseStore<TransactionHashToBlockHashSchema>
   // EthTransactionHash -> TransactionHash
   ethTransactionHashToTransactionHash: IDatabaseStore<ethTransactionHashToTransactionHashSchema>
+  // EthTransactionHash -> Transaction Receipt
+  evmTransactionReceipts: IDatabaseStore<EvmReceiptSchema>
 
   stateManager: IronfishStateManager
 
@@ -145,6 +149,12 @@ export class BlockchainDB extends TransactionalDatabase {
       name: 'ett',
       keyEncoding: BUFFER_ENCODING,
       valueEncoding: BUFFER_ENCODING,
+    })
+
+    this.evmTransactionReceipts = this.db.addStore({
+      name: 'er',
+      keyEncoding: BUFFER_ENCODING,
+      valueEncoding: new EvmReceiptValueEncoding(),
     })
 
     this.stateManager = new IronfishStateManager(this.db)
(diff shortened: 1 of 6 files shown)


ironfish/src/blockchain/database/blockchaindb.ts
```diff
@@ -422,6 +432,18 @@ export class BlockchainDB extends TransactionalDatabase {
...
   }
 
+  async putEvmReceipt(
+    ethTransactionHash: Buffer,
+    receipt: EvmReceiptValue,
+    tx?: IDatabaseTransaction,
+  ): Promise<void> {
+    return this.evmTransactionReceipts.put(ethTransactionHash, receipt, tx)
+  }
+
```
jowparks, line 442: Gonna need a get field, might as well add it here

ironfish/src/blockchain/database/evmReceiptValue.ts
```diff
@@ -0,0 +1,141 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import type { IDatabaseEncoding } from '../../storage/database/types'
+import { Log } from '@ethereumjs/evm'
+import { RunTxResult } from '@ethereumjs/vm'
+import bufio from 'bufio'
+
```
jowparks, line 8: nit: why not just call it `EvmReceipt`?

ironfish/src/blockchain/database/evmReceiptValue.ts
```diff
@@ -0,0 +1,141 @@
...
+  type: number
+  status: number
+  contractAddress?: Buffer
+  logsBloom: Buffer
+  logs: Log[]
+  gasUsed: bigint
+  cumulativeGasUsed: bigint
+
+  // fields below are included in Geth serialization, but we may not need them
+
```
jowparks, line 19: Can we derive `effectiveGasPrice` from somewhere else? Seems like it is worth storing even if it will always be 0 for now.

ironfish/src/blockchain/database/evmReceiptValue.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    const tx = new LegacyTransaction({
+      nonce: 0n,
+      to: GLOBAL_CONTRACT_ADDRESS,
+      gasLimit: 1000000n,
+      gasPrice: 0n,
+      value: 500n,
+      data: encodedFunctionData,
+    })
+
+    const signed = tx.sign(Buffer.from(evmAccount.privateKey.replace(/0x/g, ''), 'hex'))
```
jowparks, line 49: nit: EthUtils.remove0x

jowparks (approved):
LGTM

ironfish/src/blockchain/database/evmReceiptValue.ts
hughy: There's a derivation in the ethereumjs implementation of `eth_getTransactionReceipt`: 
```
    const effectiveGasPrice = tx.supports(Capability.EIP1559FeeMarket)
      ? (tx as FeeMarketEIP1559Transaction).maxPriorityFeePerGas <
        (tx as FeeMarketEIP1559Transaction).maxFeePerGas - block.header.baseFeePerGas!
        ? (tx as FeeMarketEIP1559Transaction).maxPriorityFeePerGas
        : (tx as FeeMarketEIP1559Transaction).maxFeePerGas -
          block.header.baseFeePerGas! +
          block.header.baseFeePerGas!
      : (tx as LegacyTransaction).gasPrice
 ```
 
 I think for now it should always be `tx.gasPrice` since we're using legacy transactions, but it looks like in general we can derive it from the transaction and block.

ironfish/src/blockchain/database/evmReceiptValue.ts
hughy: I was basing this off of other `*Value` interfaces that we use in databases (`AccountValue`, `AssetValue`, etc.). I think if we end up using this interface outside of just storing data in the database then we should drop the `Value`.

ironfish/src/blockchain/database/blockchaindb.ts
hughy: Added!

PR #5338, opened 2024-08-27
eth_getBlockByNumber

jowparks:
## Summary
Gets block by hex number, ie block two is `0x2`. If second param is true, will return serialized transactions. If false, just returns hashes.
https://www.quicknode.com/docs/ethereum/eth_getBlockByNumber
## Testing Plan
tests pass
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/eth/getBlockByNumber.ts
```diff
@@ -0,0 +1,131 @@
...
+      gasLimit: '0x0',
+      gasUsed: '0x0',
+      hash: EthUtils.prefix0x(blockHeader.hash.toString('hex')),
+      logsBloom: '0x',
+      miner: '0x',
+      mixHash: '0x',
+      nonce: '0x',
+      number: EthUtils.prefix0x(blockHeader.sequence.toString(16)),
+      parentHash: EthUtils.prefix0x(blockHeader.previousBlockHash.toString('hex')),
+      receiptsRoot: '0x',
```
jowparks, line 114: @hughy is this root available or the `transactionsRoot`? I think not, just verifying.

hughy: receipts root isn't available right now, but we could create the tree and root for receipts on a block if we need to. I think `transactionsRoot` should be the same as `transactionCommitment` from the block header

ironfish/src/utils/eth.ts
```diff
@@ -14,4 +14,8 @@ function numToHex(num: number | bigint): string {
   return prefix0x(num.toString(16))
 }
 
-export const EthUtils = { prefix0x, remove0x, numToHex }
+function bytesToHex(bytes: Uint8Array): string {
```
hughy, line 17: Does the `bytesToHex` from the ethereumjs repo work here?

hughy (approved):

PR #5346, opened 2024-08-28
adds eth_getTransactionReceipt RPC

hughy:
## Summary

see provider RPC docs (e.g.,
https://www.quicknode.com/docs/ethereum/eth_getTransactionReceipt)

defines EthRcpLog type and util for generating from transaction, block header, and receipt

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/rpc/routes/eth/getTransactionReceipt.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    await routeTest.node.wallet.scan()
+
+    const ethHash = evmDescriptionToLegacyTransaction(transaction.evm!).hash()
+    const result = await routeTest.client.eth.getTransactionReceipt(
+      Buffer.from(ethHash).toString('hex'),
+    )
+
+    expect(result.status).toEqual(200)
+    expect(result.content).toMatchObject({
+      transactionHash: EthUtils.prefix0x(Buffer.from(ethHash).toString('hex')),
```
jowparks, line 58: I have been doing this one a lot too, we should prob have a util for it, we can do in diff PR though.

ironfish/src/rpc/routes/eth/getTransactionReceipt.test.ts
```diff
@@ -0,0 +1,74 @@
...
+      transactionIndex: '0x1',
+      blockHash: EthUtils.prefix0x(block1.header.hash.toString('hex')),
+      blockNumber: '0x2',
+      from: EthUtils.prefix0x(evmSenderAddress.toString()),
+      to: EthUtils.prefix0x('a'.repeat(40)),
+      cumulativeGasUsed: '0x5208',
+      effectiveGasPrice: '0x0',
+      gasUsed: '0x0',
+      contractAddress: null,
+      logs: [],
```
jowparks, line 68: Do we expect no logs here? If not then can we update test to have receipt that does return logs?

jowparks (approved):
Just would like to see logs tested, otherwise looks good!

ironfish/src/rpc/routes/eth/getTransactionReceipt.test.ts
hughy: I think native asset transfers don't emit any logs. I'll update this test to use an unshield transaction so that we can test with logs.

PR #5348, opened 2024-08-29
eth_accounts

jowparks:
## Summary
https://www.quicknode.com/docs/ethereum/eth_getTransactionReceipt
## Testing Plan
tests pass
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/eth/accounts.ts
+++ b/ironfish/src/rpc/routes/eth/accounts.ts
@@ -1,0 +1,38 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import * as yup from 'yup'
+import { Assert } from '../../../assert'
+import { FullNode } from '../../../node'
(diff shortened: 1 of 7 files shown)


danield9tqh (approved):

ironfish/src/rpc/routes/eth/accounts.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    const addresses: string[] = []
+    // TODO update once wallet stores evm public address
+    const accounts = node.wallet.listAccounts()
+    for (const account of accounts) {
+      if (!account.spendingKey) {
+        continue
+      }
+      const evmPrivateKey = Uint8Array.from(Buffer.from(account.spendingKey, 'hex'))
+      const evmSenderAddress = Address.fromPrivateKey(evmPrivateKey)
```
hughy, line 38: nit - I think we can use `account.ethAddress` here

hughy (approved):

ironfish/src/rpc/routes/eth/accounts.ts
jowparks: Nice I didn't think that change was merged

PR #5353, opened 2024-09-03
returns unwrapped data from http adapter for json rpc requests

hughy:
## Summary

json rpc responses include only a single object rather than the nested objects in our typical responses

checks or the 'jsonrpc' tag in http rpc requests and returns the response data as a flat object for json rpc requests

## Testing Plan

manual testing:
```
hughy@Hughs-MacBook-Pro ~ % curl http://localhost:8021/eth/ethRouter \
  -X POST \
  -H "Content-Type: application/json" \
  --data '{"method":"eth_blockNumber","params":[],"id":1,"jsonrpc":"2.0"}'
{"jsonrpc":"2.0","id":1,"result":{"number":"0x52"}}%                                                                                                            hughy@Hughs-MacBook-Pro ~ % curl https://docs-demo.quiknode.pro/ \    
  -X POST \
  -H "Content-Type: application/json" \
  --data '{"method":"eth_blockNumber","params":[],"id":1,"jsonrpc":"2.0"}'
{"jsonrpc":"2.0","id":1,"result":"0x13b72f6"}%   
```

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/request.ts
+++ b/ironfish/src/rpc/request.ts
@@ -2,16 +2,20 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 import { Event } from '../event'
+import { JSONUtils } from '../utils'
 
-export function isJsonRcpRequest(body: string | undefined): boolean {
+export function isJsonRcpRequest(body?: string): boolean {
   if (!body) {
     return false
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-  const data = JSON.parse(body)
+  const [data, error] = JSONUtils.tryParse(body)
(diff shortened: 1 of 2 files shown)


ironfish/src/rpc/request.ts
```diff
@@ -1,9 +1,19 @@
...
-
 import { Event } from '../event'
 
+export function isJsonRcpRequest(body: string | undefined): boolean {
+  if (!body) {
+    return false
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  const data = JSON.parse(body)
```
NullSoldier, line 12: we should use tryParse to avoid the exception case.

danield9tqh (approved):

PR #5364, opened 2024-09-12
Rahul/make secret optional multisig secrets

patnir:
## Summary

We need to make the secret field optional to accommodate the ledger device usecase. The secret never leaves the device. 

We can write a migration where we add a flag that indicates if the secret exists or not. 

But we thought of an alternative that doesn't require a migration - store an empty buffer if the secret doesn't exist. When deserializing, if the buffer is all zeros, then you know the secret is undefined. 

I'm thinking through all of the implications of this (not sure it is a patterns we have used before). We are definitely storing extra bytes (but not so much that it makes a large difference). All zeros for this value isn't valid, so that shouldn't be a problem. Are there other issues here I am not thinking of?

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/walletdb/walletdb.ts
```diff
@@ -1460,7 +1460,7 @@ export class WalletDB {
     return (await this.getMultisigSecretByName(name, tx)) !== undefined
   }
 
-  async *getMultisigSecrets(tx?: IDatabaseTransaction): AsyncGenerator<MultisigSecretValue> {
+  async *getMultisigSecrets(tx?: IDatabaseTransaction): AsyncGenerator<MultisigIdentityValue> {
```
hughy, line 1463: nit - should we rename these methods to go along with the rename for the datastore?

ironfish/src/wallet/wallet.ts
```diff
@@ -1424,7 +1424,7 @@ export class Wallet {
       const multisigSecret = await this.walletDb.getMultisigSecret(
         Buffer.from(accountValue.multisigKeys.identity, 'hex'),
       )
-      if (!multisigSecret) {
+      if (!multisigSecret || !multisigSecret.secret) {
```
hughy, line 1427: since we do this check in a couple of places it might be worthwhile to keep a method `getMultisigSecret` that returns just the secret

hughy (approved):

hughy (approved):

PR #5369, opened 2024-09-12
updates ironfish to use latest ironfish-frost changes, etc.

hughy:
## Summary

integrates latest updates to `ironfish-frost` to support Ledger: https://github.com/iron-fish/ironfish-frost/pull/81/

upgrades napi to accommodate dependency conflicts

adds deserialization helper functions to support round3_min

fixes test failures from encryption and error message changes

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


patnir (approved):

ironfish-rust-nodejs/src/multisig.rs
```diff
@@ -78,7 +83,7 @@ pub fn create_signing_commitment(
         &signers,
     );
 
-    let bytes = signing_commitment.serialize();
+    let bytes = signing_commitment.unwrap().serialize();
```
mat-if, line 86: Should we bubble up the error instead of unwrapping here?

ironfish-rust-nodejs/src/multisig.rs
```diff
@@ -415,6 +432,7 @@ pub fn dkg_round2(
         &round1_public_packages,
         thread_rng(),
     )
+    .map_err(|e| IronfishError::new_with_source(IronfishErrorKind::FrostLibError, e))
```
mat-if, line 435: I think we should look into adding fmt::Display for the errors returned from the frost lib in std mode, then I think we can skip spamming these map_err everywhere. Doesn't need to happen right now, just thinking out loud.

ironfish-rust/Cargo.toml
```diff
@@ -43,7 +43,7 @@ chacha20poly1305 = "0.10.1"
 crypto_box = { version = "0.9", features = ["std"] }
 ff = "0.12.0"
 group = "0.12.0"
-ironfish-frost = { git = "https://github.com/iron-fish/ironfish-frost.git", branch = "main" }
+ironfish-frost = { git = "https://github.com/iron-fish/ironfish-frost.git", branch = "main", features = ["std", "signing"] }
```
mat-if, line 46: Do we want to use the default features here, so we don't need to define custom features? The difference I guess is that ironfish-frost has dkg, but we aren't enabling it here https://github.com/iron-fish/ironfish-frost/blob/main/Cargo.toml#L27

ironfish-rust/src/frost_utils/account_keys.rs
```diff
@@ -37,8 +37,10 @@ pub fn derive_account_keys(
         SaplingKey::new(*group_secret_key).expect("failed to derive group secret key");
 
     // Authorization key (ak), obtained from the multisig setup process
-    let authorizing_key = Option::from(SubgroupPoint::from_bytes(&authorizing_key.serialize()))
-        .expect("failied to derive authorizing key");
+    let mut bytes: [u8; 32] = [0; 32];
+    bytes.copy_from_slice(&authorizing_key.serialize().unwrap());
```
mat-if, line 41: We should re-factor this to return a result instead of using an unwrap and expect on the next line

ironfish-rust/src/frost_utils/split_spender_key.rs
```diff
@@ -33,7 +35,7 @@ pub fn split_spender_key(
         split_secret(spender_key, identities, min_signers, thread_rng())?;
 
     assert_eq!(
-        public_key_package.verifying_key().serialize(),
+        public_key_package.verifying_key().serialize().unwrap(),
```
mat-if, line 38: unwrap -> result

ironfish-rust/src/transaction/unsigned.rs
```diff
@@ -229,7 +229,9 @@ impl UnsignedTransaction {
 
         let serialized_signature = authorizing_group_signature.serialize();
 
-        let transaction = self.add_signature(serialized_signature)?;
+        let mut bytes = [0; 64];
+        bytes.copy_from_slice(&serialized_signature.unwrap());
```
mat-if, line 233: unwrap -> result

mat-if (approved):

ironfish-rust-nodejs/src/multisig.rs
hughy: I think we do have fmt::Display in std mode 🤔 

I'll see if we can easily get rid of these

ironfish-rust/Cargo.toml
hughy: Yeah I think we can use the default features here. We also support dkg in the CLI, and I can't remember what's behind the dkg feature in ironfish-frost

PR #5370, opened 2024-09-13
feat(rust-nodejs): Create napi struct for xchacha20poly1305 key

rohanjadvani:
## Summary

Add a napi struct for the xchacha20poly1305 keys that can be used for both a master key and to derive child keys.

## Testing Plan

n/a

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-rust-nodejs/src/xchacha20poly1305.rs
```diff
@@ -2,12 +2,144 @@
...
+use ironfish::xchacha20poly1305::{
+    self, EncryptOutput, XChaCha20Poly1305Key, KEY_LENGTH as KEY_SIZE, SALT_LENGTH as SALT_SIZE,
+    XNONCE_LENGTH as XNONCE_SIZE,
+};
 use napi::{bindgen_prelude::*, JsBuffer};
 use napi_derive::napi;
 
 use crate::to_napi_err;
 
+#[napi]
```
mat-if, line 14: Do you think it'd be helpful / less confusing if we namespace these exports? Similar to how we do with multisig - definitely something I didn't know about before 
https://github.com/iron-fish/ironfish/blob/staging/ironfish-rust-nodejs/src/multisig.rs#L60
https://github.com/iron-fish/ironfish/blob/staging/ironfish-rust-nodejs/index.d.ts#L279

mat-if (approved):

ironfish-rust-nodejs/src/xchacha20poly1305.rs
rohanjadvani: Yeah, I can update it

PR #5372, opened 2024-09-13
feat: Import multisig identity during account import

patnir:
## Summary

When importing a multisig account to a new datadir, the secret is likely to not exist in this data dir. so that means the encrypted communication components of the multisig commands cannot be run. 

This PR changes account import to look for the identity, if it doesn't exist then to import the identity

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


hughy (approved):
is there an easy way to unit test this?

ironfish/src/wallet/wallet.ts
```diff
@@ -1433,12 +1436,26 @@ export class Wallet {
...
+      secret = Buffer.from(accountValue.multisigKeys.secret, 'hex')
     }
 
     if (name && this.getAccountByName(name)) {
       throw new DuplicateAccountNameError(name)
     }
 
+    if (secret) {
+      const duplicateSecret = await this.walletDb.getMultisigSecretByName(accountValue.name)
+      if (duplicateSecret && duplicateSecret.equals(secret)) {
```
hughy, line 1452: should this be `&& !duplicateSecret.equals(secret)`?

patnir: updated the logic here + test

ironfish/src/wallet/wallet.ts
```diff
@@ -1492,6 +1515,24 @@ export class Wallet {
         await this.walletDb.setAccount(account, tx)
       }
 
+      if (secret) {
+        const identitySerialized = new multisig.ParticipantSecret(secret)
+          .toIdentity()
+          .serialize()
+        const multisigIdentity = await this.walletDb.getMultisigIdentity(identitySerialized)
```
hughy, line 1522: We're doing these operations (deserializing the secret, deriving the identity, reading from the db) twice in this method for the same secret. Can we combine these checks into one inside the db transaction?

patnir: Done, I thought there might a slight downside to doing it in the transaction, but its actually better to do so. i'm passing the tx to all the getter methods which will provide a better experience incase there is another db transaction touching that secret

PR #5373, opened 2024-09-14
feat: Import multisig hw identity

patnir:
## Summary

Also cleans up identities after each account import test

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/wallet/wallet.ts
```diff
@@ -1526,6 +1533,28 @@ export class Wallet {
         }
       }
 
+      if (identity) {
```
hughy, line 1536: do you think we could combine this with the `if (secret)` block, above? if there is a secret passed in then we can derive the identity

ironfish/src/wallet/wallet.ts
```diff
@@ -1526,6 +1533,28 @@ export class Wallet {
...
+        const existingIdentity = await this.walletDb.getMultisigIdentity(identity)
+
+        if (!existingIdentity) {
+          const duplicateIdentity = await this.walletDb.getMultisigSecretByName(
```
hughy, line 1540: missing `tx` from these two reads

ironfish/src/wallet/wallet.ts
patnir: I wrote this so that they both wouldn't be defined at the same time (only 1 or the other). treating each account type hardware vs signer separately. 

does it still make sense to combine these?

hughy: i think it makes sense to combine them because `identity` will always be defined if `secret` is defined. the `if (secret)` case becomes the `if (identity)` case if we derive the `identity` from the secret

hughy (approved):

PR #5379, opened 2024-09-17
Displays whether a participant has a secret or not

patnir:
## Summary

This is going to be a useful feature when we start to integrate ledger because all identities from ledger will not export their secret.

Updates both the CLI and related RPC

image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


hughy (approved):
Do you think `Has Secret` or `Ledger` is more helpful/informative to the user?

PR #5384, opened 2024-09-18
use the testnet api url for asset verification when on testnet network

mat-if:


ironfish/src/assets/assetsVerificationApi.ts
```diff
@@ -130,6 +130,14 @@ export class AssetsVerificationApi {
   }
 }
 
+export function getDefaultAssetVerificationEndpoint(networkId?: number): string {
+  if (networkId === 0) {
+    return 'https://testnet.api.ironfish.network/assets/verified_metadata'
+  }
+
+  return 'https://api.ironfish.network/assets/verified_metadata'
```
mat-if, line 138: While it might be more correct to only return this for ID 1, I wanted to get this working for testnet primarily. Changing this would require some more thinking and refactoring on what to do otherwise

hughy (approved):

PR #5387, opened 2024-09-18
adds ledger flag to participant creation

hughy:
## Summary

adds dkgGetIDentity method to ledger util module

adds '--ledger' flag to 'wallet:multisig:participant:create' command

uses 'wallet/multisig/importParticipant' RPC to add the participant identity to the walletDb

## Testing Plan

- manual testing with `wallet:multisig:participant:create --ledger`

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/multisig/participant/create.ts
+++ b/ironfish-cli/src/commands/wallet/multisig/participant/create.ts
@@ -37,7 +37,7 @@ export class MultisigIdentityCreate extends IronfishCommand {
 
     let identity
     if (flags.ledger) {
-      identity = await this.createParticipantWithLedger()
+      identity = await this.getIdentityFromLedger()
     }
 
     let response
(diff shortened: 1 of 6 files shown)


ironfish-cli/src/commands/wallet/multisig/participant/create.ts
```diff
@@ -50,4 +69,20 @@ export class MultisigIdentityCreate extends IronfishCommand {
     this.log('Identity:')
     this.log(response.content.identity)
   }
+
+  async createParticipantWithLedger(): Promise<Buffer> {
```
patnir, line 73: getParticipant/ IdentityFromLedger might be more accurate?

hughy: renamed to `getIdentityFromLedger`

PR #5400, opened 2024-09-20
implements ledger multisig backup command

hughy:
## Summary

adds a CLI command, 'wallet:multisig:ledger:backup', to create an encrypted backup of multisig keys from the ironfish dkg ledger app

users can restore the keys to their ledger app if they reinstall the app on their device or overwrite the multisig keys in the app

## Testing Plan
manual testing: used the command to create key backup

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/multisig/ledger/backup.ts
+++ b/ironfish-cli/src/commands/wallet/multisig/ledger/backup.ts
@@ -26,6 +26,8 @@ export class MultisigLedgerBackup extends IronfishCommand {
     this.log(encryptedKeys.toString('hex'))
     this.log()
     this.log('Please save the encrypted keys show above.')
-    this.log('Use `ironfish wallet:multisig:ledger:restore` if you need to restore the keys to your Ledger.')
+    this.log(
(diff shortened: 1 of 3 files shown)


ironfish-cli/src/commands/wallet/multisig/ledger/backup.ts
```diff
@@ -0,0 +1,28 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { IronfishCommand } from '../../../../command'
+import { Ledger } from '../../../../utils/ledger'
+
+export class MultisigLedgerBackup extends IronfishCommand {
+  static description = `Backup encrypted multisig keys from a Ledger device`
```
mat-if, line 8: Trying to think of a way to make this more in-line with the new description format. Would something like `show encrypted multisig key from a Ledger device` be accurate? As far as I can tell, the "backup" is really "log to user", if I'm understanding this correctly.

mat-if (approved):
minor nit comment but looks good 👍

PR #5401, opened 2024-09-20
adds cli command to restore ledger multisig backup

hughy:
## Summary

if the ironfish dkg ledger app is uninstalled, or if it is used to create a second set of keys, then the multisig keys on the device may be lost

the 'wallet:multisig:ledger:backup' command creates an encrypted key backup that can be restored to the device

the 'wallet:multisig:ledger:restore' command restores the backed up keys to the ledger

## Testing Plan
manual testing:
- created backup using 'wallet:multisig:ledger:backup' (#5400 )
- uninstalled and reinstalled ironfish dkg app
- ran 'wallet:multisig:ledger:restore' to restore keys and printed our keys to verify that they are the same as before

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/multisig/ledger/restore.ts
+++ b/ironfish-cli/src/commands/wallet/multisig/ledger/restore.ts
@@ -1,0 +1,45 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Flags } from '@oclif/core'
+import { IronfishCommand } from '../../../../command'
+import * as ui from '../../../../ui'
+import { Ledger } from '../../../../utils/ledger'
+
+export class MultisigLedgerRestore extends IronfishCommand {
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/commands/wallet/multisig/ledger/restore.ts
```diff
@@ -0,0 +1,45 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+import { Flags } from '@oclif/core'
+import { IronfishCommand } from '../../../../command'
+import * as ui from '../../../../ui'
+import { Ledger } from '../../../../utils/ledger'
+
+export class MultisigLedgerRestore extends IronfishCommand {
+  static description = `Restore encrypted multisig keys to a Ledger device`
```
mat-if, line 10: Same comment as the other PR, but I think the phrasing is fine here
```suggestion
  static description = `restore encrypted multisig keys to a Ledger device`
```

mat-if (approved):

PR #5410, opened 2024-09-21
Adding ledger support for dkg:create

patnir:
## Summary

Also added the ability to reuse an existing identity for the DKG process. The user can specify the account name separately from the participant name.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
+++ b/ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
@@ -2,75 +2,203 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-import { RpcClient } from '@ironfish/sdk'
+import {
+  deserializePublicPackage,
+  deserializeRound2CombinedPublicPackage,
+} from '@ironfish/rust-nodejs'
+import { AccountFormat, Assert, encodeAccountImport, RpcClient } from '@ironfish/sdk'
+import { Flags } from '@oclif/core'
 import { IronfishCommand } from '../../../../command'
 import { RemoteFlags } from '../../../../flags'
 import * as ui from '../../../../ui'
+import { Ledger } from '../../../../utils/ledger'
 
 export class DkgCreateCommand extends IronfishCommand {
   static description = 'Interactive command to create a multisignature account using DKG'
 
   static flags = {
     ...RemoteFlags,
+    participant: Flags.string({
+      char: 'n',
+      description: 'The name of the secret to use for encryption during DKG',
+    }),
+    newAccount: Flags.string({
+      char: 'a',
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
```diff
@@ -2,102 +2,228 @@
...
 export class DkgCreateCommand extends IronfishCommand {
   static description = 'Interactive command to create a multisignature account using DKG'
 
   static flags = {
     ...RemoteFlags,
+    participant: Flags.string({
+      char: 'n',
+      description: 'The name of the secret to use for encryption during DKG',
+    }),
+    newAccount: Flags.string({
```
hughy, line 25: I think that handling the account name separately from the participant name opens us up to a lot of difficult edge cases if/when the names don't match

We might need to add RPCs for renaming and/or deleting identities from the wallet db in case users get 'stuck' with an identity name they don't want, but I think we should do whatever we can to try to make the names match

ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
```diff
@@ -123,12 +249,51 @@ export class DkgCreateCommand extends IronfishCommand {
...
+    this.log('\nRound 1 Encrypted Secret Package:\n')
+    this.log(secretPackage.toString('hex'))
+    this.log()
+
+    this.log('\nRound 1 Public Package:\n')
+    this.log(publicPackage.toString('hex'))
+    this.log()
+
+    this.log('Next step:')
+    this.log('Send the round 1 public package to each participant')
```
hughy, line 280: this causes double printing of these messages

ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
```diff
@@ -2,102 +2,228 @@
...
+  async getIdentityFromLedger(
+    ledger: Ledger,
+    client: RpcClient,
+    name?: string,
+  ): Promise<{
+    name: string
+    identity: string
+  }> {
+    // TODO(hughy): support multiple identities using index
+    const identity = await ledger.dkgGetIdentity(0)
```
hughy, line 170: nit - maybe we should log that we're requesting the identity? We ask the user to approve the request, but it might not be clear what the request is

hughy (approved):

ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
patnir: done, changed the prompt message

ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
patnir: removed

hughy (approved):

PR #5418, opened 2024-09-24
Enable transfering in a mint in the CLI

NullSoldier:
## Summary

This allows you to both mint new coins, and transfer them in the same transaction using the CLI. This is something that was already possible using the SDK or the RPC.

This is done by adding outputs with the same amount as the mint amount.

Example to mint 1 new asset called `foo-coin` and transfer the 1 new coin to `eccfd79c841e02cda4e2c53f3a149070f1bc8c7f76a41886d268b6ec8ed2ae3b`

```
fishy wallet:mint \
--transferTo eccfd79c841e02cda4e2c53f3a149070f1bc8c7f76a41886d268b6ec8ed2ae3b \
--transferMemo "this is a memo" \
--name foo-coin  \
--metadata meta  \
--amount 1  \
--fee 0.00000001
```

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/mint.ts
+++ b/ironfish-cli/src/commands/wallet/mint.ts
@@ -98,7 +98,7 @@ This will create tokens and increase supply for a given asset.`
       description: 'The public address of the account to transfer ownership of this asset to.',
     }),
     transferTo: Flags.string({
-      description: 'transfer all newly minted coins to this public address.',
+      description: 'transfer all newly minted coins to this public address',
     }),
     transferToMemo: Flags.string({
       description: 'The memo of transfer when using transferTo',
--- a/ironfish-cli/src/commands/wallet/mint.ts
+++ b/ironfish-cli/src/commands/wallet/mint.ts
@@ -97,6 +97,12 @@ This will create tokens and increase supply for a given asset.`
     transferOwnershipTo: Flags.string({
       description: 'The public address of the account to transfer ownership of this asset to.',
     }),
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/wallet/mint.ts
```diff
@@ -97,6 +97,12 @@ This will create tokens and increase supply for a given asset.`
     transferOwnershipTo: Flags.string({
       description: 'The public address of the account to transfer ownership of this asset to.',
     }),
+    transferTo: Flags.string({
+      description: 'transfer all newly minted coins to this public address.',
```
patnir, line 101:
```suggestion
      description: 'Transfer all newly minted coins to this public address.',
```

patnir:
https://testnet.explorer.ironfish.network/transaction/c92a1ad7fb352b66c7bc119943886228aa3c27d58f3602758457204f511f33e7

Was able to do this successfully!

patnir (approved):

ironfish-cli/src/commands/wallet/mint.ts
NullSoldier: Descriptions for flags should all be lower case since the CLI audit.

From our style guide...

> Ensure that descriptions are provided for all flags, that the descriptions are in lowercase, that they are concise (so as to fit on narrow screens), and that they do not end in a period to match other flag descriptions.

PR #5422, opened 2024-09-24
chainport config update with mainnet fields

patnir:
## Summary

Mainnet support for chainport CLI

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/chainport/send.ts
+++ b/ironfish-cli/src/commands/wallet/chainport/send.ts
@@ -2,10 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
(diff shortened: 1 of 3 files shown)


ironfish-cli/src/commands/wallet/chainport/send.ts
```diff
@@ -84,7 +84,7 @@ export class BridgeCommand extends IronfishCommand {
 
     const networkId = (await client.chain.getNetworkInfo()).content.networkId
 
-    if (networkId !== TESTNET.id) {
+    if (networkId !== TESTNET.id && networkId !== MAINNET.id) {
       this.error(`Chainport transactions are only available on testnet.`)
```
leanthebean, line 88: wouldn't the error message change here too?

patnir: you're right, my bad

PR #5425, opened 2024-09-25
separates Ledger DKG logic from single signer

hughy:
## Summary

defines a separate class, LedgerDkg, for interacting with the Ledger Ironfish DKG app

updates all DKG and multisig commands to use LedgerDkg

adds aliased dependencies @zondax/ledger-js-dkg and @zondax/ledger-ironfish-dkg to use the versions of these packages that we need for the DKG app

restores the previous version of the Ledger class for use with the single signer Ledger app

uses v0.1.2 of @zondax/ledger-ironfish for compatibility with the single signer app

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/wallet/multisig/commitment/create.ts
```diff
@@ -124,9 +124,9 @@ export class CreateSigningCommitmentCommand extends IronfishCommand {
     unsignedTransaction: UnsignedTransaction,
     signers: string[],
   ): Promise<void> {
-    const ledger = new Ledger(this.logger)
+    const ledger = new LedgerDkg(this.logger)
     try {
-      await ledger.connect(true)
+      await ledger.connect()
```
patnir, line 129: i thought this is supposed to be true

ironfish-cli/src/utils/ledger.ts
```diff
@@ -358,15 +269,164 @@ export class Ledger {
...
+  logger: Logger
+  PATH = "m/44'/1338'/0"
+
+  constructor(logger?: Logger) {
+    this.app = undefined
+    this.logger = logger ? logger : createRootLogger()
+  }
+
+  connect = async () => {
+    const transport = await TransportNodeHid.create(3000, 3000)
```
patnir, line 283:
```suggestion
    const transport = await TransportNodeHid.create(3000)
```

mat-if (approved):

patnir (approved):

PR #5426, opened 2024-09-25
Error for too many participants

patnir:
## Summary

At the moment, ledger dkg only supports up to 4 participants. This commit adds an error message when the user tries to create a dkg with more than 4 participants.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
+++ b/ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
@@ -269,6 +269,10 @@ export class DkgCreateCommand extends IronfishCommand {
       throw new Error('Total number of participants must be at least 2')
     }
 
+    if (ledger && totalParticipants > 4) {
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
```diff
@@ -269,6 +269,10 @@ export class DkgCreateCommand extends IronfishCommand {
       throw new Error('Total number of participants must be at least 2')
     }
 
+    if (ledger && totalParticipants > 4) {
+      throw new Error('Ledger dkg only supports up to 4 participants')
```
mat-if, line 273:
```suggestion
      throw new Error('DKG with Ledger supports a maximum of 4 participants')
```

mat-if (approved):

NullSoldier (approved):

PR #5443, opened 2024-09-26
inputNumberPrompt and usage

patnir:
## Summary

Adds prompt util specific to numbers 

The problem with parse int is that even if you pass it something like "1234123asdf" it will return a success and return 1234123.
This is not what we want. We want to make sure that the user has entered a valid number.

We also want to retry the prompt if the user enters an invalid number or a decimal when an integer is expected.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
+++ b/ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
@@ -10,8 +10,8 @@ import {
   ACCOUNT_SCHEMA_VERSION,
   AccountFormat,
   Assert,
-  encodeAccountImport,
   RpcClient,
+  encodeAccountImport,
 } from '@ironfish/sdk'
 import { Flags } from '@oclif/core'
 import { IronfishCommand } from '../../../../command'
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
```diff
@@ -304,11 +308,11 @@ export class DkgCreateCommand extends IronfishCommand {
       errorOnDuplicate: true,
     })
 
-    input = await ui.inputPrompt('Enter the number of minimum signers', true)
-    const minSigners = parseInt(input)
-    if (isNaN(minSigners) || minSigners < 2) {
-      throw new Error('Minimum number of signers must be at least 2')
-    }
+    const minSigners = await ui.inputNumberPrompt(
```
hughy, line 311: Do we still need to check whether `minSigners < 2`?

patnir: added back

hughy (approved):

PR #5449, opened 2024-09-27
Update zondax ironfish to 0.5.1

patnir:
## Summary

We previously had to have two versions of this SDK available because the newer one did not work with single signer.

Zondax fixed the bugs and now we don't need to manage two separate dependencies now.

Created a base ledger class for the common functionality.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/utils/ledger.ts
```diff
@@ -257,131 +239,57 @@ export class LedgerDkg {
...
     await this.tryInstruction((app) => app.dkgRestoreKeys(encryptedKeys))
   }
 }
 
-export class Ledger {
-  app: IronfishApp | undefined
-  logger: Logger
-  PATH = "m/44'/1338'/0"
-
+export class Ledger extends LedgerBase {
```
NullSoldier, line 248: nit: should be differentiated that this is the single signer some how so it's not confused with LedgerBase.

hughy (approved):
tested single signer, dkg, and multisig signing

ironfish-cli/package.json
```diff
@@ -150,4 +148,4 @@
     "url": "https://github.com/iron-fish/ironfish/issues"
   },
   "homepage": "https://ironfish.network"
-}
+}
```
NullSoldier, line 151: nit: revert this & check your editor settings so it's not modifying and removing the end of file newline like all of our other files have.

ironfish-cli/src/utils/ledger.ts
```diff
@@ -17,36 +17,32 @@ import TransportNodeHid from '@ledgerhq/hw-transport-node-hid'
...
-  default as IronfishDkgApp,
-  KeyResponse,
-  ResponseAddress as ResponseAddressDkg,
-  ResponseDkgRound1,
-  ResponseDkgRound2,
-  ResponseIdentity,
-  ResponseProofGenKey as ResponseProofGenKeyDkg,
-  ResponseViewKey as ResponseViewKeyDkg,
-} from '@zondax/ledger-ironfish-dkg'
 import { ResponseError } from '@zondax/ledger-js'
```
NullSoldier, line 29: This is a bug because we're importing from a library that we don't have a dependency on. We shouldn't import from dependencies of our dependencies. We need to include this in our package.json.

ironfish-cli/package.json
patnir: done

ironfish-cli/package.json
```diff
@@ -68,9 +68,7 @@
     "@oclif/plugin-warn-if-update-available": "3.1.8",
     "@types/keccak": "3.0.4",
     "@types/tar": "6.1.1",
-    "@zondax/ledger-ironfish": "0.1.2",
-    "@zondax/ledger-ironfish-dkg": "npm:@zondax/ledger-ironfish@0.4.0",
-    "@zondax/ledger-js": "^1.0.1",
```
NullSoldier, line 73: Bring this back

hughy: We added this dependency because we needed access to `ResponseError`. Managing two versions of `@zondax/ledger-ironfish` caused problems with resolving to the right version, but we needed our own dependency for `ResponseError`

NullSoldier: I think that's fine, but you cannot use an import without referencing it in your package because if the library you depend on stops including it your code will stop compiling. You are making the code base depend on that library by importing it, therefore you need to add it to the package to indicate the dependency.

mat-if: A good way to see this in action is to comment out `import { ResponseError } from '@zondax/ledger-js'` in ledger.ts, then press CMD + . to trigger the autofix. Without this being in the package.json, it will not attempt to auto-import anything. If this is in the package.json, it correctly suggests to add the import.

patnir: @mat-if tried this, it is not in this package.json . They haven't exposed it in `"@zondax/ledger-ironfish"`. I asked zondax to do it, and they said they should release it today

patnir:
Going to merge this now and made a note to import ResponseError from the top level dependency once zondax makes the change. Context: https://iflabs.slack.com/archives/C06FQDY6CNT/p1727824902604019

PR #5457, opened 2024-09-28
Use Iron Fish API for bridge info

dguenther:
## Summary

Switches from using the Chainport API directly to using a proxy through [our API](https://github.com/iron-fish/ironfish-api/pull/1767) to avoid breaking old CLI versions when Chainport deploys their new API.

Fixes IFL-2977

## Testing Plan

* [x] Loaded incoming and outgoing transactions on preprod
* [x] Went through the send flow up until confirmation (but didn't send) on prod
* [x] Went through the send flow up until confirmation on preprod (The preprod Iron Fish metadata API is currently down)

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/chainport/send.ts
+++ b/ironfish-cli/src/commands/wallet/chainport/send.ts
@@ -283,7 +283,7 @@ export class BridgeCommand extends IronfishCommand {
   ) {
     const { flags } = await this.parse(BridgeCommand)
 
-    ux.action.start('Fetching bridge transaction fees')
+    ux.action.start('Fetching bridge transaction details')
     const txn = await fetchChainportBridgeTransaction(
       networkId,
       amount,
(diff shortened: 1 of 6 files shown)


ironfish-cli/src/commands/wallet/chainport/send.ts
```diff
@@ -280,12 +278,20 @@ export class BridgeCommand extends IronfishCommand {
...
     to: string,
     amount: bigint,
-    asset: ChainportVerifiedToken,
+    asset: ChainportToken,
     assetData: RpcAsset,
   ) {
     const { flags } = await this.parse(BridgeCommand)
 
-    const txn = await fetchChainportBridgeTransaction(networkId, amount, to, network, asset)
+    ux.action.start('Fetching bridge transaction fees')
```
patnir, line 286:
```suggestion
    ux.action.start('Fetching bridge transaction details')
```

patnir:
image

was able to test sending a bridge transactoin successfully

patnir:
was not able to confirm, but derek mentioned one of the apis is down

image

patnir:
Only comment is that each step seemed kind of slow (which might be an issue with our infra)

patnir:
Future consideration, seems like we should probably stop the CLI command here: 

image

dguenther:
> was not able to confirm, but derek mentioned one of the apis is down
> 

This is no longer the case, they should all be working now.

PR #5458, opened 2024-09-29
Fix wallet:chainport:send fee selection when -a is used

dguenther:
## Summary

`wallet:chainport:send`'s fee selector wasn't respecting the `-a` flag.

Fixes IFL-3003

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/utils/fees.ts
+++ b/ironfish-cli/src/utils/fees.ts
@@ -20,7 +20,7 @@ import { promptCurrency } from './currency'
 export async function selectFee(options: {
   client: Pick<RpcClient, 'wallet'>
   transaction: CreateTransactionRequest
-  account?: string
+  account: string | undefined
   confirmations?: number
   logger: Logger
 }): Promise<RawTransaction> {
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/utils/fees.ts
```diff
@@ -20,7 +20,7 @@ import { promptCurrency } from './currency'
 export async function selectFee(options: {
   client: Pick<RpcClient, 'wallet'>
   transaction: CreateTransactionRequest
-  account?: string
+  account: string | undefined
```
patnir, line 23: this is effectively the same thing right?

dguenther: Not quite, with `string | undefined` and no `?`, the developer is required to specify account in the object (at least in the ways that we use this function). I added this change because I've seen this bug pretty much every time we've used `selectFee`: https://github.com/iron-fish/ironfish/pull/3919

patnir (approved):

PR #5464, opened 2024-10-02
multisig server signing

hughy:
## Summary

defines common interface for storing dkg sessions and signing sessions in server memory

adds network messages for signing flow: starting a signing session, signing commitments, signature shares, and retrieving the status of a signing session

updates server and client implementations for new message types

updates 'sign' command:
- adds '--server' flag
- adds '--sessionId' flag
- prompts user to enter a session id or start a new session if server is set, but not session id
- connects to server, starts or joins a session, and uses messages to/from server to complete signing

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/multisig/sign.ts
+++ b/ironfish-cli/src/commands/wallet/multisig/sign.ts
@@ -4,17 +4,22 @@
 
 import { multisig } from '@ironfish/rust-nodejs'
 import {
+  Assert,
   CurrencyUtils,
   Identity,
+  parseUrl,
+  PromiseUtils,
   RpcClient,
   Transaction,
   UnsignedTransaction,
 } from '@ironfish/sdk'
 import { Flags, ux } from '@oclif/core'
+import dns from 'dns'
 import { IronfishCommand } from '../../../command'
 import { RemoteFlags } from '../../../flags'
 import * as ui from '../../../ui'
 import { LedgerDkg } from '../../../utils/ledger'
+import { MultisigTcpClient } from '../../../utils/multisig/network'
 import { renderUnsignedTransactionDetails, watchTransaction } from '../../../utils/transaction'
 
 // todo(patnir): this command does not differentiate between a participant and an account.
@@ -42,6 +47,13 @@ export class SignMultisigTransactionCommand extends IronfishCommand {
     ledger: Flags.boolean({
       default: false,
       description: 'Perform operation with a ledger device',
+    }),
+    server: Flags.string({
+      description: "multisig server to connect to. formatted as '<host>:<port>'",
+    }),
+    sessionId: Flags.string({
+      description: 'Unique ID for a multisig server session to join',
+      dependsOn: ['server'],
     }),
   }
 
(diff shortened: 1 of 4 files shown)


ironfish-cli/src/commands/wallet/multisig/sign.ts
```diff
@@ -289,19 +411,45 @@ export class SignMultisigTransactionCommand extends IronfishCommand {
...
-      errorOnDuplicate: true,
-    })
+      commitments = await ui.collectStrings('Commitment', identities.length - 1, {
+        additionalStrings: [commitment],
+        errorOnDuplicate: true,
+      })
+    } else {
+      multisigClient.submitSigningCommitment(commitment)
+
+      multisigClient.onSigningStatus.on((message) => {
```
danield9tqh, line 434: Why is there multiple ways of getting the signing commitments?

ironfish-cli/src/commands/wallet/multisig/sign.ts
```diff
@@ -314,38 +462,54 @@ export class SignMultisigTransactionCommand extends IronfishCommand {
...
+    } else {
+      multisigClient.submitIdentity(participant.identity)
 
-    this.log(
-      `Enter ${totalParticipants - 1} identities of the participants (excluding your own)`,
-    )
+      multisigClient.onSigningStatus.on((message) => {
+        identities = message.identities
+      })
+      multisigClient.onIdentity.on((message) => {
```
danield9tqh, line 491: Same here, there a reason we do it both ways?

ironfish-cli/src/commands/wallet/multisig/sign.ts
hughy: I agree it's simpler for the client if the server broadcasts the full list once it's complete instead of broadcasting them one at a time. The only thing we'd lose there is the ability to show progress (e.g., '3/4 commitments')

This polls the status in addition to listening for broadcast commitments in case you miss a broadcast by being behind or disconnected.

ironfish-cli/src/commands/wallet/multisig/sign.ts
```diff
@@ -187,25 +229,105 @@ export class SignMultisigTransactionCommand extends IronfishCommand {
     )
 
     this.log('Mutlisignature sign process completed!')
```
mat-if, line 231:
```suggestion
    this.log('Multisignature sign process completed!')
```

ironfish-cli/src/commands/wallet/multisig/sign.ts
```diff
@@ -187,25 +229,105 @@ export class SignMultisigTransactionCommand extends IronfishCommand {
...
+        }
+      })
+
+      ux.action.start('Waiting for other Signature Shares from server')
+      while (signatureShares.length < totalParticipants) {
+        multisigClient.getSigningStatus()
+        await PromiseUtils.sleep(3000)
+      }
+
+      multisigClient.onDkgStatus.clear()
```
NullSoldier, line 327: This status is used during signing?

ironfish-cli/src/commands/wallet/multisig/sign.ts
```diff
@@ -289,19 +411,45 @@ export class SignMultisigTransactionCommand extends IronfishCommand {
...
+      })
+
+      ux.action.start('Waiting for other Signing Commitments from server')
+      while (commitments.length < totalParticipants) {
+        multisigClient.getSigningStatus()
+        await PromiseUtils.sleep(3000)
+      }
+
+      multisigClient.onDkgStatus.clear()
+      multisigClient.onIdentity.clear()
```
NullSoldier, line 450: check these clears are the messages that are subscribed

ironfish-cli/src/commands/wallet/multisig/sign.ts
NullSoldier: Once we add encryption, this code will need to decrypt and verify that it's a valid commitment so the server won't actually know once all valid commitments are there. It works now because it's plain text.

NullSoldier (approved):

PR #5474, opened 2024-10-02
Use ledger action to make ledger more reliable

NullSoldier:
## Summary

This UI action will handle many of the failure cases that occur when
running ledger commands. They'll use CLI-UX to inform the user of the
current state, and what the state should be.

TODO:
- [x] Make it work seemlessly when existing actions are running such as when using the interactive multisig server
- [x] Investigate issue with lock screen spamming

## Testing Plan

- Plug in a ledger and perform multisig with a server and without a server.
- Plug in a ledger and perform multisig while hosting the server
- Plug in a ledger and perform multisig while joining the server

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


patnir:
Tested this with a bunch of cases - device falls asleep, i lock the device, and the app is not open. 

image

ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
```diff
@@ -70,15 +70,7 @@ export class DkgCreateCommand extends IronfishCommand {
...
       ledger = new LedgerMultiSigner(this.logger)
-      try {
-        await ledger.connect()
-      } catch (e) {
-        if (e instanceof Error) {
-          this.error(e.message)
-        } else {
-          throw e
-        }
-      }
```
patnir, line 81: This happens in a lot of places, needs to be replaced everywhere. Can happen later though

ironfish-cli/src/ui/ledger.ts
```diff
@@ -0,0 +1,62 @@
...
+          ux.action.start('Ledger handshake failed... trying again')
+        } else if (e instanceof LedgerClaNotSupportedError) {
+          const appName = ledger.isMultisig ? 'Ironfish' : 'Ironfish DKG'
+          ux.action.start(`Wrong Ledger app opened. Please open ${appName} APP.`)
+        } else if (e instanceof LedgerAppNotOpen) {
+          const appName = ledger.isMultisig ? 'Ironfish' : 'Ironfish DKG'
+          ux.action.start(`Unlock your Ledger and open the ${appName} APP.`)
+        } else {
+          throw e
+        }
```
patnir, line 48: seeing this work is so sick. assuming once this PR gets in we need to use ledgerAction everywhere?

patnir:
This is draft, but I would approve.

hughy (approved):
Tested out and works well!

I think as a future improvement we might move some log output out of the Ledger methods so that it doesn't repeat on retry. For example:
image

PR #5480, opened 2024-10-03
Cleanup sessions on Multisig Broker Server

mat-if:
## Summary

- Added logic on the server to check if a session has 0 connected clients when a client disconnects. If it does, it will delete the session
- Added a connected response from the server to the client when a client connects. Previously, it was sending the join session message immediately after connecting, which would lead to the message getting lost in the void

Closes IFL-3022

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
+++ b/ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
@@ -112,13 +112,22 @@ export class DkgCreateCommand extends IronfishCommand {
       })
       multisigClient.start()
 
+      let connectionConfirmed = false
+
       multisigClient.onConnectedMessage.on(() => {
-        if (sessionId) {
-          Assert.isNotNull(multisigClient)
-          multisigClient.joinSession(sessionId)
-          multisigClient.onConnectedMessage.clear()
-        }
+        connectionConfirmed = true
+        Assert.isNotNull(multisigClient)
+        multisigClient.onConnectedMessage.clear()
       })
+
+      if (sessionId) {
+        while (!connectionConfirmed) {
+          await PromiseUtils.sleep(500)
+          continue
+        }
+
+        multisigClient.joinSession(sessionId)
+      }
     }
 
     const { name: participantName, identity } = ledger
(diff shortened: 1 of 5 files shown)


ironfish-cli/src/multisigBroker/server.ts
```diff
@@ -223,6 +230,24 @@ export class MultisigServer {
     this.clients.delete(client.id)
   }
 
+  /**
+   * If a client has the given session ID and is connected, the associated
+   * session should still be considered active
+   */
+  private isSessionActive(sessionId: string): boolean {
+    for (const client of this.clients.values()) {
```
hughy, line 238: Keeping this in mind for a followup task - do you think we should store a list of client IDs on the session? Here and in `broadcast` we iterate through all of the clients and filter out the ones that aren't in the right session 🫣

ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
```diff
@@ -112,9 +112,13 @@ export class DkgCreateCommand extends IronfishCommand {
...
-      if (sessionId) {
-        multisigClient.joinSession(sessionId)
-      }
+      multisigClient.onConnectedMessage.on(() => {
+        if (sessionId) {
+          Assert.isNotNull(multisigClient)
+          multisigClient.joinSession(sessionId)
+          multisigClient.onConnectedMessage.clear()
+        }
+      })
```
hughy, line 121: Should we block on waiting for a confirmed connection? If the `connected` message gets dropped for some reason then we'll continue without ever joining the session. Maybe that's being too paranoid about dropped messages 😅

hughy (approved):

hughy (approved):

ironfish-cli/src/multisigBroker/server.ts
mat-if: Yeah I think that'd be a good change. I imagine we'll eventually get more and more logic around that, so having it easily accessible will be good.

PR #5482, opened 2024-10-03
Add ability for sessions to track specific clients

mat-if:
## Summary

Sessions now track client IDs
Server has encapsulated logic to add/remove clients to/from sessions

Removes the need to iterate over all clients to determine if a session is active or not, and for broadcasting

Closes IFL-3029

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/multisigBroker/server.ts
+++ b/ironfish-cli/src/multisigBroker/server.ts
@@ -317,8 +317,9 @@ export class MultisigServer {
       const client = this.clients.get(clientId)
       if (!client) {
         this.logger.debug(
-          `Client ${clientId} does not exist, but session ${sessionId} thinks it does`,
+          `Client ${clientId} does not exist, but session ${sessionId} thinks it does, removing.`,
         )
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/multisigBroker/server.ts
```diff
@@ -270,8 +305,18 @@ export class MultisigServer {
...
+      this.logger.debug(`Session ${sessionId} does not exist, broadcast failed`)
+      return
+    }
+
+    for (const clientId of session.clientIds) {
+      const client = this.clients.get(clientId)
+      if (!client) {
+        this.logger.debug(
+          `Client ${clientId} does not exist, but session ${sessionId} thinks it does`,
+        )
```
hughy, line 319: Should we remove the client from the session in this case? We shouldn't get to this state, but it could prevent the session from being cleaned up

hughy (approved):

hughy (approved):

PR #5488, opened 2024-10-07
Upgrade wallet:import to use ui.ledger

NullSoldier:
## Summary

This uses the new ledger UI action so that it will wait for the user to connect and unlock their ledger.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/import.ts
+++ b/ironfish-cli/src/commands/wallet/import.ts
@@ -7,6 +7,7 @@ import { IronfishCommand } from '../../command'
 import { RemoteFlags } from '../../flags'
 import { LedgerError, LedgerSingleSigner } from '../../ledger'
 import { checkWalletUnlocked, inputPrompt } from '../../ui'
+import * as ui from '../../ui'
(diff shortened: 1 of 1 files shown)


hughy (approved):

ironfish-cli/src/commands/wallet/import.ts
```diff
@@ -119,8 +120,13 @@ export class ImportCommand extends IronfishCommand {
...
       const ledger = new LedgerSingleSigner(this.logger)
-      await ledger.connect()
-      const account = await ledger.importAccount()
+
+      const account = await ui.ledger({
+        ledger,
+        message: 'Import Wallet',
+        action: () => ledger.importAccount(),
+      })
+
```
hughy, line 129: nit - it looks like import needs some approval, should we add `approval: true`?

NullSoldier: Done, thanks for auditing which ones needed approval.

PR #5490, opened 2024-10-07
resets ledger ui status between errors

hughy:
## Summary

removes the error status (e.g., 'Open Ledger App Ironfish') after the user has resolved the cause of the error

## Testing Plan
manual testing with `wallet:import --ledger`

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/ui/ledger.ts
+++ b/ironfish-cli/src/ui/ledger.ts
@@ -53,6 +53,7 @@ export async function ledger<TResult>({
     // eslint-disable-next-line no-constant-condition
     while (true) {
       try {
+        ux.action.status = undefined
         const result = await action()
         ux.action.stop()
         return result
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/ui/ledger.ts
```diff
@@ -42,6 +42,7 @@ export async function ledger<TResult>({
     // eslint-disable-next-line no-constant-condition
     while (true) {
       try {
+        ux.action.status = undefined
```
NullSoldier, line 45: I thought about this, but the only real issue here is that if the user is in an error loop this might cause it to keep flickering and make it hard to read right? The error will display for 1 second and then be deleted. Let's test this.

NullSoldier (approved):

PR #5497, opened 2024-10-07
Add flags for minSigners and totalParticipants for dkg:create

mat-if:
## Summary

Makes it a little easier to get through the setup phase for power users

## Testing Plan

- Test with server and with flags: Should not prompt for total participants and min signers
`wallet:multisig:dkg:create --server <...> --totalParticipants 2 --minSigners 2`
- Test with server and without flags: Should prompt for total participants and min signers
`wallet:multisig:dkg:create --server <...>`
- Test without server and with flags: Should not prompt for total participants and min signers
`wallet:multisig:dkg:create --totalParticipants 2 --minSigners 2`
- Test without server and without flags: Should prompt for total participants and min signers
`wallet:multisig:dkg:create`
- Test with sessionId and with flags: Should throw an error due to these flags being exclusive
`wallet:multisig:dkg:create --sessionId foo --totalParticipants 2 --minSigners 2`
- Test with sessionId and without flags: Should not throw an error
`wallet:multisig:dkg:create --sessionId foo`
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
+++ b/ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
@@ -92,7 +92,7 @@ export class DkgCreateCommand extends IronfishCommand {
     let multisigClient: MultisigClient | null = null
     if (flags.server) {
       let sessionId = flags.sessionId
-      if (!sessionId) {
+      if (!sessionId && !flags.totalParticipants && !flags.minSigners) {
         sessionId = await ui.inputPrompt(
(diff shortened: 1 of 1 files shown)


hughy (commented):
one additional suggestion: if a user sets either `--totalParticipants` or `--minSigners` they _must_ start a new session, so we should skip the `sessionId` prompt

hughy (approved):

PR #5511, opened 2024-10-08
allows tryInstruction to skip error checks

hughy:
## Summary

calling another instruction like 'getVersion' between 'reviewTransaction' and 'dkgGetCommitments' will cause the transaction hash stored in device memory to be cleared and cause 'dkgGetCommitments' to fail

adds an 'unsafe' option to bypass these calls in 'tryInstruction' and changes 'dkgGetCommitments' and 'dkgSign' to use the unsafe option

these methods will now work (provided they are called after 'reviewTransaction'), but they will not have the benefits of the error handling from 'appInfo' and 'getVersion'

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/ledger/ledger.ts
+++ b/ironfish-cli/src/ledger/ledger.ts
@@ -29,10 +29,7 @@ export class Ledger {
     this.isMultisig = isMultisig
   }
 
-  async tryInstruction<T>(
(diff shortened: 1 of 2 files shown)


NullSoldier:
We're closing this in favor of https://github.com/iron-fish/ironfish/pull/5512

PR #5520, opened 2024-10-08
adds default multisig broker server

hughy:
## Summary

changes '--server' to boolean flag in 'wallet:multisig:sign' and 'wallet:multisig:dkg:create'

adds '--hostname' and '--port' flags in place of previous '--server' flag to supply server hostname and port

defaults 'hostname' to 'multisig.ironfish.network' and 'port' to 9035

adds support for connection strings with '--connection' flag

adds util function to parse all connection options from flags

prints connection string to console after starting session

Closes IFL-3043

## Testing Plan
```
? Enter a name for the multisig account: test0
? Enter the ID of a multisig session to join, or press enter to start a new session: 
? Enter the passphrase for the multisig session: test
? Enter the total number of participants: 2
? Enter the number of minimum signers: 2

Started new DKG session:
91b68c04-87ea-42c6-9f7a-4f9723f1e045

DKG session connection string:
tcp://91b68c04-87ea-42c6-9f7a-4f9723f1e045:test@multisig.ironfish.network:9035
```

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
+++ b/ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
@@ -61,11 +61,9 @@ export class DkgCreateCommand extends IronfishCommand {
     }),
     sessionId: Flags.string({
       description: 'Unique ID for a multisig server session to join',
-      dependsOn: ['server'],
     }),
     passphrase: Flags.string({
       description: 'Passphrase to join the multisig server session',
-      dependsOn: ['server'],
     }),
     tls: Flags.boolean({
       description: 'connect to the multisig server over TLS',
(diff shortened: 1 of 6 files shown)


mat-if (approved):
Some observations:
- sessionId depends on server, so you cant do `wallet:multisig:dkg:create --sessionId ...`. This made sense prior, but now, I wonder if a better experience would be to assume connection to a server if given a session id. I think this logic holds for passphrase as well.
- The various permutations of the connection string appear to work (omitted session, and also omitted password on subsequent tries and it worked as expected) which was 👌 
- I alllllmost feel like we can do away with `--server` altogether, or combine server and host or something. Especially if all of the other flags imply server=true. I need to think through it more

hughy:
>     * sessionId depends on server, so you cant do `wallet:multisig:dkg:create --sessionId ...`. This made sense prior, but now, I wonder if a better experience would be to assume connection to a server if given a session id. I think this logic holds for passphrase as well.

I'll update so that `sessionId` and `passphrase` don't depend on `server` and so that they imply `server == true`

>     * The various permutations of the connection string appear to work (omitted session, and also omitted password on subsequent tries and it worked as expected) which was 👌

👌👌

>     * I alllllmost feel like we can do away with `--server` altogether, or combine server and host or something. Especially if all of the other flags imply server=true. I need to think through it more

The only reason I can think of to keep `--server` is if you want to use a server, but you want to enter `sessionId` and `passphrase` through prompts 🤷

PR #5521, opened 2024-10-09
multisig broker client handles session not found

hughy:
## Summary

defines error ids for error messages from multisig broker server

client throws SessionIdNotFound error when it receives an error message from the server with the matching error id

Closes IFL-3051

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


hughy:
Closing: error 'id' should be message id so that client can cancel retries; error code should be separate

PR #5540, opened 2024-10-11
multisig broker server disconnects on invalid message

hughy:
## Summary

if the broker server receives a message that it cannot parse then it should disconnect from the client that sent the message

the server cannot send an 'ack' message to the client if it cannot parse the message and cannot send an error message with the message id of the errant message, so the client may continue retrying the malformed message if the server does not disconnect

for now the server does not implement client banning

Closes IFL-3062

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/multisigBroker/server.ts
+++ b/ironfish-cli/src/multisigBroker/server.ts
@@ -186,7 +186,8 @@ export class MultisigServer {
       )
 
       if (parseError) {
-        this.sendErrorMessage(client, 0, `Error parsing message`)
+        client.close(parseError)
+        this.clients.delete(client.id)
         return
       }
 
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/multisigBroker/server.ts
```diff
@@ -186,7 +186,8 @@ export class MultisigServer {
       )
 
       if (parseError) {
-        this.sendErrorMessage(client, 0, `Error parsing message`)
+        client.close(parseError)
```
mat-if, line 189: What do you think about adding a debug log here for potential future debugging?

mat-if (approved):

ironfish-cli/src/multisigBroker/server.ts
hughy: I think that's a great idea!

mat-if (approved):

PR #5547, opened 2024-10-14
Clear identities when a user disconnects during the identity collection phase

mat-if:
## Summary

If you disconnect and re-connect  during the identity phase, and add/remove --ledger flag or otherwise do something that would change the identity, the DKG process will keep the old identity around and move forward to the next stage, making it impossible to complete the process.

We can safely delete the identity from the session when a user disconnects so long as it happens during the identity collection phase. If it has moved past that phase, there are more implications, so I did not handle it in this PR.

This PR also opens the door for other improvements by adding the identity to the server client.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/multisigBroker/server.ts
+++ b/ironfish-cli/src/multisigBroker/server.ts
@@ -307,6 +307,34 @@ export class MultisigServer {
       return
     }
 
+    // If the session is in the collecting identities phase, we can safely
+    // remove a disconnected client's identity from the list. Otherwise, if a
+    // client disconnects and then reconnects, they can possibly be counted as
+    // multiple identities, leaving the session in a bad state.
+    if (client.identity != null) {
+      if (isDkgSession(session)) {
+        if (
+          session.status.round1PublicPackages.length === 0 &&
+          session.status.round2PublicPackages.length === 0
+        ) {
+          const identIndex = session.status.identities.indexOf(client.identity)
+          if (identIndex > -1) {
+            session.status.identities.splice(identIndex, 1)
+          }
+        }
+      } else if (isSigningSession(session)) {
+        if (
+          session.status.signatureShares.length === 0 &&
+          session.status.signingCommitments.length === 0
+        ) {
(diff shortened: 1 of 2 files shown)


ironfish-cli/src/multisigBroker/server.ts
```diff
@@ -307,6 +307,34 @@ export class MultisigServer {
...
 
+    // If the session is in the collecting identities phase, we can safely
+    // remove a disconnected client's identity from the list. Otherwise, if a
+    // client disconnects and then reconnects, they can possibly be counted as
+    // multiple identities, leaving the session in a bad state.
+    if (client.identity != null) {
+      if (isDkgSession(session)) {
+        if (
+          session.status.round1PublicPackages.length === 0 &&
+          session.status.round2PublicPackages.length === 0
```
hughy, line 318: nit - we could probably get away with just checking `round1PublicPackages` here

this does make me wonder if we want the session to act more like a state machine and track which round/step it's on. we might be able to use that to force retries of round1, too if we blocked round1 package submissions if the state doesn't match

ironfish-cli/src/multisigBroker/server.ts
```diff
@@ -307,6 +307,34 @@ export class MultisigServer {
...
+        ) {
+          const identIndex = session.status.identities.indexOf(client.identity)
+          if (identIndex > -1) {
+            session.status.identities.splice(identIndex, 1)
+          }
+        }
+      } else if (isSigningSession(session)) {
+        if (
+          session.status.signatureShares.length === 0 &&
+          session.status.signingCommitments.length === 0
```
hughy, line 328: nit - should we check only `signingCommitments`?

hughy (approved):
there could be some race conditions here with a client disconnecting after enough identities have been submitted, but just before the first round1 package comes in

I think this is an improvement though in that it will help to prevent one of the error cases we ran into: accidentally starting with one identity before switching to a different one

PR #5554, opened 2024-10-16
adds allowed identities set to multisig server sessions

hughy:
## Summary

server blocks any client from joining the session if its identity is not in the set of allowed identities

sends the (encrypted) list of account identities to the server when starting a signing session. guards against user accidentally joining signing session with the wrong account

throws an error when joining the session with an identity that is not in the allowed set (does not retry)

allowedIdentities is optional for dkg sessions, but there's not a clear way yet of specifying this set

Closes IFL-3074

## Testing Plan
- tried to join session with the wrong account/identity
- joined session and signed transaction with correct identity

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/multisigBroker/clients/client.ts
+++ b/ironfish-cli/src/multisigBroker/clients/client.ts
@@ -178,11 +178,18 @@ export abstract class MultisigClient {
     maxSigners: number,
     minSigners: number,
     identity: string,
+    allowedIdentities?: string[],
   ): void {
     this.sessionId = uuid()
     this.passphrase = passphrase
     const challenge = this.key.encrypt(Buffer.from('DKG')).toString('hex')
-    this.send('dkg.start_session', { maxSigners, minSigners, challenge, identity })
+    this.send('dkg.start_session', {
+      maxSigners,
+      minSigners,
+      challenge,
+      identity,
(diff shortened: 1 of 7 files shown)


ironfish-cli/src/multisigBroker/clients/client.ts
```diff
@@ -178,23 +178,37 @@ export abstract class MultisigClient {
     maxSigners: number,
     minSigners: number,
     identity: string,
+    allowedIdentities?: string[],
```
mat-if, line 181: It feels a _little_ strange that the DKG functions take these params even though they are technically unsupported. I'm not sure if I think that's better, or if the DKG flow functions should not take the param at all.

mat-if (approved):

ironfish-cli/src/multisigBroker/clients/client.ts
hughy: I think the only reason to have `allowedIdentities` in DKG is to try to prevent one person from joining the session more than once. 🤔 

In practice I think it's going to be difficult to provide a good experience for setting the list of allowed identities for DKG, so I'm going to remove it from the DKG methods for now

PR #5557, opened 2024-10-17
threadsafe sapling params & provide convenience method for sapling load

jowparks:
## Summary

Convenience method for sapling load, pull from `ironfish-optimize`
Load a sapling object configured to a BLS12 jubjub curve. This is currently
the only pairing for which a jubjub curve has been defined, and is the
default implementation.

Provided as a convenience method so clients don't have to depend
explicitly on zcash_primitives just to define a JubjubBls12 point.
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/sapling_bls12.rs
+++ b/ironfish-rust/src/sapling_bls12.rs
@@ -4,11 +4,16 @@
 
 use crate::Sapling;
 use lazy_static::lazy_static;
+use std::sync::Arc;
 
 pub use blstrs::Scalar;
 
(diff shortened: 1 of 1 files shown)


jowparks:
From Andrea:
```
The parameters are read-only (always accessed behind & references), so why is Arc needed here?
Arc is needed to implement shared mutable references, but shared read-only references are safe by default.
```

```
some syntax sugar is needed (such as using &*SAPLING instead of SAPLING)
```

closing

PR #5561, opened 2024-10-17
hex encode/decode pushed to ironfish-zkp

jowparks:
## Summary
Moves `hex_to_bytes`, `bytes_to_hex`, and `hex_to_vec_bytes` to `ironfish-zkp` so that we can use it for ser/de of structs there too.

As `ironfish-zkp` is dependency of `ironfish-rust`, this will be available in both crates now.
## Testing Plan
tests pass
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/errors.rs
+++ b/ironfish-rust/src/errors.rs
@@ -33,6 +33,7 @@ pub enum IronfishErrorKind {
     FailedXChaCha20Poly1305Decryption,
     FailedXChaCha20Poly1305Encryption,
     FailedHkdfExpansion,
+    HexError,
     IllegalValue,
     InconsistentWitness,
     InvalidAssetIdentifier,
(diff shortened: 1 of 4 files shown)


ironfish-rust/src/lib.rs
```diff
@@ -25,6 +25,7 @@ pub mod xchacha20poly1305;
 pub use {
     ironfish_frost::frost,
     ironfish_frost::participant,
+    ironfish_zkp::hex::{bytes_to_hex, hex_to_bytes, hex_to_vec_bytes},
```
andiflabs, line 28: Instead of putting this line here, you could put it in `ironfish-rust/src/serializing/mod.rs`, so that you don't need to change all the imports everywhere.

andiflabs (changes requested):

andiflabs (approved):

PR #5564, opened 2024-10-18
handles reconnect/rejoin status in session managers

hughy:
## Summary

adds 'onDisconnected' event emitter to multisig broker client

uses 'onDisconnected' events in sessionManager to wait from confirmation that client has reconnected to server and rejoined session

by using 'waitForConnectedMessage' and 'waitForJoinedSession' we display ux status output to the user _and_ throw an error if the session can't be rejoined

updates session managers to conditionally start ux actions in polling loops. this makes it so that the session manager won't clear the ux status output from waiting to reconnect or rejoin a session while polling for session status

Closes IFL-3090

## Testing Plan
1. start server using 'wallet:multisig:server'
2. connect to server for dkg session with 'wallet:multisig:dkg:create --hostname localhost', wait until session joined
3. stop and restart server process
4. wait for 'dkg:create' command to reconnect and then throw an error because the session id isn't found on the new server

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/multisigBroker/clients/client.ts
+++ b/ironfish-cli/src/multisigBroker/clients/client.ts
@@ -59,6 +59,7 @@ export abstract class MultisigClient {
   private tryConnectUntil: number | null = null
 
   readonly onConnected = new Event<[]>()
+  readonly onDisconnected = new Event<[]>()
   readonly onDkgStatus = new Event<[DkgStatusMessage]>()
   readonly onSigningStatus = new Event<[SigningStatusMessage]>()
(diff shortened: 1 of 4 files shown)


hughy:
Small UX thing to take a look at:
image
When waiting on reconnecting or rejoining the `ux.action.status =` updates in the polling loops update the status, which is a bit confusing

mat-if (approved):
The client change will need to be propagated to the new repo, I think, but looks good.

PR #5565, opened 2024-10-18
Circuit struct serialization

jowparks:
## Summary
Adds circuit struct ser/de so that downstream consumers (ie oreowallet) can interact with these primitives directly.

ser/de used here currently for external proving service:
https://github.com/oreoslabs/oreowallet-mono/blob/developer/crates/prover/src/handlers.rs

Seemed reasonable to move this into `ironfish-zkp`
## Testing Plan
tests pass
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-zkp/src/circuits/mint_asset.rs
+++ b/ironfish-zkp/src/circuits/mint_asset.rs
@@ -15,7 +15,7 @@ use crate::{
 };
 use byteorder::{ReadBytesExt, WriteBytesExt};
 
-use super::util::Reader;
+use super::util::FromBytes;
 
 pub struct MintAsset {
     /// Key required to construct proofs for a particular spending key
@@ -28,13 +28,13 @@ pub struct MintAsset {
 
 impl MintAsset {
     pub fn write<W: Write>(&self, mut writer: W) -> std::io::Result<()> {
-        if let Some(proof_generation_key) = &self.proof_generation_key {
+        if let Some(ref proof_generation_key) = self.proof_generation_key {
             writer.write_u8(1)?;
             writer.write_all(proof_generation_key.to_bytes().as_ref())?;
         } else {
             writer.write_u8(0)?;
         }
-        if let Some(public_key_randomness) = &self.public_key_randomness {
+        if let Some(ref public_key_randomness) = self.public_key_randomness {
             writer.write_u8(1)?;
             writer.write_all(public_key_randomness.to_bytes().as_ref())?;
         } else {
@@ -242,27 +242,24 @@ mod test {
         let deserialized_mint_asset = MintAsset::read(&buffer[..]).unwrap();
 
         assert_eq!(
-            mint_asset.proof_generation_key.is_some(),
-            deserialized_mint_asset.proof_generation_key.is_some()
+            mint_asset.proof_generation_key.clone().unwrap().ak,
+            deserialized_mint_asset
+                .proof_generation_key
+                .clone()
+                .unwrap()
+                .ak
         );
         assert_eq!(
-            mint_asset.public_key_randomness.is_some(),
-            deserialized_mint_asset.public_key_randomness.is_some()
+            mint_asset.proof_generation_key.clone().unwrap().nsk,
+            deserialized_mint_asset
+                .proof_generation_key
+                .clone()
+                .unwrap()
+                .nsk
         );
-
-        if let (Some(pk1), Some(pk2)) = (
-            &mint_asset.proof_generation_key,
-            &deserialized_mint_asset.proof_generation_key,
-        ) {
-            assert_eq!(pk1.ak, pk2.ak);
-            assert_eq!(pk1.nsk, pk2.nsk);
-        }
-
-        if let (Some(r1), Some(r2)) = (
-            &mint_asset.public_key_randomness,
-            &deserialized_mint_asset.public_key_randomness,
-        ) {
-            assert_eq!(r1, r2);
-        }
+        assert_eq!(
+            mint_asset.public_key_randomness.unwrap(),
+            deserialized_mint_asset.public_key_randomness.unwrap()
+        );
     }
 }
--- a/ironfish-zkp/src/circuits/mint_asset.rs
+++ b/ironfish-zkp/src/circuits/mint_asset.rs
@@ -3,6 +3,7 @@ use bellperson::{
     Circuit,
 };
 use ff::PrimeField;
+use std::io::{Read, Write};
 use zcash_proofs::{
     circuit::ecc,
     constants::{PROOF_GENERATION_KEY_GENERATOR, SPENDING_KEY_GENERATOR},
@@ -12,6 +13,9 @@ use crate::{
     constants::{proof::PUBLIC_KEY_GENERATOR, CRH_IVK_PERSONALIZATION},
     ProofGenerationKey,
 };
+use byteorder::{ReadBytesExt, WriteBytesExt};
+
+use super::util::Reader;
 
 pub struct MintAsset {
     /// Key required to construct proofs for a particular spending key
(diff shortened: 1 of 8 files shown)


ironfish-zkp/src/circuits/mint_asset.rs
```diff
@@ -22,6 +26,39 @@ pub struct MintAsset {
...
 
+impl MintAsset {
+    pub fn write<W: Write>(&self, mut writer: W) -> std::io::Result<()> {
+        if let Some(proof_generation_key) = &self.proof_generation_key {
+            writer.write_u8(1)?;
+            writer.write_all(proof_generation_key.to_bytes().as_ref())?;
+        } else {
+            writer.write_u8(0)?;
+        }
+        if let Some(public_key_randomness) = &self.public_key_randomness {
```
andiflabs, line 37: Ditto

ironfish-zkp/src/circuits/mint_asset.rs
```diff
@@ -181,4 +218,51 @@ mod test {
...
+            assert_eq!(pk1.ak, pk2.ak);
+            assert_eq!(pk1.nsk, pk2.nsk);
+        }
+
+        if let (Some(r1), Some(r2)) = (
+            &mint_asset.public_key_randomness,
+            &deserialized_mint_asset.public_key_randomness,
+        ) {
+            assert_eq!(r1, r2);
+        }
```
andiflabs, line 266: This can be rewritten in a more compact way as:

```rust
assert_eq!(mint_asset.proof_generation_key.unwrap().ak, deserialized_mint_asset.proof_generation_key.unwrap().ak);
assert_eq!(mint_asset.proof_generation_key.unwrap().nsk, deserialized_mint_asset.proof_generation_key.unwrap().nsk);
assert_eq!(mint_asset.public_key_randomness.unwrap(), deserialized_mint_asset.public_key_randomness.unwrap());
````

File: ironfish-zkp/src/circuits/output.rs [hunk 3]
```diff
@@ -49,6 +52,87 @@ pub struct Output {
     pub ar: Option<jubjub::Fr>,
 }
 
+impl Output {
+    pub fn write<W: Write>(&self, mut writer: W) -> std::io::Result<()> {
+        if let Some(value_commitment) = &self.value_commitment {
```
andiflabs, line 57: Same comment as above (`if let Some(ref value_commitment) = ... `)

ironfish-zkp/src/circuits/output.rs
```diff
@@ -366,4 +454,92 @@ mod test {
...
+            // Ser/de
+            let mut writer = vec![];
+            output.write(&mut writer).unwrap();
+            let deserialized_output: Output = Output::read(&writer[..]).unwrap();
+
+            assert_eq!(
+                output.value_commitment.is_some(),
+                deserialized_output.value_commitment.is_some()
+            );
+            if let Some(ref value_commitment) = output.value_commitment {
```
andiflabs, line 510: Same comment as above: instead of checking `is_some()` and then using an `if` (whose condition will always be true), use `unwrap()` directly.

ironfish-zkp/src/circuits/spend.rs
```diff
@@ -50,6 +54,127 @@ pub struct Spend {
...
+        } else {
+            writer.write_u8(0)?;
+        }
+        if let Some(ar) = &self.ar {
+            writer.write_u8(1)?;
+            writer.write_all(ar.to_bytes().as_ref())?;
+        } else {
+            writer.write_u8(0)?;
+        }
+        writer.write_all((self.auth_path.len() as u32).to_le_bytes().as_ref())?;
```
andiflabs, line 89: Is it possible that `auth_path` does not fit in a `u32`?

ironfish-zkp/src/circuits/spend.rs
```diff
@@ -50,6 +54,127 @@ pub struct Spend {
...
+                        res.push(flag);
+                    }
+                    None => {
+                        res.push(0);
+                    }
+                }
+                res
+            })
+            .collect::<Vec<u8>>();
+        writer.write_all(bytes.as_ref())?;
```
andiflabs, line 109: This can potentially perform several allocations, which are not necessary. A traditional `for` loop can avoid that:

```rust
for auth_path in &self.auth_path {
    match auth_path {
        Some((val, flag)) => { writer.write_u8(1)?; writer.write_all(val.to_bytes_le())?; writer.write_u8(*flag as u8); }
        None => writer.write_u8(0)?;
    }
}
```

ironfish-zkp/src/circuits/util.rs
```diff
@@ -143,3 +144,35 @@ pub fn assert_valid_asset_generator<CS: bellperson::ConstraintSystem<blstrs::Sca
 
     Ok(())
 }
+
+pub trait Reader: Sized {
+    fn read<R: std::io::Read>(reader: &mut R) -> Result<Self, std::io::Error>;
```
andiflabs, line 149: `reader: &mut R` -> `mut reader: R`

ironfish-zkp/src/circuits/util.rs
```diff
@@ -143,3 +144,35 @@ pub fn assert_valid_asset_generator<CS: bellperson::ConstraintSystem<blstrs::Sca
 
     Ok(())
 }
+
+pub trait Reader: Sized {
```
andiflabs, line 148: Why is this trait needed exactly?

Also, I'd suggest a name like `FromBytes` or similar, to convey the message that this is a trait that is used to _construct_ structs.

ironfish-zkp/src/primitives/proof_generation_key.rs
```diff
@@ -68,6 +68,13 @@ impl ProofGenerationKey {
         Ok(ProofGenerationKey(ZcashProofGenerationKey { ak, nsk }))
     }
 
+    pub fn read<R: std::io::Read>(reader: &mut R) -> Result<Self, std::io::Error> {
```
andiflabs, line 71: `reader: &mut R` -> `mut reader: R`

ironfish-zkp/src/primitives/value_commitment.rs
```diff
@@ -26,6 +29,29 @@ impl ValueCommitment {
...
+
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut res = vec![];
+        res.extend(self.value.to_le_bytes());
+        res.extend(self.randomness.to_bytes());
+        res.extend(self.asset_generator.to_bytes());
+        res
+    }
+
+    pub fn read<R: std::io::Read>(reader: &mut R) -> Result<Self, std::io::Error> {
```
andiflabs, line 41: `reader: &mut R` -> `mut reader: R`

ironfish-zkp/src/primitives/value_commitment.rs
```diff
@@ -26,6 +29,29 @@ impl ValueCommitment {
         (self.asset_generator.clear_cofactor() * jubjub::Fr::from(self.value))
             + (*VALUE_COMMITMENT_RANDOMNESS_GENERATOR * self.randomness)
     }
+
+    pub fn to_bytes(&self) -> Vec<u8> {
```
andiflabs, line 33: Why not having a `write()` like other structs?

Also, I might be wrong, but I think the bytes representation of this struct is fixed-size, so this could return an array instead of allocating a vec

ironfish-zkp/src/circuits/mint_asset.rs
```diff
@@ -22,6 +26,39 @@ pub struct MintAsset {
     pub public_key_randomness: Option<jubjub::Fr>,
 }
 
+impl MintAsset {
+    pub fn write<W: Write>(&self, mut writer: W) -> std::io::Result<()> {
+        if let Some(proof_generation_key) = &self.proof_generation_key {
```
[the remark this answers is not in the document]
jowparks: Nice, thanks for the tip. I wouldn't have guessed it would be copying here, thought it would just be unwrapping the optional. No harm in updating to this syntax

ironfish-zkp/src/circuits/util.rs
jowparks: Why is this way preferable? I was reading that the way you suggest takes ownership, whereas the current way is borrowed, is it better in the case of readers to take ownership?

ironfish-zkp/src/circuits/util.rs
jowparks: How else would you do it without a trait extension? These are externally owned types. If I try to define an implementation directly it gives me a compiler error.

ironfish-zkp/src/circuits/spend.rs
jowparks: Seems you are right, this would be a bug. Looking at some of our typescript:
https://github.com/iron-fish/ironfish/blob/ca60831b8556978164e7f66a3d5359a59fbc3609/ironfish/src/primitives/rawTransaction.ts#L284

ironfish-zkp/src/circuits/util.rs
andiflabs: `Read` is already automatically implemented for `&mut R`: https://doc.rust-lang.org/std/io/trait.Read.html#impl-Read-for-%26mut+R. This means that if you have a method with signature `fn foo<R: Read>(reader: R) -> ...`, both `foo(reader)` and `foo(&mut reader)` are accepted. Conversely, if you have `fn foo<R: Read>(reader: &mut R) -> ...`, only `foo(&mut reader)` is accepted. Therefore, accepting `R` allows opening the door to more types.

In the particular case of `Read` (or `Write`), this is especially useful because readers (or writers) may be wrappers that contain references to other structs, so requiring a `&mut R` often results in a reference to a reference, and this effect is amplified when adding layers of wrappers.

ironfish-zkp/src/circuits/util.rs
andiflabs: I see: you're trying to add some utility methods to the external structs, so that the rest of the code can be shorter (I missed this when I first read the code). It makes senses then. I would just not make this trait public, but rather change its visibility to `pub(crate)`.

andiflabs (approved):

PR #5568, opened 2024-10-21
renderSpan: add more time intervals

j-s-n:
## Summary
Added years, months, and days to renderSpan
Estimated years to 365 days and months to 30 days.

## Testing Plan
Lock and unlock a wallet with flags for various time intervals.
Confirmed it returns the highest 2 time units (year & month, month & day, etc.)

## Documentation
N/A

## Breaking Change
N/A

### DIFF
--- a/ironfish/src/utils/time.ts
+++ b/ironfish/src/utils/time.ts
@@ -2,11 +2,15 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+import { tr } from 'date-fns/locale'
 import { MathUtils } from './math'
 
 const MS_PER_SEC = 1000.0
 const MS_PER_MIN = 60.0 * 1000.0
 const MS_PER_HOUR = 60.0 * 60.0 * 1000.0
+const MS_PER_DAY = 24 * 60.0 * 60.0 * 1000.0
+const MS_PER_MONTH = 30 * 24 * 60.0 * 60.0 * 1000.0
+const MS_PER_YEAR = 12 * 30 * 24 * 60.0 * 60.0 * 1000.0
 
 /**
  *
@@ -31,6 +35,9 @@ const renderEstimate = (done: number, total: number, speed: number): string => {
     forceSecond: true,
     forceMinute: true,
     forceHour: true,
+    forceDay: true,
+    forceMonth: true,
+    forceYear: true,
     hideMilliseconds: true,
   })
 }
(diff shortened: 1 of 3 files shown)


ironfish/src/utils/time.ts
```diff
@@ -58,6 +68,29 @@ const renderSpan = (
 
   const parts = []
   let magnitude = 0
+  // Do we want to include years? Is it ok to estimate at 365 days?
```
NullSoldier, line 71: Yes I think it's ok to include years and to estimate at 365 days

ironfish/src/utils/time.ts
```diff
@@ -58,6 +68,29 @@ const renderSpan = (
...
+  if (time >= MS_PER_YEAR && (magnitude <= 8 || options?.forceYear)) {
+    const years = Math.floor(time / MS_PER_YEAR)
+    time -= years * MS_PER_YEAR
+    parts.push(`${years.toFixed(0)}y`)
+    magnitude = Math.max(magnitude, 7)
+  }
+
+  // Do we want to include months? Is it ok to estimate at 30 days?
```
NullSoldier, line 79: Yes, I think it's ok to include months and to estimate at 30 days a month.

NullSoldier:
I think we should add a few tests in the test file for these new cases.

ironfish/src/utils/time.ts
j-s-n: Update year calculation to use 365 days.

ironfish/src/utils/time.ts
j-s-n: Kept month calculation at 30 days and updated month abbreviation to 'M'. capital M is commonly used in other libraries (e.g. y M d h m s ms)

j-s-n:
> I think we should add a few tests in the test file for these new cases.

added tests for renderSpan and renderEstimate. Updated renderEstimate to return largest 3 intervals (y, M, d) for longer periods and renderSpan outputs largest 2 intervals (e.g. y, M / M, d / d, h)

NullSoldier:
Ok the PR looks good. Assuming the tests pass we should merge this in.

NullSoldier (approved):

ironfish/src/utils/time.ts
```diff
@@ -58,6 +67,26 @@ const renderSpan = (
 
   const parts = []
   let magnitude = 0
```
NullSoldier, line 69:
```suggestion
  let magnitude = 0

```

PR #5569, opened 2024-10-21
Add createdAt to createAccount RPC

danield9tqh:
## Summary
Allow account creation to take a createdAt object

## Testing Plan
Unit tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/create.test.slow.ts
+++ b/ironfish/src/rpc/routes/wallet/create.test.slow.ts
@@ -10,7 +10,7 @@ import { createRouteTest } from '../../../testUtilities/routeTest'
 import { RPC_ERROR_CODES } from '../../adapters'
 import { RpcRequestError } from '../../clients/errors'
 
-describe('Route wallet/create', () => {
+describe('Route wallet/createAccount', () => {
   jest.setTimeout(15000)
   const routeTest = createRouteTest()
 
@@ -20,6 +20,10 @@ describe('Route wallet/create', () => {
 
   it('should create an account', async () => {
     await routeTest.node.wallet.createAccount('existingAccount', { setDefault: true })
+    const createdAtHead = {
+      hash: routeTest.node.chain.head.hash,
+      sequence: routeTest.node.chain.head.sequence,
+    }
 
     const name = uuid()
 
@@ -35,6 +39,7 @@ describe('Route wallet/create', () => {
     expect(account).toMatchObject({
       name: name,
       publicAddress: response.content.publicAddress,
+      createdAt: createdAtHead,
     })
   })
 
@@ -87,5 +92,35 @@ describe('Route wallet/create', () => {
     })
 
     expect(scanSpy).toHaveBeenCalled()
+  })
+
+  it('should set account createdAt if passed', async () => {
+    const name = uuid()
+    const createdAt = {
+      hash: '00000000000163b1c632498ac8f86cfd33f65c5d4e43e5001c75d5acb4ada4de',
+      sequence: 10,
+    }
+
+    const response = await routeTest.client.wallet.createAccount({
+      name,
+      createdAt,
+    })
+
+    expect(response.status).toBe(200)
+    expect(response.content).toMatchObject({
+      name: name,
+      publicAddress: expect.any(String),
+      isDefaultAccount: true,
+    })
+
+    const account = routeTest.node.wallet.getAccountByName(name)
+    expect(account).toMatchObject({
+      name: name,
+      publicAddress: response.content.publicAddress,
+      createdAt: {
+        hash: Buffer.from(createdAt.hash, 'hex'),
+        sequence: createdAt.sequence,
+      },
+    })
   })
 })
--- a/ironfish/src/rpc/routes/wallet/create.test.slow.ts
+++ b/ironfish/src/rpc/routes/wallet/create.test.slow.ts
@@ -96,10 +96,7 @@ describe('Route wallet/createAccount', () => {
 
   it('should set account createdAt if passed', async () => {
     const name = uuid()
-    const createdAt = {
-      hash: '00000000000163b1c632498ac8f86cfd33f65c5d4e43e5001c75d5acb4ada4de',
-      sequence: 10,
-    }
+    const createdAt = 10
 
     const response = await routeTest.client.wallet.createAccount({
       name,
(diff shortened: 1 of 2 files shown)


ironfish/src/rpc/routes/wallet/createAccount.ts
```diff
@@ -45,10 +59,23 @@ routes.register<typeof CreateAccountRequestSchema, CreateAccountResponse>(
   CreateAccountRequestSchema,
   async (request, context): Promise<void> => {
     AssertHasRpcContext(request, context, 'wallet')
+    if (
+      request.data.createdAt?.hash === undefined ||
+      request.data.createdAt?.sequence === undefined
+    ) {
+      request.data.createdAt = undefined
```
hughy, line 66: will yup validation enforce both fields being defined this if we add `defined()` to the `hash` and `sequence` fields?

ironfish/src/rpc/routes/wallet/createAccount.ts
```diff
@@ -18,7 +18,15 @@ import { AssertHasRpcContext } from '../rpcContext'
...
  */
 
-export type CreateAccountRequest = { name: string; default?: boolean }
+export type CreateAccountRequest = {
+  name: string
+  default?: boolean
+  createdAt?: {
+    hash: string
+    sequence: number
+  }
```
hughy, line 27: we're in the process of deprecating `hash` as part of `createdAt`. the `importAccount` RPC only uses a number for `createdAt`. should we do the same here?

ironfish/src/rpc/routes/wallet/createAccount.ts
```diff
@@ -45,10 +59,23 @@ routes.register<typeof CreateAccountRequestSchema, CreateAccountResponse>(
   CreateAccountRequestSchema,
   async (request, context): Promise<void> => {
     AssertHasRpcContext(request, context, 'wallet')
+    if (
+      request.data.createdAt?.hash === undefined ||
+      request.data.createdAt?.sequence === undefined
+    ) {
+      request.data.createdAt = undefined
+    }
```
NullSoldier, line 67: You can delete this entire block of code according to your spec above.

ironfish/src/rpc/routes/wallet/createAccount.ts
```diff
@@ -45,10 +59,23 @@ routes.register<typeof CreateAccountRequestSchema, CreateAccountResponse>(
...
+
+    const createdAt = request.data.createdAt && {
+      hash: Buffer.from(request.data.createdAt.hash, 'hex'),
+      sequence: request.data.createdAt.sequence,
+    }
+
+    const setDefault = !context.wallet.hasDefaultAccount || (request.data.default ?? false)
```
NullSoldier, line 74: random nit: but I prefer small refactors like this split out to other PRs. This is not related to the PR.

NullSoldier: Also because now that this code changed, the code is actually very strange that the RPC is auto setting the default account if one is not set when an account is created. This should absolutely not be happening and should be removed in another PR.

ironfish/src/rpc/routes/wallet/createAccount.ts
danield9tqh: The yup spec? Yup doesn't allow required fields nested in an optional object. So if we passed in the following
```
{ "name": "account1" }
```
the request object `data` would come in as
```
{
  "name": "account1"
  "createdAt": {
    hash: undefined
    sequence: undefined
  }
}
```

I could probably make this code a little cleaner but it would be more complex if we actually wanted to validate that if you pass in `createdAt` field that both `hash` and `sequence` are defined. This is one of the main problems with yup I've run into. It gets even more difficult if you want to make `createdAt` nullable (which it techincally should be since you can set `createdAt` to null in the wallet)

ironfish/src/rpc/routes/wallet/createAccount.ts
danield9tqh: I can move this refactor to another PR. I'm still confused on how it changes the functionality though. In the previous code if there was no default account before this RPC call, it would set this account as the default. Seems the same as what this code is doing

NullSoldier: There's no change in functionality here. I re-read previous comments and don't see any mention of that. That's the issue I'm mentioning, it's a refactor that doesn't change functionality for an orthagonal piece of code unrelated to this change. (using the internal default setting rather than duplicating it)

ironfish/src/rpc/routes/wallet/create.test.slow.ts
```diff
@@ -88,4 +93,34 @@ describe('Route wallet/create', () => {
...
+    expect(account).toMatchObject({
+      name: name,
+      publicAddress: response.content.publicAddress,
+      createdAt: {
+        hash: Buffer.from(createdAt.hash, 'hex'),
+        sequence: createdAt.sequence,
+      },
+    })
+  })
 })
```
NullSoldier, line 126: This file was for testing /create but someone changed it to use the new RPC route which means the old /create isn't being tested. On top of that, the file name is now wrong.

ironfish/src/rpc/routes/wallet/createAccount.ts
NullSoldier: I pulled the PR down and it works just fine for me, you just have the YUP schema defined incorrectly. For optional sub fields you need this syntax.

```
export const CreateAccountRequestSchema: yup.ObjectSchema<CreateAccountRequest> = yup
  .object({
    name: yup.string().defined(),
    default: yup.boolean().optional(),
    createdAt: yup
      .object({
        hash: yup.string().defined(),
        sequence: yup.number().defined(),
      })
      .notRequired()
      .default(undefined),
  })
  .defined()
```

Once you do this, you can delete the entire block of code setting the createdAt as undefined.

ironfish/src/rpc/routes/wallet/createAccount.ts
danield9tqh: So shouldn't make a difference if we just set it to an empty buffer? `Buffer.alloc(32, 0)`

ironfish/src/rpc/routes/wallet/create.test.slow.ts
NullSoldier: Ok we're going to fix this in another PR.

NullSoldier (approved):

hughy (approved):

PR #5570, opened 2024-10-21
Add zod routes to RPC

danield9tqh:
## Summary
Add ability to implement zod routes

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/package.json
+++ b/ironfish/package.json
@@ -49,7 +49,8 @@
     "sqlite3": "5.1.6",
     "uuid": "8.3.2",
     "ws": "8.12.1",
-    "yup": "0.29.3"
+    "yup": "0.29.3",
+    "zod": "3.23.8"
   },
   "scripts": {
     "build": "tsc -b tsconfig.build.json",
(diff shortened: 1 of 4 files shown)


dguenther (approved):

NullSoldier:
Can you explain what the purpose of this is? It's just changing a schema library to a new one without any explanation.

This is a sort of auto-reject PR without further explanation if the goal is just to change libraries for the sake of change.

1. What does this random other library do that our current one doesn't do?
2. What issues are developers facing that this solves?

danield9tqh:
The main downside of yup right now is it doesn't support some of the typescript types that we would ideally want to handle in our requests. Here's one example: https://github.com/iron-fish/ironfish/pull/5569/files#r1811059203 but there are others 
I've run into too. I can look through the code for some examples

Also
- the node app uses zod and we can re-use some schemas there between the two codebases
- zod works better with typescript in general we can use utils like z.infer<> to create the request type automatically from the schema.

NullSoldier:
> The main downside of yup right now is it doesn't support some of the typescript types that we would ideally want to handle in our requests. Here's one example: https://github.com/iron-fish/ironfish/pull/5569/files#r1811059203 but there are others I've run into too. I can look through the code for some examples
> 
> Also
> 
> * the node app uses zod and we can re-use some schemas there between the two codebases
> * zod works better with typescript in general we can use utils like z.infer<> to create the request type automatically from the schema.

I just commented on the linked example, which does not seem valid.

Also, yup has infer too but we explicitly don't use it for various reasons around ambiguety. I had issues with it back in the day but we're on 0.29.x and now it's at 1.0 so we could easily just upgrade and test how well infer has improved.

PR #5572, opened 2024-10-22
shard unit tests in CI for faster CI runs

mat-if:
## Summary

Shards the main test suite across 3 shards, and slow tests across 2 shards, resulting in a CI runtime reduction from around 15 minutes to around 6.5-7 minutes: https://github.com/iron-fish/ironfish/actions/runs/11486776859

## Testing Plan

Re-ran CI workflow multiple times across 2 branches

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,9 @@ jobs:
   test:
     name: Test
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        shard: [1,2]
 
     steps:
       - name: Check out Git repository
@@ -63,7 +66,7 @@ jobs:
         run: yarn --non-interactive --frozen-lockfile
 
       - name: Run tests
-        run: yarn test:coverage --maxWorkers=2 --workerIdleMemoryLimit=2000MB
+        run: yarn test:coverage --maxWorkers=2 --workerIdleMemoryLimit=2000MB --shard=${{ matrix.shard }}/${{ strategy.job-total }}
 
       - name: Check for missing fixtures
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -63,7 +66,7 @@ jobs:
         run: yarn --non-interactive --frozen-lockfile
 
       - name: Run tests
-        run: yarn test:coverage --maxWorkers=2 --workerIdleMemoryLimit=2000MB
+        run: yarn test:coverage --maxWorkers=2 --workerIdleMemoryLimit=2000MB --shard=${{ matrix.shard }}/${{ strategy.job-total }}
```
andiflabs, line 69: LGTM, but it's not clear to me where `strategy.job-total` comes from, or how it is configured

andiflabs (approved):

.github/workflows/ci.yml
mat-if: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#strategy-context

This has some documentation about the strategy context. It's not _ideal_, because if we for some reason add another matrix, for example for NodeJS versions, we will need to come up with a different solution here. At that point, we may have to build a custom script to manage it because it's not clear to me that github actions has a way to get the length of a specific matrix array or anything.

mat-if: Ok, came up with a slightly less brittle solution - defining the matrix jobs as '1/3', '2/3', '3/3', etc. It doesn't feel great but it's less likely to randomly blow up in the future.

PR #5589, opened 2024-10-28
feat(IFL-2710): Show default account in account list

jowparks:
## Summary
defaults to showing table header so that we can display default account column as well.
## Testing Plan
```
...
  {
    "name": "testnet genesis",
    "id": "b0557e1c-16e1-4f2b-b491-38e87422df58",
    "head": {
      "hash": "0000025a50df477c9699d4149558857b6955f84b0e7032b3880fd6612baa7076",
      "sequence": 738626,
      "inChain": true
    },
    "scanningEnabled": true,
    "viewOnly": false,
    "default": true
  },
```

```
 Account         Default
 ─────────────── ───────
 foo                    
 nodeapptest            
 testnetnodeapp         
 testnet                
 testnet genesis ✓      
 rahuldaniel            
 nodeappdkg             
 50a26d7134        
```

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/index.ts
+++ b/ironfish-cli/src/commands/wallet/index.ts
@@ -44,6 +44,10 @@ export class AccountsCommand extends IronfishCommand {
           header: 'Account',
           minWidth: 11,
         },
+        default: {
+          get: (row) => (row.default ? chalk.green('✓') : ''),
+          header: 'Default',
+        },
         viewOnly: {
           get: (row) => (row.viewOnly ? chalk.green('✓') : ''),
           header: 'View Only',
@@ -73,7 +77,6 @@ export class AccountsCommand extends IronfishCommand {
       {
         ...flags,
         printLine: this.log.bind(this),
-        'no-header': flags['no-header'] ?? !flags.extended,
       },
     )
 
--- a/ironfish/src/rpc/routes/wallet/serializers.ts
+++ b/ironfish/src/rpc/routes/wallet/serializers.ts
@@ -170,5 +170,6 @@ export async function serializeRpcAccountStatus(
       : null,
     scanningEnabled: account.scanningEnabled,
     viewOnly: !account.isSpendingAccount(),
+    default: wallet.getDefaultAccount()?.id === account.id,
   }
 }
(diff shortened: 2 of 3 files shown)


patnir (approved):
Wow this is so simple but so needed

mat-if (approved):

ironfish-cli/src/commands/wallet/index.ts
```diff
@@ -36,11 +36,14 @@ export class AccountsCommand extends IronfishCommand {
       return []
     }
 
+    const defaultAccount = await client.wallet.getDefaultAccount()
```
hughy, line 39: I think the extra request is OK, but we could avoid it if we add some `default` indicator to the response of `getAccountsStatus`

hughy (approved):

ironfish-cli/src/commands/wallet/index.ts
```diff
@@ -36,11 +36,14 @@ export class AccountsCommand extends IronfishCommand {
...
+    const defaultName = defaultAccount.content.account?.name
+
     ui.table(
       response.content.accounts,
       {
         name: {
-          get: (row) => row.name,
+          get: (row) => (row.name === defaultName ? `${row.name} (default)` : row.name),
```
andiflabs, line 46: Does this affect the JSON output too?

andiflabs: For JSON, we probably want to leave the name as is, and maybe add an extra field somewhere to point to the default account

ironfish-cli/src/commands/wallet/index.ts
jowparks: Yea I thought of that, but it felt a little dirty to added a default flag to all account responses and have all of the `default=false` and just the one to `default=true`

ironfish-cli/src/commands/wallet/index.ts
jowparks: ie add key
```
  {
    "name": "testnet genesis",
    "id": "b0557e1c-16e1-4f2b-b491-38e87422df58",
    "head": {
      "hash": "0000004b952bb2d7e7b0cb494d444961a05a928b839b7ba7a8707bd052073000",
      "sequence": 738608,
      "inChain": true
    },
    "scanningEnabled": true,
    "viewOnly": false
  },
```

andiflabs (approved):

PR #5591, opened 2024-10-28
chore(IFL-2707): conditionally logs scan starts and ends

hughy:
## Summary

only logs scan starts and ends in debug output if the start and end sequences differ

reduces noise from debug output when running a synced node where each block produces a new scan with equal start and end sequences

Closes IFL-2707

## Testing Plan
ran synced node in verbose node before and after applying changes

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/scanner/walletScanner.ts
+++ b/ironfish/src/wallet/scanner/walletScanner.ts
@@ -105,9 +105,12 @@ export class WalletScanner {
       this.state = scan
       unlock()
 
-      this.logger.debug(
-        `Scan starting from block ${scan.start.sequence} to ${scan.start.sequence}`,
-      )
+      const logScanState = scan.start.sequence !== scan.end.sequence
+      if (logScanState) {
+        this.logger.debug(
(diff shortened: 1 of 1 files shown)


ironfish/src/wallet/scanner/walletScanner.ts
```diff
@@ -105,9 +105,12 @@ export class WalletScanner {
       this.state = scan
       unlock()
```
jowparks, line 107: nit:
```
logScanState && this.logger.debug(
  `Scan starting from block ${scan.start.sequence} to ${scan.end.sequence}`,
)
```

hughy: eslint places `logScanState &&` on a separate line:
```
logScanState &&
        this.logger.debug(
          `Scan starting from block ${scan.start.sequence} to ${scan.end.sequence}`,
        )
```

I think the `if` statement is a little clearer, and it's only adding one line for the closing bracket `}`

PR #5597, opened 2024-10-30
Feature-gate all code in ironfish-rust that requires Sapling parameters

andiflabs:
## Summary

This commit adds a new `transaction-proofs` Cargo feature (enabled by default) to the `ironfish-rust` crate. When this feature is disabled, all the structs and methods that create or verify zero-knowledge proofs are also disabled, hence making the Sapling parameters not needed.

This way, developers have an option to use some of the `ironfish-rust` features (like transaction decryption) without having to deploy the Sapling parameters, which are several MBs in size.

When `transaction-proofs` is disabled, all the transaction builder structs (like `ProposedTransaction`, `SpendBuilder, `OutputBuilder`, `MintBuilder`, ...) and related methods are unavailable, because those structs/methods all generate proofs.

This commit also cleans up and reorganizes some code, in particular:

* renamed the `ironfish::transaction::utils` module to `ironfish::transaction::verify`, because this module only contains methods for verification;
* moved some of the transaction verification methods from `ironfish::transaction` to `ironfish::transaction::verify`;
* moved `ProposedTransaction` to a dedicated submodule;

## Testing Plan

* `cargo test -p ironfish`
* `cargo test -p ironfish --no-default-features`

## Documentation

N/A

## Breaking Change

May be a breaking change for dependent that use `default-features = false`.


ironfish-rust/Cargo.toml
```diff
@@ -24,9 +24,12 @@ workspace = true
 workspace = true
 
 [features]
```
jowparks, line 26: nit: should this be called `proving` or `proof`? I am not sure `proofing` is correct here

ironfish-rust/src/transaction/tests.rs
```diff
@@ -2,44 +2,52 @@
...
     },
 };
+#[cfg(feature = "transaction-proofing")]
 use ff::Field;
+#[cfg(feature = "transaction-proofing")]
 use group::GroupEncoding;
-use ironfish_frost::dkg::{round1 as round1_dkg, round2 as round2_dkg, round3 as round3_dkg};
-use ironfish_frost::participant::Secret;
+#[cfg(feature = "transaction-proofing")]
 use ironfish_frost::{
```
jowparks, line 30: Why is the frost stuff behind this flag? Is it just unused outside of proof context?

andiflabs: Yes, it's what you said: the frost tests require creating new transactions, which requires the feature to be enabled. If the feature is disabled, the tests cannot run, and these imports are unused.

ironfish-rust/Cargo.toml
andiflabs: Renamed to `transaction-proofs`

PR #5604, opened 2024-10-30
adds multisigAccount boolean to account status

patnir:
## Summary

signifies to the user that the account is a multisig account. This will be helpful in many usecases in the CLI and Node App.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/serializers.ts
+++ b/ironfish/src/rpc/routes/wallet/serializers.ts
@@ -12,6 +12,7 @@ import {
 } from '../../../wallet/exporter/multisig'
 import { AssetValue } from '../../../wallet/walletdb/assetValue'
 import { DecryptedNoteValue } from '../../../wallet/walletdb/decryptedNoteValue'
+import { isSignerMultisig } from '../../../wallet/walletdb/multisigKeys'
 import { TransactionValue } from '../../../wallet/walletdb/transactionValue'
 import {
   RpcAccountAssetBalanceDelta,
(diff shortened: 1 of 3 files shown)


ironfish/src/rpc/routes/wallet/serializers.ts
```diff
@@ -157,6 +158,9 @@ export async function serializeRpcAccountStatus(
   account: Account,
 ): Promise<RpcAccountStatus> {
   const head = await account.getHead()
+  const isMultisigAccount = account.multisigKeys
+    ? isSignerMultisig(account.multisigKeys)
+    : false
```
hughy, line 163: Should we also set this to true if `isHardwareSignerMultisig`? In the future we may also want to indicate that those accounts are Ledger accounts, but they're also multisig accounts

It's less clear to me how we should handle coordinator accounts 🤔

patnir: yeah i was thinking maybe we want a `type` field instead? but i think that might push into the domain of your work with jason

hughy: What would the different values of `type` be? Something like this?
```
- signer
- view-only
- multisig-signer
- multisig-coordinator
- ledger
- multisig-ledger
```

We're planning on refactoring how the wallet manages multisig identities, but I don't think we'll be adding or removing any types of accounts

PR #5606, opened 2024-10-30
logs warning if bootstrap node on different network

hughy:
## Summary

adds onMessage event handler to log a warning message if a bootstrap node is on a different network than the user's node

outputs the warning without requiring verbose output and specifies that the peer with the network mismatch is a bootstrap node

updates 'Peer.dispose' to use 'clearAfter' on 'onMessage' so that this second handler will still fire to log the warning even after disconnecting from the bootstrap node

Closes IFL-1217

## Testing Plan
run a testnet node and use the `--bootstrap` or `-b` flag to specify a mainnet bootstrap node, observe warning message
image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/network/peerNetwork.ts
+++ b/ironfish/src/network/peerNetwork.ts
@@ -17,6 +17,7 @@ import {
 import { PeerStore } from '../fileStores/peerStore'
 import { createRootLogger, Logger } from '../logger'
 import { MetricsMonitor } from '../metrics'
+import { renderNetworkName } from '../networks'
 import { FullNode } from '../node'
 import { IronfishPKG } from '../package'
 import { Platform } from '../platform'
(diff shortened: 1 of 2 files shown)


ironfish/src/network/peerNetwork.ts
```diff
@@ -368,11 +369,25 @@ export class PeerNetwork {
...
         host: url.hostname,
         port,
         whitelist: true,
       })
+
+      const warnNetworkMismatch = (message: NetworkMessage) => {
+        if (message instanceof IdentifyMessage) {
+          if (message.networkId !== this.networkId) {
+            this.logger.warn(
+              `Bootstrap node ${node} is on network ${message.networkId} while we are on network ${this.networkId}`,
```
mat-if, line 382: What do you think of printing the network name for the default networks? Similar to how we do it in the status command? https://github.com/iron-fish/ironfish/blob/master/ironfish-cli/src/commands/status.ts#L127-L128

hughy: I think that's an excellent idea!

mat-if (approved):

PR #5611, opened 2024-11-01
chore(IFL-2545): adds default limit for table output to 50

jowparks:
## Summary
- adds default limit for table output to 50
- prints message at end of table if limit is reached
- updates commands to handle limit
- Intentionally excluded table command updates for: multisig, genesis, and combine notes
- Did not update print to occur on csv/json

## Testing Plan
Tested all updated commands with `--limit 1`
```
     Timestamp                  Status       Type     Hash                                                             Expiration Submitted Sequence Fee Paid ($IRON) Asset                     Amount          
 ─── ────────────────────────── ──────────── ──────── ──────────────────────────────────────────────────────────────── ────────── ────────────────── ──────────────── ───────────────────────── ────────────────
     10/28/2024 2:45:48 PM PDT  confirmed    send     dc6116865d0e4e92df1e75c94afd951d6d2ca40eddd71b0900bcd270fa149bce 738792     738778             0.00000010       NORI              (f3b28) -1              
...
[see more rows by using --limit flag]
```
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/notes/index.ts
+++ b/ironfish-cli/src/commands/wallet/notes/index.ts
@@ -1,7 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-import { CurrencyUtils, RpcAsset } from '@ironfish/sdk'
+import { CurrencyUtils, RpcAsset, RpcWalletNote } from '@ironfish/sdk'
 import { Flags } from '@oclif/core'
 import { IronfishCommand } from '../../../command'
 import { RemoteFlags } from '../../../flags'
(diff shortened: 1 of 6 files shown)


ironfish-cli/src/commands/wallet/notes/index.ts
```diff
@@ -33,74 +33,74 @@ export class NotesCommand extends IronfishCommand {
...
     for await (const note of response.contentStream()) {
+      if (notes.length >= flags.limit) {
+        break
+      }
       if (!assetLookup.has(note.assetId)) {
         assetLookup.set(
           note.assetId,
           (await client.wallet.getAsset({ id: note.assetId, account })).content,
         )
       }
```
jowparks, line 46: These were rendering entire tables as individual rows. I updated to collect from the stream, not sure if this is desirable as at very large number of rows could cause memory issue (also output will not stream), but seemed like a reasonable tradeoff.

ironfish-cli/src/ui/table.ts
```diff
@@ -210,6 +216,9 @@ class Table<T extends Record<string, unknown>> {
       }
       this.options.printLine(` ${rowValues.join(' ')}`)
     }
+    if (this.options.limit && this.options.limit !== 0 && rows.length >= this.options.limit) {
```
andiflabs, line 219: Nit: I wonder if `this.options.limit` can be negative and, if so, if we should change `!== 0` to `<= 0`

andiflabs (approved):

ironfish-cli/src/ui/table.ts
jowparks: Wouldn't hurt to change it to this!

PR #5624, opened 2024-11-06
supports migrations on encrypted wallets

hughy:
## Summary

adds a template migration , 033-encrypted-wallet-template, to demonstrate how to write wallet migrations on wallet databases that may be encrypted

now that account data may be encrypted we need to be careful to decrypt the account before running the migration and to re-encrypt the migrated data before writing it back to the database

adds EncryptedWalletMigrationError thrown when a migration tries to access an encrypted account without the wallet passphrase to allow client code (e.g., 'migrations:start' command) to handle decryption flow

optionally passes wallet passphrase through migrator to an individual migration to decrypt/encrypt account data in migrations

updates 'start' command to catch EncryptedWalletMigrationError and prompt for wallet passphrase. passes wallet passphrase through node openDB flow to suppoprt running migrator with wallet passphrase

NOTE: this example assumes that the schema for the masterKey store does not change and assumes that the schema for encrypted account data does not change. if the masterKey schema changes then that change must be addressed in a separate migration. if the encrypted account schema changes, then that change should be addressed separately from any change to the decrypted account schema in a separate migration.

Closes IFL-3124

## Testing Plan
1. encrypt wallet on `staging` branch
2. change to this branch, run either `migrations:start` or `start --upgrade`

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/start.ts
```diff
@@ -223,7 +230,28 @@ export default class Start extends IronfishCommand {
...
+            true,
+            {
+              password: true,
+            },
+          )
+        } else {
+          throw e
+        }
+      }
+    }
```
hughy, line 254: Not thrilled about doing this in the `start` command or about passing the passphrase down through `NodeUtils`. Alternatively we could catch the error here and tell the user to run `ironfish migrations:start` and enter their passphrase there

rohanjadvani (approved):

ironfish-cli/src/commands/start.ts
NullSoldier: I would much prefer to do that than add the password flow here to start.

Mainly because if someone has their wallet encrypted unfortunately they just need to run migrations manually. I don't like the idea of adding interactive elements to the start command.

ironfish/src/node.ts
```diff
@@ -354,12 +354,16 @@ export class FullNode {
     return node
   }
 
-  async openDB(): Promise<void> {
+  async openDB(walletPassphrase?: string): Promise<void> {
```
NullSoldier, line 357: again, it sounds weird but I'd prefer to exclude this and make the error nicer.

hughy:
Closing this PR in favor of splitting into separate issues:
- Adding error handling and prompting logic to 'migrations' commands
- Separating account encryption/decryption logic from data models
- Adding example/template migration

PR #5630, opened 2024-11-08
stores identity in multisig account value

hughy:
## Summary

updates the MultisigKeys interface to store the multisig account identity in the account value

adds a migration, 033, that derives the account identity from the account secret already stored in the account

updates importAccount to derive identity from secret at time of import. deriving the identity instead of using the identity included in the import maintains backwards compatibility

Closes IFL-3118

## Testing Plan
1. export multisig account as JSON on staging, see that it does not contain identity
2. export multisig account as JSON on this branch, see that it does contain identity
3. import exports back into new datadirs on this branch, verify that imports succeed

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/migrations/data/033-multisig-keys-identity.ts
```diff
@@ -0,0 +1,172 @@
...
+
+        const masterKey = new MasterKey(masterKeyValue)
+        await masterKey.unlock(walletPassphrase)
+
+        // Decrypt encrypted account data
+        const decrypted = masterKey.decrypt(account.data, account.salt, account.nonce)
+        decryptedAccount = oldEncoding.deserializeDecrypted(decrypted)
+
+        // Apply migration to decrypted account data
+        logger.info(`  Migrating account ${decryptedAccount.name}`)
```
jowparks, line 65: This forking logic is unfortunate, but I can't think of a better way to do it for encrypted/unencrypted.

ironfish/src/migrations/data/033-multisig-keys-identity.ts
```diff
@@ -0,0 +1,172 @@
...
+        const migrated = this.accountBackward(decryptedAccount)
+
+        await stores.old.accounts.put(decryptedAccount.id, migrated, tx)
+      }
+    }
+  }
+
+  accountBackward(newValue: NewDecryptedAccountValue): OldDecryptedAccountValue {
+    const oldValue = newValue
```
jowparks, line 169: Is there no downside of keeping the identity around if the backward migration runs? I think it seems fine from a DB perspective.

ironfish/src/migrations/data/033-multisig-keys-identity/new/accountValue.ts
```diff
@@ -0,0 +1,229 @@
...
+export type AccountValue = EncryptedAccountValue | DecryptedAccountValue
+export class AccountValueEncoding implements IDatabaseEncoding<AccountValue> {
+  serialize(value: AccountValue): Buffer {
+    if (value.encrypted) {
+      return this.serializeEncrypted(value)
+    } else {
+      return this.serializeDecrypted(value)
+    }
+  }
+
```
jowparks, line 46: Do we need to do this for every account migration? Have separate encrypted/decrypted serde? That is brutal, but I think I remember you talking about this and finding it was required.

ironfish/src/migrations/data/033-multisig-keys-identity.ts
hughy: Since the value will be written using the old version of `MultisigKeysEncoding` it won't include the identity when it writes to the DB

ironfish/src/migrations/data/033-multisig-keys-identity/new/accountValue.ts
hughy: Yeah we have separate serde in the database now, so we will always need to handle both for migrations 😭

ironfish/src/migrations/data/033-multisig-keys-identity.ts
hughy: I couldn't think of a better way either

ironfish/src/wallet/walletdb/multisigKeys.ts
```diff
@@ -82,7 +86,7 @@ export function isSignerMultisig(multisigKeys: MultisigKeys): multisigKeys is Mu
 export function isHardwareSignerMultisig(
   multisigKeys: MultisigKeys,
 ): multisigKeys is MultisigHardwareSigner {
-  return 'identity' in multisigKeys
+  return 'identity' in multisigKeys && !('secret' in multisigKeys)
```
jowparks, line 89: Will this come back as true for view only standard multisig?

jowparks (approved):
Seems good, will be nice to see this tested in staging.

ironfish/src/wallet/walletdb/multisigKeys.ts
hughy: The standard multisig account will have the `secret` stored with its multisig keys, so this would come back as false

ironfish/src/migrations/data/033-multisig-keys-identity.ts
hughy: One alternative would be to refactor the way we encrypt accounts to encrypt the individual fields. If, say, the spending key and view keys were encrypted but still stored as separate fields then it might be easier to handle migration of an encrypted account in cases where we don't need access to the encrypted keys.

That's definitely a larger project though, and comes with additional considerations around how we treat encrypted data. Something to keep in mind for future work 😁

ironfish/src/wallet/wallet.ts
```diff
@@ -1510,15 +1510,20 @@ export class Wallet {
...
         }
         secret = multisigIdentity.secret
         identity = Buffer.from(accountValue.multisigKeys.identity, 'hex')
       } else if (isMultisigSignerImport(accountValue.multisigKeys)) {
         secret = Buffer.from(accountValue.multisigKeys.secret, 'hex')
+        // Derive identity from secret for backwards compatibility: legacy
+        // MultisigKeysImport may not include identity
         identity = new multisig.ParticipantSecret(secret).toIdentity().serialize()
+        multisigKeys = {
+          ...accountValue.multisigKeys,
```
NullSoldier, line 1524: I think this should just be `...multisigKeys` here to not over-ride any changes made between lines 1498 and this line. If someone else modifies multisigKeys it'll be lost here.

ironfish/src/wallet/wallet.ts
```diff
@@ -1510,15 +1510,20 @@ export class Wallet {
         }
 
         multisigKeys = {
-          keyPackage: accountValue.multisigKeys.keyPackage,
-          publicKeyPackage: accountValue.multisigKeys.publicKeyPackage,
+          ...accountValue.multisigKeys,
```
NullSoldier, line 1513: same here as below so we don't lose any changes to multisigKeys someone else may make in the future. We should only use multisigKeys and not accountValue after line 1498.

ironfish/src/wallet/walletdb/multisigKeys.ts
NullSoldier: This is a bug that exists previously but this should also be checking for `publicKeyPackage`.

hughy: Why should this check for `publicKeyPackage`? Since all variations of the `MultisigKeys` type have `publicKeyPackage` it doesn't help to discriminate between them

NullSoldier: If you don't it's possible to think this is a MultisigHardwareSigner but it's possible that another type that has identity, does not have secret, and has a different field than public key package will pass these checks and incorrectly be labeled a hardware signer. If someone adds a type like that, they'll need to track down this code as the cause of the bug.

hughy: Updated so that the discriminator functions each check all fields.

NullSoldier:
I tested this migration by creating a coordinator, 3 participants, and 1 normal single signer account.

I ran the migration and checked the data, I also encrypted my wallet and checked the migrations again.

All cases seem to work. I'm going to LGTM this PR but one more person should manually test the migrations.

NullSoldier (approved):

PR #5631, opened 2024-11-11
passes wallet passphrase from CLI to migrator

hughy:
## Summary

adds an error type, EncryptedWalletMigrationError, and catches errors of that type in the three commands that run migrations: 'migrations:start', 'migrations:revert', and 'start'

prompts user to enter wallet passphrase and passes passphrase through to migrator

to support passing the passphrase down to the migrator in the 'start' command we pass the passphrase through 'NodeUtils.waitForOpen' and the the node's 'openDB' method

supports implementing database migrations that handle encrypted wallet logic. for example, throwing the EncryptedWalletMigrationError on an encrypted account and using the wallet passphrase passed to 'migrate' to decrypt and re-encrypt the wallet

Closes IFL-3125

## Testing Plan
- tested separately with changes in #5624 

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish-cli/src/commands/start.ts
```diff
@@ -223,7 +229,19 @@ export default class Start extends IronfishCommand {
...
-    await NodeUtils.waitForOpen(node, () => this.closing)
+    try {
+      await NodeUtils.waitForOpen(node, () => this.closing)
+    } catch (e) {
+      if (e instanceof EncryptedWalletMigrationError) {
+        this.logger.error(e.message)
+        this.logger.error(
+          'Run `ironfish migrations:start` to enter wallet passphrase and migrate wallet databases',
+        )
+        this.exit(1)
```
rohanjadvani, line 240: Ok nice, this looks different from the original PR so the migrations will need to run separately.

rohanjadvani (approved):

ironfish-cli/src/commands/start.ts
hughy: Yeah I updated this flow to require migrations to run separately based on Jason's comment on the original PR: https://github.com/iron-fish/ironfish/pull/5624#discussion_r1837324957

PR #5649, opened 2024-11-13
Rahul/supress websocket connection reattempt

patnir:
## Summary

image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


ironfish/src/network/peers/peer.ts
```diff
@@ -294,7 +294,11 @@ export class Peer {
    */
   setWebSocketConnection(connection: WebSocketConnection): void {
     if (this.state.type !== 'DISCONNECTED' && this.state.connections.webSocket) {
-      this.logger.warn('Already have a WebSocket connection, ignoring the new one')
+      this.logger.debug(
+        `Already have a WebSocket connection with ${
+          connection.address?.host || 'Unknown host'
```
dguenther, line 299: Yeah I'm fine with bumping down this log level. I think given we've let the error ride for a while without investigating it, it'd be better to bump down to a debug to clean up our log output.

The only thing I'd suggest changing here is I think we should log this.displayName instead of connection.address.host, since peers are unique based on their identity rather than their network address. Or at least log something like "Peer {displayName} already has a websocket connection to {this.address}, ignoring new one at ${connection.address}"

patnir: Done!

dguenther (approved):

PR #5651, opened 2024-11-13
Add filter start and end support

NullSoldier:
## Summary

This lets you filter dates based on a date input such as 04/20/2023

## Testing Plan

```
ironfish wallet:transactions --no-account --filter.start 09/1/2024 --filter.end 10/01/2024
```

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/commands/wallet/transactions/index.ts
+++ b/ironfish-cli/src/commands/wallet/transactions/index.ts
@@ -63,11 +63,11 @@ export class TransactionsCommand extends IronfishCommand {
       helpGroup: 'OUTPUT',
     }),
     'filter.start': Flags.string({
-      description: 'include transactions after this date (inclusive). Example: 04/20/2023',
+      description: 'include transactions after this date (inclusive). Example: 2023-04-01',
       parse: (input) => Promise.resolve(new Date(input).toISOString()),
     }),
     'filter.end': Flags.string({
-      description: 'include transactions before this date (exclusive). Example: 05/20/2023',
+      description: 'include transactions before this date (exclusive). Example: 2023-05-01',
       parse: (input) => Promise.resolve(new Date(input).toISOString()),
     }),
   }
--- a/ironfish-cli/src/commands/wallet/transactions/index.ts
+++ b/ironfish-cli/src/commands/wallet/transactions/index.ts
@@ -62,6 +62,14 @@ export class TransactionsCommand extends IronfishCommand {
       options: ['notes', 'transactions', 'transfers'],
       helpGroup: 'OUTPUT',
     }),
+    'filter.start': Flags.string({
+      description: 'include transactions after this date (inclusive). Example: 04/20/2023',
+      parse: (input) => Promise.resolve(new Date(input).toISOString()),
+    }),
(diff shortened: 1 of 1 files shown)


ironfish-cli/src/commands/wallet/transactions/index.ts
```diff
@@ -62,6 +62,14 @@ export class TransactionsCommand extends IronfishCommand {
       options: ['notes', 'transactions', 'transfers'],
       helpGroup: 'OUTPUT',
     }),
+    'filter.start': Flags.string({
+      description: 'include transactions after this date (inclusive). Example: 04/20/2023',
```
andiflabs, line 66: Nit: I would change the example to an ISO date (to be more understandable to an international audience)

mat-if (approved):

ironfish-cli/src/commands/wallet/transactions/index.ts
```diff
@@ -62,6 +62,14 @@ export class TransactionsCommand extends IronfishCommand {
...
+      parse: (input) => Promise.resolve(new Date(input).toISOString()),
```
andiflabs, line 67: `input` gets converted to a Date, then to an ISO string, then back to a Date, then to an integer. I wonder if we could just remove the `.toISOString()` here and have this `parse` function return a Date?

andiflabs (approved):

ironfish-cli/src/commands/wallet/transactions/index.ts
```diff
@@ -62,6 +62,14 @@ export class TransactionsCommand extends IronfishCommand {
...
+    }),
+    'filter.end': Flags.string({
```
hughy, line 69: nit - should the names for these be more date-specific? `start` and `end` could also refer to sequences. I suppose we could handle that if/when we ever add filtering on sequence

hughy (approved):

ironfish-cli/src/commands/wallet/transactions/index.ts
NullSoldier: You need to create a new custom flag for this that supports dates. I can do this but I'll need to open a separate Pr that will then block this. I can start working on that now but it's outside the goal of this individual PR.

ironfish-cli/src/commands/wallet/transactions/index.ts
NullSoldier: Done!

ironfish-cli/src/commands/wallet/transactions/index.ts
NullSoldier: I'll work on adding this

andiflabs: No need to add this functionality as far as I'm concerned--I was just wondering if it was possible. I was assuming that `parse` could return any type

ironfish-cli/src/commands/wallet/transactions/index.ts
andiflabs: Suggestion: before/after

PR #5652, opened 2024-11-13
WalletScanner skips accounts with createdAt until createdAt sequence is reached

danield9tqh:
## Summary
When an account has the following 
createdAt: 10
head: null

the wallet scanner should not take any action on this account until it reaches block 10. Currently the wallet scanner will still connect blocks 1-10 (without decrypting transactions) but it still takes some time and it doesn't provide any value over just connecting block 10 directly

## Testing Plan
Unit tests + running local wallet tests

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/wallet/scanner/walletScanner.ts
+++ b/ironfish/src/wallet/scanner/walletScanner.ts
@@ -9,8 +9,8 @@ import type { HeadValue } from '../walletdb/headValue'
 import { Config } from '../../fileStores'
 import { Logger } from '../../logger'
 import { Mutex } from '../../mutex'
-import { BlockHeader, Transaction } from '../../primitives'
-import { BufferUtils, HashUtils } from '../../utils'
+import { BlockHeader, GENESIS_BLOCK_SEQUENCE, Transaction } from '../../primitives'
+import { HashUtils } from '../../utils'
 import { WorkerPool } from '../../workerPool'
 import { ChainProcessorWithTransactions } from './chainProcessorWithTransactions'
 import { BackgroundNoteDecryptor } from './noteDecryptor'
@@ -32,8 +32,13 @@ export class WalletScanner {
   /**
    * A snapshot of the accounts that have `scanningEnabled` set to true. Used
    * to tell what accounts should be scanned, and from what block.
+   * If `scanFrom` is 'always-scan' then the account's head is the same as the
+   * WalletScanner's head, and blocks should always be added
    */
-  private scanningAccounts = new Array<{ account: Account; scanFrom: HeadValue | null }>()
+  private scanningAccounts = new Array<{
+    account: Account
+    scanFrom: { head: HeadValue | null } | 'always-scan'
+  }>()
 
   constructor(options: {
     logger: Logger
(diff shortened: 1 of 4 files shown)


ironfish/src/wallet/scanner/walletScanner.ts
```diff
@@ -33,7 +33,10 @@ export class WalletScanner {
    * A snapshot of the accounts that have `scanningEnabled` set to true. Used
    * to tell what accounts should be scanned, and from what block.
    */
-  private scanningAccounts = new Array<{ account: Account; scanFrom: HeadValue | null }>()
+  private scanningAccounts = new Array<{
+    account: Account
+    scanFrom: { head: HeadValue | null } | null
```
andiflabs, line 38: The logic in this module is tricky sometimes. I haven't read the changes yet, but I know I will have a lot of trouble differentiating between `scanFrom === null` and `scanFrom.head === null`. Can we find a way to make the meaning of the values more intuitive? Maybe something like this (if it makes sense):

```
scanFrom: { head: HeadValue } | 'always-scan' | 'something-else'
```

hughy (commented):
Based on the description it seems like this should be getting rid of `connectOnlyAccounts` altogether -- if there's no reason to decrypt there's no reason to connect

ironfish/src/wallet/scanner/walletScanner.ts
danield9tqh: Yea I think that's a good idea. you're right that's the main thing I'm trying to differentiate with the change though

ironfish/src/wallet/scanner/walletScanner.ts
```diff
@@ -177,34 +182,32 @@ export class WalletScanner {
...
+      }
+
+      if (scanFrom.head === null) {
+        if (account.createdAt === null && blockHeader.sequence === GENESIS_BLOCK_SEQUENCE) {
+          candidate.scanFrom = 'always-scan'
+        }
+
+        if (account.createdAt && blockHeader.sequence >= account.createdAt.sequence) {
+          candidate.scanFrom = 'always-scan'
         }
```
andiflabs, line 199: Nit: I think this can be simplified as:

```ts
const createdAtSequence = account.createdAt?.sequence ?? GENESIS_BLOCK_SEQUENCE
if (blockHeader.sequence >= createdAtSequence) {
    candidate.scanFrom = 'always-scan'
}
```

andiflabs (approved):

ironfish/src/wallet/scanner/walletScanner.ts
```diff
@@ -177,34 +182,32 @@ export class WalletScanner {
...
+      if (scanFrom.head?.hash.equals(blockHeader.previousBlockHash)) {
+        candidate.scanFrom = 'always-scan'
       }
-      await this.wallet.connectBlockForAccount(account, blockHeader, [], false)
     }
 
+    const decryptAndConnectAccounts = this.scanningAccounts
+      .filter(({ scanFrom }) => scanFrom === 'always-scan')
+      .map(({ account }) => account)
+
```
hughy, line 210: It seems like we use `'always-scan'` as a way to to divide `scanningAccounts` into two lists: one list that the walletScanner is currently scanning for (the `always-scan` accounts), and a list of accounts that are waiting for the walletScanner to get to them.

I wonder if it would be easier to keep the accounts separate or if it would make it harder to follow?

danield9tqh: Might be easier but popping and pushing the accounts across arrays would be more verbose. I can look into something like this though

ironfish/src/wallet/scanner/walletScanner.ts
```diff
@@ -308,20 +312,40 @@ export class WalletScanner {
...
+        if (atSequence) {
+          return atSequence
+        }
+      } else {
+        return { hash: scanFrom.hash, sequence: scanFrom.sequence }
       }
     }
-    return earliestHead
+
+    return 'none'
```
hughy: Does `'none'` mean that all of the `scanFroms` are ahead of the chain head?

ironfish/src/wallet/scanner/walletScanner.ts
```diff
@@ -308,20 +312,40 @@ export class WalletScanner {
...
+      },
     )
   }
 
-  private getEarliestHead(): HeadValue | null {
-    let earliestHead = null
-    for (const { scanFrom: head } of this.scanningAccounts) {
-      if (!head) {
-        return null
+  private async getEarliestHead(): Promise<HeadValue | null | 'none'> {
```
hughy, line 324: Does `'none'` mean that all of the `scanFrom`s are ahead of the chain head?

hughy: Would it work to return the chain head in that case instead of the special `'none'` value?

ironfish/src/wallet/scanner/walletScanner.ts
```diff
@@ -33,7 +33,10 @@ export class WalletScanner {
    * A snapshot of the accounts that have `scanningEnabled` set to true. Used
    * to tell what accounts should be scanned, and from what block.
    */
-  private scanningAccounts = new Array<{ account: Account; scanFrom: HeadValue | null }>()
+  private scanningAccounts = new Array<{
+    account: Account
+    scanFrom: { sequence: number; hash?: Buffer } | 'cursor'
```
hughy, line 38: Does `'cursor'` mean that `scanFrom` is equal to the head of the wallet scanner?

ironfish/src/wallet/scanner/walletScanner.ts
danield9tqh: I think its both
1. earliest is ahead of the chain
2. there are no scanning accounts

danield9tqh: I'm not sure if returning the head would work for this case where we break https://github.com/iron-fish/ironfish/pull/5652/files#diff-4a43e4a9c484bea01d7248d51bc40948e7bbdd13c5bef3fdbfe0a02c7210f85dR138. If we set chain processor to head of the chain then maybe that's not the same as the chain processor's head (because it was set previously)

Idk its a good thought. I think I'll keep it as none for now though so don't have to think through / refactor that break statement

ironfish/src/wallet/scanner/walletScanner.ts
danield9tqh: yea, any thoughts on better name?

PR #5653, opened 2024-11-13
adds 'format' option to account import

hughy:
## Summary

adds a field to the wallet/importAccount RPC that allows clients to specify the format of the account (Base64Json, JSON, Mnemonic, or SpendingKey)

if format is specified in the request then the decoding is only attempted for the specified format

results in more specific error messages instead of an error message containing the errors for each encoder

## Testing Plan
- adds unit tests

## Documentation

```
[X] Yes
```

https://github.com/iron-fish/website/pull/755

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):
I manually tested this. With this change, I now get a better error which is "Invalid Mnemonic". We should remember to update the node app after this.

PR #5666, opened 2024-11-21
Account name for balance command is arg

patnir:
## Summary

Previously, it was a flag. This didn't conform to the standard we have for other commands like address.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


danield9tqh (approved):

hughy (commented):
I think `wallet:balance` and `wallet:balances` are borderline cases as far as the style guide is concerned with flags vs. args: https://github.com/iron-fish/ironfish/blob/6693fe5f913850867c0a3063f41ab4bff9891ab9/ironfish-cli/STYLE_GUIDE.md?plain=1#L24

I'm curious whether @mat-if and @NullSoldier considered making this an arg during the CLI audit?

NullSoldier:
Yeah, our stlye guide does indicate this should be a flag here to be consistent with our account commands. The only time the command should take an account as an arg is when the account name is the primary operand such as `ironfish wallet:use <account>`

PR #5667, opened 2024-11-22
handles Ledger 'invalid dkg status' errors

hughy:
## Summary

if a user attempts to sign a transaction, but their ledger app doesn't have any multisig keys persisted, then the signing operation will fail with an 'invalid dkg status' error. this can happen if the user reinstalls the app or if the user starts creating a new account on the device using bkg, but doesn't finish

the only ways to recover from this error are to restore an encrypted backup to the device, or to create a new multisig account using the device. since the error shows up during signing, we can infer that the user expects to have an account on the device and should be instructed to restore the backup

Closes IFL-3151

## Testing Plan
manual testing:
1. create multisig ledger account
2. create a transaction to sign using the multisig account(s)
3. reinstall the ironfish dkg app to clear the account keys from the device's storage
4. attempt to sign the transaction to produce the error

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/ledger/ledger.ts
+++ b/ironfish-cli/src/ledger/ledger.ts
@@ -26,6 +26,7 @@ export const IronfishLedgerStatusCodes = {
   PANIC: 0xe000,
   EXPERT_MODE_REQUIRED: 0x6984,
   DKG_EXPERT_MODE_REQUIRED: 0xb027,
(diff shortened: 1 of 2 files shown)


NullSoldier (approved):
I tested this manually and the new output is a LOT more clear what is happening.

PR #5668, opened 2024-11-22
do not retry ExitErrors in ui.retryStep

hughy:
## Summary

if the command is exiting due to user action (e.g., choosing not to continue) or an un-retryable problem then we should not prompt to retry

## Testing Plan
manual testing:
try to sign a transaction with ledger device locked. when prompted to unlock device and retry choose 'Cancel'

before:
image

after:
image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):
Tested manually, looks good.

PR #5672, opened 2024-11-25
optionally retries in cli commands if user rejects Ledger request

hughy:
## Summary

if a user rejects an action on their Ledger, the command will now ~exit~ ask the user to retry instead of automatically retrying and prompting the user again to approve the action

## Testing Plan

before:
image

after:
image

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/src/ui/ledger.ts
+++ b/ironfish-cli/src/ui/ledger.ts
@@ -72,47 +72,24 @@ export async function ledger<TResult>({
           // is trying to enter their pin. When we run into this error, we
           // cannot send any commands to the Ledger in the app's CLA.
           ux.action.stop('Ledger Locked')
-
-          const confirmed = await ui.confirmList(
+          await confirmRetryAction(
             'Ledger Locked. Unlock and press enter to retry:',
-            'Retry',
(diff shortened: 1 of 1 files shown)


patnir (approved):

NullSoldier:
I'm fine with this, but the reason we decided to retry is in case the user accidentally hits deny instead of accept and the pain of starting the process all over from scratch because someone accidentally rejected is a pretty painful UX experience. That being said, if you think this is a better UX I'm ok with it.

hughy:
> I'm fine with this, but the reason we decided to retry is in case the user accidentally hits deny instead of accept and the pain of starting the process all over from scratch because someone accidentally rejected is a pretty painful UX experience. That being said, if you think this is a better UX I'm ok with it.

I think a compromise here may be to prompt the user to retry instead of retrying automatically. That protects against accidentally rejecting a request, but avoids unexpected behavior if the user intentionally rejects the request.

NullSoldier (approved):
Tested this manually, it works really well. Also it reduces a lot of duplicated code.

PR #5692, opened 2024-12-06
deletes multisig identity on account import

hughy:
## Summary

since the identity will always be stored on the multisig identity we don't need to maintain the record of the identity separately in the multisigIdenitities store

keeping identities in this store can be confusing for future use when generating new accounts

deletes any existing identity that matches the identity on the imported account instead of adding an identity to the store if it is missing

removes DuplicateIdentityNameError, which is no longer thrown

removes tests that checked duplicate identity name logic on import

Closes IFL-3117

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


hughy:
We still rely on reading from the `multisigIdentities` datastore during the signing flow, so we shouldn't delete identities until the signing flow no longer relies on that datastore.

dguenther (approved):

PR #5694, opened 2024-12-11
adds support for fetching account tx by range

hughy:
## Summary

adds optional startSequence and endSequence request parameters to the 'wallet/getAccountTransactions' RPC to support fetching transactions by a range of block sequences (inclusive)

if startSequence is set but not endSequence, then all the maximum 32-bit integer value is used for endSequence

if endSequence is set but not startSequence, then the genesis block sequence (1) is used for startSequence

## Testing Plan
adds unit test

## Documentation

```
[X] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish/src/rpc/routes/wallet/getAccountTransactions.ts
+++ b/ironfish/src/rpc/routes/wallet/getAccountTransactions.ts
@@ -19,6 +19,8 @@ export type GetAccountTransactionsRequest = {
   account?: string
   hash?: string
   sequence?: number
+  startSequence?: number
+  endSequence?: number
   limit?: number
   offset?: number
   reverse?: boolean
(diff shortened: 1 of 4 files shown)


ironfish/src/rpc/routes/wallet/getAccountTransactions.ts
```diff
@@ -86,9 +90,17 @@ routes.register<typeof GetAccountTransactionsRequestSchema, GetAccountTransactio
...
-      ? account.getTransactionsBySequence(request.data.sequence)
-      : account.getTransactionsByTime(undefined, { reverse: request.data.reverse })
+    let transactions
+    if (request.data.startSequence || request.data.endSequence) {
+      transactions = account.getTransactionsBySequenceRange(
+        request.data.startSequence,
+        request.data.endSequence,
+      )
+    } else if (request.data.sequence) {
+      transactions = account.getTransactionsBySequence(request.data.sequence)
```
NullSoldier, line 100: I'm guessing we can just delete this now assuming `getTransactionsBySequenceRange` is inclusive which it looks like it is.

hughy: It is inclusive, so we can delete `getTransactionsBySequence`

NullSoldier (approved):

PR #5701, opened 2024-12-21
Added Ledger to Account Create, Import, Migration of Existing

j-s-n:
## Summary
Added ledger flag to newly created accounts (create, import, multisig, DKG) and migrated db.

## Testing Plan
Ledger imports were not workingon my laptop so will need to confirm those next.
1. Create wallet
	1. false hard coded
	2. tested
2. Import wallet
	1. ledger 
		1. false hard coded
		2. tested
	3. ledger flag
		1. true hard coded
		2. tested
	4. import file
		1. wallet
			1. ledger value decoded
			2. tested
		2. ledger
			1. ledger value decoded
			2. tested
		3. long prompt
			1. ledger value decoded
3. Multisig
	1. particpants wallets
		1. false hard coded
		2. tested
4. DKG
	1. wallet
		1. false hard coded
		2. tested
	2. ledger
		1. true hard coded
		2. tested
5. Edge Cases
	1. Migration
		1. old wallets add false ledger flag
	2. Old wallet imports
		1. decoding added conditional false ledger property if it doesnt exist

## Documentation
N/A
```

## Breaking Change
Don't think so. Ledger flag is only being added to new and existing entries in db.
```


ironfish/src/wallet/walletdb/multisigIdentityValue.ts
```diff
@@ -19,6 +19,7 @@ export interface MultisigIdentityValue {
    * while allowing distinction between undefined and actual secrets during deserialization.
    */
   secret?: Buffer
+  ledger: boolean
```
hughy, line 22: Since this field isn't optional we need a migration to update existing stored identities.

hughy: I think we may actually want to avoid adding `ledger` to this datastore. I think we should avoid importing just the identity from a ledger device without also importing a full account, and the goal of https://github.com/iron-fish/ironfish/pull/5692 is to not have multisig identities stored for imported accounts

ironfish-rust-nodejs/index.d.ts
```diff
@@ -320,6 +320,7 @@ export namespace multisig {
     incomingViewKey: string
     outgoingViewKey: string
     proofAuthorizingKey: string
+    ledger: boolean
```
NullSoldier, line 323: This is an unused change and will be deleted the next time this file is re-generated. This is auto-generated by NAPI-RS so you cannot modify it manually.

ironfish/src/wallet/walletdb/walletdb.ts
```diff
@@ -45,7 +45,7 @@ import { MultisigIdentityValue, MultisigIdentityValueEncoder } from './multisigI
 import { ParticipantIdentity, ParticipantIdentityEncoding } from './participantIdentity'
 import { TransactionValue, TransactionValueEncoding } from './transactionValue'
 
-const VERSION_DATABASE_ACCOUNTS = 32
+const VERSION_DATABASE_ACCOUNTS = 33
```
NullSoldier, line 48: This will need to be bumped and the migration needs to be renamed because this PR just merged in migration 33.

https://github.com/iron-fish/ironfish/pull/5630

ironfish/src/wallet/walletdb/accountValue.test.ts
```diff
@@ -30,6 +30,7 @@ describe('AccountValueEncoding', () => {
       },
       scanningEnabled: true,
       proofAuthorizingKey: key.proofAuthorizingKey,
+      ledger: false,
```
NullSoldier, line 33: maybe we should flip this to true in one of these tests to check the serialization is working.

ironfish-cli/src/commands/wallet/multisig/dkg/create.ts
```diff
@@ -529,6 +529,7 @@ export class DkgCreateCommand extends IronfishCommand {
       name: accountName,
       createdAt: null,
       spendingKey: null,
+      ledger: false,
```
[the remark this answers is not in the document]
j-s-n: thanks. i corrected this.

ironfish/src/migrations/data/033-add-ledger-to-accounts.ts
```diff
@@ -0,0 +1,169 @@
...
+  DecryptedAccountValue as NewDecryptedAccountValue,
+} from './033-add-ledger-to-accounts/new/accountValue'
+import {
+  AccountValueEncoding as OldAccountValueEncoding,
+  DecryptedAccountValue as OldDecryptedAccountValue,
+  EncryptedAccountValue as OldEncryptedAccountValue,
+} from './033-add-ledger-to-accounts/old/accountValue'
+import { GetStores } from './033-add-ledger-to-accounts/stores'
+
+export class Migration033 extends Migration {
```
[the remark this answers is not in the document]
j-s-n: moved all of the related files to migration 34, but will discuss if the migration logic will be implemented.

ironfish/src/migrations/data/033-add-ledger-to-accounts.ts
```diff
@@ -0,0 +1,169 @@
...
+        await stores.new.accounts.put(decryptedAccount.id, migrated, tx)
+      }
+    }
+  }
+
+  // Implement logic to migrate (decrypted) account data to the new schema
+  accountForward(oldValue: OldDecryptedAccountValue): NewDecryptedAccountValue {
+    const newValue = {
+      ...oldValue,
+      ledger: false,
```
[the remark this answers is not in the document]
j-s-n: will discuss if the migration logic will be implemented or if we can do a placeholder migration

ironfish/src/wallet/walletdb/walletdb.ts
j-s-n: corrected all references to 34 after 33 was merged

ironfish/src/wallet/walletdb/accountValue.test.ts
j-s-n: test passed with ledger as true

ironfish/src/migrations/data/034-add-ledger-to-accounts.ts
```diff
@@ -0,0 +1,169 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
danield9tqh, line 1: Does there need to be a database migration for this? It looks like we're adding the ledger flag to the `flags` field which is an unsigned 8 bit integer defaulted to 0. So if we are writing `false` for all the accounts in the migration then it would not be changing the database value at all right? So this migration wouldn't change anything? Maybe I'm missing something

j-s-n: Hugh and I discussed this, but haven't made any changes yet. It would still be functional without the migration since its adding the ledger to the flags field, but it might be helpful to have a placeholder migration for documentation for adding that change.

danield9tqh (approved):

PR #5705, opened 2025-01-17
getAccountsTransactions.test: Removed sort from streams back transactions for a given block sequence range

j-s-n:
## Summary
Removed the sort for the accountTransactions since it is already received in order. I regenerated the fixtures and the random hash lexographically sorted ended up swapping the order of the transactions.

## Testing Plan
I regenerated fixtures without sort and test passed

## Documentation

N/A

## Breaking Change

N/A

### DIFF
--- a/ironfish/src/rpc/routes/wallet/getAccountTransactions.test.ts
+++ b/ironfish/src/rpc/routes/wallet/getAccountTransactions.test.ts
@@ -114,9 +114,9 @@ describe('Route wallet/getAccountTransactions', () => {
       block3.transactions[0].hash(),
     ]
     const accountTransactions = await AsyncUtils.materialize(response.contentStream())
-    const accountTransactionHashes = accountTransactions
-      .map(({ hash }) => Buffer.from(hash, 'hex'))
-      .sort()
+    const accountTransactionHashes = accountTransactions.map(({ hash }) =>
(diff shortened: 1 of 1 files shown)


hughy (commented):
I'm wondering if we should sort both `accountTransactionHashes` and `blockTransactionHashes` before comparing. This test seems more concerned with all of the transactions being present than it is with all of them being in the same order as they are on the blocks.

j-s-n:
> I'm wondering if we should sort both `accountTransactionHashes` and `blockTransactionHashes` before comparing. This test seems more concerned with all of the transactions being present than it is with all of them being in the same order as they are on the blocks.

loadTransactionsBySequence range returns the transactions in the sequence order, so we can omit the sort since the hard coded blockTransactionHashes are also in sequence order.

hughy:
> > I'm wondering if we should sort both `accountTransactionHashes` and `blockTransactionHashes` before comparing. This test seems more concerned with all of the transactions being present than it is with all of them being in the same order as they are on the blocks.
> 
> loadTransactionsBySequence range returns the transactions in the sequence order, so we can omit the sort since the hard coded blockTransactionHashes are also in sequence order.

Sounds good to me!

hughy (approved):

PR #5709, opened 2025-01-21
account.test - getTransactionsByTime: added buffer comparison to sort

j-s-n:
## Summary
Original test was lexicographically sorting the hashed transactions, which could misorder the transactions based on their string representations. Added in sorting based on the buffer compare method to compare the bytes

## Testing Plan
Deleted and regenerated account.test.fixture and reran test 10 times.
Staging - failed
account.test - 10/10 passed

## Documentation

No

## Breaking Change

No


hughy (approved):

hughy (approved):
Looks good to me, but I think the comments above the changed lines about the sorting are incorrect

j-s-n:
> Looks good to me, but I think the comments above the changed lines about the sorting are incorrect

reverse is sorting by b compare a and chron sort is a compare b

PR #5711, opened 2025-01-22
reads identity from ledger in dkg round1

hughy:
## Summary

when running DKG using a Ledger device, the user's identity must match the identity from the Ledger device.

rather than reading an identity from the node based on a name we can read the identity directly from the Ledger.

reads identity from Ledger in 'wallet:multisig:dkg:round1' and 'wallet:multisig:dkg:create' instead of reading the identity from the node

## Testing Plan
- created multisig account using ledger

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```


dguenther (approved):
looks good!

PR #5731, opened 2025-02-12
Add methods to ProposedTransaction and description builders to build zk circuits

andiflabs:
## Summary

Needed by Oreowallet.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust/src/transaction/outputs.rs
+++ b/ironfish-rust/src/transaction/outputs.rs
@@ -74,6 +74,24 @@ impl OutputBuilder {
         ExtendedPoint::from(self.value_commitment.commitment())
     }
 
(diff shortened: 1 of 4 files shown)


ironfish-rust/src/transaction/outputs.rs
```diff
@@ -74,6 +74,24 @@ impl OutputBuilder {
         ExtendedPoint::from(self.value_commitment.commitment())
     }
 
+    pub(crate) fn build_circuit(
```
patnir, line 77: should (crate) be removed here too?

patnir (approved):

PR #5743, opened 2025-02-18
Update wasm README with deployment instructions

patnir:
## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-rust-wasm/README.md
+++ b/ironfish-rust-wasm/README.md
@@ -20,11 +20,6 @@ This package provides WebAssembly bindings for Iron Fish's core cryptographic fu
 - High-performance cryptographic operations
 - Cross-platform compatibility
 
-## Deployment Instructions 
-
-1. Create a PR to bump the version in `Cargo.toml` for the `staging` branch. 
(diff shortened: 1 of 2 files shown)


ironfish-rust-wasm/README.md
```diff
@@ -20,6 +20,11 @@ This package provides WebAssembly bindings for Iron Fish's core cryptographic fu
 - High-performance cryptographic operations
 - Cross-platform compatibility
 
+## Deployment Instructions 
+
+1. Create a PR to bump the version in `Cargo.toml` for the `staging` branch. 
+2. After merging the PR, run the following Github Actions workflow: https://github.com/iron-fish/ironfish/actions/workflows/deploy-npm-ironfish-rust-wasm.yml
```
andiflabs, line 26: LGTM, but I'd like to point out that this will be visible to all people on https://www.npmjs.com/package/@ironfish/ironfish-wasm, which may not be what we want. Some projects use a separate `DEVELOPMENT.md` file (or similar) for this purpose

andiflabs (approved):

PR #5753, opened 2025-03-04
Fix docker builds by pinning rust toolchain

NullSoldier:
## Summary

This pins the rust version to our current toolchain of 1.76.0 because the sh.rustup.rs script automatically installs stable. This means if stable drifts from our toolchain version it will automatically start failing new builds.

## Testing Plan

Just run `RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain=1.76.0` with different versions, or excluded.

## Documentation

```
[ ] Yes
```

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/ironfish-cli/Dockerfile
+++ b/ironfish-cli/Dockerfile
@@ -6,7 +6,7 @@ RUN \
     apt-get update && \
     apt-get install jq rsync -y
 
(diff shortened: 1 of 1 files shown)


patnir (approved):

leanthebean (approved):
LGTM

